PR #14, opened 2018-08-15
Resurrect old notes around class modifiers.

munificent:
Leaf mentioned that it might be useful to bring these notes back from the old evolution repo, so here they are. I didn't commit them directly because I wasn't sure if this was where you wanted and I wanted to make sure that Erik was OK with me putting his words out there.

What do you think?

### DIFF
--- a/working/class-capabilities/class-capabilities.md
+++ b/working/class-capabilities/class-capabilities.md
@@ -1,0 +1,164 @@
+With the in-progress [explicit mixin declaration syntax][6] and experiments
+around [sealed classes][11], we are inching towards a world where users have
+syntax to directly control the capabilities a class exposes. If we're going to
+go there, it might be worth considering the full set of capabilities and what
+keywords/modifiers we might use for them.
+
+This is *not* a proposal. It's just a note to capture some earlier informal
+discussions. *If* we decide to go in this direction, it might be useful source
+material.
+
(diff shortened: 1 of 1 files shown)


yjbanov:
What a fantastic document! Creates a solid framework for thinking about class capabilities. However, I couldn't tell what the resolution is for the "can extend + can implement + private methods" issue. If I have control over my class' capabilities, then the compiler should flag the case when my class allows `implements` but makes it impossible to actually implement it due to private methods accessed from outside the class.

eernstg:
LGTM!

lrhn:
I'm split on whether I'd prefer this as an *issue* to encourage discussion, instead of a dead document that is stored by itself.

Discuss! :)

munificent:
> If I have control over my class' capabilities, then the compiler should flag the case when my class allows implements but makes it impossible to actually implement it due to private methods accessed from outside the class.

Ah, that's interesting. It would be nice for the tools to give better guidance here.

> I'm split on whether I'd prefer this as an issue to encourage discussion, instead of a dead document that is stored by itself.

I'm going to go ahead and land it if that's OK. I think it's a non-goal to start a live discussion on this because, as far as I know, we aren't currently planning to take action on it. When/if we do, we can start a thread then and point to the doc.

Issue #11, opened 2018-08-13
Experiment - Class annotation for disallowing use as a super-type, a.k.a "sealed classes"


srawlins:
Users have expressed a desire for the ability to mark a class as "final" or "sealed," so that it is unavailable as a super-class. That is, it is an error for a class to extend, implement, or mix in a "sealed class." Specifically, the Angular Dart team wishes to seal certain classes that have very undefined behavior when users subclass.

There is [good discussion](https://github.com/dart-lang/sdk/issues/27372) on an MVP, make-something-cheap-available-to-users request for a `@sealed` annotation.

## An experiment

All that is being suggested here, after some discussion <sup>1</sup>, is an annotation in the [meta package](https://pub.dartlang.org/packages/meta), `@sealed`, enforced with some analyzer Hint codes.

## Use cases

The primary use case is to _remove burden from_, and _give assurance to_, authors of public classes. Today, as all classes are "open," there is an expectation from users that they can extend, implement, and mix in classes from other packages safely. For this to be true, authors must actually have extensions, implementations, and mix ins in mind, unless they write "DO NOT EXTEND/IMPLEMENT" in their Documentation comments.

A "sealed classes" feature can _remove burden from_ authors by allowing for a class that doesn't need to support the possibility that other classes use it as a super-class. Authors can write a class

* whose methods call other public methods,
* whose methods refer to private members,

without worrying about how to support sub-classes.

A "sealed classes" feature can _give assurance to_ an author, when considering how a change to the class will affect users. An author can:

* change a method to refer to a public method,
* add public methods,

without worrying about how the change will affect potential sub-classes.

## Definition

A sealed class, one which has been annotated with `@sealed` from the meta package, can only be extended by, implemented by, or mixed into a class declared in the same "package" (to be defined) as the sealed class. The consequences of "sealing" a class are primarily in static analysis, and would be implemented in the analyzer package:

* It is a Hint if the `@sealed` annotation occurs on anything other than a class.
* Given a class _C_ declared in a package _P_,
  * It is a Hint if a class which is declared in a package other than _P_ extends, implements, or mixes in _C_.
  * It is a Hint if a class, declared in _P_, and either extending, implementing, or mixing in _C_, is not also annotated with `@sealed`.

### Why a Hint?

[...]

PR #17, opened 2018-08-16
Add a working introduction to Bazel.

matanlurey:
I've added this as a working reference for future language enhancements or experiments that intend to tackle the concept of either packages-in-the-language, modules-in-the-language, visibility restrictions, "friends", etc - i.e. such as https://github.com/dart-lang/language/issues/11.

### DIFF
--- a/working/bazel/module-structure.md
+++ b/working/bazel/module-structure.md
@@ -1,4 +1,4 @@
-The following is a reference for [`Bazel`](https://bazel.build/) (internal
+The following is a reference for [`Bazel`](https://bazel.build/) (internally
 called `Blaze`) structures modules of Dart code, especially when compared to
 external structures (i.e. via `pub`).
 
@@ -57,7 +57,8 @@ Once _inside_ a package, such as `//ninjacat/app`, you can expect the following:
     > BUILD
 ```
 
-(If the package happens to be a Dart web _entrypoint_, you might also see `web/`.)
+(If the package happens to be a Dart web _entrypoint_, you might also see `web/`
+and for VM binaries, you might also see `bin/`.)
 
 However, there is another important concept, [`targets`][4]:
 
@@ -94,7 +95,9 @@ dart_test(
 ```
 
 Here we have _3_ targets:
-* `app`, which potentially is the entrypoint to our package/application.
+* `app`, which potentially is code that wraps together application-specific
+  code before being used later in the `main()` function of something in either
+  `web/` or `bin/`.
 * `flags`, which contains some common code for setting/getting flags.
 * `flags_test`, which tests that `flags` is working-as-intended.
 
@@ -224,3 +227,8 @@ dart_library(
     ],
 )
 ```
+
+Unfortunately this is quite common when you take into account the concept of
+pub's `dev_dependencies: ...`. If you have a testing only package (
+`angular_test`, `build_test`, `flutter_test`), which depends on the main package
+but is also used within the main packages tests you have a cyclic dependency.
--- a/working/bazel/module-structure.md
+++ b/working/bazel/module-structure.md
@@ -1,6 +1,6 @@
 The following is a reference for [`Bazel`](https://bazel.build/) (internally
-called `Blaze`) structures modules of Dart code, especially when compared to
-external structures (i.e. via `pub`).
+called `Blaze` at Google) structures modules of Dart code, especially when
+compared to external structures (i.e. via `pub`).
 
 This is *not* a proposal, nor a commitment to build out further Bazel support
 externally, but rather a reference to help guide discussions around visibility
@@ -36,18 +36,19 @@ Bazel [defines a _package_][2] as so:
 
 [2]: https://docs.bazel.build/versions/master/build-ref.html#packages_targets
 
-As one can see, already _package_ is very different than a [`pub` package][3].
+As one can see, already _package_ is very different from a [`pub` package][3].
 
 [3]: https://www.dartlang.org/guides/libraries/create-library-packages
 
 We map the concept of `pub` packages with the following rules:
 
 * A `package:<a>.<b>/uri.dart` resolves to `//a/b/lib/uri.dart`.
-* Any other `package:<name>/uri.dart` resolves to `//third_party/dart/<name>`.
+* Any other `package:<name>/uri.dart` resolves to `//third_party/dart/<name>`,
+  i.e. without a `.` in the name.
 
 ### Inside a Package
 
-Once _inside_ a package, such as `//ninjacat/app`, you can expect the following:
+Inside a package, such as `//ninjacat/app`, you can expect the following:
 
 ```
 > ninjacat/
@@ -102,7 +103,7 @@ Here we have _3_ targets:
 * `flags_test`, which tests that `flags` is working-as-intended.
 
 This concept already is quite different than a `pub` package, where all of the
-files in `lib/` are accessible once y ou have a dependency on that package. In
+files in `lib/` are accessible once you have a dependency on that package. In
 fact, a common issue externally is that `pubspec.yaml` (sort of similar to
 `BUILD`) is not granular enough, leading to the creation of "micro packages" 
 that have a single file or capability.
--- a/working/bazel/module-structure.md
+++ b/working/bazel/module-structure.md
@@ -1,0 +1,226 @@
+The following is a reference for [`Bazel`](https://bazel.build/) (internal
+called `Blaze`) structures modules of Dart code, especially when compared to
+external structures (i.e. via `pub`).
+
+This is *not* a proposal, nor a commitment to build out further Bazel support
+externally, but rather a reference to help guide discussions around visibility
+modifiers and terminology around modular/reusable Dart code.
+
+## Monolothic Repository
+
+In ["Why Google Stores Billions of Lines of Code in a Single Repository"][1]
+the author explain the benefits of a _Monolothic Repository_ ("Mono Repo").
+Many of the constraints described in this document are due to this decision,
+and be expected to be _unchangeable_ - i.e. it is quite outside of the scope
+of any language (including Dart) to change these requirements.
+
+[1]: https://cacm.acm.org/magazines/2016/7/204032-why-google-stores-billions-of-lines-of-code-in-a-single-repository/fulltext
+
+## Typical Structure
+
+In the Google mono-repository, there is a single root directory. Let's call it
+`google` - and it is commonly referenced by tools and Bazel with a prefix of
+`//`, so `//ninjacat` is logically located at `/google/ninjacat`. This root
+directory is called a _workspace_.
+
+Bazel [defines a _package_][2] as so:
+
+> The primary unit of code organization in a workspace is the _package_. A
+> package is a collection of related files and a specification of the
+> dependencies among them.
+>
+> A package is defined as a directory containing a file named `BUILD`,
+> residing beneath the top-level directory in the workspace. A package includes
+> all files in its directory, plus all subdirectories beneath it, except those
+> which themselves contain a BUILD file.
+
+[2]: https://docs.bazel.build/versions/master/build-ref.html#packages_targets
+
+As one can see, already _package_ is very different than a [`pub` package][3].
+
+[3]: https://www.dartlang.org/guides/libraries/create-library-packages
+
+We map the concept of `pub` packages with the following rules:
+
+* A `package:<a>.<b>/uri.dart` resolves to `//a/b/lib/uri.dart`.
+* Any other `package:<name>/uri.dart` resolves to `//third_party/dart/<name>`.
+
+### Inside a Package
+
+Once _inside_ a package, such as `//ninjacat/app`, you can expect the following:
+
+```
+> ninjacat/
+  > app/
+    > lib/
+    > test/
+    > BUILD
+```
+
+(If the package happens to be a Dart web _entrypoint_, you might also see `web/`.)
+
+However, there is another important concept, [`targets`][4]:
+
+> A package is a container. The elements of a package are called _targets_. Most
+> targets are one of two principal kinds, files and rules. Additionally, there
+> is another kind of target, package groups, but they are far less numerous.
+
+[4]: https://docs.bazel.build/versions/master/build-ref.html#targets
+
+So, imagine the following, in the `BUILD` file:
+
+```
+dart_library(
+    name = "app",
+    srcs = ["lib/app.dart"],
+    deps = [
+         ":flags",
+    ],
+)
+
+dart_library(
+    name = "flags",
+    srcs = ["lib/flags.dart"],
+)
+
+dart_test(
+    name = "flags_test",
+    srcs = ["test/flags_test.dart"],
+    deps = [
+        ":flags",
+        "//third_party/dart/test",
+    ],
+)
+```
+
+Here we have _3_ targets:
+* `app`, which potentially is the entrypoint to our package/application.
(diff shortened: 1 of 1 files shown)


working/bazel/module-structure.md
```diff
@@ -0,0 +1,226 @@
...
+    deps = [
+        "//third_party/dart/pageloader",
+    ],
+)
+```
+
+In this case, `package:ninjacat.common.widgets.login.testing` exposes a set of
+test-only libraries that can be used, only in _test_ targets. This pattern isn't
+replicable externally at all, but is _very_ common internally to avoid bringing
+in test utilities and code into production applications.
```
jakemac53: Interesting insight - I could imagine a `dev_only` flag which could be set which only allows a package to be used as a dev dependency or as a direct dependency of other `dev_only` packages - this would be non-breaking for pub (but a breaking change per package when they opt in).

working/bazel/module-structure.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Notable Differences
+
+### No cyclical _targets_
```
jakemac53: It might be worth mentioning that this is actually quite common when you take into account dev dependencies. If you have a testing only package (angular_test, build_test), which depends on the main package but is also used within the main packages tests (very common) now you have a cyclic dependency.

matanlurey: Done.

working/bazel/module-structure.md
```diff
@@ -0,0 +1,226 @@
...
+
+```
+> ninjacat/
+  > app/
+    > lib/
+    > test/
+    > BUILD
+```
+
+(If the package happens to be a Dart web _entrypoint_, you might also see `web/`.)
```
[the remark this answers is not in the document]
matanlurey: Done

working/bazel/module-structure.md
```diff
@@ -0,0 +1,226 @@
...
+[2]: https://docs.bazel.build/versions/master/build-ref.html#packages_targets
+
+As one can see, already _package_ is very different than a [`pub` package][3].
+
+[3]: https://www.dartlang.org/guides/libraries/create-library-packages
+
+We map the concept of `pub` packages with the following rules:
+
+* A `package:<a>.<b>/uri.dart` resolves to `//a/b/lib/uri.dart`.
+* Any other `package:<name>/uri.dart` resolves to `//third_party/dart/<name>`.
```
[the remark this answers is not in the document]
matanlurey: Ack

working/bazel/module-structure.md
```diff
@@ -0,0 +1,226 @@
+The following is a reference for [`Bazel`](https://bazel.build/) (internal
```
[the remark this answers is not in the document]
matanlurey: Done

working/bazel/module-structure.md
matanlurey: Ack

working/bazel/module-structure.md
```diff
@@ -0,0 +1,226 @@
...
+    srcs = ["test/flags_test.dart"],
+    deps = [
+        ":flags",
+        "//third_party/dart/test",
+    ],
+)
+```
+
+Here we have _3_ targets:
+* `app`, which potentially is the entrypoint to our package/application.
```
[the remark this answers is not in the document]
matanlurey: Clarified a bit, thanks

matanlurey:
/cc @leafpetersen @lrhn @munificent Is this useful enough to check-in as-is?

Is there anything else you _wish_ you knew (or could reference) that could be added on a follow-on?

working/bazel/module-structure.md
```diff
@@ -0,0 +1,234 @@
...
+  code before being used later in the `main()` function of something in either
+  `web/` or `bin/`.
+* `flags`, which contains some common code for setting/getting flags.
+* `flags_test`, which tests that `flags` is working-as-intended.
+
+This concept already is very different than a `pub` package, where all of the
+files in `lib/` are sort of a single "implicit" target. In fact, a common issue
+externally is that `pubspec.yaml` (sort of similar to `BUILD`) is not granular
+enough, leading to the creation of "micro packages" that have a single file or
+capability.
```
munificent: I tend to think of those micro packages as an anti-pattern. :( They can be useful if you *really* need to have different subsets of dependencies, but otherwise they just make maintenance and consumption of those packages harder.

working/bazel/module-structure.md
```diff
@@ -0,0 +1,234 @@
...
+test-only libraries that can be used, only in _test_ targets. This pattern isn't
+replicable externally at all, but is _very_ common internally to avoid bringing
+in test utilities and code into production applications.
+
+## Notable Differences
+
+### No cyclical _targets_
+
+Dart, as the language, allows cyclical dependencies between `.dart` files 
+libraries). Bazel does _not_ allow cyclical dependencies between packages
```
munificent: Nit: "(" before "libraries".

working/bazel/module-structure.md
```diff
@@ -0,0 +1,234 @@
...
+dart_library(
+    name = "bar",
+    srcs = ["lib/bar.dart"],
+    dependencies = [
+        "//ninjacat/common/foo",
+    ],
+)
+```
+
+Unfortunately this is quite common when you take into account the concept of
```
munificent: It's only unfortunate if you want to treat an entire pub package as a single unit of compilation/modularity, which was never the goal.

A pub package's granularity is the unit of "shippable API product". It's a version manager, not a build tool. Finding a pub package lacking as the unit of compilation is like complaining that you can't recline an entire car. That's not the point. It's a container around things that can be reclined. :)

working/bazel/module-structure.md
matanlurey: I think the most common cases, so far, which we haven't been able to help with:

* `{name}_test` packages (see below, basically avoiding a dependency on `package:test`)
* X-plat packages (i.e. a different implementation for web, I/O, flutter) might want different deps

working/bazel/module-structure.md
```diff
@@ -0,0 +1,234 @@
...
+
+Here we have _3_ targets:
+* `app`, which potentially is code that wraps together application-specific
+  code before being used later in the `main()` function of something in either
+  `web/` or `bin/`.
+* `flags`, which contains some common code for setting/getting flags.
+* `flags_test`, which tests that `flags` is working-as-intended.
+
+This concept already is very different than a `pub` package, where all of the
+files in `lib/` are sort of a single "implicit" target. In fact, a common issue
```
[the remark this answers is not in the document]
matanlurey: I clarified the wording a bit to explain what I actually meant :)

working/bazel/module-structure.md
matanlurey: Done.

working/bazel/module-structure.md
matanlurey: But you can't say micro packages are an anti-pattern _and_ say `pubspec.yaml` is only about version solving :-/

working/bazel/module-structure.md
munificent: I understand the concerns here, but my impression is that fears around dependencies are somewhat overblown. Just depend on the union of the packages that the various platforms need. In almost all cases, the version solver should be able to handle them.

I know that hasn't always worked out well in practice, largely because the analyzer went through a period of very enthusiastic breaking version releases and many things transitively depend on it. But that's settled down and things should work pretty reliably now. Also, @nex3's new version solver is much better at handling these cases

working/bazel/module-structure.md
munificent: I think what we probably want are three levels of granularity:

1. **Shippable versioned API product.** These have names, release schedules, dependencies on other products. Cycles are allowed. Pub packages.

2. **Build target.** This is the unit of compilation and modularity. Likely the unit of encapsulation for things like "sealed". Possibly having a privacy control at this level. Acyclic. This doesn't currently exist in a user-facing way.

3. **Source file.** A single file of source code with its own top-level scope. Privacy and namespacing works at this level. Dart libraries. Cycles are allowed (and heavily used in practice).

working/bazel/module-structure.md
jakemac53: Now that a build system of some sort is essentially required for all Dart implementations, we should be a lot more careful about huge sets of extra dependencies.

The entire package graph and all the sources of all packages has to be considered during the build, and this can cause considerable bloat for things like DDC/Kernel Builders which operate on all files in all packages. Even if they never actually run because those files aren't imported, we still have to set up the asset graph such that they _could_ run on anything, and some initialization and finalizations steps do have to iterate all _potential_ nodes in the graph.

working/bazel/module-structure.md
jakemac53: The levels of granularity you describe @munificent are exactly what we have in bazel, with the only difference being that packages are defined as directories with BUILD files, instead of coming from pub. But conceptually they are the same.

What @matanlurey and probably most people who have tried to integrate with a Dart compiler wants/needs, is a formal and consistent way of defining the **Build target** portion externally. I believe this should be integrated with pub, for a couple reasons:

- You need to have separate dependencies per build target in order to efficiently create an acyclic graph of build targets. 
  - The only alternatives I can come up with are constantly re-analyzing your actual Dart import structure to create custom targets (what package:build does, at significant expense), or forcing all end users to create static configuration for all their _transitive_ dependencies, which is just not realistic.
- Allows the build target structure to be statically configured, which is a huge win for build systems. Neither bazel nor package:build_runner allow you to read files before declaring your inputs/outputs, so this is a huge pain point today.
- Module granularity/structure is consistent across tools.
- No extra dependencies for platforms/features you don't use.
- Bazel has proven that configuring build targets in the package itself works well (enough). If you need a more granular target you can always send a PR to a package to create it. See my next point for where this potentially falls down externally though...

All this being said there is one potentially  _huge_ issue with trying to push the bazel model on the external world, which is figuring out how to actually enforce the acyclic requirement of the build targets. With bazel and a single version restriction for packages, this is not really an issue. But for pub packages any new non-breaking release of a package could introduce a build target cycle for any package that depends on that package accidentally. Maybe the version solver could take that into account (throw out versions if they introduce a build target cycle?), but it doesn't seem trivial.

matanlurey: Correct. (We should probably add a doc about build system constraints somewhere)

working/bazel/module-structure.md
```diff
@@ -0,0 +1,234 @@
+The following is a reference for [`Bazel`](https://bazel.build/) (internally
```
lrhn: internally -> internally in Google

(This repository is not aimed at Googlers only).
Consider whether it's important to external readers at all, it probably isn't.

Much of the document seems to be about Google-only design choices, so maybe instead make it explicit that this document is mainly about *Blaze at Google*, but that it also applies to Bazel build systems.

working/bazel/module-structure.md
```diff
@@ -0,0 +1,234 @@
...
+called `Blaze`) structures modules of Dart code, especially when compared to
+external structures (i.e. via `pub`).
+
+This is *not* a proposal, nor a commitment to build out further Bazel support
+externally, but rather a reference to help guide discussions around visibility
+modifiers and terminology around modular/reusable Dart code.
+
```
lrhn: Who is this document intended for?
That is, who would care about Bazel modules at all? And why should they? (So, I guess I'm asking for motivation - to make Bazel compatibility visible in language discussions - and expected audience to be spelled out).

I'm guess it is for people like me, who has no working knowledge of Bazel or Blaze, but who work with Dart pub and language packages all day long, and who might affect how well future language changes matches the Bazel model.

working/bazel/module-structure.md
```diff
@@ -0,0 +1,234 @@
...
+## Monolothic Repository
+
+In ["Why Google Stores Billions of Lines of Code in a Single Repository"][1]
+the author explain the benefits of a _Monolothic Repository_ ("Mono Repo").
+Many of the constraints described in this document are due to this decision,
+and be expected to be _unchangeable_ - i.e. it is quite outside of the scope
+of any language (including Dart) to change these requirements.
+
+[1]: https://cacm.acm.org/magazines/2016/7/204032-why-google-stores-billions-of-lines-of-code-in-a-single-repository/fulltext
+
```
lrhn: Does this mean that Bazel use is a Google-only requirement that nobody else should worry about? (Again, who is the target audience?)

working/bazel/module-structure.md
```diff
@@ -0,0 +1,234 @@
...
+[2]: https://docs.bazel.build/versions/master/build-ref.html#packages_targets
+
+As one can see, already _package_ is very different than a [`pub` package][3].
+
+[3]: https://www.dartlang.org/guides/libraries/create-library-packages
+
+We map the concept of `pub` packages with the following rules:
+
+* A `package:<a>.<b>/uri.dart` resolves to `//a/b/lib/uri.dart`.
+* Any other `package:<name>/uri.dart` resolves to `//third_party/dart/<name>`.
```
lrhn: Does "any other" here mean a package with no `.` in the name?

working/bazel/module-structure.md
```diff
@@ -0,0 +1,234 @@
...
+
+### Inside a Package
+
+Once _inside_ a package, such as `//ninjacat/app`, you can expect the following:
```
lrhn: Drop "Once". It confuses me :)

working/bazel/module-structure.md
```diff
@@ -0,0 +1,234 @@
...
+```
+
+Here we have _3_ targets:
+* `app`, which potentially is code that wraps together application-specific
+  code before being used later in the `main()` function of something in either
+  `web/` or `bin/`.
+* `flags`, which contains some common code for setting/getting flags.
+* `flags_test`, which tests that `flags` is working-as-intended.
+
+This concept already is quite different than a `pub` package, where all of the
```
lrhn: than -> from
?

working/bazel/module-structure.md
```diff
@@ -0,0 +1,234 @@
...
+files in `lib/` are accessible once y ou have a dependency on that package. In
```
lrhn: spurious space in "y ou"

lrhn:
I have no issue with a document like this if the purpose and scope is clear. 
I'm guessing it should be scoped as background material for discussions, aimed at people with no experience with Blaze at Google or Bazel build systems, because we do see it as a goal to remain *compatible* with those.

working/bazel/module-structure.md
matanlurey: Done.

(Bazel is currently Google only for the most part, but that is _rapidly_ changing. For example SpaceX uses it as their own build system, and many other medium/larger companies are considering it)

working/bazel/module-structure.md
matanlurey: Done.

working/bazel/module-structure.md
matanlurey: Done.

working/bazel/module-structure.md
matanlurey: Yes. Clarified.

working/bazel/module-structure.md
matanlurey: No. Clarified.

working/bazel/module-structure.md
matanlurey: Done.

matanlurey:
re: @lrhn's comments, I am taking that is an OK to merge :)

Happy to make more edits post-commit.

lrhn:
I'm not sure the *location* is great. I see other PRs putting such files into a "background" directory.
We can move it when we decide on a repo structure for non-language-feature documents.

PR #26, opened 2018-09-13
Create `resources` directory and move background files from `working` to it.

lrhn:


lrhn:
I think "resources" is better.

Issue #47, opened 2018-10-15
Spread Collections


munificent:
Solution for #46.

[Feature specification](https://github.com/dart-lang/language/blob/master/accepted/future-releases/spread-collections/feature-specification.md).

Most other languages have a "spread" or "splat" syntax that interpolates the elements of an existing collection into a new collection. In JS, it's a prefix `...`. In Ruby, Python, and a couple of others, it's a prefix `*`.

I propose we follow JS and use `...`.

This proposal is now accepted. Implementation work is being tracked in #164.

zoechi:
Do you think there is place in this proposal to also consider
https://github.com/flutter/flutter/issues/17862
where null values are dropped when spread?

munificent:
I don't think it's a good idea to silently discard null values. That might be the right thing for some APIs, but others may consider null values to be useful and meaningful. If the spread syntax always discarded them, it means you could never use a spread with those APIs.

zoechi:
I meant perhaps something like null-aware operators combined with spread to ignore `null` values, not ignoring them in general. 
Just thought it might make sense to consider this use case when designing the spread feature.

zoechi:
I guess an extension method/getter like `someList.nonNullValues` would do anyway

cbazza:
Just keep in mind that when designing a spread operator it goes hand in hand with the syntactically identical rest operator which goes hand in hand with destructuring syntax ;)

munificent:
Yes, the proposal doesn't mention that, but it was designed with destructuring and rest parameters in mind.

munificent:
The proposal for this has now landed: https://github.com/dart-lang/language/blob/master/working/0047.%20Spread%20Collections/spread-collections.md

alorenzen:
AngularDart makes heavy use of const lists as part of our configuration for the template compiler. In order to support items stored in another const list, we allow a "recursive" union type:
`List<T | List>`

According to your proposal, `Spread elements are not allowed in const lists or maps.`

Could this be adjusted to allow const elements to be spread in const lists?

For example, 
```
const foo = [Foo];

static bar = [Bar];

const containingFoo = [
  ...foo, // Allowed
  ...bar, // Not allowed
];
```

xster:
I don't know if it's the right protocol but what @zoechi pointed out is a common Flutter issue (though it may need an orthogonal category of solution). Branching that thread to #62

nshahan:
[...]

PR #48, opened 2018-10-15
Add proposal for "Spread Collections".

munificent:
Here it is!

### DIFF
--- a/working/0047. Spread Collections/spread-collections.md
+++ b/working/0047. Spread Collections/spread-collections.md
@@ -220,9 +220,12 @@ var ints = <int>[...numbers];
 ```
 
 This works because the individual elements in `numbers` do happen to have the
-right type even though the list that contains them does not. However, the spread
-object does need to be "spreadable"&mdash;it must be some kind of Iterable for a
-list literal or a Map for a map literal.
+right type even though the list that contains them does not. As long as the
+spread object is "spreadable"&mdash;it implements Iterable&mdash; there is no
+static error. This is true even if the object being spread is a user-defined
+class that implements Iterable but isn't even a subtype of List. For spreading
+into map literals, we require the spread object to be a class that implements
+Map, but not necessarily a subtype of the map being spread into.
 
 It is a static error if:
 
@@ -260,7 +263,7 @@ class InfiniteSequence implements Iterable<int> {
   }
 }
 
-const forever = [InfiniteSequence()];
+const forever = [...InfiniteSequence()];
 ```
 
 ### Type inference
@@ -271,15 +274,16 @@ Inference propagates upwards and downwards like you would expect:
     then the downwards inference context type of a spread element in that list
     is `Iterable<T>`.
 
-*   If a spread element in a list literal has type `Iterable<T>` for some `T`,
-    then the upwards inference element type is `T`.
+*   If a spread element in a list literal has static type `Iterable<T>` for some
+    `T`, then the upwards inference element type is `T`.
 
 *   If a map literal has a downwards inference type of `Map<K, V>` for some `K`
     and `V`, then the downwards inference context type of a spread element in
     that map is `Map<K, V>`.
 
-*   If a spread element in a map literal has type `Map<K, V>` for some `K` and
-    `V`, then the upwards inference key type is `K` and the value type is `V`.
+*   If a spread element in a map literal has static type `Map<K, V>` for some
+    `K` and `V`, then the upwards inference key type is `K` and the value type
+    is `V`.
 
 ## Dynamic Semantics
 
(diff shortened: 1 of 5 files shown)


working/0047. Spread Collections/spread-collections.md
```diff
@@ -0,0 +1,574 @@
...
+                    padding: const EdgeInsets.only(top: 24.0, bottom: 4.0),
+                    child: Text('Steps', style: headingStyle)
+                  ),
+                ]
+              ),
+// More...
+```
+
+Here, the `...` almost looks like it's part of the *next* element, the TableRow,
+than the preceding one. Given this, I think `...` is the better choice.
```
natebosch: Did you want to say "prefix" here?

munificent: Oops, fixed!

mit-mit:
Associated feature issue: https://github.com/dart-lang/language/issues/47

working/0047. Spread Collections/spread-collections.md
```diff
@@ -0,0 +1,574 @@
...
+
+  Iterator<int> get iterator {
+    return () sync* {
+      var i = 0;
+      while (true) yield i ++;
+    }();
+  }
+}
+
+const forever = [InfiniteSequence()];
```
lrhn: missing `...` ?

working/0047. Spread Collections/spread-collections.md
```diff
@@ -0,0 +1,574 @@
...
+
+### Type inference
+
+Inference propagates upwards and downwards like you would expect:
+
+*   If a list literal has a downwards inference type of `List<T>` for some `T`,
+    then the downwards inference context type of a spread element in that list
+    is `Iterable<T>`.
+
+*   If a spread element in a list literal has type `Iterable<T>` for some `T`,
```
lrhn: ... has static type ...

working/0047. Spread Collections/spread-collections.md
```diff
@@ -0,0 +1,574 @@
...
+
+        1.  Evaluate `value.entries.iterator` to a value `iterator`.
+
+        1.  Loop:
+
+            1.  If `iterator.moveNext()` returns `false`, exit the loop.
+
+            1.  Evaluate `iterator.current` to a value `newEntry`.
+
+            1.  Call `map[newEntry.key] = value`.
```
lrhn: value -> newEntry.value

lrhn:
LGTM with comments.

working/0047. Spread Collections/spread-collections.md
```diff
@@ -0,0 +1,574 @@
...
+
+```dart
+var numbers = <num>[1, 2, 3];
+var ints = <int>[...numbers];
+```
+
+This works because the individual elements in `numbers` do happen to have the
+right type even though the list that contains them does not. However, the spread
+object does need to be "spreadable"&mdash;it must be some kind of Iterable for a
+list literal or a Map for a map literal.
```
leafpetersen: Consider a comment to point out that this also doesn't require that the type of the thing being spread be assignable to the type of the List.  e.g. spreading MyList<Object> into a List<int> (neither is a subtype of the other).

working/0047. Spread Collections/spread-collections.md
```diff
@@ -0,0 +1,574 @@
...
+*   A spread element in a map literal has a static type that is not assignable
+    to `Map<Object, Object>`.
+
+*   If a map spread element's static type implements `Map<K, V>` for some `K`
+    and `V` and `K` is not assignable to the key type of the map or `V` is not
+    assignable to the value type of the map.
+
+If implicit downcasts are disabled, then the "is assignable to" parts here
+become strict subtype checks instead.
+
```
leafpetersen: There is a specified error preventing duplicate keys in map literals when the keys are constant.  Should we consider that for nested map literals + spread?  Should the following be an error?

```
  var a = {3 : true,
               ...{3 : false}};
```

working/0047. Spread Collections/spread-collections.md
munificent: Done.

working/0047. Spread Collections/spread-collections.md
munificent: Done.

working/0047. Spread Collections/spread-collections.md
munificent: Done.

working/0047. Spread Collections/spread-collections.md
munificent: Done.

working/0047. Spread Collections/spread-collections.md
munificent: I thought about duplicate keys, but the static checking for that is even pretty brittle for flat literals since it only works for literal keys (obviously):

```dart
var key = "key";
var map = {key: 1, key: 2}; // <-- No error.
```

So I figured the duplicate key checking wasn't very high value and wasn't worth extending into spread. Also, it might get weird if we introduce "conditional spread" features.

lrhn: I believe the rule about duplicate keys only apply to constant maps:

> It is a compile-time error if two keys of a constant map literal are equal.

so since spreads do not apply to constant maps, that issue is moot.

PR #56, opened 2018-10-23
Clarify file names

mit-mit:


lrhn:
There is also 
> `[name-of-feature]-implementation-plan.md`
on line 80.

mit-mit:
Good catch, updated

Issue #72, opened 2018-10-31
Terminating Tokens


munificent:
This is a proposed solution #69, optional semicolons.

I'm calling it "terminating tokens" to differentiate it from other semicolon insertion rules we might investigate. In this proposal, the rules are roughly:

* A "terminating token" is a token with a newline between it and the previous token.

* In places where a semicolon is currently allowed, if the next token is a terminating token, no semicolon is required.

* This leads to ambiguity since the existing grammar rules ignore whether or not a token is terminating. To resolve this, the grammar is changed so that a newline *cannot* be ignored in each place where that leads to a known ambiguity.

The "known ambiguity" part means that newlines are *only* used to act as implicit semicolons or to resolve ambiguities. A newline won't *cause* a syntax error on the next line. This lets us ignore most newlines, which minimizes the amount of migration and breakage.

On the other hand, the rules are fairly complex and subtle. We may not catch all of the ambiguities, and it may make it harder to introduce syntax in the future because it might cause a new ambiguity.

lrhn:
Example:
```dart
{ v
-- x}
```
Since `--` would be a terminating token (`e1` is a full expression, and `--` can start a new statement), then:
```dart
{ v
-- }
```
would have the same termination behavior as the first example, at least if you only look one token ahead.
I don't think grammar based disambiguation at "known ambiguity" positions will be sufficient to avoid the new line from causing a syntax error.

MarkOSullivan94:
> In places where a semicolon is currently allowed, if the next token is a terminating token, no semicolon is required

If I've understood this correctly it sounds like there'll be a mix of code with semicolons and code without semicolons? If so, it sounds like it could be pretty confusing.

gamebox:
@MarkOSullivan94 Sounds like a case for dartfmt :-)

mit-mit:
Moving this to icebox for now; for more details, see https://github.com/dart-lang/language/blob/master/resources/optional-semicolons-prototype.md

PhilParisot:
@mit-mit this is a feature I was really looking forward to, and can add a lot of "Coolest kid on the block" factor if done properly. People I've spoken to all thought Dart felt dated, it sounds dumb and trivial but making semicolons optional adds a lot of flash to a language.

andrewackerman:
[...]

PR #73, opened 2018-10-31
Add working spec for "terminating tokens".

munificent:
This is the maximally compatible robust approach to optional semicolons.
I'll start working on a simpler more conservative alternative soon.

This PR is just the proposal doc. I have a bunch of other supplementary material:

* A handful of small scripts for scraping a corpus and counting occurrences of certain forms.

* A fork of the front_end and dart_style packages with a prototype implementation of the rules.

* A Makefile for running tests on a big corpus of code.

The scripts reuse some existing helper code I wrote for other ui-as-code language features, so I'd have to duplicate some stuff to migrate it all into here. The front_end and dart_style packages are fairly large. Do you think it's worth bringing this stuff in?

### DIFF
--- a/working/terminating-tokens/feature-specification.md
+++ b/working/terminating-tokens/feature-specification.md
@@ -1,0 +1,1665 @@
+# Terminating Tokens
+
+Make Dart cleaner and less error-prone by eliminating pointless semicolon
+syntax, while avoiding the insanity of JavaScript's semicolon insertion rules.
+
+## Motivation
+
+For such a tiny feature&mdash;literally just eliminating some `;` at the ends of
+some lines&mdash;people have *really* strong feelings about using newlines as
+statement separators. Because of that, 2/3 of this proposal is about persuading
+you that we *should* and *can* do it. If you already think optional semicolons
+is a good idea, feel free to skip the next two sections and go straight to the
+proposal. Otherwise, I'll do my best to convince you first.
+
+### User desire
+
+The simplest motivation is that many users already want it. It was one of the
+[very first feature requests for Dart][sdk#34], and has 20 👍 and 3 ❤️. (That's
+not a lot, but keep in mind this bug was closed years ago.) The [more recent
+issue][sdk#30347] has 40 👍, 18 🎉, and 11 ❤️, compared to 7 😕 and 2 👎.
+
+[sdk#34]: https://github.com/dart-lang/sdk/issues/34
+[sdk#30347]: https://github.com/dart-lang/sdk/issues/30347
+
+BASIC, F#, Go, Groovy, Haskell, JavaScript, Julia, Lua, Python, R, Ruby, Scala,
+and the various shell languages all omit semicolons. Dart's main competitors are
+TypeScript, Swift, and Kotlin. They all do too.
+
+The odds that all of the Dart users who starred those bugs, the designers of
+those other languages, and their users are all delusional is pretty slim. This
+may not be enough to convince you that significant newlines are *better*, since
+there are also plenty of successful languages that *don't* have them, but it
+should send a clear signal that they are a reasonable idea.
+
+### User errors
+
+One aspect of good syntax design is that it minimizes user error. Humans are
+fallible no matter what, but a good syntax has fewer bumps in the carpet to trip
+over. Those of us well-versed in semicolons are so used to writing them that we
+probably forgot how *un*-natural they were at first. That bump is still there,
+we've just learned to sidestep it. New Dart users, though, still have to learn
+that. Eliminating them can make that process easier.
+
+**TODO: Are there any studies on this we can reference?**
+
+Even experienced manual semicolon inserters still make mistakes. After years of
+programming in Dart, I still forget the mandatory semicolon when assigning a
+lambda to a variable:
+
+```dart
+someCallback = () {
+  ...
+} // <-- Error!
+```
+
+Flutter code, with its deep nesting of expressions and lambdas exacerbates this. Look at the series of closing delimiters at the end of:
+
+```dart
+Widget build(BuildContext context) {
+  return CupertinoTabScaffold(
+    tabBuilder: (context, int index) {
+      return CupertinoTabView(
+        builder: (context) {
+          return CupertinoPageScaffold(
+            navigationBar: CupertinoNavigationBar(
+              middle: Text('Page 1 of tab $index'),
+            ),
+            child: Center(
+              child: CupertinoButton(
+                child: const Text('Next page'),
+                onPressed: () {
+                  Navigator.of(context).push(
+                    CupertinoPageRoute<Null>(
+                      builder: (context) {
+                        return CupertinoPageScaffold(
+                          navigationBar: CupertinoNavigationBar(
+                            middle: Text('Page 2 of tab $index'),
+                          ),
+                          child: Center(
+                            child: CupertinoButton(
+                              child: const Text('Back'),
+                              onPressed: () { Navigator.of(context).pop(); },
+                            ),
+                          ),
+                        );
+                      },
+                    ),
+                  );
+                },
+              ),
+            ),
+          );
+        },
+      );
+    },
+  );
+}
+```
+
+The commas are optional, but each of those five semicolons is required. Omit
+one, and it's an error. Place one on any of the other lines where they aren't
+required, and it's an error.
+
+The fact that it's an *error* to omit them shows how needless they are in many
+places. The language *already knows* the statement has ended, which is why it's
+able to tell you you forget the semicolon at exactly that point. The semicolons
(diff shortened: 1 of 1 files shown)


working/terminating-tokens/feature-specification.md
```diff
@@ -0,0 +1,1665 @@
...
+that its stewards are taking it in the direction they want. They are looking for
+signals about what kind of principles or philosophy the shepherds of the
+language have so they can predict if the language will continue to be a good fit
+for them.
+
+A good example of this is [tail call optimization][tco]. Every time a new
+language appears on the scene, people show up asking if it does "proper tail
+calls". (In Dart, it was [issue #29][sdk#29].) TCO is a neat feature that can
+enable some nice implementation patterns. But, if your language has loops (which
+they almost all do), it doesn't let you do anything you can't already do. It's a
```
leafpetersen: As nitted before, this isn't true, so we really shouldn't say it.

working/terminating-tokens/feature-specification.md
```diff
@@ -0,0 +1,1665 @@
...
+    Even so, the formatter will produce this formatting if the user didn't write
+    braces and the body doesn't fit on one line. More than half of these are in
+    the Flutter repository, which doesn't use dartfmt or fully follow the style
+    guide.
+
+*   **17** (0.008%) have bodies that are empty statements.
+
+[style guide]: https://www.dartlang.org/guides/language/effective-dart/style#do-use-curly-braces-for-all-flow-control-structures
+
+It's rare that a user deliberately wants an empty statement as the body of a
```
leafpetersen: I still don't understand this paragraph (see my previous email).

mit-mit:
This is the draft feature spec for https://github.com/dart-lang/language/issues/72

working/terminating-tokens/feature-specification.md
```diff
@@ -0,0 +1,1665 @@
...
+primary use case for splitting a string literal into multiple adjacent strings:
+
+*   The corpus has **11,256** adjacent strings. By that, I mean strings that
+    immediately follow another string literal, so `"a" "b" "c"` is two adjacent
+    strings.
+
+*   **11,222** (99.698%) have a newline between them and the previous string.
+
+So, in this case, we don't follow eagerness and ignore a newline between
+adjacent strings.
```
lrhn: So we will mis-interpret:
```dart
var x = "some string"
"some other string".charCodes.forEach(importantAction)
```
?
What can the user do to avoid this? Parenthesizing either string literal. Use a literal semicolon (that's the smaller solution, so that's what I would do, but then the formatter cannot remove it).
That's also a thing to consider: There will be semicolons that cannot be removed (empty statements can be turned into `{}`, but not removed). If we remove the empty statement  `;` completely, and reserve only `{}` for that (the *other* empty statement), then that issue should be gone.
(Let's do that no matter what! Please!)

working/terminating-tokens/feature-specification.md
```diff
@@ -0,0 +1,1665 @@
...
+    which have the `[` on the next line. Both are the result of chained index
+    operators like:
+
+    ```dart
+    exifData[tag] = ExifConstants.stringValues["Components"]
+            [exifData[tag][0]]
+    ```
+
+    One is in an expression context where the newline is ignored and one is not.
+
```
lrhn: There is also `--` and `++` which can both end and start an expression (prefix and suffix operators):
```dart
a
++
b
```
can be parsed in two ways with optional semicolons. 
I doubt that ever happens in practice, so using the eagerness principle should be safe.

working/terminating-tokens/feature-specification.md
```diff
@@ -0,0 +1,1665 @@
...
+```dart
+main() {
+  foo()
+  bar()
+}
+```
+
+Here, `bar` is a terminating token and behaves as if a semicolon appeared before
+it, after the `)` on `foo()`. The "should be significant" is vague. More
```
leafpetersen: Should this be "could be significant", or should the the "could be" a few lines up be a "should be"?  Or is this referring somewhere else, in which case clarify?

working/terminating-tokens/feature-specification.md
```diff
@@ -0,0 +1,1665 @@
...
+precisely:
+
+*   If a newline character appears after the preceding token's lexeme and
```
leafpetersen: There's something missing from this paragraph - it feels vacuous.  All newline characters appear between two tokens, no?  Maybe something about "and this token is one of a specific set of tokens"?

working/terminating-tokens/feature-specification.md
munificent: Oof, yes. The proposal will elide the newline there.

> What can the user do to avoid this? Parenthesizing either string literal.

Yeah, parenthesizing would work.

> (Let's do that no matter what! Please!)

SGTM. Empty statement is weird.

working/terminating-tokens/feature-specification.md
munificent: Sorry, it's referring to the "should be ignored". I'd rearranged some text and switched that around. Fixed.

working/terminating-tokens/feature-specification.md
```diff
@@ -0,0 +1,1665 @@
...
+    function() { print("hi") }
+    ```
+
+*   Otherwise, it is not.
+
+### Terminator rules
+
+To use terminating tokens we introduce a couple of special grammar terminals:
+
+*   The `TERM` rule matches and consumes an explicit semicolon. It also matches
```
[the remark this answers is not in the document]
munificent: These are handled by the combination of:

* "[TERM] matches but does not consume a terminating token"
* "if the token is `}`, it **is** terminating"

So a closing bracket always behaves like there's an implicit semicolon before it, whether or not there is a newline.

working/terminating-tokens/feature-specification.md
munificent: Yes, but not all pairs of tokens do have newlines between them. :) If there's no newline between a token and the next, then the latter token is not terminating (unless it's `}`).

working/terminating-tokens/feature-specification.md
MarkOSullivan94: I'm curious... what would happen in the case of those editors and IDEs which automatically force the code on to a new line when the width of the line gets too long?

Can just imagine in those cases and the cases of the code shown above by @Irhn it will get very confusing.

mit-mit:
What's next for this PR?

working/terminating-tokens/feature-specification.md
```diff
@@ -0,0 +1,1656 @@
...
+    is just banging code into their editor and not thinking about
+    formatting&mdash;that's dartfmt's job after all&mdash;how likely are they to
+    get what they want out? If they copy and paste a big chunk of code somewhere else, how often does it end up still correct?
+
+In an ideal world, we'd hit 100% on all four of those. Short of radically
+changing the language into something like Lisp, that's impossible. But we can
+run experiments on a corpus of Dart code to try to empirically measure these to
+some degree.
+
+The corpus I used is the Flutter repository, the Dart SDK repository, and the
```
Stargator: Why mention Flutter? Yes, it uses Dart, but so does GoogleAdware and Google Adsense. Additionally, there's the AngularDart team, plus all the frameworks the community has built.

Is this change being requested solely to benefit Flutter uses? I would urge that either these other uses of Dart are included or otherwise limit the mentions of Flutter so it's not so specific to one user base. Though you do go through packages on Pub, so I don't know why Flutter had to specifically be called out and the rest excluded.

munificent: This change is part of the "UI as code" charter, which is detailed more [here](https://github.com/munificent/ui-as-code) and [here](https://github.com/munificent/ui-as-code/blob/master/Motivation.md). It's not *only* to benefit Flutter, but if it *doesn't* benefit Flutter, then I don't think we have the resources to work on it right now. In other words "useful for Flutter" is a necessary condition, but we certainly hope that all "UI as code" features are useful for much *more* than Flutter too.

Flutter's repo is an important use case because it's a very large, very popular corpus of Dart code and is also one that does not currently use dartfmt and has its own formatting style. So it's a good test case for how robust the optional semicolon rules are across a variety of Dart code.

mit-mit:
@munificent can we merge this?

munificent:
I need to circle back on semicolons and tie up the loose ends. I think the right thing to do for this is to add a note to the top saying we probably aren't going to pursue this approach and then land it for reference.

@leafpetersen and @lrhn, does that sound OK?

leafpetersen:
SGTM

working/terminating-tokens/feature-specification.md
Stargator: Hmm, thanks for the quick response @munificent. I appreciate the openness and you provided some clarity

PR #77, opened 2018-11-02
First draft of "Control Flow Collections" proposal.

munificent:


working/control-flow-collections/feature-specification.md
```diff
@@ -0,0 +1,913 @@
...
+composed. That enables some interesting patterns and techniques:
+
+```dart
+[for (var x in hor) for (var y in vert) Point(x, y)]
+```
+
+This produces the Cartesian product of all points in the rectangle.
+
+```dart
+[for (var i in integers) if (i % 2 == 0) i * i]
```
lrhn: Nitpick: `i % 2 == 0` -> `i.isEven`

working/control-flow-collections/feature-specification.md
```diff
@@ -0,0 +1,913 @@
...
+      for (var i = 0; i < 1; i++) []
+    ];
+    ```
+
+    Produces a `List<List<String>>` containing three empty `List<String>`.
+
+*   The upwards inference element type of an `if` list element without an `else`
+    is the type of the "then" element.
+
+*   The upwards inference element type of an `if-else` list element is the least
```
lrhn: Does it ever matter that you find the LUB of these two first, before finding the LUB of that and the other elements?
If so, maybe it would be nicer to treat the if/else as introducing two elements in the following element type LUB.

working/control-flow-collections/feature-specification.md
```diff
@@ -0,0 +1,913 @@
...
+
+1.  Else, if `element` is a `for-in` element:
+
+    1.  Evaluate the iterator expression to a value `sequence`.
+
+    1.  Evaluate `sequence.iterator` to a value `iterator`.
+
+    1.  Loop:
+
+        1.  If `iterator.moveNext()` returns `false`, exit the loop.
```
lrhn: Here you also need boolean conversion/checking. Even if `sequence` implements `Iterable` (and I'm not sure you've said that it should), it can return `null`.

In practice, we will likely abstract the current semantics of a `for` statement over the "body", so we can use the same specification for `for` statements and `for` literal elements.
That should handle all the book-keeping of for loops. We do want them to act exactly the same, so it's safer to avoid spec duplication :)

working/control-flow-collections/feature-specification.md
```diff
@@ -0,0 +1,913 @@
...
+
+    ```dart
+    {
+      for (var e in things)
+        someExpression(e): anotherExpression(e)
+    }
+    ```
+
+    This may fall down if the two closures are complex but I think that's rare
+    in practice.
```
lrhn: (or the fix can name the functions first and use them in the map comprehension)

working/control-flow-collections/feature-specification.md
```diff
@@ -0,0 +1,913 @@
...
+## Next Steps
+
+As always, the immediate next step of a proposal is running it past the language
+leads and stakeholders.
+
+### Usability studies
+
+This feature has some good things going for it:
+
+*   It is just syntax sugar. A front end should be able to compile this down to
```
lrhn: It's not just syntactic sugar if you allow if-expressions in constant lists/maps.
(Or at least, it might require exponential blow-up to represent the same list without the feature).

For non-constants, it is just syntactic sugar.

working/control-flow-collections/feature-specification.md
```diff
@@ -0,0 +1,913 @@
...
+  const? typeArguments? '{' mapLiteralEntryList? '}' ;
+
+mapLiteralEntryList:
+  mapLiteralEntry ( ',' mapLiteralEntry )* ','?
+  ;
+
+mapLiteralEntry:
+  expression ':' expression |
+  'if' '(' expression ')' mapLiteralEntry ( 'else' mapLiteralEntry )? |
+  'for' '(' forLoopParts ')' mapLiteralEntry
```
lrhn: Consider allowing `await for` in an async function.
We do allow `await` in the expressions, and the user can get the same effect as:
```dart
await for (var s in stream) something(s)
```
as 
```dart
for (var s in await stream.toList()) something(s)
```
it's just less efficient and direct.

working/control-flow-collections/feature-specification.md
munificent: I thought about this a little. As far as I know, LUB is associative so it shouldn't matter, but I wouldn't be surprised if there were subtleties I'm missing.

I defaulted to defining this locally instead of flattening because the latter felt a little weird when you consider that the elements of a list may have access to a variable that is out of scope at the top level of the list:

```dart
[for (var i in stuff) if (cond) i.something else i.other]
```

I wasn't sure if flattening the then and else parts would make things weird given that `i` is only meaningful inside the `for`.

working/control-flow-collections/feature-specification.md
```diff
@@ -0,0 +1,913 @@
...
+  value: (demo) => demo.buildRoute,
+);
+```
+
+Into:
+
+```dart
+return {
+  for (var demo in kAllGalleryDemos)
+    '${demo.routeName}': demo.buildRoute,
```
srawlins: That colon there is suspiciously reminiscent of an argument list, eh?

```dart
  return Row(
    children: [
      IconButton(icon: Icon(Icons.menu)),
      Expanded(child: title),
      IconButton(
          icon: Icon(Icons.search),
          if (hasOnPressed) onPressed: onPressed),
    ]
  );
```

(Would be part of a separate proposal, hope I'm not spoiling the punchline!)

working/control-flow-collections/feature-specification.md
munificent: Done.

Unifying the spec for `for` statements and `for` elements sounds great.

working/control-flow-collections/feature-specification.md
munificent: True, but I think that would make the code *worse* in most users eyes.

working/control-flow-collections/feature-specification.md
munificent: Done.

working/control-flow-collections/feature-specification.md
munificent: Done.

working/control-flow-collections/feature-specification.md
munificent: OK, took a stab at adding this.

working/control-flow-collections/feature-specification.md
```diff
@@ -0,0 +1,913 @@
...
+command.add(mainPath);
+```
+
+Into:
+
+```dart
+var command = [
+  engineDartPath,
+  frontendServer,
+  for (var root in fileSystemRoots) '--filesystem-root=$root'
```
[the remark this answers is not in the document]
munificent: Oops, yes! :D

working/control-flow-collections/feature-specification.md
munificent: See the last section of the doc. :)

working/control-flow-collections/feature-specification.md
```diff
@@ -0,0 +1,976 @@
...
+```dart
+IconButton(
+  icon: Icon(Icons.menu),
+  tooltip: 'Navigation menu',
+  if (isAndroid) padding: const EdgeInsets.all(20.0),
+)
+```
+
+Without [rest parameters][], `for` isn't useful and `if` probably isn't
+feasible for positional arguments. But even without rest params, it's possible
```
yjbanov: What if positional arguments are made optionally "nameable"? Could that help with feasibility?

munificent: ...yes, but making positional arguments optionally nameable also has its own feasibility problems. :) That's a fairly gnarly design problem, I think.

working/control-flow-collections/feature-specification.md
```diff
@@ -0,0 +1,913 @@
...
+      IconButton(icon: Icon(Icons.menu)),
+      Expanded(child: title),
+  ];
+
+  if (isAndroid) {
+    buttons.add(IconButton(icon: Icon(Icons.search)));
+  }
+
+  return Row(
+    children: children,
```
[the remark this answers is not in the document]
munificent: Oops, fixed.

working/control-flow-collections/feature-specification.md
```diff
@@ -0,0 +1,976 @@
...
+However, the human side of this is less certain. I've tried to make the behavior
+intuitive by piggy-backing on syntax users already understand. But I worry that:
+
+*   Users will find it confusing to see `if` or `for` inside a collection
+    literal.
+
+*   Even after understanding it, users may not *like* the syntax.
+
+*   Users may be unhappy that the syntax doesn't go *far enough*. This feature
+    may lead them to expect, say, `while` to work inside a collection. They may
```
srawlins: This is the first instance of the word "while" in this doc. Is it worth it to speak above as to why only two of the control flow statements are considered here? (No `while`, `do`, or `switch`).

munificent: Good call. Done.

working/control-flow-collections/feature-specification.md
```diff
@@ -0,0 +1,1021 @@
...
+
+However, the human side of this is less certain. I've tried to make the behavior
+intuitive by piggy-backing on syntax users already understand. But I worry that:
+
+*   Users will find it confusing to see `if` or `for` inside a collection
+    literal.
+
+*   Even after understanding it, users may not *like* the syntax.
+
+*   Users may be unhappy that the syntax doesn't go *far enough*. This feature
```
InMatrix: I think this is one of the key questions we should answer in a usability study. We probably can show the user a few correct examples and then ask the user to write some new examples where going "too far" could provide some legitimate benefit. We'd like to see where the user perceives the limit is. Another aspect we can evaluate as part of this study is how we communicate errors once the user steps out of the supported syntax. It could be useful for the diagnostics guidelines project, too. Cc: @bwilkerson

working/control-flow-collections/feature-specification.md
```diff
@@ -0,0 +1,1021 @@
...
+
+This procedure handles elements in both list and map literals because the only
+difference is how an atomic list element or map entry is handled. The control
+flow parts are the same so are unified here.
+
+1.  If `element` is an `if` element:
+
+    1.  Evaluate the condition expression to a value `condition`.
+
+    1.  Subject `condition` to boolean conversion to a value `result`.
```
leafpetersen: by boolean conversion you mean do any implied downcast and check for null?

working/control-flow-collections/feature-specification.md
```diff
@@ -0,0 +1,1021 @@
...
+```dart
+var closures = [for (var i = 1; i < 4; i++) () => i];
+for (var closure in closures) print(closure());
+// Prints "1", "2", "3".
+```
+
+### Static errors
+
+The static semantics of collection `if` and `for` mostly follow their statement
+analogues.
```
leafpetersen: Is it worth spelling out that it is a static error to have a list element inside of a map literal, or a map element inside of a list literal, or is this considered a parse error?

working/control-flow-collections/feature-specification.md
```diff
@@ -0,0 +1,1021 @@
...
+If we also support a spread syntax, we can compose that to include multiple
+elements based on a single `if` condition:
+
+```dart
+Widget build(BuildContext context) {
+  return Row(
+    children: [
+      IconButton(icon: Icon(Icons.menu)),
+      Expanded(child: title),
+      if (isAndroid) ...[
```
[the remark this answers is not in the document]
lrhn: This is a spread element. My guess is that the grammar will be:

> listElement ::= 
>    expression
>  | `...` expression
>  | `for` `(` forHeader `)` listElement
>  | `if` `(` expression `)` listElement (`else` listElement)?
>  ;

(so, the former of your suggestions).

working/control-flow-collections/feature-specification.md
```diff
@@ -0,0 +1,1021 @@
...
+```
+
+### Static errors
+
+The static semantics of collection `if` and `for` mostly follow their statement
+analogues.
+
+If is a static error when:
+
+*   The collection is a list and the type of any of the body elements may not be
```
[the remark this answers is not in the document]
lrhn: Fix it where possible.
If you do:
```dart
Iterable<int> l = [if (b) "hello" : []];
```
then the context type determined the list type to be `List<int>` so the element type is `int`. We should move that constraint into the branches.
If there is no element type, then it's not possible to fail, and we will use the upwards inference to guess the actual list type, and then I hope we will unify both `"hello"` and `[]` with any other elements, not synthesize `Object` early and then combine that with other elements (if it matters).

working/control-flow-collections/feature-specification.md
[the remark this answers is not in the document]
munificent: What Lasse says. The body of an `if` is a `listElement`. The spread proposal [introduces the same grammar rule](https://github.com/dart-lang/language/blob/master/working/spread-collections/feature-specification.md#syntax) and puts the `...` under there. So if we do both, those get merged together and `listElement` can contain `...`, `if`, `for`, or an expression.

PR #79, opened 2018-11-03
Move over "Enhanced Type Promotion" proposal.

munificent:
Was originally at: https://github.com/dart-lang/sdk/pull/29624

### DIFF
--- a/working/enhanced-type-promotion/feature-specification.md
+++ b/working/enhanced-type-promotion/feature-specification.md
@@ -1,0 +1,630 @@
+# Enhanced Type Promotion
+
+Author: Brian Wilkerson
+
+Language team shepherd: Bob Nystrom
+
+Status: Strawman
+
+The Dart language features type promotion. This lets you avoid pointless
+explicit casts in some cases where it's obvious that a variable has a more
+specific type than its declared (or originally inferred) type. For example:
+
+```dart
+int stringLength(Object o) {
+  if (o is String) {
+    return o.length; // OK!
+  } else {
+    throw "Not a string.";
+  }
+}
+```
+
+Without type promotion, the marked line would have a static error since Object
+does not have a `.length` getter. Type promotion notes that an is String test
+was applied to `o`, and promotes or "re-types" `o` to have type String inside
+the then clause of the if statement.
+
+The type promotion rules in Dart are useful, but very restricted. Some
+restriction is necessary because dataflow based type inference can get quite
+complex in some cases. But, in practice, the current rules are too restrictive.
+For example, this does not work:
+
+```dart
+int stringLength(Object o) {
+  if (o is! String) {
+    throw "Not a string.";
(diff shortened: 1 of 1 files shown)


mdebbar:
Do types promote into closures?
```dart
class Foo {
  void foo() => print('foo');
}

var action;
if (x is Foo) {
  action = () {
    x.foo(); // Is this ok?
  }
} else {
  action = () => null;
}
scheduleMicrotask(action);
```
Type promotion in the above scenario is safe if `x` is final, but is dangerous otherwise because could be reassigned another value before `action` is invoked.

working/enhanced-type-promotion/feature-specification.md
```diff
@@ -0,0 +1,630 @@
...
+      if (o is String) {
+        if (callClosure) closure();
+        print(o.length);
+      }
+    }
+    ```
+
+    At the point where `.length` is accessed, `o` may be a String, or it may be
+    an int if `callClosure` was true. To avoid cases like this, the spec defines
+    certain variables to be off limits for type promotion.
```
mdebbar: It would be great to elaborate a bit here. Under what conditions do we consider  a variable safe to type-promote?
```dart
void foo(x) async {
  closure() {
    x = 123;
  }

  if (x is String) {
    print(x.length); // Safe
  }

  if (x is String) {
    if (something) closure();
    print(x.length); // Unsafe
  }

  if (x is String) {
    await something;
    print(x.length); // Unsafe
  }
}
```

mdebbar: Another example, a bit contrived:
```dart
var x = 'abc';

class A {
  bool get foo {
    x = 123;
    return true;
  }
}

void main() {
  var a = A();
  if (x is String && a.foo) {
    print(x.length); // Unsafe!
  }
}
```

lrhn:
Landing this, if for nothing else, then as background material.

working/enhanced-type-promotion/feature-specification.md
munificent: Yeah, these are good points. This proposal isn't quite active right now and I'm not the author, so I don't think we'll tweak it right now. But if we summon it back to life, I think we should be more explicit about the things you bring up.

bwilkerson: Line 120 says that the definition of which variables are safe to type promote is unchanged by this proposal, so it's what the spec says:

> The static type system ascribes a static type to every expression. In some cases, the type of a local variable (which can be a formal parameter) may be promoted from the declared type, based on control flow.

With the following caveat added in many places:

> ...  then the type of v is known to be T in s2, unless any of the following are true
> - v is potentially mutated in s1,
> - v is potentially mutated within a function other than the one where v is declared, or
> - v is accessed by a function defined in s1 and v is potentially mutated anywhere in the scope of v.

PR #87, opened 2018-11-08
Initial draft of implementation plan for set literals

leafpetersen:
Here's a first cut at an implementation plan for set literals.  I experimented with specifying the tasks as a DAG instead of linear phases, not sure it's worth the overhead, but we can see how people like it.  

The schedule needs input from the implementation team leads.

@lrhn @mit-mit @eernstg @munificent @JekCharlsonYu 

Please take a look.  Feel free to push updates or changes.  We can land this and then ask for feedback, or we can try soliciting feedback and sign off from the implementation teams on this PR as somebody (Michael?) recently suggested.

### DIFF
--- a/accepted/future-releases/set-literals/implementation-plan.md
+++ b/accepted/future-releases/set-literals/implementation-plan.md
@@ -26,7 +26,7 @@ The CFE implements parsing support for the new literal syntax.
 ### Task2 : Task1 (CFE implementation)
 
 The CFE implements support for set literals (errors and warnings, constant
-evaluation, and any required changed to support the backends).
+evaluation, and any required changes to support the backends).
 
 ### Task3 : Task1 (Analyzer implementation)
 
@@ -47,17 +47,17 @@ dart2js implements backend support.
 
 ### Task7 : Task0 (Intellij support)
 
-Intellij implements syntax highlighting.
+IntelliJ parser supports set literals
 
 ### Task8 : Task1 (Grok support)
 
 Grok implements any required support.
 
-### Task9 : Task1 (Dartfmt support)
+### Task9 : Task3 (Dartfmt support)
 
 Dartfmt implements formatting support.
 
-### Task9 : Task1 (Dartdoc support)
+### Task9 : Task3 (Dartdoc support)
 
 Dartdoc implements support.
 
--- a/accepted/future-releases/set-literals/implementation-plan.md
+++ b/accepted/future-releases/set-literals/implementation-plan.md
@@ -1,0 +1,102 @@
+# Implementation Plan for Set literals
+
+Relevant documents:
+ - [Tracking issue](TODO)
+ - [Full proposal](https://github.com/dart-lang/language/blob/master/accepted/future-releases/set-literals/feature-specification.md)
+
+## Implementation and Release plan
+
+### Release flags
+
+The implementation of this change will happen behind
+an
+[*experiments flag*](https://github.com/dart-lang/sdk/blob/master/docs/process/experimental-flags.md).
+Tools need to be passed the flag `--enable-experiment=set-literals` for the
+changes to be enabled.
+
+### Task0 (Prerequisite)
+
+All tools add support for experimental flags, if not yet supported.  This is
+tracked as Phase 0 of [this issue](https://github.com/dart-lang/language/issues/60).
+
+### Task1 : Task0 (Parsing support)
+
+The CFE implements parsing support for the new literal syntax.
+
+### Task2 : Task1 (CFE implementation)
+
+The CFE implements support for set literals (errors and warnings, constant
+evaluation, and any required changed to support the backends).
+
+### Task3 : Task1 (Analyzer implementation)
+
+The analyzer implements static checking for set literals (errors and wrarnings,
+constant evaluation).
+
+### Task4 : Task3 (DDC implementation)
+
+DDC implements backend support.
+
+### Task5 : Task2 (VM support)
+
+VM implements backend support.
+
+### Task6 : Task2 (dart2js support)
+
+dart2js implements backend support.
+
+### Task7 : Task0 (Intellij support)
+
+Intellij implements syntax highlighting.
+
+### Task8 : Task1 (Grok support)
+
+Grok implements any required support.
+
+### Task9 : Task1 (Dartfmt support)
+
+Dartfmt implements formatting support.
+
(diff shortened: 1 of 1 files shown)


accepted/future-releases/set-literals/implementation-plan.md
```diff
@@ -0,0 +1,102 @@
...
+ - [Full proposal](https://github.com/dart-lang/language/blob/master/accepted/future-releases/set-literals/feature-specification.md)
+
+## Implementation and Release plan
+
+### Release flags
+
+The implementation of this change will happen behind
+an
+[*experiments flag*](https://github.com/dart-lang/sdk/blob/master/docs/process/experimental-flags.md).
+Tools need to be passed the flag `--enable-experiment=set-literals` for the
```
mit-mit: Maybe keep the flag singular (`set-literal`)?

leafpetersen: OK by me, done.

leafpetersen:
cc @vsmenon @sigmundch @jmesserly @kmillikin @a-siva @mraleph @devoncarew @stereotype441 @bwilkerson @jcollins-g @jwren 

Can you all please take a look at the plan and the schedule and either sign off or propose changes?

accepted/future-releases/set-literals/implementation-plan.md
```diff
@@ -0,0 +1,102 @@
...
+tracked as Phase 0 of [this issue](https://github.com/dart-lang/language/issues/60).
+
+### Task1 : Task0 (Parsing support)
+
+The CFE implements parsing support for the new literal syntax.
+
+### Task2 : Task1 (CFE implementation)
+
+The CFE implements support for set literals (errors and warnings, constant
+evaluation, and any required changed to support the backends).
```
bwilkerson: nit: "changed" --> "changes"

accepted/future-releases/set-literals/implementation-plan.md
```diff
@@ -0,0 +1,102 @@
...
+
+VM implements backend support.
+
+### Task6 : Task2 (dart2js support)
+
+dart2js implements backend support.
+
+### Task7 : Task0 (Intellij support)
+
+Intellij implements syntax highlighting.
```
bwilkerson: I believe that syntax highlighting is actually implemented in server. I don't know whether you care abut that level of detail.

But one thing that's missing is the need to update the IntelliJ parser to handle the new syntax.

accepted/future-releases/set-literals/implementation-plan.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Task8 : Task1 (Grok support)
+
+Grok implements any required support.
+
+### Task9 : Task1 (Dartfmt support)
```
bwilkerson: This depends on Task3.

accepted/future-releases/set-literals/implementation-plan.md
```diff
@@ -0,0 +1,102 @@
...
+
+Dartfmt implements formatting support.
+
+### Task9 : Task1 (Dartdoc support)
```
bwilkerson: This depends on Task3.

accepted/future-releases/set-literals/implementation-plan.md
devoncarew: Yes, would re-word to 'IntelliJ parser supports set literals'.

leafpetersen: Done.

accepted/future-releases/set-literals/implementation-plan.md
leafpetersen: Done.

accepted/future-releases/set-literals/implementation-plan.md
leafpetersen: Done.

accepted/future-releases/set-literals/implementation-plan.md
leafpetersen: Done

leafpetersen:
Landing this for reference, though I'm not sure if there's been further iteration offline?

munificent:
It's not clear which of these tasks can be done in parallel and which can't (though the schedule sort of implies it). Is it worth splitting it into phases? Numbering the tasks makes it seem like they're sequential.

leafpetersen:
> It's not clear which of these tasks can be done in parallel

It's apparently not clear from the syntax, but the schedule explicitly includes a dependency graph. 
`Task9 : Task3, Task4` means that `Task9` depends on `Task3` and `Task4`.

munificent:
Oh, sorry, I glossed over that part. It appears you are defining a Makefile. Are tabs significant in this document? 😉

kmillikin:
Task 0 (experimental flags) is just getting underway, but it doesn't block Task 1 from starting.

Task 1 (syntactic support in the CFE parser) has been implemented.  I'm not sure if it's complete yet but it looks like it.  It hasn't been wired up to Fasta.  It is selected by setting a variable in the code (that is, not exposed to users).

Task 2 (CFE implementation) consists of two pieces.  (1) we will add syntactic support to Kernel for SetLiteral, analogous to ListLiteral which is already there.  This requires some implementation in the type inferrer.  (2) we will implement a Kernel to Kernel transformation that translates `SetLiteral(e0, e1, ...)` into `new Set()..add(e0)..add(e1)...` and translates `const SetLiteral(c0, c1, ...)` into `const ImmutableSet(const {c0: null, c1: null, ...})` where ImmutableSet is an internal core library class.

Backends can build native support for SetLiteral but it is not necessarily required to launch the feature.

We have a realistic estimate of three weeks to complete it.

bwilkerson:
> Task 1 (syntactic support in the CFE parser) has been implemented. I'm not sure if it's complete yet but it looks like it.

I believe that it's complete. Analyzer is using it and has a complete implementation modulo one bug in type inference (that appears to be more general than the set support).

leafpetersen:
> It appears you are defining a Makefile. Are tabs significant in this document? 😉

Oh man, I now so regret not calling these things Makefiles instead of implementation plans.  We could have even written them in make.

munificent:
Galaxy brain idea: We could build a little command line app that takes a GitHub issue ID as an argument and exits with a non-zero exit code if the issue is still open. Then this could literally be a Makefile with a target that "builds" successfully only when all of the related issues are closed.

PR #100, opened 2018-11-14
Move "Spread Collections" to accepted and start implementation plan.

munificent:
I didn't put dates into it yet because I'm not sure what we want to say
there, but this is a start.

cc @JekCharlsonYu


lrhn:
LGTM. (Since the Set-literal proposal is ahead of this one now, we should also also mention Set-literals in the feature specification).

accepted/future-releases/spread-collections/implementation-plan.md
```diff
@@ -0,0 +1,145 @@
...
+Owner: rnystrom@google.com ([@munificent](https://github.com/munificent/) on GitHub)
+
+Relevant links:
+
+* [Tracking issue](https://github.com/dart-lang/language/issues/47)
+* [Proposal](https://github.com/dart-lang/language/blob/master/accepted/future-releases/spread-collections/feature-specification.md)
+
+## Phase 0 (Prerequisite)
+
+### Experiment support
```
leafpetersen: Let's just take this as given, delete this section.

accepted/future-releases/spread-collections/implementation-plan.md
```diff
@@ -0,0 +1,145 @@
...
+This feature can likely be implemented entirely in the front end, so back-end
+support may not be needed. If it does require Kernel changes, the back end will
+need to handle those changes.
+
+### Analyzer
+
+The analyzer implements parsing the new syntax, type checking it, and
+evaluating `...` in constant collections.
+
+## Phase 2 (Tool Implementation)
```
leafpetersen: Add dartdoc.

leafpetersen:
@mit-mit @JekCharlsonYu This is on the Q1 implementation list.  We'll need to work with the implementation teams to get a schedule and get it on their OKRs.

mit-mit:
Hi @munificent can you address the few comments so we can get this landed? (It's OK to land with just the draft impl. plan; we can refine that later)

accepted/future-releases/spread-collections/implementation-plan.md
munificent: Done.

accepted/future-releases/spread-collections/implementation-plan.md
munificent: Is there anything dartdoc needs to do for this? It doesn't affect the signatures of anything.

munificent:
I made, I think, all of the changes needed for set literals, but let me know if I missed anything. The interesting case is syntactic ambiguity when there isn't a type context:

```dart
var wat = {...a};
```

I think that should be inferred as map, instead of making it an error, to be consistent with empty `{}`.

I'll merge this in, but let me know if there are other changes you'd like.

accepted/future-releases/spread-collections/implementation-plan.md
leafpetersen: It affects that set of AST nodes it can see in constants, so I think it needs to be prepared to handle it or it may break.

Issue #39, opened 2018-10-09
Make it an error to use a type variable non-covariantly in a superinterface


eernstg:
This is a proposal for how to deal with the issue described in #38.

Related [feature specification](https://github.com/dart-lang/language/blob/master/accepted/future-releases/contravariant-superinterface-2018/feature-specification.md).

As described in #38, the ability to use a type variable contravariantly in a superinterface creates an "anti-parallel" subtype relationship for a given generic class and a direct superinterface thereof (and, of course, the same thing can happen indirectly in many ways). This creates various complications for the static analysis and the enforcement of the heap invariant ("soundness"). It easily generalizes to the invariant case, hence we'll target "non-covariant" occurrences here.

## Discussion

I believe that it would be beneficial for Dart to introduce a new compile-time error, and I suspect that it would break a very small amount of code. Here is the proposed error:

**Let _C_ be a generic class that declares a formal type parameter _X_, and assume that _T_ is a direct superinterface of _C_. It is a compile-time error if _X_ occurs contravariantly or invariantly in _T_.**

For example:
```dart
class A<X> {}
class B<X> extends A<void Function(X)> {} // Compile-time error.
class C<X> extends A<void Function<Y extends X>(Y)> {} // Compile-time error.
```

With that error in place, it is no longer possible to create the "anti-parallel" subtyping relationship that we used in #38 to violate the heap invariant.

## Alternative Proposal

One other approach could be used: We could specify that an upcast from `B<T>` for some `T` to a parameterized type based on `A` must yield `A<void Function(Null)>` rather than `A<void Function(T)>`. In general, we would perform such an upcast by replacing the relevant type variable by `Null` rather than replacing it by the actual type argument `T`. This would be sound, but it might also give rise to some confusion. I believe that a compile-time error is the better choice for superinterfaces.

## A Useful Property

If we introduce the above-mentioned error then we will have the following property: **"A well-bounded type will have well-bounded superinterfaces".**

That is a useful property because it enables the use of super-bounded types along with the ordinary subtype relationships, without forcing all types thus obtained to need a well-boundedness check.

To see this, we state without evidence that the current checks on regular-bounded types will ensure that superinterfaces are regular-bounded. That is, the bounds on the type parameters in a generic class declaration must be so strong that they enforce the bounds in each superinterface for the given actual type arguments.

[...]

PR #114, opened 2018-11-29
New files added, for contravariant-superinterface-2018

eernstg:
Introduce documents in `accepted/future-releases` for feature #39.

### DIFF
--- a/accepted/future-releases/contravariant-superinterface-2018/implementation-plan.md
+++ b/accepted/future-releases/contravariant-superinterface-2018/implementation-plan.md
@@ -1,4 +1,4 @@
-# Implementation Plan for the Q4 2018 Superinterface Contravariance Error.
+# Implementation Plan for the Q1 2019 Superinterface Contravariance Error.
 
 Relevant documents:
  - [Tracking issue](https://github.com/dart-lang/language/issues/113)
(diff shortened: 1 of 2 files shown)


accepted/future-releases/contravariant-superinterface-2018/implementation-plan.md
```diff
@@ -0,0 +1,30 @@
+# Implementation Plan for the Q1 2019 Superinterface Contravariance Error.
+
+Relevant documents:
+ - [Tracking issue](https://github.com/dart-lang/language/issues/113)
+ - [Feature specification](https://github.com/dart-lang/language/blob/master/accepted/future-releases/contravariant-superinterface-2018/feature-specification.md)
+
+
+## Implementation and Release plan
```
leafpetersen: I think we should use an experiment flag, even though breakage chances are low.  It makes sure we roll out consistently.  Maybe "covariant-only-superinterfaces"?

mit-mit: +1 to using that flag name

mit-mit:
LGTM if the flag mentioned above is added to the spec

accepted/future-releases/contravariant-superinterface-2018/implementation-plan.md
eernstg: Added flag to plan.

PR #116, opened 2018-11-29
Update generalized typedef feature spec to allow usage of typedef as class

eernstg:
This updates the feature spec of generalized type aliases to allow all class-like usages, including `F.myStaticGetter` and `F<int>.myStaticMethod(42)` where `F` is a type alias and `F` resp. `F<int>` may denote a class `C` or a parameterized type `C<...>` where `C` is a generic class.


accepted/future-releases/generalized-typedef-2018/feature-specification.md
```diff
@@ -114,32 +119,34 @@ evaluated as an expression, it is subject to instantiation to bound.
...
-   `D` is accessible to the current library and has no prefix.*)
+(`e is A`), and in a type cast (`e as A`).
+
+A type alias application of the form _F_ or the form
+_F&lt;T<sub>1</sub>..T<sub>k</sub>&gt;_ denoting a class can be used in a
+position where a class is expected: in the `extends`, `with`, `implements`,
+and `on` clause of a class or mixin declaration; and in an instance
+creation expression (*e.g., `new F()`, `const F<String>.named('Hello!')`*).
+
+A type alias application _A_ of the form _F_ or the form
```
leafpetersen: @lrhn Do we intend to support the actual syntax `F<T>.staticMethod` now (as opposed to just implicitly via `G.staticMethod` where `G` is a type alias for `F<T>`?  I don't particularly object, but if so we should call out that the non-typedef version is allowed now as well, since this is a change.  I presume there are no parsing issues, since named parameters essentially cover this corner of the grammar already?

lrhn: I would say no — for now.

I do want to support `something<typearg>` as a syntax in more places (`List<int>` as a type literal, `object.genericMethod<int>` as an instantiated tear-off, likely other), but I don't see an advantage in adding piecemeal syntax support for it until we do the full thing.

There is no current *benefit* to allowing the syntax `F<T>.foo` for static access (we currently only allow it only for constructor invocations); you can always do `F.foo` and get the same thing, so this would require be a parser change for no actual benefit.

eernstg: Sounds good! ;-)

I uploaded a new version where it is an error to invoke a static method using syntax like `F<int>.m(42)` where `F` is a type alias.

PR #129, opened 2018-12-11
Change type requirements for `{}` being a set literal.

lrhn:
Use `LinkedHashSet<Null>` as the comparison, not `Set<Null>`. This ensures that:
```dart
LinkedHashSet<int> s = {};
```
is valid. Since we guarantee to return a `LinkedHashSet`, that is perfectly fine code.

### DIFF
--- a/accepted/future-releases/set-literals/feature-specification.md
+++ b/accepted/future-releases/set-literals/feature-specification.md
@@ -1,5 +1,6 @@
 # Set Literals Design Document
 Author: lrn@google.com
+Version: 1.1
 
 Solution for [Set Literals Problem](http://github.com/dart-lang/language/issues/36).
 Based on feature proposal [Issue 37](http://github.com/dart-lang/language/issues/37)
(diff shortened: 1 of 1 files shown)


mit-mit:
@lrhn can we land this?

stereotype441:
> `LinkedHashSet` and `LinkedHashMap` are both defined in `dart:collection`. So far the analyzer has only needed to know about `dart:core` and `dart:async` in order to analyze code. I don't know how big the impact on the analyzer would be of requiring it to be able to also access code from `dart:collection`, but it won't be free and it will likely postpone our ability to implement this feature until sometime in Q1.

Hmm, I agree that we probably can't modify the analyzer to adapt to this change before January--most members of the analyzer team will be leaving for the holidays later this week and there's too much other work on our plate.

However, regarding the analyzer's need to know about `dart:collection`, that's actually unchanged by this proposal; even if we don't make this change, the spec already says that the analyzer should consider the "exact type" of a non-const set literal to be `LinkedHashSet<T>`.  It looks like the current analyzer implementation gets that wrong, and I've filed https://github.com/dart-lang/sdk/issues/35431 to track the bug.  It's a rare enough corner case that I don't think we necessarily need to fix it before enabling the "set literals" feature.

The change described in this PR also only affects a rare corner case.  So if the language team decides to land this PR, I think it would be fine to treat it the same way; as a bug in the analyzer that we don't necessarily need to fix before enabling the "set literals" feature.

stereotype441:
Note: I filed https://github.com/dart-lang/sdk/issues/35439 as a reminder to change the analyzer implementation.

PR #147, opened 2018-12-18
Remove support for "for" in const collections.

munificent:


stereotype441:
Thanks for the heads up!  This will make the implementation easier :)

Issue #141, opened 2018-12-17
Should Object be nullable by default?


leafpetersen:
This issue is to discuss the question of whether the `Object` type should be nullable by default when we enable NNBD types.

Arguments I've hard so far:

For:
  - Since `null` supports all of the operations that are defined on `Object`, it works perfectly well anywhere an `Object` is expected.
  - It's not clear what tangible benefit you get from being able to say that something can be any type, except Null.

Against:
  - There is no way to express the type of a non-null object without this.  So you can't write a generic class which requires only non-nullable type arguments.  Does this matter?
  - You won't get definite assignment warnings on variables of type `Object`.

cc @lrhn @munificent @eernstg

kevmoo:
From an end-user perspective, having `Object` be "different" here would be a bit confusing.

As an experiment, `JsonDecoder` now is `Converter<String, Object>`. `null` is a totally valid output.

In a world without nullable `Object`, what would the 2nd type argument be?

leafpetersen:
> In a world without nullable `Object`, what would the 2nd type argument be?

`Object?`

kevmoo:
Oh! So, `Object` would just be like `Customer` – that's what I hoped.

So put me in the against camp. 😄

leafpetersen:
A minor concern that I have with making `Object` nullable is that if we have neither non-nullable `Object` nor `T!`, then there is no user visible type that we can ascribe to `x` in the following code if we promote:

```dart
foo(Object nullable) {
  if (nullable != null) {
    var x = nullable;  
  }
}
```

On the other hand, there is essentially no difference between `Object?` and `Object!` in terms of methods available on them, so it's not all that useful to know that `x` is non-nullable.

eernstg:
My general preference is to let `Object` denote all objects other than the null object (that is, to make it non-null).

> There is no way to express the type of a non-null object without this

Right, to me that's the main reason. But there's a technical as well as a conceptual side to it.

Technically, we'd otherwise need a special exception to even denote the set 'all objects other than the null object'. As mentioned, we'd use `Object!`, but that would already be a special exception (because `!` is otherwise never used, or at least needed, on a statically resolved type).

Technically, we are also well positioned to claim that `Object` does not include the null object: Just like all other types today, they are implicitly subject to the transformation `T` --> `T | Null` almost everywhere. So the only breaking change would be that `null is Object` will now be false.

[...]

PR #151, opened 2018-12-20
Subtyping rules for nullable types and legacy migration types

leafpetersen:
This updates the informal description of the subtyping rules (including the derivation of the algorithmic syntax directed rules) for non-null by default types.  It also adds rules for handling the legacy types planned for use during the migration when non-opted-in libraries interact with opted-in libraries.

cc @stereotype441 @bwilkerson @kmillikin @lrhn

### DIFF
--- a/resources/type-system/subtyping.md
+++ b/resources/type-system/subtyping.md
@@ -65,7 +65,7 @@ type variables only occur statically (never at runtime).
 
 The type `Null` represents the type of the `null` constant.
 
-The type `Never` type is the uninhabited bottom type.
+The type `Never` represents the uninhabited bottom type.
 
 The type `T?` represents the nullable version of the type `T`, interpreted
 semantically as the union type `T | Null`.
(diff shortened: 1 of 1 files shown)


resources/type-system/subtyping.md
```diff
@@ -121,12 +134,31 @@ We say that a type `T0` is a subtype of a type `T1` (written `T0 <: T1`) when:
...
 
-- **Left Bottom**: `T0` is `Null`
+- **Left Bottom**: `T0` is `Never`
+
+- **Right Legacy Top**: `T1` is a legacy top type (i.e. `Object*`, `dynamic*`,
+  or `void*`).
+
+- **Left Legacy Null** `T0` is `Null*`
+
+- **Left and Right Legacy** `T0` is `S0*` and `T1` is `S1*` and `S0 <: S1`.
```
eernstg: This would not allow us to conclude `List<Null*> <: List<int*>`, because that will require `Null <: int`. Wouldn't that break too much existing code?

(Left Legacy and Right Legacy can't do it, either.)

resources/type-system/subtyping.md
```diff
@@ -121,12 +134,31 @@ We say that a type `T0` is a subtype of a type `T1` (written `T0 <: T1`) when:
 
 - **Right Top**: `T1` is a top type (i.e. `Object`, `dynamic`, or `void`).
```
eernstg: It's being discussed [here](https://github.com/dart-lang/language/issues/141) whether `Object` should be a top type, or it should be spelled `Object?`. If the latter is chosen, 'Right Top' should only mention `dynamic` and `void`.

resources/type-system/subtyping.md
```diff
@@ -33,9 +33,12 @@ The set of types under consideration are as follows:
 - `dynamic`
 - `void`
 - `Null`
+- `Never`
 - `Function`
 - `Future<T>`
 - `FutureOr<T>`
+- `T?`
+- `T*`
```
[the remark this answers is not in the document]
leafpetersen: I have a revision that explores how to deal with `T!`, I'll try to get back to it tomorrow.

resources/type-system/subtyping.md
leafpetersen: Yes, agreed, I realized that my treatment of `T*` in this CL is completely broken, I have revised it from first principles in a follow on.  Will send out shortly.

PR #152, opened 2018-12-20
Update roadmap with migration details.

leafpetersen:
Added a detailed summary of my understanding of what our agreed on roadmap looks like, clarifying as many details as possible.  Also added an overview of what the external migration process should look like.

cc @mit-mit @dgrove @vsmenon @srawlins @Hixie


leafpetersen:
I'm going to go ahead and land this, I'll follow up on any comments in a followup.

PR #162, opened 2019-01-04
Update the implementation plan for control flow collections

devoncarew:
- split the Grok work out from IntelliJ (they're pretty disjoint)
- refactor categories to call out where the work will happen
- also mention VS Code (its syntax highlighting grammar is based on the one from Atom; if one needs as update, they both will, and the files are similar enough that a fix into one can be used as the basis for a fix into the other)

### DIFF
--- a/accepted/future-releases/control-flow-collections/implementation-plan.md
+++ b/accepted/future-releases/control-flow-collections/implementation-plan.md
@@ -66,9 +66,15 @@ for this.
 DDC may need to support canonicalizing constant collections with spread
 operators.
 
-### IntelliJ/Grok
+### IntelliJ
 
 Update the IntelliJ parser to support the new syntax.
+
+### Grok
+
+Update Grok to handle the new AST.
+
+### Analysis server
 
 Update to use the latest analyzer with support for the feature. There are a
 handful of usability features that would be nice.
@@ -148,6 +154,11 @@ Atom. This same grammar is also used for syntax highlighting on GitHub. Update
 this to handle the new syntax.
 
 [atom]: https://github.com/dart-atom/dart
+
+### VS Code
+
+Update the syntax highlighting grammer to support the control flow syntax (and,
+likely apply a very similar diff to the Atom grammer above).
 
 ### VM
 
--- a/accepted/future-releases/control-flow-collections/implementation-plan.md
+++ b/accepted/future-releases/control-flow-collections/implementation-plan.md
@@ -157,8 +157,8 @@ this to handle the new syntax.
 
 ### VS Code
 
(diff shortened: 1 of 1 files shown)


accepted/future-releases/control-flow-collections/implementation-plan.md
```diff
@@ -66,10 +66,16 @@ for this.
 DDC may need to support canonicalizing constant collections with spread
 operators.
 
-### IntelliJ/Grok
+### IntelliJ
```
devoncarew: cc @alexander-doroshko

accepted/future-releases/control-flow-collections/implementation-plan.md
```diff
@@ -66,10 +66,16 @@ for this.
...
 
 Update the IntelliJ parser to support the new syntax.
 
+### Grok
+
+Update Grok to handle the new AST.
```
devoncarew: cc @chngsprk, fyi

accepted/future-releases/control-flow-collections/implementation-plan.md
```diff
@@ -66,10 +66,16 @@ for this.
...
+
+### Analysis server
```
devoncarew: cc @stereotype441

accepted/future-releases/control-flow-collections/implementation-plan.md
```diff
@@ -149,6 +155,11 @@ this to handle the new syntax.
 
 [atom]: https://github.com/dart-atom/dart
 
+### VS Code
```
devoncarew: cc @dantup (work hasn't started here yet; when it does, we may need to update the grammar file)

accepted/future-releases/control-flow-collections/implementation-plan.md
stereotype441: sgtm

accepted/future-releases/control-flow-collections/implementation-plan.md
```diff
@@ -149,6 +155,11 @@ this to handle the new syntax.
...
+
+Update the syntax highlighting grammer to support the control flow syntax (and,
```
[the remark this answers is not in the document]
devoncarew: done!

accepted/future-releases/control-flow-collections/implementation-plan.md
DanTup: I pasted the samples from this file into VS Code and it seems like this may be handled as we'd want without any changes:

screenshot 2019-01-08 at 10 46 15 am

PR #163, opened 2019-01-04
Remove references to constant evaluation of `for`

stereotype441:
`for` is prohibited in constant expressions.


munificent:
Thanks for catching that!

Issue #167, opened 2019-01-09
Should we do bottom-up disambiguation of set/map literals with spreads?


munificent:
This is syntactically ambiguous:

```dart
{...a}
```

It could be a map or set. Right now, the proposal follows the set literal proposal and says the context type can be used to disambiguate:

```dart
Set s = {...a}; // Set
Map m = {...}; // Map
```

It makes sense for the set literal proposal to only use the context type to disambiguate empty literals, because it's empty. What else could you use? But it seems weird for spreads:

```dart
var list = [1, 2, 3];
var set = {...list};
```

As it's proposed, this will generate a type error here. There is no context type, so the `{...list}` is inferred as a map literal. Then you get a type error trying to spread a list into a map literal.

This feels pretty bad to me. Should we also allow bottom-up inference if there isn't a context type? Would doing so cause any problems I'm not foreseeing? If not, I'm happy to update the proposal.

cc @lrhn @leafpetersen

lrhn:
We knew the rules would have to be generalized for spreads/if/for elements.

For something that is syntactically indistinguishable, we should use types to determine the meaning. Then all following steps can use the static type of the expression to see what the choice was.

I would first use *context type* if available. 
If the basetype of the context type is a subtype of `Iterable<Object>` (and it's not also a map, as usual), then let `T` be the type of `Iterable` that it actually implements.
Then it's a set literal, all sub-parts are *elements* (not *entries*), and the context type of each element is `T`.

If not, it's a map literal and each sub-part is an *entry*. For map entries, we'll probably have to talk about "key context type" and "value context type" for the map entries. We'll need a new syntactic category for map entries, because we now have sub-entries in if/for constructs.


For a non-empty set-or-map literal and the context type is not available, I'd look at the type of the individual element-or-entry expressions, recursively, and give each of them a vote.

For `... p`, if `e` has a static type that is a subtype of `Iterable<Object>` and not a subtype of `Map<Object,Object>`, then it votes *set*.
Otherwise if `e` has a static type that is a subtype of `Map<Object, Object>`, then it votes *map*.
If neither, then it doesn't vote either way.

For `if (e1) p1`, analyze p1 and use its votes.
For `if (e1) p1 else p2` analyze p1 and p2 and use all their votes.
For `for (... in e1) p1`, analyze p1 and use its votes.
For `expression` vote *set*.
For `expression: expression` vote *map*.

[...]

PR #168, opened 2019-01-09
Allow spread element types to disambiguate a set/map literal.

munificent:
This is a stab at addressing #167. Thoughts?

### DIFF
--- a/accepted/future-releases/spread-collections/feature-specification.md
+++ b/accepted/future-releases/spread-collections/feature-specification.md
@@ -331,18 +331,28 @@ An expression like:
 ```
 
 Is syntactically parsed as `mapOrSetLiteral`. To determine whether it actually
-is a map or set, the surrounding context is used. Given an `mapOrSetLiteral`
-with context type C:
+is a map or set, the upwards and downwards type inference contexts are used.
 
-*   If `Set<Null>` is assignable to C, and `Map<Null, Null>` is not assignable
-    to C, then the collection is a set literal.
+Given a `mapOrSetLiteral`:
 
-*  Otherwise, it is a map literal.
+1.  If `Map<Null, Null>` is assignable to the literal's context type, then it
+    is a map literal.
+
+2.  Else, if `Set<Null>` is assignable to the literal's context type, then the
+    collection is a set literal.
+
+3.  Else, if it has spread elements and all of the spread element expression
+    types are subtypes of `Iterable`, then it is a set literal.
+
+    **TODO: Is dynamic a subtype of Iterable? If so, this needs to be tweaked.
+    If all the spreads are dynamic, then it should be a map literal.**
+
+4.  Otherwise, it is a map literal.
 
 In other words, if it can only be a set, it is. Otherwise, it's a map. In cases
-where the context type is not specific enough to disambiguate, we could make it
-an error instead of defaulting to map. However, that would be inconsistent with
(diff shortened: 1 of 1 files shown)


accepted/future-releases/spread-collections/feature-specification.md
```diff
@@ -331,18 +331,28 @@ An expression like:
...
 
-*   If `Set<Null>` is assignable to C, and `Map<Null, Null>` is not assignable
-    to C, then the collection is a set literal.
+Given a `mapOrSetLiteral`:
 
-*  Otherwise, it is a map literal.
+1.  If `Map<Null, Null>` is assignable to the literal's context type, then it
+    is a map literal.
+
+2.  Else, if `Set<Null>` is assignable to the literal's context type, then the
```
leafpetersen: Note that this is not the criteria that is specified for empty set literals, but there's active discussion about this here: https://github.com/dart-lang/sdk/issues/35431

accepted/future-releases/spread-collections/feature-specification.md
```diff
@@ -331,18 +331,28 @@ An expression like:
...
+    collection is a set literal.
+
+3.  Else, if it has spread elements and all of the spread element expression
```
leafpetersen: I would phrase this differently.  I think it's probably better to say that that *if* there is a non-empty context, you resolve based on the context type (as above, or however we decide).   Only if there is no context type do you resolve based on the upwards inference.  Otherwise, I think it's guaranteed to be an error anyway, and you're just making things confusing by still trying to use upwards inference.

accepted/future-releases/spread-collections/feature-specification.md
```diff
@@ -331,18 +331,28 @@ An expression like:
...
+    types are subtypes of `Iterable`, then it is a set literal.
+
+    **TODO: Is dynamic a subtype of Iterable? If so, this needs to be tweaked.
```
leafpetersen: No, it's not.

accepted/future-releases/spread-collections/feature-specification.md
lrhn: Agree. You only need to address the case where there is no context type and no type arguments. The rest are already handled and is unaffected by spreads or comprehensions.

I think we need to rewrite everything around this so that we only have "setOrMapLiteral" as a syntactic category. It's going to be complicated to have:
```
setLiteral ::= `{` setElements `}`
mapLiteral ::= `{ mapEntries `}`
mapEntry ::= expression `:` expression |  `...` expression | `if` `(` expression `)` mapEntry (`else` mapEntry)? | `for` `(` ... `in` expression `)` mapEntry ;
setElement ::= expression  |  `...` expression | `if` `(` expression `)` setElement (`else` setElement)? | `for` `(` ... `in` expression `)` setElement ;
setOrMapPart ::= |  `...` expression | `if` `(` expression `)` setOrMapPart (`else` setOrMapPart)? | `for` `(` ... `in` expression `)` setOrMapPart ;
```
because they *do* overlap unless deep-down in the structure there is either an `e` element or an `e1:e2` entry.

So, I'd go for unifying sets and maps, and then use the context type or content parts to determine which one it is. And I wouldn't separate spreads and if/for constructs, they both need the same treatment anyway.

Let me try doing that.

munificent:
I'm gonna go ahead and close this in favor of the Grand Unified Map/Set doc that's coming.

PR #191, opened 2019-01-23
Tweak type inference on spread.

munificent:
- Specify that the element type is dynamic when spreading dynamic.

- Make it a compile-time error to spread Object.

### DIFF
--- a/accepted/future-releases/spread-collections/feature-specification.md
+++ b/accepted/future-releases/spread-collections/feature-specification.md
@@ -317,6 +317,10 @@ It is a static error if:
     and `V` and `K` is not assignable to the key type of the map or `V` is not
     assignable to the value type of the map.
 
+*   A spread element has static type `Object`. *We don't allow implicit
+    downcasting from `Object` to some `Iterable` or `Map` type because we may
+    not know the element, key, or value types to cast it to.*
+
 If implicit downcasts are disabled, then the "is assignable to" parts here
 become strict subtype checks instead.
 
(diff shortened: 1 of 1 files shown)


accepted/future-releases/spread-collections/feature-specification.md
```diff
@@ -317,6 +317,10 @@ It is a static error if:
     and `V` and `K` is not assignable to the key type of the map or `V` is not
     assignable to the value type of the map.
 
+*   A spread element has static type `Object`. *We don't allow implicit
```
lrhn: It is suspicious to single out `Object`. 
I assume it's because we know that it's a type that both `Iterable<X>` and `Map<Y,Z>` are assignable to, but so is `FutureOr<Object>`.

So, if this generalizes, I'd say that we don't allow *downcast* at all. We do allow `dynamic` because we will handle the type dynamically, but we don't allow `Object`+implicit downcast, or `FutureOr<Object>` or `FutureOr<dynamic>`?

(I'm not sure it's true that we don't know the down-cast type. If we decide otherwise that this is a `Set`, then we will infer a `T` so that it's a `Set<T>`, and that is the `T` we will add an `as Iterable<T>` for at the downcast.
Is the problem that we don't know *yet* at the point where we are doing the inference, and we want to insert the implicit downcast at that point?)

munificent: Oof, yeah. I completely forgot about FutureOr<T>.

There are a couple of pieces I have in mind:

* I don't think you should be able to spread Object.
* You should be able to spread dynamic.
* It's OK to spread an Iterable whose element type is a supertype of the containing collection. In that case, it will downcast each element. So this is allowed:

    ```dart
    List<num> nums = [1, 2, 3];
    List<int> ints = [...nums];
    ```

* It's not OK to spread an Iterable of an unrelated type. This is not allowed:

    ```dart
    List<Strings> strings = ["s", "t", "r"];
    List<int> ints = [...strings]; // ERROR. WAT.
    ```

I *think* what I'm describing here basically mirrors how Dart 2 handles a for-in statement `for (T i in f)` where:

* Spread expression `e` :: Iterable expression `f`
* Collection element type :: Iteration variable type `T`

At least, that was my intent. Took another stab at the prose. Let me know what you think.

PR #197, opened 2019-01-29
NNBD per library opt out

danrubel:
This updates the NNBD roadmap with a description of how a library may be opted **out** within a package that has been opted into NNBD.

### DIFF
--- a/working/0110-incremental-sound-nnbd/roadmap.md
+++ b/working/0110-incremental-sound-nnbd/roadmap.md
@@ -95,14 +95,14 @@ these checks should be provably redundant and do not need to be inserted.
 
 Concretely, opting in to weak null checking will proceed as follows:
 - The programmer opts into the weak null checks by marking their package as
-  opted in.
+  opted in. They can opt out of weak null checks on a per library basis.
 - This will cause warnings about null-safety violations within the package (only).
   - If the opted-in code uses other packages that have opted in, then the
     programmer will see null-safety warnings from any misuses of the opted in
     APIs from the other packages.
-  - If the opted-in code uses other packages that have not opted in, the
-    programmer will not see any null-safety warnings from uses of those APIs.
-    They are free to treat those APIs as being null-accepting or
+  - If the opted-in code uses code in libraries or packages that have not been
+    opted in, the programmer will not see any null-safety warnings from uses
+    of those APIs. They are free to treat those APIs as being null-accepting or
     non-null-accepting as appropriate.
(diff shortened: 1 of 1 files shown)


working/0110-incremental-sound-nnbd/roadmap.md
```diff
@@ -95,14 +95,14 @@ these checks should be provably redundant and do not need to be inserted.
 
 Concretely, opting in to weak null checking will proceed as follows:
 - The programmer opts into the weak null checks by marking their package as
-  opted in.
+  opted in. They can opt out of weak null checks on a per library basis.
```
lrhn: I'd expect this to also opt them out of the syntax. It's opting out of the entire feature, not just static checking.

working/0110-incremental-sound-nnbd/roadmap.md
```diff
@@ -175,8 +175,15 @@ and faster code.
 
 Non-nullable types will be controlled by a language opt-in as
 described [elsewhere](https://github.com/dart-lang/language/issues/93).  In
-short, a library may opt in via syntax in the code, or a package may opt in in
-entirety.  Opting in applies only to the library or package so marked.
+short, a package may be opted-in and specific libraries within that package
+may be opted-out.
+
+Within a package that has been opted-in, specific libraries may be opted-out
+by adding `@pragma('nnbd-legacy')` to that library's `library` directive.
```
lrhn: Interesting choice. I'm not particularly happy about using pragmas that have semantic effects, but it would allow us to get the feature without adding any new language features for version-opt-in.

(We should allow library declarations with no name, so you can add annotations and comments on the library without having to give it a unique name).

leafpetersen:
I commented on #94 about the open questions, including this proposal to use `@pragma`.  I'd suggest we try to resolve via discussion there.

working/0110-incremental-sound-nnbd/roadmap.md
danrubel: Agreed. That is my intent. I'll update the PR to indicate as much.

working/0110-incremental-sound-nnbd/roadmap.md
danrubel: Per comments in https://github.com/dart-lang/language/issues/94, perhaps there is a better choice than @pragma. I'm happy with whatever mechanism is decided, but think it's important to have the ability to opt-out a library.

danrubel:
Dropping this for now until we figure out if and how this can be accomplished.

PR #198, opened 2019-01-29
Corrected legacy subtyping

leafpetersen:
The previous development of legacy subtyping didn't give all the right equations.  I think this one does the right thing, based on the interpretation of `T*` as behaving as either of `T` or `T?`.


leafpetersen:
cc @munificent @stereotype441

leafpetersen:
cc @kmillikin

PR #201, opened 2019-02-01
add lint rule deliverable

pq:
Adds a section for the linter, suggesting a rule that flags opportunities to use spread operators.

See also: https://github.com/dart-lang/linter/issues/1402.

/cc @munificent @leafpetersen


munificent:
Thanks!

PR #220, opened 2019-02-13
Add new spread benchmark and write up analysis of results.

munificent:
I was feeling stuck without enough information to make an informed recommendation, so I went ahead and got as much benchmark data and tried my best to understand what the numbers were saying. Here's what I got.

What do you think? If you're OK with it, I'll also run it by the leads and see what they think. (And also so they can bikeshed my benchmarking methodology, naturally. 😄)

### DIFF
--- a/accepted/future-releases/spread-collections/performance.md
+++ b/accepted/future-releases/spread-collections/performance.md
@@ -21,6 +21,17 @@ summzarize the trends.
 
 *   Desugar maps to `Map.keys` and `Map.[]`.
 
+**Decided: After discussing this with the language leads, we've agree to:**
+
+*   Keep desugaring iterables to `iterator`, but give implementations the
+    *option* of using `length` and `[]` if the iterable is a List, Queue, or
+    Set.
+
+*   Keep desugaring maps to `entries`. We think user-defined Map classes are
+    pretty rare, as is spreading maps. Implementations can always apply
+    whatever optimizations they want when they detect that a map is a built-in
+    map type.
+
 The benchmark is [here][code], and the full data is [here][sheet] (not visible
 outside of Google, sorry).
 
--- a/accepted/future-releases/spread-collections/performance.md
+++ b/accepted/future-releases/spread-collections/performance.md
@@ -1,0 +1,400 @@
+A key part of the spread proposal is the `...` syntax "desugars" to calls to a
+public API on the object being spread. This is important because it allows users
+to spread their own collection types. In other words, it makes the spread syntax
+programmable.
+
+That raises the question of *what* public API it should call. The API should be
+simple to implement, but also efficient to execute. Whatever it desugars to,
+we'll be stuck with forever. Since it's calling into user code which can have
+arbitrary side effects, any change to the semantics could be breaking.
+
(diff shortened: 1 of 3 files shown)


accepted/future-releases/spread-collections/performance.md
```diff
@@ -0,0 +1,400 @@
...
+    internal iterator may paint us into a corner with future control flow
+    changes. This isn't good enough to win out past that, even ignoring the poor
+    JIT performance.
+
+*   `List.[]` is much faster when the collection is empty: 3x Aot, 3x dart2js
+    and 10x on the JIT (!). This is probably showing the benefit of not
+    allocating the iterator. It's somewhat faster for non-empty collections too:
+    1.7x AoT, 1.2x dart2js, 1.3x JIT.
+
+    This looks pretty nice. I think empty collections will happen fairly
```
lrhn: If we allow the iteration to do `isEmpty` (or `isNotEmpty` or `length`) on the spread collection, then we can special-case the empty collection too.

accepted/future-releases/spread-collections/performance.md
```diff
@@ -0,0 +1,400 @@
...
+collection types. This produces something like a 5-dimensional hypercube of
+configurations, so I sliced the data a bunch of different ways and tried to
+summzarize the trends.
+
+**TL;DR: I recommend, we:**
+
+*   Desugar to `List.length` and `List.[]` at compile time if we statically know
+    the Iterable object being spread implements List.
+
+*   Desugar maps to `Map.keys` and `Map.[]`.
```
lrhn: Unless we can encourage all map implementations to make this *efficient* (not do an extra lookup), I really don't want to make it the mandatory approach.

accepted/future-releases/spread-collections/performance.md
munificent: I did try that out but then realized this isn't a safe operation. If you call `isEmpty` on a WhereIterable before iterating it, for example, it will invoke the filter closure for the elements multiple times, which may have side effects.

accepted/future-releases/spread-collections/performance.md
```diff
@@ -0,0 +1,400 @@
...
+justify it. I don't see that here.
+
+**Recommendation: `Map.keys` is clearly faster and is more idiomatic. Use that
+instead of `Map.entries`.**
+
+## Protocols
+
+I considered another option: We could define three *protocols*: "iterable",
+"list", and "map". Each defines how a conforming implementation of the
+corresponding classes behaves. For example, if `Iterable.isEmpty` returns
```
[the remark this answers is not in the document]
munificent: Sure, but the protocol is about what the methods on those interfaces *mean*, not just their static signatures. It's things like "if `length` returns *x*, then `iterator` should return an Iterator where calling `moveNext()` returns `true` *x* times and then `false".

munificent:
Thanks for all the comments and discussion this morning! I added a brief section explaining what we've decided to do and then I'm going to go ahead and land it as it is. It represents a piece of investigation work and not a canonical source of truth, so I think it's OK if not everything is resolved.

PR #231, opened 2019-02-21
Write up summary of optional semicolon prototypes.

munificent:
There's nothing actionable about this doc, so it doesn't need to be perfect. But I think it's worth landing to record my thoughts before they evaporate.


mit-mit:
Looks great, thanks very much for writing this!

munificent:
OK, I'm just going to go ahead and merge it. :)

PR #233, opened 2019-02-25
Moving dartLangSpec.tex to this repo

eernstg:
This is the branch https://dart-review.googlesource.com/c/sdk/+/93540, that is, it includes the single-line change that disables the generalized type aliases.


eernstg:
Note that this setup does not yet support all make goals, but the basic `make pdf` (which is also the default `make`) works.

PR #237, opened 2019-02-25
Small tweaks to .travis.yml

kevmoo:


mit-mit:
Nice, like the grep change :-)

PR #240, opened 2019-02-25
Only run on updates to the master branch

kevmoo:
PRs are covered separately


mit-mit:
Ah, so this will still run if someone sends a PR on the `.tex` file?

kevmoo:
Yup yup, but no double-build for PRs that are also branches in this repo.

PR #246, opened 2019-02-26
Introduce context types (was Gerrit CL 92782)

eernstg:
We currently (Aug 2019) have 65 occurrences of `context type` in the language specification, but it is never defined. This PR adds a section where that concept is defined in terms of a list of cases.

This PR is intended to introduce context types as a stand-alone concept, such that we can talk about the context type in other parts of the specification. In loose terms, we could say that we assume that type inference has already been completed (that is actually what the spec says already), leaving certain parts of the program untouched. E.g., when a generic function is used where the context type is a non-generic function, we can then specify that generic function instantiation takes place (again, this is what we already do).

This means that the current PR will improve on the completeness of the specification, but we will still need to make adjustments when we specify type inference in total.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -5,7 +5,6 @@
 \usepackage{syntax}
 \usepackage{amssymb}
 \usepackage[fleqn]{amsmath}
-\usepackage{amssymb}
 \usepackage{semantic}
 \usepackage{dart}
 \usepackage{hyperref}
@@ -47,7 +46,7 @@
 %   type, not just function types.
 % - Clarify that 'Constant Constructors' is concerned with non-redirecting
 %   generative constructors only.
-% - Disable type alias generalization, because that feature is not in 2.2.
+% - Specify the notion of a context type.
 %
 % 2.1
 % - Remove 64-bit constraint on integer literals compiled to JavaScript numbers.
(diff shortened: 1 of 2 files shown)


specification/dartLangSpec.tex
```diff
@@ -7427,7 +7428,7 @@ \subsection{Sets}
 \item
 For each $i \in 1 .. n$ in numeric order,
 the expression $e_i$ is evaluated producing object $v_i$.
-\item A fresh object (\ref{generativeConstructors}) $s$
+\item A fresh instance (\ref{generativeConstructors}) $s$
```
eernstg: Reason for change: This line is the only occurrence of `fresh object`, it's `fresh instance` everywhere else.

specification/dartLangSpec.tex
```diff
@@ -15671,7 +16105,7 @@ \subsection{Type Void}
   }
 \item
   \commentary{%
-  In an arrow function body \code{=> $e$},
+  In an arrow function body `\code{=> $e$}',
```
eernstg: Reason for change: For readability, we use single quotes around some source code snippets that may otherwise be hard to read. Here: `=>` and `e` separated by a space is not obviously a thing (it looks like two things).

PR #249, opened 2019-02-27
Cleaned up "return with" such that we use "object" rather than "value"

eernstg:


specification/dartLangSpec.tex
```diff
@@ -6794,12 +6796,14 @@ \subsubsection{String Interpolation}
...
 \begin{itemize}
-\item{} Evaluate $e_i$ to an object $o_i$.
-\item{} Invoke the \code{toString} method on $o_i$ with no arguments, and let
- $r_i$ be the returned value.
-\item{} If $r_i$ is not an instance of the built-in type \code{String}, throw an \code{Error}.
+\item Evaluate $e_i$ to an object $o_i$.
+\item Invoke the \code{toString} method on $o_i$ with no arguments,
+  and let $r_i$ be the returned object.
+\item If $r_i$ is not an instance of the built-in type \code{String},
+  throw an \code{Error}.
```
lrhn: In Dart 2, that has to mean that the returned value is `null`, because the return type of `Object.toString` is `String`.
(In Dart 2.x, the return type will be non-nullable `String` and this can be removed completely).

specification/dartLangSpec.tex
```diff
@@ -8862,16 +8888,16 @@ \subsubsection{Function Expression Invocation}
 \begin{itemize}
 \item \code{$im$.isMethod} evaluates to \code{\TRUE{}}.
 \item \code{$im$.memberName} evaluates to the symbol \code{\#call}.
-\item \code{$im$.positionalArguments} evaluates to an unmodifiable list with the values
-resulting from evaluation of
-\code{<Object>[$a_1, \ldots,\ a_n$]}.
+\item \code{$im$.positionalArguments} evaluates to an unmodifiable list with the objects
```
lrhn: -> containing the objects

(like below).

specification/dartLangSpec.tex
```diff
@@ -1214,9 +1214,9 @@ \section{Functions}
 
   \commentary{%
   Because Dart supports dynamic function invocations,
-  we cannot guarantee that a function that does not return a value
+  we cannot guarantee that a function that does not return an object
   will not be used in the context of an expression.
-  Therefore, every function must return a value.
+  Therefore, every function must return an object.
```
lrhn: -> must either throw or return an object.

specification/dartLangSpec.tex
```diff
@@ -11779,7 +11811,7 @@ \subsubsection{Statement Completion}
 then, unless otherwise stated, the execution of $s$ stops
 at that point and completes in the same way.
 \commentary{%
-For example, if execution of the body of a \DO{} loop returns a value,
+For example, if execution of the body of a \DO{} loop returns an object,
```
lrhn: Below there is an `\Q{... will be inserted here}`. If that is visible, make it a comment and a TODO. It's either there or not, it has no value to eventually be there.

specification/dartLangSpec.tex
eernstg: Adjusted to say `If $r_i$ is the null object, a dynamic error occurs.`

specification/dartLangSpec.tex

specification/dartLangSpec.tex

specification/dartLangSpec.tex
eernstg: There is a TODO comment already, and the `\Q{...}` _is_ visible, so I just removed the `\Q{...}`.

PR #255, opened 2019-03-06
Restored change from error to warning from CL 85959

eernstg:
The change from 'compile-time error' to 'static warning' that we performed in CL 85959 was accidentally lost when 'invalid_returns.md' was integrated into the language specification. This CL just changes those few words back to 'static warning' (and removes a couple of spurious `{}`).

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -12638,7 +12638,7 @@ There are obviously situations where code does not fall through, and yet does no
 
 \begin{dartCode}
 \SWITCH{} (x) \{
-  \CASE{} 1: \TRY{} \{ $\ldots$ \RETURN{}; \} \FINALLY{} \{ $\ldots$ \RETURN{}; \}
+  \CASE{} 1: \TRY{} \{ $\ldots$ \RETURN; \} \FINALLY{} \{ $\ldots$ \RETURN; \}
 \}
 \end{dartCode}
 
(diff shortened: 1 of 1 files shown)


lrhn:
I still think this should be an error. 
@leafpetersen

eernstg:
Given that tools have reported an error for this situation for a while (we currently get expectation 'StaticWarning' and actual outcome 'CompileTimeError' for, e.g., analyzer-linux-release           language_2/void/return_future_future_or_void_async_error1_test/none), we seem to be in a good position to just keep the specification unchanged (so it remains an error, now per decision rather than by accident), that is: we'd just delete this pull request.

We might want to consider all the _other_ warnings an extra time (maybe some of them can also be errors), but in that case we'd need to change the specification, so that would be done in a separate PR.

leafpetersen:
> (we currently get expectation 'StaticWarning' and actual outcome 'CompileTimeError' for, e.g., analyzer-linux-release language_2/void/return_future_future_or_void_async_error1_test/none)

This seems to have been an (accidental?) regression between 2.0 (where the analyzer reports a warning) and 2.1 (where it is an error).   Maybe reasonable to move forward with it since it's been this way for a while?  Not great that this slipped through though.

eernstg:
Abandoned. At this point we're more likely to change even more warnings to errors, rather than changing these two errors to warnings.

PR #261, opened 2019-03-11
Add extension method survey as markdown document.

lrhn:


working/0041 - static extension methods/lrn-static-extension-survey.md
```diff
@@ -0,0 +1,177 @@
...
+
+We will look to other languages for inspiration, but Dart is fairly unique in having reified type arguments and _not_ having overloading on method signature. Solutions that work for other languages may not fit Dart as well.
+
+
+# Other Languages
+
+A number of other languages have something that can be described as static extension methods.
+
+
+## C#
```
munificent: Might be worth pointing out that C# is working on adding support for extension getters, static methods, etc.: https://github.com/dotnet/roslyn/issues/11159

working/0041 - static extension methods/lrn-static-extension-survey.md
```diff
@@ -0,0 +1,177 @@
...
+{
+    Console.WriteLine(input.ToString());
+    return input;
+}
+```
+
+
+Since C# implements generics by specialization for value types and it reifies the type for reference types, the type `T` here will be the actual type of the receiver.
+
+Only non-nullable value types (not reference types) can be made "nullable".
```
munificent: C# 8.0 adds support for nullable reference types: https://devblogs.microsoft.com/dotnet/nullable-reference-types-in-csharp/

PR #269, opened 2019-03-14
Clarify spec for const ifElements.

munificent:
cc @MichaelRFairhurst

### DIFF
--- a/accepted/future-releases/unified-collections/feature-specification.md
+++ b/accepted/future-releases/unified-collections/feature-specification.md
@@ -654,12 +654,12 @@ They are defined as:
 *   An `ifElement` is a constant element if its condition is a constant
     expression evaluating to a value of type `bool` and either:
 
-    *   If the condition evaluates to `true`, then when the "then" element is a
-        constant expression and the "else" element, if it exists, is a
+    *   The condition evaluates to `true`, then when the "then" element is a
+        constant expression, and the "else" element (if it exists) is a
         potentially constant element.
 
-    *   If the condition evaluates to `false`, then when the "then" element is a
-        potentially constant element and the "else" element, if it exists, is a
+    *   The condition evaluates to `false`, then when the "then" element is a
+        potentially constant element, and the "else" element (if it exists) is a
         constant element.
 
 *   An `ifElement` is potentially constant if its condition, "then" element, and
--- a/accepted/future-releases/unified-collections/feature-specification.md
+++ b/accepted/future-releases/unified-collections/feature-specification.md
@@ -655,12 +655,12 @@ They are defined as:
     expression evaluating to a value of type `bool` and either:
 
     *   If the condition evaluates to `true`, then when the "then" element is a
-        constant expression and any "else" element is a potentially constant
-        element.
+        constant expression and the "else" element, if it exists, is a
+        potentially constant element.
 
     *   If the condition evaluates to `false`, then when the "then" element is a
-        potentially constant element and any "else" element is a constant
-        element.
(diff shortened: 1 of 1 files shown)


accepted/future-releases/unified-collections/feature-specification.md
```diff
@@ -655,12 +655,12 @@ They are defined as:
     expression evaluating to a value of type `bool` and either:
 
     *   If the condition evaluates to `true`, then when the "then" element is a
```
MichaelRFairhurst: I'm not sure the "when" and "if" are necessary here.

I think this sentence boils, logically, down to:

```
constIfElement = constBoolCondition &&
  (
    (conditionTrue && constThenElement && potentiallyConstElseElement)
    ||
    (conditionFalse && constElseElement && potentiallyConstThenElement)
  )
)
```

So I think in these "either" blocks you have three predicates ANDed together, and it might be clearer to write it like that then as an "if when"

Even though that's valid too, it requires imperative instead of functional programming:

```
isValid = constBoolCondition;
if (isValid) {
  switch(conditionBlockValue) {
    case true: return constThenElement && potentiallyConstElseElement;
    case false: return constElseElement && potentiallyConstThenElement;
  }
}
return isValid;
```

Which approach implementors use is up to them, but I think functional programming in EnglishSpecLanguage++ is clearer than imperative programming in EnglishSpecLanguage++.

Unless I'm spouting nonsense here.

munificent: Your formulation is about right, except that it doesn't handle the else clauses being absent. In particular, it should be clear that omitting the else clause does *not* mean that neither case matches and it's impossible for an ifElement to be constant. Also, I'm not really sure how to map a switch to English. :)

I think the change here is at least a marginal improvement.

leafpetersen: I actually agree this is a mess to read.  It currently says:

An A is B if C and either:
  if X then when Y and Z.

To which I scratch my head and say "if X then when Y and Z what?"

Can we just rewrite this to:
```
 ... and either:
      *  The condition evaluates to `true`; the "then" element is a constant expression; and the "else" element, if it exists, is a potentially constant element.
      *  Or the condition evaluates to false....
```

munificent: Done.

accepted/future-releases/unified-collections/feature-specification.md
```diff
@@ -654,13 +654,13 @@ They are defined as:
 *   An `ifElement` is a constant element if its condition is a constant
     expression evaluating to a value of type `bool` and either:
 
-    *   If the condition evaluates to `true`, then when the "then" element is a
-        constant expression and any "else" element is a potentially constant
-        element.
+    *   The condition evaluates to `true`, then when the "then" element is a
```
leafpetersen: delete "then when"?

accepted/future-releases/unified-collections/feature-specification.md
```diff
@@ -654,13 +654,13 @@ They are defined as:
...
+        constant expression, and the "else" element (if it exists) is a
+        potentially constant element.
 
-    *   If the condition evaluates to `false`, then when the "then" element is a
-        potentially constant element and any "else" element is a constant
-        element.
+    *   The condition evaluates to `false`, then when the "then" element is a
```
leafpetersen: delete "then when"?

munificent: Oops, meant to. Sorry.

PR #282, opened 2019-03-21
Feedback on language version selection strawman

munificent:
I wasn't sure the best way to provide feedback on the strawman.

So I stuffed my notes into the doc itself. That way I can send out a
PR that shows my notes and then in the PR, you and others can reply and
discuss.

I don't intend to land this.

### DIFF
--- a/working/0093 - Gradual Language Change Migration/0094 - Per Library Language Version Selection/strawman.md
+++ b/working/0093 - Gradual Language Change Migration/0094 - Per Library Language Version Selection/strawman.md
@@ -12,9 +12,9 @@ However, since Dart 2.0, a breaking language change is unlikely to be practicall
 
 Breaking the language will require a large (and growing) body of code to be migrated.
 
-We still want to do breaking changes, because some language changes are worth it. They just have to be worth _a lot_ to be viable. 
+We still want to do breaking changes, because some language changes are worth it. They just have to be worth _a lot_ to be viable.
 
-In order to improve the cost/benefit ratio, we can also try to reduce the _cost_. 
+In order to improve the cost/benefit ratio, we can also try to reduce the _cost_.
 
 One approach, which we should definitely invest in, is to create tools to migrate code from one version of the language to another. Such a tool should be able to take any library or package and migrate it seamlessly to a new version of the language. Users of an unmigrated package can then migrate it themselves, perhaps aided by `pub`, when they choose to depend on the package.
 
@@ -24,11 +24,11 @@ This document describes another approach to reducing the migration cost of intro
 
 The end goal is still to migrate all code to the newest language version.
 
-The goal of this feature is not to make future or experimental features available. We use flags for that. 
+The goal of this feature is not to make future or experimental features available. We use flags for that.
 It is not intended to do conditional compilation based on SDK version.
-This feature is intended for shipping code running against released SDKs and language versions, 
-allowing unmigrated code targeting an older language version running alongside migrated code 
-for the newest language version supported by the available SDK. 
+This feature is intended for shipping code running against released SDKs and language versions,
+allowing unmigrated code targeting an older language version running alongside migrated code
+for the newest language version supported by the available SDK.
 
 ## Language Versioning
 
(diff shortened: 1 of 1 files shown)


working/0093 - Gradual Language Change Migration/0094 - Per Library Language Version Selection/strawman.md
```diff
@@ -38,7 +38,17 @@ A package selects the _language version level_ it expects using its `pubspec.yam
 A Dart SDK tool must reject a any version higher than what the SDK supports. A Dart SDK may also have a lowest version that it
 supports being backwards compatible with.
 
-We will allow an individual library to select a lower language version level than the one selected for its package, for the case where that library has not yet been migrated to the newer semantics. 
+**Bob: This is unclear, since the SDK constraint specifies a *range* of
+versions. I think what this should do is that the *minimum* version of the
+constraint determines the default language level all code in the package can
+assume. If a Dart SDK tool's version is not within the constraint at all, that's
+an error.**
```
eernstg: I suppose there would be a computation which takes an SDK constraint from pubspec.yaml (and maybe there can also be other sources) and computes a language version level. Maybe it's as simple as "from a range, take the minimum; if that minimum is x.y.z then the language version level is x.y".

But I agree that this computation should be mentioned explicitly.

working/0093 - Gradual Language Change Migration/0094 - Per Library Language Version Selection/strawman.md
```diff
@@ -38,7 +38,17 @@ A package selects the _language version level_ it expects using its `pubspec.yam
...
+
+We will allow an individual library to select a lower language version level than the one selected for its package, for the case where that library has not yet been migrated to the newer semantics.
+
+**Bob: Should tools make it an error for a library to select a *higher* version?
+Doing so will always cause problems because users of older SDKs could end up
+with this library but not be able to use it.**
```
eernstg: I believe line 119 confirms this. Maybe it would be nice to move that sentence up here.

working/0093 - Gradual Language Change Migration/0094 - Per Library Language Version Selection/strawman.md
```diff
@@ -50,6 +60,16 @@ library foo with 2.2;
 
 ```
 
+**Bob: This looks nice, but to the lexer, that's a floating point literal. It would be Very Bad if an implementation couldn't tell "2.2" and "2.20" apart. Should we quote it "2.2" just to be safe?**
```
eernstg: Or we could just decide that language level versions are integers, which would also make it even more obvious that it's not the same thing as the pubspec SDK constraint.

working/0093 - Gradual Language Change Migration/0094 - Per Library Language Version Selection/strawman.md
```diff
@@ -78,14 +118,28 @@ The `pub` tool configures this default language level for a package based on its
...
 
 (If we allow un-packaged code to act as if it was in a package, e.g., for testing, it will inherit the package's language level, which is reasonable since that code is usually in the same pub package, just not in the exported Dart package).
 
 It *is* an option to only support per-package versioning, and not have a per library versioning option. It may increase the migration cost for a package, which encourages putting it off until the latest possible moment, but it may also encourage *completing* a migration sooner instead of leaving a few less-important files hanging.
 
+**Bob: One concern is that package authors are disincentivized from migrating.
+To migrate, they logically must raise the minimum SDK constraint, but that in
+turn makes it impossible for users on older SDKs to use their package.
+Hopefully, the new features are delightful enough to mitigate this, but we
+should keep it in mind.**
```
eernstg: If package P1 (which allows for language version levels k1..k2) imports something from package P2 (which allows for language version levels m1..m2) then it seems reasonable to use language level n which is some number in the intersection of k1..k2 and m1..m2 (as opposed to insisting on level k1 and then failing if k1 isn't in the range m1..m2).

Is it even likely that we would be able to allow for any kind of version mix without going into version resolution of a similar kind as that of pub?

mit-mit:
cc @lrhn @munificent what the next step here?

munificent:
No action was needed, it was just a way to provide feedback on the landed doc.

Issue #170, opened 2019-01-10
Type Patterns


eernstg:
In response to #169, this is a proposal for a mechanism called _type patterns_. A type pattern _P_ may contain certain syntactic elements that introduce a new type variable (look for `var`), and it supports a check for whether a given type _matches_ the pattern _P_.

## Examples

The basic construct defined in this proposal is a type patterns clause. Here are some examples of type patterns clauses (and a special case thereof called a type pattern), one per line, along with some comments giving hints about the meaning of the given construct.

```dart
// Match a single, fixed type with a type pattern.
int
List<int>
void Function()

// Match several fixed types with a type patterns clause.
A, B
Iterable<int>, List<dynamic>

// Match any type, and bind `X` to it.
var X

// Match any type `T` which satisfies `T <: C<T>`, and bind `X` to it.
var X extends C<X>

// Match any type of the form `Iterable<T>` and bind `X` to `T`;
// also, match `List<T>` and bind `X` to `T`.
Iterable<var X>

// Match any type of the form `T Function()` and bind `X` to `T`;
// also, match `T function([int])` and bind `X` to `T`.
var X Function()

// Match any type of the form `T Function(S, [int])`,
// binding `X` to `T` and `Y` to `S`;
// also, match `num Function(double, [num, num])`,
// binding `X` to `num` and `Y` to `double`.
var X Function(var Y, [int])

// Error: cannot introduce two type variables with the same name
// in one type patterns clause.
var X, List<var X>

// Match `Map<num, num>` and bind `X` to `num`;
// match `Map<num, int>` and bind `X` to `num`;
// do not match `Map<int, num>`.
Map<var X, X>

// Match `List<int>`, binding `X` to `int`; do not match `List<String>`.
List<var X extends num>
```

First note that every type is also a type pattern, so it is always possible to use any particular type as a pattern. The intuition behind this is that it matches the specified type.

Next, when a type is a subtype of a type pattern which is also derivable from `<type>`, it also matches. So `int` matches `num`, and all types match `dynamic`.

Finally, a type pattern containing a primitive type pattern (that is, a type pattern where `var` occurs) matches the types as described in the comments above, which generally means that `T` matches `P` when it is possible to choose a value for each type variable introduced by a primitive type pattern in `P`, and `T` matches the result.

## Syntax

This proposal will make use of the following extensions to to Dart grammar:

```
<typePatterns> ::= <typePattern> (',' <typePattern>)*

<typePattern> ::= <functionTypePatternTails>
[...]

PR #284, opened 2019-03-22
Type-pattern based feature spec for scoped static extension methods

eernstg:
This PR adds the file 'feature-specification-using-type-patterns.md' to the directory which is concerned with static scoped extension methods.

It is an initial version of a feature specification of static scoped extension methods where only the core is covered (the extension methods themselves and their static and dynamic semantics, but no static methods or other "auxiliary" features that extensions could have).

It is based on using type patterns (dart-lang/language#170) to specify how to extract the actual type arguments of the given receiver, and it should contain enough detail to justify the claim that we can implement this feature, with this semantics and static analysis, and trust that it will not introduce soundness issues.


eernstg:
Abandoned: We chose to use an approach based on type inference rather than type patterns, and also to use the statically known receiver type rather than the dynamic type.

PR #291, opened 2019-03-29
Adjusted spec of For-In to insist on Iterable

eernstg:
The specification of a for-in statement used to rely on having an `iterator` member on the iterable, and on being able to do `moveNext()` and `current` on the returned object. With this change, we insist that the iterable must actually have type `Iterable<T>` for some `T`.

Note that deployed code already requires this (verified with `dart` and `dart2js`), and it is already specified for an asynchronous for-in that it is a dynamic error for the iterable to not be a `Stream`. So this is more consistent, not breaking in practice, and more strongly typed.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -12294,7 +12294,7 @@ It is a compile-time error if the static type of $c$ may not be assigned to \cod
 Let $D$ be derived from \syntax{<finalConstVarOrType>?}
 and let $n0$ be an identifier that does not occur anywhere in the program.
 A for statement of the form \code{\FOR{} ($D$ \id{} \IN{} $e$) $s$}
-is then equivalent to the following code,
+is then treated as the following code,
 where $\id_1$ and $\id_2$ are fresh identifiers:
 
 \begin{normativeDartCode}
@@ -12310,7 +12310,7 @@ $T$ $\id_1$ = $e$;
 If the static type of $e$ is a top type
 (\ref{superBoundedTypes})
 then $T$ is \code{Iterable<dynamic>},
-and otherwise $T$ is the static type of $e$.
+otherwise $T$ is the static type of $e$.
 It is a compile-time error if $T$ is not assignable to \code{Iterable<dynamic>}.
 
 \commentary{
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -12291,8 +12291,7 @@ It is a compile-time error if the static type of $c$ may not be assigned to \cod
 \LMLabel{for-in}
 
 \LMHash{}%
-Let $D$ be derived from \syntax{<finalConstVarOrType>?}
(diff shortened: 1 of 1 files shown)


lrhn:
The advantage of this change is that we can use `Iterable<T>` as context type for the iterable expression `e` of `for (T x in e)`. We currently can't because it might not be an iterable.

My guess is that we are already doing this, which is why implementations fail when it's not actually an iterable.
That's also supported by `for (var x in {});` not throwing. By the specification, this should be trying to iterate a map, but since it succeeds, we are obviously iterating a set. That means the context type must already be an iterable type.

specification/dartLangSpec.tex
```diff
@@ -36,6 +38,8 @@
 %   Function.
 % - Generalize specification of type aliases such that they can denote any
 %   type, not just function types.
+% - Added requirement that the iterator of a for-in statement must have
```
lrhn: Move this to 2.3 and change version to 2.3

specification/dartLangSpec.tex
```diff
@@ -12287,24 +12291,33 @@ \subsubsection{For-in}
...
-where $T$ is the static type of \code{$e$.iterator}.
-
-\commentary{
-It follows that it is a compile-time error if $D$ is empty and \id{} is a final variable.
+\noindent
+If the static type of $e$ is a top type
+(\ref{superBoundedTypes})
+then $T$ is \code{Iterable<dynamic>},
+otherwise $T$ is the static type of $e$.
+It is a compile-time error if $T$ is not assignable to \code{Iterable<dynamic>}.
```
lrhn: Add a TODO here saying that this is a language-induced implicit downcast, so we must change this for NNBD.

specification/dartLangSpec.tex
```diff
@@ -12287,24 +12291,33 @@ \subsubsection{For-in}
...
 \begin{normativeDartCode}
-\VAR{} $n0$ = $e$.iterator;
-\WHILE{} ($n0$.moveNext()) \{
-\ \ $D$ \id{} = $n0$.current;
-\ \ $s$
+$T$ $\id_1$ = $e$;
+\VAR{} $\id_2$ = $id_1$.iterator;
+\WHILE{} ($\id_2$.moveNext()) \{
+\ \ $D$ \id{} = $\id_2$.current;
+\ \ $S$
```
lrhn: Put `S` into a block, `{ S }`. 
Otherwise the `id` declaration is moved to the same level as the $S$, and that would cause a conflict for degenerate examples like `for (var id in []) var id = 2;`

specification/dartLangSpec.tex
eernstg: Done

specification/dartLangSpec.tex

specification/dartLangSpec.tex
eernstg: Discussed IRL, keeping the current text.

PR #292, opened 2019-03-29
Static type-pattern based feature spec for scoped static extension methods

eernstg:
This PR is a **static** variant of the PR #284. It adds the file 'feature-specification-using-static-type-patterns.md' to the directory which is concerned with static scoped extension methods.

It is an initial version of a feature specification of static scoped extension methods where only the core is covered (the extension methods themselves and their static and dynamic semantics, but no static methods or other "auxiliary" features that extensions could have).

It is based on using type patterns (#170) to specify how to extract the type arguments of the **statically known type** of the given receiver.

This means that extension methods can be considered to be a thin layer of syntactic sugar on top of global functions and their invocation.

However, static type patterns are more powerful than the static+dynamic ones used by #284, because we do not have to enforce that the type patterns are 'subtype robust', which means that they can work as a somewhat more powerful constraint language.


eernstg:
Abandoned: We chose to use an approach based on type inference rather then type patterns.

PR #293, opened 2019-03-29
Initial cut at an NNBD specification

leafpetersen:
My initial cut at a specification.  I've tried to lay out a specific choice for each of the remaining decisions that we have to take.


leafpetersen:
cc @lrhn @munificent @eernstg @mit-mit

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,374 @@
...
+
+
+## Syntax
+
+The grammar of types is extended to allow any type to be suffixed with a `?`
+(e.g. `int?`) indicating the nullable version of that type.
+
+A new primitive type `Never`.  This type is denoted by the built-in type
+declaration `Never`, with the same syntactic and scoping treatment as with other
+built-in types such as `Null`.
```
lrhn: Nit: This means it's declared in `dart:core`. This differs from `FutureOr` which is defined in `dart:async`, so it's not like *all* other built-in types.

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,374 @@
...
+setters, getters, and operators on `Object`.
+
+It is an error to read a field or tear off a method from an expression whose
+type is potentially nullable and not `dynamic`, except for the methods and
+fields on `Object`.
+
+It is an error to call an expression whose type is potentially nullable and not
+`dynamic`.
+
+It is an error for an instance field with no static initializer to have a
```
lrhn: What does "static initializer" mean?
Should perhaps be:

It is a compile-time error if a class has an instance field with a potentially nullable type and no initializer expression, and it has a non-redirecting generative constructor that does not intitialize that variable through an initializing formal or initializer list entry.

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,374 @@
...
+potentially nullable type unless the variable or field is marked with the `lazy`
+modifier.
+
+It is an error if a local variable that is potentially non-nullable and is not
+marked `lazy` is used before it is definitely assigned (see Definite Assignment
+below).
+
+It is an error if a method, function, getter, or closure with a potentially
```
lrhn: ... if the body of a ...
closure -> function expression

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,374 @@
...
+It is an error if a required named parameter has a default value.
+
+It is an error if a named parameter that is part of a `required` group is not
+bound to an argument at a call site.
+
+It is an error to call the default `List` constructor with a length argument and
+a type argument which is potentially non-nullable.
+
+For the purposes of errors and warnings, the null aware operators `?.` and `?..`
+are checked as if the receiver of the operator had non-nullable type.
```
lrhn: Should it be an error to use `?.` or `?..` on a non-nullable type?

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,374 @@
...
+It is an error for a class to extend, implement, or mixin the type `Never`.
+
+It is an error to call a method, setter, or getter on a receiver of static type
+`Never` (including via a null aware operator).
+
+It is an error to apply an expression of `Never` in the function position in a
+function call.
+
+It is an error if the static type of `e` in the expression `throw e` is
+potentially nullable.
```
lrhn: We can drop this. There is no actual problem with throwing Null.
This doesn't even allow us to disallow catching potentially nullable types, or even nullable types, because we don't have a type-constructor that makes a type non-nullable, and we would want:
```dart
foo<T>(Object o) { 
  try { 
     if (o != null) throw o;
  } on T catch (e) {
     Expect.identical(o, e);
  }
}
```
to work, and there is no type we can use instead of the potentially nullable `T` there.

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,374 @@
...
+
+It is not an error for the body of a `lazy` field to reference `this`.
+
+It is an error for a toplevel variable, a static field, or a formal parameter to
+be declared `lazy`.
+
+It is an error if the type `T` in the **on-catch** clause `on T catch` is
+potentially nullable.
```
lrhn: Not sure we can do that. It would leave some situations where you don't have a type to write because you know it's `T&Object`, but you can't write that.

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,374 @@
...
+are as follows.
+
+A check of the form `e == null` or of the form `e is Null` where `e` has static
+type `T` promotes the type of `e` to `Null` in the `true` continuation, and to
+**NonNull**(`T`) in the
+`false` continuation.
+
+A check of the form `e != null` or of the form `e is T` where `e` has static
+type `T?` promotes the type of `e` to `T` in the `true` continuation, and to
+`Null` in the `false` continuation.
```
lrhn: We might also want:

A check of the form `e is T` where `e` has static type `FutureOr<T>` promotes `e` to `T` in the `true` continuation, and to `Future<T>` in the `false` continuation.

(Or is this just assumed part of the TODO above, since it's not null-related?)

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,374 @@
...
+
+The **NonNull** function defines the null-promoted version of a type, and is
+defined as follows.
+
+- **NonNull**(Null) = Never
+- **NonNull**(_C_<_T_<sub>1</sub>, ... , _T_<sub>_n_</sub>>) = _C_<_T_<sub>1</sub>, ... , _T_<sub>_n_</sub>>  for class *C* other than Null
+- **NonNull**(FutureOr<_T_>) = FutureOr<_T_>   
+- **NonNull**(_T_<sub>0</sub> Function(...)) = _T_<sub>0</sub> Function(...)
+- **NonNull**(Function) = Function
+- **NonNull**(Object) = Object
```
lrhn: Technically `Object` is handled by the _C_ case above.

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,374 @@
...
+- **NonNull**(Never) = Never
+- **NonNull**(dynamic) = dynamic
+- **NonNull**(void) = void   
+- **NonNull**(_X_) = X & **NonNull**(B), where B is the bound of X.
+- **NonNull**(_X_ & T) = X & **NonNull**(T)
+- **NonNull**(_T_?) = **NonNull**(_T_)
+- **NonNull**(_T_\*) = **NonNull**(_T_)
+
+#### Type promotion, Definite Assignment, and Definite Completion
```
lrhn: Duplicate section.

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,374 @@
...
+
+#### Null aware operator
+
+An expression of the form `e?.<tail>` where `<tail>` is any sequence of
+selectors evaluates to `null` if `e` evaluates to `null`, and otherwise
+evaluates to the same value as `e.<tail>`.
+
+An expression of the form `e?..<tail>` where `<tail>` is any sequence of
+selectors evaluates to `null` if `e` evaluates to `null`, and otherwise
+evaluates to the same value as `e..<tail>`.
```
lrhn: Need something more to handle `x?..foo()..bar()`. This should elide both calls.

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,374 @@
...
+    initializer expression overwrites any intermediate written values.
+  - Evaluating the initializer expression may cause an exception to be thrown.
+    If the variable was written to before the exception was thrown, the value of
+    the variable on subsequent reads is the last written value.  If the variable
+    was not written before the exception was thrown, then the next read attempts
+    to evaluate the initializer expression again.
+  - If a variable or field is read from during the process of evaluating its own
+    initializer expression, and no write to the variable has occurred, the read
+    is treated as a first read and the initializer expression is evaluated
+    again.
```
lrhn: I don't think we should recurse here. If a lazy variable is read during the evaluation of its own initializer expression, the read should throw (a `CyclicInitiaizationError`) like it currently does.

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,374 @@
...
+
+A toplevel or static variable with an initializer is evaluated as if it was
+marked `lazy`.  Note that this is a change from pre-NNBD semantics in that:
+  - Throwing an exception during initializer evaluation no longer sets the
+    variable to `null`
+  - Reading the variable during initializer evaluation is no longer a checked
+    error.
```
lrhn: "checked"?

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,374 @@
...
+### Type reification
+
+All types reified in legacy libraries are reified as legacy types.  Runtime
+subtyping checks treat them according to the subtyping rules specified
+separately.
+
+### Runtime checks and weak checking
+
+When weak checking is enabled, runtime type tests (including explicit and
+implicit casts) shall succeed with a warning whenever the runtime type test
```
lrhn: Not sure what this means. What does "succeed with a warning mean", and what is the *result* of the test?
Does it succeed without a warning if the test succeeds with non-changed types?

This sounds like we have *runtime warnings* and each test needs to do two type checks (although we can perhaps merge that into one check algorithm with a ternary result). It sounds *expensive*. For the warning to be useful, it must at minimum remember the source position of the test, which is an extra overhead on application size.

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,374 @@
...
+`Null`, and `required` named parameters were treated as optional.
+
+### Exports
+
+If an unmigrated library re-exports a migrated library, the re-exported symbols
+retain their migrated status (that is, downstream migrated libraries will see
+their migrated types).
+
+It is an error for a migrated library to re-export symbols from an unmigrated
+library.
```
lrhn: This can hold back migration. 
Also, if an unmigrated library can export migrated "symbols", this might not even export any unmigrated types.

Maybe:
It is an error for a migrated library to re-export declarations with a static type containing unmigrated types.

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,374 @@
...
+means that an unmigrated library can bring together otherwise incompatible
+methods.  When choosing the most specific signature during interface
+computation, all nullability and requiredness annotations are ignored, and the
+`Never` type is treated as `Null`.
+
+In a migrated library, override checking must check that an override is
+consistent with all overridden methods from other migrated libraries in the
+super-interface chain, since a legacy library is permitted to override otherwise
+incompatible signatures for a method.
+
```
lrhn: (What do we do if a migrated class extends an unmigrated class, and does not override an umigrated method? Will the type in the subclass interface be the `*`-erasure of the original type? Will it be enforced?)

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,374 @@
...
+type' ::= functionType
+          | qualified typeArguments? 
+
+type ::= type' `?`?
+```
+
+#### Conditional expression ambiguities
+
+Conditional expressions inside of braces are ambiguous between sets and maps.
+That is, `{ a as int ? - 3 : 3 }` can be parsed as a set literal `{ (a as int)
```
stereotype441: Suggestion: rephrase this example into a form where the preferred parse would not produce a compile-time error.  E.g.: `{ a as bool ? -3 : 3 }` instead of `{ a as int ? -3 : 3 }`.  That way this example can be translated directly to a test case.

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,374 @@
...
+
+The grammer of expressions is extended to allow any expression to be suffixed
+with a `!`.
+
+The grammer of instance fields and local variables is extended to allow any
+declaration to include the modifer `lazy`.
+
+The grammar of function types is extended to allow an additional group of named
+parameters prefaced with the word `required` (e.g. `int Function(int, {int?
+y}, required {int z})`.
```
danrubel: Should the above be changed to "allow the `required` modifier before each named parameters"
and `{required int z}` ?

leafpetersen: Yes.

danrubel: For consistency, with `@required`, I recommend that the new `required` modifier come before all other modifiers for that parameter (e.g. `required final int x`). Sound reasonable?

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,374 @@
...
+
+A new primitive type `Never`.  This type is denoted by the built-in type
+declaration `Never`, with the same syntactic and scoping treatment as with other
+built-in types such as `Null`.
+
+The grammer of expressions is extended to allow any expression to be suffixed
+with a `!`.
+
+The grammer of instance fields and local variables is extended to allow any
+declaration to include the modifer `lazy`.
```
danrubel: Per discussion, we are allowing `late` rather than `lazy`.
Also, where does the modifier appear in a string of modifiers?
```
   final late int x;
```
or
```
   late final int x;
```
?

working/0110-incremental-sound-nnbd/feature-specification.md
leafpetersen: Done.  Left the question of keyword vs built-in identifier as a TODO.

working/0110-incremental-sound-nnbd/feature-specification.md
leafpetersen: done.

working/0110-incremental-sound-nnbd/feature-specification.md
leafpetersen: done.

working/0110-incremental-sound-nnbd/feature-specification.md
leafpetersen: Not sure.  Maybe?  It's nice to get some warning at least, but it makes changing a type from nullable to non-nullable more of a breaking change.  e.g. in

```
  library lib;
  int ? x;
  int? foo() => 3
```

you could change both `int?`s to `int` with no downstream breakage unless we make it an error to do `foo()?.isEven` if `foo` returns int.

working/0110-incremental-sound-nnbd/feature-specification.md
leafpetersen: > There is no actual problem with throwing Null.

Why do you say there's no problem throwing null?  Doing so causes a NullThrownError, which is the "throw" equivalent of a null-pointer error.  Is this really ever what the user wanted?

> This doesn't even allow us to disallow catching potentially nullable types, or even nullable types

Why can't we disallow catching nullable types?  There's no reason to allow `on int? catch (e) {` because `e` will never be `null`, no matter what.  I guess technically you could want to do `on FutureOr<int?> catch(e)`, but.... really?

working/0110-incremental-sound-nnbd/feature-specification.md
leafpetersen: Yes, part of TODO.

working/0110-incremental-sound-nnbd/feature-specification.md
leafpetersen: Ok - called it out there though to make it clear.  I prefer technically clear to technically correct.  ;)

working/0110-incremental-sound-nnbd/feature-specification.md
leafpetersen: We had a long discussion of this on the board, and I thought we had resolved this in favor of not doing the cyclic initialization error checks.  Have you changed your mind on this?

working/0110-incremental-sound-nnbd/feature-specification.md
leafpetersen: We do not check for it.

working/0110-incremental-sound-nnbd/feature-specification.md
leafpetersen: Do you really think this is likely?  Migrating before your upstream packages migrate is already the less common use case. Exporting symbols is a fringe use case.  Exporting symbols that were exported from yet another library seems like a fringe on a fringe case.  It seems much simpler to me to say that we just don't support it.

working/0110-incremental-sound-nnbd/feature-specification.md
leafpetersen: I think that inherited methods should stay as legacy types.  So:
```
// @dart=2.3
library foo
class A {
  int foo(int x) => null;
}

class B extends A {
}

void test() {
  int x = foo(null); // Ok
}
 ```

File: working/0110-incremental-sound-nnbd/feature-specification.md (hunk 5 above)
Comment (lrhn) [reply working/0110-incremental-sound-nnbd/feature-specification.md]: I mean that if we choose to remove the artificial restriction on throwing `null`, we could. Things would still work. We added the rule because some people didn't think it worked well with the catch syntax at the time, `catch (SomeError e)`, where `SomeError` looked like a type annotation and would therefore be assumed to catch null. To avoid confusion, you could no longer throw `null`. The current syntax `on SomeError catch (e)` doesn't have this problem so we could just allow throwing `null` again.

If we keep throwing `null` disallowed, then perhaps we can disallow catching (some) definitely nullable types, but not potentially nullable types. We probably must allow `on X`, a type variable, or `on dynamic`. I don't see any big advantage to disallowing `on Object?`, the `on` clause of `catch` blocks generally works like an `is` check, so we can allow anything there that an `is` check allows. I don't think adding restrictions is worth the complexity, and I doubt we'll catch any errors with those restrictions.

Comment (leafpetersen) [reply working/0110-incremental-sound-nnbd/feature-specification.md]: > If we keep throwing `null` disallowed

Are we considering changing this?  I certainly wasn't.

File: working/0110-incremental-sound-nnbd/feature-specification.md [hunk 19]
```diff
@@ -0,0 +1,374 @@
...
+[roadmap](https://github.com/dart-lang/language/blob/master/working/0110-incremental-sound-nnbd/roadmap.md).
+
+This proposal draws on the proposal that Patrice Chalin wrote
+up [here](https://github.com/dart-archive/dart_enhancement_proposals/issues/30),
+and on the proposal that Bob Nystrom wrote
+up
+[here](https://github.com/dart-lang/language/blob/master/resources/old-non-nullable-types.md).
+
+
+## Syntax
```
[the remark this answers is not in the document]
leafpetersen: I've added the proposed grammar changes to this PR, subject to review.

specification/dartLangSpec.tex
```diff
@@ -826,7 +826,8 @@ \section{Variables}
 
 <declaredIdentifier> ::= <metadata> \COVARIANT{}? <finalConstVarOrType> <identifier>
 
-<finalConstVarOrType> ::= \FINAL{} <type>?
+<finalConstVarOrType> ::= \LATE{} \FINAL{}? <type>?
```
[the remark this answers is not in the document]
danrubel: For processing efficiency, we have exactly one modifier sequence. `external` must appear before `static` or `covariant` which much appear before `final`, etc.

While we could add this as the first modifier grouping that can appear in either order, I'd rather not open that door. Please pick one modifier order and allow the other to be an error/warning.

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,399 @@
...
+
+The grammar of types is extended to allow any type to be suffixed with a `?`
+(e.g. `int?`) indicating the nullable version of that type.
+
+A new primitive type `Never`.  This type is denoted by the built-in type
+declaration `Never` declared in `dart:core`.
+
+The grammer of expressions is extended to allow any expression to be suffixed
```
gamebox: grammar

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,399 @@
...
+with a `!`.
+
+The modifier `late` is added as a built-in identifier.  The grammer of top level
```
gamebox: grammar

specification/dart.sty
```diff
@@ -45,6 +46,7 @@
 \def\IF{\keyword{if}}
 \def\IN{\keyword{in}}
 \def\IS{\keyword{is}}
+\def\LATE{\keyword{late}}
```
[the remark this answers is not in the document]
leafpetersen: @bwilkerson was advocating for this to be a keyword on the grounds that it makes error recovery much more robust.  I don't have strong feelings about this - I see arguments on both sides.  Taking away variable names from programmers is painful for them (and for us), but making error recovery less robust is a tax on everything.

specification/dartLangSpec.tex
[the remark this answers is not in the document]
leafpetersen: In other places we're pretty opinionated about ordering (e.g. `static final`) and someone (I think @bwilkerson or @stereotype441) was claiming that the Java community started enforcing this as a convention at some point just because the inconsistency was painful.  @munificent thoughts on this?

specification/dartLangSpec.tex
```diff
@@ -9479,10 +9485,14 @@ \subsubsection{Cascaded Invocations}
 where $e$ is an expression and \metavar{suffix} is a sequence of operator, method, getter or setter invocations.
 
 \begin{grammar}
-<cascadeSection> ::= `..' (<cascadeSelector> <argumentPart>*)
+<cascadeSequence> ::= (`?..' \alt `..`) <cascadeSection> <cascadeSectionContinuation>*
```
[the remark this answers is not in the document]
leafpetersen: done.

specification/dartLangSpec.tex
```diff
@@ -11496,6 +11506,8 @@ \subsection{Assignable Expressions}
 
 <assignableSelector> ::= <unconditionalAssignableSelector>
   \alt `?.' <identifier>
+  \alt `!'
```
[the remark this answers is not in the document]
leafpetersen: > I don't think `x!` should be _assignable_. Put it into `<selector>` instead.

If you do this, then you can't write `x!.y = 3`, which I think we want to be able to write.  Unless I'm missing something?  But I think the `x!.y` can't be derived as a selector, you have to go through assignableSelector.  So I was planning on adding this here, and then disallowing things like `x! = 3` semantically.  

> You can write garbage like `x..foo() = 42` with the existing grammar too,

How does this derive?  I don't see how `x..foo()` gets derived as an `<assignableExpression>`.

specification/dartLangSpec.tex
```diff
@@ -14106,15 +14118,18 @@ \subsection{Static Types}
...
+<typeNotVoid> ::= <functionType> `?'?
+  \alt <typeNotVoidNotFunction>
 
-<typeNotVoidNotFunction> ::= <typeName> <typeArguments>?
-  \alt \FUNCTION{}
+<typeNotFunction> ::= \VOID{}
+  \alt <typeNotVoidNotFunction>
+
+<typeNotVoidNotFunction> ::= <typeName> <typeArguments>? `?'?
+  \alt \FUNCTION{} `?'?
```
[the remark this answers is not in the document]
leafpetersen: I think disallowing `dynamic?` and `void?` is just going to make this (much) worse.  I didn't set out to disallow `void?` - it just kind of falls out of the existing weirdness of how `void` is treated.  If we want to allow it, I can explicitly allow it by changing to:

```
<typeNotFunction> ::= \VOID{} `?'?
  \alt <typeNotVoidNotFunction>
```

working/0110-incremental-sound-nnbd/feature-specification.md
leafpetersen: thanks, done.

working/0110-incremental-sound-nnbd/feature-specification.md
leafpetersen: Updated with a more precise specification, PTAL.

working/0110-incremental-sound-nnbd/feature-specification.md
lrhn: Ack, no, just didn't remember it. My first instinct is still to disallow what is almost certainly an infinite recursion or an early throw (a second attempt to store an initializer expression's value for a final field will throw). 
I can see the arguments for not doing extra state-management (with a need for try/finally), but it just feels so very fragile.

But, yes, this is what we agreed on.

working/0110-incremental-sound-nnbd/feature-specification.md
lrhn: Maybe write it as "no longer throws".

working/0110-incremental-sound-nnbd/feature-specification.md
lrhn: Why is it that we won't support it? That is, what is the problem we are trying to solve by disallowing the export?

Is it purely methodological problems: Library A is migrated, but exports declarations from unmigrated library B. Library C uses library A and wants to migrate, so they might need to migrate twice, once for A and once more when B is migrated. We don't want that, so we disallow A from exporting declarations from B until B has been migrated.

If so, I'm not sure how that differs from an unmigrated library exporting declarations of a migrated library. If library A exports declarations from library B and library B migrates first, then the user in library C, which imports library A, may need to migrate twice as well, once for the B changes, and once more for the A changes.

They can choose to not migrate library C until both A and B have been migrated, but that applies to both cases.

Or is there a technical problem? I can't see one because you can get into the exact same state by importing library A and B separately, rather than relying on the export of B by A, and we need to be able to handle that.

working/0110-incremental-sound-nnbd/feature-specification.md
lrhn: That's probably the only sane choice. If you don't write anything, we can't guess how to migrate the legacy function, so we just have to keep it at the legacy signature.

working/0110-incremental-sound-nnbd/feature-specification.md
lrhn: I was always considering it (I never cared for the restriction), but with NNBD it has begun to actually matter. We need to decide whether `try { ... } on int? { ... }` is allowed, whether it's useful, and how that affects the type of `e` in `try { ... } catch (e) { ... }`.

On one hand, disallowing `null` in some situations is exactly what NNBD is for, and if we think that throwing `null` is most likely an error, we should keep disallowing it. (That is probably the case, but I want to make sure).

On the other hand, we now have a way to distinguish between `on int` and `on int?`, so any compunctions we had about whether the `on` type should or shouldn't match `null` is now gone. If that was the only reason to not allow throwing `null` (it was the only one given at the time, but maybe just because it was sufficient), then we could just allow throwing `null`.

If we allow throwing `null`, then the default `on` clause of a `try` would be `on Object?`. That also means that `try { ... } catch (e, s) { ... }` would give `e` the static type `Object?`, but `Object?` is probably not really distinguishable from `Object` in practice (they allow the same methods, and only differ in assignability in that `Object?` is not assignable to `Object`).

If we disallow throwing `null`, then we also disallow potentially nullable typed expressions in throws (the context type of the `throw` operand is `Object`).

Do we also disallow nullable types in `try`/`on` clauses? That is, can you write:
```dart
try {
  ...
} on Exception? catch (e, s) {
   // e has static type Exception?
}
```
or do we disallow it because it's known to be unnecessary.
Would we also disallow potentially nullable types? (Likely yes, but it means that:
```dart
T catcher<T>(void Function() thrower) {
  try {
    thrower();
  } on T catch (e) {
    return e;
  }
  throw ArgumentError.value(thrower, "thrower", "Didn't throw");
}
```
would be illegal, and you have to make it `T catcher<T extends Object>(...)`. That's also more correct, so it's not necessarily a problem.


All in all, I think we should think about this just once more, just to be sure. 
It's probably correct that `throw null` is a mistake, a value accidentally being `null`, and disallowing it is doing the user a service.

specification/dartLangSpec.tex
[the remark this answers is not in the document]
lrhn: If you change `selector` to:
```ebnf
<selector> ::= `!'
 \alt <argumentPart>
 \alt <assignableSelector>
```
*and* either `assignableSelectorPart` to:
```
<assignableSelectorPart> ::= (<argumentPart> | `!')* <assignableSelector>
```
(from `<assignableSelectorPart> ::= <argumentPart>* <assignableSelector>`)
or change that and `assignableExpression` to:
```ebnf
<assignableeExpression> ::= <primary> <assignableSelectorPart> 
  \alt \SUPER{} <unconditionalAssignableSelector>
  \alt <constructorInvocation> <assignableSelectorPart>
  \alt <identifier>

<assignableSelectorPart> ::= <selector>* <assignableSelector>
```
then you can derive `x!.y = 3`.

You are correct that just changing `<selector>` si not enough because `assignableSelectorPart` doesn't use `selector`.

---

About `..foo() = 42`, the grammar for `cascadeSection` is
```ebnf
<cascadeSection> ::= `..' (<cascadeSelector> <argumentPart>*)
  \gnewline{} (<assignableSelector> <argumentPart>*)*
  \gnewline{} (<assignmentOperator> <expressionWithoutCascade>)?
```
and with:
 *  `<cascadeSelector>` &mapsto; `<identifier>` &mapsto; `foo`
 * `<argumentPart>` &mapsto; `<arguments>` &mapsto; `(` `)`
 * `<assignmentOperator>` &mapsto; `=`
 * `<expressionWithoutCascade>` &mapsto;<sup>*</sup> `42`

you can derive `x .. foo () = 42` by having one `<argumentPart>`, zero `(<assignableSelector> <argumentPart>*)` and one `(<assignmentOperator> <expressionWthoutCascade>)` as allowed by the multiplicities.
There is no `<assignableExpression>` involved.

The problem is that `<argumentPart>` is a selector and not an assignable selector, but the cascade grammar allows inserting it before the assignment. The issue is the cascade grammar, and as you say the `<assignableExpression>` doesn't have that issue.
I think my suggestion above is better (but would obviously like someone else to check that it still allows all the right things).

The code is rejected by our tools anyway, even though it satisfies the grammar. It is then treated as equivalent to `x.foo() = 42`, which is itself invalid. Example:
```dart
main() {
  42..toString() = 4;
} 
```
gives the CFE error:
```
casc.dart:2:18: Error: Can't assign to this.
  42..toString() = 4;
```
and the analyzer error:
```
Analyzing casc.dart...
  error • Missing selector such as '.<identifier>' or '[0]' at casc.dart:2:5 • missing_assignable_selector
  error • A value of type 'int' can't be assigned to a variable of type 'String' at casc.dart:2:20 • invalid_assignment
2 errors found.
```
(The second error is spurious—there is no variable of type string in the program—which suggests that the analyzer isn't entirely comfortable with this misuse of syntax).

specification/dartLangSpec.tex
```diff
@@ -1405,13 +1406,13 @@ \subsection{Formal Parameters}
...
-  \alt `(' <normalFormalParameters> `,' <optionalFormalParameters> `)'
-  \alt `(' <optionalFormalParameters> `)'
+  \alt `(' <normalFormalParameters> `,' <optionalOrNamedFormalParameters> `)'
+  \alt `(' <optionalOrNamedFormalParameters> `)'
 
 <normalFormalParameters> ::= \gnewline{}
   <normalFormalParameter> (`,' <normalFormalParameter>)*
 
-<optionalFormalParameters> ::= <optionalPositionalFormalParameters>
+<optionalOrNamedFormalParameters> ::= <optionalPositionalFormalParameters>
```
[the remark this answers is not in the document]
bwilkerson: Personally, I'd like to see us allow all four varieties. Whether that happens sooner or later is a separate question, but it makes it much easier to migrate users to a new API if named and optional positional can be mixed, so I'd prefer sooner.

specification/dartLangSpec.tex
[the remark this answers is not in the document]
bwilkerson: That was me.

specification/dartLangSpec.tex
```diff
@@ -1479,8 +1480,8 @@ \subsubsection{Optional Formals}
 \begin{grammar}
 <defaultFormalParameter> ::= <normalFormalParameter> (`=' <expression>)?
 
-<defaultNamedParameter> ::= <normalFormalParameter> (`=' <expression>)?
-  \alt <normalFormalParameter> ( `:' <expression>)?
+<defaultNamedParameter> ::= \REQUIRED{}? <normalFormalParameter> (`=' <expression>)?
+  \alt \REQUIRED{}? <normalFormalParameter> ( `:' <expression>)?
```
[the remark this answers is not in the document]
kasperpeulen: What about a non nullable named parameter, that has no default value, is this inferred as being required? 

I hope so because that would mean that we can get rid of allmost all cases of required.

lrhn:
There is on easy way to make a `<functionFormalParameter>` nullable. > I suggest:
```
<functionFormalParameter> ::= \gnewline{}
    <metadata> \COVARIANT{}? <type>? <identifier> <formalParameterPart> `?'?
```
The final optional `?` would make the function type of the parameter into a nullable function type, so:
```dart
  E firstWhere(bool test(E element), {E orElse()?});
```
would be valid.

When trying to convert dart:convert and dart:collection, I had a lot of optional function parameters which need to be nullable.

bwilkerson:
Would it be better to change it to something like the following?
```
E firstWhere(bool Function(E element) test, {E Function()? orElse});
```

lrhn:
@bwilkerson 
Please define "better" :)

Unless we intend to remove the old-style function parameter format (I don't), we should fully support it.
It is highly annoying if you can write `foo(int compare(int a, int b))`, but have to rewrite it to `foo(int Function(int a, int b)? compare)` when you want it to be nullable. The non-orthogonality is jarring.

As a person migrating a large library (I'm working on a proof-of-concept migration of the core platform libraries), I found it *very* convenient to just put a `?` after the `)`, independently of the parameter format. I actually first wrote `foo(int compare(int a, int b)?)`, then started worrying whether it was allowed. That suggests to me that it is a reasonable syntax, and that it is needed.

I personally *like* the old-style function parameter syntax, and prefer it over writing the type separately. I will be sad if I am forced to the "worse" syntax just because the type must be nullable.

So, to answer your question from a personal perspective: *Emphatic no!*

bwilkerson:
Ok. The consistency argument is valid and persuasive. If the language team agrees to update the syntax to allow this, please let us know explicitly so that we can update the parser and other code paths.

What I meant by "better" is that I thought we were actively encouraging users to switch to the newer syntax and that it would therefore be better in the sense that the SDK would follow our own advice. I might be wrong about any of number of things in that line of reasoning. :-)

stereotype441:
> Ok. The consistency argument is valid and persuasive. If the language team agrees to update the syntax to allow this, please let us know explicitly so that we can update the parser and other code paths.

I went ahead and filed #364 to track this question.

specification/dartLangSpec.tex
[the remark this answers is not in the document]
eernstg: I'm reverting the changes I introduced into Dart.g based on Lasse's proposal to swap `final` and `late` and deleted the 'Suggested change' that I used to have in the previous comment.  Similar changes were made in other comments below.

specification/dartLangSpec.tex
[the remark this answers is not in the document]
lrhn: It is top of my list for improvements going on 6.5 years now (https://github.com/dart-lang/sdk/issues/7056, sadly only the nr 27 most wanted feature based on :+1: count).

specification/dartLangSpec.tex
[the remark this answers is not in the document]
lrhn: I can see a number of patterns emerge for myself after trying to convert code to NNBD:
* Local variables differ. I usually write `late final` without an intializer (late assignment), and `final late` with an initializer (lazy initialization).
* For class-level variables, I always write `final late` (because they are late assignments).
* No top-level variables yet. 

All in all, if we only allow one ordering, I think I'd prefer `final late` to `late final`, after having actually written code with it.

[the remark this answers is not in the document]
leafpetersen: Do you want to allow `late var x`, `var late x`?  What about `int late x` and `late int x`?  I'm a little confused about how to think about this.

specification/dartLangSpec.tex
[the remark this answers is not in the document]
leafpetersen: @eernstg yes.

@kasperpeulen when we discussed this, the feeling was that it was error prone that cutting and pasting a method signature to a type would change its meaning:

```
int foo({int x}) => x; // Implicitly required
int Function({int x}) f = foo; // This is an error, since the parameter in the function type is optional
```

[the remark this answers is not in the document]
leafpetersen: @eernstg this is already marked as a semantic error below.

specification/dartLangSpec.tex
```diff
@@ -1848,8 +1849,13 @@ \section{Classes}
   \alt \EXTERNAL{}? <operatorSignature>
   \alt (\EXTERNAL{} \STATIC{}?)? <functionSignature>
   \alt \STATIC{} (\FINAL{} | \CONST{}) <type>? <staticFinalDeclarationList>
+  \alt \STATIC{} \LATE{} \FINAL{}? <type>? <initializedIdentifierList>
```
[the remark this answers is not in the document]
leafpetersen: Done.

specification/dartLangSpec.tex
```diff
@@ -1848,8 +1849,13 @@ \section{Classes}
...
+  \alt \STATIC{} (\VAR{} | <type>) <initializedIdentifierList>
+  \alt \COVARIANT{} (\VAR{} | \LATE{}) <initializedIdentifierList>
```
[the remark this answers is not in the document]
leafpetersen: Done.

specification/dartLangSpec.tex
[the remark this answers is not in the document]
leafpetersen: I think changing this to stop special-casing void is orthogonal to this change, so leaving this as is.

specification/dartLangSpec.tex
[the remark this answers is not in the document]
leafpetersen: @lrhn If I do as you suggest, then I don't think you can have `!` in a cascade section anymore, right?

```
a..foo()!.bar=3;
```

I think this cascade parses with my grammar as:

```
cascadeSelector -> identifier -> foo
argumentPart* -> argumentPart -> ()
assignableSelector -> !
assignableSelector -> .bar
assignmentOperator -> =
expressionWithoutCascade -> conditionalExpression -> ifNullExpression -> logicalOrExpression -> logicalAndExpression -> equalityExpression -> relationExpression -> bitwiseOrExpression -> bitwiseXOrExpression -> bitwiseAndExpression -> shiftExpression -> additiveExpression -> multiplicativeExpression -> unaryExpression -> StayWithMeAlmostThere -> primary -> literal -> numericLiteral
```

How does it parse with yours?

working/0110-incremental-sound-nnbd/feature-specification.md
```diff
@@ -0,0 +1,524 @@
...
+- **NonNull**(_T_?) = **NonNull**(_T_)
+- **NonNull**(_T_\*) = **NonNull**(_T_)
+
+#### Extended Type promotion, Definite Assignment, and Definite Completion
+
+These are extended as per separate proposal.
+
+### Runtime semantics
+
+#### Null assertion operator
```
stereotype441: Would it be better to call this a "null check operator"?  Every time I hear "null assertion operator", I get momentarily confused because it sounds like it's an operator that asserts that its expression is null, whereas in fact it's the opposite.

specification/dartLangSpec.tex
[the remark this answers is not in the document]
lrhn: Never had to think about `var`, and I can see how that's confusing.

I use `late final type name` or `final late type name`. If I omit the type (which only makes sense for a declaration with an initializer), it will be `final late name = ...`.

Since `var` is different from `final` in that you cannot have a type with `var`, the options are only:
* `var late x;` and
* `late var x;`.
The latter *reads* better because you actually read it as "late variable x", which works with natural language, where "variable late x" does not. That is, "var" occurs as a noun when you read it.
That's unlike "final" which is read as an adjective, and therefore can be swapped with the adjective "late".

Modifiers before type, so definitely `late int x;`, not `int late x`, and I treat `var` as a placeholder for a type, not a modifier, so probably `late var x` too.
(I would be perfectly happy if we allowed `final var x`, and treated `final x` as a shorthand for that, then `var` really *is* a placeholder for a type).

So, what I want is:
 * `late int x`
 * `late var x`
 * `final int x`
 * `final x`
 * `final late x`
 * `final late int x`
 * `late final x`
 * `late final int x`

If I can't get both orderings, then I'd prefer `final late (int) x` to `late final (int) x`. 
There is no technical issue with allowing both, so it would be because someone wants there to be only one way to do things. You can enforce that with lints if they really want it, no need to restrict the language for everybody else (or lock down the style guide before people have had a chance to actually use the feature).

specification/dartLangSpec.tex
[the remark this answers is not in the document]
lrhn: Let's pick the first grammar I proposed:
```ebnf
<cascadeSequence> ::= (`?..' \alt `..') <cascadeSection> (`..' <cascadeSection>)*

<cascadeSection> ::= <cascadeSelector> (<cascadeAssignment> | <selector>* (<assignableSelector> <cascadeAssignment>)?)

<selector> ::= `!' 
  \alt <argumentPart>
  \alt <assignableSelector>

<assignableSelector> ::= (`?.' \alt `.') <identifier>
  \alt `?.'? `[' expression `]'

<cascadeAssignment> ::= <assignmentOperator> <expressionWithoutCascade>
```
Then the cascade section of
```dart
a...foo()!.bar = 3
```
parses as
* cascadeSection &mapsto; cascadeSelector selector* assignableSelector cascadeAssignment
* cascadeSelector &mapsto; identifier (`foo`)
* selector* &mapsto; selector selector
* selector &mapsto; argumentPart &mapsto;* `(` `)`
* selector &mapsto; `!`
* assignableSelector &mapsto; `.` identifier (`bar`)
* cascadeAssignment &mapsto; `=` expressionWithoutCascade
* expressionWithoutCascade &mapsto;* integerLiteral (3)

We only want the actually assignable selectors to be assignable, and `!` does not introduce a LHS, so it should not be an assignable selector. Only `.id`, `[e]` and `?.[e]` are assignable because it's the only suffixes that can denote a setter (counting `operator[]=` as a settter).

specification/dartLangSpec.tex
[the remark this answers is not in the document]
bwilkerson: > ... no need to restrict the language for everybody else ...

I was not particularly arguing to restrict the language (not that I haven't on other occasions :-). I merely pointed out two things of possible interest.

First, that Dart imposes a single order for all other sequences of modifiers and that we might want to be consistent. If we think that it's worth allowing the modifiers `late` and `final` to appear in either order, then perhaps we should allow other modifiers to appear in arbitrary order as well.

Second, that Java started out allowing modifiers to appear in any order, and the community strongly preferred having a single order for them. This suggests that having a single order for modifiers might improve readability rather than reduce it, and hence it might make sense to specify a single order for these modifiers just as we've done elsewhere in the grammar.

working/0110-incremental-sound-nnbd/feature-specification.md
leafpetersen: Ok, I've added the following:

```
It is a warning to use a null aware operator (`?.`, `?..`, `??`, `??=`, or
`...?`) on a non-nullable receiver.
```

working/0110-incremental-sound-nnbd/feature-specification.md
leafpetersen: Ok, filed [an issue](https://github.com/dart-lang/language/issues/385) on this, let's discuss and resolve it there.

specification/dartLangSpec.tex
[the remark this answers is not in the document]
leafpetersen: We resolved this offline in favor of `late final` being the canonical ordering.

working/0110-incremental-sound-nnbd/feature-specification.md
leafpetersen: Yes to all of the above, except that it's not clear to me why you need to remember more information for the warning than you do for the error that would otherwise be thrown.

working/0110-incremental-sound-nnbd/feature-specification.md
leafpetersen: It seems to me to add a lot of complication if migrated libraries can have subset of their APIs which are unmigrated.

specification/dartLangSpec.tex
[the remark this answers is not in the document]
leafpetersen: > Will we want to allow optional positional parameters and _required_ named parameters on the same function?

This seems to me raise the same kind of calling convention questions that other combinations raise.

```dart
class A {
  void foo({required int x}) {}
}

class B extends A {
  void foo({int x}) {}
}

class C extends A {
  void foo([int y], {required int x}) {}
}
```

A call site that looks like `f(A x) => x.foo(x : 3)` doesn't know whether the receiver expects an optional named parameter, or a required named parameter, or a required named parameter and one or more optional parameters.

I'm not entirely unsympathetic to broadening our calling conventions, but I think that's a separate change that we need to make with full buy in from the various teams.

working/0110-incremental-sound-nnbd/feature-specification.md
leafpetersen: done.

specification/dartLangSpec.tex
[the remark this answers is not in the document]
leafpetersen: Okay, I've taken your suggestion, PTAL.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -0,0 +1,536 @@
...
+non-nullable.
+
+We say that a type `T` is **potentially nullable** if `T` is not non-nullable.
+Note that this is different from saying that `T` is nullable.  For example, a
+type variable `X extends Object?` is a type which is potentially nullable but
+not nullable.  Note that `T*` is always potentially nullable by this definition.
+
+We say that a type `T` is **potentially non-nullable** if `T` is not nullable.
+Note that this is different from saying that `T` is non-nullable.  For example,
+a type variable `X extends Object?` is a type which is potentially non-nullable
```
a14n: Shouldn't it be `X extends Object` instead of  `X extends Object?`?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -0,0 +1,536 @@
...
+
+It is an error for a class to extend, implement, or mixin a type of the form
+`T?` for any `T`.
+
+It is an error for a class to extend, implement, or mixin the type `Never`.
+
+It is an error to call a method, setter, or getter on a receiver of static type
+`Never` (including via a null aware operator).
+
+It is an error to apply an expression of `Never` in the function position in a
```
a14n: > an expression of`Never`

looks strange. Aren't there some words missing?

leafpetersen: fixed, thanks.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: No.  Note that `Object` is not nullable.

`X extends Object` **is** potentially non-nullable, since it is not nullable, but it is **also** non-nullable, since it is a subtype of `Object`.  

`X extends Object?` on the other hand is potentially non-nullable, since it is not nullable (`Null` is not a subtype of it) , but it is not *known* to be non-nullable, since it could be instantiated with a nullable type.

That may all seem confusing, but basically:

nullable type -> definitely could get Null
  - e.g. int?
non-nullable type -> definitely will never get Null
  - e.g. int
potentially nullable type -> might or might not get "instantiated" such that it becomes nullable
  - e.g. X extends Object?
  - e.g. int*
potentially non-nullable type -> might or might not get "instantiated such it becomes non-nullable. 
  - e.g. X extends Object?
  - e.g. int*

leafpetersen:
I'm going to go ahead and land this, so that we have a baseline for further changes to be diff'ed against.  I think we have all of the major discussion threads either resolved or captured in separate issues.

accepted/future-releases/nnbd/feature-specification.md
a14n: Thanks for the explanation.

PR #302, opened 2019-04-05
Update section "Constant Constructors"

### LINKED ISSUE #289: Language specification: Clarify section 'Constant Constructor'
Cf. https://github.com/dart-lang/sdk/issues/36315, section 'Constant Constructor' in the language specification explicitly mentions that a 'constant constructor' has an initializer list and a superinitializer, but fails to make it explicit that it must then be a non-redirecting generative constructor.

Consequently, [this sentence](https://github.com/dart-lang/language/blob/965c37cc3a225abca63567c127840df0bf492613/specification/dartLangSpec.tex#L3428) looks like it is applicable to _every_ kind of constructor which is constant, but it is actually only applicable for generative constant constructors. We can restrict that again to non-redirecting generative constructors, because we can't have a redirecting generative constructor without also having a non-redirecting generative constructor in the same class.

Step 1: Change the text to say 'generative constant constructors' rather than 'constant constructors' in this section. Step 2: For any properties lost by this change (things said about other kinds of constructors) which are actually intended to hold, check that the relevant sections do say those things already, and otherwise add it there.

eernstg:
This CL updates the language specification section 'Constant Constructors' to cover the right kinds of constructors only, and moves a long commentary on potentially constant expressions to the section 'Constants' where potentially constant expressions are defined. This resolves #289.

Note that the grammar rule for `<literal>` was also moved as well: From 'Constants' (where there was no justification for having it) to just before the rule for `<primary>`, which is the place where `<literal>` occurs in the grammar. The individual kinds of literals (such as `<booleanLiteral>` and `<setLiteral>`) have always been defined elsewhere, each in their own section on the associated kind of object.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -3430,7 +3430,9 @@ Constant constructors have stronger constraints than other constructors.
 For instance, all the work of a non-redirecting generative constant constructor
 must be done in its initializers
 and in the initializing expressions of
-the instance variables of the enclosing class.%
+the instance variables of the enclosing class
+(and the latter may already have happened earlier,
+because those initializing expressions must be constant).%
 }
 
 \LMHash{}%
@@ -3449,15 +3451,18 @@ The above refers to both locally declared and inherited instance variables.%
 }
 
 \LMHash{}%
-If a non-redirecting generative constant constructor is declared by a class $C$,
+If a non-redirecting generative constant constructor $k$
+is declared by a class $C$,
 it is a compile-time error
 for an instance variable declared in $C$
 to have an initializing expression that is not a constant expression.
 
 \commentary{%
-A superclass of $C$ cannot have such an initializing expression either,
-because it must necessarily declare
-a non-redirecting generative constant constructor.%
+A superclass of $C$ cannot have such an initializing expression $e$ either.
+If it has a non-redirecting generative constant constructor
+then $e$ is an error,
+and if it does not have such a constructor
+then the (implicit or explicit) superinitializer in $k$ is an error.%
 }
 
 \LMHash{}%
@@ -5983,7 +5988,7 @@ The rules for identity make it impossible for a Dart programmer to observe wheth
 \commentary{%
 All usages of 'constant' in Dart are associated with compile time.
 A potentially constant expression is an expression that will generally yield
-a constant value when the value of certain parameters is given.
+a constant value when the values of certain parameters are given.
 The constant expressions is a subset of the potentially constant expressions
 that \emph{can} be evaluated at compile time.%
 }
@@ -6075,7 +6080,7 @@ not a constant expression (\ref{const}).
 that occurs in a constant context, is a potentially constant expression if $T$ is a constant type expression, and $e_1$, \ldots{} , $e_n$ are constant expressions.
 It is further a constant expression if the list literal evaluates to an object.
 
-\item A constant set literal (\ref{set}),
+\item A constant set literal (\ref{sets}),
 \code{\CONST{} <$T$>\{$e_1$, \ldots{}, $e_n$\}}, or
 \code{<$T$>\{$e_1$, \ldots{}, $e_n$\}}
 that occurs in a constant context, is a potentially constant expression if $T$ is a constant type expression, and $e_1$, \ldots{} , $e_n$ are constant expressions.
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -3411,89 +3414,54 @@ \subsubsection{Constant Constructors}
...
 
-%\commentary{Spell out subtleties: a constant constructor call within the initializer of a constant constructor is treated as a ordinary constructor call (a new), because the arguments cannot be assumed constant anymore. In practice, this means two versions are compiled and analyzed. One for new and one for const.}
-
-\commentary{
-All the work of a constant constructor must be handled via its initializers.
+\commentary{%
+Constant constructors have stronger constraints than other constructors.
+For instance, all the work of a non-redirecting generative constant constructor
+must be done in its initializers
+and in the initializing expressions of
```
lrhn: Not even in the initializing expressions of its instance variables - those must all be compile-time constant, so there is no work there.

specification/dartLangSpec.tex
```diff
@@ -3411,89 +3414,54 @@ \subsubsection{Constant Constructors}
...
+for an instance variable declared in $C$
+to have an initializing expression that is not a constant expression.
 
-\commentary{
-The problem is not that the assignments to \code{w} are not potentially constant; they are.
-However, all these run afoul of the rules for constant lists (\ref{lists}), maps (\ref{maps}) and objects (\ref{const}), all of which independently require their subexpressions to be constant expressions.
+\commentary{%
+A superclass of $C$ cannot have such an initializing expression either,
+because it must necessarily declare
+a non-redirecting generative constant constructor.%
```
lrhn: Technically not true. it can have a non-constant field initializer and no non-redirecting generative constant constructor. It's an error, but that's stated elsewhere (later).
So, rephrase: 

  A superclass of $C$ cannot have such an initializing expression either
  if it has a non-redirecting generative constant constructor, 
  as required by the super-invocation of the non-redirection generative constant
  constructor of this class in the following paragraph.

In general, don't state anything about what a program can or can't do. It can do everything, some of it causes errors.

specification/dartLangSpec.tex
```diff
@@ -5991,19 +5969,23 @@ \subsection{Object Identity}
 \subsection{Constants}
 \LMLabel{constants}
 
-\commentary{
-All usages of the word 'constant' in Dart are associated with compile time.
+\commentary{%
+All usages of 'constant' in Dart are associated with compile time.
 A potentially constant expression is an expression that will generally yield
 a constant value when the value of certain parameters is given.
```
lrhn: Should this be "the values of certain parameters are given"?

specification/dartLangSpec.tex
```diff
@@ -6180,7 +6163,7 @@ \subsection{Constants}
   evaluates to an instance of \code{String}.
 
 % New in 2.1.
-\item An expression of the form \code{$e$ as $T$} is potentially constant
+\item An expression of the form \code{$e$\,\,as\,\,$T$} is potentially constant
```
lrhn: Why is this neccessary? It's highly annoying if \code{...} does not provide the correct spacing (mono-spaced preferably). I do not want to write `\,\,` instead of space in code in general, so why is it necessary here?

specification/dartLangSpec.tex
```diff
@@ -6281,66 +6293,120 @@ \subsection{Constants}
...
+  \CONST{} C(p, q): x = q, y = p + 100, z = p + q;
+\}
+\end{dartCode}
 
 \commentary{
-As an example, consider:
+The assignment to \code{x} is allowed under the assumption that \code{q} is constant (even though \code{q} is not, in general a compile-time constant).
+The assignment to \code{y} is similar, but raises additional questions.
+In this case, the superexpression of \code{p} is \code{p + 100}, and it requires that \code{p} be a numeric constant expression for the entire expression to be considered constant.
+The wording of the specification allows us to assume that \code{p} evaluates to an integer.
```
lrhn: In a constant setting, but not in a non-constant invocation.

specification/dartLangSpec.tex
```diff
@@ -6281,66 +6293,120 @@ \subsection{Constants}
...
+A similar argument holds for \code{p} and \code{q} in the assignment to \code{z}.
```
lrhn: Only because you already derived that `p` was integer, otherwise they could be strings.

specification/dartLangSpec.tex
eernstg: Constant expression evaluation is allowed to take place late. Added a few words clarify this.

specification/dartLangSpec.tex

specification/dartLangSpec.tex
eernstg: We have 73 occurrences of `\,` in `\code{...}`, generally when the `\code` occurs in text. The reason for having them is that the spacing is not always proportional (in particular `!   nextthing` gets a crazy large space), and also that it's generally confusing that the spaces (even when proportional) are so much larger in `a   code   example` in the middle of a sentence. Let's take that issue separately.

specification/dartLangSpec.tex

specification/dartLangSpec.tex
eernstg: Added text to indicate this.

specification/dartLangSpec.tex
eernstg: I kept this one as-is: Even if the argument goes '... \code{p} be a numeric or string constant ...' it's not unreasonable to say that it is similar.

Issue #177, opened 2019-01-14
Scoped Class Extensions


eernstg:
In response to #40, this is a proposal for a mechanism, _scoped class extensions_, that allows class hierarchies to be extended with new instance methods. This proposal is intended to be compatible with the scoped static extension method mechanism (#41), in the sense that those two proposals can be combined into a single one, or they can be adopted one after the other. Scoped static extension methods can do things that are not possible with scoped class extensions, and vice versa, so it makes sense to have both.

## Motivation

**Like #41**, this is a proposal for a mechanism that allows developers to add new methods to existing  receivers without editing the corresponding existing declarations.

The special **advantage of this proposal** is the ability to enhance an existing class hierarchy with new methods that are subject to object-oriented dispatch. In other words, whereas scoped static extension methods are similar to static methods, this proposal enables something which is similar to adding instance methods to the target class hierarchy, without editing it.

The [**visitor** design pattern](https://en.wikipedia.org/wiki/Visitor_pattern) is a well-known software engineering idiom for which a main selling point is that it allows developers to, sort of, add a new instance method to an existing class hierarchy. This is considerably less convenient than a real instance method, however:

- With a visitor, the target hierarchy (that is, the classes whose instances we wish to visit) must implement special support for double dispatch. That is, if we wish to visit instances of a class `C` with a visitor of type `Visitor`, then the class `C` must declare an `accept(Visitor v)` method that invokes the method in the visitor that corresponds to `C`.
- With a visitor, extra ceremony must be applied in order to specify a behavior which is similar to object-oriented overriding (e.g., every visit method must have a default implementation that makes an emulated "superinvocation", and care must be taken for repeated overrides).
- With a visitor, it is necessary for clients to use a different syntax in order to "invoke the added method": (they typically need to do `myVisitor.accept(myReceiver)`), and with a given visitor type, it is impossible to directly specify different signatures, so the approach taken could often be to make the visitor generic and use that to specify the return type, and then to store arguments to the invocation in the visitor itself, etc.
- Finally, it is necessary to _edit_ the visitor itself in order to broaden the support such that the visitor can be used on any new kinds of objects that weren't taken into account (and maybe weren't even written) when the visitor was written.

In other words, we already have a design pattern that is well-known for being able to "add a new method" to a given class hierarchy, but it is quite inconvenient to use.

[...]

Issue #41, opened 2018-10-11
Static Extension Methods


lrhn:
UPDATE: Feature specification is available here:
https://github.com/dart-lang/language/blob/master/accepted/future-releases/static-extension-methods/feature-specification.md

Original in-line proposal:
----

Possible solution strategy for #40. This issue doesn't (currently) have an concrete proposal.

Scoped static extension methods are successfully used in C# and Kotlin to add extra functionality to existing classes.

The idea is that a declaration introduces a static extension method in a scope. The extension method is declared with a *name* (and signature) and on a *type*, and any member access with that name (and signature) on something with a matching static type, will call the extension method.
Example (C# syntax):
```c#
class Container { // Because all methods in C# must be inside a class.
  public static String double(this String input)   // `this` marks it as an extension method on String
  { 
    return input + input; 
  }
}
...
   String x = "42";
   String y = x.double();  // "4242"
```
The method is a *static* method, all extension methods do is to provide a more practical way to invoke the method.

In Kotlin, the syntax is:
```Kotlin
fun String String.double() {
    return this + this;
}
```
The `String.` in front of the name marks this as an extension method on `String`, and the body can access the receiver as `this`. Apart from these syntactic differences, the behavior is the same.

It's possible to declare multiple conflicting extension methods. Say, you add a `floo` extension method to both `List` and `Queue`, and then I write `myQueueList.floo()`. It's unclear which one to pick. In case of a conflict, the usual approach is to pick the one with the most specific receiver type (prefer the one on `List` over the one on `Iterable`), and if there is no most specific receiver type, it's a static error.
C# and Kotlin both allow overriding by signature, so there risk of conflict is lower than it would be in Dart, but the same reasoning can apply.

So, Dart should also have scoped extension methods. We can probably get away with a Kotlin-like syntax, for example:
```dart
String String.double() => this + this;
```
Extension methods can be used on any *type*, including function types and `FutureOr` (although probably not very practically on the latter). It can also be a generic function.
Example:
```dart
T List<T>.best<T>(bool preferFirst(T one, T other)) =>
    this.fold((T best, T other) => preferFirst(best, other) ? best : other);
...
List<int> l = ...;
print(l.best((a, b) => a > b));
```
In this case, the type argument should probably be inferred from the static type of the receiver (which it wouldn't be if the receiver was just treated like an extra argument).

Another option is to allow:
```dart
T List<var T>.best(bool preferFirst(T one, T other)) => ...
```
[...]

PR #303, opened 2019-04-08
Initial draft of static extension methods design document

lrhn:


mit-mit:
Feature is https://github.com/dart-lang/language/issues/41

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,375 @@
...
+extension BoxList<T> on Box<List<T>> { foo() {} }
+extension BoxSpec on Box<List<num>> { foo() {} }
+...
+  List<int> x = ...;
+  x.foo();
+```
+
+Here all three extensions apply. The most specific one is `BoxSpec` which is specialized for lists of numbers. If `BoxSpec` had not been in scope, then neither of `BoxCom` or `BoxList` would be more specific than the other. Their `on` types, when instantiated to bounds, are `Box<Iterable<Comparable<dynamic>>>` and `Box<List<dynamic>>` which are unrelated by subtyping.
+
+This is also why we use the instantiated-to-bounds type for comparison. If we used the actual type, bound to the type variable by the current use-case, we would consider `BoxList<int>` to be more specific than `BoxSpec`, but in practice, the `BoxList<int>` extension would not be able to *use* the integer-ness of the type in its code, so `BoxSpec`, which can specialize its code for lists of numbers, is the most precise match.
```
leafpetersen: Counter-argument: if you used the actual type, and `BoxSpec` wasn't in scope, then `BoxList` would be most specific, and it could take advantage of the `Listness` of the type.  But if you use the instantiated-to-bounds type, then neither is most specific and you get an error.

I'm not sure what the right answer is here.  Do we know what other languages do?

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,375 @@
...
+
+### Overriding Access
+
+If two or more extensions apply to the same member access, or if a member of the receiver type prevents using an extension method, or if the extension is shadowed, then it is possible to force an extension member invocation:
+
+```dart
+MyList(object).quickSort();
+```
+
+The syntax looks like a constructor invocation, but it does not create a new object.
```
leafpetersen: Presumably if `MyList` is generic you should be able to pass the explicit generic arguments as well.

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,375 @@
...
+}
+...
+  MySmart.smartHelper(someObject);  // valid
+```
+
+### Semantics of Invocations
+
+If an extension is found to be the one applying to a member invocation, then at run-time, the invocation will perform a method invocation of the corresponding instance member of the extension, with `this` bound to the receiver value and type parameters bound to the types found by static inference.
+
+If the receiver is `null`, then that invocation is an immediate run-time error unless the `on` type of the extension has a trailing `?`.
```
leafpetersen: Ah, I maybe understand your previous comment better - you're not assuming NNBD.  In post NNBD world, does this mean "unless the on type is nullable"?  Or "unless the on type is potentially nullable"?

e.g.

```dart
extension E1<T extends Object?> on T {
  void m() {};
}
extension E2 on FutureOr<Object?> {
  void n() {};
}
void test() {
  null.m(); // OK?
  null.n(); // OK?
}
```

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,375 @@
...
+
+With NNBD types, we will not allow a non-nullable extension `on` type to apply to a member invocation with a nullable receiver type.
+
+### Semantics of Extension Members
```
leafpetersen: I think I buy all of this, but it's very finicky stuff: by which I mean, we should think hard about this, and then think hard about it again... :)

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,375 @@
...
+
+The *type* can be any valid Dart type, but not a single type variable. It can refer to the type parameters of the extension. It can be followed by `?` which means that it allows `null` values.
+
+The member declarations can be any static or instance member declaration except for instance variables and constructors.
+
+### Scope
+
+Dart's static extension methods are *scoped*. They only apply to code that the extension itself is accessible to. Being in accessible means that using the extension's name must denote the extension. The extension is not in scope if another declaration with the same name shadows the extension, if the extension's library is not imported, if the library is imported and the extension is hidden, or the library is only imported with a prefix. In other words, if the extension had been a class, it is only in scope if using the name would denote the class&mdash; which will allow you to call static methods on the class, which is exactly what we are going to do.
```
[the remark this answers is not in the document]
lrhn: You can alwasy do scope override with import prefixes, you just have to do the *explicit* `quiver.CoolList(myList).foo()` override. You don't get the implicit suffix invocation when you override.

It is true that a conflict of name will prevent you from accessing the extension, just as it would prevent you from accessing a class. If you have two different classes names `CoolList`, then you can't call a static method on either.

We could choose to allow conflicts where all the names are extensions, as long as the member resolution gives a unique result, so if you have two `CoolList` extensions in scope, but only one of them has a `foo` member, then `[42].foo()` is not considered a conflict. Or if both have a `foo`, but one is more specific than the other, then it is not considered a conflict.
I am wary at allowing this because it would suggest that declaring those in the same scope should also be allowed (you can have them in the same scope, so why not declared them in the same scope?), and then people might think the following is a reasonable design:
```dart
extension CoolList<T> on List<T> { T foo() ... }
extension CoolList on List<int> { int foo() ...int specific code ... }
extension CoolList on List<String> { String foo() ...String specific code ... }
```
If library authors start doing that, then you *cannot* do an explicit override any more because that depends on explicit name resolution.

I think keeping the "at most one definition of a name in each scope" rule is important for language sanity.

That leaves us with the ability to *open* more than one extension with the same name in a specific scope, which means renaming.

We can introduce the "extension alias" you suggest. We could also go one step further and introduce a renaming operation directly in the import modifiers:
```dart
import "package:list/list.dart" show CoolList;
import "package:quiver/quiver.dart" show CoolList as QuiverList;
// or: import "package:quiver/quiver.dart" hide CoolList as QuiverList;
```
That would allow renaming any import. It's a trivial rename, all uses of the new name will refer to the imported declaration.
In the importing library, if the name `QuiverList` is in scope and not shadowed, then the quiver `CoolList` extension applies.

I'm sure people will have lots of fun with `import "dart:core" hide dynamic as Any;`. :smiling_imp:

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,375 @@
...
+```ebnf
+<extension> ::= 
+  `extension' <identifier> <typeParameters>? `on' <type> `?'? `{'
+     memberDeclaration*
+  `}'
+```
+
+Such a declaration introduces its *name* (the identifier) into the surrounding scope. The name does not denote a type, but it can be used to denote the extension itself in various places. The name can be hidden or shown in `import` or `export` declarations.
+
+The *type* can be any valid Dart type, but not a single type variable. It can refer to the type parameters of the extension. It can be followed by `?` which means that it allows `null` values.
```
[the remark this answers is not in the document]
lrhn: Erik also couldn't find any reason not to, and I don't remember why I was worried about it.
It's otherwise equivalent to `on Object` (or on the bound of the type variable), but you get the static receiver type as a type. That should be simpler than having to deconstruct the receiver type.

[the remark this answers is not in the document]
leafpetersen: The other option I've kicked around is always providing the self type.  e.g., we just say that there's always an implicit `Self` type variable in scope that is the matched type.  Then, as you say, you could just put it on `Object`.  I think for this use case it's simpler to just allow `T` as a pattern.  What I'm not sure of is whether 

```
extension<T> on TypePattern {}
```

applies to exactly the same set of things as:

```
extension<T extends TypePattern> on T {}
```

It seems like it ought to?  And if so, you can always get the self type.  But if not, then you can't both get the self type and have a specific pattern.

accepted/future-releases/static-extension-methods/design-document.md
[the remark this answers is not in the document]
leafpetersen: > Then again, I'm not sure why we will need to give more ways to avoid scope conflict for extensions than we do for other declarations.

My argument would be that prefixing other kinds of things doesn't change their affordances.  There is nothing you can do with a class `Foo` that you can't do with `foo.Foo`, as far as I know.  And similarly for functions, getters, setters, etc.  But with extensions, there is: prefixing it stops the implicit open of the extension in scope.  I'm really worried that if providing extensions as part of APIs becomes really common, then basically everyone will have to name mangle their extension names to be globally unique or else clients won't be able to use them as intended.  

I'd be fine with import based renaming.  I'd be fine with explicit renaming.  

Another approach would be to allow "opening" and extension explicitly.  So you would do:

```dart
import "package:list/list.dart" show CoolList;  // Bring CoolList into scope, but not CoolList.foo
open CoolList;  // Bring CoolList.foo into the extension method resolution scope.
```

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,375 @@
...
+More precisely, an extension declaration is a top-level declaration with a grammar similar to:
+
+```ebnf
+<extension> ::= 
+  `extension' <identifier> <typeParameters>? `on' <type> `?'? `{'
+     memberDeclaration*
+  `}'
+```
+
+Such a declaration introduces its *name* (the identifier) into the surrounding scope. The name does not denote a type, but it can be used to denote the extension itself in various places. The name can be hidden or shown in `import` or `export` declarations.
```
leafpetersen: What do you think about making the name optional?  I really do think there is a good use case for file local extensions, where I want to define a bunch of helper operations on some types, but I don't plan to export them.  Having to pick a meaningless name that doesn't conflict with anything else in the file or in the imported extensions seems like pointless work:  why not just allow me to say:

```
extension on int {
  // My int helper methods here
}
extension on String {
  // My String helper methods here
}
```

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,375 @@
...
+
+The *type* can be any valid Dart type, but not a single type variable. It can refer to the type parameters of the extension. It can be followed by `?` which means that it allows `null` values.
+
+The member declarations can be any static or instance member declaration except for instance variables and constructors.
+
+### Scope
+
+Dart's static extension methods are *scoped*. They only apply to code that the extension itself is accessible to. Being in accessible means that using the extension's name must denote the extension. The extension is not in scope if another declaration with the same name shadows the extension, if the extension's library is not imported, if the library is imported and the extension is hidden, or the library is only imported with a prefix. In other words, if the extension had been a class, it is only in scope if using the name would denote the class&mdash; which will allow you to call static methods on the class, which is exactly what we are going to do.
+
+### Extension Member Resolution
```
leafpetersen: It's not clear to me from this section whether you intend to allow tear offs are not.  I am in favor, I think it's the uniform thing to do, and useful, but either way we should be clear.

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,449 @@
...
+
+The declaration introduces an extension. The extension's `on` type defines which types are being extended.
+
+For any member access, `x.foo`, `x.bar()`, `x.baz = 42`, `x(42)` or `x + y`, the language first checks whether the static type of `x` has a member with the same base name as the operation. That is, if it has a corresponding instance member, respectively, a `foo` method or getter or a `foo=` setter. a `bar` member or `bar=` setter, a `baz` member or `baz=` setter, a `call` method, or a `+` method. If so, then the operation is unaffected by extensions. *This check does not care whether the invocation is otherwise correct, based on number or type of the arguments, it only checks if there is a member at all.*
+
+(The types `dynamic` and `Never` are considered as having all members, the type `void` is always a compile-time error when used in a receiver position, so none of these can ever be affected by static extension methods).
```
leafpetersen: In discussions with @srawlins today, some use cases came up that made me question whether treating dynamic calls as never resolving to an extension is really the right thing to do.  The concrete example was JSON, where you could imagine defining a package which adds `.toInt` etc to `dynamic`.  It also occurred to me that if you use extension methods to define things like kotlin's `let`, `apply`, and `also`, it's a bit weird that you can't use them on things of type dynamic.  Is it really useful to keep extensions that apply to everything from applying to dynamic?

Hixie:
I'm concerned about the "COMEFROM" nature of extension methods (where do i go to find where this method i'm using is implemented?).

I think overall in the Flutter framework we'd probably avoid using it, except for one thing, which is adding methods to specific specializations. For example, right now `Animation<T>` has a `drive` method that asserts `T == double`. I'd be happy to add an `extension AnimationDouble on Animation<double>` that added `drive`.

My biggest question is, how will this look in the dartdocs?

accepted/future-releases/static-extension-methods/design-document.md
lrhn: Absolutely, or have it inferred like here.

accepted/future-releases/static-extension-methods/design-document.md
lrhn: In NNBD-world, if the `on` type is nullable, then it applies to a nullable static receiver type, if not, it doesn't. There won't be any unsafe `null` errors (except in the presence of legacy unsafe-nullable `*` types).

So, the example is OK. Since `null` is a subtype of `Object?` and of `FutureOr<Object?>`, the `n` and `m` extension methods match the `null.m()` (with `T` inferred as `Null`) and `null.n()` invocations.

Since the `on` types are nullable, the `this` references inside the methods will be nullable as well, so it's no surprise that they might be `null`.

If the receiver type is potentially nullable

So, using NNBD:
```dart
extension E1 on Object { m1() { assert(this != null); } }
extension E2 on Object?  { m2() {} }
void main() {
  Object o1 = 1;
  o1.m1(); // OK, Object <: Object
  o1.m2(); // OK, Object <: Object?
  Object? p1 = 1;
  p1.m1(); // compile-time error, there is no m1 method matching Object?.
  p1.m2(); // OK, Object? <: Object?
  Object? p2 = null;
  p2.m1(); // compile-time error, there is no m1 method matching Object?.
  p2.m2(); // OK, Object? <: Object?.
}
```

Using non-NNBD semantics, where the `?` on the type means "accepts null":
```dart
extension E1 on Object { m1() { assert(this != null); } }
extension E2 on Object?  { m2() {} }
void main() {
  Object o1 = 1;
  o1.m1(); // OK, o1 != null
  o1.m2(); // OK.
  Object? o2 = null;
  o2.m1(); // run-time error, o2 == null at runtime.
  o2.m2(); // OK, o2 == null at runtime is accepted and this == null.
}
```

accepted/future-releases/static-extension-methods/design-document.md
lrhn: I'm not sure how often library-private extensions occur. I'd wait and see if there is a market for this feature. All you need to do is write `_1` instead if you really don't care about the name.

accepted/future-releases/static-extension-methods/design-document.md
lrhn: The design is based on the idea that existing methods takes precedence over extension methods because that way adding extension methods cannot break existing code. 
It will allow us to add extension methods to the platform libraries, and not break anyone who uses an existing member with the same name on their own subtype of the type.

That reasoning should apply to dynamic invocations too. If someone writes `dynamicValue.foo` anywhere, we should not let an `extension o on Object? { int get foo => 42; }` override their existing behavior.

If we are willing to drop that protection, maybe we can do so everywhere, and let extension methods take precedence over existing methods. That would be a very different feature (one which allows overriding `String` methods).
It will probably also mean that we cannot add extension methods on any subclassable platform type to any existing platform library.

lrhn:
> My biggest question is, how will this look in the dartdocs?

It will be documented where the extension is declared. The DartDoc tool may want to special-case the situation where an extension is defined in the same library as a class that it applies to, and give the class at least a link to the extension methods.

Apart from that, we just have to figure out how to document the type. I'll leave that to the DartDoc team, they have much more experience with presenting the relevant information.

Perhaps `extension Foo<T extends Comparable<T>> on List<T> { void fastSort() { ... } }` could be documented as 
```dart
extension Foo on List<T extends Comparable<T>> 
instance members:
  void fastSort() { ... }
```
?

accepted/future-releases/static-extension-methods/design-document.md
lrhn: tear-offs are described below now.

accepted/future-releases/static-extension-methods/design-document.md
leafpetersen: > All you need to do is write `_1`

I know, and that's horrid.  We can certainly wait on allowing it, but there's an opportunity cost to waiting.  If we don't do this now, odds are very high we'll never do it.  Every change has a fixed cost.

I would expect library private extensions to be approximately as common as library private functions/static methods, since that's what we're doing: giving people a way to write functions that have a nice call syntax.

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,493 @@
...
+Here we perform type inference equivalent to what we would do for:
+
+```dart
+class MyList<T extends Comparable<T>> {
+  MyList(List<T> argument);
+  void quickSort() { ... }
+}
+...
+Uint8List bytes = ...;
+... MyList(bytes).quickSort() ... // Infer type argument of MyList here.
```
leafpetersen: You should make clear that the context type for the invocation is not used as part of this inference process.

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,493 @@
...
+```
+
+This is inference based on static types only. The inferred type argument becomes the value of `T` for the function invocation that follows.
+
+If the inference fails, or if the synthetic constructor invocation would not be valid for any other reason, then the extension does not apply. If the extension does not have an instance member with the base name `quickSort`, it does not apply.
```
leafpetersen: "If an extension does not have..."?  Otherwise, I'm not sure what extension we're talking about.

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,493 @@
...
+
+Example:
+
+```dart
+extension MySmart on Object {
+  smart() => smartHelper(this);  // valid
+  static smartHelper(Object o) { ... }
+}
+...
+  MySmart.smartHelper(someObject);  // valid
```
leafpetersen: Do we disallow references to generic type parameters in static methods?  Otherwise we need to allow `MyGenericSmart<int>.smartHelper(someObject)`.

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,493 @@
...
+
+```dart
+extension MyUnaryNumber on List<Object> {
+  bool get isEven => this.length.isEven;
+  bool get isOdd => !this.isEven;
+  static bool isListEven(List<Object> list) => list.isEven;
+}
+```
+
+Here the `list.isEven` is  guaranteed to hit the `isEven` of the same extension (unless someone puts an `isEven` member on `List`), an extension has more specificity than any other extension inside itself.
```
leafpetersen: We should probably say explicitly that you can use `MyUnaryNumber(object).isEven` inside the definition of `MyUnaryNumber` to reach the extension even in the presence of an instance method.

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,493 @@
...
+
+The behavior is *only* for the `this` operator. Doing something like `var self = this; self.isEven;` will not give preference over members of `List` (if `List` had an `isEven`, then `self.isEven` would invoke that, where `this.isEven` will invoke the extension method). It is as if `this` has a type representing the extension, even if no such type actually exists.
+
+This behavior makes the extension act similarly to a class, which will hopefully make it easier for users to understand the resolution.
+
+Like for a class or mixin member declaration, the names of the extension members, both static and instance, are in the *lexical* scope of the extension member. That is why `MySmart` above can invoke the static `smartHelper` without prefixing it by the extension name. In the same way, *instance* members are in the lexical scope.
```
leafpetersen: By *instance* members, you mean other extensions?

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,493 @@
...
+  bool get isOdd => !this.isEven;
+  static bool isListEven(List<Object> list) => list.isEven;
+}
+```
+
+Here the `list.isEven` is  guaranteed to hit the `isEven` of the same extension (unless someone puts an `isEven` member on `List`), an extension has more specificity than any other extension inside itself.
+
+About *`this`*: Inside an instance member of an extension, the static type of the `this` operator is the `on` type. However, for member invocations, we give preference to extension methods of the current extension, *even over members of the `on` type*. 
+
+Here, the `this.isEven` is a member invocation on `this`, so it *first* tries to resolve `isEven` against the current extension. If it finds a member with the same base name on the extension, that member is invoked with the same `this` value. If not, then it is considered a normal member invocation on an object with static type `List<Object>,` and if `List<Object>` does not have the necessary member, then all applicable extensions are checked (which will definitely not match the current extension since we already checked that).
```
leafpetersen: I'm not sure what to think about this.  It feels a bit odd to resolve these differently from external uses, but it also feels a bit odd for unqualified references to resolve to the instance method.

accepted/future-releases/static-extension-methods/design-document.md
munificent: Yeah, Dart already forbids that. Static members exist on the *class's* namespace, and don't hang off the instantiated type or have access to the class's type arguments.

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,493 @@
...
+  MySmart.smartHelper(someObject);  // valid
+```
+
+### Semantics of Invocations
+
+If an extension is found to be the one applying to a member invocation, then at run-time, the invocation will perform a method invocation of the corresponding instance member of the extension, with `this` bound to the receiver value and type parameters bound to the types found by static inference.
+
+If the receiver is `null`, then that invocation is an immediate run-time error unless the `on` type of the extension has a trailing `?`.
+
+With NNBD types, it is a run-time error if the receiver is `null` and the instantiated `on` type of the selected extension does not allow `null`. For sound non-nullable types, this can be excluded statically. If a receiver expression is nullable, then the matching `on` type must be nullable too. If a receiver expression is soundly non-nullable, then the value cannot be `null` at run-time. During migration, we will have unsound nullable types like `int*` which should be matched by `extension Wot on int {…}`, and in that case we will need a run-time check to avoid passing `null` to that extension.
```
munificent: This paragraph is ordered a little funny. The first sentence makes it sound like there's no static checking of nullable receivers in NNBD but then you get to the end and it seems that's only related to legacy types? It might be good to reorder this to be clearer.

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,493 @@
...
+
+### Semantics of Extension Members
+
+When executing an extension instance member, we stated earlier that the member is invoked with the original receiver as `this` object. We still have to describe how that works, and what the lexical scope is for those members.
+
+Inside an extension declaration, we give preference to that extension over any other extension. At any otherwise invalid invocation, if the current extension applies, then it is considered more specific than all other extensions that may apply. This applies to static methods too.
```
munificent: This feels unnecessarily special to me. Is it really worth having different resolution rules inside an extension and outside?

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,493 @@
...
+Example:
+
+```dart
+extension MyList<T> on List<T> {
+  void add(T value, {int count = 1}) { ... }
+  void add2(T value1, T value2) { ... }
+}
+```
+
+You cannot *access* this member in a normal invocation, so it could be argued that you shouldn't be allowed to add it. We allow it because we do not want to make it a compile-time error for a type to add a method just because an extension is already adding a method with the same name. it will likely be a problem if any code *uses* the method, but only that code needs to change (perhaps using an override).
```
munificent: "for a type to add a method" -> "to add an instance method to an existing class".

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,493 @@
...
+
+That also means that an extension can shadow members of the `on` type.  To implement `add2` above, we don't want to call `this.add`, instead we want to use the `add` of `List` directly. To do that, we allow `super` invocations to target the underlying `on` type directly, with *no* extensions applying.
+
+```dart
+void add2(T value1, T value2) {
+  super..add(value1)..add(value2);
```
munificent: I'm worried this encourages a mental that really isn't correct. Extensions aren't on a magic subclass of the type. If they were, then a non-self call would hit the extension first and not the instance method.

Are you sure this `super()` stuff is needed at all?

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,493 @@
...
+- `extension X on FutureOr<int>`
+- `extension X on int?`
+
+For the first two, we could put the static members on the `Iterable` class, but since the extension does not apply to *all* iterables, it is not clear that this is correct.
+
+For `int Function(int)` and `FutureOr<int>`, it's unclear how to call such a static method at all. We can denote`int Function(int)` with a type alias, but putting static members on type aliases is a new concept.  We could put the static method on `Function`, but that's not particularly discoverable, and why not require that they are put on `Functon` explicitly. For `FutureOr`, we could allow static members on `FutureOr` (which is a denotable type), but again it seems spurious. For `int?`, we could put the method on `int`, but why not  just require that it's on `int`.
+
+The issue here is that the type patterns used by `on` are much more powerful than what is necessary to put static members on class types.
+
+It would probably be more readable to introduce a proper static extension declaration:
```
munificent: You could also simply disallow static members in extensions on non-simple class types.

That prohibits the perhaps useful cases of wanting a static member that can only be called through the extension itself, but I'm not sure that's that useful anyway.

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,493 @@
...
+The grammar then becomes:
+
+```ebnf
+<extension> ::= 
+  `extension' <identifier>? <typeParameters>? `on' <type> `?'? `{'
+     memberDeclaration*
+  `}'
+```
+
+This is a simple feature, but with very low impact. It only allows you to omit a single private name for an extension that is only used in a single library. Unless there is a documented demand for this feature, it doesn't seem worth the effort.
```
munificent: For what it's worth, it is an annoying source of friction to come up with a name for an extension, and is probably even more annoying when it's only used locally. Anonymous extensions would help, though only for private ones.

I guess a simpler solution is for us to just establish a convention and tell people to follow it so they don't have to think too much about the name.

accepted/future-releases/static-extension-methods/design-document.md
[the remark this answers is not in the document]
lrhn: It should work on the same things, yes.

If `foo(TypePattern x) {}; foo(expression);` is allowed, then so is `foo<T extends TypePattern>(T x) {}; foo(expression);` for exactly the same expressions. We use exactly the same kind of type inference to match the `on` type as we do for such a function invocation.

As for getting both, I had hoped that the following would work:
```dart
extension Foo<E, S extends List<E>> on S {
   S get self => this;
   List<E> clone() => List<E>.of(this);
}
```
but I seem to get `dynamic` for `E` when I try this with a function. :frowning_face:

accepted/future-releases/static-extension-methods/design-document.md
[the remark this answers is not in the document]
lrhn: I have a scope override for opting in to an extension, but none for opting out. You need to remove the extension from the scope for that. That is probably unavoidable since we use the same syntax for everything, so there is no syntax we can write to do "the normal thing", because that is the syntax we are using already.

accepted/future-releases/static-extension-methods/design-document.md
leafpetersen: Yes, unnamed extensions would not be exported.  The semantics would be that the compiler invents a unique library private name to fill in for the missing name.

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,493 @@
...
+```ebnf
+<extension> ::= 
+  `extension' <identifier> <typeParameters>? `on' <type> `?'? `{'
+     memberDeclaration*
+  `}'
+```
+
+Such a declaration introduces its *name* (the identifier) into the surrounding scope. The name does not denote a type, but it can be used to denote the extension itself in various places. The name can be hidden or shown in `import` or `export` declarations.
+
+The *type* can be any valid Dart type, including a single type variable. It can refer to the type parameters of the extension. It can be followed by `?` which means that it allows `null` values. When Dart gets non-nullable types by default (NNBD), this `?` syntax is removed and subsumed by nullable types like `int?` being allowed in the `<type>` position.
```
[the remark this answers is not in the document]
leafpetersen: No higher order type parameters!! (yet).

accepted/future-releases/static-extension-methods/design-document.md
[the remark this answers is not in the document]
eernstg: Here's a version that uses a wrapper object rather than a function (I suspect that this will be an even more faithful model for how type inference is performed, given the rest of this document):

```dart
class Foo<E, S extends List<E>> {
  S _self;
  Foo(this._self);
  S get self => _self;
  List<E> clone() => List<E>.of(_self);
}

main() {
  List<int> xs = [1];
  // Emulate extension method invocation `xs.clone()`.
  // At first, use the direct desugaring as wrapper:
  var ys = Foo(xs).clone();
  print(ys.runtimeType); // 'List<dynamic>'.
}
```

In this situation we can't expect type inference to apply any constraints to the first type argument of the `Foo` instance creation (`E`), so it becomes `dynamic`. It would surely be a massively breaking change (and not particularly useful) to "optimize" `E` to `int` just because the second type argument has been inferred to be `List<int>` (assuming that we would do so for type inference everywhere, in order to make sure that we keep the set of special exceptions for extension methods as small as possible).

I believe that the design will satisfy Leaf's rule about applicability ('applies to exactly the same'), but also that we shouldn't expect inference to work the same for `extension<T> on TypePattern {}` and for `extension<T extends TypePattern> on T {}`. So I'd recommend that we just don't aim for having this property, such that we can use the normal rules for type inference with extensions.

accepted/future-releases/static-extension-methods/design-document.md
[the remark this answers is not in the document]
eernstg: We could use generalized type aliases to perform renaming, just like they can rename classes:

```dart
import `pkg:list` as list;
import `pkg:quiver` as quiver;

typedef CoolList<X> = list.CoolList<X>;
typedef QuiverList<X> = quiver.CoolList<X>;
```

These type aliases would be semantically isolated from other type aliases (which is good because it means that we don't promise being able to do anything new just because we're accessing a renamed extension), but it might also be useful because it can be used to pre-resolve some conflicts in a way that will work for the given application:

```dart
import `pkg:list` as list;
import `pkg:quiver` as quiver;

typedef CoolList<X> = list.CoolList<X>;
typedef QuiverList<X extends SomeQuiverishType> = quiver.CoolList<X>;
```

So if `quiver.CoolList` admits all type arguments, but this application will actually only use it when `X extends SomeQuiverishType` then we could use the above to ensure that there is no conflict with `list.CoolList` for any type argument which does not extend `SomeQuiverishType`.

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,482 @@
...
+
+- The override can also be used for extensions imported with a prefix (which are not otherwise in scope): `prefix.ExtensionName(object).method(args)`.
+
+- An invocation of an extension method throws if the receiver is `null` unless the `on` type has a trailing `?`. With NNBD types, the invocation throws if the receiver is `null` and the instantiated `on` type of the selected extension does not accept `null`. (In most cases, this case can be excluded statically, but not for unsafely nullable types like `int*`).
+
+- Otherwise an invocation of an extension method runs the instance method with `this` bound to the receiver and with type variables bound to the types found by type inference (or written explicitly for an override invocation). The static type of `this` is the `on` type of the extension.
+
+- Inside an instance extension member, extension members accessed by unqualified name are treated as extension override accesses on `this`. Otherwise invocations on `this` are treated as any other invocations on the same static type.
```
eernstg: As mentioned in comment on line 263, I believe that this rule would be consistent with the approach proposed in #328, but not consistent with the current rules for the treatment of unqualified identifiers and `this`.

In short, the current rules stipulate that an unqualified name `m` that resolves to an instance member in the lexical scope will _not_ be taken to denote that instance member directly, it will be desugared to `this.m` and then processed as such.

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,482 @@
...
+
+We could allow multiple types in the `extension` `on` clause as well. It would have the following consequences:
+
+- An extension only applies if the receiver type is a subtype of *all* `on` types.
+- An extension is more specific than another if for every `on` type in the latter, there is an `on` type in the former which is a proper subtype of that type, or the two are equivalent, and the former is a proper subtype of the latter when instantiated to bounds.
+- The trailing `?` makes the most sense if it is applied only once (it's the extension which accepts and understands `null` as a receiver), but for forwards compatibility, we will need to put it on every `on` type individually. All `on` types must be nullable in order to accept a nullable receiver.
+- There is no clear type to assign to `this` inside an instance extension method. For a mixin that's not a problem because it introduces a type by itself, and the combined super-interface is only used for `super` invocations. For extension, a statement like `var self = this;` needs to be assigned a useful type.
```
eernstg: Superinvocations are otherwise no longer mentioned in this document. Maybe it should be deleted from here, or some extra explanation should be added in order to specify what it would mean?

My first thought would be that `this.m()` would invoke `m` on the syntactic receiver of the current extension method invocation, and that would be checked against the combined interface of the `on` types, and there's nothing "superinvocated" about this.

accepted/future-releases/static-extension-methods/design-document.md
eernstg: The `super` stuff is no longer present in the proposal (except for a tiny bit in the sections about variants), so I believe we can resolve this thread.

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,482 @@
...
+  bool get isOdd => !isEven;
+  static bool isListEven(List<Object> list) => list.isEven;
+}
+```
+
+Here the `list.isEven` will find that `isEven` of `MyUnaryNumber` applies, and unless there are any other extensions in scope, it will call that.
+
+The unqualified `length` of `isEven` is not defined in the current lexical scope, so is equivalent to  `this.length`, which is valid since `List<Object>` has a `length` getter.
+
+The unqualified `isEven` of `isOdd` resolves lexically to the `isEvent` getter above it, so it is equivalent to `MyUnaryNumber(this).isEven`,  even if there are other extensions in scope which define an `isEven` on `List<Object>`.
```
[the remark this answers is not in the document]
lrhn: I don't think we need anything new, even #328. we can consistently say that a lexically bound identifier refers to the thing in the lexical scope. Whether we do that by a rewrite to `C.foo`, `this.foo` or `Extension(this).foo` is a matter of *what* is in the lexical scope (a static member, an instance member or an extension member).

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,482 @@
...
+- Extensions are declared using the syntax:
+
+  ```ebnf
+  <extension> ::= `extension' <identifier><typeParameters>? `on' <type> `?'?
+     `{'
+       <memberDeclaration>*
+     `}'
+  ```
+
+  where `extension` becomes a built-in identifier, `<type>` must not be a type variable, and `<memberDeclaration>` does not allow instance fields or constructors. It does allow static members.
```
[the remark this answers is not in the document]
lrhn: Agree, I allowed it elsewhere, but missed it here.

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,493 @@
...
+
+For `x.best()`, the most specific one is `BoxList`. Because `Box<List<int>>` is a proper subtype of both ` Box<iterable<int>>` and `Box<List<num>>`, we expect `BoxList` to be the best implementation. The return type causes `v` to have type `int`. If we had chosen `BoxSpec` instead, the return type could only be `num`, which is why we choose the most specific instantiated type as the winner. 
+
+For `y.best()`, the most specific extension is `BoxSpec`. The instantiated `on` types that are compared are `Box<Iterable<num>>` for `BoxCom` and `Box<List<num>>` for the two other. Using the instantiate-to-bounds types as tie-breaker, we find that `Box<List<Object>>` is less precise than `Box<List<num>>`, so the code of `BoxSpec` has more precise information available for its method implementation. The type of `w` becomes `Box<List<num>>`.
+
+In practice, unintended extension method name conflicts are likely to be rare. Intended conflicts happen where the same author is providing more specialized versions of an extension for subtypes, and in that case, picking the extension which has the most precise types available to it is considered the best choice.
```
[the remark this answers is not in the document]
eernstg: FYI, language issue #177 is a proposal for extension methods with dynamic resolution (invoking the implementation that best matches the dynamic type of the receiver). It's designed to be a compatible extension (such that we can add it to the static extension methods feature later on). So the option should still be open.

accepted/future-releases/static-extension-methods/design-document.md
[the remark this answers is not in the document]
eernstg: Static extensions are not (except for some variants discussed briefly at the end) related in any way that resembles the subtype and subclass relation of classes, but I think we need a foundation for how to deal with custom bindings of `this` that will also work for classes (and everything else) in a consistent manner.

In a class, I believe that it would be confusing (and inconvenient) if you can call methods of the enclosing class when they are in the lexical scope, but not if they are inherited from a superclass or just added to the interface of the class by the `implements` clause. #328 specifies that we select the instance member of the enclosing class in the case where the interface of that class has a member of the requested name, and that's not something that we would get from a rule that is based solely on lexical scoping.

accepted/future-releases/static-extension-methods/design-document.md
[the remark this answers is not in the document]
leafpetersen: It feels really off to me to use typedef for renaming these since they... aren't types.  It feels really misleading and confusing.  If we want a declaration, maybe:

```dart
extension CoolList<X> = list.CoolList<X>;
```

?

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,501 @@
...
+  List<int> x = ....;
+  x.doTheSmartThing(print);
+```
+
+Here both the extensions apply, but the `SmartList` extension is more specific than the `SmartIterable` extension because `List<dynamic>` &lt;: `Iterable<dynamic>`.
+
+Example:
+
+```dart
+extension BoxCom<T extends num> on Box<Iterable<T>> { T best() {...} }
```
sigmundch: did you mean `on Iterable<T>` (remove the surrounding `Box<..>` here and below? or change teh code further down to say `Box<List<int>> x = ...`?)

lrhn: Not sure what I intended, but the `Box` type is unnecessary for the example, so I'll remove it.

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,504 @@
...
+extension Bar<T> extends Foo<T> on List<T> { twizzle() { ... } } 
+```
+
+If both of these extensions apply, then pick the one that extends the other. If there are more related extensions which apply, then pick one which transitively extends all the others.
+
+If there isn't exactly one among applicable extensions which extends all the rest, the conflict is a compile-time error.
+
+This approach allows related extensions to declare functionality on a number of types, without accidentally allowing a conflict with an unrelated extension.
+
+The extending extension must be defined on a subtype of its super-extension. In the above, the `on` type of `Bar<T>` is `List<T>`, which is a subtype of the `on` type of `Foo<T>`, which is `Iterable<T>`. Also, the declared extension methods in the extending extension must be valid overrides of the same-named super-extensions extension methods, as if it was a subclass relationship.
```
leafpetersen: Do you mean subtype in general, or proper subtype?  Can I do
```
extension Baz<T> extends Foo<T> on Iterable<T> { twizzle() { ... }}
```

What about

```
extension A<T> on T { twizzle() { ... }}
extension B<T> extends A<T> on T { twizzle() { ... } }
```

The latter would give you a way to override completely generic extensions, which we wouldn't otherwise have, I think?

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,504 @@
...
+
+The above resolution strategy uses an implicit "more specific" relation between applicable extensions to select one of them (when possible). This may cause surprises when two unrelated extensions happen to both apply.
+
+Alternatively, we can allow extensions to declare that they are *related*, and only allow conflicts between related extensions, which are presumably aware of each other.
+
+The syntax would be:
+
+```dart
+extension Foo<T> on Iterable<T> { twizzle() {...} fromp() { ... }}
+extension Bar<T> extends Foo<T> on List<T> { twizzle() { ... } }
```
leafpetersen: Do we need "implements" as well?  Otherwise, it feels like I can get stuck:

```
extension Foo on A { twizzle() {...} }
extension Bar on I  { twizzle() { ... } } 
class B extends A implements I {}
extension Baz extends Foo on A { twizzle() {...} } 
B().twizzle() // Error, because Baz doesn't cover I, right?
```

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,504 @@
...
+}
+```
+
+This is equivalent to giving the extension a fresh private name.
+
+We may need to make `on` a built-in identifier, and not allow those as names of extensions, then there should not be any parsing issue. Even without that, the grammar should be unambiguous because `extension on on on { … }` and `extension on on { … }` are distinguishable, and the final type cannot be empty. It may be *harder* to parse, though.
+
+This is a simple feature, but with very low impact. It only allows you to omit a single private name for an extension that is only used in a single library.
+
+### Scope
```
leafpetersen: Should we provide a way to show/hide individual members of an extension?  Otherwise you're somewhat stuck if you have two extensions that provide overlapping names and you want to pick and choose.

accepted/future-releases/static-extension-methods/design-document.md
lrhn: Probbly not. Interfaces are useful for generalizing over virtual lookups without having any specific implementation. A static invocation is always a concrete implementation, and is unaffected by interfaces.

accepted/future-releases/static-extension-methods/design-document.md
lrhn: Just subtype. You *can* completely shadow the super-extension, but I don't think avoiding that isn't worth much effort.

accepted/future-releases/static-extension-methods/design-document.md
leafpetersen: Just to be sure I understand correctly:

pre-NNBD:

```dart
extension E<T> on T {
  void foo(){}
}

void main() {
  null.foo();
}
```

is a runtime error, but post-NNBD will be accepted?

lrhn: Yes. If you *want* the pre-migration behavior, you have to migrate it to:
```dart
extension E<T extends Object> on T {
  void foo() {}
}
```
(which is also valid pre-migration and does the same thing),
and if you want the post-NNBD behavior prior to migration, you have to write;
```dart
extension E<T> on T? {
  void foo() {}
}
```
(which will still work post-migration, the `?` will just be redundant).

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,523 @@
...
+
+Dart's static extension methods are *scoped*. They only apply to code that the extension itself is accessible to. Being accessible means that using the extension's name must denote the extension. The extension is not in scope if another declaration with the same name shadows the extension, if the extension's library is not imported, if the library is imported and the extension is hidden, or the library is only imported with a prefix. In other words, if the extension had been a class, it is only in scope if using the name would denote the class&mdash; which will allow you to call static methods on the class, which is exactly what we are going to do.
```
leafpetersen: Is a private extension in a library's export scope?  If not, we should probably be sure to say so explicitly.  If so, does importing that library bring any public methods in that private extension into scope?

lrhn:
Landing so the text can be used as reference.

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,612 @@
...
+
+```dart
+extension<T> on List<T> {
+  void quadruple() { ... }
+}
+```
+
+This is equivalent to giving the extension a fresh private name.
+
+We may need to make `on` a built-in identifier, and not allow those as names of extensions, then there should not be any parsing issue. Even without that, the grammar should be unambiguous because `extension on on on { … }` and `extension on on { … }` are distinguishable, and the final type cannot be empty. It may be *harder* to parse, though.
```
leafpetersen: What's the plan to resolve this question?  Base it on experimentation in the parser?

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,612 @@
...
+...
+  List<int> x = ...;
+  var v = x.best();
+  List<num> y = ...;
+  var w = y.best();
+```
+
+Here all three extensions apply to both invocations.
+
+For `x.best()`, the most specific one is `BestList`. Because `List<int>` is a proper subtype of both ` iterable<int>` and `<List<num>`, we expect `BestList` to be the best implementation. The return type causes `v` to have type `int`. If we had chosen `BestSpec` instead, the return type could only be `num`, which is one of the reasons why we choose the most specific instantiated type as the winner.
```
leafpetersen: ` iterable<int>` -> `Iterable<int>`

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,612 @@
...
+  bool get isOdd => !isEven;
+  static bool isListEven(List<Object> list) => list.isEven;
+}
+```
+
+Here the `list.isEven` will find that `isEven` of `MyUnaryNumber` applies, and unless there are any other extensions in scope, it will call that. (Or unless someone adds an `isEven` member to `List`, but that's a breaking change, and then, if still necessary, this code can change the call to `MyUnaryNumber(list).isEven`.)
+
+The unqualified `length` of `isEven` is not defined in the current lexical scope, so is equivalent to  `this.length`, which is valid since `List<Object>` has a `length` getter.
+
+The unqualified `isEven` of `isOdd` resolves lexically to the `isEvent` getter above it, so it is equivalent to `MyUnaryNumber(this).isEven`,  even if there are other extensions in scope which define an `isEven` on `List<Object>`.
```
leafpetersen: `isEvent` -> `isEven`.  Also, maybe make explicit that even if someone added an isEven instance member to List, that the unqualified isEven reference would resolve to the extension method?

accepted/future-releases/static-extension-methods/design-document.md
```diff
@@ -0,0 +1,612 @@
...
+
+Invocations on `this` use the same extension method resolution as any other code. Most likely  the current extension will be the only one in scope which applies. It definitely applies to its own declared `on` type.
+
+Like for a class or mixin member declaration, the names of the extension members, both static and instance, are in the *lexical* scope of the extension member body. That is why `MySmart` above can invoke the static `smartHelper` without prefixing it by the extension name. In the same way, *instance* member declarations (the extension members) are in the lexical scope. 
+
+If an unqualified identifier lexically resolves to an extension method of the surrounding extension, then that identifier is not equivalent to `this.id`, rather the invocation is equivalent to an explicit invocation of that extension method on `this` (which we already know has a compatible type for the extension): `Ext<T1,…,Tn>(this).id`, where `Ext` is the surrounding extension and `T1` through `Tn` are its type parameters, if any. The invocation works whether or not the names of the extension or parameters are actually accessible, it is not a syntactic rewrite.
```
leafpetersen: A corner case here: if I write an extension on `dynamic`, I think this implies that internal unqualified references to the members that I am defining will resolve to the extension members and not as dynamic calls.  Worth calling out, and testing for?

PR #316, opened 2019-04-15
simplify syntax for library opt-out

danrubel:
This change simplifies the syntax for designating a library language version to make scanning and parsing more efficient. Rather than accepting a language version anywhere in any type of comment, this proposes that a language version can only be specified in a single line comment of the form
```
// @dart = 2.3
```
with nothing else in the comment.


lrhn:
I have integrated this change in a larger change. Please check https://github.com/dart-lang/language/pull/332

danrubel:
@lrhn Thanks!

PR #325, opened 2019-04-23
Clarify implementation step for validating codegen

natebosch:
We need to make sure that the build systems which use the analyzer to
resolve code work correctly with the new feature - if it requires adding
a flag when we construct the `AnalysisDriver` we need to know about it.

### DIFF
--- a/templates/implementation_meta_issue.txt
+++ b/templates/implementation_meta_issue.txt
@@ -23,9 +23,8 @@ The implementation plan for this issue is <link to implementation plan>.
 - [ ] Issue for implementation in Cider:
 - [ ] Issue for implementation in Dartfmt:
 - [ ] Issue for implementation in Dartdoc:
-- [ ] Issue for implementation in Angular compiler:
+- [ ] Issue for implementation in build systems and Angular compiler:
 - [ ] Issue for language tests:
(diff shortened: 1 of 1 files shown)


templates/implementation_meta_issue.txt
```diff
@@ -28,3 +28,4 @@ The implementation plan for this issue is <link to implementation plan>.
 - [ ] Issue for documentation:
 - [ ] Issue for specification:
 - [ ] Issue for code cleanup:
+- [ ] Issue for validating behavior of codegen:
```
munificent: This sounds pretty nebulous. Can we word it more concretely?

How about:

```
- [ ] Issue to validate that build package handles feature:
- [ ] Issue to validate that AngularDart template compiler handles feature:
```

?

leafpetersen: We have an existing entry for Angular above.  Should we just add package build as an additional "Implementation issue" to be filed above?

natebosch: They don't need to be separate. Updated pr to add "build system" wording next to Angular.

munificent:
LGTM. Leaf?

PR #326, opened 2019-04-23
Fix typo in assignableExpression grammar

leafpetersen:
I think this is what is intended?


munificent:
If this was a typo, it makes me feel a little better about my mistake in the CHANGELOG. :)

PR #332, opened 2019-04-26
Update versioning feature specification.

lrhn:
- Restrict override syntax to only `// @dart = 2.3` format.
- Specify experiment flag behavior.
- Update `.packages` file format to contain entire SDK version. Needed by experiment flag behavior.
- Change `*:defaultPackage` to `:defaultPackage` because `*` was already a valid package name in the [`.packages` file specification](<https://github.com/lrhn/dep-pkgspec/blob/master/DEP-pkgspec.md>).

### DIFF
--- a/accepted/future-releases/language-versioning/language-versioning.md
+++ b/accepted/future-releases/language-versioning/language-versioning.md
@@ -1,28 +1,28 @@
 # Dart Language Versioning
 
 [lrn@google.com](@lrhn)<br>
-Version: 1.0-rc2<br>
+Version: 1.1<br>
 Status: Preliminary Design, Under Discussion.
 
 ## Motivation
 
 The Dart language is evolving with new language features added in most new releases.
-However, since Dart 2.0, a breaking language change is unlikely to be practically possible.
-Any breaking the language will require a large (and growing) body of code to be migrated.
+Since Dart 2.0, a breaking language change is unlikely to be practically possible.
+Any breaking of the language requires a large (and growing) body of code to be migrated.
 
-We still want to do breaking changes, because some language changes are worth it. They just have to be worth _a lot_ to be viable. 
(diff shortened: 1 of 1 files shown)


danrubel:
Your changes look better. Thanks!

danrubel:
Closed the wrong issue. Reopening...

lrhn:
PTAL everybody.

accepted/future-releases/language-versioning/language-versioning.md
```diff
@@ -68,11 +67,17 @@ If a `pubspec.yaml` does not contain a minimum SDK version, the default language
...
 
-Examples:
+This does mean that if a later language version change the *comment syntax* in a non-backwards compatible way, we may have to add further restrictions at that point. That is unlikely to happen. 
+
+Part files may be marked, but are not required to be. They must always be treated the same way as the library they belong to, so it is a compile-time error if a part file has a different language level override than its library. Tools that work on individual part files, like the formatter, can use a marker in the part file, but if there isn't one, the tool should make an effort to find the library that the file belongs to, and use the version of that library. If the part uses a URI-based `part of` declaration, the library file should be accessible. If it uses a name-based `part of` declaration, the tool might not be able to find the library from the part file alone, and it might be prudent to add a language level override to the part file as well.
```
leafpetersen: This seems like the worst of both worlds.  Do we want tools to support this or not?  If all we can tell users is "tools may or may not support this", then they de facto have to write it, and yet at the same time we're asking tools to put effort into trying to support it sometimes?

lrhn: We can go with "part files must have the mark" too. I'm warming to the idea. After all, we don't mind making it uncomfortable to *not* migrate.

Let's do that.

leafpetersen: SGTM, thanks!

PR #377, opened 2019-05-29
Move nnbd roadmap to accepted.

leafpetersen:
Start with just the roadmap, I'll land the draft feature spec shortly.


mit-mit:
One comment: Maybe remove this part:

```
Alternatives to a sound system
TODO: discuss some additional options here:

Typescript/C# style heavy type system with unsound holes
Static only type system
Simple variable/parameter/return value only annotations
Dynamic checking
```

And maybe add a

PR #388, opened 2019-06-03
First cut at analyzer preferred subtyping algorithm

leafpetersen:
Just exploration right now.  The analyzer folks wanted to explore re-factoring the algorithmic subtyping to make it easier to share code between the legacy and new type systems, here's my attempt at working their suggestions into a version of the rules.

cc @MichaelRFairhurst @stereotype441

### DIFF
--- a/resources/type-system/subtyping.md
+++ b/resources/type-system/subtyping.md
@@ -242,6 +242,140 @@ that the choice of common variable names avoid capture.  It is valid to choose
 the `Xi` or the `Yi` for `Zi` so long as capture is avoided*
 
 
+## Algorithmic subtyping: version 2
+
+This version of the algorithmic rules re-orders the rules to move all of the
+rules relevant to non-nullability as early as possible.  It also simplifies the
+algorithm somewhat by allowing specific rules to "fall through" to later rules
+when it is clearer to do so.  This presentation is intended to be equivalent to
+the above.  It is motivated by pragmatic concerns from the implementations.
+
+### Rules
+
+We say that a type `T` is **nullable** if `Null <: T`.  This is equivalent to
+the syntactic criterion that `T` is any of:
+  - `Null`
+  - `S?` for some `S`
+  - `S*` for some `S` where `S` is nullable
+  - `FutureOr<S>` for some `S` where `S` is nullable
+  - `dynamic`
+  - `void`
+
+We say that a type `T` is a **top** type if `T` is `dynamic`, `void`, `Object?`,
+or `FutureOr<S>` where `S` is a top type.
+
+**Note that these rules are designed to be read top down as an algorithm**
+
+We say that a type `T0` is a subtype of a type `T1` (written `T0 <: T1`) when:
+
+- **Reflexivity**: if `T0` and `T1` are the same type then `T0 <: T1` otherwise continue
+  - *Note that this check is necessary as the base case for primitive types, and
+    type variables but not for composite types.  In particular, algorithmically
+    a structural equality check is admissible, but not required
+    here. Pragmatically, non-constant time identity checks here are
+    counter-productive*
+
+- **Right Top**: if `T1` is a top type then `T0 <: T1` otherwise continue
+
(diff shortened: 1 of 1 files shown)


resources/type-system/subtyping.md
```diff
@@ -242,6 +242,140 @@ that the choice of common variable names avoid capture.  It is valid to choose
...
+when it is clearer to do so.  This presentation is intended to be equivalent to
+the above.  It is motivated by pragmatic concerns from the implementations.
+
+### Rules
+
+We say that a type `T` is **nullable** if `Null <: T`.  This is equivalent to
+the syntactic criterion that `T` is any of:
+  - `Null`
+  - `S?` for some `S`
+  - `S*` for some `S` where `S` is nullable
```
MichaelRFairhurst: note that we don't have to cover this case ourselves in the analyzer; we definite `[S/S'?]S* = S'?`.

resources/type-system/subtyping.md
```diff
@@ -242,6 +242,140 @@ that the choice of common variable names avoid capture.  It is valid to choose
...
+  false)
+
+- **Left Legacy Top**: if `T0` is `S*` where `S` is a top type then the
+  subtyping does not hold (return false)
+
+- **Left Nullable 1**: if `T0` is `Null`, `Null*`, or `S0?` and `T1` is not
+  nullable then the subtyping does not hold (return false)
+
+- **Left Nullable 2**: otherwise, if `T0` is `S0?` then the subtyping holds iff
+  `S0 <: T1` (return true or false)
```
MichaelRFairhurst: I think ideally we wouldn't have to extract `S0` from `S0?`, because it requires a copy of `S0` which may be a moderately complex type like a function type.

That said, I think we are mostly worried about extracting `S0` from `S0*`, since legacy types will be much more common than nullable types for a long time. So unwrapping `S0?` may be totally doable.

resources/type-system/subtyping.md
```diff
@@ -242,6 +242,140 @@ that the choice of common variable names avoid capture.  It is valid to choose
...
+
+- **Left Bottom**: if `T0` is `Null`, `Null?`, `Null*`, `Never`, `Never?`, or
+  `Never*` then `T0 <: T1` (return true)
+  - Note that we rely here on the fact that if `T0` is `Null`, `Null?`, `Null*`
+    or `Never?` then we know by the fact that `Left Nullable 1` above did not
+    apply that `T1` is nullable.
+
+- **Left Legacy** if `T0` is `S0*` then:
+  - `T0 <: T1` iff `S0 <: T1` (return true or false)
```
MichaelRFairhurst: Yeah so this rule is a potential performance problem for us.

Do we need to redo all the prior steps if we don't allow representation of `T?*` etc?

resources/type-system/subtyping.md
```diff
@@ -242,6 +242,140 @@ that the choice of common variable names avoid capture.  It is valid to choose
...
+
+- **Right Legacy** `T1` is `S1*` then:
+  - `T0 <: T1` iff `T0 <: S1?` (return true or false)
+
+- **Right Nullable**: if `T1` is `S1?` then `T0 <: T1` if:
+  - `T0` is nullable
```
MichaelRFairhurst: I think maybe there's a missing negative here somewhere or something. Am I correct in reading that these rules currently say `FuureOr<int?> <: String?`?

MichaelRFairhurst (commented):
Thanks for getting on this so quickly Leaf! Definitely some differences between what we've come up with, which tells me that your approach of refactoring the existing rules gives us a really really important sense of confidence once we have the right rules.

resources/type-system/subtyping.md
fishythefish: This is equivalent to `Null <: T0`, but I think the correct condition is
```suggestion
  - `T0 <: Null`
```

munificent:
This looks to be quite old. Can we abandon it?

PR #390, opened 2019-06-06
Removed --enable-experiment flag for non-function type aliases

eernstg:
Cf. discussion by email 'Re: New DartK crash on 2 co19 tests': This CL removes the 'nonfunction-type-aliases' flag from the [implementation plan](https://github.com/dart-lang/language/blob/master/accepted/future-releases/generalized-typedef-2018/implementation-plan.md#release-flag).

The intention is that we use this issue to clarify whether we will have that flag. It is not currently implemented by the vm nor by the analyzer, but it is mentioned in some co19 tests, e.g., [this one](https://github.com/dart-lang/co19/blob/3755b2251b57dc216bc45c9a572896cad4f9152d/Language/Types/Type_Aliases/syntax_t01.dart#L17).

If we wish to have the flag then we just drop this CL and tools should then have it; if we wish to remove the flag then we land this CL.

### DIFF
--- a/accepted/future-releases/generalized-typedef-2018/implementation-plan.md
+++ b/accepted/future-releases/generalized-typedef-2018/implementation-plan.md
@@ -7,20 +7,10 @@ Relevant documents:
 ## Implementation and Release plan
 
 This feature is non-breaking, because it is concerned with the introduction of
-support for new syntactic forms.
-Still, we will introduce it using an
-[experiments flag](https://github.com/dart-lang/sdk/blob/master/docs/process/experimental-flags.md)
-in order to enable a controlled deployment.
-
+support for new syntactic forms. 
+Hence, there is no `--enable-experiment` flag for it.
 
 ### Phase 0 (Preliminaries)
-
-#### Release Flag
-
-The flag
-`--enable-experiment=nonfunction-type-aliases`
-must be passed for the changes to be enabled.
-In this phase, support for that flag is added to all tools.
 
 #### Tests
 


kmillikin:
You can have as many experimental flags as you want.  You just can't test them before they exist.

In general, I'd rather hold off on having tests of a feature until we start implementing it.  Before then, any behavior that those tests exhibit will just be accidental.  And tests that we don't (or can't) yet run have historically been full of bugs in the tests themselves.

eernstg:
> You just can't test them before they exist

This means that we _must_ have the flag for every feature which is planned, and on top of that we need to ensure that the flag is implemented before the tests are introduced.

That's the opposite of what we discussed in the email thread (which was having the feature, but no flag), and if we must have the flag then this CL should simply be abandoned.

> In general, I'd rather hold off on having tests of a feature until we start implementing it.

That makes sense, but the other side of the coin was that Sergey and Irina could use this topic as a flexible task (when this feature was postponed because other features had a higher priority): They could start writing tests because this feature was planned and well understood, and then the tests would be available whenever the implementation started.

Maybe it would suffice to add support for recognizing the flag as the first step, even though the implementations don't support the feature (fully, or at all)?

kmillikin:
Yes, you must have the enable-experiment flag for every experimental feature which is planned.  It's not really two separate decisions, it's just one: is this feature developed behind an experimental flag?  If yes, you have to have the flag.

And it's necessary to add support for recognizing the flag as a first step, before the implementations support the feature.  There's no practical way we can develop a feature behind a flag when the flag doesn't exist!

eernstg:
This was a miscommunication -- abandoning the pull request.

PR #403, opened 2019-06-14
Fixed two typo level latex errors

eernstg:
Just two typo fixes that prevent LaTeX'ing.


leafpetersen:
Landing this so we get a regenerated spec.

eernstg:
Thanks!

PR #413, opened 2019-06-20
Correct and clarify potentially non-nullable definition

leafpetersen:
Adding the constraint that to be considered nullable, a type `T` must not be a subtype of `Object` (in other words, not a * of a nullable type).

Also adding S* to the syntactic list of non-nullable types, as implied by the semantic definition.

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -129,9 +129,10 @@ We say that a type `T` is **potentially nullable** if `T` is not non-nullable.
 Note that this is different from saying that `T` is nullable.  For example, a
 type variable `X extends Object?` is a type which is potentially nullable but
 not nullable.  Note that `T*` is potentially nullable by this definition if `T`
-is potentially nullable.  The potentially nullable types include all of the
-types which are either definitely nullable, potentially instantiable to a
-nullable type, or which may be migrated to a nullable type.
+is potentially nullable - so `int*` is not potentially nullable, but `X*` where
+`X extends int?` is.  The potentially nullable types include all of the types
+which are either definitely nullable, potentially instantiable to a nullable
+type, or for which any migration results in a potentially nullable type.
 
 We say that a type `T` is **potentially non-nullable** if `T` is not nullable.
 Note that this is different from saying that `T` is non-nullable.  For example,
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -103,15 +103,23 @@ This is equivalent to the syntactic criterion that `T` is any of:
   - `dynamic`
   - `void`
 
+Nullable types are types which are definitively known to be nullable, regardless
+of instantiation of type variables, and regardless of any choice of replacement
+for the `*` positions (with `?` or nothing).
+
 We say that a type `T` is **non-nullable** if `T <: Object`.
 This is equivalent to the syntactic criterion that `T` is any of:
   - `Never`
   - Any function type (including `Function`)
-  - Any interface type except `dynamic`, `void`, and `Null`.
+  - Any interface type except `Null`.
   - `S*` for some `S` where `S` is non-nullable
   - `FutureOr<S>` where `S` is non-nullable
   - `X extends S` where `S` is non-nullable
   - `X & S` where `S` is non-nullable
+
+Non-nullable types are types which are either definitively known to be
+non-nullable regardless of instantiation of type variables, or for which
+replacing the `*` positions with nothing will result in a non-nullable type.
 
 Note that there are types which are neither nullable nor non-nullable.  For
 example `X extends T` where `T` is nullable is neither nullable nor
@@ -121,7 +129,9 @@ We say that a type `T` is **potentially nullable** if `T` is not non-nullable.
 Note that this is different from saying that `T` is nullable.  For example, a
 type variable `X extends Object?` is a type which is potentially nullable but
 not nullable.  Note that `T*` is potentially nullable by this definition if `T`
-is potentially nullable.
+is potentially nullable.  The potentially nullable types include all of the
+types which are either definitely nullable, potentially instantiable to a
+nullable type, or which may be migrated to a nullable type.
 
 We say that a type `T` is **potentially non-nullable** if `T` is not nullable.
 Note that this is different from saying that `T` is non-nullable.  For example,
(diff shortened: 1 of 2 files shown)


accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -94,20 +94,21 @@ library section below.
...
-We say that a type `T` is **non-nullable** if `T <: Object`.  This is equivalent
-to the syntactic criterion that `T` is any of:
-  - `Object`, `int`, `bool`, `Never`, `Function`
-  - Any function type
-  - Any class type or generic class type
+We say that a type `T` is **non-nullable** if `T <: Object`.
+This is equivalent to the syntactic criterion that `T` is any of:
+  - `Never`
+  - Any function type (including `Function`)
+  - Any interface type except `dynamic`, `void`, and `Null`.
```
lrhn: Nitpickery: I don't consider `Function` a function type, nor `dynamic` or `void` as interface types. :)

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -94,20 +94,21 @@ library section below.
...
   - `FutureOr<S>` for some `S` where `S` is nullable
   - `dynamic`
   - `void`
 
-We say that a type `T` is **non-nullable** if `T <: Object`.  This is equivalent
-to the syntactic criterion that `T` is any of:
-  - `Object`, `int`, `bool`, `Never`, `Function`
-  - Any function type
-  - Any class type or generic class type
+We say that a type `T` is **non-nullable** if `T <: Object`.
```
lrhn: Making `int*` definitely and potentially non-nullable.

That makes sense if being nullable is only used for restricting member access.

However, below it says:

> It is an error if a top level variable, static variable, or instance field with
potentially non-nullable type has no initializer expression and is not
initialized in a constructor via an initializing formal or an initializer list
entry, unless the variable or field is marked with the `late` modifier.

That would make `int x;` invalid in a non-migrated library because the type of the field is `int*` which is not nullable, and therefore potentially non-nullable.

I think we migth *need* the fourth category of types for `*` types: They are both definitely nullable and definitely non-nullable (and not potentially either).

If we use definite nullable/non-nullable to enable features, and potentially nullable/non-nullable to detect errors them, then that would allow `*` types to work everywhere, without error triggering an error.
(It's going to be tricky to get right, though).

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -94,20 +94,21 @@ library section below.
 
 ### Static errors
 
-We say that a type `T` is **nullable** if `Null <: T`.  This is equivalent to
-the syntactic criterion that `T` is any of:
+We say that a type `T` is **nullable** if `Null <: T` and not `T <: Object`.
```
lrhn: So, to be sure, the only types affected by this change are *T** where *T* <: `Object`.

Any other type which is <: `Object` is also not a supertype of `Null`.


This definition of nullable and non-nullable splits types into *three* groups, not four like before, where it was possible to be both nullable and non-nullable (and not potentially either).

1. non-nullable and potentially non-nullable (not nullable), T <: Object.
1. potentially nullable (not nullable) and potentially non-nullable (not nullable), not T<: Object, not Null <: T
1. nullable and potentially non-nullable (not non-nullable), not T <: Object, Null <: T.

So: to definitely (and potentially) nullable, definitely (and potentially) non-nullable, or potentially either.

That does seem to make sense.

lrhn:
Just noticed
> For the purposes of errors and warnings, the null aware operators `?.` and `?..` are checked as if the receiver of the operator had non-nullable type.

Does it have to say *which* non-nullable type? 
In particular, for `FutureOr<int?> x; x?.then(...);`, is that allowed? And if so, at which type? (Or is that just disallowed because `FutureOr` doesn't allow member invocations not on `Object` either?)

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: Is Null an interface type?

leafpetersen:
> Just noticed
> 
> > For the purposes of errors and warnings, the null aware operators `?.` and `?..` are checked as if the receiver of the operator had non-nullable type.
> 
> Does it have to say _which_ non-nullable type?

Coincidentally I had just noticed this as well.  I think we should specify it as **NonNull(T)** where T is the static type of the receiver.  I think this matches the behavior you'd get from promotion in the desugared version.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: We don't rely on * types for error suppression in legacy libraries - it's simply the case that none of these new errors apply.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: I could split this out further into:
  - definitely nullable (Null <: T and not T <: Object)
  - definitely non-nullable (T <: Object, not Null <: T)
  - assumed non-nullable (T <: Object)
     - the non-nullable types, plus T* where T is assumed non-nullable
  - potentially nullable (not T <: Object or Null <: T)
  - potentially non-nullable (not Null <: T or T <: Object)

I'm not sure this is useful though, so instead I'm just collapsing assumed non-nullable and non-nullable.  Does that make sense?

accepted/future-releases/nnbd/feature-specification.md
lrhn: OK, so the quoted text does not apply to legacy libraries at all. Then it may work. I'll have to think about it.

accepted/future-releases/nnbd/feature-specification.md
lrhn: So far it has been. It behaves just as any other (sealed) interface type, so there hasn't been any need to special-case it.

accepted/future-releases/nnbd/feature-specification.md
lrhn: If we don't need the extra cases, then let's not introduce them.
As long as this works, I'm fine with it.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: Well, it is special cased in the type system - you can't get the right subtyping otherwise.  That's why I get confused... :)

leafpetersen:
Landing this.  cc @scheglov @lrhn @stereotype441 Let me know if there is more discussion to follow up on here.

PR #420, opened 2019-06-25
Elaborate on the `ExtensionName(object)` override notation.

lrhn:
Made it explicit that `ExtensioName(object)` is a valid expression. It evaluates to the value of `object` and has the same static type, but if used for member lookup, it instead does lookup on the extension members.
It works in cascades: `var x = ExtensionName(object)..method1()..method2();` should invoke extension methods, then evaluate to `object`.

Also stated that `this` can be promoted by type promotion, which is useful, but not essential. 
Can we promote `this` in instance methods? If not, maybe we should just drop it here too.

### DIFF
--- a/accepted/future-releases/static-extension-methods/feature-specification.md
+++ b/accepted/future-releases/static-extension-methods/feature-specification.md
@@ -1,6 +1,6 @@
 # Dart Static Extension Methods Design
 
-lrn@google.com<br>Version: 1.0<br>Status: Design Proposal
+lrn@google.com<br>Version: 1.1<br>Status: Design Proposal
 
 This is a design document for *static extension methods* for Dart. This document describes the most basic variant of the feature, then lists a few possible variants or extensions.
 
@@ -28,7 +28,7 @@ For these discoverability and readability reasons, static extension methods will
 
 The extension methods are *static*, which means that we use the static type of an expression to figure out which method to call, and that also means that static extension methods are not virtual methods.
 
-The methods we allow you to add this way are normal methods, operators, getter and setters.
+The methods we allow you to add this way are normal methods, operators, getter and setters. As such, the feature should really be called "Static Extension Members". For historical reasons, we will stick with the "Static Extension Methods" name.
 
 ## Declaring Static Extension Methods
 
@@ -315,11 +315,19 @@ or if you imported the extension with a prefix to avoid name collision:
 prefix.MyList<String>(object).quickSort();
 ```
 
-The syntax looks like a constructor invocation, but it does not create a new object.
+The syntax looks like a constructor invocation, but it does not create a new object. Instead it evaluates `object` to a value, then calls the `quickSort` extension method of `MyList` on that object.
 
 If `object.quickSort()` would invoke an extension method of `MyList`, then `MyList(object).quickSort()` will invoke the exact same method in the same way.
 
 The syntax is not *convenient*&mdash;you have to put the "constructor" invocation up front, which removes the one advantage that extension methods have over normal static methods. It is not intended as the common use-case, but as an escape hatch out of unresolvable conflicts.
+
+If the expression `MyList(object)` is evaluated for its value, it evaluates to the value of  `object`. It only works as a scope override for member lookup, similarly to how a class name works differently for static member lookup and evaluation to a `Type` object.
+
+A cascade expression like `var o = MyList(expr)..quickSort()..quickSort()` works correctly, the static members of `MyList` are invoked by the cascade, and then `o` is bound to the value of `expr`. As such, the static type of `MyList(object)` is the static type of `object`, it is just that the static type is not what is used for member lookup.
+
+_It is an valid alternative to entirely disallow evaluating `MyList(object)` for its value, but that would preclude, or at least make it more complicated, to allow cascades on extensions._
+
+You can write `Extension1(Extension2(object))`. That will evaluate `Extension2(Object)` for its value and then allow lookups on `Extension1` only.
 
 ### Static Members and Member Resolution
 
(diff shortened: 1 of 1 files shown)


leafpetersen:
> Can we promote `this` in instance methods? If not, maybe we should just drop it here too.

https://github.com/dart-lang/sdk/issues/32120

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -315,12 +315,20 @@ or if you imported the extension with a prefix to avoid name collision:
 prefix.MyList<String>(object).quickSort();
 ```
 
-The syntax looks like a constructor invocation, but it does not create a new object.
+The syntax looks like a constructor invocation, but it does not create a new object. Instead it evaluates `object` to a value, then calls the `quickSort` extension method of `MyList` on that object.
```
leafpetersen: Consider making it explicit that `MyList(o).foo` is an error if `foo` is not an extension member of `MyList`, even if `foo` *is* an instance member on `o`?

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -356,7 +364,7 @@ During NNBD migration, we will have unsound nullable types like `int*` which sho
 
 When executing an extension instance member, we stated earlier that the member is invoked with the original receiver as `this` object. We still have to describe how that works, and what the lexical scope is for those members.
 
-Inside an extension method body, `this` does not refer to an instance of a surrounding type. Instead it is bound to the original receiver, and the static type of `this` is the declared `on` type of the surrounding extension (which may contain unbound type variables).
+Inside an extension method body, `this` does not refer to an instance of a surrounding type. Instead it is bound to the original receiver, and the static type of `this` is the declared `on` type of the surrounding extension (which may contain unbound type variables). The type of `this` can be promoted by type promotion (`if (this is String) …`)as if it was a final variable.
```
leafpetersen: We don't do this in general for `this`.  We could, and it might be a good idea.

accepted/future-releases/static-extension-methods/feature-specification.md
lrhn: Will do.

accepted/future-releases/static-extension-methods/feature-specification.md
lrhn: Let's drop here, and maybe do it in full generality at some point.

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -315,12 +315,20 @@ or if you imported the extension with a prefix to avoid name collision:
...
 
 If `object.quickSort()` would invoke an extension method of `MyList`, then `MyList(object).quickSort()` will invoke the exact same method in the same way.
 
 The syntax is not *convenient*&mdash;you have to put the "constructor" invocation up front, which removes the one advantage that extension methods have over normal static methods. It is not intended as the common use-case, but as an escape hatch out of unresolvable conflicts.
 
+If the expression `MyList(object)` is evaluated for its value, it evaluates to the value of  `object`. It only works as a scope override for member lookup, similarly to how a class name works differently for static member lookup and evaluation to a `Type` object.
+
+A cascade expression like `var o = MyList(expr)..quickSort()..quickSort()` works correctly, the static members of `MyList` are invoked by the cascade, and then `o` is bound to the value of `expr`. As such, the static type of `MyList(object)` is the static type of `object`, it is just that the static type is not what is used for member lookup.
```
lrhn: I'm actually not happy with the static type here.

If, hypothetically, we want to use the current extension declarations as extension types as well, then `MyList(expr)` should have `MyList` as static type at that point, allowing only the extension members to be used. If we give it type `List` now, then someone might start doing `(MyList(expr)..extensionMember()).listMember()`, and that would *stop* working if `MyList(expr)` changes static type.

So, for now I'll recommend that the static type of `MyList(expr)` as a value expression is `void`. You can still do `MyList(expr).foo()` or `MyList(expr)..foo()..bar()` because lookups won't use the static type (just like `ClassName.foo` does static member lookup, but the static type of `ClassName` by itself is `Type`).

We should discourage `MyList(expr)` as a value expression anywhere other than in a cascade because it's just equivalent to `expr as void`, but the `void` type would prevent users from using the value of the cascade for now, until we perhaps get extension types.

lrhn:
Changed static type of `ExtensionName(object)` to `void` to preserve the option of making `ExtensionName` an extension *type* later.

PTAL

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -554,7 +566,7 @@ Since it's possible to add extensions on superclass (including `Object`), it wou
 
 - Otherwise, the single most-specific extension's member is invoked with the extension's type parameters bound to the types found by inference, and with `this ` bound to the receiver.
 
-- An extension method can be invoked explicitly using the syntax `ExtensionName(object).method(args)`. Type arguments can be applied to the extension explicitly as well, `MyList<String>(listOfString).quickSort()`. Such an invocation overrides all extension resolution. It is a compile-time error if `ExtensionName` would not apply to the `object.method(args)` invocation if it was in scope. 
+- An extension method can be invoked explicitly using the syntax `ExtensionName(object).method(args)`. Type arguments can be applied to the extension explicitly as well, `MyList<String>(listOfString).quickSort()`. Such an invocation overrides all extension resolution. It is a compile-time error if `ExtensionName` would not apply to the `object.method(args)` invocation if it was in scope. It is a compile-time error if the extension `ExtensionName` doesn't declare a member named `method`.
```
leafpetersen: I don't know what the sentence starting with "It is a compile-time error if `ExtensionName`" is supposed to mean, consider rephrasing?

Also, it's not clear to me where we specify that `ExtensionName(object)` is an error if the resolved `on` type for `ExtensionName` is not assignable to the static type of `object`, which I presume we want?

accepted/future-releases/static-extension-methods/feature-specification.md
leafpetersen: Not sure how I feel about this.  It's definitely not what I would do ignoring the extension type possibility, and does take away a useful affordance.  It will also be quite surprising to users, I think, if they encounter it.  On the other hand, it's a pretty corner case thing.

Do you really expect to want to piggy back extension types on this?  It feels unlikely to work out well in general to me, but maybe.

leafpetersen:
Is this abandoned?

leafpetersen:
Ah, not abandoned, I think, but the description text is out of date.

lrhn:
I actually think it can be abandoned. Most of the content has been superseded by other updates, and extracting the rest is not worth working on this CL for. I'll just see if something is still useful and add it to another CL.

PR #427, opened 2019-06-26
Specify that torn-off extension methods are never equal.

lrhn:


leafpetersen:
> This seems too strong as it prevents GVN, or is that permitted somewhere else?

We could allow but not require identity.  I think so far the Dart spec has tried to make identity and dis-identity deterministic, but in general I've been in favor of not specifying dis-identity.

yjbanov:
I think this semantic will break down when we start talking about [immutables](https://github.com/dart-lang/language/blob/master/working/0125-static-immutability/feature-specification.md) (or value types, or data classes), where it is highly desirable to have the language provide built-in hash code and equality. Consider:

```dart
typedef DoMath = int Function(int);

data class Arithmetic {
  final DoMath mathDoer;
}

data class Adder {
  final int n;

  int addTo(int x) => x + n;
}

main() {
  var twoAdder = Adder(n: 2);
  var arithmetic1 = Arithmetic(mathDoer: twoAdder.addTo);
  var arithmetic2 = Arithmetic(mathDoer: twoAdder.addTo);

  // It would be weird for the following assertion to fail.
  assert(arithmetic1 == arithmetic2);
}
```

lrhn:
The text is actually a little too vague for my taste because it merely says that equality is identity. It does not say that a tear-off of an extension method creates a new object each time it is evaluated. I think it *should* say that too. We do not want some platforms to treat some tear-off as identical, and another to not do that. So, I'll add that too.

Each tear-off evaluation creates a new function object. Those function objects are only equal to themselves.

PR #434, opened 2019-06-28
Lower and upper bounds adjusted for NNBD

leafpetersen:
This takes the specification of lower and upper bounds as implemented and adds in:
  - `Never`
  - Nullable types `T?`
  - Legacy types `T*`
  - Required named parameters
  
It does not try to fix up any of the known issues, but rather to add the minimal delta to account for NNBD.

cc @scheglov @stereotype441 @stefantsov @lrhn @eernstg @munificent

### DIFF
--- a/resources/type-system/upper-lower-bounds.md
+++ b/resources/type-system/upper-lower-bounds.md
@@ -2,7 +2,11 @@
 
 leafp@google.com
 
-This documents the currently implemented upper and lower bound computation.
+This documents the currently implemented upper and lower bound computation,
+modified to account for explicit nullability and the accompanying type system
+changes (including the legacy types).  In the interest of backwards
+compatibility, it does not try to fix the various issues with the existing
+algorithm.  
 
 ## Types
 
@@ -23,13 +27,17 @@ top types.
 - **TOP**(`FutureOr<T>`) is **TOP**(T)
 - **TOP**(T) is false otherwise
 
+The **OBJECT** predicate is true for any type which is in the equivalence class
+of `Object`.
+
+
 - **OBJECT**(`Object`) is true
 - **OBJECT**(`dynamic`) is true
 - **OBJECT**(`void`) is true
 - **OBJECT**(`FutureOr<T>`) is **OBJECT**(T)
 - **OBJECT**(T) is false otherwise
 
-The **BOTTOM** predicate is true for things in the equivalence class of bottom.
+The **BOTTOM** predicate is true for things in the equivalence class of `Never`.
 
 - **BOTTOM**(`Never`) is true
 - **BOTTOM**(`X&T`) is true iff **BOTTOM**(`T`)
@@ -79,7 +87,7 @@ The **MOREBOTTOM** predicate defines a total order on bottom and `Null` types.
 
 ## Upper bounds
 
-We define the upper bound of two types T1 and T2 to be **UP**(T1,T2) as follows.
+We define the upper bound of two types T1 and T2 to be **UP**(`T1`,`T2`) as follows.
 
 
 - **UP**(`T`, `T`) = `T`
--- a/resources/type-system/upper-lower-bounds.md
+++ b/resources/type-system/upper-lower-bounds.md
@@ -16,20 +16,34 @@ document
 The **TOP** predicate is true for any type which is in the equivalence class of
 top types.
 
-- **TOP**(`Object`) is true
+- **TOP**(`T?`) is true iff **TOP**(`T`) or **OBJECT**(`T`)
+- **TOP**(`T*`) is true iff **TOP**(`T`) or **OBJECT**(`T`)
 - **TOP**(`dynamic`) is true
 - **TOP**(`void`) is true
 - **TOP**(`FutureOr<T>`) is **TOP**(T)
 - **TOP**(T) is false otherwise
 
-The **BOTTOM** predicate is true for either of the two bottom types.
+- **OBJECT**(`Object`) is true
+- **OBJECT**(`dynamic`) is true
+- **OBJECT**(`void`) is true
+- **OBJECT**(`FutureOr<T>`) is **OBJECT**(T)
+- **OBJECT**(T) is false otherwise
 
-- **BOTTOM**(`Null`) is true
-- **BOTTOM**(`bottom`) is true
+The **BOTTOM** predicate is true for things in the equivalence class of bottom.
+
+- **BOTTOM**(`Never`) is true
+- **BOTTOM**(`X&T`) is true iff **BOTTOM**(`T`)
+- **BOTTOM**(`X extends T`) is true iff **BOTTOM**(`T`)
 - **BOTTOM**(`T`) is false otherwise
 
-The **MORETOP** predicate defines a total order on the equivalence class of top
-types.
+The **NULL** predicate is true for things in the equivalence class of `Null`
+
+- **NULL**(`Null`) is true
+- **NULL**(`T?`) is true iff **NULL**(`T`) or **BOTTOM**(`T`)
+- **NULL**(`T*`) is true iff **NULL**(`T`) or **BOTTOM**(`T`)
+- **NULL**(`T`) is false otherwise
+
+The **MORETOP** predicate defines a total order on top and `Object` types.
 
 - **MORETOP**(`void`, `T`) = true
 - **MORETOP**(`T`, `void`) = false
@@ -37,7 +51,31 @@ types.
 - **MORETOP**(`T`, `dynamic`) = false
 - **MORETOP**(`Object`, `T`) = true
 - **MORETOP**(`T`, `Object`) = false
+- **MORETOP**(`T?`, `S?`) = **MORETOP**(`T`, `S`)
+- **MORETOP**(`T?`, `S`) = true
+- **MORETOP**(`T`, `S?`) = false
+- **MORETOP**(`T*`, `S*`) = **MORETOP**(`T`, `S`)
+- **MORETOP**(`T*`, `S`) = true
+- **MORETOP**(`T`, `S*`) = false
 - **MORETOP**(`FutureOr<T>`, `FutureOr<S>`) = **MORETOP**(T, S)
+
+The **MOREBOTTOM** predicate defines a total order on bottom and `Null` types.
+
+- **MOREBOTTOM**(`Never`, `T`) = true
+- **MOREBOTTOM**(`T`, `Never`) = false
+- **MOREBOTTOM**(`Null`, `T`) = true
+- **MOREBOTTOM**(`T`, `Null`) = false
+- **MOREBOTTOM**(`T?`, `S?`) = **MOREBOTTOM**(`T`, `S`)
+- **MOREBOTTOM**(`T?`, `S`) = true
+- **MOREBOTTOM**(`T`, `S?`) = false
+- **MOREBOTTOM**(`T*`, `S*`) = **MOREBOTTOM**(`T`, `S`)
+- **MOREBOTTOM**(`T*`, `S`) = true
+- **MOREBOTTOM**(`T`, `S*`) = false
+- **MOREBOTTOM**(`X&T`, `X&S`) = **MOREBOTTOM**(`T`, `S`)
+- **MOREBOTTOM**(`X&T`, `S`) = true
+- **MOREBOTTOM**(`S`, `X&T`) = false
+- **MOREBOTTOM**(`X extends T`, `X extends S`) = **MOREBOTTOM**(`T`, `S`)
+
 
 ## Upper bounds
 
@@ -45,16 +83,51 @@ We define the upper bound of two types T1 and T2 to be **UP**(T1,T2) as follows.
 
 
 - **UP**(`T`, `T`) = `T`
-- **UP**(`T1`, `T2`) = `T1` if:
-  - **TOP**(`T1`) and **TOP**(`T2`)
-  - and  **MORETOP**(`T1`, `T2`)
-- **UP**(`T1`, `T2`) = `T2` if:
-  - **TOP**(`T1`) and **TOP**(`T2`)
-  - and  not **MORETOP**(`T1`, `T2`)
+- **UP**(`T1`, `T2`) where **TOP**(`T1`) and **TOP**(`T2`) = 
+  - `T1` if **MORETOP**(`T1`, `T2`)
+  - `T2` otherwise
 - **UP**(`T1`, `T2`) = `T1` if **TOP**(`T1`)
-- **UP**(`T1`, `T2`) = `T1` if **BOTTOM**(`T2`)
 - **UP**(`T1`, `T2`) = `T2` if **TOP**(`T2`)
-- **UP**(`T1`, `T2`) = `T2` if **BOTTOM**(`T1`)
+
+- **UP**(`T1`, `T2`) where **BOTTOM**(`T1`) and **BOTTOM**(`T2`) = 
+  - `T1` if **MOREBOTTOM**(`T1`, `T2`)
+  - `T2` otherwise
+- **UP**(`T1`, `T2`) = `T1` if **BOTTOM**(`T1`)
+- **UP**(`T1`, `T2`) = `T2` if **BOTTOM**(`T2`)
+
+- **UP**(`T1`, `T2`) where **NULL**(`T1`) and **NULL**(`T2`) = 
+  - `T1` **MOREBOTTOM**(`T1`, `T2`)
+  - `T2` otherwise
+
+- **UP**(`T1`, `T2`) where **NULL**(`T1`) = 
+  - `T2` if  `T2` is nullable
+  - `T2?` otherwise
+
+- **UP**(`T1`, `T2`) where **NULL**(`T2`) = 
+  - `T1` if  `T1` is nullable
+  - `T1?` otherwise
+
+- **UP**(`T1`, `T2`) where **OBJECT**(`T1`) and **OBJECT**(`T2`) =
+  - `T1` if **MORETOP**(`T1`, `T2`)
+  - `T2` otherwise
+
+- **UP**(`T1`, `T2`) where **OBJECT**(`T1`) = 
+  - `T1` if `T2` is non-nullable
+  - `T1?` otherwise
+
+- **UP**(`T1`, `T2`) where **OBJECT**(`T2`) = 
+  - `T2` if `T1` is non-nullable
+  - `T2?` otherwise
+
+- **UP**(`T1*`, `T2*`) = `S*` where `S` is **UP**(`T1`, `T2`)
+- **UP**(`T1*`, `T2?`) = `S?` where `S` is **UP**(`T1`, `T2`)
+- **UP**(`T1?`, `T2*`) = `S?` where `S` is **UP**(`T1`, `T2`)
+- **UP**(`T1*`, `T2`) = `S*` where `S` is **UP**(`T1`, `T2`)
+- **UP**(`T1`, `T2*`) = `S*` where `S` is **UP**(`T1`, `T2`)
+
+- **UP**(`T1?`, `T2?`) = `S?` where `S` is **UP**(`T1`, `T2`)
+- **UP**(`T1?`, `T2`) = `S?` where `S` is **UP**(`T1`, `T2`)
+- **UP**(`T1`, `T2?`) = `S?` where `S` is **UP**(`T1`, `T2`)
 
 - **UP**(`X1 extends B1`, `T2`) = 
   - `T2` if `X1 <: T2`
@@ -98,10 +171,13 @@ We define the upper bound of two types T1 and T2 to be **UP**(T1,T2) as follows.
      - `R0` is **UP**(`T0`, `T1`)
      - `B2i` is `B0i`
      - `P2i` is **DOWN**(`P0i`, `P1i`)
+     - `Named0` contains `R0i xi` if `R1i xi` is a required named parameter in `Named1`
+     - `Named1` contains `R1i xi` if `R0i xi` is a required named parameter in `Named0`
      - `Named2` contains exactly `R2i xi` for each `xi` in both `Named0` and `Named1` 
         - where `R0i xi` is in `Named0`
         - where `R1i xi` is in `Named1`
         - and `R2i` is **DOWN**(`R0i`, `R1i`)
+        - and `R2i xi` is required if `xi` is required in either `Named0` or `Named1`
 
 - **UP**(`T Function<...>(...)`, `S Function<...>(...)`) = `Function` otherwise
 - **UP**(`T1`, `T2`) = `T2` if `T1` <: `T2`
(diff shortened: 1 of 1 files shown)


resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+- **NULL**(`T*`) is true iff **NULL**(`T`) or **BOTTOM**(`T`)
+- **NULL**(`T`) is false otherwise
+
+The **MORETOP** predicate defines a total order on top and `Object` types.
+
+- **MORETOP**(`void`, `T`) = true
+- **MORETOP**(`T`, `void`) = false
+- **MORETOP**(`dynamic`, `T`) = true
+- **MORETOP**(`T`, `dynamic`) = false
+- **MORETOP**(`Object`, `T`) = true
```
eernstg: I do not understand why this would be `Object` rather than `Object?`. With `Object` we get things like `MORETOP(Object, Object?)`, and that definitely seems wrong.

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+- **MORETOP**(`T`, `Object`) = false
+- **MORETOP**(`T?`, `S?`) = **MORETOP**(`T`, `S`)
+- **MORETOP**(`T?`, `S`) = true
```
eernstg: So `MORETOP(void?^{k+1}, void?^k)` for all `k`, and `MORETOP(void?, void*^m)` for all `m`. Looks like we have the "natural" relationship except that `void` is above all "modified voids". Perhaps quirky, but I assume this is benign.

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+- **MORETOP**(`FutureOr<T>`, `FutureOr<S>`) = **MORETOP**(T, S)
+
+The **MOREBOTTOM** predicate defines a total order on bottom and `Null` types.
+
+- **MOREBOTTOM**(`Never`, `T`) = true
+- **MOREBOTTOM**(`T`, `Never`) = false
+- **MOREBOTTOM**(`Null`, `T`) = true
+- **MOREBOTTOM**(`T`, `Null`) = false
+- **MOREBOTTOM**(`T?`, `S?`) = **MOREBOTTOM**(`T`, `S`)
+- **MOREBOTTOM**(`T?`, `S`) = true
```
eernstg: Typo?
```suggestion
- **MOREBOTTOM**(`T`, `S?`) = true
```

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+- **MOREBOTTOM**(`T`, `S?`) = false
```
eernstg:
```suggestion
- **MOREBOTTOM**(`T?`, `S`) = false
```

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+- **MOREBOTTOM**(`T*`, `S*`) = **MOREBOTTOM**(`T`, `S`)
+- **MOREBOTTOM**(`T*`, `S`) = true
```
eernstg:
```suggestion
- **MOREBOTTOM**(`T`, `S*`) = true
```

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+- **MOREBOTTOM**(`T`, `S*`) = false
```
eernstg:
```suggestion
- **MOREBOTTOM**(`T*`, `S`) = false
```

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+- **MOREBOTTOM**(`X&T`, `X&S`) = **MOREBOTTOM**(`T`, `S`)
+- **MOREBOTTOM**(`X&T`, `S`) = true
```
eernstg: This is surprising; we'd have `MOREBOTTOM(X & Null, Y extends Never)` for any X and Y, which seems to allow us to pass `null` as an argument where the parameter type is actually `Never`.

Maybe we'll never have any arguments of type `Never`, but it looks like a source of unsoundness in principle.

Line 84 admits a symmetric example with the same issue.

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+
+- **UP**(`T`, `T`) = `T`
+- **UP**(`T1`, `T2`) where **TOP**(`T1`) and **TOP**(`T2`) = 
+  - `T1` if **MORETOP**(`T1`, `T2`)
+  - `T2` otherwise
+- **UP**(`T1`, `T2`) = `T1` if **TOP**(`T1`)
+- **UP**(`T1`, `T2`) = `T2` if **TOP**(`T2`)
+
+- **UP**(`T1`, `T2`) where **BOTTOM**(`T1`) and **BOTTOM**(`T2`) = 
+  - `T1` if **MOREBOTTOM**(`T1`, `T2`)
```
eernstg:
```suggestion
  - `T2` if **MOREBOTTOM**(`T1`, `T2`)
```

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+  - `T2` otherwise
```
eernstg:
```suggestion
  - `T1` otherwise
```

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+- **UP**(`T1`, `T2`) = `T1` if **BOTTOM**(`T1`)
```
eernstg:
```suggestion
- **UP**(`T1`, `T2`) = `T2` if **BOTTOM**(`T1`)
```

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+- **UP**(`T1`, `T2`) = `T2` if **BOTTOM**(`T2`)
```
eernstg:
```suggestion
- **UP**(`T1`, `T2`) = `T1` if **BOTTOM**(`T2`)
```

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+  - `T1` if **MORETOP**(`T1`, `T2`)
+  - `T2` otherwise
+
+- **UP**(`T1`, `T2`) where **OBJECT**(`T1`) = 
+  - `T1` if `T2` is non-nullable
+  - `T1?` otherwise
+
+- **UP**(`T1`, `T2`) where **OBJECT**(`T2`) = 
+  - `T2` if `T1` is non-nullable
+  - `T2?` otherwise
```
eernstg: Explanation related to comment on `OBJECT`: For lines 118-128, we know that `not TOP(T1)` and `not TOP(T2)`, which was the reason why I suggested not to include `dynamic` and `void` in `OBJECT`.

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+
+- **DOWN**(`T1`, `T2`) where **BOTTOM**(`T1`) and **BOTTOM**(`T2`) =
+  - `T1` if **MOREBOTTOM**(`T1`, `T2`)
+  - `T2` otherwise
+- **DOWN**(`T1`, `T2`) = `T2` if **BOTTOM**(`T2`)
+- **DOWN**(`T1`, `T2`) = `T1` if **BOTTOM**(`T1`)
+
+
+- **DOWN**(`T1`, `T2`) where **NULL**(`T1`) and **NULL**(`T2`) =
+  - `T1` **MOREBOTTOM**(`T1`, `T2`)
```
eernstg:
```suggestion
  - `T1` if **MOREBOTTOM**(`T1`, `T2`)
```

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+  - `T2` otherwise
+
+- **DOWN**(`T1`, `T2`) where **NULL**(`T1`) =
+  - `T1` if  `T2` is nullable
+  - `Never` otherwise
```
eernstg: It's surprising that `DOWN(Null, int*)` is `Never`. Is that useful?

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+
+- **DOWN**(`T1`, `T2`) where **NULL**(`T2`) =
+  - `T2` if  `T1` is nullable
+  - `Never` otherwise
```
eernstg: Symmetric surprise: `DOWN(int*, Null)` is `Never`.

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+
+- **DOWN**(`T1`, `T2`) where **OBJECT**(`T1`) and **OBJECT**(`T2`) =
+  - `T1` if **MORETOP**(`T2`, `T1`)
+  - `T2` otherwise
+
+- **DOWN**(`T1`, `T2`) where **OBJECT**(`T1`) =
+  - `T2` if `T2` is non-nullable
+  - `Never` otherwise
```
eernstg: Surprised again: `DOWN(Object, int?)` is `Never`? This time I would definitely expect `int`. Symmetric example exists for line 241.

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+
+- **DOWN**(`T1`, `T2`) where **OBJECT**(`T2`) =
+  - `T2` if `T1` is non-nullable
+  - `Never` otherwise
+
+- **DOWN**(`T1*`, `T2*`) = `S*` where `S` is **DOWN**(`T1`, `T2`)
+- **DOWN**(`T1*`, `T2?`) = `S*` where `S` is **DOWN**(`T1`, `T2`)
+- **DOWN**(`T1?`, `T2*`) = `S*` where `S` is **DOWN**(`T1`, `T2`)
+- **DOWN**(`T1*`, `T2`) = `S` where `S` is **DOWN**(`T1`, `T2`)
+- **DOWN**(`T1`, `T2*`) = `S` where `S` is **DOWN**(`T1`, `T2`)
```
eernstg: Line 246, 247 forget the `*`. We could have `DOWN(int*, int?)` and get `int`, but it seems likely that we would want `int*`. WDYT?

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+- **MOREBOTTOM**(`T?`, `S?`) = **MOREBOTTOM**(`T`, `S`)
+- **MOREBOTTOM**(`T?`, `S`) = true
+- **MOREBOTTOM**(`T`, `S?`) = false
+- **MOREBOTTOM**(`T*`, `S*`) = **MOREBOTTOM**(`T`, `S`)
+- **MOREBOTTOM**(`T*`, `S`) = true
+- **MOREBOTTOM**(`T`, `S*`) = false
+- **MOREBOTTOM**(`X&T`, `X&S`) = **MOREBOTTOM**(`T`, `S`)
+- **MOREBOTTOM**(`X&T`, `S`) = true
+- **MOREBOTTOM**(`S`, `X&T`) = false
+- **MOREBOTTOM**(`X extends T`, `X extends S`) = **MOREBOTTOM**(`T`, `S`)
```
eernstg: We would need a rule where the two type variables differ (otherwise they'd presumably have the same bound as well).

Similarly, if we can't match different type variables in the `X&T` case then some top types will presumably just be unrelated (and then we don't have a total order).

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
+# Dart 2.0 Upper and Lower bounds
```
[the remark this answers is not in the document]
leafpetersen: Done.

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+changes (including the legacy types).  In the interest of backwards
+compatibility, it does not try to fix the various issues with the existing
+algorithm.  
+
+## Types
+
+The syntactic set of types used in this draft are a slight simplification of
+full Dart types, as described in the subtyping
+document
+[here](https://github.com/dart-lang/language/blob/master/resources/type-system/subtyping.md)
```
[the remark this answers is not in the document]
leafpetersen: > That document does not mention multiple modifiers, like `T??` or `T?*`, but they were hinted at yesterday, so I'm not sure about their status.

These are handled inductively - for a type `T?` `T` itself may always be of the form `S?` for some `S`.

> Another thing is that basically all rulesets below are defined in such a way that they _must_ assume prioritization: Every non-first rule starts with 'otherwise'. I think this should be mentioned.

Fixed by merge from the non NNBD version, let me know if you think it needs more commentary.

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+- **TOP**(`FutureOr<T>`) is **TOP**(T)
+- **TOP**(T) is false otherwise
+
+The **OBJECT** predicate is true for any type which is in the equivalence class
+of `Object`.
+
+
+- **OBJECT**(`Object`) is true
+- **OBJECT**(`dynamic`) is true
+- **OBJECT**(`void`) is true
```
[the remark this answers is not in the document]
leafpetersen: Fixed.

resources/type-system/upper-lower-bounds.md
```diff
@@ -0,0 +1,390 @@
...
+The **NULL** predicate is true for things in the equivalence class of `Null`
+
+- **NULL**(`Null`) is true
+- **NULL**(`T?`) is true iff **NULL**(`T`) or **BOTTOM**(`T`)
+- **NULL**(`T*`) is true iff **NULL**(`T`) or **BOTTOM**(`T`)
+- **NULL**(`T`) is false otherwise
+
+The **MORETOP** predicate defines a total order on top and `Object` types.
+
+- **MORETOP**(`void`, `T`) = true
```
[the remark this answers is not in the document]
leafpetersen: I don't think it matters.  We're imposing a total order on equivalent types, so we can choose any order we want.  We've generally chosen the order that simplifies the structure of the type, which is why I chose this order for `?` and `*`.  That is, we choose `LUB(dynamic, FutureOr<dynamic>)` to be `dynamic`, since it's a "simpler" type.  So similarly, this gives us that `LUB(dynamic, dynamic?) == dynamic`.  Either choice is valid.

resources/type-system/upper-lower-bounds.md
leafpetersen: Yeah, this is all a bit annoying.  I think it works as written, because:
  - You only initially call this when you either have two types satisfying **TOP** or two types satisfying **OBJECT**
  - Within that constraint, the only other thing that matters is that we impose a total order, which we do.

So you are correct, that we have **MORETOP**(`Object`, `Object?`), but you never ask that question directly.  Instead, you must have asked something like **MORETOP**(`Object?`, `Object??`), which recursively reduces to the above question.  So we get the answer we want, just in a weird way.

I'll think about whether I can split this out.  It feels a bit ugly though - I need a total order on all TOP types, and that means that I recursively I need to be able order an OBJECT type relative to a TOP type.

resources/type-system/upper-lower-bounds.md
leafpetersen: Actually, I think I want **MORETOP**(`void?^{k}`, `void?^{k+1}`), just to be consistent with the way we treat other things.  Refactored this, PTAL.

resources/type-system/upper-lower-bounds.md
leafpetersen: done.

resources/type-system/upper-lower-bounds.md
leafpetersen: done.

resources/type-system/upper-lower-bounds.md
leafpetersen: done.

resources/type-system/upper-lower-bounds.md
leafpetersen: done.

resources/type-system/upper-lower-bounds.md
leafpetersen: Well, `X&Null` is neither a bottom type nor a Null type, so that specific example doesn't arise.  I think this is another instance of the same weirdness you commented above, where in order to get a total order on Null types, I need to (or at least am choosing to) impose a total order on the union of the Null and the Never types.  

An example that could arise would be:  MOREBOTTOM(Null, Never?), which arises as the base case of the query MOREBOTTOM(Null?, Never??) . Again, we get a reasonable answer, but it's a bit weird how we get it.

An alternative way that I could formulate this and MORETOP is to explicitly handle the case that one of the arguments is BOTTOM and one is NULL.  So add rules that look like

MOREBOTTOM(T, S) = T if BOTTOM(T) and NULL(S)
MOREBOTTOM(S, T) = S if BOTTOM(S) and NULL(T)
 
I think I prefer to just using the total order, even though it's weird?

resources/type-system/upper-lower-bounds.md
leafpetersen: This is a bit ugly, unfortunately.  It means that we can't be completely symmetric and still respect alpha-variance.  That is, what do I do with MOREBOTTOM(`X extends Never`, `Y extends Never`)?  The only difference that I could order based on is the name of the type variable, which I really don't want to do .  Or I guess something silly like which one was declared first but.... yuck.

This is another point in favor of normalization - they both nicely normalize to `Never`.

resources/type-system/upper-lower-bounds.md
leafpetersen: done.

resources/type-system/upper-lower-bounds.md
leafpetersen: done.

resources/type-system/upper-lower-bounds.md
leafpetersen: done.

resources/type-system/upper-lower-bounds.md
leafpetersen: done.

resources/type-system/upper-lower-bounds.md
leafpetersen: ack

resources/type-system/upper-lower-bounds.md
leafpetersen: done.

resources/type-system/upper-lower-bounds.md
leafpetersen: None of this is particularly useful, but you're right, that's not as nice as it could be.  PTAL.

resources/type-system/upper-lower-bounds.md
leafpetersen: PTAL

resources/type-system/upper-lower-bounds.md
leafpetersen: Yeah, we could do better here.  PTAL

resources/type-system/upper-lower-bounds.md
leafpetersen: Well, the *, ? case is already handled in the lines above.  I don't know that there's a right answer here: if `S <: T2, S <: T1*` then `S* <: T2, S* <: T1*`, so either choice works.

resources/type-system/upper-lower-bounds.md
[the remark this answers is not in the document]
eernstg: Right, there's no soundness issue. It's just a matter of whether the given total order will be subtle and surprising to developers, or they won't even notice it because it's all so natural. ;)

But `void` and `dynamic` are perhaps extra robust here, because any equational theory we might add would be expected to satisfy `void? == void` and `dynamic? == dynamic`; `void*` would still be protected rather well (by rules against using such a value), but `dynamic*` might mean "`Object`, with unchecked member access", so the order `dynamic > dynamic*` is right.

resources/type-system/upper-lower-bounds.md
eernstg: Couldn't **MORETOP** use `Object?` in the two cases I mentioned, and then we'd have a `**MOREOBJECT**` which is simply

**MOREOBJECT(FutureOr^{k+1}<Object>, FutureOr^k<Object>) for all k >= 0.

and then we wouldn't need to look outside the definition of **MORETOP** to see that **MORETOP**(Object, Object?)` won't happen.

resources/type-system/upper-lower-bounds.md
eernstg: I think that works.

resources/type-system/upper-lower-bounds.md
eernstg: Right, and this one actually matters.

resources/type-system/upper-lower-bounds.md
eernstg: Thinking about this again, and in terms of existential types, I think it's fine as it is.

resources/type-system/upper-lower-bounds.md
eernstg: > `X&Null` is neither

Ah, yes, sorry. The robust intuition is "a **BOTTOM** type is `Never`, spelled in different ways". The **NULL** types are a bit more tricky: `Never*` is such a type, and it could stand for `Never`:

```dart
// Opt-out.
Function(X) f<X>(Function(X) f) => f;

// Opt-in.
main() => (f<Never>((_) => true))(null);
```

So we have the dual of the top quirk: **MOREBOTTOM**(`Never`, `X & Never`), even though we generally have `X & T <: T`. Again, the most simple types are in the most extreme positions, so it's consistent.

I think it's fine to keep the rules on BOTTOM-union-NULL.

resources/type-system/upper-lower-bounds.md
leafpetersen: I think what you are asking is: "why not have a MORETOP which only applies to the top types, and a MOREOBJECT which only applies to the Object types".  I think you can do this, but then you need to do a bunch of work to make sure that MORETOP doesn't get recursively called on Object types.  It's not enough to just pattern match on `Object?`, because you can have `FutureOr<Object>?` as an argument, and you need to be sure not to recursively call MORETOP on that, because it's not a top type.  I think to make this work, you do something like this:

**MORETOP**(`T?`, `S?`) where **TOP**(`T`) and **TOP**(`S`)  = **MORETOP**(`T`, `S`)
**MORETOP**(`T?`, `S?`) where **OBJECT**(`T`) and **OBJECT**(`S`) = **MOREOBJECT**(`T`, `S`)
**MORETOP**(`T?`, `S?`) where **OBJECT**(`T`) and **TOP**(`S`) = true
**MORETOP**(`T?`, `S?`) where **TOP**(`T`) and **OBJECT**(`S`) = false

and similarly for `T*` and `S*`.

I think that works fine as well.  It's a little more verbose.  And it's kind of inefficient to be checking for TOP and OBJECT on each recursive call, but in practice I doubt it really matters.  I could be convinced to specify it that way though.  Thoughts?

PR #438, opened 2019-07-04
Remove `?` from syntax. Specify override syntax details.

lrhn:


accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -118,7 +122,7 @@ The declaration introduces an extension. The extension's `on` type defines which
...
 
-(The types `dynamic` and `Never` are considered as having all members, the type `void` is always a compile-time error when used in a receiver position, so none of these can ever be affected by static extension methods. The type `Function` and all function types are considered as having a `call` member on top of any members inherited from `Object`. Methods declared on `Object` are available on all types and can therefore never be affected by extensions).
+(The types `dynamic` is considered as having all members, the type `void` and `Never` are always compile-time errors when used in a receiver position, so none of these can ever be affected by static extension methods. The type `Function` and all function types are considered as having a `call` member on top of any members inherited from `Object`. Methods declared on `Object` are available on all types and can therefore never be affected by extensions).
```
eernstg: Typo:

```suggestion
(The type `dynamic` is considered as having all members, the type `void` and `Never` are always compile-time errors when used in a receiver position, so none of these can ever be affected by static extension methods. The type `Function` and all function types are considered as having a `call` member on top of any members inherited from `Object`. Methods declared on `Object` are available on all types and can therefore never be affected by extensions).
```

Leaf wasn't quite happy about `are always compile-time errors` for `Never`. Maybe the text could simply avoid to state an answer to that question?

lrhn: True, it doesn't matter which behavior we choose for Never: Disallow any member access or allow any member access, because either will prevent extension members from applying. (Disallowing any member access is not the same as sayting that there isn't any member, it's a hard error whenever you try to do `.foo` on anything with a static type of `void` (and possibly `Never`).

Interestingly, we'll probaly *allow* `ExtName(neverExpression)`. There is no harm in that, it has a well-defined static behavior, and it will never get run at run-time anyway.

eernstg: Just noted one more thing: Leaf actually wrote 'It is an error to call a method, setter, or getter on a receiver of static type `Never` (including via a null aware operator).' in the NNBD feature spec.

lrhn: I'll keep it unspecified here since it won't make a difference for extension methods until we do get NNBD.

lrhn:
@danrubel @bwilkerson 
Updated version.

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -50,19 +50,21 @@ More precisely, an extension declaration is a top-level declaration with a gramm
...
 
 An extension declaration with a non-private name is included in the library's export scope, and a privately named extension is not. It is a compile-time error to export two declarations, including extensions, with the same name, whether they come from declarations in the library itself or from export declarations (with the usual exception when all but one declaration come from platform libraries). Extension *members* with private names are simply inaccessible in other libraries.
 
+We may want to make `extension` a built-in identifier. Is not necessary for disambiguation, but it may make parsing easier.
```
leafpetersen: Do we or don't we?

leafpetersen:
cc @stefantsov @stereotype441 @mraleph as heads up on the change.

accepted/future-releases/static-extension-methods/feature-specification.md
danrubel: Yes, `extension` is currently a built-in identifier and allows me to implement better parser recovery... especially when braces are mismatched.

lrhn: Is `on` also a built-in identifier, or was that not necessary?

PR #446, opened 2019-07-11
Specify some details for extension methods.

lrhn:
- Specify that extensions cannot declare members with the same name as `Object` members.
- Specify that extension override "expressions" cannot be used anywhere they are evaluated for their value, only as receivers of member lookups.
- Specfiy that `extension` is a built-in identifier, and `on` is not.
- Specify that the name of an extension must not be a built-in identifier.

### DIFF
--- a/accepted/future-releases/static-extension-methods/feature-specification.md
+++ b/accepted/future-releases/static-extension-methods/feature-specification.md
@@ -59,13 +59,13 @@ Such a declaration introduces its *name* (the identifier) into the surrounding s
 
 The *type* can be any valid Dart type, including a single type variable. It can refer to the type parameters of the extension.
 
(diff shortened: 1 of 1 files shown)


accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -323,6 +325,8 @@ The syntax is not *convenient*&mdash;you have to put the "constructor" invocatio
 
 An expression of the form `MyList(object)` or `MyList<String>(object)` must *only* be used for extension member access. It is a compile-time error to use it in any other way, similarly to how it is a compile-time error to use a *prefix* for anything other than member access. This also means that you cannot use an override expression as the receiver of a cascade, because a cascade does evaluate its receiver to a value. Unlike a prefix, it doesn't have to be followed by a `.` because extensions can also declare operators, but it must be followed by a `.`, a declared operator, or an arguments part (in case the extension implements `call`).
 
+If an extension override expression is the first operand of a `==` operator, then the evaluation of that operator does *not* check whether the operands are `null` before invoking the `==` method of the extension. It is treated as a completely normal static method invocation. Checking for `null` before invoking an `operator==` implementation is reserved for *instance* members, not static extension members.
```
leafpetersen: Is `MyExt(object) != whatever` allowed?  Presumably it's an error if the return type of `MyExt.==` is not bool?

leafpetersen:
This LGTM if we want to take this approach, but we may want to discuss a bit more.  You have a goal that this be forward compatible with extension types using the same syntax, and I think this interpretation may not play well with that.

PR #456, opened 2019-07-17
Normalization

leafpetersen:
This is an initial cut at a procedure for computing a normal form for type terms.


resources/type-system/normalization.md
```diff
@@ -0,0 +1,93 @@
...
+  - if `S` is `Null` then `Future<Null>?`
+  - else `FutureOr<S>`
+- **NORM**(`T?`) = 
+  - let `S` be **NORM**(`T`)
+  - if `S` is a top type then `S`
+  - if `S` is `Never` then `Null`
+  - if `S` is `Never*` then `Null`
+  - if `S` is `Null` then `Null`
+  - if `S` is `FutureOr<R>` and `R` is nullable then `S`
+  - if `S` is `FutureOr<R>*` and `R` is nullable then `FutureOr<R>`
```
eernstg: Similarly to line 65, the requirement `nullable(R)` could be reduced to `Null <: R`.

resources/type-system/normalization.md
```diff
@@ -0,0 +1,93 @@
...
+  - if `S` is `R?` then `R?`
+  - if `S` is `R*` then `R?`
+  - else `S?`
+- **NORM**(`T*`) = 
+  - let `S` be **NORM**(`T`)
+  - if `S` is a top type then `S`
+  - if `S` is `Null` then `Null`
+  - if `S` is `R?` then `R?`
```
eernstg: This is the case where normalization may eliminate the permission to access members on a receiver of type `T*` --> `R?*` --> `R?`.

resources/type-system/normalization.md
```diff
@@ -0,0 +1,93 @@
...
+  - if `S` is `R*` then `R*`
+  - else `S*`
+- **NORM**(`X extends T`) =
+  - let `S` be **NORM**(`T`)
+  - if `S` is `Never` then `Never`
+  - else `X extends T`
```
eernstg: `X extends S`, presumably?

resources/type-system/normalization.md
```diff
@@ -0,0 +1,93 @@
...
+- **NORM**(`X & T`) =
+  - let `S` be **NORM**(`T`)
+   - if `S` is `Never` then `Never`
+   - if `S` is a top type then `X`
```
eernstg: This seems to eliminate the permission (no-implicit-casts) to assign a variable of type `X & T` to a variable of type `Object` in the case where `S` is `Object*`.

resources/type-system/normalization.md
```diff
@@ -0,0 +1,93 @@
...
+equivalent types.  In the absence of legacy (*) types, it should be the case
+that for any two types which are mutual subtypes, their normal forms are either
+one of the primitive top types, or are syntactically identical.
+
+This is based on the following equations:
+- `T?? == T?`
+- `T?* == T?`
+- `T*? == T?`
+- `T** == T*`
+- `Null? == Null`
```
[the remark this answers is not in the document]
lrhn: That should probably be true.
A `*` type is different from its base type and its nullable base type because it can be used as either, co- and contra-variantly.
However if `Null?` and `Null` are equivalent, then that distinction doesn't matter and `Null*` should be equivalent to both.

resources/type-system/normalization.md
```diff
@@ -0,0 +1,93 @@
...
+- `Never? == Null`
```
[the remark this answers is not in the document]
lrhn: `Never*` may be used as either `Never` or `Never?`, which `Null` cannot - you cannot assign `Null` to `Never`, that's a downcast, but you can assign `Never*`.

resources/type-system/normalization.md
```diff
@@ -0,0 +1,93 @@
...
+## Normalization
+
+The **NORM** relation defines the canonical representative of classes of
+equivalent types.  In the absence of legacy (*) types, it should be the case
+that for any two types which are mutual subtypes, their normal forms are either
+one of the primitive top types, or are syntactically identical.
+
+This is based on the following equations:
+- `T?? == T?`
+- `T?* == T?`
```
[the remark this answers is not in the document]
lrhn: Here `T?*` is assignable to either `T?` or `T??`, which doesn't matter since they are equivalent, so `T?*` it's equivalent to `T?` wrt. assignability and usability.

resources/type-system/normalization.md
```diff
@@ -0,0 +1,93 @@
...
+- `T*? == T?`
+- `T** == T*`
+- `Null? == Null`
+- `Never? == Null`
+- `dynamic? == dynamic`
+- `void? == void`
```
lrhn: `dynamic*` == `dynamic`
`void*` == `void`
?

resources/type-system/normalization.md
```diff
@@ -0,0 +1,93 @@
...
+
+Applying these equations recursively and making a canonical choice when multiple
+equations apply gives us something like the following:
+
+- **NORM**(`T`) = `T` if `T` is primitive
+- **NORM**(`FutureOr<T>`) =
+  - let `S` be **NORM**(`T`)
+  - if `S` is a top type then `S`
+  - if `S` is `Object` then `S`
+  - if `S` is `Object*` then `S`
```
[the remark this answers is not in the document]
lrhn: I would not expect that. The top types are `void`, `dynamic` and `Object?`. There are other types which happen to be a super-type of a top type, but these all normalize to a top-type.

`FutureOr<Object*>` is assignable to `FutureOr<Object>` which is not a top-type, without a down-cast, so it's not properly a top-type.

resources/type-system/normalization.md
```diff
@@ -0,0 +1,93 @@
...
+  - let `S` be **NORM**(`T`)
+  - if `S` is a top type then `S`
+  - if `S` is `Never` then `Null`
+  - if `S` is `Never*` then `Null`
+  - if `S` is `Null` then `Null`
+  - if `S` is `FutureOr<R>` and `R` is nullable then `S`
+  - if `S` is `FutureOr<R>*` and `R` is nullable then `FutureOr<R>`
+  - if `S` is `R?` then `R?`
+  - if `S` is `R*` then `R?`
+  - else `S?`
```
lrhn: No handling of S being `Object*` here. So either it is considered a top type, or it should be handled here too.

resources/type-system/normalization.md
lrhn: This can only happen in a migrated library. A non-migrated library will see a neutered type for the migrated library's API with only `*` types, so it cannot create a `T?*`.

In a migrated library, you can create an instance of a non-migrated library with a `T?` as type argument. That might give you `T?*` as an API type. You should treat that as `T?`, because it may definitely be `null`.

nshahan:
Can this PR get cleaned up and landed? This doc is useful for the implementations and would be helpful source of truth in the repo.

resources/type-system/normalization.md
```diff
@@ -0,0 +1,93 @@
...
+
+We assume that type aliases are fully expanded, and that prefixed types are
+resolved to a canonical name.
+
+## Normalization
+
+The **NORM** relation defines the canonical representative of classes of
+equivalent types.  In the absence of legacy (*) types, it should be the case
+that for any two types which are mutual subtypes, their normal forms are either
+one of the primitive top types, or are syntactically identical.
```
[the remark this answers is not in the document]
leafpetersen: Added a comment about this.

resources/type-system/normalization.md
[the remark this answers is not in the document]
leafpetersen: It's an error to access a member on `T?*` as well.

resources/type-system/normalization.md
leafpetersen: No, if we have `FutureOr<int*>?`, we don't want to reduce that to `FutureOr<int*>`.

resources/type-system/normalization.md
leafpetersen: done.

resources/type-system/normalization.md
leafpetersen: `Object*` is not a top type, so I don't think this applies?

resources/type-system/normalization.md
leafpetersen: done.

resources/type-system/normalization.md
leafpetersen: Not sure I follow.  If `S` is `Object*`, then this says the result is `Object?` which is what we want, no?

leafpetersen:
This is now landed.

Issue #440, opened 2019-07-05
Several related proposals for definite assignment and type promotion with NNBD


stereotype441:
This issue captures some in-person discussion @lrhn and I had yesterday and today.  A number of proposals were floated, and I'd like to hear others' opinions.  Lasse, please let me know if I get some details wrong.

### Proposal: promotions are recorded as a chain.
It would be nice to be able to write the following code:
```dart
// Example 1
void f(Object x) {
  if (x is! num) throw ...; // Promotes `x` to `num`
  if (x is int) { // Promotes `x` to `int`
    ...
  } else {
    if (x is! double) throw ...; // Promotes `x` to `double`.
  }
  x + 1; // Ok because `x` is known to be `num`.
}
```
(Note that all these examples assume the code has been "opted in", so that types are non-nullable by default, and enhanced type promotion and definite assignment analysis are enabled). 

Consider what happens when type promotion analysis reaches the end of the second `if` statement.  It has to join a control flow path in which `x` is promoted to `int` with a control flow path in which `x` is promoted to `double`.  We would like it to remember the promotion to `num`.

The proposed way to do this is for the promotion process to record each promotion as a chain of types, where each type in the chain is a subtype of the next link in the chain.  So `if (x is! num) throw ...;` promotes `x` from `Object` to `num <: Object`; `if (x is int)` promotes to `int <: num <: Object`; and `if (x is! double) throw ...;` promotes to `double <: num <: Object`.  Joining two control flow paths is now a simple* matter of rewinding mismatched parts of the two chains until a matching type is found.

(*Ok, maybe it's not actually that simple.  How do we join, for example, `int <: Object` with `int <: num <: Object`?  What if the two chains form a diamond, e.g. `List<int> <: List<num> <: Iterable<num>` and `List<int> <: Iterable<int> <: Iterable<num>`?)

### Proposal: some assignments promote.
It would be nice for type promotion to account for assignments, so e.g. one can do something like this:
```dart
// Example 2
void f(Object o) {
  if (o is! String) o = o.toString();
  o.length; // Ok because `o` has promoted type `String`
}
```

However, not every assignment should promote.  Sometimes the user wants to explicitly declare a variable to have a more general type in order to avoid accidentally making use of more specific methods, and we want to preserve that ability.  For example:
```dart
// Example 3
void f() {
  Iterable<int> x = ...;
  x = <int>[1, 2, 3];
  x[0] = 4; // Compile-time error: `Iterable<int>` does not support `operator[]`.
[...]

PR #473, opened 2019-07-26
Flow analysis preliminaries

leafpetersen:
This is not for submittal, just using this as a place for discussion of the setup for flow analysis, and the policy choices.

### DIFF
--- a/resources/type-system/flow-analysis-v0.md
+++ b/resources/type-system/flow-analysis-v0.md
@@ -1,0 +1,339 @@
+# Flow Analysis for Non-nullability
+
+paulberry@google.com, leafp@google.com
+
+## Terminology and Notation
+
+We use the generic term *node* to denote an expression, statement, top level
+function, method, constructor, or field.  We assume that all nodes are labelled
+uniquely in some unspecified fashion such that mappings can be maintained from
+nodes to the results of flow analysis.
+
+The flow analysis assumes that the set of variables which are assigned in any
+given node has been pre-computed, and can be queried by a predicate **Assigned**
+such that for a node `N` and a variable `v`, **Assigned**(`N`, `v`) is true iff
+`N` syntactically contains an assignment to `v` (regardless of reachability of
+that assignment).
+
+### Models
+
+A *type test site* is a location in the program where a variable's type is
+tested, either via an `is` expression or a cast.
+
+A *variable model*, denoted `VariableModel(declaredType, promotedTypes,
+testSites, assigned, unassigned, writeCaptured)`, represents what is statically
+known to the flow analysis about the state of a variable at a given point in the
+source code.
+
+- `declaredType` is the type assigned to the variable at its declaration site
+  (either explicitly or by type inference).
+
+- `promotedTypes` is an ordered set of types that the variable has been promoted
+  to, with the final entry in the ordered set being the current promoted type of
+  the variable.  Note that each entry in the ordered set must be a subtype of
+  all previous entries, and of the declared type.
+
+- `testSites` is an ordered list of *type test site*s representing the set of
+  type tests that are known to have been performed on the variable in all code
+  paths leading to the given point in the source code.
+
+- `assigned` is a boolean value indicating whether the variable is known to
+  have been definitely assigned at the given point in the source code.
+
+- `unassigned` is a boolean value indicating whether the variable is known not
+  to have been definitely assigned at the given point in the source code.  (Note
+  that a variable may not be both definitely assigned and definitely
+  unassigned).
+
+- `writeCaptured` is a boolean value indicating whether a closure might exist at
+  the given point in the source code, which could potentially write to the
+  variable.
+
+A *flow model*, denoted `FlowModel(reachable, variableInfo)`, represents what
+is statically known to flow analysis about the state of the program at a given
+point in the source code.
+
+  - `reachable` is a stack of boolean values modeling the reachability of the
+  given point in the source code.  The ith element of the stack (counting from
+  the top of the stack) indicates whether the given program point is reachable
+  from the ith enclosing control flow split in the program.  If the bottom
+  element of `reachable` is `false`, then the given point is definitively known
+  by flow analysis to be unreachable from the start of the method under
+  analysis.  If it is `true`, then the analysis cannot eliminate the possibility
+  that the given point may be reached by some path.  Each other element of the
+  stack models the same property, starting from some control flow split between
+  the start of the program and the current node.
+
+  - `variableInfo` is a mapping from variables in scope at the given point to
+  their associated *variable model*s.
+
+  - We will use the notation `{x: VM1, y: VM2}` to denote a map associating the key
+    `x` with the value `VM1`, and the key `y` with the value `VM2`.
+
+  - We will use the notation `VI[x -> VM]` to denote the map which maps every
+    key in `VI` to its corresponding value in `VI` except `x`, which is mapped
+    to `VM` in the new map (regardless of any value associated with it in `VI`).
+
+  - We will use the notation `[a, b]` to denote a list containing elements `a`
+    and `b`.
+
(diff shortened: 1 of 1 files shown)


resources/type-system/flow-analysis-v0.md
```diff
@@ -0,0 +1,339 @@
...
+
+- `writeCaptured` is a boolean value indicating whether a closure might exist at
+  the given point in the source code, which could potentially write to the
+  variable.
+
+A *flow model*, denoted `FlowModel(reachable, variableInfo)`, represents what
+is statically known to flow analysis about the state of the program at a given
+point in the source code.
+
+  - `reachable` is a stack of boolean values modeling the reachability of the
```
leafpetersen: So what's going on here?  This is my attempt to incorporate Paul's extended reachability into the basic framework.  I think this is roughly what the extended reachability was doing, except that instead of keeping the stack of previous reachability values on the call stack, I'm keeping it explicitly in the analysis.  An implementation could do either.  So  far, I've found the book-keeping a little cleaner here.  Note that I'm possibly applying this in more places than Paul had originally intended?  For example, I think the following is a motivating example:

```
Object x;
if (false && x is int) {
  x.isEven
}
```

I'm not sure whether Paul intended this to promote in the body or not, but it seemed reasonable, and I *think* if I have this set up right, then we will get that.

resources/type-system/flow-analysis-v0.md
```diff
@@ -0,0 +1,339 @@
...
+  as `join(M, M)`.
+
+- `join(M1, M2)`, where `M1` and `M2` are flow models, represents the union of
+  two flow models and is defined as follows:
+
+  We define `join(M1, M2)` where `M1 = FlowModel(r1, VI1)` and `M2 =
+  FlowModel(r2, VI2))` and `pop(r1) = pop(r2) = r0` for some `r0` to be `M3` where:
+    - if `top(r1)` is true and `top(r2)` is false, then `M3` is `FlowModel(r0, VI1)`.
+    - if `r1` is false and `r2` is true, then `M3` is `FlowModel(pop(r2), VI2)`.
+    - otherwise `M3 = FlowModel(r3, VI3)` where:
```
leafpetersen: I'm not 100% clear whether I'm doing the right thing here.  It's clear what to do if one of the branches is provisionally reachable and the other is not, and it's clear what to do if both are provisionally reachable.  But what about if both are not provisionally reachable?  My guess here is that the right thing to do is to treat them both as provisionally reachable - that is, keep both of their states.  But I'm not 100% sure about that.  

```dart
  foo(Object x, bool b) {
    throw "Unimplemented";
    if (b) {
      if (x is! int) return;
      throw "Unimplemented";
   } else {
     if (x is! int return;
     throw "Unimplemented";
  }
  x.isEven; // should work?
}
```

resources/type-system/flow-analysis-v0.md
```diff
@@ -0,0 +1,339 @@
...
+
+- `exit(M)` represents the model corresponding to a program location which is
+  unreachable, but is otherwise modeled by flow model `M = FlowModel(r, VI)`,
+  and is defined as `FlowModel(push(pop(r), false), VI)`
+
+### Promotion
+
+Promotion policy is defined by the following operations on flow models.
+
+Policy:
```
leafpetersen: I've tried to capture the policy options for assignment promotion that were proposed [here](https://github.com/dart-lang/language/issues/440) here.  What other options should I include?

resources/type-system/flow-analysis-v0.md
```diff
@@ -0,0 +1,339 @@
...
+    - and `captured` is false
+    - and `promoted = S::l` or (`promoted = []` and `declared = S`)
+    - and `T <: S` and not `S <: T`
+    - and `T` is a type of interest for `x` in `sites`
+
+  - We say that a variable `x` is demotable via assignment of an expression of
+    type `T` given variable model `VM` if
+    - `VM = VariableModel(declared, promoted, sites, assigned, unassigned, captured)`
+    - and `captured` is false
+    - and promoted contains `T`
```
leafpetersen: Allow demotion to anything that we had previously promoted to?  Probably also need to include the declared type here.

leafpetersen:
cc @lrhn @johnniwinther @scheglov @eernstg @munificent @stereotype441 

I'm starting to work on pulling together Paul's proposal and the various discussions that have been going on into a spec for the flow analysis based type promotion etc.  This is some initial machinery and policy discussion that I thought would be useful to get some early feedback on.

resources/type-system/flow-analysis-v0.md
```diff
@@ -0,0 +1,339 @@
...
+
+Analysis of an expression `N` assumes that `before(N)` has been computed, and
+uses it to derive `after(N)`, `null(N)`, `notNull(N)`, `true(N)`, and
+`false(N)`.
+
+If `N` is an expression, and the following rules specify the values to be
+assigned to `true(N)` and `false(N)`, but do not specify values for `null(N)`,
+`notNull(N)`, or `after(N)`, then they are by default assigned as follows:
+  - `null(N) = exit(after(N))`.
+  - `notNull(N) = join(true(N), false(N))`.
```
stereotype441: I think this has to be `notNull(N) = join(split(true(N)), split(false(N)))`

resources/type-system/flow-analysis-v0.md
```diff
@@ -0,0 +1,339 @@
...
+
+### Statements
+
+- **Conditional statement**: If `N` is a conditional statement of the form `if
+  (E) S1 else S2` then:
+  - Let `before(E) = before(N)`.
+  - Let `before(S1) = split(true(E))`.
+  - Let `before(S2) = split(false(E))`.
+  - Let `after(N) = join(after(S1), after(S2))`.
+  - Let `true(N) = join(true(S1), true(S2))`.
```
stereotype441: Since statements don't have values, we shouldn't need `true(N)`, `false(N)`, `null(N)`, or `notNull(N)`.  Just `after(N)` is sufficient.

resources/type-system/flow-analysis-v0.md
```diff
@@ -0,0 +1,339 @@
...
+
+- `assigned` is a boolean value indicating whether the variable is known to
+  have been definitely assigned at the given point in the source code.
+
+- `unassigned` is a boolean value indicating whether the variable is known not
+  to have been definitely assigned at the given point in the source code.  (Note
+  that a variable may not be both definitely assigned and definitely
+  unassigned).
+
+- `writeCaptured` is a boolean value indicating whether a closure might exist at
```
stereotype441: I'm starting to implement the logic for `writeCaptured` in the prototype, and I wanted to let you know about two corner cases I've run into.  First, what if a write happens in one closure and a read happens in another:
```dart
f(Object x) {
  void Function() h;
  var g = () {
    if (x is int) { // (1)
      h();
      x.isEven; // (2)
    }
  };
  h = () { x = "hello"; }; // (3)
  g();
}
```
Even though the if-test at (1) appears lexically before the closure that captures a write to `x` is created (at (3)), since (1) is itself inside a closure, there's no guarantee that it will execute prior to the write being captured.  So promotion is unsafe in this case, and (2) should be a compile-time error.  I'm handling this situation by saying that prior to analysis, we gather a list of all variables that are written to in all closures.  Then, when entering any closure, analysis sets `writeCaptured` to true for any variables in that list.

Second, what if a write capturing closure happens in a loop, after a promotion attempt?
```dart
f(Object x) {
  void Function()? g;
  while (true) {
    if (x is int) { // (1)
      if (g != null) g();
      x.isEven; // (2)
    }
    g = () { x = 'hello'; };
  }
}
```
In this case the promotion at (1) is unsafe, because the second time through the loop, there is a write captured.  So there should be a compile-time error at (2).  I'm handling this situation by saying that when entering a loop, we set `writeCaptured` to true for any variables that are captured ahywhere within the loop.

PR #483, opened 2019-08-01
Clarify static extension method access

eernstg:
The feature spec does not explicitly make accesses to static members using explicit extension member syntax (like `E(e).staticMember`) an error, it just says that a static member _can_ be accessed using the normal static member syntax `E.staticMember`.

This PR changes the rule for overriding access (like `E(e).m`) such that it is explicitly mentioned that the member (`m`) must be an extension _instance_ member.


leafpetersen:
Is this waiting on an LGTM from @lrhn ?

eernstg:
Yes, I just wanted buy-in from Lasse. May come any minute.

lrhn:
LGTM.

PR #484, opened 2019-08-01
Update .travis.yml

mit-mit:
Upgrade to Ubuntu 18.04


mit-mit:
This should give us TeX 2017 according to https://www.tug.org/texlive/debian.html

PR #512, opened 2019-08-09
Transferred several grammar adjustments from Dart.g to the spec

eernstg:
This is a language specification update that transfers several elements from Dart.g to the language specification.

The grammar rule for `<declaration>` contains a correction which makes a difference: It allows for `late final x;` as well as `late final x = 42;`, which wasn't allowed previously.

A `<topLevelVariableDeclaration>` has been added, because the use of `<variableDeclaration>` created the problem that it allowed `const x;` and `final x;` as top level declarations, and we already have a separate alternative in `<topLevelDefinition>` to allow for `const` and `final` variables as we want to allow them at the top level.

At the same time, a handful of declarations do not have `<metadata>` at the beginning any more (because it's now handled `<metadata> <topLevelDefinition>` in general, such that we also allow metadata for `external <getterSignature>;` and several other cases where it was not supported, according to the spec).

The most controversial part could be the change to `<assignableSelectorPart>`, but I believe that the form I've used is simpler and capable of deriving the same terms: It uses the ability of `<selector>` to be a `!` also in an `<assignableSelectorPart>`, and this is something that `<selector>` can already do (without any changes from me).

Finally, the `<topLevelDefinition>` has also been adjusted to allow `late final x;` at top level.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -1833,7 +1833,8 @@ A \Index{class} defines the form and behavior of a set of objects which are its
 Classes may be defined by class declarations as described below, or via mixin applications (\ref{mixinApplication}).
 
 \begin{grammar}
-<classDefinition> ::= <metadata> \ABSTRACT{}? \CLASS{} <identifier> <typeParameters>?
+<classDefinition> ::=
+  \ABSTRACT{}? \CLASS{} <identifier> <typeParameters>?
   \gnewline{} <superclass>? <interfaces>?
   \gnewline{} `{' (<metadata> <classMemberDefinition>)* `}'
   \alt <metadata> \ABSTRACT{}? \CLASS{} <mixinApplicationClass>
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -1850,20 +1851,21 @@ \section{Classes}
...
-  \alt \STATIC{} \CONST{} <type>? <staticFinalDeclarationList>
-  \alt \COVARIANT{} \LATE{}? (\VAR{} | <type>) <initializedIdentifierList>
-  \alt \LATE{}? \FINAL{} <type>? <initializedIdentifierList>
-  \alt \LATE{}? (\VAR{} | <type>) <initializedIdentifierList>
+  \alt \EXTERNAL{}? <operatorSignature>
+  \alt \STATIC{} (\FINAL{} | \CONST{}) <type>? <staticFinalDeclarationList>
+  \alt \STATIC{} \LATE{} \FINAL{} <type>? <initializedIdentifierList>
+  \alt (\STATIC{} | \COVARIANT{}) \LATE{}? <varOrType>
+       <initializedIdentifierList>
+  \alt \LATE{}? (\FINAL{} <type>? | <varOrType>) <initializedIdentifierList>
```
leafpetersen: I don't understand this re-organization.  Your comment is:

> The grammar rule for <declaration> contains a correction which makes a difference: It allows > for late final x; as well as late final x = 42;, which wasn't allowed previously.

but the text on the left includes ` \alt \LATE{}? \FINAL{} <type>? <initializedIdentifierList>` which permits both of those derivations.  What's happening here?

leafpetersen: Ok, per comments elsewhere, this is apparently about `static late final` without an initializer?  What about just splitting ` \alt \STATIC{} \LATE{}? \FINAL{} <type>? <staticFinalDeclarationList>` into

```
 \alt \STATIC{} \FINAL{} <type>? <staticFinalDeclarationList>
 \alt \STATIC{} \LATE{} \FINAL{} <type>? <initializedIdentifierList>
```

I really find these much more readable when the prefix of the alternatives reads off directly, even if it's more verbose.

specification/dartLangSpec.tex
```diff
@@ -13536,23 +13539,26 @@ \section{Libraries and Scripts}
...
+  \alt <getterSignature> <functionBody>
+  \alt <setterSignature> <functionBody>
+  \alt (\FINAL{} | \CONST{}) <type>? <staticFinalDeclarationList> `;'
+  \alt \LATE{} \FINAL{} <type>? <initializedIdentifierList> `;'
+  \alt <topLevelVariableDeclaration> `;'
 
-<getOrSet> ::= \GET{}
-  \alt \SET{}
+<topLevelVariableDeclaration> ::=
+  \LATE{}? <varOrType> <identifier> (`=' <expression>)?
```
leafpetersen: Can't the tail of this be replaced with `initializedIdentifierList`?

specification/dartLangSpec.tex
```diff
@@ -13536,23 +13539,26 @@ \section{Libraries and Scripts}
...
-  \alt <variableDeclaration> `;'
+  \alt <getterSignature> <functionBody>
+  \alt <setterSignature> <functionBody>
+  \alt (\FINAL{} | \CONST{}) <type>? <staticFinalDeclarationList> `;'
+  \alt \LATE{} \FINAL{} <type>? <initializedIdentifierList> `;'
+  \alt <topLevelVariableDeclaration> `;'
 
-<getOrSet> ::= \GET{}
-  \alt \SET{}
+<topLevelVariableDeclaration> ::=
```
leafpetersen: Any reason this is a separate definition instead of inline like the rest?

eernstg: No, inlined it.

specification/dartLangSpec.tex
eernstg: Of course! Done.

specification/dartLangSpec.tex
eernstg: I've split them all, that is: (const | final), (static | covariant), (final type? | varOrType).

Issue #508, opened 2019-08-08
Introduce a strict-inference static analysis mode.


srawlins:
The goal is to catch more error-prone code during static analysis; for example, code which is guaranteed to error at runtime. This often two secondary effects: (1) reducing errors in developer understanding of such code, and (2) reducing dynamic dispatch in code where a developer did not think dynamic dispatch was occurring.

Here are some examples of such code:

```
fn(List<int> strings) => print(strings.first.isEven);

void main() {
  var args = ["one", "two", "three"];
  var useArgs = true;
  fn(useArgs ? args : []);
}
```

Depending on the value of `useArgs`, the last line is guaranteed to result in a runtime error. The surprise to the developer is that  `args` was even allowed on that line! The problem here is that the developer thought that inference would infer the type of `[]` from the type of `args`. A "strict-inference" mode changes static analysis rules such that inference of `[]`  does not  _fall back_ to dynamic, but fails.

Another common example that comes up on mailing lists is inference in [Iterable.fold](https://api.dartlang.org/stable/2.4.1/dart-core/Iterable/fold.html). Note  the signature of `Iterable<E>#fold<T>`:

> `T fold <T>(T initialValue, T combine(T previousValue, E element))`

```dart
var a = [1, 2, 3].fold(true, (s, x) => s + x);
```

There are  currently no compile-time errors in this code, but it will fail at runtime. A "strict-inference" mode would fail to infer the callback's type at compile-time. If the developer changes the type of `a` to an explicit `bool`, or adds an explicit type annotation to `fold`, they will then see more clearly the error in their code.

The proposed "strict-inference" mode will report errors (e.g. Hints in analyzer) at the declaration of values whose types cannot be inferred. Here is an inexhaustive list (an exhaustive list will come with a more formal specification):

* a variable declared without a type (e.g. with `var` or `final`) and no initializer.
* a field declared without a type and no initializer.
* a function parameter declared without a type, excluding a method parameter whose types are inferred from a  parent class (inherited?  inferred?).
* an empty collection literals whose type(s) cannot be inferred from "above."
* an instantiation of a generic class, where the type argument(s) are neither explicitly provided, nor can be inferred from downwards inference or upwards inference, e.g. `var f = Future.error("Boo");`
* an invocation of a generic function,where  the type argument(s)  are neither explicitly provided, nor can be inferred from downwards inference or upwards inference.

[...]

PR #518, opened 2019-08-12
Add strict-inference specification

srawlins:
We need a formal specification for  "strict-inference", so that it can be properly implemented and 
 tested (e.g. two in progress CRs, [112142](https://dart-review.googlesource.com/c/sdk/+/112142) and [111740](https://dart-review.googlesource.com/c/sdk/+/111740)). @leafpetersen wrote that we can land a spec in this language repository.

Fixes https://github.com/dart-lang/language/issues/508

### DIFF
--- a/working/0508 - strict-inference/strict-inference.md
+++ b/working/0508 - strict-inference/strict-inference.md
@@ -1,0 +1,347 @@
+# Strict inference static analysis option
+
+This document specifies the "Strict inference" mode enabled with a static
+analysis option. As a static analysis option, we only intend to implement this
+feature in the Dart Analyzer. Under this feature, when there is not enough
+information available to infer an expression's type, rather than "falling back"
+to `dynamic` (or the type's  bound), inference is considered to have failed,
+and an analyzer "Hint" is reported at the location of the expression.
+
+## Enabling strict inference
+
+To enable strict inference, set the `strict-inference` option to `true`, under
+the Analyzer's  `language` section:
+
+```yaml
+analyzer:
+  langauge:
+    strict-inference: true
+```
+
+## Motivation
+
+It is possible to write Dart code that passes all static analysis and
+compile-time checks that is guaranteed to result in runtime errors. Commone
+examples include runtime type errors, and no-such-method errors. New and
+experienced Dart developers alike write such code, and are surprised to see
+such errors at runtime, which look like they should be caught at compile time.
+
+The strict inference mode aims to highlight such code during static analysis.
+We can look at two common examples:
+
+```dart
+fn(List<int> numbers) => print(numbers.first.isEven);
+
+void main() {
+  var args = ["one", "two", "three"];
+  var useArgs = true;
+  fn(useArgs ? args : []);
+}
+```
+
+Depending on the value of `useArgs`, the last line will result in a runtime
+error. The developer thinks that `args` is an appropriate value to send to
+`fn`, because surely static analysis would have reported if this wasn't so. In
+real world cases of this code, `fn` and `main` may be authored by different
+developers, and may be located in separate packages. `args` may have been
+created in a third location, by a third developer, with a complex generic type.
+Static analysis is supposed to help developers catch where they have
+misunderstood an API or the type of an object they are handling.
+
+The developer might think that the type of `[]` is unimportant, or that it is
+inferred from the `numbers` parameter type, or that it is inferred from the
+expression to the left of `:`. None of these are true. The type of `[]` is
+inferred from the types of its elements. As their are none, it "falls back" to
+`dynamic`. Then type of the conditional expression `useArgs ? args : []` is
+inferred from the "then" and "else" expressions, `args` (`List<String>`) and
+`[]` (`List<dynamic>`), resulting in LUB of the two, `List<dynamic>`.
+
+Static analysis allows a `List<dynamic>` argument for a `List<int>`, as an
+implicit cast. At runtime, however, `args` (a `List<String>`) fails to cast to
+`List<int>`.
+
+To prevent such an error, an empty collection literal (`[]`, `{}`) without an
+explicit type argument, and whose type cannot be inferred from downwards
+inference is considered to have an inference failure. In this example, the
+strict inference failure would report that the type of `[]` cannot be inferred,
+and suggest that the developer add an explicit type argument. The developer
+would like ly add `<String>`, thinking "`args` is a `List<String>` and I think
+that `fn` accepts a `List<String>`, so I'll make the empty list alternative
+also a `List<String>`." At this point, existing static analysis will inform the
+developer that a `List<String>` cannot be passed where a `List<int>` is
+expected.
+
+Let's look at a second example. Consider the signature of [`Iterable.fold`]:
+
+> T fold <T>(T initialValue, T combine(T previousValue, E element))
+
+```dart
+void main() {
+  var a = [1, 2, 3].fold(true, (s, x) => s + x);
+}
+```
+
+There are no compile-time, static analysis errors in this code, but it is
+guaranteed to produce a failure at runtime, when `true + 1` is executed. The
+issue here is similar to the previous one: the developer likely thinks that the
+type of `s` (`T`) will be inferred from the type of `initialValue`, and that
+static analysis would report any issue with that type. But inference doesn't
+flow between parameters like that.  Instead, while trying to infer `T`, there
+is not enough information from downwards inference (the type of `a`), and there
+is not enough information from upwards inference (the type of `(s, x) => s +
+x`). Inference gives up on the type of the function literal, and gives it
+`dynamic Function(dynamic, int)`. Inference combines that type with `bool`, the
+type of `true`, and settles on `dynamic` for T.
+
+The issue would be revealed with either an explicit type on `fold`, an explicit
+type for `a`, which would help to infer the type of `s`, or an explicit type
+for `s`:
+
+```dart
+void main() {
+  // Each of these produce an existing error:
+  // "The operator '+' isn't defined for the class 'bool'."
+  var b = [1, 2, 3].fold<bool>(true, (s, x) => s + x);
+  bool c = [1, 2, 3].fold(true, (s, x) => s + x);
+
+  // This produces an existing error:
+	// "Couldn't infer type parameter 'T'. Tried to infer 'Object' for 'T' which
+  // doesn't work: Parameter 'combine' declared as 'T Function(T, int)' but
+  // argument is 'int Function(int, int)'. The type 'Object' was inferred from:
+  // Parameter 'initialValue' declared as 'T' but argument is 'bool'. Consider
+  // passing explicit type argument(s) to the generic."
+  var d = [1, 2, 3].fold(true, (int s, x) => s + x);
+}
+```
+
+In strict inference mode, the inference failure on `(s, x) => s + x` will be
+reported, enouraging the developer to add a type to `s`, `a`, or `fold`,
+revealing their misunderstaing of the types.
+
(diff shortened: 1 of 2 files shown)


working/0508 - strict-inference/strict-inference.md
```diff
@@ -0,0 +1,347 @@
...
+Static analysis allows a `List<dynamic>` argument for a `List<int>`, as an
+implicit cast. At runtime, however, `args` (a `List<String>`) fails to cast to
+`List<int>`.
+
+To prevent such an error, an empty collection literal (`[]`, `{}`) without an
+explicit type argument, and whose type cannot be inferred from downwards
+inference is considered to have an inference failure. In this example, the
+strict inference failure would report that the type of `[]` cannot be inferred,
+and suggest that the developer add an explicit type argument. The developer
+would like ly add `<String>`, thinking "`args` is a `List<String>` and I think
```
bwilkerson: nit: "like ly" --> "likely"

working/0508 - strict-inference/strict-inference.md
```diff
@@ -0,0 +1,347 @@
...
+  // argument is 'int Function(int, int)'. The type 'Object' was inferred from:
+  // Parameter 'initialValue' declared as 'T' but argument is 'bool'. Consider
+  // passing explicit type argument(s) to the generic."
+  var d = [1, 2, 3].fold(true, (int s, x) => s + x);
+}
+```
+
+In strict inference mode, the inference failure on `(s, x) => s + x` will be
+reported, enouraging the developer to add a type to `s`, `a`, or `fold`,
+revealing their misunderstaing of the types.
```
bwilkerson: nit: "misunderstaing" --> "misunderstanding"

working/0508 - strict-inference/strict-inference.md
```diff
@@ -0,0 +1,350 @@
...
+}
+```
+
+The failure is reported on the identifier of the variable or field.
+
+### Function parameter
+
+A function parameter declared without a type (via `var`, `final` or without a
+modifier), which does not inherit a type (in the case of a method), and whose
+type cannot be inferred from downwards inference (in the case of a function
```
bwilkerson: > ... whose type cannot be inferred from downwards inference ...

I think you intended to say a little more than that (here and elsewhere). Specifically "inferred as something other than `dynamic`".

working/0508 - strict-inference/strict-inference.md
```diff
@@ -0,0 +1,350 @@
...
+  final f;      // Inference failure
+  final g = 7;  // OK
+
+  C(this.f);
+
+  static var s; // Inference failure
+}
+```
+
+The failure is reported on the identifier of the variable or field.
```
bwilkerson: We have generally taken the approach that language documents specify what is an isn't valid (in this case, what is or isn't an inference failure), but leave details such as where the error is reported to the tools to decide. I'd prefer that we drop lines like this one from this specification.

working/0508 - strict-inference/strict-inference.md
```diff
@@ -0,0 +1,350 @@
...
+function-typed function parameter without a return type is an inference
+failure. The return type of non-recursive local functions can always be
+inferred from downwards or upwards inference, as it will have a body, and the
+return type of the body is known (even if there are inference failures within).
+
+TODO(srawlins): Examples
+
+## Cascading failures
+
+Cascading inference failures are not reported. That is, when inference cannot
```
bwilkerson: It's probably sufficient, and perhaps better, to simply say that this proposal does not change the way type inference works, only what is or is not reported to the user.

working/0508 - strict-inference/strict-inference.md
srawlins: Done

working/0508 - strict-inference/strict-inference.md
srawlins: Done

working/0508 - strict-inference/strict-inference.md
srawlins: Fine by me! Done.

working/0508 - strict-inference/strict-inference.md
srawlins: Even dynamic would be fine. For example:

```dart
fn(void callback(dynamic x), dynamic x) { callback(x); }
fn((var x) => print(x));
```

Strict inference will not report any inference failures; the parameter type on the second line is inferred to be dynamic. The first line is filled with explicit dynamics.

Or the method case:

```dart
class C {
  void f2(dynamic a) => print(a);
}

class D extends C {
  @override
  void f2(a) => print(a); // OK; inherits dynamic type
}
```

working/0508 - strict-inference/strict-inference.md
srawlins: I've clarified this at the top, good  suggestion.

I think I would like this section at the bottom, because these questions come up a lot. Originally Jenny even implemented strict-inference to cascade things like `var a; var b = [a];`  so that `b` also has an inference failure. The return-type example came up in internal discussions.

I made up the for-each and field-initializing parameter examples while writing the rest of the spec, so I don't feel strongly about them. But they serve as interesting examples of "there is no type here, and yet this never means an inference failure."

WDYT?

bwilkerson: I think it's fine if you want to leave the section, but it might be good to clarify that all of the things mentioned in this section fall out naturally from the fact that type inference has inferred `dynamic` and that there is no additional requirements being added by this section. (Basically the normative / non-normative distinction from the language spec.)

srawlins: OK, I've added that distinction, and clarified the text.

srawlins:
@leafpetersen @munificent I'd love a review from one of you, as you've reviewed previous iterations.

> How many of these cases are covered by NNBD?

Do you mean in terms of "Extended Type promotion, Definite Assignment, and Definite Completion"? I don't think these have been specified yet. I see a [9-month old proposal](https://github.com/dart-lang/language/blob/master/working/enhanced-type-promotion/feature-specification.md) from you and Bob. Should I look at this to see how it would interact?

> Are there any incompatibilities between this proposal and NNBD?

I don't think so; not as NNBD is written today, since the type promotion stuff isn't in the spec yet. If, e.g. top-level return types are to be inferred under NNBD, then we could relax that inference failure.

working/0508 - strict-inference/strict-inference.md
```diff
@@ -0,0 +1,362 @@
+# Strict inference static analysis option
```
leafpetersen: I don't think working/ is the right sub-directory for this long term.  The usual model is that working is for us to noodle and iterate in, then things moved to accepted/ and then from there to 2.x/ .  If that's what you have in mind for this, that's fine, but I think it probably makes most sense to add this to resources/type-system .  WDYT?

working/0508 - strict-inference/strict-inference.md
```diff
@@ -0,0 +1,362 @@
...
+```yaml
+analyzer:
+  langauge:
+    strict-inference: true
+```
+
+## Motivation
+
+It is possible to write Dart code that passes all static analysis and
+compile-time checks that is guaranteed to result in runtime errors. Commone
```
leafpetersen: Commone -> Common

working/0508 - strict-inference/strict-inference.md
```diff
@@ -0,0 +1,362 @@
...
+examples include runtime type errors, and no-such-method errors. New and
+experienced Dart developers alike write such code, and are surprised to see
+such errors at runtime, which look like they should be caught at compile time.
+
+The strict inference mode aims to highlight such code during static analysis.
+We can look at two common examples:
+
+```dart
```
leafpetersen: This example doesn't work.  Downwards inference does fill in the type for [] in this example.

working/0508 - strict-inference/strict-inference.md
```diff
@@ -0,0 +1,362 @@
...
+## Motivation
+
+It is possible to write Dart code that passes all static analysis and
+compile-time checks that is guaranteed to result in runtime errors. Commone
+examples include runtime type errors, and no-such-method errors. New and
+experienced Dart developers alike write such code, and are surprised to see
+such errors at runtime, which look like they should be caught at compile time.
+
+The strict inference mode aims to highlight such code during static analysis.
+We can look at two common examples:
```
leafpetersen: I'd suggest maybe phrasing this as:

Consider two motivating examples.

### Example: Under-constrained list literals.
....
### Example: Under-constrained generic method invocations.

working/0508 - strict-inference/strict-inference.md
```diff
@@ -0,0 +1,362 @@
...
+`fn`, because surely static analysis would have reported if this wasn't so. In
+real world cases of this code, `fn` and `main` may be authored by different
+developers, and may be located in separate packages. `args` may have been
+created in a third location, by a third developer, with a complex generic type.
+Static analysis is supposed to help developers catch where they have
+misunderstood an API or the type of an object they are handling.
+
+The developer might think that the type of `[]` is unimportant, or that it is
+inferred from the `numbers` parameter type, or that it is inferred from the
+expression to the left of `:`. None of these are true. The type of `[]` is
```
leafpetersen: This isn't true here, there is a downwards context which fills in the type of [].

working/0508 - strict-inference/strict-inference.md
```diff
@@ -0,0 +1,362 @@
...
+}
+```
+
+There are no compile-time, static analysis errors in this code, but it is
+guaranteed to produce a failure at runtime, when `true + 1` is executed. The
+issue here is similar to the previous one: the developer likely thinks that the
+type of `s` (`T`) will be inferred from the type of `initialValue`, and that
+static analysis would report any issue with that type. But inference doesn't
+flow between parameters like that.  Instead, while trying to infer `T`, there
+is not enough information from downwards inference (the type of `a`), and there
```
leafpetersen: This isn't quite an accurate description of what happens.  What happens is that the downwards context provides no information.  Upwards inference constrains `T` to be a super type of `bool` from the first argument.  For the second argument, inference assumes that `s` and `x` have type `dynamic`, and hence infers `dynamic Function(dynamic, int)` as the type of the function literal.  This in turn constraints `T` to be a supertype of `dynamic`.  The LUB of `bool` and `dynamic` is `dynamic`, and so that is what is chosen for `T`.

working/0508 - strict-inference/strict-inference.md
```diff
@@ -0,0 +1,362 @@
...
+The issue would be revealed with either an explicit type on `fold`, an explicit
+type for `a`, which would help to infer the type of `s`, or an explicit type
+for `s`:
+
+```dart
+void main() {
+  // Each of these produce an existing error:
+  // "The operator '+' isn't defined for the class 'bool'."
+  var b = [1, 2, 3].fold<bool>(true, (s, x) => s + x);
+  bool c = [1, 2, 3].fold(true, (s, x) => s + x);
```
leafpetersen: Also by:

var b = [1, 2, 3].fold(true, (int s, int x) => s + x);

working/0508 - strict-inference/strict-inference.md
```diff
@@ -0,0 +1,362 @@
...
+class E extends C {
+  @override
+  void f2(var a) => print(a); // OK
+}
+
+void fA(String cb(var a)) => print(cb(7)); // Inference failure
+void fB(String cb(int x)) => print(cb(7)); // OK
+
+// Typedef parameters cannot be specified with `var`.
+typedef Callback = void Function(int a); // OK
```
leafpetersen: Consider `void Function(int)` instead?

working/0508 - strict-inference/strict-inference.md
srawlins: Done.

PR #521, opened 2019-08-13
Add spec for strict-raw-types

### LINKED ISSUE #516: Introduce a strict-raw-types static inference mode.
The goal is to catch more error-prone code during static analysis; for example, code which is guaranteed to error at runtime. This often two secondary effects: (1) reducing errors in developer understanding of such code, and (2) reducing dynamic dispatch in code where a developer did not think dynamic dispatch was occurring.

We define a "raw type" as a type with omitted type argument(s). Dart fills in such types with their bounds, or dynamic if there are no bounds. Here are some examples:

```dart
void main() {
  List a = [1, 2, 3];
  a.forEach((n) => print(n.isEven));
  fn(a);
}

void fn(List strings) => strings.forEach((s) => print(s.toUpperCase()));
```

On the first line of `main`, `a` is declared as a `List<dynamic>`. The type of `[1, 2, 3]` is inferred from `List<dynamic>`. There is no implicit cast. On the second line, `n.isEven` is a dynamic dispatch call, since the elements of `a` are treated as dynamic. On the third line, when `fn` is called, the runtime will throw a NoSuchMethodError, trying to call `toUpperCase` on the numbers. There are currently no compile-time, static analysis checks that would reveal a problem.

The proposed "strict-raw-types" mode will report errors (e.g. Hints in the analyzer) at the declaration of _most_ types with omitted type arguments. Here are some examples of "raw types" that would be reported:

* A variable declaration, like `List a = [1, 2, 3];`
* A type argument, like `Future<List> a;`
* A function parameter, like `main(List args) {}`, `typedef cb = void Function(Future)`
* A function return type, like `Future main() {}`, `typedef  cb = Future Function()`
* A class declaration, like  `class MyList extends List {}`, `mixin Foo on List`

There are some cases where a type will not be considered a "raw type":

* Omitted type arguments on types which are annotated with `@optionalTypeArgs` from the meta package. There are some classes which are often used in ways where type arguments are not known statically, or they are found in heterogeneous collections. If such a class is annotated with `@optionalTypeArgs`, then it will not be reported as a "raw type."
* A type on the right side of an `is` or `as` expression; in experiments it was found that this just forced developers to write `is List<dynamic>` and `as Map<dynamic, dynamic>` over and over and over. This increases verbosity, without improving correctness or understandability.

Raw types are currently one type of static analysis report found with the "implicit-dynamic: false" static analysis mode. However, multiple attempts to enable this mode on large code bases have shown it to be too onerous to comply with; "strict raw types" are just one piece of "inplicit-dynamic: false", and should be must easier to comply with.

This issue is a continuation of dart-lang/sdk#33749. Thanks to @munificent, @leafpetersen, and @matanlurey for specifying most of the above.

srawlins:
Fixes #516

### DIFF
--- a/working/0516 - strict-raw-types/strict-raw-types.md
+++ b/working/0516 - strict-raw-types/strict-raw-types.md
@@ -1,0 +1,125 @@
+# Strict raw types static analysis option
+
+This document specifies the "Strict raw types" mode enabled with a static
+analysis option. As a static analysis option, we only intend to implement this
+feature in the Dart Analyzer. Under this feature, a type with omitted type
+argument(s) is defined as a "raw type." Dart fills in such type arguments with
+their bounds, or `dynamic` if there are no bounds.
+
+## Enabling strict raw types
+
+ To enable strict raw types, set the `strict-raw-types` option to `true`, under
+ the Analyzer's `language` section:
+
+ ```yaml
+analyzer:
+  langauge:
+    strict-raw-types: true
+```
+
+## Motivation
+
+It is possible to write Dart code that passes all static type analysis and
+compile-time checks that is guaranteed to result in runtime errors. Common
+examples include runtime type errors, and no-such-method errors. Developers are
+often surprised to see such errors at runtime, which look like they should be
+caught at compile time.
+
+The strict raw types mode aims to highlight such code during static analysis.
+We can look at some common examples:
+
+```dart
+void main() {
+  List a = [1, 2, 3];
+}
+```
+
+Developers often think that inference fills in the type of `a` from the right
+side of the assignment. It may look like `a` has the type `List<int>`. But Dart
+fills in omitted type arguments, like `E` on `List`, with `dynamic` (or the
+corresponding type parameter's bound); `List a;` is purely a shorthand for
+`List<dynamic> a;`. Inference then flows from `a` onto the expression on the
+right side of the assignment. This is more obvious in another example:
+
+```dart
+void main() {
+  List a = [1, 2, 3]..forEach((e) => print(e.length));
+  var b = [4, 5, 6]..forEach((e) => print(e.length));
+}
+```
+
+The first statement does not result in any static analysis errors, since the
+type of the list is inferred to be `List<dynamic>`. Instead, the code results
+in a runtime no-such-method error, when the `length` getter is called on an
+`int`.
+
+The second statement, however, allows the type of the list to be inferred from
+its elements, as `List<int>`, which results in a static analysis type error,
+which notes that the getter `length` is not defined on `int`.
+
+Raw types can also lead to unintended dynamic dispatch:
+
+```dart
+void main() {
+  List a = [1, 2, 3];
+  a.forEach((e) => print(e.isEven));
+}
+```
+
+The devloper likely does not realize that the parameter `e` of the callback is
+`dynamic`, and that the call to `isEven` is a dynamic dispatch.
+
+Reporting strict raw types encourages developers to fill in omitted type
+arguments, hopefully with something other than `dynamic`. In cases where the
+only good type is `dynamic`, then including it as an explicit type argument
+avoids the raw type.
+
+## Conditions for a raw type Hint
+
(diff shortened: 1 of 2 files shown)


working/0516 - strict-raw-types/strict-raw-types.md
```diff
@@ -0,0 +1,125 @@
+# Strict raw types static analysis option
```
leafpetersen: As per comments in the strict inference review, consider moving this to resources/type-system?  Or alternatively add a resources/strict-checks/ directory?

working/0516 - strict-raw-types/strict-raw-types.md
```diff
@@ -0,0 +1,125 @@
...
+their bounds, or `dynamic` if there are no bounds.
+
+## Enabling strict raw types
+
+ To enable strict raw types, set the `strict-raw-types` option to `true`, under
+ the Analyzer's `language` section:
+
+ ```yaml
+analyzer:
+  langauge:
```
leafpetersen: should be language

working/0516 - strict-raw-types/strict-raw-types.md
```diff
@@ -0,0 +1,125 @@
...
+This document specifies the "Strict raw types" mode enabled with a static
+analysis option. As a static analysis option, we only intend to implement this
+feature in the Dart Analyzer. Under this feature, a type with omitted type
+argument(s) is defined as a "raw type." Dart fills in such type arguments with
+their bounds, or `dynamic` if there are no bounds.
+
+## Enabling strict raw types
+
+ To enable strict raw types, set the `strict-raw-types` option to `true`, under
+ the Analyzer's `language` section:
```
leafpetersen: Is there a command line option version of this flag?

working/0516 - strict-raw-types/strict-raw-types.md
```diff
@@ -0,0 +1,125 @@
...
+Raw types can also lead to unintended dynamic dispatch:
+
+```dart
+void main() {
+  List a = [1, 2, 3];
+  a.forEach((e) => print(e.isEven));
+}
+```
+
+The devloper likely does not realize that the parameter `e` of the callback is
```
leafpetersen: devloper => developer

working/0516 - strict-raw-types/strict-raw-types.md
```diff
@@ -0,0 +1,125 @@
...
+`dynamic`, and that the call to `isEven` is a dynamic dispatch.
+
+Reporting strict raw types encourages developers to fill in omitted type
+arguments, hopefully with something other than `dynamic`. In cases where the
+only good type is `dynamic`, then including it as an explicit type argument
+avoids the raw type.
```
leafpetersen: Perhaps: "avoids the raw type, and makes the dynamic behavior more explicit in the code."

working/0516 - strict-raw-types/strict-raw-types.md
srawlins: Done.

working/0516 - strict-raw-types/strict-raw-types.md
srawlins: No, there isn't presently. I can't find the review comment, but I think @bwilkerson recommended just wiring an analysis option, because we're primarily dogfooding this internally, where it will exclusively be wired via an analysis options file.

 I don't think there was an intention to never add a command-line flag, it was just simpler not to for now.

working/0516 - strict-raw-types/strict-raw-types.md
srawlins: Done.

working/0516 - strict-raw-types/strict-raw-types.md
srawlins: Done.

working/0516 - strict-raw-types/strict-raw-types.md
srawlins: I like it. Done.

PR #537, opened 2019-08-22
Restructure and make things more explicit.

lrhn:
Restructure so explicit invocation is defined first, implicit invocation desugars (semantically) to an explicit invocation.
Make member invocation definition more explicit.
Introduce "corresponding class".

This should not *change* any behavior, only make it more explicit.
@bwilkerson @mraleph @johnniwinther


accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -46,214 +46,205 @@ extension MyFancyList<T> on List<T> {
...
-```
+If `E` is the single most specific accessible and applicable extension for a member invocation *i* with target expression `e`, which would then be a compile-time error by itself,  then we treat the target expression as if it was the extension application of `E` to `e`, and if `E` is generic, also providing the type arguments inferred for `E` in checking that it was applicable. This makes the member invocation equivalent to an explicit extension member invocation. This happens even if the *name* `E` is not accessible, so this is not a syntactic rewrite.
 
-Here the inference occurs just as if the extension had been declared like:
+Implicit extension member invocation can also apply to individual *cascade* invocations. A cascade is treated as if each cascade section was a separate member invocation on an expression with the same value as the cascade receiver expression (the one before the first `..`). This means that a cascade like `o..foo()..bar()` may perform an implicit extension member invocation on `o` for `foo()` and a normal invocation on `o` for `bar()`. There is no way to specify the corresponding explicit member invocation, so this is again not a syntactic rewrite.
```
eernstg: Note that cascades are themselves specified in terms of a syntactic rewrite (`evaluate t.suffix to an object`), and the rewrite would give rise to an extension invocation using the rules for normal member access. So there's no need to talk about syntactic rewrites here.

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -46,214 +46,205 @@ extension MyFancyList<T> on List<T> {
...
-class SuperList<T> {
-  final List<T> $this;
-  SuperList(this.$this);
-  R foldRight<R>(R base, R Function(T, R) combine) { ... }
-}
-```
+##### Accessibility
 
-and it was invoked as `SuperList(strings).foldRight(0, count)`.
+An extension is *accessible* for an expression if it has a name which is not private to a different library and the extension declaration is declared or imported into a scope which is a (recursive) parent scope of the lexical scope of the expression. That currently means imported into the default import scope of the current library, or declared at the top-level of the library itself.
```
eernstg: Even mentioning that the extension has a name which is private to another library breaks the rules for namespace management: Otherwise, we never need to even consider what roles such a name could play.

So the cost associated with this kind of language is that we must start from scratch, leaving all scope rules etc. in the dust.

A fresh name of the extension would allow us to rely on standard namespace management, which would also allow us to connect to other parts of the language, e.g., when an explicitly resolved extension method invocation is inferred and checked as-if via a corresponding class. Otherwise that class would also have a name that "doesn't work" according to the standard scope-and-namespace rules.

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -46,214 +46,205 @@ extension MyFancyList<T> on List<T> {
...
 
-Or, alternatively, like the extension method had been declared like:
+You can *avoid* making the extension in-scope for a library by either not importing any library exporting the extension, importing such a library and hiding the extension using `hide` or `show`, or importing such a library only with a prefix.
 
-```dart
-R Function<R>(R, R Function(T, R)) SuperList$foldRight<T>(List<T> $this) =>
-    <R>(R base, R Function(T, R) combine) { ... };
-```
+An extension *is* in scope if the name is *shadowed* by another declaration (a class or local variable with the same name shadowing a top-level or imported declaration, a top-level declaration shadowing an imported extension, or a non-platform import shadowing a platform import).
```
eernstg: This assigns a completely new meaning to the phrase `in scope`. We will then need to revisit all locations where that phrase is used (the spec has 206 occurrences), and make sure that they still mean the right thing, or that they use a different phrase. Maybe use a different phrase for the new concept?

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -46,214 +46,205 @@ extension MyFancyList<T> on List<T> {
...
 
-and it was invoked as `SuperList$foldRight(strings)(0, count)`.
+An extension *is* in scope if it is imported, and the extension name conflicts with one or more other imported declarations.
```
eernstg: Same issue as line 210.

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -46,214 +46,205 @@ extension MyFancyList<T> on List<T> {
...
+
+- The invocation is an *instance* member invocation. That is the case if the expression `e` does not denote a prefix or a class, mixin or extension declaration *(then the member invocation would be a static invocation)*. An instance member invocation on `e` will always begin by evaluating `e` to an object, and then continue by performing an instance member invocation on that object.
+- The type *S*  does not have a member with the basename  *m*. For this, the types `dynamic`and `Never` are considered as having all member names, and an expression of type `void` cannot occur as the target of a member invocation, so none of these can ever have applicable extensions. Function types and the type `Function` are considered as having a `call` member. *This ensure that if there is an applicable extension, the existing invocation would otherwise be a compile-time error*. Members of `Object` exists on all types, so they can never be the target of implicit member invocations (they can also not be declared as extension members).
```
eernstg: The NNBD feature spec makes member accesses on `Never` an error, so it should not be necessary to say that it has all members.

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -46,214 +46,205 @@ extension MyFancyList<T> on List<T> {
...
+- The extension application `E(x)` would be valid (not a compile-time error) where `x` is a fresh variable with static type *S* (to avoid type inference for any type parameters of `E` from affecting the already determined static type of `e`).
```
eernstg: Can't use `E(x)` if `E` may not be in scope (according to the current meaning of that phrase).

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -46,214 +46,205 @@ extension MyFancyList<T> on List<T> {
...
 
-1. The latter extension is declared in a platform library, and the former extension is not, or
-2. they are both declared in platform libraries or both declared in non-platform libraries, and
-3. the instantiated type (the type after applying type inference from the receiver) of *T*<sub>1</sub> is a subtype of the instantiated type of *T*<sub>2</sub> and either
+1. The `E2` extension is declared in a platform library and the `E1` extension is not, or
+2. either both or neither are declared in platform libraries and
+3. *T<sub>1</sub>* is a subtype of of *T<sub>2</sub>* and either
 4. not vice versa, or
-5. the instantiate-to-bounds type of *T*<sub>1</sub> is a subtype of the instantiate-to-bounds type of *T*<sub>2</sub> and not vice versa.
+5. the instantiate-to-bounds `on` type of `E1` is a subtype of the instantiate-to-bounds `on` type of `E2` and not vice versa.
```
eernstg: Instantiation to bounds is not defined if `E1`, `E2` is not in scope.

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -437,17 +391,17 @@ int Function(int) func = (int x) => Foo(b).baz<int>(x);
 
 *Torn off extension methods are never equal unless they are identical*. Unlike instance methods, which are equal if it's the same method torn off from the same object (unless it's an instantiated tear-off of a generic function), torn off extension methods may close over the type variables of the extension as well. To avoid distinction between generic and non-generic extensions, no two torn off extension methods are equal, even if they are torn off from the same extension on the same object at the same static type.
 
-Extension methods torn off *constant* receiver expressions are not constant expressions. They also create a new function object each time the tear-off expression is evaluated.
+Extension methods torn off a *constant* receiver expressions are not constant expressions. They also create a new function object each time the tear-off expression is evaluated.
```
eernstg: Typo: `expression[s]`. Might read better as `An extension method torn off ...` in order to stay in singular everywhere.

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -437,17 +391,17 @@ int Function(int) func = (int x) => Foo(b).baz<int>(x);
...
 
-An explicitly overridden extension method access, like `Foo<Bar>(b).baz`, also works as a tear-off. 
+An explicitly extension method invocation like `Foo<Bar>(b).baz`, also creates a tear-off.
```
eernstg: Typo: `explicit[ly]`, or maybe `explicitly resolved`.

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -462,7 +416,7 @@ extension Tricky on int {
...
+In detail: Any expression of the form `e1(args)` or `e1<types>(args)` where `e1` does not denote a method, and where the static type of `e1` is not a function type, an interface type declaring a `call` method, or `dynamic,` will currently be a compile-time error. If the static type of `e1` is an interface type declaring a `call` *getter* or a `call=` *setter*, then this stays a compile-time error. Otherwise we check for extensions applying to the static type of `e1` and declaring a `call` member. If one such most specific extension exists, and it declares a `call` extension method, then the expression is equivalent to `e1.call(args)` or `e1.call<typeS>(args)`. Otherwise it is still a compile-time error.
```
eernstg: I think 'and it declares a `call` extension method' is redundant here.

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -46,214 +46,205 @@ extension MyFancyList<T> on List<T> {
...
 ```
 
-More precisely, an extension declaration is a top-level declaration with a grammar similar to:
+More precisely, an extension declaration is a declaration with a grammar similar to:
 
 ```ebnf
-<extension> ::= 
-  `extension' <identifier>? <typeParameters>? `on' <type> `{'
-     memberDeclaration*
+<extensionDefinition> ::=
```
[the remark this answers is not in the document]
lrhn: Done.

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -46,214 +46,205 @@ extension MyFancyList<T> on List<T> {
...
+- Declares a constructor.
+- Declares an instance variable.
+- Declares an abstract member.
+- Declares a member with a formal parameter marked `covariant`.
 
-If an extension declaration is only used locally in a library, there might be no need to worry about naming conflicts or overrides. In that case, then name identifier can be omitted (hence the `<identifier>?` in the grammar above).
+(The *basename* of a declaration is the declared name of the declaration for variable, method, getter and most operator declarations, and it's the declared name without the trailing `=` for setter declarations and the `[]=` operator.)
 
-Example:
+The member declarations can be any non-abstract static or instance member declaration (like the ones allowed in `class` or `mixin` declarations) except for instance variables and constructors, and that they cannot have the same *base name* as a member declared by `Object` and cannot have `covariant` parameters. The member declarations themselves must otherwise satisfy the same requirements as member declarations of class and mixin declarations (no parameter name conflicts, no type parameter bound cycles, etc.)
```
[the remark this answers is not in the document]
lrhn: Gone.

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -46,214 +46,205 @@ extension MyFancyList<T> on List<T> {
...
 
-```dart
-extension<T> on List<T> {
-  void quadruple() { ... }
-}
-```
+Abstract members are not allowed since the extension declaration does not introduce an interface, and constructors are not allowed because the extension declaration doesn't introduce any type that can be constructed. Instance variables are not allowed because there won't be any memory allocation per instance that the extension applies to. We could implement instance variables using an `Expando`, but it would necessarily be nullable, so it would still not be an actual instance variable. Users who want that functionality can still add it manually using getter/setter declarations. Members with the same base name as members of `Object` are not allowed because some of them are accessed directly by the language semantics, and it is potentially confusing and error-prone if extension members could have the same name and a wildly different signature.
```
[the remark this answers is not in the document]
lrhn: Done.

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -46,214 +46,205 @@ extension MyFancyList<T> on List<T> {
...
 
-This is equivalent to giving the extension a fresh private name.
+An extension declaration with a non-private name is included in the library's export scope, and a privately named extension is not. It is a compile-time error to export two declarations, including extensions, with the same name, whether they come from declarations in the library itself or from export declarations (with the usual exception when all but one declaration come from platform libraries). Extension *members* with private names are simply inaccessible in other libraries.
+
+We make `extension` a built-in identifier. Is not necessary for disambiguation, but it makes error-recovery in parsers much easier.
 
 If we make `on` a built-in identifier, then there should not be any parsing issue. Even without that, the grammar should be unambiguous because `extension on on on { … }` and `extension on on { … }` are distinguishable, and the final type cannot be empty. It may be *harder* to parse.
 
 This is a simple feature, but with very low impact. It only allows you to omit a single private name for an extension that is only used in a single library.
```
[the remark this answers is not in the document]
lrhn: Done.

accepted/future-releases/static-extension-methods/feature-specification.md
lrhn: Rewritten.

accepted/future-releases/static-extension-methods/feature-specification.md
lrhn: Rewritten to not mention scopes.

accepted/future-releases/static-extension-methods/feature-specification.md
lrhn: No longer uses "in scope".

accepted/future-releases/static-extension-methods/feature-specification.md
lrhn: Ditto.

accepted/future-releases/static-extension-methods/feature-specification.md
lrhn: Moved it to be with `void` instead.

accepted/future-releases/static-extension-methods/feature-specification.md
lrhn: added "in a scope where E denotes the extension". Then it does work.

accepted/future-releases/static-extension-methods/feature-specification.md
lrhn: `E1` denotes the extension, not the name of the extension, and instantiate-to-bounds type is defined on the extension declaration itself, so yes it is defined :)

accepted/future-releases/static-extension-methods/feature-specification.md
lrhn: Made singularly correct.

accepted/future-releases/static-extension-methods/feature-specification.md
lrhn: rewritten.

accepted/future-releases/static-extension-methods/feature-specification.md
lrhn: It's not redundant because the extension is applicable if it declares an extension *member* named `call`, but the call is only valid if it declares an extension *method* named `call`.
If the most specific extension has a `call` getter, then it is still a compile-time error.

I have emphasized "*method*".

lrhn:
Addressed comments, PTAL.

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -46,214 +46,205 @@ extension MyFancyList<T> on List<T> {
...
-```dart
-class MyList<T extends Comparable<T>> {
-  MyList(List<T> argument);
-  void quickSort() { ... }
-}
-...
-List<Duration> times = ...;
-... MyList(times).quickSort() ... // Infer type argument of MyList here.
-```
+The static type of a member invocation on an extension application is the return type of the extension member with the corresponding member name of the invocation, with the explicit or inferred type arguments of the extension application replacing the type parameters.
```
leafpetersen: "type parameters bound by the extension, and the explicit or inferred type arguments of the invoked member replacing the type parameters bound by the member".

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -46,214 +46,205 @@ extension MyFancyList<T> on List<T> {
...
 
-### Generic Parameter Inference
+### Implicit Extension Member Invocation
 
-If both the extension and the method is generic, then inference must infer the extension type parameters first, to figure out whether the extension applies, and only then start inferring method type parameters. As mentioned above, the inference is similar to other cases of chained inference. 
+Extension members can be invoked *implicitly* (without mentioning the extension by name) as if they were members of the `on` type of the extension. This is intended as the primary way to use extensions, with explicit extension member invocation as a fallback for cases where the implicit extension resolution doesn't do what the user want. 
 
-Example:
+An implicit extension member invocation occurs for a simple or composite member invocation with a target expression `e` iff there exists a unique *most specific* extension declaration which is *accessible* and *applicable* to the member invocation (see below).
```
leafpetersen: I think this leaves the behavior of composite member invocations underspecified (or possibly incorrectly specified) when the setter comes from the extension and the getter from elsewhere (or vice versa).  e.g.

```dart
class A {
  int get x => 3;
  A operator+(int _) => null;
}

extension E on A {
  void set x(int _) {}

}

void main() {
  var a = new A();
  print(a += 3);
}
```

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -46,214 +46,205 @@ extension MyFancyList<T> on List<T> {
...
-    return base;
-  }
-}
-...
-  List<String> strings = ...;
-  int count(String string, int length) => length + string.length;
-  ...
-  var length = strings.foldRight(0, count);
-```
+If `E` is the single most specific accessible and applicable extension for a member invocation *i* with target expression `e`, which would then be a compile-time error by itself,  then we treat the target expression as if it was the extension application of `E` to `e`, and if `E` is generic, also providing the type arguments inferred for `E` in checking that it was applicable. This makes the member invocation equivalent to an explicit extension member invocation. This happens even if the *name* `E` is not accessible, so this is not a syntactic rewrite.
```
leafpetersen: Again, this doesn't sound quite right for composite invocations with split origins.

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -46,214 +46,205 @@ extension MyFancyList<T> on List<T> {
...
 
-### Extension Conflict Resolution
+- The invocation is an *instance* member invocation. That is the case if the expression `e` does not denote a prefix or a class, mixin or extension declaration *(then the member invocation would be a static invocation)*, and it is not an explicit extension application. An instance member invocation on `e` will always begin by evaluating `e` to an object, and then continue by performing an instance member invocation on that object.
+- The type *S*  does not have a member with the basename  *m*. For this, the type `dynamic`is considered as having all member names, and an expression of type `Never` or `void` cannot occur as the target of a member invocation, so none of these can ever have applicable extensions. Function types and the type `Function` are considered as having a `call` member. *This ensure that if there is an applicable extension, the existing invocation would otherwise be a compile-time error*. Members of `Object` exists on all types, so they can never be the target of implicit member invocations _(they can also not be declared as extension members)_.
```
leafpetersen: This seems to imply that an extension setter `x=` will never be applicable if the `on` type has an `x` getter, right?  I think this is a change.  The analyzer and the CFE allow this.

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -673,6 +627,18 @@ The use of `typedef` for something which is not a type may be too confusing. Ano
 extension MyWidgetList<T extends Widget> = prefix.MyList<T>;
 ```
 
+### Avoiding Scope Conflicts
+
+We allow importing two extensions with the same name into the same scope, and both of them *work*. They cannot be referenced by name, but they can be implicitly invoked. This breaks the current invariant that we cannot have two declarations with the same name in the same scope.
+
+To retain this invariant, we can define a "fresh extension alias" for each extension declaration. It is a second *fresh name* for the extension (a name not occurring anywhere else in the program), which will be a fresh *private* name if the extension's name is private (or omitted, in which case we only need the one fresh name), and a fresh non-private name if the extension's name is non-private. Any implicit extension invocation uses the *fresh extension alias* name to refer to the extension (so an `expr.foo()` which resolves to a method on the extension  `Ext` is equivalent to `$fresh$Ext(expr).foo()`).
```
leafpetersen: Why not just use a prefix, instead of inventing a new concept?  Import it once prefixed to get a name for it, and once unprefixed to bring the members into scope?

leafpetersen: If you do use fresh names, best to make them private.  They shouldn't be exported.  See below.

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -673,6 +627,18 @@ The use of `typedef` for something which is not a type may be too confusing. Ano
...
 ```
 
+### Avoiding Scope Conflicts
+
+We allow importing two extensions with the same name into the same scope, and both of them *work*. They cannot be referenced by name, but they can be implicitly invoked. This breaks the current invariant that we cannot have two declarations with the same name in the same scope.
+
+To retain this invariant, we can define a "fresh extension alias" for each extension declaration. It is a second *fresh name* for the extension (a name not occurring anywhere else in the program), which will be a fresh *private* name if the extension's name is private (or omitted, in which case we only need the one fresh name), and a fresh non-private name if the extension's name is non-private. Any implicit extension invocation uses the *fresh extension alias* name to refer to the extension (so an `expr.foo()` which resolves to a method on the extension  `Ext` is equivalent to `$fresh$Ext(expr).foo()`). 
+
+A show/hide modifier on an import or export for the name *N* must also show or hide any fresh alias which is an alias for something named *N*. So, for each fresh name, we need to be able to determine the non-fresh name it corresponds to.
```
leafpetersen: I don't think it makes sense to "show" a fresh name, except if you're thinking in a whole program context.  There is no reasonable scheme whereby I can modularly compile two independent modules and arrange for the two sets of fresh names chosen to be globally fresh.  The "freshness" is local to the compilation unit.  So there's no point in exporting the "fresh" names - they're not guaranteed not to collide anyway.

accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -46,214 +46,205 @@ extension MyFancyList<T> on List<T> {
...
 
-You can *avoid* making the extension in-scope for a library by either not importing any library exporting the extension, importing such a library and hiding the extension using `hide` or `show`, or importing such a library only with a prefix.
+An *extension application* is an expression of the form `E(expr)` or `E<typeArgs>(expr)` where `E` denotes an extension declaration (that is, `E` a simple or qualified identifier which refers to the extension declaration).
 
-An extension *is* in scope if the name is *shadowed* by another declaration (a class or local variable with the same name shadowing a top-level or imported declaration, a top-level declaration shadowing an imported extension, or a non-platform import shadowing a platform import).
+An extension application is subject to static type inference.  If `E` is an extension declared as `extension E<X...> on T {...}`, then the type inference for an extension application is exactly the same as that the type inference performed for the same syntax considered as a constructor invocation on a class declared as:
```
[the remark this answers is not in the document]
lrhn: Does the context type ever affect the receiver expression of `E(o).m`? I was assuming that there is no context type for `E(o)` in that expression because there is no way to pass a context type through an application of something that we don't know what is (yet). Then, after `E(o)` was inferred with no context type, we use the context type and type of `E(o)` to do inference on the `.m` invocation.

accepted/future-releases/static-extension-methods/feature-specification.md
lrhn: What is a "composite invocation with split origins"?

It *should* be correct. If it isn't, then some part of the definitions leading here is insufficient.

Since all extension lookup is by member basename, and the applicability rule uses the basename of the interface to see if any extension is applicable, then we will either find one most specific extension for both read and write, or none at all.

accepted/future-releases/static-extension-methods/feature-specification.md
lrhn: This is not a *change*, it has always been the intent.
The existing specification states:

> For any member access, `x.foo`, `x.bar()`, `x.baz = 42`, `x(42)`, `x[0] = 1` or `x + y`, including `null`-aware and cascade accesses which effectively desugar to one of those direct accesses, and including implicit member accesses on this, the language first checks whether the static type of `x` has a member with the same `base name` as the operation. That is, if it has a corresponding instance member, respectively, a `foo` method or getter or a `foo=` setter. a `bar `member or `bar=` setter, a `baz` member or `baz=` setter, a `call` method, a `[]=` operator or a `+` operator. If so, then the operation is unaffected by extensions.

Notice the use of "base name", and explicitly mentioning both `foo` and `foo=`.

If it is not allowed by tools, that is an error.

The only difference I might have made here is that I treat `[]` and `[]=` as having the same basename. (And I'm not entirely sure about that, but it's probably a good idea to treat `[]=` as a setter in as many ways as possible - every time we didn't, it has come back to bite us).

accepted/future-releases/static-extension-methods/feature-specification.md
lrhn: The goal of the idea in this section is not to solve the reference problem, but to avoid having two elements with the same name in the same scope. I have rewritten other text so that the available extensions are not based on being "in scope" but rather on being among the members of an import declaration, whether that member is actually added to the library scope or not.

This was just a potential other approach, and it might not be necessary.

lrhn: The goal is to make them exported, so they can be used to reference the extension even when the declared name is not in scope.

accepted/future-releases/static-extension-methods/feature-specification.md
lrhn: I'm sure a compiler can find a way (perhaps encode the library URI in the name).

I think I'll just drop this idea again. It was intended as a different view on how scoping works, making extensions accessible if their fresh name was in scope, whether or not their declared name was, but it's dangerous to thread a new feature through the needleholes cut for a different feature.

accepted/future-releases/static-extension-methods/feature-specification.md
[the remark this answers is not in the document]
leafpetersen: For direct property accesses, there is never a context type.  For cascades, there may be.  So if this doesn't apply to cascades, then this is implicit, but I think it does?

accepted/future-releases/static-extension-methods/feature-specification.md
leafpetersen: The CFE allowed this as of 2 pm yesterday, are you sure dart2js and and the VM are correctly rejecting it?

accepted/future-releases/static-extension-methods/feature-specification.md
leafpetersen: > Since all extension lookup is by member basename

This is going to need some clarification and tests.  It sounds like you intend all of the member invocations in this program to be errors:

```dart
class A {
  int get x => 3;
}

extension E on A {
  void set x(int _) {}
  void set y(int _) {}

}

extension E2 on A {
  int get y => 3;
}

void main() {
  var a = new A();
  a.x;
  a.x = 4;
  a.x += 3;
  a.y;
  a.y = 4;
  a.y += 3;
}
```

As of 2pm yesterday, all of these except the last one were accepted by all implementations (the last was rejected by the CFE).

Reading the spec it sounds like, on the other hand, the following should be accepted, since both extensions apply, and one is more specific?

```dart
class B {}
class C extends B {}

extension EB on B {
  int get y => 3;
}

extension EC on C {
  int set y = 3;
}

void test() {
  var b = new C();
  // b.x; should be error?
  b.x = 4;
  // b.x += 3; Should be error?

}
```

accepted/future-releases/static-extension-methods/feature-specification.md
leafpetersen: I don't believe that the compiler can find a way (other than probabilistically, by choosing a large random number, in which case it is impossible for another module to *predict* what name is chosen).  If something external to the compiler (like the programmer) is required to associate a globally unique key with every compilation unit (are library URIs globally unique?  I think they aren't since you can use paths, right?) then yes, you can bootstrap off of that, but the compiler *cannot* produce a globally unique name in a predictable way without an external source of predictable uniqueness.

accepted/future-releases/static-extension-methods/feature-specification.md
leafpetersen: I think prefixing solves this.  No matter what, you need to have a way of associating extension member names (of which there will be lots with the same name in "scope") with extension declarations.  All I'm saying is, don't do that association based on the unqualified name of the extension declaration: instead, do it based on a fully qualified name.  Then you never have to talk about there being two extension declarations with the same name in scope.  You just have things like `p0.E`, `p1.E` and `p1.E2` in scope, and if you see an unqualified reference to `E2` you resolve it to `p1.E2`, but if you see an unqualified reference to `E` you fail to resolve it.  And now you have whatever mechanism you use for resolving extension member names to extension declarations map `foo` to `p0.E.foo` and `bar` to `p1.E.bar`, and everybody is happy.

accepted/future-releases/static-extension-methods/feature-specification.md
leafpetersen: Just validated that as of this morning, the analyzer reports no errors, and the CFE only reports two errors (marked in comments) on the following program:

```dart
class A {
  int get x => 3;
}

extension E on A {
  void set x(int _) {}
  void set y(int _) {}

}

extension E2 on A {
  int get y => 3;
}

class B {}
class C extends B {}

extension EB on B {
  int get y => 3;
}

extension EC on C {
  void set y(int _) {}
}

void test() {
  var b = new C();
  b.y; // Should be an error?
  b.y = 4; // Should be accepted?
  b.y += 3; // Should be error?  CFE yes

}

void main() {
  var a = new A();
  a.x; 
  a.x = 4; // Should be an error?
  print(a.x += 3); // Should be an error?
  a.y; // Should be an error?
  a.y = 4; // Should be an error?
  print(a.y += 3); // Should be an error? CFE yes
}
```

lrhn:
Ping Leaf?

accepted/future-releases/static-extension-methods/feature-specification.md
leafpetersen: I think maybe you were misdiagnosing this? Maybe the CFE implemented extension method resolution at all at the time - and it was telling you that it didn't know anything about a setter named `x`?  In any case, in a build from last week, the CFE emitted no errors for the example you list above.

accepted/future-releases/static-extension-methods/feature-specification.md
leafpetersen: Agreed, this comment was written under the understanding that setters and getters were resolved separately, instead of by basename.

Issue #543, opened 2019-08-24
Operator '[]=' for the class 'Object'


LeandroDoldan:
### Problem
AFAIK, something like the following is impossible to implement:

```
class Updater<T> {
  T entity;

  Updater(this.entity);

  update(String key, dynamic value) {
    entity[key] = value;
  }
}
```

Because the following error is thrown:
```
The operator '[]=' isn't defined for the class 'Object'.
Try defining the operator '[]='.
```

### Request
Please, implement the operator '[]=' for the class 'Object'.

Regarding typing, this is a simple TypeScript-like suggestion:

```
class Updater<T> {
  T entity;

  Updater(this.entity);

  update<K keyof T>(K key, dynamic value) {
    entity[key] = value;
  }
}
```

Thank you!

lrhn:
The `Object` class does not have any `[]` operator. I am assuming you want a behavior similar to JavaScript, but Dart is not JavaScript. You cannot look up members using `[]` and a string name `o["foo"]`, you have to write `o.foo`.
This is a deliberate choice, and if you want to access a member by a name that is only known as a string at runtime, that is reflection and you need to use `dart:mirrors` for it (on a platform supporting `dart:mirrors`).

There are no plans to allow reflection directly on `Object`.

Aside from that, it would break every other class which currently extend `Object` and implement `operator[]` (like `Map`).

rakudrama:
@LeandroDoldan Something like this might work for you:

```dart
// T is the type of things held, instead of the type of the collection.
class Updater<T> {
  Map<String, T> entity;
  Updater(this.entity);
  update(String key, T value) {
    entity[key] = value;
  }
}
```

You could write dynamic code that is against the spirit of Dart. I would advise against this - dynamic invocation is often less efficient and you get no safety or code completion benefits that you would get with stronger typing.
```dart
class Updater<T> {
  T entity;
  Updater(this.entity);
  update(String key, T value) {
    (entity as dynamic)[key] = value;
  }
}
```

I do think it is an interesting idea to 'destructure' the type. If Dart had that feature it might look like

```dart
class Updater<T extends Map<K, V>> {
  T entity;
  Updater(this.entity);
  update(K key, V value) {
    entity[key] = value;
  }
}
```

@lrhn Sometimes it is awkward to 'pass around' several types as type parameters, perhaps the language could support that better.

eernstg:
@rakudrama, just FYI, the 'destructure the type' feature is at the core of the type patterns proposal (#170).

[...]

PR #548, opened 2019-08-27
Bundle getter/setter names for lexical lookups

eernstg:
This PR updates the language specification such that it specifies "bundling" of getters and setters for all parts of Dart except imports and exports, that is, whenever an identifier is looked up in the enclosing lexical scopes.

It defines a new concept of 'lexical lookup' which will look for `id` as well as `id=` and select the lexically nearest declaration with one of those names, and make it an error if we are looking up `id` and the selected scope has an `id=` but no `id`, and vice versa; or if we look up `id`/`id=` and nothing is in scope, conclude that we must use an instance member, and the class interface doesn't have the required getter-or-setter, and so on. It's a substantial rewrite, clarification, and bug fix.

In any case, the newly specified approach is what we already did/intended for 'Identifier Reference'.

It is also what the CFE already does for 'Unqualified Invocation'. (In this case, the spec required a simple search for `id` and ignored conflicts with `id=` in intermediate scopes, but the new text and the CFE behavior is more consistent with the rest of the language.)

For assignments, we used to spec a nearly consistent search for `id` as well as `id=`, and tools are actually slightly more consistent (in their bundling of getters and setters, as specified now).

All in all, this should not be a breaking change in any other way than for unqualified invocation.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -1922,6 +1922,22 @@ The enclosing scope of an instance member declaration is the instance scope of t
 The enclosing scope of a static member declaration is the static scope of the class in which it is declared.
 
 \LMHash{}%
+The current instance
+(\commentary{and hence its members})
+can only be accessed at specific locations in a class:
+We say that a location $\ell$
+\IndexCustom{has access to \THIS{}}{has access to this@has access to \THIS{}}
+if{}f $\ell$ is inside the body of a declaration of
+an instance member or a generative constructor,
+or in the initializing expression of a \LATE{} instance variable declaration.
+
+\commentary{%
+Note that an initializing expression for a non-\LATE{} instance variable
+does not have access to \THIS,
+and neither does any part of a declaration marked \STATIC.%
+}
+
+\LMHash{}%
 Every class has a single superclass except class \code{Object} which has no superclass.
 A class may implement a number of interfaces by declaring them in its implements clause (\ref{superinterfaces}).
 
@@ -2030,8 +2046,8 @@ a member with basename (\ref{classMemberConflicts}) $C$.
 If a generic class named $G$ declares a type variable named $X$,
 it is a compile-time error
 if $X$ is equal to $G$,
-if $G$ has a member whose basename is $X$,
-and if $G$ has a constructor named \code{$G$.$X$}.
+or if $G$ has a member whose basename is $X$,
+or if $G$ has a constructor named \code{$G$.$X$}.
 
 
 \subsection{Instance Methods}
@@ -2110,7 +2126,8 @@ in a trade-off where some static type safety is lost.
 \LMLabel{operators}
 
 \LMHash{}%
-\IndexCustom{Operators}{operators} are instance methods with special names.
+\IndexCustom{Operators}{operators} are instance methods with special names,
+except for operator \lit{[]=} which is an instance setter.
 
 \begin{grammar}
 <operatorSignature> ::= \gnewline{}
@@ -3823,6 +3840,8 @@ This section specifies these errors.
 \LMHash{}%
 The \Index{basename} of a getter or method named $n$ is $n$;
 the basename of a setter named \code{$n$=} is $n$.
+The basename of an operator named $n$ is $n$,
+except for operator \code{[]=} whose basename is \code{[]}.
 
 \LMHash{}%
 Let $C$ be a class.
@@ -10574,8 +10593,8 @@ Perform a lexical lookup of \code{\id=} from the location of \id.
   \code{\THIS.\id{} = $e$}.
 
   \commentary{%
-    In this case it is known that $a$ occurs in
-    the body of an instance member or a generative constructor,
+    In this case it is known that $a$ has access to \THIS{}
+    (\ref{classes}),
     and the interface of the enclosing class has a member named \code{\id=}.
     Both the static analysis and evaluation proceeds with
     \code{\THIS.\id{} = $e$},
@@ -11770,8 +11789,11 @@ Evaluation of an assignable expression of the form \code{\SUPER{}[$e_2$]} is equ
 This section specifies how to look up a name
 based on the enclosing lexical scopes.
 This is known as a \Index{lexical lookup}.
-When \id{} is an identifier, it may look up a name $n$ of the form \id{}
+When \id{} is an identifier or \code{[]},
+it may look up a name $n$ of the form \id{}
 as well as of the form \code{\id=}.
+When \id{} is an operator name other than \code{[]},
+it will look up \id{}.
 
 \LMHash{}%
 A lexical lookup yields a result which is
@@ -11793,9 +11815,9 @@ as detailed below.%
 
 \LMHash{}%
 Consider the situation where a name $n$ has basename \id{}
-(\ref{classMemberConflicts}),
-an identifier \id{} occurs at a location $\ell$,
-and a lexical lookup of $n$ is performed from $\ell$.
+(\ref{classMemberConflicts})
+where \id{} is an identifier or an operator name,
+and a lexical lookup of $n$ is performed from a given location $\ell$.
 
 \commentary{%
 We specify a name and a location from where a lexical lookup is performed.
@@ -11808,7 +11830,8 @@ and the location that determines which scopes are the enclosing ones.%
 }
 
 \LMHash{}%
-When we consider an occurrence of an identifier \id{} and say that
+When we consider an occurrence of
+an identifier or operator name \id{} and say that
 a lexical lookup of \id{} is performed,
 it is understood that the lookup is performed from
 the location of said occurrence of \id.
@@ -11826,8 +11849,9 @@ let $D$ be that declaration.
 \commentary{%
 A non-local variable declaration named \id{} will implicitly induce
 a getter \id{} and possibly a setter \code{\id=} into the enclosing scope.
-This means that $D$ may denote an implicitly induced getter or setter,
-which is significant in the case where an error must arise
+This means that $D$ may denote an implicitly induced getter or setter
+rather than the underlying variable declaration.
+That is significant in the case where an error must arise
 because the lookup was for one kind, but only the other kind exists.%
 }
 
@@ -11854,16 +11878,10 @@ the following errors apply:
   % - inside an instance variable initializer;
   % - in an expression in the initializer list of a constructor
   %   (be it a superinitializer, a field initializer, or an assertion).
-  It is a compile-time error if $\ell$ is not inside
-  %% TODO(eernst): Come NNBD, adjust text for `late` variable initializers.
-  the body of an instance member or a generative constructor.
-  \commentary{%
-    Note that an initializing expression for an instance variable
-    is not inside such a body.%
-  }
+  It is a compile-time error if $\ell$ does not have access to \THIS{}
+  (\ref{classes}).
 \item
-  When $\ell$ is inside the body of
-  an instance member or a generative constructor,
+  When $\ell$ has access to \THIS{}
   it is a compile-time error if the interface of the enclosing class
   does not have a member named $n$.
   \commentary{%
@@ -11890,20 +11908,14 @@ unless $D$ is an instance member or a local variable
 %% TODO(eernst): Come NNBD, `this` is accessible in a `late` instance variable
 %% initializer, so they must also be included in the first case.
 \LMHash{}%
-% Case: `this` is accessible.
-When $\ell$ occurs in
-the body of an instance member or a generative constructor,
+When $\ell$ has access to \THIS{}
+(\ref{classes}),
 it is a compile-time error if $D$ is an instance member,
 and the interface of the enclosing class does not have a member named $n$.
-% Case: `this` is not accessible.
-Otherwise, it is a compile-time error if $D$ is an instance member.
+When $\ell$ does not have access to \THIS,
+it is a compile-time error if $D$ is an instance member.
 % An instance variable initializer is in the instance scope, so it is
 % actually possible for $D$ to be an instance member here.
-
-\commentary{%
-Note that an initializing expression for an instance variable
-is not inside such a body.%
-}
 \EndCase
 
 \rationale{%
@@ -12176,8 +12188,8 @@ In this situation $e$ is treated as
 \code{\THIS.\id}.
 
 \commentary{%
-In this case it is known that $e$ occurs in
-the body of an instance member or a generative constructor,
+In this case it is known that $e$ has access to \THIS{}
+(\ref{classes}),
 and the interface of the enclosing class has a member named \id.
 Both the static analysis and evaluation proceeds with
 \code{\THIS.\id},
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -11874,9 +11874,8 @@ In this case, at least one declaration with basename \id{}
 is in scope at the location $\ell$.
 
 \LMHash{}%
-It is a compile-time error if $D$ is not an instance member
-and the name of $D$ is not $n$,
-except when $n$ is \code{\id=} and $D$ declares a local variable
+It is a compile-time error if the name of $D$ is not $n$,
+unless $D$ is an instance member or a local variable
 (\commentary{which may be a formal parameter}).
 
 %% TODO(eernst): Come NNBD, `this` is accessible in a `late` instance variable
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -11874,13 +11874,16 @@ In this case, at least one declaration with basename \id{}
 is in scope at the location $\ell$.
 
 \LMHash{}%
+It is a compile-time error if $D$ is not an instance member
+and the name of $D$ is not $n$,
+except when $n$ is \code{\id=} and $D$ declares a local variable
+(\commentary{which may be a formal parameter}).
+
+\LMHash{}%
 If $\ell$ does not occur in
 %% TODO(eernst): Come NNBD, update text for `late` variable initializers.
 the body of an instance member or a generative constructor,
-it is a compile-time error if the name of $D$ is not $n$,
-except when $n$ is \code{\id=} and $D$ declares a local variable
-(\commentary{which may be a formal parameter});
-and it is a compile-time error if $D$ is an instance member.
+it is a compile-time error if $D$ is an instance member.
 % An instance variable initializer is in the instance scope, so it is
 % actually possible for $D$ to be an instance member here.
 
@@ -11892,10 +11895,7 @@ is not inside such a body.%
 \LMHash{}%
 Otherwise $\ell$ occurs in
 the body of an instance member or a generative constructor.
-In this case it is a compile-time error
-if $D$ is not an instance member and the name of $D$ is not $n$;
-and it is a compile-time error
-if $D$ is an instance member,
+In this case it is a compile-time error if $D$ is an instance member,
 and the interface of the enclosing class does not have a member named $n$.
 \EndCase
 
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -10598,7 +10598,8 @@ proceeds as follows:
 
 \begin{itemize}
 \item
-  In the case where \id{} is a local variable $v$,
+  In the case where \id{} is a local variable $v$
+  (\commentary{which may be a formal parameter}),
   the expression $e$ is evaluated to an object $o$,
   and the variable $v$ is bound to $o$.
   If no error occurs, the value of $a$ is $o$.
@@ -10861,7 +10862,7 @@ Except for errors inside $e_1$ and references to the name $x$,
 exactly the same compile-time errors that would be caused by
 \code{$x$.$v$ = $e_2$}
 are also generated in the case of $a$.
-%% TODO(eernst): Also, we should mention other cases, e.g., there is no getter `z.v`.
+Moreover, it is a compile-time error if $T$ does not have a getter named $v$.
 The static type of $a$ is the least upper bound of
 the static type of \code{$e_1$.$v$} and the static type of $e_2$.
 
@@ -10882,7 +10883,8 @@ Consider a compound assignment $a$ of the form \code{$e_1$[$e_2$] ??= $e_3$}.
 Exactly the same compile-time errors that would be caused by
 \code{$e_1$[$e_2$] = $e_3$}
 are also generated in the case of $a$.
-%% TODO(eernst): We should mention other cases, e.g., there is no `operator []`.
+Moreover, it is a compile-time error
+if the static type of $e_1$ does not have an `\code{operator []}'.
 The static type of $a$ is the least upper bound of
 the static type of \code{$e_1$[$e_2$]} and the static type of $e_3$.
 
@@ -10906,7 +10908,9 @@ Consider a compound assignment $a$ of the form \code{\SUPER.$v$ ??= $e$}.
 Exactly the same compile-time errors that would be caused by
 \code{\SUPER.$v$ = $e$}
 are also generated in the case of $a$.
-%% TODO(eernst): We should mention other cases, e.g., there is no getter `\SUPER.v`.
+Moreover, exactly the same compile-time errors that would be caused by
+evaluation of the expression \code{\SUPER.$v$}
+are also generated in the case of $a$.
 The static type of $a$ is the least upper bound of
 the static type of \code{\SUPER.$v$} and the static type of $e$.
 
@@ -11818,6 +11822,16 @@ In the situation where $S$ has
 exactly one declaration with basename \id,
 let $D$ be that declaration.
 
+\commentary{%
+In other words, if we are looking up a name $n$ with basename \id,
+we stop searching if we find any declaration named \id{} or \code{\id=}.
+If, in that scope, there are declarations for both \id{} and \code{\id=},
+we return the one which has the requested name $n$.
+In the case where only one declaration is present, we return it,
+even though it may have the name \id{} when $n$ is \code{\id=} or vice versa.
+That situation will cause an error, as specified below.%
+}
+
 \LMHash{}%
 \Case{$D$ does not exist}
 When no declaration with basename \id{} is in scope at the location $\ell$,
@@ -11832,7 +11846,12 @@ the following errors apply:
   % - in an expression in the initializer list of a constructor
   %   (be it a superinitializer, a field initializer, or an assertion).
   It is a compile-time error if $\ell$ is not inside
+  %% TODO(eernst): Come NNBD, adjust text for `late` variable initializers.
   the body of an instance member or a generative constructor.
+  \commentary{%
+    Note that an initializing expression for an instance variable
+    is not inside such a body.%
+  }
 \item
   When $\ell$ is inside the body of
   an instance member or a generative constructor,
@@ -11856,11 +11875,17 @@ is in scope at the location $\ell$.
 
 \LMHash{}%
 If $\ell$ does not occur in
-the body of an instance member or generative constructor,
+%% TODO(eernst): Come NNBD, update text for `late` variable initializers.
+the body of an instance member or a generative constructor,
 it is a compile-time error if the name of $D$ is not $n$,
 and it is a compile-time error if $D$ is an instance member.
 % An instance variable initializer is in the instance scope, so it is
 % actually possible for $D$ to be an instance member here.
+
+\commentary{%
+Note that an initializing expression for an instance variable
+is not inside such a body.%
+}
 
 \LMHash{}%
 Otherwise $\ell$ occurs in
@@ -11967,7 +11992,7 @@ Now proceed as described in the first applicable case from the following list:
 Note that a lexical lookup will never yield a declaration of
 an instance member.
 In each case where it is determined that there is no error
-and and instance member is the result of the lookup,
+and an instance member is the result of the lookup,
 the member signature from the interface of the enclosing class is yielded.
 
 The reason for this is that there may not be a declaration, e.g.,
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -11790,11 +11790,8 @@ Evaluation of an assignable expression of the form \code{\SUPER{}[$e_2$]} is equ
 This section specifies how to look up a name
 based on the enclosing lexical scopes.
 This is known as a \Index{lexical lookup}.
-When \id{} is an identifier or \code{[]},
-it may look up a name $n$ of the form \id{}
-as well as of the form \code{\id=}.
-When \id{} is an operator name other than \code{[]},
-it will look up \id{}.
+When \id{} is an identifier,
+it may look up a name $n$ of the form \id{} as well as of the form \code{\id=}.
 
 \LMHash{}%
 A lexical lookup yields a result which is
@@ -11817,7 +11814,7 @@ as detailed below.%
 \LMHash{}%
 Consider the situation where a name $n$ has basename \id{}
 (\ref{classMemberConflicts})
-where \id{} is an identifier or an operator name,
+where \id{} is an identifier,
 and a lexical lookup of $n$ is performed from a given location $\ell$.
 
 \commentary{%
@@ -11831,8 +11828,7 @@ and the location that determines which scopes are the enclosing ones.%
 }
 
 \LMHash{}%
-When we consider an occurrence of
-an identifier or operator name \id{} and say that
+When we consider an occurrence of an identifier \id{} and say that
 a lexical lookup of \id{} is performed,
 it is understood that the lookup is performed from
 the location of said occurrence of \id.
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -3486,7 +3486,7 @@ then the (implicit or explicit) superinitializer in $k$ is an error.%
 \LMHash{}%
 The superinitializer that appears, explicitly or implicitly,
 in the initializer list of a constant constructor
-must specify a constant constructor of
+must specify a generative constant constructor of
 the superclass of the immediately enclosing class,
 or a compile-time error occurs.
 
@@ -8851,17 +8851,19 @@ Note that the type argument list is omitted when $r = 0$ (\ref{generics}).
 }
 
 \LMHash{}%
-For the static analysis of $i$,
-a lexical lookup of \id{} is performed
-(\ref{lexicalLookup}):
+Perform a lexical lookup of \id{}
+(\ref{lexicalLookup})
+from the location of $i$.
 
 \LMHash{}%
 \Case{Lexical lookup yields a declaration}
-In this case the lexical lookup of \id{}
-yields a declaration $D$.
+Let $D$ be the declaration yielded by the lexical lookup of \id{}.
+
 \begin{itemize}
 \item
-  Consider the case where $D$ is a type declaration.
+  When $D$ is a type declaration, that is,
+  a declaration of a class, mixin, type alias, or type parameter,
+  the following applies:
   If $D$ is a declaration of a class $C$
   that has a constructor named $C$
   then the meaning of $i$ depends on the context:
@@ -8872,9 +8874,9 @@ yields a declaration $D$.
   \code{\CONST\,\,$i$};
   if $i$ does not occur in a constant context
   then $i$ is treated as \code{\NEW\,\,$i$}.
-  Otherwise a compile-time error occurs
-  (\commentary{that is, if $D$ does not declare a class,
-    or it declares a class that has no constructor named $C$}).
+  If $D$ is not a class declaration,
+  or it declares a class named $C$ that has no constructor named $C$,
+  a compile-time error occurs.
 \item
   Otherwise, if $D$ is a declaration of
   a local function,
@@ -8885,8 +8887,9 @@ yields a declaration $D$.
   a function expression invocation
   (\ref{functionExpressionInvocation}).
 \item
-  Otherwise, if $D$ is a static method, getter, or variable declaration
-  in the enclosing class $C$,
+  Otherwise, if $D$ is
+  a static method, static getter, or class variable declaration
+  in the enclosing class or mixin $C$,
   $i$ is treated as
   (\ref{notation})
   \code{$C$.$i$}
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -10474,84 +10508,157 @@ \subsection{Assignment}
...
-It is a compile-time error if $d$ denotes
-a prefix object, type declaration, or function declaration.
 
-\LMHash{}%
-If $d$ is the declaration of a local variable, the expression $e$ is evaluated to an object $o$.
-Then, the variable $v$ is bound to $o$.
-If no error occurs, the value of the assignment expression is $o$.
+\begin{itemize}
+\item
+  In the case where \id{} is a local variable $v$,
```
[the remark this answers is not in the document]
eernstg: Yes, this is exactly one of those locations where we otherwise have a (\commentary{which can be a formal parameter}). Fixed.

specification/dartLangSpec.tex
```diff
@@ -10746,9 +10858,12 @@ \subsubsection{Compound Assignment}
 Consider a compound assignment $a$ of the form \code{$e_1$.$v$ ??= $e_2$}.
 Let $T$ be the static type of $e_1$ and let $x$ be a fresh variable of type $T$.
 Except for errors inside $e_1$ and references to the name $x$,
-exactly the same compile-time errors that would be caused by \code{$x$.$v$ = $e_2$} are also generated in the case of $a$.
+exactly the same compile-time errors that would be caused by
+\code{$x$.$v$ = $e_2$}
+are also generated in the case of $a$.
 %% TODO(eernst): Also, we should mention other cases, e.g., there is no getter `z.v`.
```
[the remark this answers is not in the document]
eernstg: That must be a renaming glitch. It should say `x.v`. The point is that the dynamic semantics will evaluate `x.v`, so we should have a compile-time error when there is a setter but no getter in the interface of the type of `x`. Adjusted the text to say this, rather than having a TODO that might as well just be resolved here.

Made the same change to the following two cases (lines 10885, 10909).

specification/dartLangSpec.tex
```diff
@@ -11641,6 +11761,225 @@ \subsection{Assignable Expressions}
...
+}
+
+\LMHash{}%
+When we consider an occurrence of an identifier \id{} and say that
+a lexical lookup of \id{} is performed,
+it is understood that the lookup is performed from
+the location of said occurrence of \id.
+
+\LMHash{}%
+Let $S$ be the innermost lexical scope containing $\ell$
```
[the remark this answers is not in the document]
eernstg: The interactions between namespaces are not covered by scope nesting when it comes to imports or exports, because we have no notion of taking the union with lexical scoping (but we do have special conflict resolution rules for multiple imports/exports providing declarations with the same name).

But I suppose the relationship between "the union of all imported namespaces" (which includes hide/show, and conflict resolution) and the library declarations in the current library: The current ones just shadow the imported ones, similarly to nested scopes.

But this is of course the place where we haven't arrived at a conclusion yet: We agree that lexical scoping should consistently bundle setters/getters, as specified in this PR, but it might be too much of a breaking change to do the same thing for the relationship between imports and the current library declarations.

We should discuss the import/export issues in the context of #534, and focus on lexical scoping here, but I agree that we should keep the overall consistency in mind.

So, #534 in its current form is as consistent as possible with the lexical scoping rules proposed in this PR. In particular, it consistently considers `id`/`id=` name pairs when it detects name clashes among declarations from different sources. However, it's a matter of a small change in like 4-5 sentences to make it use a plain name matching approach (such that it's OK to import getter `x` from one library and setter `x=` from another one, or to have the getter locally and import the setter, etc.).

When it comes to the significance of these matters for the current PR, I think it's irrelevant: The import/export rules in #534 specify how to compute the `library namespace` of the current library. That's the outermost scope, and we never proceed to look into import directives in order to find additional names. They're already included when we first encounter that 'library namespace'.

For 'lib1.dart', 'lib2.dart', #534 in its current form specifies that the `x` getter is removed from the imported namespace from 'lib1'. The more forgiving variant that matches on plain names would include the `x` getter, and allow `print(x)` to access it, because both `x` and `x=` would then be in the library scope.

> CFE is treating the imports as being in a different lexical scope

I think it's dangerous to even consider that interpretation, because lexical scoping doesn't have a union (and the associated symmetric conflict handling step), and we need that for imports.

For 'lib4', #534 in the current form considers `x` and `x=` to be a conflict, so both are removed from the imported namespace (and then a `NAME_CONFLICT`  binding is added to the library scope, such that we don't get into the situation where `x` is missing entirely, and some instance method body does `x++` and finds an inherited instance variable when it should say "You're trying to use `x`, but that name is ambiguous").

So my intention is to clear up the mess. ;-)

We can do this by using this PR to get a consistent approach to lexical scoping, and (1) #534 in its current form in order to treat imports/exports as consistently as possible, or (2) adjust #543 to use plain name clashes (and allow mixing getters/setters from different sources freely), in order to avoid some breakage.

I think the text in line 11812 can be used without changes, based on the notion of the 'library scope' of the current library, and relying on #534 in one form or another to explicitly spell out what's in that library scope.

The current text about `library scope` and `scope of a library` and `top level scope` does not contradict this.

So I think we could choose to resolve this comment and clarify the remaining issues in the context of #534, but I'll keep it open to make sure that you're OK with that.

specification/dartLangSpec.tex
```diff
@@ -11641,6 +11761,225 @@ \subsection{Assignable Expressions}
...
+\Case{$D$ does not exist}
+When no declaration with basename \id{} is in scope at the location $\ell$,
+the following errors apply:
+
+\begin{itemize}
+\item
+  % Here is the list of cases covered by this item: \ell is..
+  % - inside a library function, getter, setter, variable initializer;
+  % - inside a static method, getter, setter, variable initializer;
+  % - inside an instance variable initializer;
```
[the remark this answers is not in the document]
eernstg: That's actually one of the reasons why I preserved the listing of cases, such that we can re-check the list whenever something new happens. Added a TODO comment, similar to the ones that we have had for locations affected by the integration of a given feature spec.

specification/dartLangSpec.tex
```diff
@@ -11641,6 +11761,225 @@ \subsection{Assignable Expressions}
...
+  % - in an expression in the initializer list of a constructor
+  %   (be it a superinitializer, a field initializer, or an assertion).
+  It is a compile-time error if $\ell$ is not inside
+  the body of an instance member or a generative constructor.
```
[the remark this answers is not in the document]
eernstg: I'm using the term 'initializing expression' when referring to `e` in `var x = e`, not 'body'. So if that terminology is used consistently we should be OK.

Checking: We have 111 occurrences of `body`: They are all referring to the body of a function or of a constructor; or the body of a class or typedef; or the body of a loop.

We also have `A member is abstract if it has no body` (which would be wrong if we take it to mean that `int x;` is abstract).

An instance variable is an instance member (`The instance members of a class are its instance methods, getters, setters and instance variables`).

I added commentary to make this distinction explicit.

specification/dartLangSpec.tex
```diff
@@ -11641,6 +11761,225 @@ \subsection{Assignable Expressions}
...
+  % - a local variable (\commentary{why may be a formal parameter});
+  % - a local function.
+  Otherwise, the lexical lookup yields $D$.
+\end{itemize}
+
+\commentary{%
+Note that a lexical lookup will never yield a declaration of
+an instance member.
+In each case where it is determined that there is no error
+and and instance member is the result of the lookup,
```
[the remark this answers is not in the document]

specification/dartLangSpec.tex
```diff
@@ -11641,6 +11761,225 @@ \subsection{Assignable Expressions}
...
+\LMHash{}%
+Let $S$ be the innermost lexical scope containing $\ell$
+which has a declaration with basename \id.
+In the case where $S$ has
+a declaration named \id{} as well as a declaration named \code{\id=},
+let $D$ be the declaration named $n$.
+In the situation where $S$ has
+exactly one declaration with basename \id,
+let $D$ be that declaration.
+
```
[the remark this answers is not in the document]

specification/dartLangSpec.tex
```diff
@@ -8848,53 +8851,84 @@ \subsubsection{Unqualified Invocation}
 }
 
 \LMHash{}%
-It is a compile-time error if $i$ occurs inside a top-level or static function
-(be it function, method, getter, or setter)
-or a top-level or static variable initializer,
-and there is no lexically visible declaration named \id{} in scope.
+For the static analysis of $i$,
```
lrhn: I'm not particularly fond of starting a sentence with a ... preposition phrase? (Is that a thing).

Maybe just:

> First perform a lexical lookup of \id{} (\ref{lexicalLookup}) from the location of $i$.

(A lexical lookup needs a context, you can't do "lexical lookup" in the abstract).

specification/dartLangSpec.tex
```diff
@@ -8848,53 +8851,84 @@ \subsubsection{Unqualified Invocation}
...
+a lexical lookup of \id{} is performed
+(\ref{lexicalLookup}):
 
 \LMHash{}%
-If there exists a lexically visible declaration named \id,
-let $D_{id}$ be the innermost such declaration.
-Then:
+\Case{Lexical lookup yields a declaration}
+In this case the lexical lookup of \id{}
```
lrhn: Is that a definition, an assertion or just unnecessary flavor text? It seems to just repeat the Case text, and it's unclear why it does that.

Maybe just:
> Let $D$ be the declaration yielded by the lexical lookup of \id{}.

specification/dartLangSpec.tex
```diff
@@ -8848,53 +8851,84 @@ \subsubsection{Unqualified Invocation}
...
+yields a declaration $D$.
 \begin{itemize}
-\item Consider the situation where $D_{id}$ is a type declaration.
-  If $D_{id}$ is a declaration of a class $C$
+\item
+  Consider the case where $D$ is a type declaration.
```
lrhn: Just:

> If $D$ is a type declaration

(and say what a "type declaration" is. I assume it means a class declaration, mixin declaration or type alias declaration. Does it include a type variable?).

specification/dartLangSpec.tex
```diff
@@ -8848,53 +8851,84 @@ \subsubsection{Unqualified Invocation}
...
-  then $i$ is equivalent to \code{\CONST\,\,$i$};
+  then $i$ is treated as
+  (\ref{notation})
+  \code{\CONST\,\,$i$};
   if $i$ does not occur in a constant context
-  then $i$ is equivalent to \code{\NEW\,\,$i$}.
+  then $i$ is treated as \code{\NEW\,\,$i$}.
   Otherwise a compile-time error occurs
-  (\commentary{that is, if $D_{id}$ does not declare a class,
+  (\commentary{that is, if $D$ does not declare a class,
```
lrhn: This doesn't *clearly* address the case where $D$ is a type declaration, but not a class declaration.
The `Otherwise a compile-time error occurs` looks like it binds to the immediately preceding `if`. (Nobody reads the commentary!)

Maybe change it to:
> If $D$ is not a class declaration, or the declared class $C$ has no constructor named $C$, then a compile-time error occurs.

specification/dartLangSpec.tex
```diff
@@ -8848,53 +8851,84 @@ \subsubsection{Unqualified Invocation}
...
-\item Otherwise, if $D_{id}$ is a static method of the enclosing class $C$,
-  $i$ is equivalent to
-  \code{$C$.\id<$A_1, \ldots,\ A_r$>($a_1, \ldots,\ a_n,\ x_{n+1}$: $a_{n+1}, \ldots,\ x_{n+k}$: $a_{n+k}$)}.
-\item Otherwise, if $i$ occurs in an instance method body,
-  $i$ is equivalent to the ordinary method invocation
-
-  \code{\THIS{}.\id<$A_1, \ldots,\ A_r$>($a_1, \ldots,\ a_n,\ x_{n+1}$: $a_{n+1}, \ldots,\ x_{n+k}$: $a_{n+k}$)}.
+\item
+  Otherwise, if $D$ is a static method, getter, or variable declaration
+  in the enclosing class $C$,
```
lrhn: enclosing class -> enclosing class or mixin

(and soon extension)

specification/dartLangSpec.tex
```diff
@@ -8848,53 +8851,84 @@ \subsubsection{Unqualified Invocation}
...
+\EndCase
+
+\Case{Lexical lookup yields a member signature}
+When the lexical lookup of \id{} yields a member signature,
+$i$ is treated as
+(\ref{notation})
+the ordinary method invocation
+\code{\THIS.$i$}
+(\ref{ordinaryInvocation}).
+
```
lrhn: (Be prepared for extensions: If $D$ is an extension member invocation, then you can't just rewrite it with a prefix, you have to replace `id` by `Ext<typeArgs>(id)`.)

specification/dartLangSpec.tex
```diff
@@ -10474,84 +10508,158 @@ \subsection{Assignment}
...
-(\commentary{which may be declared explicitly or induced by an instance variable, etc}).
-Let $T$ be the static type of $v$ when $v$ denotes a variable,
-otherwise let $T$ be the static type of the formal parameter of the setter \code{$v$=}.
-It is a compile-time error if the static type of $e$ may not be assigned to $T$.
-The static type of $a$ is the static type of $e$.
+\Case{\code{\id{} = $e$}}
+Consider an assignment $a$ of the form \code{\id{} = $e$},
+where \id{} is an identifier.
+
+If \id{} is a local variable $v$
```
lrhn: Why not talk about the result of lexical lookup here? How do you *know* that *id* "is a local variable" without doing lexical lookup.

specification/dartLangSpec.tex
```diff
@@ -10474,84 +10508,158 @@ \subsection{Assignment}
...
+
+\begin{itemize}
+\item
+  When the lexical lookup yields a declaration $D$,
+  it is guaranteed to be a setter
+  (\commentary{that may be explicit or induced implicitly by a variable})
+  because other declarations do not have a name
+  of the form \code{\id=}.
+
+  If $D$ is the declaration of a static setter in class $C$
```
lrhn: class or mixin

specification/dartLangSpec.tex
```diff
@@ -10474,84 +10508,158 @@ \subsection{Assignment}
...
+  Otherwise, a compile-time error occurs,
+  unless the static type of $e$ is assignable to the parameter type of $D$.
+\item
+  When the lexical lookup yields a member signature,
+  $a$ is treated as
+  (\ref{notation})
+  \code{\THIS.\id{} = $e$}.
+
+  \commentary{%
+    This occurs when no declaration named \code{\id=} is in scope,
```
lrhn: > This also occurs  ...

(because otherwise it sounds like it only occurs then, and not when you actually find an instance setter first).

specification/dartLangSpec.tex
```diff
@@ -10474,84 +10508,158 @@ \subsection{Assignment}
...
-If $d$ is the declaration of a local variable, the expression $e$ is evaluated to an object $o$.
-Then, the variable $v$ is bound to $o$.
-If no error occurs, the value of the assignment expression is $o$.
+\begin{itemize}
+\item
+  In the case where \id{} is a local variable $v$
+  (\commentary{which may be a formal parameter}),
+  the expression $e$ is evaluated to an object $o$,
+  and the variable $v$ is bound to $o$.
+  If no error occurs, the value of $a$ is $o$.
```
lrhn: Just say: 

> Then $a$ evaluates to the object $o$ (\ref{expressionEvaluation}).

Errors are stated elsewhere to automatically interrupt evaluation, only mention them when they are introduced or when you need to handle them, otherwise they propagate automatically.

specification/dartLangSpec.tex
```diff
@@ -10474,84 +10508,158 @@ \subsection{Assignment}
...
+  \commentary{%
+    $v$ is not a final variable,
+    because a compile-time error would then have occurred.
+    But a dynamic type error could occur, e.g.,
+    when $e$ has type \DYNAMIC{} and a run-time check on $o$ fails.%
```
lrhn: As I see it, that error is part of evaluation of $e$ with a context type that introduces an implicit downcast, not part of the assignment. So, that does not need to be mentioned here.

specification/dartLangSpec.tex
```diff
@@ -10474,84 +10508,158 @@ \subsection{Assignment}
...
-It is a compile-time error if $d$ denotes
-a prefix object, type declaration, or function declaration.
 
-\LMHash{}%
-If $d$ is the declaration of a local variable, the expression $e$ is evaluated to an object $o$.
-Then, the variable $v$ is bound to $o$.
-If no error occurs, the value of the assignment expression is $o$.
+\begin{itemize}
+\item
+  In the case where \id{} is a local variable $v$
```
lrhn: Again, do lexical lookup to figure that out.

specification/dartLangSpec.tex
```diff
@@ -10474,84 +10508,158 @@ \subsection{Assignment}
...
+    $v$ is not a final variable,
+    because a compile-time error would then have occurred.
+    But a dynamic type error could occur, e.g.,
+    when $e$ has type \DYNAMIC{} and a run-time check on $o$ fails.%
+  }
+\item
+  In the case where the lexical lookup of \code{\id=}
+  from the location of \id{}
+  yields a declaration $D$,
+  $D$ is a top level setter $s$
```
lrhn: > then $D$ is necessarily a top level setter $s$.

specification/dartLangSpec.tex
```diff
@@ -10474,84 +10508,158 @@ \subsection{Assignment}
...
+  (\commentary{possibly implicitly induced by a variable}).
 
-\commentary{
-If $v$ is a final variable, a compile-time error has occurred and execution is unspecified.
-But a program with no compile-time errors may incur a dynamic type error.
-}
+  The expression $e$ is evaluated to an object $o$.
+  Then the setter $s$ is invoked
+  with its formal parameter bound to $o$.
+  If no error occurs, the value of $a$ is $o$.
```
lrhn: > Then $a$ evaluates to the object $o$.

(with no mention of errors).

specification/dartLangSpec.tex
```diff
@@ -10474,84 +10508,158 @@ \subsection{Assignment}
...
-is not a subtype of the actual type
-(\ref{actualTypeOfADeclaration})
-of $v$.
+\LMHash{}%
+Evaluation of an assignment $a$ of the form \code{$p$.\id{} = $e$}
+proceeds as follows:
+The expression $e$ is evaluated to an object $o$.
+Then the setter denoted by \code{$p$.\id} is invoked
+with its formal parameter bound to $o$.
+If no error occurs, the value of $a$ is $o$.
```
lrhn: > Then $a$ evaluates to the object $o$.

(same all similar places below, I'll stop saying it now).

specification/dartLangSpec.tex
```diff
@@ -11641,6 +11765,246 @@ \subsection{Assignable Expressions}
...
+
+\LMHash{}%
+This section specifies how to look up a name
+based on the enclosing lexical scopes.
+This is known as a \Index{lexical lookup}.
+When \id{} is an identifier, it may look up a name $n$ of the form \id{}
+as well as of the form \code{\id=}.
+
+\LMHash{}%
+A lexical lookup does not succeed or fail.
```
lrhn: That's meaningless.
Does it "not *just* succeed or fail"?

Generally don't waste words saying what *isn't* true. Especially, don't *start* with it.

specification/dartLangSpec.tex
```diff
@@ -11641,6 +11765,246 @@ \subsection{Assignable Expressions}
...
+}
+
+\LMHash{}%
+Consider the situation where a name $n$ has basename \id{}
+(\ref{classMemberConflicts}),
+an identifier \id{} occurs at a location $\ell$,
+and a lexical lookup of $n$ is performed from $\ell$.
+
+\commentary{%
+We need to specify a name and a location from where a lookup is performed.
```
lrhn: Drop "need". We *do*. 
We probably wouldn't do it unless we needed to, but you don't need to defend the choice :).

specification/dartLangSpec.tex
```diff
@@ -11641,6 +11765,246 @@ \subsection{Assignable Expressions}
...
+For instance,
```
lrhn: Instance of what? I don't see the connection to the previous statement.

specification/dartLangSpec.tex
```diff
@@ -11641,6 +11765,246 @@ \subsection{Assignable Expressions}
...
+  which has the name $n$.
+  \commentary{%
+    In this case it is guaranteed that $\ell$ occurs inside
+    the body of an instance member or generative constructor,
+    and said member exists.%
+  }
+\item
+  Consider the case where $D$ is a formal type parameter declaration.
+  It is a compile-time error if $\ell$ occurs inside
+  a static method, getter, setter, or variable initializer.
```
lrhn: No longer true for a *generic* static method, which can declare its own type parameter.
So, this is only for type parameters of a class or mixin.

(Also, what about:
```dart
typedef Foo<T> = T Function(T);
```
Do we do lexical lookup on the `T`'s? If so, they also find a type parameter declaration.

specification/dartLangSpec.tex
```diff
@@ -11677,113 +12041,209 @@ \subsection{Identifier Reference}
...
 <qualified> ::= <identifier> (`.' <identifier>)?
 \end{grammar}
 
 \LMHash{}%
 A built-in identifier is one of the identifiers produced by the production
 \synt{BUILT\_IN\_IDENTIFIER}.
-It is a compile-time error if a built-in identifier is used as the declared name of a prefix, class, type parameter or type alias.
-It is a compile-time error to use a built-in identifier other than \DYNAMIC{} in a type annotation or type parameter.
+It is a compile-time error if a built-in identifier is used as
+the declared name of a prefix, class, type parameter or type alias.
```
lrhn: or mixin

specification/dartLangSpec.tex
```diff
@@ -11677,113 +12041,209 @@ \subsection{Identifier Reference}
...
+It is a compile-time error to use a built-in identifier
+other than \DYNAMIC{} or \FUNCTION{}
+in a type annotation or a type parameter bound.
```
lrhn: What does "*in* a type annotation or type parameter bound" mean?

Is `Function` "in" `List<notFunction> x;`?

Does this paragraph actually say anything? If you cannot declare a type named as a built-in identifier, then you can't use a built-in identifier as a type (unless it's `dynamic` or `Function`, which are types). So, what are we saying here that isn't already invalid?

specification/dartLangSpec.tex
```diff
@@ -11677,113 +12041,209 @@ \subsection{Identifier Reference}
...
-Furthermore, types should begin with an uppercase letter (see the appendix) and so no clash should occur in any Dart user program anyway.
+This eliminates many confusing situations,
+both for human readers and during parsing.%
 }
 
 \LMHash{}%
-It is a compile-time error if either of the identifiers \AWAIT{} or \YIELD{} is used as an identifier in a function body marked with either \ASYNC{}, \code{\ASYNC*} or \code{\SYNC*}.
+It is a compile-time error if either of the identifiers \AWAIT{} or \YIELD{}
+is used as an identifier in a function body
+marked with either \ASYNC{}, \code{\ASYNC*} or \code{\SYNC*}.
```
lrhn: (Nitpick, should affect parameters, but not nested functions.
```dart
foo(int await) async {  // should probably not be allowed
   bar(int await) { print(await); } // Should actually be allowed.
}
```

(I'll stop digressing now).

specification/dartLangSpec.tex
```diff
@@ -11677,113 +12041,209 @@ \subsection{Identifier Reference}
...
+which is an identifier \id{}
+is determined as follows:
 
 \LMHash{}%
-Let $d$ be the innermost declaration in the enclosing lexical scope whose name is \id{} or \code{\id=}.
-If no such declaration exists in the lexical scope, let $d$ be the declaration of the inherited member named \id{} if it exists.
+\Case{Lexical lookup yields a declaration}
+In this case the lexical lookup
+(\ref{lexicalLookup})
+for \id{} yields a declaration $D$.
```
lrhn: Repeating the case text.

Maybe just:

> Let $D$ be the declaration yielded by the lexical lookup of \id.

specification/dartLangSpec.tex
```diff
@@ -11677,113 +12041,209 @@ \subsection{Identifier Reference}
...
-\item If $d$ is a local variable (\commentary{which can be a formal parameter}) then $e$ evaluates to the current binding of \id.
-\item If $d$ is a static method, top-level function or local function then $e$ evaluates to the function object obtained by closurization (\ref{functionClosurization}) of the declaration denoted by $d$.
-\item If $d$ is the declaration of a class variable, static getter or static setter declared in class $C$, then evaluation of $e$ is equivalent to evaluation of the property extraction (\ref{propertyExtraction}) \code{$C$.\id}.
-\item If $d$ is the declaration of a library variable, top-level getter or top-level setter, then evaluation of $e$ is equivalent to evaluation of the top level getter invocation (\ref{topLevelGetterInvocation}) \id.
-\item Otherwise, if $e$ occurs inside a top level or static function (be it function, method, getter, or setter) or variable initializer, evaluation of $e$ causes a \code{NoSuchMethod} to be thrown.
-\item Otherwise, evaluation of $e$ is equivalent to evaluation of the property extraction (\ref{propertyExtraction}) \code{\THIS.\id}.
+\item
+  If $D$ is a class, mixin, type alias or type parameter,
```
lrhn: > ... parameter declaration,

specification/dartLangSpec.tex
```diff
@@ -11677,113 +12041,209 @@ \subsection{Identifier Reference}
...
+  the static type of $e$ is \code{Type}.
+\item
+  If $D$ is the declaration of
+  a library variable or getter,
```
lrhn: Can you find a library variable? Won't it always be the induced getter?

specification/dartLangSpec.tex
```diff
@@ -11677,113 +12041,209 @@ \subsection{Identifier Reference}
...
+  the static type of $e$ is the static type of the
+  top level getter invocation \id{}
+  (\ref{topLevelGetterInvocation}).
+\item
+  If $D$ is a static method, library function, or local function,
+  the static type of $e$ is the function type of $D$.
```
lrhn: Here we don't address instantiated tear-offs based on context type.

specification/dartLangSpec.tex
```diff
@@ -11677,113 +12041,209 @@ \subsection{Identifier Reference}
...
+\item
+  If $D$ is the declaration of a static variable or getter
```
lrhn: Again, you'll only find the implicit getter, right?

specification/dartLangSpec.tex
```diff
@@ -11677,113 +12041,209 @@ \subsection{Identifier Reference}
...
+  and $D$ occurs in the class $C$,
+  the static type of $e$ is the static type of the getter invocation
+  (\ref{propertyExtraction})
+  \code{$C$.\id}.
```
lrhn: That's always the return type of $D$, so maybe just write that?

specification/dartLangSpec.tex
```diff
@@ -11677,113 +12041,209 @@ \subsection{Identifier Reference}
...
+In this case the lexical lookup
+(\ref{lexicalLookup})
+for \id{} yields a member signature $s$.
+%
+In this situation $e$ is treated as
+(\ref{notation})
+\code{\THIS.\id}.
+
+\commentary{%
+This occurs when no declaration named \id{} is in scope,
```
lrhn: > This also occurs ...

specification/dartLangSpec.tex
```diff
@@ -11677,113 +12041,209 @@ \subsection{Identifier Reference}
...
+  the value of $e$ is an object implementing the class \code{Type}
+  which reifies the corresponding type.
+\item
+  If $D$ is a type parameter $X$ then the value of $e$ is
+  the value of the actual type argument corresponding to $X$
+  that was passed to the generative constructor that created
+  the current binding of \THIS{}.
+\item
+  If $D$ is the declaration of
+  a library variable, top-level getter, or top-level setter,
```
lrhn: It can't be a setter, right? (Or variable).

specification/dartLangSpec.tex
```diff
@@ -11677,113 +12041,209 @@ \subsection{Identifier Reference}
...
+  then evaluation of $e$ is equivalent to evaluation of an invocation of
+  the top level getter \id{}
+  (\ref{topLevelGetterInvocation}).
+\item
+  If $D$ is a constant variable of one of the forms
```
lrhn: local variable or also static/top-level constant variables?

specification/dartLangSpec.tex
```diff
@@ -11677,113 +12041,209 @@ \subsection{Identifier Reference}
...
+  then the value of $e$ is the value of the constant expression $e'$.
+\item
+  If $D$ is a declaration of
+  a top-level function, static method, or local function,
+  then $e$ evaluates to the function object obtained by closurization
+  (\ref{functionClosurization})
+  of $D$.
+\item
+  If $D$ is a local variable (\commentary{which can be a formal parameter})
+  then $e$ evaluates to the current binding of \id.
```
lrhn: > ... current binding of that variable. 

(There might be more than one \id in the program, the lookup distinguishes them, so use the result of the lookup. The name itself is underspecified unless you reintroduces scopes).

specification/dartLangSpec.tex
```diff
@@ -13950,7 +14410,7 @@ \subsection{Exports}
 %% natural to define it at the first usage in terms of the page number.
 \LMHash{}%
 We define an operation for
-combining namespaces with disjoint sets of keys  as follows. 
+combining namespaces with disjoint sets of keys  as follows.
```
lrhn: Two spaces between "keys  as".

specification/dartLangSpec.tex
eernstg: I tried to get rid of 'Consider the case' but we need to make sure that it establishes a premise for the rest of this bullet point (it shouldn't be a reasonable interpretation that it's only valid for half of the bullet point). I think I found a way to preserve that property which is still less verbose.

Changed to use an explicit list of non-class type declarations at the end.

specification/dartLangSpec.tex

specification/dartLangSpec.tex
eernstg: Indeed! Done.

specification/dartLangSpec.tex
eernstg: Added a TODO marked 'extension methods'.

specification/dartLangSpec.tex

specification/dartLangSpec.tex
eernstg: Adjusted to avoid saying whether the declaration was found, stating instead the important guarantees for this situation. Then it's up to the inquisitive reader to look at \ref{lexicalScope} to verify that it's always _as intended_ that we are now accessing an instance member rather than any other kind of entity.

specification/dartLangSpec.tex

specification/dartLangSpec.tex

specification/dartLangSpec.tex

specification/dartLangSpec.tex

specification/dartLangSpec.tex
eernstg: Done. There weren't any more places. ;-)

specification/dartLangSpec.tex
eernstg: But it doesn't fail at all. It _always_ delivers a result. (So it's also meaningless to say that it 'succeeds'.)

A compile-time error may occur during the lexical lookup, but we usually do not specify propagation of errors (for instance, when specifying `e++` we don't say 'if a compile-time error occurred during the determination of the static type of `e` then `e++` is also a compile-time error').

I reorganized the text such that it talks about the positive outcome first, and then explains why we don't have to mention the case `if the lexical lookup failed` in the locations where such a lookup is performed.

specification/dartLangSpec.tex
eernstg: This is about the _reason_ for mentioning $\ell$ in the definition (it's a new concept in the specification). The spec never talked about 'locations' before, that was always just implicit in the fact that the spec generally talks about a snippet of code which has a location because it's known to be part of the program (or it's a desugaring which is, implicitly, considered to be in the same location as the original).

I rewrote the whole paragraph, avoiding `need`, and the result is clearer.

specification/dartLangSpec.tex
eernstg: All rewritten.

specification/dartLangSpec.tex
eernstg: Right! Changed to say `of a class or a mixin`, which leaves type aliases and generic methods untouched.

For the `typedef` we definitely need to perform a scoped lookup:

```dart
typedef F<S> = T Function<T>(S);
```

specification/dartLangSpec.tex

specification/dartLangSpec.tex
eernstg: I think it won't make anybody happier to allow that nested case. I'd prefer to leave it as it is. ;-)

specification/dartLangSpec.tex
eernstg: It used to be `as a type annotation or type parameter`, and it was adjusted to `in a type annotation..` because we wouldn't want to allow `List<static> x;`.

I added `as an identifier` to rule out the case `notFunction`.

Given that we'd want to have reasonably clear error messages, I think it's useful to say that we cannot use things like `List<static>` for syntactic reasons. If we do not say this then I'm sure you are right that it will turn out to be a compile-time error anyway, because we can't find a type that has the name `static`, but I still think it's going to be a less tool-team-friendly approach if we reach that result during static analysis (resolution), and require that the parser must survive this mal-terms first.

So I do think that it's useful to have a rule about these `<type>` cases, not just rules against declaration of types with those names.

specification/dartLangSpec.tex

specification/dartLangSpec.tex

specification/dartLangSpec.tex

specification/dartLangSpec.tex
eernstg: Made it explicit that it is all of them.

specification/dartLangSpec.tex
eernstg: Now using a named variable $v$, such that it is implied that we know more than the name \id. I believe this makes it reasonable to say that it determines the location where its binding (value) can be looked up, corresponding to the scope where we found it at compile-time.

specification/dartLangSpec.tex
eernstg: Fixed.

specification/dartLangSpec.tex

specification/dartLangSpec.tex

specification/dartLangSpec.tex
eernstg: The section about generic function instantiation is written in such a way that it relies on "a compile-time error occurs except when the following step succeeds:" and then it explains how to perform generic function instantiation based on the types that we found for the target term (which is required to be a generic function type).

So it is actually consistent that we don't mention generic function instantiation right here.

I added a commentary in order to remind the reader that it may occur.

specification/dartLangSpec.tex
```diff
@@ -3481,7 +3484,7 @@ \subsubsection{Constant Constructors}
 }
 
 \LMHash{}%
-The superinitializer that appears, explicitly or implicitly, 
+The superinitializer that appears, explicitly or implicitly,
 in the initializer list of a constant constructor
 must specify a constant constructor of
```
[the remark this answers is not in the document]
eernstg: Line 3057 says that `It is a compile-time error if class $S$ does not declare a generative constructor named $S$ (respectively \code{$S$.\id})`, which implies in that context that a superinitializer must invoke a _generative_ constructor. Added it anyway, seems like a non-harmful reminder.

specification/dartLangSpec.tex
eernstg: Resolved in the same way as the other location with similar text.

specification/dartLangSpec.tex
eernstg: I already fixed the setter issue (don't know how it pops up here now).

But the variable issue is a little bit messy. We do have implicitly induced setters and getters, but we do not consistently say that those are the ones that we obtain from lookups. So a lookup may yield a variable declaration. So it would be a substantially bigger change to say consistently that lookups can never yield a non-local variable. Please create a separate issue if you wish to push that change.

specification/dartLangSpec.tex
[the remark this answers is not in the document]
eernstg: I think this issue has been resolved: Imports are specified in #534 in terms of a separate mechanism that does not involve scopes or nesting, and I believe that it would just introduce a bit of unnecessary complexity if we were to unite the scope-nesting mechanism and the mechanism used to construct the library namespace from imports and local declarations.

specification/dartLangSpec.tex
eernstg: With the rewrite of the whole paragraph, I believe this issue has been resolved.

specification/dartLangSpec.tex
eernstg: I'll mark this as resolved; please create an issue in order to get the enhancement where `await` is allowed in such places where it is currently an error.

specification/dartLangSpec.tex
eernstg: The latest commit on this branch cleans up the references to variables. They are now exclusively referring to local variables, or they talk about getters/setters, along with `\commentary{which may be implicitly induced by a ... variable}` where applicable.

specification/dartLangSpec.tex
eernstg: Cf. comment on line 12104.

specification/dartLangSpec.tex
eernstg: For spec simplicity, we can delete the sentence 12061-12063 and let tools implicitly rely on the fact that `List<static>` and such must be an error (and we don't specify whether they make the parser survive it and then fail to find a type, or they make it a syntax error).

Or we can keep it, because it documents that tools can rely on such things to be a compile-time error no matter what, which may simplify things for human readers of code as well as parsers.

@lrhn, what do you prefer?

lrhn:
LGTM

PR #557, opened 2019-09-02
Feature spec of explicit variance

eernstg:
This is a first draft version of the feature specification of explicit variance support in Dart: (1) Explicit variance modifiers `out`/`inout`/`in` that generic type declarations can use on their type parameters, and (2) an explicit invariance modifier `exactly` on type arguments that client code can use in order to obtain statically safe use of legacy classes (using the current kind of covariance which requires dynamic checks).

It is rather permissive, in the sense that it allows for things like `class B<out X> extends A<X> {}` where `A` is a legacy class (say, `class A<X> { foo(X x) {}}`), which means that we _can_ have a variable `b` of type `B<num>`, and we can call `b.foo(1.23)`, and it will throw at run-time.

We could also say that the method `foo` is inaccessible in `B` (just like Java wildcards: we could "filter" out the part of the API whose member signatures violate the given variance). But I suspect that it will work better in practice to catch such cases with a lint.

Member declarations involving type parameters that have explicit variance modifiers are checked strictly (so `class B<out X> { void foo(X x) {...}}` is a compile-time error), but it is of course always possible to check such properties in the body and throw (`void foo(Object? o) { X x = o as X; ...}`), and, with that, I saw no reason to prohibit `covariant` parameters.

### DIFF
--- a/accepted/future-releases/variance/feature-specification-use-site-variance.md
+++ b/accepted/future-releases/variance/feature-specification-use-site-variance.md
@@ -8,8 +8,7 @@ Status: Draft
 ## CHANGELOG
 
 2019.10.15:
-- Created this as a variant of the feature specification that only has
-  use-site invariance (using `exactly`).
+- Created this as a variant of the feature specification that only has use-site invariance (`exactly`): This proposal has all of `out`/`inout`/`in`.
 
 
 ## Summary
@@ -39,7 +38,7 @@ The grammar is adjusted as follows:
     ('extends' <typeNotVoid>)?
 
 <varianceModifier> ::= // New.
-    'out' | 'inout' | 'in'
+    'out' | 'inout' '?'? | 'in'
 
 <typeArguments> ::= // Modified.
     '<' <typeArgumentList> '>'
@@ -48,7 +47,7 @@ The grammar is adjusted as follows:
     <typeArgument> (',' <typeArgument>)*
 
 <typeArgument> ::= // New.
-    (<varianceModifier> '?'?)? <type> |
+    <varianceModifier>? <type> |
     '*'
 ```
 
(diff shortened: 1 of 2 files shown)


accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,418 @@
...
+
+The grammar is adjusted as follows:
+
+```
+<typeParameter> ::= // Modified.
+    <metadata> <typeParameterVariance>? <typeIdentifier>
+    (('extends'|'super') <typeNotVoid>)?
+
+<typeParameterVariance> ::= // New.
+    'out' | 'inout' | 'in'
```
leafpetersen: Since out and inout have no special status, presumably it's valid to use them as type parameter names?  And even with variance modifiers?  e.g. `class C<out out> {}` is a valid declaration of a class with one covariant type parameter named `out`?

leafpetersen: cc @kallentu

eernstg (commented):
Review response. Resolved several threads (please reopen if needed); a couple of threads are still open.

I haven't made any changes in the direction of supporting other variances than invariance in the use-site part, I'd expect that to be handled in another iteration.

accepted/future-releases/variance/feature-specification-use-site-variance.md
```diff
@@ -0,0 +1,453 @@
...
+
+Let _D_ be the declaration of a class or mixin, and let _X_ be a type parameter declared by _D_.
+
+We say that _X_ _is covariant_ if it has no variance modifier or it has the variance modifier `out`; that it _is invariant_ if it has the variance modifier `inout`; and that it _is contravariant_ if it has the variance modifier `in`.
+
+If _X_ has the variance modifier `out` then it is a compile-time error for _X_ to occur in a non-covariant position in a member signature in the body of _D_, except that it is not an error if it occurs in a covariant position in the type annotation of a covariant formal parameter (*note that this is a contravariant position in the member signature as a whole*). *For instance, _X_ can not be the type of a method parameter (unless covariant), and it can not be the bound of a type parameter of a generic method.*
```
leafpetersen: We should probably make it explicit that this covers setters as well.  Same for line 202.

leafpetersen: We also should make it explicit that `covariant` on a field covers the setter, but *not* the getter.  So we have that:

```
class C<in T> {
   covariant C<T> f1; // Error, the C<T> appears in an invalid position in the getter
   final covariant C<T> f2; // Error as above
}

class D<out T> {
  covariant D<T> f1; // Ok: the getter is properly covariant, and the setter is covered by the covariant annotation.
}
```

cc @kallentu

accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,418 @@
...
+
+*The variance for each type parameter of a type alias is restricted based on the body of the type alias. Explicit variance modifiers may be used to document how the type parameter is used on the right hand side, and they may be used to impose more strict constraints than those implied by the right hand side.*
+
+We say that a type parameter _X_ of a type alias _F_ _is covariant/invariant/contravariant_ if it has the variance modifier `out`/`inout`/`in`, respectively. We say that it is _covariant/contravariant_ if it has no variance modifier, and it occurs only covariantly/contravariantly, respectively, on the right hand side of `=` in the type alias (*for an old-style type alias, rewrite it to the form using `=` and then check*). Otherwise (*when _X_ has no modifier, but occurs invariantly or both covariantly and contravariantly*), we say that _X_ _is invariant_.
```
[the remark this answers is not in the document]
leafpetersen: Either way, this definition needs to account for it.  As written, there's a contradiction between the parenthetical comment in the last sentence and the rest of the last sentence: the non-parenthetical part says "Otherwise we say that _X_ _is invariant_", but the parenthetical part says "when _X_... occurs invariantly or both covariantly and contravariantly".  The bi-variant case is covered by the first, but not the second.

plammens:
With this specification, is there a way to manually "override" or ignore (in the sense of doing a dynamic type cast) a wrong in/out position?

For example, consider the following class:

```dart
class Filter<in E> {
  Filter(this.predicate)

  final bool Function(E) predicate;

  Iterable<E> apply(Iterable<E> items) => items.where(predicate);
} 
```

Ideally the type parameter should be contravariant, because a `Filter<Fruit>` should be a `Filter<Apple>`, because if the predicate accepts fruits it certainly accepts apples.
But as written above it would be illegal according to the spec because `E` appears in an out position in the return type of the `apply` method.

However, we know that the `apply` method will never alter the actual runtime type of the given elements, as it is just a filter, so indeed a `Filter<Fruit>` would return `Apple`s with `apply` if you give it `Apple`s, and not any other type of `Fruit`.
This is probably not something the type checker can determine, so, would there be a way of ignoring this illegal "out" position, allowing `E` to be contravariant?

plammens:
Also I think `MyClass<SomeTypeArgument>` should always be a subtype of `MyClass<dynamic>`, regardless of variance of the type parameter, because `dynamic` is not some kind of root supertype but instead it just means disable static type checking for this parameter.

With the current `variance` experiment, I can't assign for instance a `Filter<int>` to a `Filter<dynamic>`.

lrhn:
> because `dynamic` is not some kind of root supertype 

It is, actually. The `dynamic` type is a top type, so it's a supertype of all Dart types.

If you have `class Filter<in T> { void add(T value) {}}`, and we allowed assigning a `Filter<int>` to `Filter<dynamic>`, then we get unsoundness:
```dart
var f1 = Filter<int>();
Filter<dynamic> f2 = f1; // Not allowed today, and isn't sound.
Filter<Object?> f3 = f2; // Definitely valid, not way to disallow.
Filter<String> f4  = f3; // Allowed because of contravariance.
f4.add("banana"); // Not sound.
```

The point of *sound* variance is to allow a call like `f4.add("banana")` to not do extra runtime type checks, because soundness ensures that any value assigned to `Filter<String>` will *definitely* accept a `String` as argument to `add`.

plammens:
> It is, actually. The dynamic type is a top type, so it's a supertype of all Dart types.

Yes, yes, my bad, I realized this later. 

In any case the type system is still unsound because of covariant type parameters by default 😅

munificent:
There is a variance spec already in the repo that's being actively worked on (https://github.com/dart-lang/language/tree/main/working/0524-variance), so I'm guessing this PR is (extremely) stale and can be abandoned?

PR #560, opened 2019-09-03
Specify induced setters/getters for late variables

### LINKED ISSUE #523: Do late final variables with an initializer induce a setter?
Based on input from @johnniwinther, we need to consider the following:

A late final variable may or may not have an initializer. With an initializer it supports lazy initialization; without an initializer it supports late initialization by assignment in the scope of the variable (but not subsequent mutation). But some corner cases aren't completely clear.

The NNBD feature specification says [here](https://github.com/dart-lang/language/blob/master/accepted/future-releases/nnbd/feature-specification.md#late-fields-and-variables) that we have the following error:

> A write to a field or variable which is marked `final` and `late` is a runtime error
> unless the field or variable was declared with no initializer expression, and there
> have been no previous writes to the field or variable (including via an initializing
> formal or an initializer list entry).

Replacing 'unless' by 'if' and negating the tail we get this form:

A write to a field or variable which is marked `final` and `late` is a runtime error if

- the field or variable was declared with an initializer expression, or
- there have been a previous write to the field or variable (including via an initializing formal or an initializer list entry).

It seems unlikely that we'd want to make it a run-time error to write to a variable based on syntactic criteria, because this could obviously be detected statically.

So we might prefer to change this such that it is a compile-time error to write to such a variable when it has an initializer (e.g., when it is declared as `late final int? x = 12;`). The motivation for this rule would be that the variable _should_ be initialized by the given initializer, and it is not allowed to do it by a plain assignment. (I cannot find evidence that such a write is already an error.)

Alternatively, we could say that it is not a compile-time error to initialize a late final variable with an initializer by a plain assignment (in fact, such an error does not seem to be specified today), and also not a run-time error (which would be a change). In this case, the initializer is a _backup_ mechanism that will kick in if and only if no initialization was performed using a plain assignment, at the time when the variable is read for the first time.

The first alternative would improve on static safety (by detecting some errors earlier); the second alternative provides fewer static guarantees, but it could be useful in practice. 

The approach which is most consistent with previous decisions in Dart is probably the first alternative, given that we do not allow an instance variable with an initializer to be initialized in a constructor (via an initializing formal or an entry in the initializer list). So we have previously made the choice against a 'backup' semantics; the upside of this as that developers can read the initializer and trust it.

If we specify that a late final variable with an initializer does not induce a setter then we get the first alternative (any assignments will simply incur a  "no such setter" error). If we specify that it does induce a setter then we can get the second alternative (that setter would then enforce that the variable is never mutated, but it can be initialized, either by evaluating the initializer or by the first assignment).

So I'd propose that we specify that a late final variable with an initializer does _not_ induce a setter.

@munificent, @leafpetersen, @lrhn, WDYT?

eernstg:
This CL adds one extra paragraph to the NNBD spec, specifying which setters/getters are implicitly induced by `late` variables.

Having this information explicit in the NNBD spec helps in a few ways:
- It enables us to rely on explicit rules in order to determine the treatment of `late` variables when used from an opted-out library _L_ (because _L_ will just rely on the getters, setters, no need for an 'erasure' operation on the declaration itself).
- It resolves #523. 

This CL also makes a couple of typo-level changes: Re-ordering keywords `late` and `final` to follow the grammar; adding parentheses on a type used in null-aware operator source transformations (following the convention that function types are right-associative).


accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -391,6 +391,13 @@ continuation.
 
 #### Late fields and variables
 
+A non-local variable declaration of the form `late final v;`, `late final T v;`,
```
leafpetersen: I think I'm in favor of not defining a setter here.

eernstg: What I'm proposing is that the forms with an initializer don't have a setter. But if the forms with _no_ initializer do not have a setter, how would they get initialized?

I would expect these forms to induce a setter (as specified in line 397), and that setter would perform the checks that enable dynamic initialization and prevent subsequent mutation.

Otherwise we'd need to come up with a substantially different specification of how to perform static checks and dynamic execution of assignments like `v = e`.

eernstg: Just noted that [this comment](https://github.com/dart-lang/language/issues/523#issuecomment-553192653) seems to express support for (1) `late final` with initializer does not induce a setter, and (2) `late final` without initializer does induce a setter, which is what this PR specifies. So maybe we agree already.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -391,6 +391,13 @@ continuation.
...
+`late final v = e;`, or `late final T v = e;`, or the same forms with `static`
+prepended, implicitly induces a getter into the enclosing scope. Moreover, the
+forms that do not include an initializing expression `e` induce a setter as
+well. For other variable declarations including `late`, the same getters and
```
lrhn: Maybe:

   other variable declarations &mapsto; non-`final` non-local variable declarations 

It's unclear what "other" covers. If my suggestion is not correct, then it definitely needs to be spelled out.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -410,24 +417,24 @@ read.
     is treated as a first read and the initializer expression is evaluated
     again.
 
-A write to a field or variable which is marked `final` and `late` is a runtime
+A write to a field or variable which is marked `late` and `final` is a runtime
```
lrhn: Is this a *dynamic* write?

Otherwise, if the variable has an initializer expression, then it's a *compile-time* error, not a run-time error.

SO maybe:

> A write to a field or variable which is declared `late` and `final` and with no intiializer expression, is a run-time error if there has been a previous write to the field or variable (including ...).

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -410,24 +417,24 @@ read.
...
 error unless the field or variable was declared with no initializer expression,
 and there have been no previous writes to the field or variable (including via
 an initializing formal or an initializer list entry).
 
-Overriding a field which is marked both `final` and `late` with a member which
+Overriding a field which is marked both `late` and `final` with a member which
 does not otherwise introduce a setter introduces an implicit setter which
 throws.  For example:
```
lrhn: That sounds spurious, and wrong. A final late field has a getter and a setter. Those are the ones you override.

Overriding the getter with another getter is valid, so it should also be valid to declare a non-late final field to override that getter.

Otherwise *all* final fields should shadow inherited setters with throwing versions. That's not a `late` property.

eernstg: > A final late field has a getter and a setter.

A final late field _with_ an initializer wouldn't have a setter.

But we could use the standard approach: Getters and setters are overridden by other getters/setters iff the latter exist.

So a late final field with an initializer has a getter and no setter, and the getter may be overridden just like other getters. A setter may be added to the interface by a subtype&mdash;no special exceptions needed.

For a late final field without an initializer there is a getter and a setter, and they can be overridden as usual.

The rule specified in line 418 differs from this by specifying that if a subclass overrides the getter and would inherit the setter then we auto-generate a setter that throws. We could do that, but I agree with @lrhn that we might as well avoid introducing a special exception for this case.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: Agreed.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: It is a compile time error if it is statically visible as such, but you can arrange (via overriding) for it not to be visible as such.  I'm not sure I follow the gist of your proposed change?

lrhn: I agree that my change technically says the same thing. At least after reading things a few extra times.

I guess my problem is, in general, with rules stated as "something happens, unless (long list of exceptions covering most actual situations)". I just find it harder to grok. Especially when some of those exceptions are compile-time errors, where it feels wrong to talk about run-time semantics of the code at all.

I am still not entirely convinced that the current text is not subtly wrong.
It seems to say that it is not a run-time error to write to a late final field which has an initializer expression. 
But you *cannot* write to such a field, it has no setter, so saying what happens or doesn't happen *when you write* is meaningless. 
It's not wrong, though. Mainly confusing.

accepted/future-releases/nnbd/feature-specification.md
eernstg: Note that my proposed change is a completely presentational matter: The grammar insists that variables have `late final` rather than `final late`, so I adjusted a couple of locations to use the same ordering.

So the discussion about the underlying rule is separate, and I'd prefer to take that discussion in a separate PR.

accepted/future-releases/nnbd/feature-specification.md
eernstg: My change on this line only ensures that `late` and `final` are used in the same order as they must be in a variable declaration.

With respect to the confusion caused by the fact that we have two cases:

-  The variable declaration has an initializing expression, so the failing operation is a dynamic invocation of a setter which does not exist (`noSuchMethod`).

- The variable declaration does _not_ have an initializing expression, so the failing operation is an invocation (static or dynamic) of a setter which throws (because there _has_ been a previous write, via a setter invocation or an initializing formal initialization, or an initialization due to a constructor initializer list).

I believe that the current text does cover these cases correctly.

Still, we could reduce the number of logical operations imposed on the reader by making the two relevant cases explicit. I changed the text in order to achieve this.

Note that the new text does not mention the case where there is an initializer: In that case there is no setter, and that is covered by standard rules both in the static and the dynamic case.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -417,10 +417,15 @@ read.
     is treated as a first read and the initializer expression is evaluated
     again.
 
-A write to a field or variable which is marked `late` and `final` is a runtime
-error unless the field or variable was declared with no initializer expression,
-and there have been no previous writes to the field or variable (including via
-an initializing formal or an initializer list entry).
+Let `v` be a `late` and `final` non-local variable without an initializing
```
leafpetersen: I'm not sure that I agree with this rewording, at least without being more explicit what we mean by "the setter".  This can be read as "to invoke the setter named `v=`", which I don't think is correct.  Consider:

```
class A {
  late final int v;
}
class B extends A {
  int v;
  B() {
    super.v = 3;
    v = 4; // This is an invocation of `v=`, but it is not an error.
  }
}
```

I'm thinking perhaps think the clearest way to do this is to specify this in terms of the induced setters, getters, and backing store.

eernstg: Just made it explicit which setter causes the run-time error; how do you like that?

leafpetersen: Sure, I can live with that.  I may follow up.

eernstg: Sounds good, thanks! Then I'll land it if Lasse is happy.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -391,6 +394,13 @@ continuation.
 
 #### Late fields and variables
 
+A non-local variable declaration of the form `late final v;`, `late final T v;`,
```
lrhn: I have a hard time reading this because it's not progressing logically. First it groups four declaration forms, then it says something for two of them, then it mentions "the rest" (non-final declarations, a syntactically simpler group) afterwards. I'm never sure what the thing I have just read really covers.

How about:

> A non-local variable declaration implicitly induces a getter into the surrounding scope.
> It also induces an implicit setter in the following case:
> * The variable declaration is non-final.
> * The variable declaration is final, late and has no initializer expression.
>
> \rationale{The final late no-initializer variable is special in that it is the only 
> final variable which can be assigned to, even if it can only be assigned once.}.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -410,24 +420,29 @@ read.
     is treated as a first read and the initializer expression is evaluated
     again.
 
-A write to a field or variable which is marked `final` and `late` is a runtime
-error unless the field or variable was declared with no initializer expression,
-and there have been no previous writes to the field or variable (including via
-an initializing formal or an initializer list entry).
+Let `v` be a `late` and `final` non-local variable without an initializing
```
lrhn: This conflates the "variable", the name and the declaration. Which is probably common in the spec, but still confusing to me :)

Maybe change "non-local variable" to "non-local variable declaration", and change "`v`" to "*d*" (a metavariable denoting the declaration) and add "with name `v`" to link a name to the declaration. Use "*d*" below instead of "`v`" when referring to the thing that may or may not have been assigned.

accepted/future-releases/nnbd/feature-specification.md
eernstg: That's more readable, done!

accepted/future-releases/nnbd/feature-specification.md
eernstg: Done.

Issue #387, opened 2019-06-03
Specify superinterface contravariance error


eernstg:
Cf. #113, the language specification needs to specify that it is a compile-time error for a super-interface to have a non-covariant occurrence of a type variable, e.g.:

```dart
class C<X> {}
class D<X> extends C<void Function(X)> {}
```

eernstg:
PR is https://github.com/dart-lang/language/pull/565.

PR #565, opened 2019-09-09
Added the "superinterface must be covariant" rule

eernstg:
This is a tiny thing which has been on the table for a long time: `class C extends List<Function<X>> {}` is an error, superinterfaces must be covariant.

Resolves issue #387.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -3831,7 +3831,7 @@ It is a compile-time error if a class $C$ has two superinterfaces
 that are different instantiations of the same generic class.
 \commentary{%
 For example, a class can not have
-both \code{List<int>} and \code{List<num>} as superinterfaces,
+both \code{List<\VOID>} and \code{List<\DYNAMIC>} as superinterfaces,
 directly or indirectly.%
 }
 
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -3841,10 +3841,11 @@ it is a compile-time error if $X$ occurs in a non-covariant position
 % Could say 'a direct superinterface', but it is easy to see that it is
 % enough to check direct superinterfaces, and it is then true
 % for indirect ones as well.
-in a superinterface of $C$.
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -3808,17 +3810,40 @@ \subsection{Superinterfaces}
...
+called to her or his attention.%
+}
+
+\LMHash{}%
+It is a compile-time error if a class $C$ has two superinterfaces
+that are different instantiations of the same generic class.
+\commentary{%
+For example, a class can not have
+both \code{List<int>} and \code{List<num>} as superinterfaces,
+directly or indirectly.%
```
lrhn: Worth mentioning `List<Object>` vs `List<dynamic>`? Is it allowed or not?

specification/dartLangSpec.tex
```diff
@@ -3808,17 +3810,40 @@ \subsection{Superinterfaces}
...
+}
+
+\LMHash{}%
+When a generic class $C$ declares a type parameter $X$,
+it is a compile-time error if $X$ occurs in a non-covariant position
+% Could say 'a direct superinterface', but it is easy to see that it is
+% enough to check direct superinterfaces, and it is then true
+% for indirect ones as well.
+in a superinterface of $C$.
```
lrhn: I think "in a superinterface" is ambiguous. It can mean in the superinterface type or in the superinterface member signatures or ....
Maybe:

> in a type parameter of a superinterface of $C$.

That's the only place it can occur, and it makes it clear(er) that we don't look through the superinterface signatures.

eernstg: Changed it to say 'in a type which specifies a superinterface'. It's dangerous to say 'in a type parameter' because this will swap the positions if the corresponding type parameter is contravariant.

specification/dartLangSpec.tex
eernstg: Done: `List<void>` and `List<dynamic>` is now mentioned as a conflict.

Issue #567, opened 2019-09-11
Specification of deferred loading `loadLibrary` function is lacking.


lrhn:
The specification doesn't actually specify the signature of `loadLibrary`, but just talks about "calling" it in the abstract, and says that the future completes "successfully", but not with which value, or that it completes with an error if loading fails (if it does). 

> We do not specify which object the future returned resolves to.

Not sure what that even means, perhaps just that we may return the same future object again, or perhaps "resolves" meant to say "completes".

The way the `loadLibrary` method on the prefix object is specified, it should be possible to tear it off and look at its run-time type.
The VM allows this, but dart2js does not, and the analyzer seems to consider it to have type `Function` or something (https://github.com/dart-lang/sdk/issues/38281).

The function should be specified to have an acual function type, and that function type should probably be `Future<void> Function()`, `Future<dynamic> Function()` or `Future<Null> Function()`.  The VM's type is `Future<dynamic> Function()` (both statically and dynamically) and it's completed with `null`. Keeing this type is probably the *simplest*. Changing it to a static type of `Future<void> Function()` is better design, but potentially breaking.

It should be specified that if loading fails, the returned future completes with an `Error`. That is never specified. Not sure it can even happen any more, but I guess it's possible for JS compilation where the second compilation artifact can go missing.

eernstg:
Cf. language PR #568.

sigmundch:
> The VM allows this, but dart2js does not

Note, both VM and dart2js allow it. This program prints `() =>  Future<dynamic>` on both:

```dart
import 'b.dart' deferred as b;
main() => print(b.loadLibrary.runtimeType);
```

For both the type is defined by how the lowering of loadLibrary is done in the CFE. There we inject a top level method called `__loadLibrary_prefix` that is then used to create the tearoff. We purposely typed it as returning dynamic when we first implemented this. See:

https://github.com/dart-lang/sdk/blob/b0fa0aebb7064b3af60a7504df832fe3bf4c9a44/pkg/front_end/lib/src/fasta/kernel/load_library_builder.dart#L58

eernstg:
Changing the return type to `Future<void>` would indicate the intended usage of `loadLibrary`. Changing it to `Future<Null>` would be truthful, but has the well-known problem that it can silently be assigned to a `Future<T>` for any `T` (with NNBD: any nullable `T`).

[...]

PR #568, opened 2019-09-11
Specify loadLibrary

eernstg:
The function `loadLibrary` currently has no specification for static analysis. This PR adds it to the relevant prefix objects, with type `Future<dynamic> loadLibrary()`.

We need to consider whether we want to use a more strict variant: `Future<void>` loadLibrary()`.

Cf. language issue #567.

### DIFF
--- a/specification/dart.sty
+++ b/specification/dart.sty
@@ -330,6 +330,9 @@
 \newcommand{\Show}[2]{\ensuremath{\ShowName({#1},{#2})}}
 \newcommand{\Hide}[2]{\ensuremath{\HideName({#1},{#2})}}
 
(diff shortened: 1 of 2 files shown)


eernstg:
Note that discussions in dart-lang/sdk#38379 support the use of `Future<void>` as the return type, and might help determine how breaking this would be.

eernstg:
Adjusted this PR to use return type `Future<void>` which may need a breaking change treatment (how much is actually broken by this change? probably very little working code), because this seems clearly preferable.

eernstg:
Closing, done as part of #1161, using type `Future<void>`.

PR #578, opened 2019-09-13
Add typedefs and generic function types to strict-inference

srawlins:
I missed both typedefs and "generic function types" (right terminology?) in the strict-inference spec. Both things can have omitted return types, and in both cases that return type is assumed to be dynamic.


srawlins:
Added!

PR #579, opened 2019-09-17
Add new specification for `.packages` file

lrhn:
Provides a single source of truth about the `.packages` file format.

Updates the existing spec by removing references to `packages` directories (dropped in Dart 2) and adding metadata support for language versioning. Also changes the wording from being a proposal with motivation and arguments for making a particular choice, to just a specification of a specific design.


leafpetersen:
Is this still active, or subsumed by the json proposal?

mit-mit:
Closing, we now have https://github.com/dart-lang/language/blob/master/accepted/future-releases/language-versioning/package-config-file-v2.md

PR #588, opened 2019-09-20
Make mixin application implement the mixin's interface.

lrhn:
The current specification of mixin application does not say that
the mixin application class implements the mixin's interface.
Instead it implements the implemented interfaces of the mixin declaration
directly.

This is clearly wrong, since the mixin application class must implement
the mixin interface (`is MyMixin` must work!).

When we add the mixin interface to the mixin application class as an immediate
super-interface, we no longer need to have the mixin's implemented interfaces
as immediate superinterfaces.

This change will affect the max-depth of the inheritance chain of the mixin
application if an implemented interface of the mixin was currently the longest
inheritance chain leading to the mixin interface. By inserting the
mixin interface, the max-depth increases by one. This may affect some
least upper bound computations.

(The fact that LUB is affected by completely opaque class hierarchy
refactorings, and that LUB is visible to end users in type inference,
suggests that the LUB computation is fundamentally broken and should be
scrapped before it hurts us more).

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -3481,7 +3481,7 @@ then the (implicit or explicit) superinitializer in $k$ is an error.%
 }
 
 \LMHash{}%
-The superinitializer that appears, explicitly or implicitly, 
+The superinitializer that appears, explicitly or implicitly,
 in the initializer list of a constant constructor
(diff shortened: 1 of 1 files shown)


lrhn:
@stereotype441 @Markzipan @leafpetersen FYI.

PR #591, opened 2019-09-23
Proposal for `package_config.json` file

lrhn:


accepted/future-releases/language-versioning/package-config-file-v2.md
```diff
@@ -0,0 +1,134 @@
...
+## New File Format
+
+The new file will be called `.packages.json` and stored in the same directory as the `.packages` file. The format is JSON text describing is a single JSON object.
+
+That object has, at least, the following properties:
+
+- `apiVersion`: A single integer marking the version of this format, currently `2` (counting `.packages` as the first). Tools should refuse to work on files with a larger version number than the ones they are aware of. Incremental changes will not require increasing the version, so new versions are not expected any time soon, but code should still prepare for it.
+
+- `defaultPackage`: A string containing a package name.
+- `packages`: A list of packages, each with the following properties:
```
munificent: Nit: "list" -> "array".

You probably want to be more explicit here. Something like:

> An array of objects. Each object has the following key-value pairs...

accepted/future-releases/language-versioning/package-config-file-v2.md
```diff
@@ -0,0 +1,134 @@
...
+  - `name`: A string containing a package name.
+  - either `path`: A string containing a relative or absolute path using `/` as separator. Every character other than `/` is used verbatim. If the path starts with a `/` or with a DOS/Windows drive letter (a single letter followed by `:/`), then it is considered absolute. If not, the path is relative to the directory of the `.packages.json` file.
```
munificent: I really think we don't want to be in the business of defining relative and absolute paths in this spec. (For example, what about UNC paths on Windows? Is a path starting with a drive letter considered an absolute path on non-Windows systems? Should it be?)

Do we need to support paths at all or can we get away with just URIs? They've already done the hard work of specifying relative and file URIs.

accepted/future-releases/language-versioning/package-config-file-v2.md
```diff
@@ -0,0 +1,134 @@
...
+
+- `defaultPackage`: A string containing a package name.
+- `packages`: A list of packages, each with the following properties:
+  - `name`: A string containing a package name.
+  - either `path`: A string containing a relative or absolute path using `/` as separator. Every character other than `/` is used verbatim. If the path starts with a `/` or with a DOS/Windows drive letter (a single letter followed by `:/`), then it is considered absolute. If not, the path is relative to the directory of the `.packages.json` file. 
+  - or `uri`: A URI reference. This is resolved relative to the (likely `file:`) URI referencing the `.package.json` file. If the URI reference has no scheme, or if it has a `file:` scheme, it is recommended to use the `path` instead.
+  - `languageVersion`: Optional. A string containing a language version (major.minor).
```
munificent: You'll need to specify the version more precisely. Alternatively, if you don't want to embed a mini language inside one of the JSON fields, you could do something like:

```json
{
  "language": {
    "version": {
      "major": 1,
      "minor": 2
    }
  }
}
```

That's pretty gross. :( But it does increase the odds of any random tool correctly parsing a version number. Also, it avoids users/tools accidentally writing:

```json
"languageVersion": 1.2
```

Which definitely does *not* mean what they think it means.

Also, I think "dart" is more useful than "language" here. We could conceivably extend this to put stuff related to other languages in here, but "Dart" is pretty precise.

accepted/future-releases/language-versioning/package-config-file-v2.md
```diff
@@ -0,0 +1,134 @@
...
+
+Tools should ignore properties that they don't recognize. This will allow us to add more properties in the future without increasing the `apiVersion` number, but not to remove properties or change the meaning of existing properties.
```
munificent: Also, tools that read in and then output a `.packages.json` file should preserve all data in the file even if the tool itself doesn't use it. This way the tool doesn't accidentally strip out extensions used by other tools.

accepted/future-releases/language-versioning/package-config-file-v2.md
```diff
@@ -0,0 +1,134 @@
...
+The following properties allows the generator to specify metadata about the file:
+
+- `generated`: A string containing an ISO-8601 UTC timestamp for when the file was generated, using the format: `YYYY-MM-DDTHH:mm:ss.sssZ`.
+- `generator`: An object containing at last the following properties:
+  - `name`: Name of the generator. Likely `pub`.
+  - `version`: The version of the generator. A [Semantic Version](https://semver.org/spec/v2.0.0.html). Pub can probably use the SDK version.
+
+These are all optional, but if they are present, they should have the expected format.
+
+If a tool wants to store extra information in the `.packages.json` file, they can do so under a key prefixed by `toolname:`, so if Pub wanted to store more information on a package, say the package version, it could do so as `"pub:pkgVersion": "1.16.0"` in the package entry. Keys containing `:` are reserved for custom tool-specific properties. This feature can only really be used by the tool generating the file (as a generated file, it may be overwritten at any time), and it should be used sparingly since unknown properties are just overhead for all other users of the file.
```
munificent: Let's not reinvent XML namespaces, poorly. I think it's sufficient to say that tools that want to store extra information should generally do so under a key that corresponds to or contains the tool's name, and authors should ensure that name is widely available in the ecosystem before claiming it.

We've never had a real problem with key conflicts in pub as far as I know.

accepted/future-releases/language-versioning/package-config-file-v2.md
```diff
@@ -0,0 +1,134 @@
...
+  - `languageVersion`: Optional. A string containing a language version (major.minor).
+
+Tools should ignore properties that they don't recognize. This will allow us to add more properties in the future without increasing the `apiVersion` number, but not to remove properties or change the meaning of existing properties. 
+
+If we remove a property that tools rely on, change the meaning of an existing property, or add more information that is *required* for resolving and understanding Dart source files, it is a breaking change&mdash;tools expecting the previously available data will no longer understand the whole picture. Such changes will require increasing the `apiVersion` number.
+
+The following properties allows the generator to specify metadata about the file:
+
+- `generated`: A string containing an ISO-8601 UTC timestamp for when the file was generated, using the format: `YYYY-MM-DDTHH:mm:ss.sssZ`.
+- `generator`: An object containing at last the following properties:
```
munificent: Schema bikeshed: How about combining these under one key:

```json
"generated": {
  "tool": "pub",
  "on": "YYYY-MM-DDTHH:mm:ss.sssZ",
  "version": "1.2.3",
}
```

accepted/future-releases/language-versioning/package-config-file-v2.md
```diff
@@ -0,0 +1,134 @@
...
+
+We use JSON for the format. JSON has parsers available for most languages (not all Dart tools are written in Dart) and it is standardized and efficiently parsable.  JSON is *not* easy to write by hand (but doable), and it does not allow comments, but this is a machine generated file that will hardly ever be edited by hand.
+
+## New File Format
+
+The new file will be called `.packages.json` and stored in the same directory as the `.packages` file. The format is JSON text describing is a single JSON object.
+
+That object has, at least, the following properties:
+
+- `apiVersion`: A single integer marking the version of this format, currently `2` (counting `.packages` as the first). Tools should refuse to work on files with a larger version number than the ones they are aware of. Incremental changes will not require increasing the version, so new versions are not expected any time soon, but code should still prepare for it.
```
natebosch: This precludes the possibility of writing 2 formats during a transition period for format upgrades. I worry that future upgrades will be even harder than this one where we will be writing both `.packages` and this file and can smoothly flip over a tool at a time.

accepted/future-releases/language-versioning/package-config-file-v2.md
```diff
@@ -0,0 +1,134 @@
...
+
+- `defaultPackage`: A string containing a package name.
```
natebosch: I think this refers to the same concept that we sometimes call "root package" in the build system and docs. Can we consider using that name here? Can we expand this description to explain exactly what this means and how tools should interpret it?

accepted/future-releases/language-versioning/package-config-file-v2.md
```diff
@@ -0,0 +1,134 @@
...
+  - A *path* or *location URI*.
+  - A *language version* (major/minor version number)
+
+On top of that, we also add extra metadata which allows us to version the format and attach information about when and how the file was created.
+
+We use JSON for the format. JSON has parsers available for most languages (not all Dart tools are written in Dart) and it is standardized and efficiently parsable.  JSON is *not* easy to write by hand (but doable), and it does not allow comments, but this is a machine generated file that will hardly ever be edited by hand.
+
+## New File Format
+
+The new file will be called `.packages.json` and stored in the same directory as the `.packages` file. The format is JSON text describing is a single JSON object.
```
pulyaevskiy: Just a naming nit to consider: 

`.packages.json` looks so similar to Node's `package.json`, so when both are in the same directory (which does happen) I can see how it can be very confusing.

I understand that this file is not specific to Pub so calling it `pubspec.packages` could be missleading too. But something like `dart.packages` (or `dart.packages.json`) would make it much more explicit.

lrhn: Naming is still open, but the name will definitely not be `packages.json`.

accepted/future-releases/language-versioning/package-config-file-v2.md
lrhn: I don't like "root package" since I don't know what it means (there is no tree structure, so why is it a root). It might commonly be used for the "current" Pub package, but this file is ignorant of Pub (it will be generated by other tools too). 

This is the package to use for files which do not otherwise belong to a package. It allows us to attach properties to non-package files which we otherwise attach to packages.

accepted/future-releases/language-versioning/package-config-file-v2.md
lrhn: Interesting idea. 
To have two formats in the same file, we would have to have two separate objects.
Say:
```
{ 
  "v2": {
    ...
  },
  "v3": {
    ...
  }
}
```
and tools should use the most recent version number they are familiar with.

I'm not sure we *want* that, though.

If the new version contains information essential for understanding the program, then the old version plain just doesn't work. Supplying it will not make things work anyway.

If the new version does not contain information that is essential, just useful, then it's an incremental change that doesn't require incrementing the version number to begin with.

So, I can't find a use-case where this would be both necessary and successful.

accepted/future-releases/language-versioning/package-config-file-v2.md
lrhn: Sure. I was thinking that all generators should have the "generated" time, but not all tools may want the rest. Still, this is probably a good idea.

accepted/future-releases/language-versioning/package-config-file-v2.md
lrhn: Reasonable.

accepted/future-releases/language-versioning/package-config-file-v2.md
lrhn: I deliberately avoided that. Do you think it is a good idea? I'm not sure I *want* tools to put arbitrary data into the file. If we do that, we'll likely see `analysis_options.yaml` getting moved into this file (I mean, why not, the analyzer will be reading the file anyway?). 
Is this what we *want*? An arbitrary registry of metadata for dart files in the related subdir?

We would also have to specify what it means, exactly, to "preserve data". Should Pub retain a package entry for a package it doesn't know about (any more)? What if that entry contains a property unknown to Pub? Most likely we will say that existing package entries *can* be removed by any tool generating package entries, but if the same package still exists, then any extra properties that Pub doesn't overwrite should be retained.

Should Pub retain extra information in the `generator` entry, or is that entirely its own?

accepted/future-releases/language-versioning/package-config-file-v2.md
lrhn: I loathe unnecessary verbosity, so let's define the format:

> A string containing a decimal numeral, a `.` and another decimal numeral, where the decimal numerals have no leading spaces unless the numeral is precisely `0`. Any string matched in its entirety by the RegExp `(0|[1-9]\d*)\.(0|[1-9]\d*)`.

That's how we specify language version inside Dart files in `// @dart = x.y` comments, and it's also a valid major/minor version of a SemVer version.

We probably won't accept double entries at all, not because of `1.2`, but because of `1.20`. It must be a string.

accepted/future-releases/language-versioning/package-config-file-v2.md
lrhn: One of the problems we had with the original format is that Pub always generated URI references that were just relative (to the file) or absolute (pointing to the pub cache) paths. 3rd party tools would read and understand them as plain *paths*, not URI references. They were apparently completely successful, so it seems it is all we actually need.

We an say that the "path" is a *URI path*, but we risk users *still* treating it as a plain path because in 99.99% of the the cases it's the same thing.
The last case is, say, when the pub cache in `C:\Users\Lasse Nielsen\LocalData\.pubcache`, and the path entry for a package becomes `C:/Users/Lasse%20Nielsen/LocalData/.pubcache/mypkg-1.0.0/lib`. Any tool using that as a plain OS path will break because `Lasse%20Nielsen` is not a directory.

So, we have to say whether the content is a URI reference, a URI path, or an OS Path. There are cases where the three differ, and using one a another will give the wrong result, fx  `c:/something`, relative to `file:///c:/src/mypkg/.dart_data/something.json`.
- If treated as URI, `c` is a scheme, and the result is the URI `c:/something`
- If treated as URI path, it is a *relative* path, and the result is file:///c:/src/mypkg/.dart_data/c:/something`
- if treated as OS path, it's an absolute path (and likley the intended approach).

We need to make sure that we specify *exactly* how to interpret the content, and knowing that 3rd party readers may treat it as a OS path, we could just say that it is one.
If we do that, then we should probably not begin defining what's relative and what's not, and leave that to the OS itself (or code that knows which OS it's on).

I don't like the "URI path" approach. It's cheating, really, and just increases the risk of someone treating it as a path, not a URI path and getting caught by `%20` escapes or similar.

So, the other reasonable approach is to say that it is a URI reference. In that case, it should be named `uri` (just for explicitness), and I'd like if Pub wrote `file://` in front of absolute paths (maybe all of them, maybe some of them, just to keep people on their toes), just to make sure that people who think everything is a path gets caught.

So, please vote :+1: for OS path, :heart: for URI reference :smile: .

accepted/future-releases/language-versioning/package-config-file-v2.md
lrhn: Using JSON terminology is probably a good idea.

accepted/future-releases/language-versioning/package-config-file-v2.md
natebosch: > I can't find a use-case where this would be both necessary and successful.

From my perspective it matches the situation we are in today.

We have many tools that consume the file and few tools that write the file. We were unable to identify and update all the tools that read the file in time to make them compatible with the new version. We chose to go for a smooth rollout and write both versions for some period of time. We specifically chose not to do a hard flip. This new scheme _only_ allows for a hard flip.

How do you envision version changes happening in the future? How will they be different from today such that a "write both" strategy is not necessary?

accepted/future-releases/language-versioning/package-config-file-v2.md
munificent: > I don't like "root package" since I don't know what it means (there is no tree structure, so why is it a root). 

It is the root of the dependency graph. It is the package that contains the entrypoint being invoked, which in turn imports all other libraries used by the application. I think "root" is a reasonable word for that.

"Default" implies that the implementation needs to "select" a package for some purpose and in the absence of other information it selects this one. I guess that is your mindset — that stuff outside of "lib" is not in any package so when it needs to guess at the package for it, it picks the default. But I don't think that's how any significant fraction of users look at the system. To users, all files under a package's root directory, even outside of "lib" are clearly and explicitly part of that package.

For example, given:

```
foo/
  lib/
    foo.dart
  example/
    foo_app.dart
bar/
  lib/
    bar.dart
  example/
    bar_app.dart
```

And a packages.json file like:

```json
{
  "defaultPackage": "foo",
  "packages": [
    {
      "name": "foo",
      "path": "foo/"
    },
    {
      "name": "bar",
      "path: "bar/"
    }
  ]
}
```

If a user somehow contrived to invoke `bar/example/bar_app.dart` using that
"packages.json" file, they would be very surprised if "bar.dart" was interpreted at the language version specified by *foo*.

You might argue that they are only surprised because they don't understand how the underlying model works. But I think the reality is that the underyling model where "package" means something different from the user-facing notion of "package" they have been taught for year is misleading. We shouldn't declare that their brains are wrong. We should design the system to be consistent with what's in their brains.

Maybe "default package" is the wrong notion entirely. All that does is determine the language version, right? How about we state that explicitly? Something like:

```json
{
  "dart": {
    "version": "2.1",
  },
  "packages": [
    {
      "name": "foo",
      "path": "foo/"
    },
    {
      "name": "bar",
      "path: "bar/"
    }
  ]
}
```

?

accepted/future-releases/language-versioning/package-config-file-v2.md
munificent: One key problem with OS paths is that they are not context free. You don't know what the string "foo\bar" means unless you know which host operating system it is intended for. That alone to me is a big enough problem that I think it's better to say these are URIs.

> So, the other reasonable approach is to say that it is a URI reference. In that case, it should be named uri (just for explicitness), and I'd like if Pub wrote file:// in front of absolute paths (maybe all of them, maybe some of them, just to keep people on their toes), just to make sure that people who think everything is a path gets caught.

+1 to all that, including naming it "uri".

accepted/future-releases/language-versioning/package-config-file-v2.md
munificent: > I deliberately avoided that. Do you think it is a good idea?

Yes, it's essentially required. If we allow tools to store their own extensions in the file (and I think we do want to do that), then we need to ensure other tools roundtrip those files must preserve those extensions.

In practice, I don't think it will be much of an issue. This file isn't intended to be the source of truth for data. Most tools will either only read it or only write it. Tools that do generate this file from scratch are not obliged to maintain extensions put in some previous incarnation of the file generated by a different tool. It's basically on the user to ensure that two tools which generate the same file somehow coordinate with each other.

lrhn: Ah, so you are not talking about pub, but explicitly about tools which modify the Pub-created file for their own purposes (adding data that only they understand).

If Pub does not retain extra data, but creates it from scratch, then other tools which add something on top are at Pub's mercy. Since Pub may be called at any time by your editor if you touch `pubspec.yaml` for any reason, you can't rely on that extra data anyway (unless those tools are called *by* Pub every time it generates the file.

If we want to preserve extra data, I think even Pub should do so, but that risks accumulating cruft over time. At least users can always delete the file and start over.

I'm not sure we *do* want to allow tools to store their own extensions in the file. It is not the Windows Registry for Dart tools, the file has a purpose. Unless we want to have a registry for  Dart tools. Then retaining the unknown content should be

accepted/future-releases/language-versioning/package-config-file-v2.md
lrhn: The current change would be a non-breaking in the new format.

The only operation supported by the previous format was to resolve a `package:` URI to a non-`package:` URI. The information needed to do that is available in the new version too, so old code is not broken.

We now add more information, which is used to tell tools how to interpret the files they find.

Future changes are either incremental and non-breaking, or breaking.

If we add more information, but nothing so essential that existing uses of the file will stop working, then it's a non-breaking change. We do not need a new version number.

If we add more information which is necessary for tools to know how files are interpreted, then it's probably a breaking change. Old tools, using only the old information, would no longer be able to correctly interpret the Dart files they find. That's where we need to increment the version, because old code will now do *wrong* things.

Keeping the old information available in the old format, along with new information in a new format, will *not change that*. The old tools are *still broken* because they are simply not capable of handling the new world. I hope that such changes will not happen, or that they will be rare, but I do think that keeping two formats at the same time will *not help*.

That said, we can always have a transitional phase with:
```dart
{
  "apiVersion": 2,
  "forwardCompat": {
    "apiVersion: 3,
    "new data here": ...
  },
  "old data here": ...
}
```
and require version-n readers seeing version m<n data to look for a "forwardCompat" entry which may hold their data. Being a general format allows us such shenanigans. I'd prefer to not decide the exact way to handle this until it becomes necessary, I'll just remember that it is possible if it becomes necessary (and if it actually solves any problem).

accepted/future-releases/language-versioning/package-config-file-v2.md
lrhn: The issue here is that there are different models in play: The Dart package and the Pub package.

The Dart language has no notion of root package or dependency graphs. It has a finite set of available Dart packages, which each contains a hierarchical structure of available Dart files that you can find using `package:` URIs. Then you run one of the available scripts to start a program. You can start more scripts, from different packages, using `Isolate.spawnUri`. Tree shaking is a compiler optimization that ensures that your deployed code only contains what is reachable from an entry point.

The Pub package is more than that, but it morally belongs to the same Pub package, and since that is what users are thinking about when they develop a package, the distinction between a Pub package and a Dart package is confusing. 

The Pub package concept is *so far* not reflected in the language or compiler tools, which is not surprising since it was developed *after* the language.

*We could change this now*. We could say that the `package_config.json` file will point to a *Pub* package directory (not its `lib/` dir), package URI resolution would use the `lib/` dir, but language versioning would apply the language version to all files in the Pub package directory tree.
I think that would be compatible with the current Pub cache layout (it contains `lib` dirs) and the Google Dart code layout (a Google Dart package has a root dir and a `lib/` dir as well).

It would hard-code the Pub layout (at least the `lib/` sub-dir) into tools which are currently agnostic to how packages are stored.

Or we could make the "lib" path configurable and have package entries like:
```json
{
  "name": "foo",
  "rootUri": "../", /* relative to JSON file */
  "libUri": "lib/", /* relative to rootUri */
  "languageVersion": "2.5"
}
```
where the `rootUri` is what we use to determine whether a file belongs to the package (is it in a subdir of this location, then it belongs to the package, and if two packages overlap, we'll resolve it somehow, perhaps by flipping a coin), and the `libUri` is a directory that we resolve `package:` URIs against (and which *must* be a subdir of the `rootUri`, no cheating!). That allows us to expand the "belongs to a package" to a directory tree (heck, we can add "includes" and "excludes" for subdirs as well if we want to, and do fancy Boolean logic to decide whether a file is "in the package" or not), and not just to the subdirectory where the Dart package files reside.

Or we can say that if `libUri` is absent, just use `rootUri`, and make `rootUri` point at the `lib/` directory for all other packages, and only care about the extra files of the current package. Pub can do what it wants here, this will give it tools to say what the "Pub package" is, and not just the "Dart package" (but it also has to say that, obviously, we do need to resolve `package:` URIs).

Should we do something like this?


(I added the "default package" because I want to allow the files in a Pub package to be considered part of the package for other (potential) future features, like package privacy or package sealed types. It was the *minimal* feature which allows one (*current*) package to have its test/ and other directories be considered in same package when you run your tests. The suggestion here would satisfy that, while also allowing a more general use. I just want to make absolutely sure that nobody will be in a position to claim that their own code is somehow part of a third-party package to get access to package private code.)

accepted/future-releases/language-versioning/package-config-file-v2.md
lrhn: I tend to agree. Saying that it is a URI reference will make things easier. Even if I say that the path uses `/` as separator, it won't prevent `"ab/cd\ef"` from being an interpretation nightmare.

So, yes, URIs. Not paths, because paths are not standardized.

accepted/future-releases/language-versioning/package-config-file-v2.md
munificent: > Ah, so you are not talking about pub, but explicitly about tools which modify the Pub-created file for their own purposes (adding data that only they understand).

Pub isn't a good example because it never reads any file that it also writes. Pub only reads pubspec.yaml files and it only writes pubspec.lock files. Thus it never roundtrips a file or has to worry about preserving extensions.

But tools like stagehand that have commands for modifying a pubspec *do* read and write the same file and have to make sure to preserve data.

accepted/future-releases/language-versioning/package-config-file-v2.md
munificent: > The Dart language has no notion of root package or dependency graphs.

For a given program invocation, there is a single library that is the entrypoint where execution began. All other libraries are reached from the import/export graph starting at that library. That sounds like a root package and a dependency graph to me.

> The Pub package concept is _so far_ not reflected in the language or compiler tools, which is not surprising since it was developed _after_ the language.

Well, *sort of*. The "package:" import scheme was added to the language by Kasper et al. specifically at the behest of Natalie and I because we were in the middle of building pub and needed it. Dart existed before pub, but the language's notion of packages did not.

> _We could change this now_. We could say that the `package_config.json` file will point to a _Pub_ package directory (not its `lib/` dir), package URI resolution would use the `lib/` dir, but language versioning would apply the language version to all files in the Pub package directory tree.

Not a bad idea. The fact that the language tries to define its own somewhat different notion of package has never seemed practically useful to me.

> It would hard-code the Pub layout (at least the `lib/` sub-dir) into tools which are currently agnostic to how packages are stored.

I'm a little hesitant to bake directory policy right into the language. We agonized over how "package:" imports should map to the interior of pub packages for a long time before settling on "lib" in large part to avoid symlink cycles. It's a *useful* policy, and I think a pretty good one, but I don't know if we want to be stuck with it forever.

> the `libUri` is a directory that we resolve `package:` URIs against (and which _must_ be a subdir of the `rootUri`, no cheating!).

Interesting idea! I do really like the idea of having the language's notion of "package" include the entire Pub package directory. Like you note, that may come into play with things like package privacy. It would be *really* nice if, for example, tests could have some kind of privileged access to things in the same package's libraries.

It might be worth thinking about a little more but my initial feeling is that I like it.

accepted/future-releases/language-versioning/package-config-file-v2.md
```diff
@@ -0,0 +1,183 @@
...
+      "languageVersion": "2.5"
+    },
+    {
+      "name": "test",
+      "path": "/users/myself/.pubcache/test-1.16.0/lib/",
+      "languageVersion": "2.5"
+    },
+    ... 
+  ],
+  "generated": "2019-09-12T12:13:14",
```
jakemac53: [nit] this doesn't match the spec above

accepted/future-releases/language-versioning/package-config-file-v2.md
```diff
@@ -0,0 +1,183 @@
...
+
+That object has, at least, the following properties:
+
+- `apiVersion`: A single integer marking the version of this format, currently `2` (counting `.packages` as the first). Tools should refuse to work on files with a larger version number than the ones they are aware of. Incremental changes will not require increasing the version, so new versions are not expected any time soon, but code should still prepare for it.
+
+- `defaultPackage`: A string containing a package name.
+- `packages`: An array of packages, each with the following properties:
+  - `name`: A string containing a package name.
+  - either `path`: A string containing a relative or absolute path OS specific path using `/` as separator. If the path is relative, it is relative to the directory of the `package_config.json` file. 
+  - or `uri`: A URI reference. This is resolved relative to the (likely `file:`) URI referencing the `.package.json` file.
```
jonasfj: I saw the discussion of _OS Path_ vs _URI_, and I see that `path` is OS specific, I think that's okay because this file is inherently OS specific, it can't be moved from one machine to another.

But I'm also okay to use `uri` instead of _OS path_, I suspect some tools will resort to regex parsing the URI, but that'll probably continue to work just as fine as it does now :see_no_evil: 

At-least using the `uri` property name will make it more obvious that it's a URI :)

lrhn: The path is not inherently OS specific, and in practice it needs to not be so.
The Dart SDK has a checked-in `.packages` file which works (and needs to wok) on any platform.

accepted/future-releases/language-versioning/package-config-file-v2.md
lrhn: Thanks.

accepted/future-releases/language-versioning/package-config-file-v2.md
```diff
@@ -0,0 +1,183 @@
...
+  - `languageVersion`: Optional. A string containing a Dart language version consisting of a decimal numeral, a `.` character and another decimal numeral, where the decimal numeral must not have a leading `0` unless the numeral is exactly `0`. That is, a string matched in its entirety by the regular expression `(0|[1-9]\d*)\.(0|[1-9]\d*)`. *This is the same format accepted by `// @dart = x.y` comments in source files.*
+
+Tools should ignore properties that they don't recognize. This will allow us to add more properties in the future without increasing the `apiVersion` number, but not to remove properties or change the meaning of existing properties. 
+
+If we remove a property that tools rely on, change the meaning of an existing property, or add more information that is *required* for resolving and understanding Dart source files, it is a breaking change&mdash;tools expecting the previously available data will no longer understand the whole picture. Such changes will require increasing the `apiVersion` number.
+
+The following properties allows the generator to specify metadata about the file:
+
+- `generated`: An object containing one or more of the following entries:
```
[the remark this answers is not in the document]
lrhn: That sounds reasonable. I've gone with the flattened style for generator.

accepted/future-releases/language-versioning/package-config-file-v2.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Background
+
+Dart currently uses a `.packages` file to configure *package URI resolution* (converting a package URI to a file name). The file uses a `.ini` file-like format with package names as keys and URI references as values. The URI references generated by Pub are always absolute or relative paths, so third-party users of the file tend to see them as paths, not URI.
+
+For the [*language versioning*](feature-specification.md) feature, we have proposed extending the format by adding fragments to the URI references and an entry with the empty string as key. This is a breaking change (the empty key was not previously accepted) and even more breaking for tools which assume the values are plain paths.
+
+Instead of breaking the existing file users, we will instead introduce a *new* file containing the information that we intended to store in the `.packages` file. This fill will use a structured format which allows extending it with further information in the future. We will still generate the `.packages` file for now, allowing users of the file to continue working until they can migrate to using the new file.
```
leafpetersen: fill -> file.

accepted/future-releases/language-versioning/package-config-file-v2.md
```diff
@@ -0,0 +1,193 @@
...
+
+The following optional properties of the top-level object allows the generator to specify metadata about the file. They can all be omitted, but if they are present, they should have the expected format.
+
+- `generated`: A string containing an ISO-8601 UTC timestamp for when the file was generated, using the format: `YYYY-MM-DDTHH:mm:ss.sssZ`, for example `"generated": "2019-09-24T12:34:56.000Z"`.
+- `generator`:  A string with a name/identifier of the generator which created the file, typically `"pub"`.
+- `generatorVersion`: A string with the version of the generator, if the generator wants to remember that information. The version must be a [Semantic Version](https://semver.org/spec/v2.0.0.html). Pub can use the SDK version. Example: `"generatorVersion": "2.5.0"`.
+
+The JSON properties may occur in any order, except that the `apiVersion` *must* be the first entry in the file. The Dart JSON serializer uses iteration order of the incoming map, so controlling the order is possible. The file *should* be "pretty printed" with newlines and indentation, to make human inspection easier, but tools consuming the file must accept JSON with or without any optional whitespace.
```
leafpetersen: What does this buy us?

accepted/future-releases/language-versioning/package-config-file-v2.md
```diff
@@ -0,0 +1,193 @@
...
+
+#### Extra Information in the File
+
+If a tool wants to store extra information in the `package_config.json` file, they can do so under a key which corresponds to or contains the name or identifier of the tool. Authors should ensure that the name is available in the ecosystem before claiming it for their tool. If Pub wanted to store more information on a package, say the package version, it could do so as `"pubPkgVersion": "1.16.0"` in the package entry. 
+
+This feature can only really be used by the tool generating the file (as a generated file, it may be overwritten at any time), and it should be used sparingly since unknown properties are just overhead for all other users of the file. Still, if a tool can cooperate with Pub and provide the information in `pubspec.yaml`, Pub might be able to put it into the configuration file as well.
+
+Any tool which modifies the file piecemeally by changing or add information, rather than writing it from scratch like Pub, should retain all properties that it does not recognize. This allows multiple tools to all cooperate about adding information to the file.
```
leafpetersen: The adverb form of piecemeal is piecemeal, because English.  ¯\_(ツ)_/¯

accepted/future-releases/language-versioning/package-config-file-v2.md
```diff
@@ -0,0 +1,193 @@
...
+### Specifying the File
+
+Tools currently allow you to specify the `.packages` file with a command line argument:
+
+```
+--packages=someFile
+-p someFile
+```
+
+This approach will work with the JSON configuration file as well. The tool must scan the file to detect whether it is a JSON file or an original `.packages` file. It is considered a JSON file if the first non-whitespace (space, tab, carriage return, linefeed) character is a `{`. The JSON file must start this way, and the existing format cannot do so; it must start with either a comment, `#`, or a non-empty package name, and package names cannot contain a `{` character.
```
leafpetersen: What do you think about allowing tools to look for a .dart_tools/package_config.json file in the same directory as the specified .packages file and use that instead?

leafpetersen: Alternatively, what about having tools (e.g. pub) which are generating both files mark the .packages file with a redirection marker so that other tools which are aware of both conventions can redirect to the non-deprecated version if they are pointed at the deprecated version?  This could possibly go in the initial comment if we wanted to make it completely non-breaking.

accepted/future-releases/language-versioning/package-config-file-v2.md
lrhn: Which "this"?

Putting the `apiVersion` first allows parsers to be more efficient than parsing the entire file first and then reading the intermediate maps. If the `apiVersion` changes, then you can know immediately which following format to expect, and use a faster parser specialized for that format ... if you want to. I like to have the option.

As for pretty-printing, it's just easier to read for humans.

accepted/future-releases/language-versioning/package-config-file-v2.md
lrhn: I believe there was widespread support for keeping the file out of the package root directory at the meeting where the location was decided. We can allow you to put it there, but then we'd have to decide which of `package_config.json` and `.dart_tool/package_config.json` takes precedence. If Pub is always going to generate the file in `.dart_tool`, then I'm not sure there is an advantage in also allowing it elsewhere.

I'm not sure which problem the redirection is trying to solve. Is it the case where a tool is explicitly pointed to the `.packages` file, fx with a command line argument like `-p .packages`, and then it figures out to use the `.dart_tool/package_config.json` instead because of the redirect?

We could make our configuration-file-finding-algorithm automatically check for `.dart_tool/package_config.json` when it detects that it's being asked to load a `.packages` file. It's currently forced to detect the content by pattern matching the syntax (luckily it can be seen from the first character).

leafpetersen: Sorry, I wasn't proposing to change the location.  The issue I'm trying to solve is that there seem to be many (external) tools that pass the location of a .packages file explicitly to another tool.  For example, I believe that many of the uses in Flutter that @jakemac53 found are in this category: they neither generate nor parse the contents of .packages, they just say "here's the .packages file!".  If the eventual tool which does parse the contents has already been updated, it seems like it might be valuable for that tool to be able to say "Ah, you have pointed me at a .packages file, but I see that there is a .dart_tool/package_config.json file in the same directory, so I will just use that instead".  But if that is considered too surprising or error prone, then we could instead only do that if the .packages file explicitly says "Tools that have been updated should use this other file, otherwise here's the legacy contents".  Hence the alternative proposal for a redirection.

accepted/future-releases/language-versioning/package-config-file-v2.md
leafpetersen: Sorry, I meant the apiVersion.  Requiring this means that you can't just use an off the shelf json serializer, right?  I guess we don't expect there to be too many generators, but still.  Maybe if we want this first, then the outermost json object should be an array?

accepted/future-releases/language-versioning/package-config-file-v2.md
```diff
@@ -0,0 +1,193 @@
...
+
+The format will look like:
+
+```json
+{
+  "apiVersion": 2,
+  "packages": [
+    {
+      "name": "myPackage",
+      "rootUri": ".",
```
jonasfj: This needs to be `../` if we assume that `package_config.json` eixsts in `./dart_tool/package_config.json` and according to this spec:
> The URI reference is resolved relative to the (likely file:) URI identifying the package_config.json file.

accepted/future-releases/language-versioning/package-config-file-v2.md
lrhn: Added an "if you are pointed to a .packages file, you *should* check for an adjacent .dart_tool/package_config.json file" clause.

accepted/future-releases/language-versioning/package-config-file-v2.md
lrhn: ACK.

accepted/future-releases/language-versioning/package-config-file-v2.md
lrhn: Changed to a recommendation instead of a requirement.

accepted/future-releases/language-versioning/package-config-file-v2.md
lrhn: Good catch.

accepted/future-releases/language-versioning/package-config-file-v2.md
lrhn: So I did it.

PR #594, opened 2019-09-24
Changed invocation of non-function objects to use a `call` getter

eernstg:
Cf. issue https://github.com/dart-lang/sdk/issues/36420: All implementations except DDC will invoke a getter named `call` or a method named `call` in the case where a function expression invocation (dynamic or statically checked) is performed where the target is not a function object.

The language specification currently specifies that only a method named `call` can be invoked in this manner (so if a getter is found it is an error, dynamic or static).

This PR makes the necessary changes to the specification to allow using a getter (to obtain a function object which is then invoked).

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -9081,16 +9081,17 @@ the returned result is then the result of evaluating $i$.
 
 \LMHash{}%
 Otherwise $o$ is not a function object.
-If $o$ has a method named \CALL{}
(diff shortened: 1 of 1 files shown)


eernstg:
Note discussion in dart-lang/sdk#36420, including https://github.com/dart-lang/sdk/issues/36420#issuecomment-545402561, which may lead to acceptance or rejection of this change.

PR #595, opened 2019-09-26
Update strict-inference re: return types of setters, overriding methods

srawlins:


resources/type-system/strict-inference.md
```diff
@@ -297,12 +297,14 @@ void main() {
 
 ### Function return types
 
-Declaring a recursive local function, a top-level function, a method, a
-typedef, a generic function type, or a function-typed function parameter
-without a return type is an inference failure. The return type of non-recursive
-local functions can always be inferred from downwards or upwards inference, as
-it will have a body, and the return type of the body is known (even if there
-are inference failures within).
+Declaring a recursive local function, a top-level function, a static method, a
```
lrhn: Is a recursive function always a problem?
Should it only be if the return type of the function depends on itself, and not, say:
```dart
foo(int x) => x <= 0 ? 0 : foo(x - 1).toString().length;
```
The return type of this function is no harder to deduce than a non-recursive function.

resources/type-system/strict-inference.md
```diff
@@ -297,12 +297,14 @@ void main() {
...
+non-overriding instance method, a typedef, a generic function type, or a
```
lrhn: "a typedef" is traditionally called a "a type alias".

The "a generic function type" case is different from the rest. Every other thing you mention is the syntactic category of a declaration, which makes sense since the sentence starts with "Declaring a ...". A "generic function type" is not a declaration, so it's not as clear what it means to "declare" one.

Why is it important that the function type is generic?

srawlins: Hmm, I may just be abusing the analyzer's AST for this concept, [GenericFunctionType](https://pub.dev/documentation/analyzer/0.37.0/dart_ast_ast/GenericFunctionType-class.html). It's not important that it be declared with type arguments. E.g.

```dart
void f6(callback()) {                 // Inference failure
  callback();
}
```

The issue there is that the return type of `callback` is implicitly `dynamic`; no inference is carried out. Is this just called a "function type?"

resources/type-system/strict-inference.md
srawlins: I thought type inference only calculated an omitted return type on a local function if that function was "non-recursive", as in, never references itself.

I don't see anything in the spec regarding this concept though...

munificent:
This PR is, uh, quite old. Is it worth landing or just abandoning?

PR #596, opened 2019-09-26
Move const-update-2018 to accepted/2.5

srawlins:


mit-mit:
Thanks, I had forgot about this one!

PR #599, opened 2019-09-27
Allow 'unused' parameters with underscores for names

srawlins:


resources/type-system/strict-inference.md
```diff
@@ -209,9 +211,10 @@ typedef Callback = void Function(int); // OK
 
 void main() {
   var f = (var a) {};      // Inference failure
+  var g = ([_, __]) {};    // OK
```
lrhn: Why are these parameters optional? If it's not important, then make them non-optional (it's just noise then, and makes people like me start questioning the reason behind it).

resources/type-system/strict-inference.md
```diff
@@ -171,7 +171,9 @@ type cannot be inferred from downwards inference (in the case of a function
 literal) is considered an inference failure. A function literal's parameter
 types are commonly inferred when assigning the literal to a variable typed with
 a typedef, or when passing the literal as an argument whose corresponding
-parameter is  function-typed.
+parameter is function-typed.
+
+"Unused" parameters named `_`, `__`, etc. never represent an inference failure.
```
lrhn: What does quote-*unused*-unquote mean?
Must the parameters actually be unused or does it apply to any parameters named `_`*? In the former case, remove the quotes, and if it's the latter then remove the `"Unused"` entirely, and perhaps add a rationale text like:
> *Such names traditionally represent variables which are ignored, and as such, their type, `dynamic` or not, is not important.*

srawlins: Good clarification; it's the latter. Done.

resources/type-system/strict-inference.md
srawlins: Just noise, I suppose :). In practice I saw it was often optional parameters defined in type aliases. Fixed.

resources/type-system/strict-inference.md
```diff
@@ -171,7 +171,11 @@ type cannot be inferred from downwards inference (in the case of a function
 literal) is considered an inference failure. A function literal's parameter
 types are commonly inferred when assigning the literal to a variable typed with
 a typedef, or when passing the literal as an argument whose corresponding
-parameter is  function-typed.
+parameter is function-typed.
+
+Parameters named `_`, `__`, etc. do not represent an inference failure. Such
+names traditionally represent variables which are ignored, and as such, their
+type, `dynamic` or not, is not important.
```
bwilkerson: Do we want to specify that they are only allowed (not an inference failure) if they are, in fact, not referenced within the function? Not doing so leaves a hole, but it's probably uncommon enough that it might not be worth specifying. On the other hand, the analysis isn't that hard, so it might be worth it.

srawlins: Goooood question. I'd say that case of using a variable named `_` is quite uncommon (in fact, we could add a lint for it 😄). I'm ok not specifying anything about the use of `_`.

srawlins:
@lrhn I'd love another review, thanks!

resources/type-system/strict-inference.md
lrhn: How does a variable "represent" an inference failure? The word "represent" is not used before in this document.

Maybe:

> Parameters named \`_\`, \`__\`, etc., are not considered inference failures if they have no type, 
> as long as they are not referenced inside the body of the function.

Or mix it into the previous paragraph:

> A function parameter declared ....
> ... is considered an inference failure, except if the variable is named \`\_\`, \`\_\_`, etc., 
> and the variable is not referenced from inside the function body.

(Or just ignore it for any variable which is not used).

srawlins:
> I think the document needs to state, somewhere, that an "inference failure" is considered a compile-time error and prevents execution. Because it doesn't say that.

Interesting. I don't think an inference failure should be a compile-time error, nor prevent execution. Right now the first paragraph of the document includes:

```
As a static analysis option, we only intend to implement this
feature in the Dart Analyzer. Under this feature, when [...], inference is considered to have
failed, and an analyzer "Hint" is reported at the location of the expression.
```

I think we only concept this mode to exist under static analysis, so that tools which "execute" code are not even necessarily aware of the analysis. It could be enabled under the new `--enable-experiment` flags which the various back ends support.

> (It doesn't define "inference failure" at all, just mentions it on line 68. If that's the introduction point, it should at least be italicized: "...considered to have an inference failure."

Good point. That first paragraph also includes "inference is considered to have failed." It was my intention to define "inference failure" there, but I don't include those exact words. I'll include them now.

resources/type-system/strict-inference.md
srawlins: I like your last suggestion. I've written text to follow this idea.

lrhn:
Makes sense. In strict mode, and inference failure is an "analysis error". It's not a real compile-time error because it's not actually preventing execution, but obviously you can treat it as an error in your build process.

PR #601, opened 2019-09-28
Starting to gather and formalize my thoughts around pattern matching.

munificent:
Like I did with ui-as-code, I thought it would be good to start with a
high level picture of what we're hoping to achieve. This doc is that.

### DIFF
--- a/working/0546-patterns/goals-and-constraints.md
+++ b/working/0546-patterns/goals-and-constraints.md
@@ -1,0 +1,462 @@
+# Goals and Constraints
+
+We're investigating adding pattern matching to Dart. If you aren't familiar
+with the concept, [this Stackoverflow page][so] might help. [This tracking
+issue][546] also helps frame things.
+
+[546]: https://github.com/dart-lang/language/issues/546
+
+Before I get into details of the design, I wanted to walk through our specific
+goals and constraints. It's not enough to just yank pattern matching out of,
+say, Haskell, and cram it into Dart. We need to define what a *good* pattern
+matching feature looks like *in the context of Dart.*
+
+[so]: https://stackoverflow.com/questions/2502354/what-is-pattern-matching-in-functional-languages
+
+*Note: In this doc, I'll use a made-up match statement and other hypothetical
+forms that could contain patterns as a way to show examples. The important part
+of the examples is the patterns themselves contained in those forms. This doc
+does not concretely propose any of those specific features.*
+
+## General Goals
+
+Before I get to specific pattern matching features, here's some higher level
+goals that I have in mind when I think about designing this feature. Most are
+not hard requirements, but soft constraints we will make trade-offs between and
+aim to maximize.
+
+### Be unambiguous and syntactically simple
+
+Of course, the grammar *must* be unambiguous. If the user writes a valid Dart
+program, it needs to mean only one thing. Further, we'd like the rules that
+determine what a given pattern means to be as simple as possible.
+
+Ideally, they would be completely context free. If not, they should at least
+avoid relying on static type analysis in order to be meaningfully parsed. It may
+be acceptable to rely on variable resolution to determine what a piece of syntax
+means, much in the way that we rely on resolution to know if `foo` in `foo.bar`
+is an import prefix or a variable.
+
+### Offer good static checking
+
+Dart is generally moving in a direction that if we can detect a user error at
+compile time, we should. The earlier a developer can find and fix a mistake, the
+better. That implies good static checking, in a few different ways:
+
+If we can determine at compile time that a given pattern will *always* fail to
+match a value, we should probably report that as an error. For example:
+
+```dart
+String s = ...
+match (s) {
+  case 1 => print("???");
+}
+```
+
+The compiler knows `s` is a String and the `1` pattern can never match any
+possible strings, so this code isn't useful.
+
+Patterns are also often used in places like variable declarations where users
+don't expect a runtime failure to occur. In that case, if a pattern *might* fail
+to match, it should probably be compile-time error:
+
+```dart
+num n = ...
+let int i = n;
+```
+
+Here, `let` is some made-up syntax that takes a pattern (`int i`) and matches it
+against the initializer. In this case, a num might not match a pattern that
+requires the value to be an int (it could be a double). Instead of failing at
+runtime with an exception, this should probably be a static error.
+
+### Check for exhaustiveness
+
+One aspect of static checking is *exhaustiveness* checking. At compile time,
+many languages can tell you if the set of patterns you are matching a value
+against fully covers all possible values that might occur. If not, it tells you
+that your patterns are not exhaustive -- some errant value may sneak through and
+match none of the cases.
+
+This is similar to how a switch statement on an enum type today gives you a
+warning if you don't cover all of the cases. Extending this to arbitrary
+patterns is a useful, powerful feature, because it means that when users add new
+cases to some enum or enum-like type, the compiler will tell them all of the
+places in code that may need to be extended to handle that case.
+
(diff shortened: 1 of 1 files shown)


working/0546-patterns/goals-and-constraints.md
```diff
@@ -0,0 +1,462 @@
...
+
+We'll presumably want to support dotted identifiers for enum cases. I imagine
+users will expect that to also work for named constants imported from libraries
+with a prefix.
+
+That raises the question of whether we want to support arbitrary constant
+expressions. That has some appeal, but means that much of the expression grammar
+would overlap the pattern grammar, which leads to ambiguity and pain. Consider,
+for example, `[1, 2]`. Is that a constant pattern containing a constant list
+literal, or a list descructuring pattern containing two literal patterns?
```
mnordine:
```suggestion
literal, or a list destructuring pattern containing two literal patterns?
```

working/0546-patterns/goals-and-constraints.md
```diff
@@ -0,0 +1,462 @@
...
+As in other languages, the syntax for these should mirror the corresponding
+collection literal expression form:
+
+```dart
+Object obj = ...
+match (obj) {
+  case (a, b) => print("tuple with elements $a and $b");
+  case [a, b] => print("list with elements $a and $b");
+  case {a, b} => print("set with elements $a and $b");
+  case (a, b) => print("tuple with elements $a and $b");
```
mnordine: Should the tuple case be here twice?

mnordine:
@munificent What about destructuring from instance fields, like so:

```dart
class Point
{
  num x,y,z;
  
  Point(this.x, this.y, this.z);
}

void main()
{
  final pt = Point(1,2,3);
  final {x,y,z} = pt;
  print('x: $x, y: $y, z: $z');
}
```

If so, maybe a tiny blurb is worth mentioning?

working/0546-patterns/goals-and-constraints.md
munificent: Oops, fixed, thanks!

working/0546-patterns/goals-and-constraints.md
munificent: Nope, fixed.

munificent:
> What about destructuring from instance fields, like so:

That's what the "Destructure instances of classes" section is about.

mnordine:
I see you have it there in a `match` context, but what about without it, as in destructing to local variables, as above? Or is that type of destructing out of scope?

working/0546-patterns/goals-and-constraints.md
```diff
@@ -0,0 +1,461 @@
...
+  Rect(this.left, this.top, this.right, this.bottom);
+}
+class Circle extends Shape {
+  final num x, y, radius;
+  Circle(this.x, this.y, this.radius);
+}
+
+printShape(Shape shape) {
+  match (shape) {
+    case Rect(left, top, right, bottom) =>
```
lrhn: What C# does here is: `Rect{Left: var left, Top: var top, Right: var right, Bottom: var bottom}`.
This separates the property name from the variable name, which is necessary when you can also pattern match the individual property values:
```c#
if (shape is Rect(Left: 0, Bottom: 0, Right: var width, Top: var height) rect) {
  ...
}
```
It's necessary to distinguish `Left: x` from `Left: var x` because the former uses the existing variable `x` as test value and the latter introduces a new `x` variable with a non-refutable pattern ([Example](https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLQKgLwPYIMpwAOhuMATADQAmIA1AD4ACZAjALABQnjAzAATMBLAOycA3pz5SBZMpOkSO05XwCSeABZRCEPgj4BePgDsIAdz4AlCAGMYACgAMFPs74sXZAJQBueSr4ASzA+e31A5CtbGDEAGwgwGBBXF2BsGBhsAFtktwRAgHMNJL4ANyh9M0CqGA0XTMJk8v0NCELigF89aK8+RQCAxhYATnsAIgARaqCSsb5abrsAOnyimHm+MYAPOYXW9phffxUO49OOc8DjGAgEMCgbXXUtHT7jq/WAQSQoPoKIGA+c7nQTWOx8ZLPbS6frSXhBa58eKJPxKOH8D56A6o5TwzENHHohHrNIZbKEqTwsEOTHImAuTGkzJZBmI1bFVnrBq9WEBWoRJZ0wxIhKA47KfnIFYHYXssVovkaAVM7LClVZCkqSVLBrCglnY54xHfaBvBUqf7rXkDATCUJyvioEWJXoAKlCuqd6qO5uU5z9nHOQA==)).

munificent:
> Or is that type of destructing out of scope?

Out of scope of this *doc*, yes. :) Here I'm just trying to lay out the constraints and goals for the patterns themselves. Defining the other language constructors that will be allowed to *contain* patterns is hinted at here but is left for later in another doc. It's a big set of features, so I'm trying to break it into digestible pieces.

I do aim to be able to have destructuring in local variable declarations.

working/0546-patterns/goals-and-constraints.md
```diff
@@ -0,0 +1,461 @@
...
+        print("Rect $left, $top, $right, $bottom");
+    case Circle(x, y, radius) =>
+        print("Circle $x, $y, $radius");
+  }
+}
+```
+
+Here, the `Rect(...)` and `Circle(...)` are matching on shapes of the
```
leafpetersen: There are interesting design choices here around what this is testing.  Exact instance of Rect? Subclass of Rect? Instance of something implementing Rect?  Instance of something conforming to the "Rect destructor protocol"?  etc.

working/0546-patterns/goals-and-constraints.md
```diff
@@ -0,0 +1,461 @@
...
+
+*   **A pattern-based if.** This is basically a lightweight, single case pattern
+    match statement. You have one pattern and a "then branch" you execute if the
+    pattern matches and an optional else branch if it doesn't.
+
+    For example, [C# allows a type pattern][csharp] inside the condition of an
+    if statement and Swift has [`if case`][swift].
+
+    [csharp]: https://docs.microsoft.com/en-us/dotnet/csharp/pattern-matching
+    [swift]: http://fuckingifcaseletsyntax.com/
```
kevmoo: 😮 Had to parse that URL 3 times...

munificent: Yeah, a little unfortunate that this was the best reference I could find:

*Achievement unlocked: Used a profane URL in a language proposal.*

working/0546-patterns/goals-and-constraints.md
```diff
@@ -0,0 +1,461 @@
...
+num n = ...
+let int i = n;
+```
+
+Here, `let` is some made-up syntax that takes a pattern (`int i`) and matches it
+against the initializer. In this case, a num might not match a pattern that
+requires the value to be an int (it could be a double). Instead of failing at
+runtime with an exception, this should probably be a static error.
+
+### Check for exhaustiveness
```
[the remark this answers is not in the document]
munificent: > whether there should be a way to write refutable patterns (possibly with explicit syntax)

By this, do you mean *at all* or just in a context like a variable declaration where it's not clear whether refutation is desired?

> We've discussed essentially giving people "implicit downcasts" back via a refutable pattern binding. 

Yeah, I think we'll want some kind of "matched downcast" as a way to emulate sum type-style matching using subclasses.

This is good feedback, thanks!

working/0546-patterns/goals-and-constraints.md
munificent: Yes, or even an arbitrary class that happens to implement some destructuring protocol that accepts a Shape as the matched value.

PR #610, opened 2019-10-08
Fix await specification to be type safe.

lrhn:
The current specification of `await` expressions awaits any `Future`. That's not type-safe, and a left-over from when we flattened futures recursively. We no longer do that, so we must also not flatten a future which is intended to be a value.

Example:
```dart
FutureOr<Future<int>> ffs = Future<int>.value(37);
var x = await ffs;
```
This code infers `Future<int>` as the type of `x`. Then evaluation awaits the `Future<int>` value of `ffs` to 37 and then fails to assign it to `x`.
There are no type warnings.

The `await` must only await the future if it is the future type which was flattened away, not if it is the inner value future, so it *must* check the type of the future (unless the static type tells us that it can only be a future of that type, not a value).

Super-types of `Future` have the usual problems which may go away with NNBD's no-implicit-downcast. You can await `dynamic`, with return type dynamic and flattening of any `Future` value. You can await `Object`, but only if the context type allows `Object`.

This is potentially a breaking change, but one which improves type safety. I'm not sure I can come up with a situation where you can unwrap the wrong future and still satisfy the type constraints without downcasts, so I think we can include this in NNBD without further consequences.


eernstg:
Just noting that there is no run-time error for this:

```dart
import 'dart:async';

FutureOr<Future<int>> ffs = Future<int>.value(37);

main() async {
  var x = await ffs;
  print(x);
}
```

Interestingly, this turns out to be because there is no dynamic type check when `x` is initialized (at least in `dart2js`), so it's actually a real soundness violation rather than a dynamic type error.

specification/dartLangSpec.tex
```diff
@@ -28,6 +28,11 @@
 % Significant changes to the specification.
 %
 % 2.4
+% - Change specification of `await` to be type-safe.
+%   The existing specification allowed predictable run-time type errors for
+%   well-typed programs like
+%       FutureOr<Future<S>> ffs = Future<S>.value(s);
+%       Future<S> fs = await ffs;
```
eernstg: This should be added at the end of the 2.4 list, not the beginning.

Also, 5 lines of explanation in the change log for an update is long. How about:

% - Change specification of `await` to be type-safe, preventing unsound constructs like
%       FutureOr<Future<int>> ffs = Future<int>.value(s);
%       Future<int> fs = await ffs;

specification/dartLangSpec.tex
```diff
@@ -11526,27 +11537,30 @@ \subsection{Await Expressions}
...
-a subtype of \code{Future}.
+a supertype or subtype of \code{Future}.
 Tools may choose to give a hint in such cases.%
 }
 
 \LMHash{}%
-The static type of $a$ is \flatten{T} (\flatten{} is defined in section \ref{functionExpressions}) where $T$ is the static type of $e$.
+The static type of $a$ is $T$.
+If $a$ has a static context type $R$, then the static context type of
+$e$ is \code{FutureOr<$R$>}.
```
eernstg: I'd prefer to have PR #246 landed, introducing context types in a systematic manner. It seems messy to introduce it one place at a time (though it may be faster ;-).

specification/dartLangSpec.tex
lrhn: DOne.

PR #611, opened 2019-10-08
Small updates to NNBD spec

leafpetersen:
This incorporates several clarifications that have come up in discussions.

cc @MichaelRFairhurst @crelier @bwilkerson @stereotype441 @rakudrama @nshahan @stefantsov

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -208,6 +208,9 @@ type which is not `dynamic`, and is not a subtype of `Iterable<dynamic>`.
 It is a warning to use a null aware operator (`?.`, `?..`, `??`, `??=`, or
 `...?`) on a non-nullable receiver.
 
+It is a warning to use the null check operator (`!`) on a non-nullable
+expression.
+
 ### Assignability
 
 The definition of assignability is changed as follows.  
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -228,7 +228,7 @@ not a subtype of the class type associated with the class in which it is defined
 constructor for any class other than `Null`).
 
 It is a warning to use a null aware operator (`?.`, `?..`, `??`, `??=`, or
-`...?`) on a non-nullable receiver.
+`...?`) on a non-nullable value.
 
 It is a warning to use the null check operator (`!`) on a non-nullable
 expression.
@@ -544,8 +544,7 @@ would have succeeded if all `?` types were ignored, `Never` were treated as
 In legacy libraries, whether run in strong or weak checking mode, instance
 checks (`e is T`) also return `false` whenever they would have done so pre-nnbd.
 Specifically, if `e` evaluates to `null`, then `e is T` always returns `false`
-unless `T` is `Null`, `Never`, `Object`, `dynamic`, `void`, or a nullable or
-legacy version of any of the above.
+unless `T <: Null` or `Object <: T`.
 
 ### Exports
 
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -105,7 +105,6 @@ The **flatten** function is modified as follows:
   - if `T` is `S?` then **flatten**(`T`) = **flatten**(`S`)`?`
   - otherwise if `T` is `S*` then **flatten**(`T`) = **flatten**(`S`)`*`
   - otherwise if `T` is `FutureOr<S>` then **flatten**(`T`) = `S`
-  - otherwise if `T` is `FutureOr<S>` then **flatten**(`T`) = `S`
   - otherwise if `T <: Future` then let `S` be a type such that `T <: Future<S>`
 and for all `R`, if `T <: Future<R>` then `S <: R`; then **flatten**('T') = `S`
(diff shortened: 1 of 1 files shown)


accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -516,6 +541,12 @@ implicit casts) shall succeed with a warning whenever the runtime type test
 would have succeeded if all `?` types were ignored, `Never` were treated as
 `Null`, and `required` named parameters were treated as optional.
 
+In legacy libraries, whether run in strong or weak checking mode, instance
+checks (`e is T`) also return `false` whenever they would have done so pre-nnbd.
+Specifically, if `e` evaluates to `null`, then `e is T` always returns `false`
+unless `T` is `Null`, `Never`, `Object`, `dynamic`, `void`, or a nullable or
+legacy version of any of the above.
```
eernstg: Would it suffice to say 'unless `T` is a type whose value is `Null`, `Never`, or a legacy top type'?

Otherwise it's not obvious that it includes cases like `e is X` where `X` is a type variable whose value is `Never`, or `e is FutureOr<Object>`, and a bunch of other cases.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -92,6 +97,18 @@ The internal representation of types is extended with a type `T*` for every type
...
+  - if `T` is `S?` then **flatten**(`T`) = **flatten**(`S`)`?`
+  - otherwise if `T` is `S*` then **flatten**(`T`) = **flatten**(`S`)`*`
+  - otherwise if `T` is `FutureOr<S>` then **flatten**(`T`) = `S`
+  - otherwise if `T <: Future` then let `S` be a type such that `T <: Future<S>`
+and for all `R`, if `T <: Future<R>` then `S <: R`; then **flatten**('T') = `S`
+  - otherwise **flatten**('T') = `T`
+
 ### Static errors
 
 We say that a type `T` is **nullable** if `Null <: T` and not `T <: Object`.
```
lrhn: (Currently "not `T <: Object`" is sufficient, right? I guess it's a good idea to have both).

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -205,9 +222,17 @@ is declared `late` and does not have an initializer.
 It is an error if the object being iterated over by a `for-in` loop has a static
 type which is not `dynamic`, and is not a subtype of `Iterable<dynamic>`.
 
+It is an error if the type of the value returned from a factory constructor is
+not a subtype of the class type associated with the class in which it is defined
+(specifically, it is an error to return a nullable type from a factory
+constructor for any class other than `Null`).
+
 It is a warning to use a null aware operator (`?.`, `?..`, `??`, `??=`, or
 `...?`) on a non-nullable receiver.
```
lrhn: "receiver" is not really accurate for `...?`. Could it be just "value"?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -205,9 +222,17 @@ is declared `late` and does not have an initializer.
...
 
+It is a warning to use the null check operator (`!`) on a non-nullable
```
lrhn: Just a warning, not an error?

accepted/future-releases/nnbd/feature-specification.md
lrhn: Is it that simple?

If we do `null is FutureOr<Never>` then that's equivalent to `null is FutureOr<Never*>` in a legacy library, but not to `null is FutureOr<Never*>*`.  Which is true, but it's not a top type.

So: `null is T` returns true if `T` is _Null_, _Never_, a top type, _S?_ for any _S_, or `FutureOr<..>` of any type where it returns true?

Maybe:

> When `e is T` occurs in a legacy library, if `e` evaluates to `null` and the legacy interpretation of the type expression `T` is _S*_, then the test evaluates to `true` if _Null_ <: _S_ or  _S_ is _Never_, and to `false` otherwise.

I wouldn't try to enumerate the infinitely many top types.
I'm actually a little surprised that `null is Never` succeeds. Are we sure we want that?

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: No.  If `T` is `X extends Object?` we have "not `T <: Object`", but we don't want to treat `X` as **nullable** - only as **potentially nullable**.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: It feels unfriendly to code generators that want to be blind about nullability to make it an error.  Also makes it more breaking to change a nullable return type to a non-nullable return type, etc.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: done.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: How about `T <: Null` or `Object <: T`?

> I'm actually a little surprised that `null is Never` succeeds. Are we sure we want that?

We certainly need `null is Never*` for backwards compatibility, and I don't see much reason to explicitly exclude `Never`.

PR #615, opened 2019-10-09
Spec fix cascades

eernstg:
The section on cascades had no specification matching expressions like `e..m1..m2` (anything containing more than one `<cascadeSection>` was either incorrectly specified or completely unspecified, depending on the interpretation of the unspecified term _suffix_).

This PR fixes that issue, and includes conditional cascades (`?..`) as well.

It uses a `let ..` in commentary (only) because we don't otherwise use that in the specification, but is prepared for using it as normative text by having a `TODO` comment to that effect.

### DIFF
--- a/specification/dart.sty
+++ b/specification/dart.sty
@@ -67,6 +67,9 @@
 \def\WITH{\keyword{with}}
 \def\YIELD{\keyword{yield}}
 
+% Used for desugaring.
+\def\LET{\keyword{let}}
+
 % Used for inline code snippets.
 \def\code#1{\texttt{#1}}
 
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -9665,23 +9665,15 @@ Notice that the wording avoids re-evaluating the receiver $o$ and the arguments 
 \EndCase
 
 
-\subsubsection{Cascaded Invocations}
-\LMLabel{cascadedInvocations}
+\subsubsection{Cascades}
+\LMLabel{cascades}
 
 \LMHash{}%
-A \IndexCustom{cascaded method invocation}{method invocation!cascaded}
-has the form \code{$e$..\metavar{suffix}} or \code{$e$?..\metavar{suffix}},
-where $e$ is an expression
-and \metavar{suffix} is derived from \synt{cascadeSection}.
-
-\rationale{%
-A \synt{cascadeSection} basically allows for accessing members
-including setters.
-The motivation for having cascaded invocations is that they can be
-!!!HERE!!!
-}
-
-is a sequence of operator, method, getter or setter invocations.
+A \Index{cascade} is a kind of expression that allows for performing
+multiple operations on a given object
+without storing it in a variable and and accessing it via a name.
+Cascades can have different syntactic forms,
+but will always contain two periods \lit{..}.
 
 \begin{grammar}
 <cascadeSequence> ::= (`?..' | `..') <cascadeSection> (`..' <cascadeSection>)*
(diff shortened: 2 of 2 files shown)


specification/dartLangSpec.tex
```diff
@@ -9665,13 +9665,15 @@ \subsubsection{Ordinary Invocation}
...
+\subsubsection{Cascades}
+\LMLabel{cascades}
 
 \LMHash{}%
-A \IndexCustom{cascaded method invocation}{method invocation!cascaded}
-has the form \code{$e$..\metavar{suffix}}
-where $e$ is an expression and \metavar{suffix} is a sequence of operator, method, getter or setter invocations.
+A \Index{cascade} is a kind of expression that allows for performing
+multiple operations on a given object
+without storing it in a variable and and accessing it via a name.
```
munificent: "and and" -> "and".

eernstg: Thanks, fixed.

eernstg:
The grammar now uses a compositional rule for cascades (such that we get the parentheses from the previous version of this PR for free).

specification/dartLangSpec.tex
```diff
@@ -9685,29 +9691,113 @@ \subsubsection{Cascaded Invocations}
...
+In both cases, the static type of $e$ is the static type of $e_0$.
+
+\LMHash{}%
+Evaluation of a cascaded member access $e$
+of the form \code{$e_0$..$s$}
+where $e$ is an \synt{expression} and
+$s$ is a \synt{cascadeSection}
+such that $e$ is not initially conditional
+proceeds as follows:
+Evaluate $e_0$ to an object $o$.
```
lrhn: Could you just insert here:

If $e_0$ is initially conditional and $o$ is the null object, then $e$ evaluates to the null object.
Otherwise let $t$ .... 

and drop the "is not initially conditional" clause above? Then the entire next section should be subsumed by that.

specification/dartLangSpec.tex
```diff
@@ -9685,29 +9691,113 @@ \subsubsection{Cascaded Invocations}
...
+so any \lit{?} on a non-innermost \lit{..} would be useless.%
+}
+
+\LMHash{}%
+Let $s$ be a \synt{cascadeSection}.
+Let $e$ be an expression of the form \code{$e_0$..$s$}
+where $e_0$ is an \synt{expression}
+such that $e$ is not initially conditional.
+Exactly the same compile-time errors are raised for $e$
+as would be raised for \code{$e_0$.$s$}.
```
lrhn: Make that  ($e_0$).$s$  because otherwise e0 == x..foo would mean that `x..foo..bar` is the same as `x..foo.bar`.

specification/dart.sty
```diff
@@ -67,6 +67,9 @@
 \def\WITH{\keyword{with}}
 \def\YIELD{\keyword{yield}}
 
+% Used for desugaring.
+\def\LET{\keyword{let}}
```
lrhn: Unused?

lrhn:
Woot. Much better (and simpler, which is good, so double better!)

specification/dartLangSpec.tex
```diff
@@ -9665,13 +9665,15 @@ \subsubsection{Ordinary Invocation}
...
 
 \LMHash{}%
-A \IndexCustom{cascaded method invocation}{method invocation!cascaded}
-has the form \code{$e$..\metavar{suffix}}
-where $e$ is an expression and \metavar{suffix} is a sequence of operator, method, getter or setter invocations.
+A \Index{cascade} is a kind of expression that allows for performing
+multiple operations on a given object
+without storing it in a variable and and accessing it via a name.
+Cascades can have different syntactic forms,
+but will always contain two periods \lit{..}.
```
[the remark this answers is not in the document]
lrhn: Sorry, missed one of my outdated comments.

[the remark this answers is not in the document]

specification/dart.sty
eernstg: Yes, removed.

specification/dartLangSpec.tex
eernstg: Did that already. It is also needed for other reasons: `(a + b).x` etc. may be ugly, but should still be handled correctly. ;-)

specification/dartLangSpec.tex
eernstg: Please follow up if needed&mdash;I have not made this change.

The reason for not doing this is that the following section spells out the shorting semantics, and if I make this change then we won't describe that semantics, we will describe the approach where we step through each layer of the cascade one section at a time.

Didn't we want to make the shorting semantics explicit?

specification/dartLangSpec.tex
```diff
@@ -9685,29 +9691,113 @@ \subsubsection{Cascaded Invocations}
...
+where $e_0$ is a \synt{conditionalExpression},
+exactly the same compile-time errors are raised for $e$
+% This rule is NNBD-ready, so we use it even though we could
+% use a slightly simpler set of rules today.
+as would be raised for \code{$e_0$?.$s$}.
+If $e$ is an expression of the form \code{$e_0$..$s$}
+where $e_0$ is an \synt{expression}
+such that $e$ is initially conditional,
+exactly the same compile-time errors are raised for $e$
+as would be raised for \code{$e_0$?.$s$}.
```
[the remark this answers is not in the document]
eernstg: Done already again.

PR #616, opened 2019-10-09
Added SML program implementing the null-shorting transformation

eernstg:
The plan is to tell UniPro that they can use this program to expand expressions, such that they can test null-shorting by writing an expression `e` and testing the resulting value for equality between `e` and the transformed expression `xlate e`, probably in a function where they pass in various different values for subexpressions, such that the same pair of `e` and `xlate e` expressions can be tested for having the same behavior with a null receiver and with a non-null receiver in each relevant location.


mit-mit:
Paging CLA bot

PR #636, opened 2019-10-23
Removed explicit shortcut semantics for cascades

eernstg:


specification/dartLangSpec.tex
```diff
@@ -9807,23 +9796,20 @@ \subsubsection{Cascades}
...
-Then $e$ evaluates to the result of
-evaluating \code{v..$s_1$..$s_2$ $\cdots$ ..$s_k$}.
+If $o$ is not the null object:
+let \code{v} be a fresh variable bound to $o$ and
+evaluate \code{v.$s$} to an object which is discarded,
+then $e$ evaluates to $o$.
+Otherwise $o$ is the null object.
+If $e$ is initially conditional then $e$ evaluates to $o$;
+%% TODO(eernst): With NNBD, this error cannot happen -- delete.
+otherwise a dynamic error occurs.
```
lrhn: How about:

> If $e$ is initially conditional and $o$ is the null object,
then $e$ evaluates to the null object.
Otherwise, let \code{v} be a fresh variable bound to $o$ and
evaluate \code{v.$s$} to an object which is discarded, and
then $e$ evaluates to $o$.

The currently proposed text incorrectly disallows `null..toString()`.

eernstg: Ah, yes, and the dynamic error will happen during `v.s` anyway. Done.

PR #638, opened 2019-10-23
Added FINAL? to grammar of formal parameters

### LINKED ISSUE #502: Grammar anomaly with `var` and parameters
The grammar allows `var` to be used as the return type in a `<fieldFormalParameter>`:

```dart
class C {
  Function x;
  C(var this.x())
  void f(var g()) {}
}
```

However, this is treated by all tools as a compile-time error. It does not seem to have an obvious and useful semantics, so this issue serves to propose that we specify it to be a compile-time error.

(Given that we do not intend to require tools to report a compile-time error as a syntax error or not, we are free to do this by changing the grammar in the specification such that it just doesn't derive `var` in that position; tools are still free to parse it without error, and then report it as a non-syntax compile-time error).

Another anomaly is that a `<fieldFormalParameter>` cannot be declared `final`:

```dart
void f1(
  final void f(), // Syntax error.
  final void Function() g, // OK.
) {}
```

So, for consistency, we should adjust that grammar rule to allow `final`.

A related topic is that a `<simpleFormalParameter>` can use `var` in the place where a type is otherwise located:

```dart
void f2(
  var x, // OK.
  void f(var y), // OK.
) {}
```

This also seems to be of little value. But it is not inconsistent. It could presumably be used, say, to indicate that the type is provided by inference (as a matter of style, or possibly with tool support), and this issue does not propose to make any changes here.

As a whole, this issue is intended to be a grammar anomaly cleanup, and the changes should be non-breaking, relative to the current behavior of our tools.

eernstg:
This PR adds an optional `final` to the syntax of formal parameters, as proposed in language issue #502 and supported by Leaf and Bob. Lasse, if you support this then we can land this PR and resolve #502. It will then need to be implemented, but it is a non-breaking change and could be done when we have the resources.

Otherwise we can just delete those two occurrences of `\FINAL?` from this PR, and land the remaining changes (which are bug fixes that came up as I checked these changes and detected differences between Dart.g and the spec).

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -863,11 +863,11 @@ Variables are storage locations in memory.
 \begin{grammar}
 <variableDeclaration> ::= <declaredIdentifier> (`,' <identifier>)*
 
-<declaredIdentifier> ::= <metadata> \COVARIANT{}? <finalConstVarOrType> <identifier>
+<declaredIdentifier> ::= \COVARIANT? <finalConstVarOrType> <identifier>
 
-<finalConstVarOrType> ::= \LATE{}? \FINAL{} <type>?
+<finalConstVarOrType> ::= \LATE? \FINAL{} <type>?
   \alt \CONST{} <type>?
-  \alt \LATE{}? <varOrType>
+  \alt \LATE? <varOrType>
 
 <varOrType> ::= \VAR{}
   \alt <type>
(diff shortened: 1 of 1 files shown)


leafpetersen:
> . It will then need to be implemented, but it is a non-breaking change and could be done when we have the resources.

Please don't land this until this change is at least in the process of being implemented.  The spec is documentation, and making the documentation inconsistent with the implementation without making any plans to bring the implementation in line is something I'd really like us to avoid.  It's too easy for things like this to fall through the cracks.  As discussed last week, perhaps we should batch up a bunch of these changes and try to land them next quarter?

eernstg:
The support for `final` on formal parameters has been changed completely today (it's not supported in general, only on parameters of primary constructors, in which case it specifies that the parameter is declaring). So this PR is completely obsolete today.

PR #643, opened 2019-10-25
Specify that a constant e.length cannot use an extension method

eernstg:
The specification of constants left this possibility open (because the syntax `e.length` could previously only call the instance getter, but now it can also invoke/tear-off an extension member). The update avoids saying 'extension method' because those are not yet in the language specification.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -34,6 +34,8 @@
 %   interpolation; specify that it is a dynamic error for `loadLibrary`
 %   to load a different library than the one which was used for
 %   compile-time checks.
+% - Specify that a constant expression `e.length` can only invoke an instance
+%   getter (in particular, it cannot execute/tear-off an extension member).
 %
 % 2.4
 % - Clarify the section 'Exports'.
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -6281,6 +6281,10 @@ It is further a constant expression if the map literal evaluates to an object.
   if $e$ is a potentially constant expression.
   It is further constant if $e$ is a constant expression that
   evaluates to an instance of \code{String}.
+  % Prevent extensions methods from running arbitrary code
+  % when the receiver $e$ has static type `Object`.
+  It is a compile-time error if the execution of \code{$e$.length}
+  would not invoke an instance getter.
 
 % New in 2.1.
 \item An expression of the form \code{$e$\,\,as\,\,$T$} is potentially constant
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -6285,8 +6285,8 @@ It is further a constant expression if the map literal evaluates to an object.
   evaluates to an instance of \code{String}.
   % Prevent extension methods from running arbitrary code
   % when the receiver $e$ has static type `Object`.
-  It is a compile-time error if the execution of \code{$e$.length}
-  would not invoke an instance getter.
+  It is a compile-time error if an expression of the form \code{$e$.length}
+  must be constant, but would not invoke an instance getter.
 
 % New in 2.1.
 \item An expression of the form \code{$e$\,\,as\,\,$T$} is potentially constant
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -35,7 +35,8 @@
 %   to load a different library than the one which was used for
 %   compile-time checks.
 % - Specify that a constant expression `e.length` can only invoke an instance
-%   getter (in particular, it cannot execute/tear-off an extension member).
+%   getter (in particular, it cannot execute/tear-off an extension member);
+%   similar changes were made for several operators.
 %
 % 2.4
 % - Clarify the section 'Exports'.
@@ -6112,9 +6113,13 @@ and \IndexCustom{constant expressions}{constant expression}
 are the following:
 
 \begin{itemize}
-\item A literal boolean, \TRUE{} or \FALSE{} (\ref{booleans}), is a potentially constant and constant expression.
+\item
+  A literal boolean, \TRUE{} or \FALSE{} (\ref{booleans}),
+  is a potentially constant and constant expression.
 
-\item A literal number (\ref{numbers}) is a potentially constant and constant expression
+\item
+  A literal number (\ref{numbers}) is
+  a potentially constant and constant expression
   if it evaluates to an instance of type \code{int} or \code{double}.
   % A too-large integer literal does not evaluate to an object.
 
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -6281,6 +6283,10 @@ \subsection{Constants}
   if $e$ is a potentially constant expression.
   It is further constant if $e$ is a constant expression that
   evaluates to an instance of \code{String}.
+  % Prevent extension methods from running arbitrary code
+  % when the receiver $e$ has static type `Object`.
+  It is a compile-time error if an expression of the form \code{$e$.length}
+  must be constant, but would not invoke an instance getter.
```
leafpetersen: Do we need to say "must be constant or potentially constant"?

eernstg: My first take on that is that we don't: Being potentially constant is otherwise a simple property (we use syntax, so `true` and `42` are p.cst.; and name-application-to-declaration-bindings, so `id` is p.cst. when `id` is a library function or static method, or a formal parameter of a const constructor; and we have a bunch of congruence rules, so `e1 + e2` is potentially constant when both `e1` and `e2` are potentially constant, etc).

For instance, `e.length` is already potentially constant if `e` is potentially constant, and we don't check whether the static type is `String` or `dynamic` (so it is not a problem if the static type of `e` is `bool`, even though that will obviously prevent the expression from being constant).

I believe the notion of being potentially constant is _only_ used in order to improve the level of confidence a developer can have in a given constant constructor initializer list: When we replace the formal parameter by the actual arguament in a `<constObjectExpression>` like `const C(e)`, we already [require](https://github.com/dart-lang/language/blob/93c7a9c390644006911cc0bff977dbd50f017bce/specification/dartLangSpec.tex#L3557) the resulting expressions in the initializer list to be constant. So we never trust a potentially constant expression to provide any guarantees, we just use the notion of potentially constant expressions to raise a flag in some cases where this substitution requirement is obviously going to fail.

So I believe that it wouldn't be necessary to impose the constraint on potentially constant expressions, and also that it would differ from the level of guarantees that we otherwise get from that concept, so it's probably not worthwhile to try to be strict here.

eernstg:
Updated to require that we invoke an instance member in several other cases (operators).

specification/dartLangSpec.tex
```diff
@@ -34,6 +34,9 @@
 %   interpolation; specify that it is a dynamic error for `loadLibrary`
 %   to load a different library than the one which was used for
 %   compile-time checks.
+% - Specify that a constant expression `e.length` can only invoke an instance
```
lrhn: Update comment too?

specification/dartLangSpec.tex
```diff
@@ -6124,201 +6131,290 @@ \subsection{Constants}
...
+  such that \lit{+} denotes an instance operator invocation.
 
-\item An expression of the form \code{-$e_1$} is a potentially constant expression
+\item
+  An expression of the form \code{-$e_1$} is a potentially constant expression
   if $e_1$ is a potentially constant expression.
   It is further a constant expression if $e_1$ is a constant expression that
-  evaluates to an instance of type \code{int} or \code{double}.
+  evaluates to an instance of type \code{int} or \code{double},
+  such that \lit{-} invokes the instance operator \code{unary-}.
```
lrhn: You say "*the* instance operator" here, but "*an* instance operator" everywhere else.

Just say that `\lit{-} denotes an instance operator invocation.` like everywhere else. The other places don't say *which* one they denote.

eernstg:
Double-checked treatment of `!`, resolved an issue with its specification (and similarly for `++e`, `--e`): Added missing parentheses, adjusted such that static semantics is included (we used to say only how those expressions are evaluated).

specification/dartLangSpec.tex

specification/dartLangSpec.tex

specification/dartLangSpec.tex
leafpetersen: > For instance, `e.length` is already potentially constant if `e` is potentially constant, and we don't check whether the static type is `String` or `dynamic` or anything else specifically (so it is not a problem if the static type of `e` is `bool`, even though that will obviously prevent the expression from being constant)

This is not true.  This class is a static error:

```dart
class A {
  final int l;
  const A(String x) : l = x.length;
  const A.error(bool x) : l = x.length;
}

void main() {
}
```

We require constant constructors to be well-typed.

So I don't understand what you are saying here.  Are you saying that we should accept this program so long as we never try to create a const instance of `A`?
```dart
extension on bool {
  int get length => 3;
}
class A {
  final int l;
  const A(bool x) : l = x.length;
}
```
That seems unhelpful to me. 

cc @lrhn

eernstg: We don't check types in order to determine whether a given expression is _potentially constant_, but we do of course check types of expressions everywhere, also constant ones.

So the class `A` that you mention has a compile-time error based on normal expression typing. But the following class `B` does not:

```dart
extension ExtObject on Object {
  String get length => 'Very long!';
}

class B {
  final l;
  const B(Object o) : l = o.length;
}
```

We don't want to allow `o.length` to run arbitrary code, so it should never invoke an extension method.

But if we don't have this PR then there is no rule that prevents us from declaring a constant constructor like the one in `B`, and even for `const B('Hest')` it would (presumably) have to run the extension method as part of the constant evaluation step, because `o.length` is statically determined to be an extension method invocation. So it doesn't help that `o` is actually a `String` when we evaluate `const B('Hest')`, because `o.length` is still an extension method invocation. (And it would be highly confusing if we were to introduce an exception to that rule, just for constant expression evaluation.)

So that's the reason for having this PR: It should prevent extension method invocation from occurring during constant expression evaluation.

What I'm saying about 'potentially constant' here is just that this is a rather simple concept (e.g., `e1 + e2` is potentially constant if `e1` and `e2` are potentially constant, no further questions asked), so any attempt to make 'potentially constant' much smarter just for this case seemed inconsistent.

However, it could actually be a useful idea to do that anyway, because it _can_ be detected during checks on the constant constructor itself that any given expression is an extension method invocation (which is a better service for developers than it would be to detect it when a `<constObjectExpression>` is encountered).

But in that case we will definitely need to mention extension methods. We could handle this by adding a `TODO` saying that we introduce the general rule for constant expressions that it is an error if an extension method is invoked. (Just so that we don't say 'extension method' before that concept has been added to the spec.)

leafpetersen: I'm a little lost in the above.  Let me try to lay out my question here as concretely as possible:

This program (program 1):
```
extension ExtObject on Object {
  int get length => 4;
}

class B {
  final l;
  const B.fromInt(int x) : l = x;
  const B(Object o) : l = const B.fromInt(3.length);
}
```
is currently accepted by the CFE but rejected by the analyzer.

This program (program 2):
```
extension ExtObject on Object {
  int get length => 4;
}

class B {
  final l;
  const B(Object o) : l = o.length;
}
```

is currently accepted by all platforms.

I believe that this PR makes program 1 an error (since it says that e.length is an error if e must be constant and length is not an instance getter).

I believe that this PR leaves program 2 as not an error (since it says nothing about e.length if e must only be potentially constant).

My question is simply: do we believe that this is the behavior that we want?  I'm a little surprised that we would issue an error in one and not the other, since both of them must be rejected if we ever try to evaluate them as constants.

If this is the behavior we want, we should specify also what the behavior is.  Currently the two implementations disagree: the analyzer ignores the extension and calls the instance getter, the CFE sees the extension and rejects during constant evaluation (probably we want the latter?):

```dart
extension ExtObject on Object {
  int get length => 4;
}

class B {
  final l;
  const B(Object o) : l = o.length;
  const B.fromInt(int x) : l = x;
}

class C {
  const C(dynamic b0, dynamic b1) : assert(identical(b0, b1));
}

void main() {
  // passes on analyzer because the constructor to B calls the string getter on o
  // fails on CFE because calling an extension method is not allowed
  const C(B("helloworld"), B.fromInt(10));
  // fails on analyzer because the assertion does not pass
  // fails on CFE because the .length extension is not permitted
    const C(B("helloworld"), B.fromInt(4)); // assertion failure
}
```

eernstg: > this PR makes program 1 an error

Almost. Program 1 is an error today because `const B.fromInt(3.length)` is a `<constObjectExpression>`, [so](https://github.com/dart-lang/language/blob/62c036cc41b10fb543102d2f73ee132d1e2b2a0e/specification/dartLangSpec.tex#L9766) `3.length` must be constant, [so](https://github.com/dart-lang/language/blob/62c036cc41b10fb543102d2f73ee132d1e2b2a0e/specification/dartLangSpec.tex#L6423) `3` must evaluate to an instance of `String`, which it doesn't. I believe that the response from the analyzer is fine (`const_with_non_constant_argument`).

If we had made it `const B.fromInt('a string'.length)` then I believe that we specify no error today (by accident, because these rules were written before extension methods were added, so there was no loophole). The analyzer does accept the program with that adjustment.

This PR would indeed ensure that the program with `'a string'.length` is an error. I think that's a desirable outcome for this case.

For program 2 I agree that this PR would not make it an error. When I said

> However, it could actually be a useful idea to do that anyway

I was responding to your input: I had been convinced that it _would_ be useful to make the checks for 'potentially constant' stronger, in order to catch this particular issue (which is easy, because the choice between instance method invocations and extension method invocations is made statically). So even though 'potentially constant' is a rather superficial check, (cf.

```dart
class A {
  final a;
  final b;
  const A(x): a = !x, b = x.length;
}

main() {}
```

which will never work and is still accepted), we can improve the 'potentially constant' check in this particular case, and I think we should do that.

> we should specify also what the behavior is

I'd take it for granted that we do not want constant evaluation semantics to include execution of arbitrary user-written code. In that case we should make it an error for a potentially constant (hence also for a constant) expression to invoke an extension method, and I certainly don't think we should give those expressions a non-standard semantics (so if `e.length` would invoke an extension method when executed at run time then it shall also invoke an extension method when encountered during constant evaluation, and we won't do that, so it's just an error and we're done).

eernstg:
I created https://github.com/dart-lang/language/pull/650 in order to make it explicit that extension method invocations are not constant expressions, and not even potentially constant.

I used the phrase 'extension methods' even though that is not yet defined in the language specification, but I expect that it will be added soon. We could add commentary to say that.

specification/dartLangSpec.tex
leafpetersen: Filed https://github.com/dart-lang/sdk/issues/39170 to track the analyzer issue.  @lrhn or @eernstg could one of you add a language test for extension methods in constants?  We probably also need to be sure and specify that this is an error.

eernstg: Dropped #650 again, because of the implementation effort. This means that it will not be detected for a potentially constant expression that it will access an extension member, so we only get an error for expressions that are actually constant (and in particular, we can have a constant constructor which gives no errors, yet can't ever be used in a constant expression; but we have that in so many other ways already, so there's nothing inconsistent about it ;-).

eernstg:
> add a language test for extension methods in constants?

Try https://dart-review.googlesource.com/c/sdk/+/123580.

PR #650, opened 2019-10-29
Prevent extension method invocation in constants

eernstg:


eernstg:
Turns out (says @askeksa-google ;-) that it is important for the implementation of 'potentially constant' checks that there is no need to know any types. So we may not wish to enhance those checks after all, because it would require a substantial implementation effort.

askeksa-google:
Indeed, in the CFE, potentially constant is checked during parsing, where the const context (normal, required or inferred) is tracked, and types are generally not available (since they may depend on later type inference). The const context of an expression is not otherwise kept around, so at the time where the type is known, the const context of the expression is not known. Implementing the suggested change would require the CFE to track const context during type inference.

leafpetersen:
Ok, I'm fine with leaving this as is.

eernstg:
Very good, I'll drop this PR.

PR #653, opened 2019-10-30
Fix markdown. Escape column end character

sgrekhov:
Not escaped `|` is treated as end of the column and it looks like

| Member invocation on target X | Corresponding member name                                    |
| :---------------------------- | :----------------------------------------------------------- |
| `X binop expr2`               | `+`, `-`, `*`, `/` , `~/`, `%`, `<`, `<=`, `>`, `>=`, `<<`, `>>`, `>>>`, `^`, `|`, `&` |


sgrekhov:
Btw, why there is no '==' operator in the list?

PR #664, opened 2019-11-05
Remove Language Versioning text from Extension Methods spec

srawlins:
I suppose I should add a revision note to the bottom? 1.5?


lrhn:
I already have a different v1.5 text, and have removed the text there. See #672

Issue #678, opened 2019-11-11
Partial classes - feature proposal


kevmoo:
A proposal to address request https://github.com/dart-lang/language/issues/252

Pending PR: https://github.com/dart-lang/language/pull/680

kevmoo:
Inspiration: https://github.com/jaredpar/csharplang/blob/partial/proposals/extending-partial-methods.md

Levi-Lesches:
Love this feature. Metaprogramming and code-gen is being discussed in #1482 and I'm working on the second version of my proposal (initially #1507) and the idea of _where_ does the code go and _how_ can it be interacted with from other libraries is a big one. My driving principle is that _generated code in a `.g.dart` file should behave the same as though it were hand-written in regular `.dart` file_.

PS: I'm in favor of `partial class X` instead of `class X partial`. It fits better with `mixin`, `abstract`, `extension`, and `var` which describe the declaration itself instead of `implements`, `extends`, and `on`, which only slightly modify the declaration (and thus go after).

insinfo:
Love this feature

Levi-Lesches:
I opened #1753 to discuss how modules might revive the discussion around partial classes by answering some fundamental question, and how code-generation boosts the need for this feature.

kevmoo:
@munificent – I'm MORE than happy to close this out and defer to https://github.com/dart-lang/language/blob/master/working/augmentation-libraries/feature-specification.md – just say the word! 😄

Levi-Lesches:
@JohnGalt1717 not sure about your use-case specifically, but you might want to look into extensions. Also, see above. 

 

> I'm MORE than happy to close this out and defer to https://github.com/dart-lang/language/blob/master/working/augmentation-libraries/feature-specification.md – just say the word! 😄

@munificent ^

kevmoo:
@munificent – I think we can merge this into the `augmentation` bits?

Issue #252, opened 2019-03-03
Partial Classes and Methods


listepo:
Would like to have something similar in a dart. Especially when generating code.

vanesyan:
Possibly, covered by #177

eernstg:
Right, scoped class extension (#177) is intended to allow third-party developers (that is, for a class `C`, developers who won't edit the code of `C` because they can't or because it's somebody else's code) to add proper instance methods to an existing type hierarchy (a typical example would be a tree of classes that `extend` each other), with the ability to inherit and/or override method implementations, and the ability at call sites to have all the usual typing properties (e.g., that a method may accept more optional arguments than the methods that it overrides).

For code generation you might be able to decide up front that you can _always_ edit the code of the target class(es), in which case you might be able to use a mechanism based on `part` files.

Such a mechanism would allow you to patch a class `C` in the main file by adding method implementations and (possibly private) declarations from the part file in a "partial" declaration of `C`, and it could basically be considered to be similar to an include file mechanism (the effective source code in the main file is obtained by adding stuff from the part files).

This is simpler than scoped class extension, but it will only work in cases where you have enough control over the target libraries to, at least, be able to introduce those `part` declarations into the target libraries and generate the part files.

Also, you can control scoped class extension because it is scoped. So if two third-party organizations wish to add a method named `foo` to the same class(es) then a part based approach will just incur some "duplicate declaration" errors at compile-time. But with scoped class extensions you'd make the choice in each library that contains a call site that you import just one of those conflicting extensions, which eliminates the conflict.

a14n:
A common pattern to generate code (with source_gen):

```dart
// file.dart
part 'file.gen.dart';
class A extends $A {
}

// file.gen.dart
part of 'file.dart';
class $A {
  myAwesoneGeneration() => null;
}
```

With this pattern it's not possible/simple to generate constructors or static member.

eernstg:
Right, and it's also not possible to add a method to `num`, `int`, and `double`.

lrhn:
C# partial classes do not allow you to add members to existing non-partial classes.

You can only have multiple parts of a class declared to be partial, and still only in the same assembly (compilation unit). In Dart, I'd say it would have to be the same *library*.

rrousselGit:
With the weird behavior of extensions methods tear-off that are never equal to themselves https://github.com/dart-lang/language/issues/425, I think partial classes is important.

Without partial classes, code-generators will instead generate extension members.
[...]

PR #680, opened 2019-11-12
Starting on partial class proposal

kevmoo:
Author: kevmoo@google.com

Proposed solution to [partial classes (#252)](https://github.com/dart-lang/language/issues/252).
Discussion about this proposal should go in [Issue #678](https://github.com/dart-lang/language/issues/678).


kevmoo:
Motivated by https://twitter.com/QQqtNni8YTHTj9l/status/1192470981018898434 to give this a "partial" stab 😀

working/0678 - partial classes/feature-brainstorm.md
```diff
@@ -0,0 +1,199 @@
...
+...and would be treated as
+
+```dart
+class Person{}
+```
+
+> Why not restrict it to one-definition-per-file? It's common for code
+generators to create one file with the output from several separate generators.
+
+### Class hierarchy and members are merged across partial classes
```
jakemac53: What order are mixins applied in? This at least needs to be defined :)

working/0678 - partial classes/feature-brainstorm.md
```diff
@@ -0,0 +1,199 @@
...
+Discussion about this proposal should go in [Issue #678](https://github.com/dart-lang/language/issues/678).
+
+## Motivation
+
+A critical use of code generation is adding boilerplate to hand-written classes.
+
+An ideal solution would provide:
+
+- Strict separation between human- and computer-generated code.
+- Allow computer-generated code to add members (functions, fields, properties,
```
jakemac53: This does have some compile time implications - all codegen has to be done before a library can be summarized/outlined.

working/0678 - partial classes/feature-brainstorm.md
```diff
@@ -0,0 +1,199 @@
...
+part 'example.g.dart';
+
+@JsonSerializable(nullable: false)
+class Person {
+  final String firstName;
+  final String lastName;
+  final DateTime dateOfBirth;
+  Person({this.firstName, this.lastName, this.dateOfBirth});
+
+  // No special wiring needed!
```
jakemac53: When looking at this class as a user how would I know the toJson/fromJson exist? The annotation is kind of a hint but annotations aren't required here.

You would have to click through to the part file to see these extra memebers and there isn't any indication that there are additional members.

This is less boilerplate though for the user :man_shrugging:

working/0678 - partial classes/feature-brainstorm.md
```diff
@@ -0,0 +1,199 @@
...
+Here's a motivating example with a speculative future implementation for
+[json_serializable](https://pub.dev/packages/json_serializable).
+
+### Before (current behavior)
+
+```dart
+// before: User-generated code
+import 'package:json_annotation/json_annotation.dart';
+
+part 'example.g.dart';
```
lrhn: Maybe this `part` declaration needs to be modified too, so tools won't complain if it *isn't* there yet. Currently, this file is just invalid Dart when `example.g.dart` does not exist. We likely want to continue complaining about missing imports, so it might be necessary to state that this file can be omitted temporarily for static analysis.

It is still worrysome that your code might change when importing the part file.
If the part file declares a public name, it might start shadowing an imported name, and the meaning of code changes between analysis without the part and analysis with the part.

Same can happen for identifiers inside instance members which suddenly see part-file introduced members that weren't there before.

It also means that code using the generated API still can't be analyzed before the part is available. If the `C.foo` method is only introduced by the generated code, with no stub, then any code which does `C c = ...;c.foo();` will not be correct without the generated file. You can add the stub, but then you really have to add *all* the stubs, which does not seem practical (or the goal of this approach).

Or perhaps the analyzer could have a mode where it can see that *some* import is missing *and* the library contains partial classes, so it should assume that any non-conflicting member access on that class is going to be fine (no more than a hint that 
> The C.foo member is not available yet. The following files are still missing: "example.g.dart"

or similar). It won't work if the entire class is introduced in the `part`.



The Dart system patch files are similar in some ways and they cannot introduce new public names, and the library itself cannot see private names from the patch file (it can define an `external` private member first, then the patch can provide the implementation). This also means that the generated code can use any private name for its implementation without risking conflicts.

That approach conflicts with goal of allowing generated code to add public members which do not have stubs, though.

working/0678 - partial classes/feature-brainstorm.md
```diff
@@ -0,0 +1,199 @@
...
+```dart
+class Person {
+  int get value => 42;
+}
+```
+
+- The "shape" of `Person` is now visible to static analysis tools even before
+  code generation is ran.
+
+### Open questions
```
lrhn: How about static members:
```dart
// file1.dart
part "part1.dart";
class C {}

// part1.dart
part of "file1.dart"
class C {
  static int counter = 0;
}
```

Since you can add constructors, you should also be able to add static members.

working/0678 - partial classes/feature-brainstorm.md
```diff
@@ -0,0 +1,199 @@
...
+class Person {
+  int get value => 42;
+}
+```
+
+...and would be treated as
+
+```dart
+class Person {
+  int get value => 42;
```
lrhn: What if the signatures do not match?

Say:
```dart
class P {
  external num get foo;
}

class P {
  int get foo => 42;
}
```
Is this valid or an error? 

It could be valid, with the `P` interface having either `num get foo;` or `int get foo;` as member signature and the implementation being `int get foo => 42;`, a valid implementation of either interface.

There probably has to be a most specific signature among the available ones, just as if we were merging different interfaces, so I'd go with `int get foo;` for the interface too.
That, or disallowing any difference.

working/0678 - partial classes/feature-brainstorm.md
kevmoo: The question

```dart
class Person with A{}
class Person with B{}
class Person with C{}
```

Should be treated as 

```dart
class Person with A, B, C{}
```

But if the `Person` definitions are across many parts, whe application of each mixin should be deterministic. Trivial thought: in order they are defined in each file, starting w/ defining library, then each part file in the order in which the directive appears in the deefining library

working/0678 - partial classes/feature-brainstorm.md
kevmoo: Good point. Adding stubs now.

working/0678 - partial classes/feature-brainstorm.md
kevmoo: ack

working/0678 - partial classes/feature-brainstorm.md
```diff
@@ -0,0 +1,199 @@
...
+
+A critical use of code generation is adding boilerplate to hand-written classes.
+
+An ideal solution would provide:
+
+- Strict separation between human- and computer-generated code.
+- Allow computer-generated code to add members (functions, fields, properties,
+  constructors, etc) to human-generated code directly – without requiring
+  subclasses, mix-ins, or manually connecting generated private members with user-created public members.
+- Allow user-created "stubs" to be added filled in by generated code. This
```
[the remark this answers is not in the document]
kevmoo: ack

working/0678 - partial classes/feature-brainstorm.md
```diff
@@ -0,0 +1,199 @@
...
+  a number of static-analysis and tooling scenarios to work without the code
```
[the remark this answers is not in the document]
kevmoo: ack

working/0678 - partial classes/feature-brainstorm.md
```diff
@@ -0,0 +1,199 @@
...
+## Design ideas
+
+### Legal to define the same type multiple times in the same library
+
+The following would be legal:
+
+```dart
+class Person{}
+class Person{}
+class Person{}
```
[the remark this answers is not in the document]
kevmoo: putting `partial` in after the declaration – trying to follow async/await and avoid a breaking change for anyone using `partial`

working/0678 - partial classes/feature-brainstorm.md
```diff
@@ -0,0 +1,199 @@
...
+...and would be treated as
+
+```dart
+class Person {
+  int get value => 42;
+}
+```
+
+- The "shape" of `Person` is now visible to static analysis tools even before
+  code generation is ran.
```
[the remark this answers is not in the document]
kevmoo: SGTM.

working/0678 - partial classes/feature-brainstorm.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Open questions
+
+* Do we want a `partial` keyword or similar?
```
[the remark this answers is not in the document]
kevmoo: Sure?

jakemac53:
Random thought I was having here - how much value does this provide over extensions? Should we improve extensions instead so you can add constructors and static methods?

lrhn:
> Random thought I was having here - how much value does this provide over extensions? Should we improve extensions instead so you can add constructors and static methods?

I think this feature could provide much that *static* extension methods do not. It's a different feature solving a different problem.

Partial classes are generally brought up in the context of generated code. If you have class which includes both generated members and pre-written members, then you currently have to either rewrite the class around the pre-written members or add one of the kinds of code to the other using mixins. The code gets muddled in unrelated glue features and you often need to repeat names to do the linking.

A partial class declaration would allow a class to be declared in total by code living in two different files. Maybe, even likely, files of the same library, but it still allows you to generate an entire file at a time and get automatic linking without the two parts having to explicitly mention each other.

The methods will still be first-class virtual methods, not static extension methods.

working/0678 - partial classes/feature-brainstorm.md
lrhn: That's consistent, even if somewhat arbitrary. It will be the only place where the order of declarations matter, so I'd prefer to not go there, and simply not allow a partial class part to add a mixin any more than I'd allow it to change the superclass otherwise.

jakemac53:
I understand the difference - but I think extension methods could still solve a lot of the same problems if they were more powerful. If you define an extension in the same library as a class then it starts to look pretty similar to actual  methods on the class. Obviously you can't add fields and whatnot, but the virtual vs static dispatch distinction matters less in that case in a practical sense.

That doesn't mean its not worth doing this - it is a very different feature to be sure.

working/0678 - partial classes/feature-brainstorm.md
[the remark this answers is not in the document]
lrhn: I'm sure it's a built-in identifier. I *think* it might be useful.
```dart
part class Foo { 
   int theFoo() => 42;
}
```
I think it reads decently.

kevmoo:
FYI: I'm excited about this feature, but I don't plan to push on it until at least Q2 (April) 2020.

Hoping I can find someone a bit more clue-full about language design to take it on 😄

listepo:
Any news?

jyardin:
Hello, thank you for this proposal!
I think it is important to make code generation (or any form of metaprogramming) more powerful in dart since you cannot rely on reflection in flutter.

I have one question though. 
In your 'After' json_serializable code example:
```dart
// after: User-generated code
import 'package:json_annotation/json_annotation.dart';

part 'example.g.dart';

@JsonSerializable(nullable: false)
class Person partial {
  final String firstName;
  final String lastName;
  final DateTime dateOfBirth;
  Person({this.firstName, this.lastName, this.dateOfBirth});

  // These stubs ensure the shape of the type is fully visible before code
  // generation to enable static analysis. The stubs can also be available
  // in the analyzer API allow code generation to use them. For instance:
  // json_serializable could drop the explicit annotations to enable/disable
  // factory and/or toJson members and instead just key off the existence of
  // the corresponding stubs.
  partial factory Person.fromJson(Map<String, dynamic json>);
  partial Map<String, dynamic> toJson();
}
```
the stubs are optional, right ?
If not, this would defeat the purpose of reducing boilerplate in my opinion.

listepo:
@kevmoo any news?

kevmoo:
Not yet. We're all working really hard to get null safety finished. What scenarios are you looking to use partial classes with?

listepo:
@kevmoo for code generation.

kevmoo:
That's our inspiring scenario. Just double-checking. 👍

kevmoo:
Ditto here @munificent – happy to close this out RE https://github.com/dart-lang/language/blob/master/working/augmentation-libraries/feature-specification.md SGTM!

mit-mit:
Closing as stale. We'll continue to explore augmentation libraries (linked from the previous comment)

PR #683, opened 2019-11-14
Allow `E(e)?.id` in extension methods.

lrhn:
It is believed that the current implementations of both analyzer and front-end allows `E(e)?.foo` where `E` is an extension. This was previously not allowed by the design document.
It has a reasonable behavior, and we have chosen to allow it. This updates the design document to do so, and specify the behavior.

There will be tests.

### DIFF
--- a/accepted/future-releases/static-extension-methods/feature-specification.md
+++ b/accepted/future-releases/static-extension-methods/feature-specification.md
@@ -1,6 +1,6 @@
 # Dart Static Extension Methods Design
 
-lrn@google.com<br>Version: 1.5<br>Status: Design Document.
+lrn@google.com<br>Version: 1.6<br>Status: Design Document.
 
 This is a design document for *static extension members* for Dart. This document describes the feature's syntax and semantics.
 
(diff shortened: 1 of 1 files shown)


accepted/future-releases/static-extension-methods/feature-specification.md
```diff
@@ -119,36 +119,48 @@ We defined the *instantiated `on` type* of `E` as the `on` type of the declarati
 
 We define the *instantiate-to-bounds `on` type* of an extension as the `on` type with type parameters replaced by the types that the instantiate-to-bounds algorithm would derive for any type parameters of the extension.
 
-A *simple member invocation* on a target expression `X` is an expression of one of the forms:
+A *simple member invocation,* null aware or not, on a target expression `X` is an expression of one of the forms:
```
eernstg: The possible forms that `X` can have are not particularly obvious. Forms that contain a selector but no `=` could be derived from `<postfixExpression>` via `<primary> <selector>*` or `<constructorInvocation> <selector>*`, and the forms containing `=` could have their left hand side derived from `<assignableExpression>`.

But cascades can occur in many of these terms, unless that possibility is explicitly ruled out.

In particular, if we derive `X.id` from `<primary> <selector>*` then `.id` may not be a syntactic subterm of `X.id`, it is the bottom&right-most subterm of a term derivedr from `<selector>*`. So we can't just say "derive `X`, and then derive `X.id` from `X`", because there is no grammar rule which can do that.

So we can't say that `X.id` and the other forms reflect the syntactic structure, which means that they'll have to stand for something else (a mere sequence of tokens?), in which case `X` might as well be `x..bar().foo = y` such that `X.id` is `x..bar().foo = y.id`.

I think we would need to characterize `X` in a way which ties it down a bit more than just "it's a sequence of tokens", but it is probably not enough to say that it is one of those `<...> <selector>*` forms.

lrhn: I think this is sufficient for now for the feature specification, but we'll have to fiddle with it when integrating into the full spec.

PR #684, opened 2019-11-14
Move old supporting docs to accepted

mit-mit:
Move old supporting static extension member docs to /accepted/2.6/static-extension-members/old/


lrhn:
Check if there are any links to these documents from the corresponding issues.
If not, LGTM.

mit-mit:
Yup, ran a search and the only links I could find was a relative (current dir) link from one of them to the other

PR #699, opened 2019-11-21
Create proposal for enhanced default constructors.

lrhn:


working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Goal
+
+We combine the forwarding constructor feature with default constructors and automatic initialization of declared instance variables. This *unifies* default constructors and mixin application forwarding constructors *and* adds more power to the default constructor for simple classes.
+
+We expect this to happen after non-nullable types have been added to the language, so the proposal is written in terms of that, with a few guidelines for how legacy libraries work where nullability matters.
+
+## Proposal
+
+Let *C* be a class declaration with name `C` and superclass *S* with name `S` and which *declares* no constructor.
```
munificent: Now that the default constructor is manifestly useful, I think we should anticipate that users will want to have access to it even in classes that define other constructors. In fact, a useful pattern would probably be to have other constructors that just forward to the now-very-useful default one. It practically replaces field initialization lists, pretty much like primary constructors in Scala and Kotlin.

At the same time, users will sometimes want to hide that default constructor, so if declaring a constructor does not hide it, we'd need some other mechanism to do so.

5-minute bikeshed proposal:

* If a class declares no constructors, you get the default one implicitly. This optimizes for the data class case.

* If a class does declare a constructor, or you wish to rename the default constructor, you can add it explicitly using `= default;`:

    ```dart
    class Point {
      num x, y;
      Point.cartesian = default;
    }
    ```

    (This might also give you a notation for making it a *const* constructor if desired.)

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
+
+It is a compile-time error if:
+
+- *C* declares a private-named and potentially non-nullable instance variable with no initializer expressions.
```
munificent: This restriction may end up being annoying. But I can see us keeping it for now and then revisiting it if we come up with a better way of defining "properties": some combination of getter/setter/backing field.

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
+Let *C* be a class declaration with name `C` and superclass *S* with name `S` and which *declares* no constructor.
+
+It is a compile-time error if:
+
+- *C* declares a private-named and potentially non-nullable instance variable with no initializer expressions. 
+- Or in **NNBD-legacy mode**,  *C* declares a private-named and final instance variable with no initializer expression.
+
+Otherwise, for each constructor *s* of the superclass, named `S` or `S.id`, satisfying all of the following:
+
+1. If *s* is a named constructor with name `S.id` then:
```
munificent: Format nit: Since this is a list of criteria and not an ordered algorithm, I'd use bullets instead of numbers.

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
+   1.  *s* is accessible ( `id` is not a private identifier or *S* is not declared in a different library), and
+   2.  *C* does not declare a static member named *id*.
+2. The *s* constructor is a generative constructor.
+3. *C* does not declare an instance variable with no initializer expression and with the same basename as a required named parameter of *s*.
```
munificent: Is this restriction necessary? It would be weird for both the sub and superclass to have redundant fields, but would still work to simply initialize sub's field with the parameter and forward it to the superclass.

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
+4. Either *C* does not declare any instance variables or *s* does not have any optional positional parameters (required because we can't declare a constructor with both optional position parameters and named parameters).
```
munificent: Ugh, we should remove that limitation. I didn't get to with UI-as-code because I bundled it in the same proposal that added rest params, but maybe we should sever that piece and make it its own proposal.

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
+
+add a generative default constructor *c* to *C* with the corresponding name `C` or `C.id`, the following parameters:
```
munificent: "the following" -> "with the following"

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
+
+and with an initializer list consisting of a single `super` invocation of either `super`, if *s* is named `S`, or `super.id`, if *s* is named `S.id`, with one argument for each parameter of *c* corresponding to a parameter of *s*, forwarding the value of the *c* parameter as an argument to the corresponding *s* parameter.
+
+The constructor *c* is `const` if *s* is `const` and *C* does not declare *any* instance variables.
```
munificent: I don't use const enough to recall all the dark corners, but it is not possible to support a default const constructor with fields? Is the problem that we don't want to silently give the user a *const* constructor unless they ask for it?

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
+- For each named parameter of *s*, *c* has a corresponding parameter with the same name, type, optionality and default value (if optional), unless a named initializing formal parameter with the same name was declared above (in which case the parameter of *s* must be optional per requirement 4. above).
+- For each positional parameter of *s*, whether optional or not, $c$ has a corresponding positional parameter with the same position, type, optionality and default value (if optional). For documentation purposes, the parameter will have the same name as the corresponding positional parameter of *s* unless that conflicts with one of the named initializing formals added in the previous point. In that case, the name will be modified by adding a `$` and a positive decimal integer numeral with no leading zeros after the name, where the numeral represents the smallest positive (&ge; 1) integer sufficient to make the name not be the same as any of the named parameters of the constructor, not the same as the name of any later positional parameter of *s*, and not the same as any earlier positional parameter of *c*.
```
munificent: Well thought out! :)

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
+
+and with an initializer list consisting of a single `super` invocation of either `super`, if *s* is named `S`, or `super.id`, if *s* is named `S.id`, with one argument for each parameter of *c* corresponding to a parameter of *s*, forwarding the value of the *c* parameter as an argument to the corresponding *s* parameter.
+
+The constructor *c* is `const` if *s* is `const` and *C* does not declare *any* instance variables.
+
+It is a *compile-time error* if there are no superclass constructors satisfying the requirements. (This is not a necessity, it's completely valid for a class to have no generative constructors, which is what currently happens if a class declares only factory constructors. We require the author to explicitly opt in to this behavior because *if* it is unintentional, it is a very hard-to-spot error.)
```
munificent: What if the class is abstract and is intended to be used as an interface? I guess in that case, it should be implementing the superclass instead of extending it anyway, so I think this makes sense?

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
+The following class declarations will be valid:
+
+```dart
+class Person {
+  final String firstName;
+  final String lastName;
+  final int age;
+}
+
+class Fireman extends Person {
```
munificent: Suggestion: Use a non-gendered example, or perhaps even a non-person one.

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
+}
+```
+
+You can instantiate a `Fireman` as:
+
+```dart
+ Fireman(firstName: "John", lastName: "Doe", age: 37, yearsOfService: 12)
+```
+
+This is still verbose, because of the named arguments.
```
munificent: I don't think that's particularly verbose. :) Flutter users will probably feel this looks entirely normal.

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
+
+You can extend classes with non-default constructors and forward implicitly to those as well:
+
+```dart
+class ColorBox extends Rectangle<int> {
```
munificent: A self-contained example would be good since users probably don't recall Rectangle offhand.

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
+
+```dart
+class MyRectangle extends Rectangle<int> {
+  MyRectangle affineTransform(int a, int b, int c, int d) => 
+      MyRectangle(a * left + b * top, c * left + d * top, 
+                  a * right + b * bottom, c * right + d * bottom);
+}
+```
+
+Here you implicitly "inherit" all constructors from the superclass without having to rewrite them.
```
munificent: This implies that adding a constructor to a class will cause new constructors to spontaneously appear in all subclasses (that don't define their own constructors). If the superclass is one you don't control, you might not want your own class signing up for that extensibility.

In particular, this interacts really interestingly with package management. Consider a package foo:

### foo 1.0.0:

#### pubspec.yaml:

```yaml
name: foo
version: 1.0.0
```

#### lib/foo.dart:

```dart
class Foo {
  Foo.one();
}
```

Later, the maintainer adds another constructor:

### foo 1.1.0:

#### pubspec.yaml:

```yaml
name: foo
version: 1.1.0
```

#### lib/foo.dart:

```dart
class Foo {
  Foo.one();
  Foo.two();
}
```

This is a non-breaking change, so they just bump the minor version. Now consider a second package bar:

### bar 1.0.0:

#### pubspec.yaml:

```yaml
name: bar
version: 1.0.0
dependencies:
  - foo: '^1.0.0'
```

### lib/bar.dart:

```dart
class Bar extends Foo {}
```

In principle, the dependency constraint that bar places on foo is fine. bar should work with any version of foo compatible with foo 1.0.0.

But imagine some app that depends on bar 1.0.0. Can it call `Bar.second()`? Well, that depends on which version of foo got selected. If some other shared dependency on foo caused it to be pinned to 1.0.0, the answer is "no". Otherwise, the answer is "yes".

Do you think it's safer to limit this to the unnamed constructor? One option, if you like my suggestion for an explicit `= default` syntax is to only use the unnamed constructor implicitly and let users opt in to named ones using that syntax.

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Notes
+
+All the field initializers are named parameters. We do not want to make them positional parameters for two reasons: Primarily because it means that we have to pick an order, and alphabetical order is arbitrary, while source order makes the code unstable in the face of refactoring (and any other order is even more arbitrary). Secondarily because it makes it even harder to consistently forward to superclass constructors which also have named parameters.
+
+In NNBD libraries, we do not require that *final* instance variables are always initialized, but we do require that *potentially non-nullable* variables are. We could require that final fields are initialized, but it's very reasonable to treat being *nullable* as meaning being optional. Given that there is an initializing formal parameter for the field, it will definitely be initialized
```
munificent: Nit: "." at end.

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
...t` only when the subclass declares no instance variables. We could allow final instance variables with no initializer expression (because then the constructor must initialize it) or with an initializer expression which is constant (because then it's valid in a `const` constructor class). The latter requirement is particularly fragile and useless. Changing the initializer expression of a variable from `final int x = a + 2;` to `final int x = a + b;` where `a` is a constant variable and  `b` is a non-constant variable with the value `2` will change the constructors of the class from constant to non-constant. Also, a final instance variable initialized to a constant value should probably just be a getter. We want to avoid accidentally making a class have constant constructors, because that may become a support burden for the author, so we only make constructors constant when the subclass is particularly simple, and where the current mixin application constructor forwarding makes the constructor constant. (It is not a *breaking* change to make more constructors constant in the future, but it is a *bad* change for library owners which intended a class to have non-constant constructors).
```
munificent: +1, though see my above suggestion about maybe letting users opt in to const default constructors.

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
+
+You only get default constructors when you write *no* constructors. If you want to forward ten superclass constructors, and change the eleventh just a bit, or add an eleventh factory constructor, then you have to write all of the other ten in full. If this is too annoying, we may be able to add brief syntax for forwarding a constructor verbatim, say `Foo.bar = super.bar`.
```
munificent: +1 to it being annoying. When possible, I like to avoid features where one minor change necessitates a cascade of other changes.

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
+
+It's always possible to write exactly the same constructors manually, so this change does not introduce any new power. It's always possible to move away from default constructors to explicitly written constructors (except for classes used as mixins, which are always going to be highly restricted). As such, feature has no back-end impact, it can be implemented entirely in the front-end. Back-ends may want to tree-shake unused constructors, or unused constructors parameters, though.
```
munificent: When you consider how this interacts with version constraints, it does give you a special power at the package level where a class inheriting a class from another package can then get a set of constructors determined by pub and not the class itself. Probably not a *good* power...

working/0698 - Enhanced Default Constructors/proposal.md
leafpetersen: I think this makes it a breaking change to add an optional parameter to a constructor.  Are we comfortable with that?

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
+2. The *s* constructor is a generative constructor.
+3. *C* does not declare an instance variable with no initializer expression and with the same basename as a required named parameter of *s*.
+4. Either *C* does not declare any instance variables or *s* does not have any optional positional parameters (required because we can't declare a constructor with both optional position parameters and named parameters).
+
+add a generative default constructor *c* to *C* with the corresponding name `C` or `C.id`, the following parameters:
+
+- For each non-private instance variable with no initializer expression declared by *C*, where `v` is the name of the instance variable, the constructor has a named initializing formal parameter `this.v` with no default value. The named parameter is *required* if the instance variable has a potentially non-nullable type.
+- For each named parameter of *s*, *c* has a corresponding parameter with the same name, type, optionality and default value (if optional), unless a named initializing formal parameter with the same name was declared above (in which case the parameter of *s* must be optional per requirement 4. above).
```
leafpetersen: This makes it a breaking change to add an optional named parameter to a constructor.

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
+1. If *s* is a named constructor with name `S.id` then:
+   1.  *s* is accessible ( `id` is not a private identifier or *S* is not declared in a different library), and
+   2.  *C* does not declare a static member named *id*.
+2. The *s* constructor is a generative constructor.
+3. *C* does not declare an instance variable with no initializer expression and with the same basename as a required named parameter of *s*.
+4. Either *C* does not declare any instance variables or *s* does not have any optional positional parameters (required because we can't declare a constructor with both optional position parameters and named parameters).
+
+add a generative default constructor *c* to *C* with the corresponding name `C` or `C.id`, the following parameters:
+
+- For each non-private instance variable with no initializer expression declared by *C*, where `v` is the name of the instance variable, the constructor has a named initializing formal parameter `this.v` with no default value. The named parameter is *required* if the instance variable has a potentially non-nullable type.
```
leafpetersen: This makes it a breaking change to switch between a field and getter unless you remember to also rewrite the constructor.

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
+# Dart Default Constructor Enhancement
```
leafpetersen: This looks very cool.  It would be good to have a section discussing the impact on breaking changes.  I'm a little concerned about the fact that this makes small changes to the class or superclass show up invisibly in the constructor in ways that are breaking.  In the superclass case, I think that means that more things are breaking changes.  In the class case, I think (but am not sure) that you can always rewrite the constructor explicitly to avoid the breaking change, but that's easy not to notice.  Maybe that's an advantage of the primary constructor syntax of Kotlin?

working/0698 - Enhanced Default Constructors/proposal.md
munificent: All the more reason to do sealed by default. :)

working/0698 - Enhanced Default Constructors/proposal.md
mraleph: I really like this feature but I think one important thing is misses is what I call, for the lack of better word, "stability" (in parallel with mathematical notion of [stability](https://en.wikipedia.org/wiki/Stability_theory)) - if you want to make small semantic change, you can achieve it with a small syntactic change. 

For example, imagine I want a constructor that behaves like "enhanced default constructor" but has few `assert`-s. 

I can't do that with few changes - I suddenly need to create the full constructor which initialises all the fields explicitly. 

Kotlin constructor syntax has stability property because features their combine and build on top of each other - rather than being "exclusive" (you either have enhanced constructor - or old school constructor... why not both?)

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
+
+and with an initializer list consisting of a single `super` invocation of either `super`, if *s* is named `S`, or `super.id`, if *s* is named `S.id`, with one argument for each parameter of *c* corresponding to a parameter of *s*, forwarding the value of the *c* parameter as an argument to the corresponding *s* parameter.
+
+The constructor *c* is `const` if *s* is `const` and *C* does not declare *any* instance variables.
+
+It is a *compile-time error* if there are no superclass constructors satisfying the requirements. (This is not a necessity, it's completely valid for a class to have no generative constructors, which is what currently happens if a class declares only factory constructors. We require the author to explicitly opt in to this behavior because *if* it is unintentional, it is a very hard-to-spot error.)
+
+It is a *compile-time error* to derive a mixin from a class unless its superclass is `Object`, it declares no generative constructors, and it declares no potentially non-nullable instance members without an initializer expression. (For NNBD-legacy code, it must declare no final fields without initializer expressions.)
+
+### Examples
```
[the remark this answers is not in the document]
lrhn: That's a good idea for a motivational document, and something I'd omit entirely for a design document ("just the facts, mam!"). Since I have examples, this is clearly a motivational document.

working/0698 - Enhanced Default Constructors/proposal.md
lrhn: There is not just *one* default constructor if your superclass has multiple generative constructors. You get initializing forwarding constructors for all of them.

Suggested syntax for default constructor:
```dart
default Point.cartesian; // no super, so use `:super()`.
```
or, if you want to forward to specify the super-constructor.
```dart
default ColorPoint: super.cartesian;
```

I guess we could say that:
* If your class has no constructors, you get the full forwarding initializing constructors.
* If your class declares at least one constructor, no explicit default constructor, no unnamed constructor, and your superclass has an unnamed constructor, you get a default forwarding unnamed constructor `default MyName: super;`.

There's still no way to *avoid* that short of declaring `default MyName._preventDefault;`. I'm not happy about that. I think I'd prefer to just make you write `default MyName;` in that case, and not give you a default constructor if you declare *any* constructor explicitly.

working/0698 - Enhanced Default Constructors/proposal.md
lrhn: The problem here is that such a variable *must* be initialized, and we *can't* initialize it with a privately named parameter. You have to specify explicitly how to initialize that field, we can't do it for you.

working/0698 - Enhanced Default Constructors/proposal.md
lrhn: That might be a good solution.
The *default* constructor keeps being the unnamed constructor, but you can declare *a* default constructor manually.

working/0698 - Enhanced Default Constructors/proposal.md
lrhn: Also reasonable when you can write `const default Foo;` and not have to write all the parameters.

working/0698 - Enhanced Default Constructors/proposal.md
lrhn: OK, I'm convinced. Will rewrite to do that.

working/0698 - Enhanced Default Constructors/proposal.md
lrhn: True. That's always going to be the case. If a class adds an extra optional parameter to its unnamed constructor, then any forwarding constructor will also get that extra parameter. That's what forwarding means.

working/0698 - Enhanced Default Constructors/proposal.md
lrhn: Not particularly. Can we have both kinds of optionals on the same constructor? (PLEASE!!!!)
Otherwise forwarding to a non-default constructor is dead in the water. Which is sad, but maybe necessary for now.

working/0698 - Enhanced Default Constructors/proposal.md
lrhn: It makes it a breaking change to add a *required* parameter to the superclass constructor, but that was always the case.

Adding a new field named `x` to the subclass may change the meaning of calling `Foo(x: 42)`, if it previously referred to a superclass constructor parameter. 
That's also not new, you change the meaning of an existing constructor. You should be aware of the super-constructor parameters. The only issue is that the change is not directly visible in the source, so you might not be aware.

That kind of conflict is unavoidable if we want to forward *any* parameter to the superclass constructor, which I do believe is valuable.

working/0698 - Enhanced Default Constructors/proposal.md
lrhn: If we can't allow that, then default initializing constructors are impossible.
It's also a breaking change to change a field to a getter/setter with an *explicitly written* constructor, because you can no longer use `this.x` to initialize it.

I think it's reasonable. You are editing a class, and the implicit constructor added to that class may change depending on the editing. You are in the perfect position to fix that too, and your tests should catch the change.

working/0698 - Enhanced Default Constructors/proposal.md
lrhn: Stability, non-chaotic syntax, ease of migration, etc. It's a good idea.

It's *hard* to do here because we are automatically generating a significant amount of syntax.

Maybe we could use the `default` word in the parameter list:

```dart
  Foo(default) : super.bar;
```
The `default` stands for named initializers for all uninitialized public non-late fields of the current class plus forwarding parameters for all parameters of the `super.bar` constructor (with some conflict handling).

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,166 @@
...
+
+A *mixin application* receives corresponding forwarding constructors for all accessible generative constructors of the superclass. The forwarding constructor has the same basename (identifer after the class name for named constructors, empty for the unnamed constructor), the same parameter signature as the superclass constructor, including default values for optional parameters, and an initializer list with a `super`-invocation to the corresponding superclass constructor which forwards each parameter as an argument to the corresponding superclass constructor parameter. The forwarding constructor is `const` if the superclass constructor is `const` *and* the mixin declares no instance variables.
+
+Since a mixin application class cannot contain static members, there is no risk of the introducing a conflict between a constructor name and a static member. 
+
+### Goal
+
+We combine the forwarding constructor feature with default constructors and automatic initialization of declared instance variables. This *unifies* default constructors and mixin application forwarding constructors *and* adds more power to the default constructor for simple classes.
```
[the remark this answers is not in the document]
lrhn: True. I was hoping to be able to unify all the concepts, but even without that, initializing default constructors are useful.

working/0698 - Enhanced Default Constructors/proposal.md
lrhn: This is not new, just different.

If my library v 1.0.0 declares:
```dart
  class C {}
```
and v1.1.0 adds:
```dart
  class C {
    static int foo() => 42;
  }
```
then someone depending on >=1.0.0 *may* get 1.1.0 and think that `C.foo` exists.
That's why we add `@Since("2.1")` to the platform libraries so that the analyzer can trigger on non-breaking features introduced after your min-dependency.

Versioning is not safe unless you lock yourself to *one* version, or always check your code against the *oldest* version you claim compatibility with.
It's not enough to use the oldest solution to the current constraints, because you might depend on some other package which depends on 1.1.0 of my package above. If they loosen their constraint to accept 1.0.0 later, then you can again end up being run on 1.0.0. So, only safe check is to analyze your code against the oldest supported version of each package that you use, even if they are not mutually compatible. (And that may not make sense at all, if packages are exporting types from each other, so maybe it's just impossible).

working/0698 - Enhanced Default Constructors/proposal.md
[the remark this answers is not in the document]
munificent: > something I'd omit entirely for a design document

I think a design doc still needs motivation, otherwise it's impossible to evaluate whether it's a *good* design.

working/0698 - Enhanced Default Constructors/proposal.md
munificent: > I think I'd prefer to just make you write `default MyName;` in that case, and not give you a default constructor if you declare _any_ constructor explicitly.

+1.

working/0698 - Enhanced Default Constructors/proposal.md
munificent: > If my library v 1.0.0 declares:
> 
> ```dart
>   class C {}
> ```
> 
> and v1.1.0 adds:
> 
> ```dart
>   class C {
>     static int foo() => 42;
>   }
> ```
> 
> then someone depending on >=1.0.0 _may_ get 1.1.0 and think that `C.foo` exists.

This case is different. If you want to call `C.foo()`, you need to depend on the version of your package that includes it, 1.1.0.

My example involves *three* libraries. My app depends on bar, which transitively depends on foo. I have *no* direct dependency on foo at all. But with the current proposal, the set of constructors that *Bar* exposes depends on the version of *foo* that is selected with my app.

working/0698 - Enhanced Default Constructors/proposal.md
munificent: Fair, though I think the scope of that is much smaller than having entirely new named constructors appear on a class.

working/0698 - Enhanced Default Constructors/proposal.md
munificent: > Can we have both kinds of optionals on the same constructor?

YESPLZ.

working/0698 - Enhanced Default Constructors/proposal.md
munificent: > Maybe we could use the `default` word in the parameter list:

Not a bad idea. I think a user suggested this several years ago if I remember right. It feels maybe a little too magic to me, but it does nicely encapsulate exactly the sugar that's wanted — a synthesized parameter list of initializing formals — and gives users maximum flexibility to use that sugar how they want.

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -13,13 +13,14 @@ Creating a class with a simple constructor is fairly short because of Dart's ini
 
 ```dart
 class Complex {
-  final double real, imaginary;
+	final double real, imaginary;
```
munificent: Nit: Tab -> spaces.

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -41,126 +42,253 @@ Since a mixin application class cannot contain static members, there is no risk
 
 ### Goal
 
-We combine the forwarding constructor feature with default constructors and automatic initialization of declared instance variables. This *unifies* default constructors and mixin application forwarding constructors *and* adds more power to the default constructor for simple classes.
+We want to allow classes with fields that can be constructed without having to write constructors. That is, we want *initializing default constructors* which takes parameters class fields (with some exceptions) and initializes the fields.
```
munificent: Some grammar errors here. Maybe "which takes parameters for each class's fields and initializes the fields from them."

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -41,126 +42,253 @@ Since a mixin application class cannot contain static members, there is no risk
...
-
-The constructor *c* is `const` if *s* is `const` and *C* does not declare *any* instance variables.
-
-It is a *compile-time error* if there are no superclass constructors satisfying the requirements. (This is not a necessity, it's completely valid for a class to have no generative constructors, which is what currently happens if a class declares only factory constructors. We require the author to explicitly opt in to this behavior because *if* it is unintentional, it is a very hard-to-spot error.)
-
-It is a *compile-time error* to derive a mixin from a class unless its superclass is `Object`, it declares no generative constructors, and it declares no potentially non-nullable instance members without an initializer expression. (For NNBD-legacy code, it must declare no final fields without initializer expressions.)
+If possible, it should be easy to migrate to and from using the feature. Having to rewrite significant parts of your class in order to move from default constructors to custom constructors is annoying. We will try to limit the amount of code you need to write when going from the simple case to a more complex case.
```
munificent: +1 for mentioning this explicitly!

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -41,126 +42,253 @@ Since a mixin application class cannot contain static members, there is no risk
...
+```
+
+would expand to;
+
+```dart
+  Person(this.lastName, {required this.firstName, required this.age}) : super()
+
+```
+
+because the explicitly supplied `lastName` parameter takes precedence over inserting an initializing formal parameter for that field.
```
munificent: Syntax bikeshed: Now that the syntax is really sugar for *part of a parameter list* and not *an entire constructor*, consider putting the syntax inside the parameter list:

```dart
  Person(this.lastName, default);
```

This would offer some interesting affordances:

```dart
  Person(this.lastName, default); // The field initializers are positional and required.
  Person(this.lastName, [default]); // The field initializers are positional and optional.
  Person(this.lastName, {default}); // The field initializers are named and optional.
  Person(this.lastName, {required default}); // The field initializers are named and required.

  Person(default, this.lastName); // Default ones come first.
```

leafpetersen: If we went this route, what about using `this` instead of `default`?  In the initializer list, you could then also allow `this.firstName` to refer to the implicit `firstName` parameter.

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,294 @@
...
+- If the forwarding constructor declares a parameter (including initializing constructor parameters introduced above) with the same name as a positional superconstructor parameter, call it *originalName*, then, for documentation purposes, the corresponding forwarding constructor parameter uses the name *originalName_n* where *n* is a decimal integer representing for the smallest integer greater than zero which makes the name different from any parameter name declared by the forwarding constructor or by the superconstructor.
+- If the forwarding constructor declares any named parameters, including if it is an initializing constructor, then all optional positional superconstructor parameters are ignored.
+- If the forwarding constructor declares any positional parameters after the superconstructor reference entry, then all positional superconstructor-derived parameters are ignored.
```
leafpetersen: I'm not sure what "positional superconstructor-derived parameters are".  Do you mean that:

```
class A {
  final int x;
  A(this.x);
}

class B extends A {
  B(super, String z);
}
```

expands the constructor in `B` to `B(String z) : super()`?

This is then presumably a compile time error?

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,294 @@
...
+- Otherwise the forwarded parameters are optional if they are optional in the superconstructor, and if so, they have the same default value, if any.
+
+Then a super-invocation is added to the initializer list where each such forwarded parameter is forwarded as a corresponding superconstructor argument.
+
+#### Example
+
+The class `Fireman`above could be given an initializing constructor by writing:
+
+```dart
+  default Fireman(super);
```
leafpetersen: Can this have an initializer list?  What about a body?

working/0698 - Enhanced Default Constructors/proposal.md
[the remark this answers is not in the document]
lrhn: That depends on whether you are *selling* the design or simply specifying what needs to be implemented. I have found that motivational text in the final design document is more confusing to implementors than they are helping. *Examples* might still be useful, but as explanatory, not motivational aides.

working/0698 - Enhanced Default Constructors/proposal.md
lrhn: Rewritten.

working/0698 - Enhanced Default Constructors/proposal.md
lrhn: No, missing an "optional" there.

working/0698 - Enhanced Default Constructors/proposal.md
lrhn: Yes to both.

working/0698 - Enhanced Default Constructors/proposal.md
lrhn: Putting the field initializer marker into the parameter list as *positional* parameters means that we get arbitrary ordering (source order most likely, because that's the only thing the user can actually control). I'd prefer to avoid that if possible.
That's why I put `default` outside, because it definitely will generate named parameters, and I didn't want to put two markers into the parameter list (the forwarding marker and the initializing marker). It gets too crowded.

If we did put it inside the parameter list, then `this` is actually a good marker, and it compares well with `super` for the forwarding parameters, and to the `this.name` for a specific field initializer. 
I'm just worried that if we *only* allow initializing parameters as named, then people will forget the `{...}` braces and just write `Foo(this)` instead of `Foo({this})`, and if we do allow them outside, users will have to order their fields for the constructor, not for what reads better.
(That said, I'm sure a lot of users will love it anyway, so maybe I'm being too careful).

working/0698 - Enhanced Default Constructors/proposal.md
lrhn: Fixed.

(Pro-tip: When transferring markdown between two computers trough a gmail draft mail, do remember to paste the *source*, not the formatted text).

working/0698 - Enhanced Default Constructors/proposal.md
lrhn: I have other issues with forwarding all named constructors, so I think I'll remove that again. Just the unnamed one, or nothing.

I should be easy to write `Foo.foo(super);` if you do want to forward anyway.

working/0698 - Enhanced Default Constructors/proposal.md
lrhn: I've reeled in the forwarding a little, so we never forward optional positional parameters if adding one would be breaking. It means that you can forward optional positional superconstructor parameters, as long as you put the `super` after all other positional parameters in the subclass constructor (and have no named parameters, obviously).

working/0698 - Enhanced Default Constructors/proposal.md
lrhn: So, brainstorm on allowing `this` as the "intialize all" marker anywhere in the parameter list.

For each class, there are a given number of auto-initializable fields (non-late, non-initializer-expression) fields. For named initialization, they must also be non-private. We probably want to omit private fields everywhere, and require you to do those manually, because otherwise we leak the private name (and leave you with no way to *avoid* the auto-initialization of a field).
Some of those initializers will have to be required

So, if you write `this` in the positional parameter part, then they will all be positional (even the boolean ones, which some people will scoff at). 

If it's in the optional positional parameter part `[this]`, then they must all be optional. If some are necessarily required, then that's an error.

If it's in the required positional parameter part, with any required parameter after it, then they must all be required. That's acceptable.

If it's at the *end* of the required positional parameter part, then we can make the tail of the parameters optional. Any required initializer makes all prior (source order) field initializers also required, but if the final ones are nullable, they can become optional.

It feels fragile.

If the `this` is in the named parameter part, then it works as now. Some are optional, those who must be are required.
If it's modified by `required`, then they're all required.

I'm not sure this buys use enough to be worth the complexity and general footgunnery.

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,285 @@
...
+
+Since a mixin application class cannot contain static members, there is no risk of the introducing a conflict between a constructor name and a static member. 
+
+### Goal
+
+We want to allow classes with fields without having to write constructors. That is, we want *initializing default constructors* which takes parameters for class fields (with some necessary exceptions) and initializes the fields.
+
+We want those constructors to be independent of the order of the fields in the class, so the constructor parameters will be *named*.
```
jakemac53: Does this mean that private fields are not allowed? We should at least call that out if so.

If they are allowed, that introduces cases where adding a _private_ field in a class is now a breaking change - because the name would collide with a public or private field initializer in any class that extends that class.

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,285 @@
...
+  final int x, y;
+  Point(this.x, this.y);
+  double get distanceFromOrigo => sqrt(x * x + y * y);
+}
+class Color3DPoint extends Point {
+  final Color color;
+  final int z;
+  final double _cachedDistanceFromOrigo;
+  // Expands to: Point(this.color, int x, int y, this.z) : super(x, y);
+  Point(this.color, super, this.z)
```
mnordine: Is this constructor supposed to be `Color3DPoint`?

working/0698 - Enhanced Default Constructors/proposal.md
```diff
@@ -0,0 +1,294 @@
...
+The class `Parcel` above could be given an initializing constructor by declaring a constructor of the form:
+
+```dart
+  default Parcel();
+```
+
+This would expand to the constructor:
+
+```dart
+  Person({required this.address, required this.weightInGrams}) : super();
```
mnordine: Still got a `Person` in there

PR #700, opened 2019-11-21
Ensure that upper bounds preserve legacy

leafpetersen:
This addresses a quirk observed by @scheglov that `UP(Never*, int*) ==  int?`, which is somewhat surprising.  This change ensures that `UP(S, T)` is never of the form `Q?` unless one of `T` or `S` is a `?` type.

cc @eernstg @munificent @lrhn


scheglov:
What is the status of this?
Should I implement it?

leafpetersen:
I'll land this, I think it's a good change.  cc @stefantsov

PR #702, opened 2019-11-23
Additional clarifications around late variables.

leafpetersen:


leafpetersen:
cc @scheglov @stefantsov @johnniwinther @liamappelbe @crelier @sigmundch @rakudrama @eernstg @munificent

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -217,7 +220,16 @@ assignable to `Object`.
...
 It is not an error for the body of a `late` field to reference `this`.
 
-It is an error for a formal parameter to be declared `late`.
+It is an error for a variable to be declared as `late` in any of the following
+positions: in a formal parameter list of any kind; in a catch clause; in the
+variable binding section of a c-style `for` loop, a `for in` loop, an `await
+for` loop, or a `for element` in a collection literal.
+
+It is an error for the initializer expression of a `late` local variable to use
+a prefix `await` expression.
```
scheglov: It is an error in this case:
```dart
main() async {
   late v = await 42;
}
```

But is it an error in this case?
```dart
main() async {
  late v = () async {
    await 42;
  };
}
```

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -217,7 +220,16 @@ assignable to `Object`.
...
+
+It is an error for a class with a `const` constructor to have a `late final`
+field.
```
scheglov: Does this include `static` fields? Or the set of modifiers should be exactly `late final`?

lrhn: No. I'm guessing that "field" means "instance variable" here. A "field" is not a formally specified term.

accepted/future-releases/nnbd/feature-specification.md
lrhn: No.

So,  maybe phrase it as:

> It is a compile-time error if the initializer expression of a `late` local variable contains an `await` expression which is not inside another function expressions.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -217,7 +220,16 @@ assignable to `Object`.
 
 It is not an error for the body of a `late` field to reference `this`.
 
-It is an error for a formal parameter to be declared `late`.
+It is an error for a variable to be declared as `late` in any of the following
+positions: in a formal parameter list of any kind; in a catch clause; in the
```
sgrekhov: @leafpetersen where exactly a catch clause use of 'late' is an error?
```dart
} on Exception catch (late1 e) {
  late2 int i;
}
```
Does this sentence disallow 'late' in late1 or late2 positions?

eernstg: The catch clause actually doesn't allow for this syntax, it's `<identifier> (',' <identifier>)`, so `late1 e` is just a syntax error.

The position `late2` must be allowed, that's like any other block of code. We might want to omit the catch clause from the above sentence and just mention in commentary that a `<catchPart>` introduces a local variable or two, but doesn't syntactically allow for `late`.

PR #707, opened 2019-11-25
Clarify the definition of UP on function types

leafpetersen:


leafpetersen:
cc @scheglov

leafpetersen:
Landing this now.

PR #733, opened 2019-12-08
Integrate several small changes to the spec

leafpetersen:
- Allow elision of default value in abstract methods
  - **CHANGE** Allow operations on `Never` and specify the typing
  - Specify the type signature for calling Object methods on nullable types
  - Specify implicit conversion behavior
  - Allow potentially constant type variables in instance checks and casts
  - Specify the error thrown by the null check operator
  - Specify `fromEnvironment` and `Iterator.current` library breaking changes

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -6,6 +6,15 @@ Status: Draft
 
 ## CHANGELOG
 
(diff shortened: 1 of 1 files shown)


leafpetersen:
cc @scheglov @johnniwinther @stereotype441 Note that there are several small **changes** to the specified errors and warnings in this update.

PR #749, opened 2019-12-17
Clarify override checking

leafpetersen:
Eliminate incorrect language implying that opted in methods inherited through legacy classes preserved their opted-in types, but clarify that override checking is still done against the original declared type (whether inherited through legacy classes or not).


leafpetersen:
cc @scheglov @johnniwinther

lrhn:
LGTM!

PR #750, opened 2019-12-17
Specify additional compile time errors around late variables

leafpetersen:


leafpetersen:
cc @stereotype441 @scheglov @johnniwinther @stefantsov

PR #763, opened 2019-12-27
Flow analysis

leafpetersen:
Draft flow analysis specification.


leafpetersen:
cc @lrhn @eernstg @munificent @scheglov

resources/type-system/flow-analysis.md
```diff
@@ -0,0 +1,713 @@
...
+Promotion policy is defined by the following operations on flow models.
+
+We say that the **current type** of a variable `x` in variable model `VM` is `S` where:
+  - `VM = VariableModel(declared, promoted, tested, assigned, unassigned, captured)`
+  - `promoted = S::l` or (`promoted = []` and `declared = S`)
+
+Policy:
+  - We say that at type `T` is a type of interest for a variable `x` in a set of
+    tested types `tested` if `tested` contains a type `S` such that `T` is `S`,
+    or `T` is **NonNull(`S`)**.
```
stereotype441: The "or `T` is **NonNull(`S`)**" part of this is not currently implemented.  Tracking in https://github.com/dart-lang/sdk/issues/40011.

resources/type-system/flow-analysis.md
```diff
@@ -0,0 +1,713 @@
...
+
+  - We say that a variable `x` is promotable via type test with type `T` given
+    variable model `VM` if
+    - `VM = VariableModel(declared, promoted, tested, assigned, unassigned, captured)`
+    - and `captured` is false
+    - and `S` is the current type of `x` in `VM`
+    - and not `S <: T`
+    - and `T <: S` or (`S` is `X extends R` and `T <: R`) or (`S` is `X & R` and
```
stereotype441: I'm not sure if all the cases involving `extends` and `&` are properly implemented.  Perhaps you could help me craft some tests?

resources/type-system/flow-analysis.md
```diff
@@ -0,0 +1,713 @@
...
+    - and `S` is the current type of `x` in `VM`
+    - and `T <: S` and not `S <: T`
+    - and `T` is a type of interest for `x` in `tested`
+
+  - We say that a variable `x` is demotable via assignment of an expression of
+    type `T` given variable model `VM` if
+    - `VM = VariableModel(declared, promoted, tested, assigned, unassigned, captured)`
+    - and `captured` is false
+    - and declared::promoted contains a type `S` such that `T` is `S` or `T` is
+      **NonNull(`S`)**.
```
stereotype441: The "or `T` is **NonNull(`S`)**" part of this isn't implemented either.  Also tracking in https://github.com/dart-lang/sdk/issues/40011

resources/type-system/flow-analysis.md
```diff
@@ -0,0 +1,713 @@
...
+  - We say that a variable `x` is promotable via type test with type `T` given
+    variable model `VM` if
+    - `VM = VariableModel(declared, promoted, tested, assigned, unassigned, captured)`
+    - and `captured` is false
+    - and `S` is the current type of `x` in `VM`
+    - and not `S <: T`
+    - and `T <: S` or (`S` is `X extends R` and `T <: R`) or (`S` is `X & R` and
+      `T <: R`)
+
+  - We say that a variable `x` is promotable via initialization given variable
```
stereotype441: Note: not yet implemented.  See https://github.com/dart-lang/sdk/issues/39876.

resources/type-system/flow-analysis.md
```diff
@@ -0,0 +1,713 @@
...
+      - `VM = VariableModel(declared, promoted, tested, true, false, captured)`.
+    - otherwise if `x` is promotable via initialization given `VM` then
+      - `VM = VariableModel(declared, [T], tested, true, false, captured)`.
+    - otherwise if `x` is promotable via assignment of `E` given `VM`
+      - `VM = VariableModel(declared, T::promoted, tested, true, false, captured)`.
+    - otherwise if `x` is demotable via assignment of `E` given `VM`
+      - `VM = VariableModel(declared, demoted, tested, true, false, captured)`.
+      - where `previous` is the suffix of `promoted` starting with the first type
+        `S` such that `T <: S`, and:
+        - if `S`is nullable and if `T <: Q` where `Q` is **NonNull(`S`)** then
```
stereotype441: I believe the "NonNull" part of this is not yet implemented.  Tracking in https://github.com/dart-lang/sdk/issues/40011.

resources/type-system/flow-analysis.md
```diff
@@ -0,0 +1,713 @@
...
+  is defined as follows:
+  - If `E` is not a promotion target, then `M`
+  - If `E` is a promotion target `x`, then
+    - Let `VM = VariableModel(declared, promoted, tested, assigned, unassigned,
+      captured)` be the variable model for `x` in `VI`
+    - If `x` is not promotable via type test to `T` given `VM`, then return `M`
+    - Else
+      - Let `S` be the current type of `x` in `VM`
+      - If `T <: S` then let `T1` = `T`
+      - Else if `S` is `X extends R` then let `T1` = `X & T`
```
stereotype441: I'm not sure whether we handle these two cases correctly (with "extends" and "&").  Could you help me come up with test cases?

resources/type-system/flow-analysis.md
```diff
@@ -0,0 +1,713 @@
...
+      - Else If `S` is `X & R` then let `T1` = `X & T`
+      - Else `x` is not promotable (shouldn't happen since we checked above)
+      - Let `VM2 = VariableModel(declared, T1::promoted, T1::tested, assigned,
+      unassigned, captured)`
+      - Return `FlowModel(r, VI[x -> VM2])`
+- `promoteToNonNull(E, M)` where `E` is an expression and `M` is a flow model is
+  defined to be `promote(E, T, M)` where `T0` is the type of `E`, and `T` is
+  **NonNull(`T0`)**.
+- `factor(T, S)` where `T` and `S` are types defines the "remainder" of `T` when
```
stereotype441: The implementation currently doesn't have any of this logic.  I'm tracking in https://github.com/dart-lang/sdk/issues/40014, however I must admit that I'm questioning whether this feature is worth the implementation effort.  The only non-contrived example I can come up with where this is useful is that it would cause `if (x is! Null)` to promote `x` in the same way that `if (x != null)` would.  Do you have a more compelling example?

resources/type-system/flow-analysis.md
```diff
@@ -0,0 +1,713 @@
...
+  - Else if `T` is `FutureOr<R>` and `Future<R> <: S` then `factor(R, S)`
+  - Else if `T` is `FutureOr<R>` and `R <: S` then `factor(Future<R>, S)`
+
+
+Questions:
+ - The interaction between assignment based **promotion** and downwards inference is
+   probably managable.  I think doing downwards inference using the current
+   type, and then promoting the variable afterwards is fine for all reasonable
+   cases.
+ - The interaction between assignment based **demotion** and downwards inference is
```
stereotype441: I have very little confidence that we're currently doing this correctly.  Do you think you could help come up with some test cases?

resources/type-system/flow-analysis.md
```diff
@@ -0,0 +1,713 @@
...
+  - Let `after(N) = assign(x, E1, after(E1))`.
+  - Let `true(N) = assign(x, E1, true(E1))`.
+  - Let `false(N) = assign(x, E1, false(E1))`.
+  - Let `null(N) = assign(x, E1, null(E1))`.
+  - Let `notNull(N) = assign(x, E1, notNull(E1))`.
+
+- **operator==** If `N` is an expression of the form `E1 == E2` then:
+  - Let `before(E1) = before(N)`
+  - Let `before(E2) = after(E1)`
+  - Let `true(N) = join(join(null(E1), null(E2)),
```
stereotype441: "Join" doesn't do what you want here--it takes two FlowAnalysis states that represent divergent control flow paths and forms the FlowAnalysis state if they were to both branch to a common execution point.  That's not what's happening with `==`.  So, for example, in the expression `a == (b = 0)`, `null(E1)` hasn't seen the assignment to `b` yet, so `join(null(E1), null(E2))` produces a model in which `b` is not necessarily definitely assigned.

What's currently implemented is much simpler: if Ei is the literal expression `null`, and Ej is a read of a variable, then `false(N) = promoteToNonNull(Ej, after(Ej))`, otherwise `false(N) = after(E2)`.  In either case `true(N) = after(E2)`.

It's possible that we could achieve what you want with restrict, e.g. something like `true(N) = join(restrict(null(E1), null(E2), ...), restrict(notNull(E1), notNull(E2), ...))`.  But I'd have to think pretty hard about it, and I'm not really convinced it would buy us enough functionality to be worth it compared to what's already implemented.

resources/type-system/flow-analysis.md
```diff
@@ -0,0 +1,713 @@
...
+  (i.e. does not throw an exception or perform a jump such as `return`, `break`,
+  or `continue`).
+
+- `true(E)`, where `E` is an expression, represents the *flow model* just after
+  execution of `E`, assuming that `E` completes normally and evaluates to `true`.
+
+- `false(E)`, where `E` is an expression, represents the *flow model* just after
+  execution of `E`, assuming that `E` completes normally and evaluates to `false`.
+
+- `null(E)`, where `E` is an expression, represents the *flow model* just after
```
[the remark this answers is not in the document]
leafpetersen: It's hard to say without knowing what the special behaviors around null are.  A less contrived example of something that I think this enables is the following:

```dart
foo(int? x, int? y) {
  if (x != null) {
    if (x == y) {
       y.isEven; // (1)
    }
  }
}
```

The line at `(1)` is allowed in my model because the true model for the comparison contains the join of the cases where either `x` and `y` are both null (which is empty since `x` is known not to be null) or that they are both non-null.  So `y` gets promoted in the continuation of the comparison.

Is this worth it?  I don't really know.  I didn't specify it this way specifically to enable this example, but it seemed to be the uniform way to do it.

resources/type-system/flow-analysis.md
```diff
@@ -0,0 +1,713 @@
...
+  union of two variable models, defined as follows:
+  - If `VM1 = VariableModel(d1, p1, s1, a1, u1, c1)` and
+  - If `VM2 = VariableModel(d2, p2, s2, a2, u2, c2)` then
+  - `VM3 = VariableModel(d3, p3, s3, a3, u3, c3)` where
+   - `d3 = d1 = d2`
+     - Note that all models must agree on the declared type of a variable
+   - `p3 = p1 ^ p2`
+     - `p1` and `p2` are totally ordered subsets of a global partial order.
+  Their intersection is a subset of each, and as such is also totally ordered.
+   - `s3 = s1 U s2`
```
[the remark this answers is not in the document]
leafpetersen: sgtm.

resources/type-system/flow-analysis.md
```diff
@@ -0,0 +1,713 @@
...
+  a boolean indicating wether the variable is written in the finally block,
+  represents the composition of two variable models through a try/finally and is
+  defined as follows:
+  - If `VMB = VariableModel(d1, p1, s1, a1, u1, c1)` and
+  - If `VMF = VariableModel(d2, p2, s2, a2, u2, c2)` then
+  - `VM3 = VariableModel(d3, p3, s3, a3, u3, c3)` where
+   - `d3 = d1 = d2`
+     - Note that all models must agree on the declared type of a variable
+   - if `b` is true then `p3 = p2`
+   - if `b` is false then if the last entry in `p1` is a subtype of the last
```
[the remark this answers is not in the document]
leafpetersen: Can you elaborate on what you mean by "drop any entires from p2 that would violate the subtyping invariant"?  

The thing that I'm not sure about here is that elsewhere I assume that the list of promoted types are totally ordered.  I'm worried that this concatenation could violate that.  But maybe what you're saying is that we have `p3 = [...p1, for (var item in p2) if subtype(item, p1.last) then item]` ?  That should guarantee a total order, and does preserve more intermediate steps in principle.

resources/type-system/flow-analysis.md
```diff
@@ -0,0 +1,713 @@
...
+     entry of `p2`, then `p3 = p1` else `p3 = p2`.  If the variable is not
+     written to in the finally block, then it is valid to use any promotions
+     from the try block in any subsequent code (since if any subsequent code is
+     executed, the try block must have completed normally).  We only choose to
+     do so if the last entry is more precise.  (TODO: is this the right thing to
+     do here?).
+   - `s3 = s1 U s2`
```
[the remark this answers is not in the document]
leafpetersen: OK.  I don't really have a good intuition about the types of interest heuristic, so I'm fine following the implementation here.

resources/type-system/flow-analysis.md
```diff
@@ -0,0 +1,713 @@
...
+     - The set of test sites is the union of the test sites on either path.
+   - `a3 = a2`
```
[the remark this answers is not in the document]
leafpetersen: That seems right to me.

resources/type-system/flow-analysis.md
```diff
@@ -0,0 +1,713 @@
...
+       the join of the model from before the try block and after the try block,
+       and so assignments that may have occurred in the try block are already
+       modelled here).
+   - `u3 = u2`
+     - A variable is definitely unassigned if it is definitely unassigned in the
+       model of the finally block (note that the finally block is analyzed using
+       the join of the model from before the try block and after the try block,
+       and so the absence of any assignments that may have occurred in the try
+       block is already modelled here).
+   - `c3 = c2`
```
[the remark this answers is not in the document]
leafpetersen: Agreed.

resources/type-system/flow-analysis.md
leafpetersen: Sounds good.

resources/type-system/flow-analysis.md
leafpetersen: Yes.

resources/type-system/flow-analysis.md
leafpetersen: I'll put discussion in the issue.

resources/type-system/flow-analysis.md
leafpetersen: Yes.

resources/type-system/flow-analysis.md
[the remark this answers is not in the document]
leafpetersen: Seeing your comment later, I see that I have mis-specified operator ==, so I'll have to think further about the mechanism.  The example code still holds as something that I think I intended to work that I'm not sure works without tracking notNull.

resources/type-system/flow-analysis.md
leafpetersen: Ack.  I made a note of this, and we can iterate and follow up.

PR #765, opened 2019-12-27
Additional errors and warnings

leafpetersen:
- Update errors for switch statements.
  - Make it an error entirely to use the default `List` constructor in opted-in
    code.
  - Clarify that setter/getter assignability uses subtyping instead of
    assignability.


accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -251,8 +282,7 @@ initializer expression and is not marked `late` is used before it is definitely
 assigned (see Definite Assignment below).
 
 It is an error if a method, function, getter, or function expression with a
```
lrhn: -> ... if *the body of* a method  ...

(We define the concept for statements above, and methods are not statements).

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,8 +293,7 @@ It is an error if a required named parameter has a default value.
 It is an error if a named parameter that is part of a `required` group is not
 bound to an argument at a call site.
 
-It is an error to call the default `List` constructor with a length argument and
-a type argument which is potentially non-nullable.
+It is an error to call the default `List` constructor.
```
lrhn: Wohoo!

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -325,6 +354,30 @@ not a subtype of the class type associated with the class in which it is defined
 (specifically, it is an error to return a nullable type from a factory
 constructor for any class other than `Null`).
 
+It is an error if any case of a switch statement except the last case (the
+default case if present) **may complete normally**.  The previous syntactic
+restriction requiring the last statement of each case to be one of a delimited
```
lrhn: Is "delimited" correct here? (Non-native English speaker woes - I associate delimited to *having delimiters*, not just being *limited*, and I'm not sure what those delimiters would be here. Maybe I'm just too limited in my view.)

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -325,6 +354,30 @@ not a subtype of the class type associated with the class in which it is defined
...
+set of statements (break, continue, return, throw, or rethrow) is removed.
+
+Given a switch statement which switches over an expression `e` of type `T`,
+where the cases are dispatched based on expressions `e0...ek`:
```
lrhn: Nitpick: I'd put the `...` outside of the code quotes.
Maybe even go for `*e*<sub>0</sub>&hellip;*e*<sub>*k*</sub>` since the *e*s are metavariables, not source code: *e*<sub>0</sub>&hellip;*e*<sub>*k*</sub>.

Or maybe that's just overkill.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -325,6 +354,30 @@ not a subtype of the class type associated with the class in which it is defined
...
+  - It is no longer required that the `ei` be instances of the same class.
```
lrhn: -> ... that the values of the `ei` expressions are instances  ...

The ei are expressions, not values, and only values can be instances of classes.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -325,6 +354,30 @@ not a subtype of the class type associated with the class in which it is defined
...
+  - It is an error if the static type of any of the `ei` is not a subtype of
+    `T`.
```
lrhn: We rarely use the static type of constant expressions for anything. Use the runtime type instead.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -325,6 +354,30 @@ not a subtype of the class type associated with the class in which it is defined
...
+  - It is an error if any of the `ei` evaluate to constants for which equality
+    is not primitive.
+  - If `T` is an enum type, it is a warning if the switch does not handle all
+    enum cases, either explicitly or via a default.
+  - If `T` is `Q?` where `Q` is an enum type, it is a warning if the switch does
+    not handle all enum cases and `null`, either explicitly or via a default.
+  - It is a warning if a switch over a nullable type does not handle `null`
+    either explicitly or via a default.
+
+It is an error if a class as a setter and a getter with the same basename where
```
lrhn: ... as a setter ... -> ... has a setter ...

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -325,6 +354,30 @@ not a subtype of the class type associated with the class in which it is defined
...
+the return type of the setter is not a subtype of the argument type of the
```
lrhn: return type of the setter -> return type of the getter

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: done.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: It is a correct usage - derived I think from the sense of the set not being open ended, but instead having fixed boundaries.  But `enumerated` is probably clearer here, so I'll use that.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: done.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: done.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: Hmm.  It worries me slightly.   If it wasn't for the potential for additional breakage, I think I'd be mildly opposed on the grounds that it ties our hands later for no good reason.  If we do generalize to patterns later, patterns will mostly be driven off of static types.  But I can't immediately think of anything that goes too wrong with just treating *references* to constants as equivalent to the constants themselves, and just ignoring their static types.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: done.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: done.

accepted/future-releases/nnbd/feature-specification.md
lrhn: I guess either should work. The only case where it might cause trouble is if we have a const variable with a more general type than the value itself, which is likely incredibly rare, and the switch expression has a more specific type. I can't imagine a use-case where that happens.

leafpetersen:
cc @johnniwinther @stefantsov @scheglov @stereotype441 

Note that this PR lands **changes** to the specifed errors and warnings around switch statements.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -487,25 +537,28 @@ defined as follows.
...
-Note that we do not equate primitive top types.  `List<void>` and
-`List<dynamic>` are still considered distinct runtime type objects.  Note that
-we also do not equate `Never` and `Null`, and we do not equate function types
-which differ in the placement of `required` on parameter types.  Because of
-this, the equality described here is not equivalent to syntactic equality on the
-`LEGACY_ERASURE` of the types.
+objects) are considered equal if and only if the runtime type objects `T1` and
+`T2` corresponds to the types `S1` and `S2` respectively, and the normal forms
+**NORM(`S1`)** and **NORM(`S2`)** are syntactically equal up to equivalence of
+bound variables and **ignoring `*` modifiers on types**.  So for example, the
```
lrhn: and ordering of named parameters?

PR #779, opened 2020-01-14
Specify strong mode/weak mode flag.

leafpetersen:
Update the spec as follows:
  - Specify the weak/strong checking flag
  - Specify the weak/strong execution defaults based on the entry point
  - Specify that it is an error to compile or execute a mixed mode program with strong checking
  - Specify that errors are no longer demoted to warnings in weak checking mode


leafpetersen:
@mit-mit Does this match your understanding of the outcome of our discussions?

cc @lrhn @munificent @eernstg 

@a-siva @johnniwinther @sigmundch @vsmenon @liamappelbe @aam This is the desired weak mode/strong mode opt in flow, does this make sense for the implementations?

@davidmorgan @natebosch @jakemac53 are there issues with infrastructure here?  Once concern that I have is that I believe that for some of our test frameworks, `main` lives in test code and not in the user code.  Can we make this work correctly, in the sense that when the users test code is opted in they get opted into strong mode but otherwise run in weak mode?

vsmenon:
This sounds good to me.  Perhaps just the terms "weak null safety" and "null safety" to emphasize that "strong" is the default and final outcome?

jakemac53:
In general bootstrapping of apps is potentially a concern here, although the only use case I am aware of that still exists is package:test.

It bootstraps the app in an isolate with a script provided as a `data:` uri. This would presumably not pick up the default language version for the package and so we would have to use the language version comment or something to opt it in, but it should be doable.

For general codegen the files would be appear as a part of the root package, and thus pick up the default language version for that package based on its sdk constraint. Unless the language version is overridden in the `main` file this would likely just work as expected.

leafpetersen:
> It bootstraps the app in an isolate with a script provided as a `data:` uri. This would presumably not pick up the default language version for the package and so we would have to use the language version comment or something to opt it in, but it should be doable.

I believe that the VM will not support different isolates running in weak vs strong mode, so I think you'd need to know up front which way you want to run the VM.  Does that still seem reasonable?

jakemac53:
> I believe that the VM will not support different isolates running in weak vs strong mode, so I think you'd need to know up front which way you want to run the VM. Does that still seem reasonable?

This will likely be a larger problem. The test package is the main isolate, and it runs by spawning new isolates. The test binary itself is shipped on pub (its in the `bin` dir of the package) and would be opted in based on the test package settings and not the user package settings.

If as a result tests can only be ran in whatever mode the `test` package was published with then that would essentially break the promise of an incremental migration path. Or we would have to force users to go through separate top level test binaries which would break existing workflows and be a general pain.

sigmundch:
I like the current proposal, and I personally like the flag name `--runtime-null-safety`. I also learned recently that this affects the CFE and the semantics of constants. Given the static nature of constants, is it still accurate to call it "runtime-null-safety"?

johnniwinther:
I don't think it works to base the mode on the `main` file. 
  1) The mode defines the encoding of constants, which in turn requires the choice between a `*_platform_weak.dill` and `*_platform_strong.dill` for compilation. This choice has to be done very early in the compiltation; probably before the main file has been found and read.
  2) The mode defines whether to emit errors or warnings. This requires the main file to be found and read before any such message is emitted. This might be case with the currently specified errors but it puts a further strain on the compilation pipeline and might break subtly.
  3) A `main` file doesn't always exist or isn't known to exist. This is the case for platform compilation and modular compilation.
  4) Changes in the `main` file can make the mode change in the middle of an incremental compilation session. This is a new strain on the incremental compilation pipeline.

jakemac53:
Good points @johnniwinther - note that the modular compiles that we do also use the incremental compiler and so we would have this issue even when the `main` file isn't changed to be opted in/out.

I am concerned how this will work for modular compilation in general unless all apps agree on a mode which isn't really possible in google3?

jakemac53:
Talked a bit with @leafpetersen this AM. We should be able support both modes with modular compilation by treating each mode as a separate "platform". So essentially each library would be compiled twice (but only if it was depended on by two apps in different modes).

We would likely end up wanting to run separate workers as well per mode, or possibly we could keep two incremental compilers around in the same worker (one for weak mode and one for strong mode).

I am quite confident this could work, but it will involve potential duplication of work.

cc @davidmorgan for thoughts on that.

leafpetersen:
Thinking about the discussion of this with @jakemac53 , I'm wondering whether the upshot is that this becomes something that we do in the build systems rather than in the compiler.  That is, the compiler just understands the flags turning on/off runtime null safety, and presumably defaults to off.  But in so far as possible, the build and run systems should opt applications in once the user entry point becomes opted in.  So for internal code, once the build target for a test of app entry point gets opted in, we'd want to switch to runtime null safety mode.  And maybe something similar for pub and flutter run?  Not sure how feasible that is.

We can live without this auto-opt in functionality, but I'm fairly convinced that the more we can automatically opt people into runtime null safety as early as possible, the more confident we can be that the ecosystem will quickly converge on a correct migration.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -545,6 +549,38 @@ per
...
+
+**Strong checking** mode respects the nullability of types at runtime in casts
+and instance checks, as defined below.  Unmigrated programs or programs
+consisting of a mix of migrated and unmigrated code may fail at runtime with new
+cast failures due to the nullability annotations in runtime type objects.
+**This mode is unsound** if any part of the program is still unmigrated.  **This
+mode is sound** if all parts of the program have been migrated.
+
+Weak vs strong runtime checking can be controlled at runtime via the
+`--[no-]runtime-null-safety` flag, where the negated version of the flag implies
```
mit-mit: Given these flags are passed to compilers, which are all about running code, I would prefer to simplify the terms by omitting the `runtime-` part. I.e.:

* --null-safety: "strong checking"
* --no-null-safety: "weak checking"

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -545,6 +549,38 @@ per
 
 ## Runtime semantics
 
+### Weak and strong semantics
+
+To allow the null safety feature to be rolled out incrementally, we define two
+modes of execution.
+
+**Weak checking** mode largely ignores the nullability of types at runtime, as
```
mit-mit: What checks do you actually get with weak checking? If there are no observable checks from a developer's viewpoint, then calling this 'weak checking' seems a bit confusing.

leafpetersen: Do you prefer just using "weak mode" and "strong mode"?  I can make a pass to use that uniformly.  The "checking" here is because it is primarily instance checks and casts that are affected by the semantics: in weak mode, instance checks and casts are "weak" in the sense that they ignore nullability, whereas in strong mode they are "strong" in the sense that they properly account for nullability.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: Done.

accepted/future-releases/nnbd/feature-specification.md
mit-mit: I'm OK with 'weak checking' then.

I still really dislike any use of 'strong' as I think it's confusing given how prominently we used that term during Dart 2. I would much prefer 'weak checking' and 'full checking'.

jakemac53: Why not keep the wording consistent with the flag?

```
To allow the null safety feature to be rolled out incrementally, we define a new mode of execution: `null-safety` mode. To enable this mode, you can pass the `--[no]null-safety` flag.

**No Null Safety** mode ...

**Null Safety** mode ...

```

sigmundch: My concern with "no null safety" is that I can't discern that from Dart 2. Naming is hard! - another idea "Lax null safety" vs "enforced null safety"

jakemac53: What about `Legacy` mode vs `Null Safety` mode?

mit-mit: @sigmundch, why is discerning this from Dart 2 in the flag important? Isn't the point of `--no-null-safety` exactly to make it "seem like" it's still Dart 2?

@jakemac53 I don't like the term "legacy". It has a dual meaning of "from the past" and "obsolete", and the latter meaning is a problematic association for code that developers have invested much into.

sigmundch: I just would like to not give the illusion that this is a feature that can be turned off. We need to effectively communicate that `--no-null-safety` will only remove the enforcement of null safety checks, but will keep around the new syntax, the actual checks, and any warnings they may produce in weak-mode.

sigmundch: In light of the recent discussions of disallowing the combined mixed-mode + strong-mode, I wonder if terminology derived from the state of libraries (mixed vs not) could more clearly convey what we mean instead of using the weak/strong terms. For example:

* **mixed null-safety mode**: libraries can start using null safety, but null-safety is weakly enforced, that is, null values can flow into non-nullable typed variables.
* **pure null-safety mode**: all libraries migrated + strong null-safety enforcement

Thoughts?

mit-mit: I like that. I've in other contexts been using the terms 'partial null safety' and 'full null safety'

mit-mit:
@leafpetersen can we go ahead and merge this?

leafpetersen:
@mit-mit PTAL.

leafpetersen:
cc @a-siva @alexmarkov @crelier @rakudrama @johnniwinther

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1480,17 +1517,12 @@ variable `X`, it is statically decidable which of the first two clauses apply in
...
-
-Note that in a program with no opted out libraries, the first clause can never
-apply.
+type `S`, an instance check `e is T` occurring in an **opted-in library** is
+evaluated as follows:
+  - Return `NNBD_SUBTYPE(S, T)`
 
-Note also that except in the case that `T` is of the form `X` or `X*` for some
-type variable `X`, it is statically decidable which clause applies.
+Note that it is an error to run a program containing opted-in libraries in
```
eernstg, line 1524: That would be `containing legacy libraries`.

leafpetersen: Yes it would, wouldn't it... :) Fixed thanks!

PR #805, opened 2020-01-30
Batched up spec updates

leafpetersen:
This batches up a number of clarifications and changes that have accumulated.  

cc @lrhn @munificent @eernstg @scheglov @stefantsov @liamappelbe @crelier @nshahan @fishythefish @sigmundch @johnniwinther

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -448,8 +448,8 @@ type variables appearing in contravariant positions instead of `Null`.
 
 The definitions of least and greatest closure are changed in null safe libraries
 to substitute `Never` in positions where previously `Null` would have been
-substituted, and `Object?` in positions where previously `Object` would have
-been substituted.
+substituted, and `Object?` in positions where previously `Object` or `dynamic`
+would have been substituted.
 
 ### Const type variable elimination
 
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -6,6 +6,23 @@ Status: Draft
 
 ## CHANGELOG
 
+2020.01.29
+  - **CHANGE** Relax the exhaustiveness check on switches.
+  - Specify the type of throw expressions.
+  - Specify the override inference exception for operator==.
+  - **CHANGE** Specify that instantiate to bounds uses `Never` instead of `Null.
+  - **CHANGE** Specify that least and greatest closure uses `Never` instead of
+    `Null.
+  - Specify that type variable elimination is performed on constants using least
+    closure.
+  - Clarify extension method resolution on nullable types.
+  - **CHANGE** Add missing cases to `NNBD_TOP_MERGE` and specify it's behavior
+    on `covariant` parameters.
+  - Fix the definition of `NORM` for un-promoted type variables
+  - Change the notion of type equality for generic function bounds to mutual
+    subtyping.
+  - **CHANGE** Specify that debug assertions are added to methods in strong
+    mode.
 
 2020.01.27
   - **CHANGE** Change to specification of weak and strong mode instance checks
@@ -380,8 +397,6 @@ where the cases are dispatched based on expressions `e0`...`ek`:
     enum cases, either explicitly or via a default.
   - If `T` is `Q?` where `Q` is an enum type, it is a warning if the switch does
     not handle all enum cases and `null`, either explicitly or via a default.
-  - It is a warning if a switch over a nullable type does not handle `null`
-    either explicitly or via a default.
 
 It is an error if a class has a setter and a getter with the same basename where
 the return type of the getter is not a subtype of the argument type of the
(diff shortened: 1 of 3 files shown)


accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -412,6 +427,54 @@ Tearing off a method from a receiver of static type `Never` produces a value of
 type `Never`.  Applying an expression of type `Never` in the function position
 of a function call produces a result of type `Never`.
 
+The static type of a `throw e` expression is `Never`.
+
+In legacy mode, an override of operator== with no explicit parameter type
```
leafpetersen: @johnniwinther @scheglov

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -412,6 +427,54 @@ Tearing off a method from a receiver of static type `Never` produces a value of
...
+type.  Otherwise, the parameter type of the overriding method is dynamic.
+
+Top level variable and local function inference is performed
+as
+[specified separately](https://github.com/dart-lang/language/blob/master/resources/type-system/inference.md).
+Method body inference is not yet specified.
+
+### Instantiate to bounds
+
+The computation of instantiation to bounds is changed to substitute `Never` for
```
leafpetersen: @eernstg Is this what we want here?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -412,6 +427,54 @@ Tearing off a method from a receiver of static type `Never` produces a value of
...
+type variables appearing in contravariant positions instead of `Null`.
+
+### Least and greatest closure
+
+The definitions of least and greatest closure are changed in null safe libraries
```
leafpetersen: cc @scheglov @stefantsov note this change

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -990,6 +1053,10 @@ as:
    - And the reverse
  - `NNBD_TOP_MERGE(Object?, dynamic)  = Object?`
    - And the reverse
+ - `NNBD_TOP_MERGE(Object*, void)  = void`
```
leafpetersen: @stefantsov Note this fix.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1206,3 +1283,13 @@ understand these changes: for example, by providing a debugging option in which
 instance checks or casts which would result in a different outcome if run in
 strong checking mode vs weak checking mode are flagged for the developer by
 logging a warning or breaking to the debugger.
+
+### Automatic debug assertion insertion
```
leafpetersen: @johnniwinther @stefantsov This change is lower priority, but something that I'd like to do if we have time before the release.

resources/type-system/subtyping.md
```diff
@@ -92,11 +92,7 @@ capture-avoiding substitution of types `T0, ..., Tk` for the type variables
 ## Type equality
 
 We say that a type `T0` is equal to another type `T1` (written `T0 === T1`) if
-the two types are structurally equal up to renaming of bound type variables,
-and equating all top types.
-
-TODO: make these rules explicit.
-
+`T0 <: T1` and `T1 <: T0`.
```
leafpetersen: cc @stefantsov @johnniwinther @fishythefish @nshahan @liamappelbe @crelier @scheglov This change makes the subtyping rules for generic methods more forgiving.  It is specifically intended to allow a generic method with a legacy bound to override one with a nullable or non-nullable bound.  It also has the effect of allowing overrides that use different variants of top types for bounds. It also accounts for ad hoc normalization (e.g. `T Function<T extends int?>()` is specified to be a subtype of `T Function<T extends int??>()` by this change, and hence this validates the normalization of `int??` to `int?`).

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -6,6 +6,23 @@ Status: Draft
 
 ## CHANGELOG
 
+2020.01.29
+  - **CHANGE** Relax the exhaustiveness check on switches.
+  - Specify the type of throw expressions.
+  - Specify the override inference exception for operator==.
+  - **CHANGE** Specify that instantiate to bounds uses `Never` instead of `Null.
```
kevmoo: missed trailing "tick" after `Null` on this line!

leafpetersen: Fixed, thanks!

resources/type-system/normalization.md
```diff
@@ -80,7 +80,7 @@ equations apply gives us something like the following:
 - **NORM**(`X extends T`) =
   - let `S` be **NORM**(`T`)
   - if `S` is `Never` then `Never`
-  - else `X extends S`
+  - else `X extends T`
```
fishythefish: Why this change? Presumably, `X extends int??` (for example) would be normalized to `X extends int?`, but IIUC this change is undoing that.

resources/type-system/subtyping.md
fishythefish: Is this change gated behind anything, or can we just update the (legacy) subtyping algorithm?

resources/type-system/normalization.md
leafpetersen: Note the difference between normalizing a *use* of a type variable and the *declaration* of a type variable.  This case covers type variable uses. It's not clear to me what it means to normalize the bound on a use.  It's fine to do so, if a representation is chosen in which each use has a copy of the bound, but the bound does not contribute to syntactic equality.   The bound does contribute to normalization, since `X extends Never` normalizes to `Never`.  Does that make sense?

resources/type-system/subtyping.md
leafpetersen: Let's just update in place.

resources/type-system/normalization.md
fishythefish: Sort of. I'm reading the various expressions in this document as (abstract) type terms, and in my mental model, these aren't intrinsically equipped with a notion of being declared vs. used. (In a Dart program, type terms certainly exist in declaration-contexts and use-contexts, but I'm interpreting normalization as a context-free operation.)

Maybe I missed a definition somewhere, but I'm not really sure how to read `X extends T` as a standalone type term. (That is, I recognize it as a *declaration* in the context of a type parameter list, but I'm not sure how to interpret an `extends` clause without that context the way I can interpret `FutureOr<X>`, for example.) In this case, I took it as shorthand for the following: I have a type variable `X` and when I query my type environment's oracle/function/mechanism for looking up bounds, it tells me that **bound**(`X`) = `T`.

leafpetersen: > In this case, I took it as shorthand for the following: I have a type variable `X` and when I query my type environment's oracle/function/mechanism for looking up bounds, it tells me that **bound**(`X`) = `T`.

Functionally, that is the correct interpretation.  The analyzer's representation doesn't really work that way.  Instead, they attach a bound to every type variable, and there's essentially a coherence property that you enforce.  But it's morally the same thing - it's just a question of whether you keep a map from type variables to bounds, or just enforce that all type variables bound at the same location have the same bound attached to them.  I've just gotten in the habit of using the `X extends T` notation instead of talking about "the bound associated with T in the environment" since that also matches up better with the implementations.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -752,10 +815,9 @@ implicitly induced by _D_ if a value has previously been assigned to `v`
 (which could be due to an initializing formal or a constructor initializer
 list, or due to an invocation of the setter).
 
-Let _D_ be a `late` and `final` local variable declaration named `v`
-without an initializing expression.  It is a run-time error, throwing an
-instance of `LateInitializationError`, to assign a value to `v` if a value
-has previously been assigned to `v`.
+Let _D_ be a `late` and `final` local variable declaration named `v`.  It is a
```
leafpetersen: cc @johnniwinther

leafpetersen:
Landing this for now, I can iterate based on feedback in followups.

PR #811, opened 2020-01-31
Mixins: No have uninitialized non-nullable fields.

lrhn:
Also remove the reference to `CastError`. The language specification doesn't mention `CastError` any more, and there is no need to reintroduce it here.

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -8,7 +8,7 @@ Status: Draft
 
 2020.01.31
   - Specify that mixins must not have uninitialized potentially non-nullable 
-    non-late fields.
+    non-late fields, and nor must classes with no generative constructors.
   - Remove reference to `CastError`. A failed `!` check is just a
     "dynamic type error" like the `as` check in the current language specification.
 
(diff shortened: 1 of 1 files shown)


lrhn:
@stefantsov FYI.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -298,10 +304,21 @@ fields on `Object`.
...
-potentially non-nullable type has no initializer expression and is not
-initialized in a constructor via an initializing formal or an initializer list
-entry, unless the variable or field is marked with the `late` modifier.
+It is an error if a top level variable or static variable with a non-nullable type 
+has no initializer expression unless the variable is marked with the `late` modifier.
+
+It is an error if a class declaration declares an instance variable with a potentially 
+non-nullable type and no initializer expression, and the class has a generative constructor 
+where the variable is not initialized via an initializing formal or an initializer list entry, 
+unless the variable is marked with the `late` modifier.
```
eernstg: It is a tricky corner case that a class `C` with no generative constructors does not allow for creation of instances of `C`, nor for declaration of subclasses of `C`, and it escapes the above rule. (It can have useful static methods, and it can be a non-superclass superinterface, and it _will_ have a factory constructor which must return one of those subtypes, so maybe someone may want to write such a thing.)

You could say that it is then no problem to have an instance variable with a potentially non-nullable type and no initializer, because there cannot ever exist an instance of `C` or any subclass thereof, so it's no problem that it has a "bad" instance variable.

Nevertheless, I'd prefer that we make this an error, because it is a confusing setup, and not a useful one (as far as I can see):

```suggestion
unless the variable is marked with the `late` modifier.
It is an error if a class declaration declares an instance variable with a potentially 
non-nullable type and no initializer expression, and the class has no generative constructors.
```

If anyone really wants that API they can declare an abstract getter and setter (or one of those newfangled abstract fields ;-) and make the class abstract, and they can then preserve the rest of the setup without changes.

lrhn: Made the class-with-no-generative-constructor and non-late-potentially-nullable-uninitialized-instance-variable combination an error too.

PR #817, opened 2020-02-05
Clarify the notion of being noSuchMethod forwarded

eernstg:
The current language specification text about a noSuchMethod forwarded member seems to imply that `m` is _not_ nsm forwarded when an implementation named `m` is inherited. As is implied by the example a few lines later, this was never the intention: We must have an implementation named `m` which is a correct override of the member signature named `m` in the interface of the class, and if we don't have that then `m` is nsm forwarded.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -2404,16 +2404,18 @@ one of the following is true:
 \end{itemize}
 
 \LMHash{}%
+For a concrete class $C$, a
+\IndexCustom{noSuchMethod forwarder}{noSuchMethod forwarder}
+is implicitly induced for each member $m$ which is noSuchMethod forwarded.
+
+\LMHash{}%
 It is a compile-time error if a concrete class $C$ has
 an unimplemented member named $m$ with signature $S$,
 and a superclass of $C$ has an accessible concrete declaration of $m$
-which is not a \code{noSuchMethod} forwarder.
+which is not a noSuchMethod forwarder.
 
 \LMHash{}%
-For a concrete class $C$, a
-\IndexCustom{\code{noSuchMethod} forwarder}{noSuchMethod forwarder}
-is implicitly induced for each member $m$ which is noSuchMethod forwarded.
-This is a concrete member of $C$
+A noSuchMethod forwarder is a concrete member of $C$
 with the signature taken from the interface of $C$ respectively $D$ above,
 and with the same default value for each optional parameter.
 It can be invoked in an ordinary invocation and in a superinvocation,
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -31,6 +31,9 @@
 %   'class declaration' everywhere, so `<classDefinition>` is inconsistent).
 % - Clarify that <libraryDeclaration> and <partDeclaration> are the
 %   start symbols of the grammar.
+% - Clarify the notion of being 'noSuchMethod forwarded': `m` is indeed
+%   noSuchMethod forwarded if an implementation of `m` is inherited, but
+%   it does not have the required signature.
 %
 % 2.6
 % - Specify static analysis of a "callable object" invocation (where
@@ -2380,9 +2383,12 @@ one of the following is true:
 
 \begin{itemize}
 \item $C$ has a non-trivial \code{noSuchMethod},
-  the interface of $C$ contains $m$,
-  and $C$ has no concrete declaration of $m$
-  (\commentary{that is, no member $m$ is declared or inherited by $C$}).
+  the interface of $C$ contains a member signature $s$ named $m$,
+  and $C$ has no concrete declaration named $m$ that correctly overrides $s$
+  (\commentary{%
+    that is, no member named $m$ is declared or inherited by $C$,
+    or one is inherited, but it does not have the required signature%
+  }).
 \item
   % Inaccessible private methods are not present in the interface of a class,
   % so we need to find a class that can access $m$.
@@ -2450,8 +2456,8 @@ because compilers control the treatment of private names.
 
 \LMHash{}%
 It is a compile-time error if a concrete class $C$ has
-a \code{noSuchMethod} forwarded method signature $S$
-for a method named $m$,
+a \code{noSuchMethod} forwarded member signature $S$
+for a member named $m$,
 and a superclass of $C$ has an accessible concrete declaration of $m$
 which is not a \code{noSuchMethod} forwarder.
 
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -2392,6 +2392,7 @@ one of the following is true:
     that is, no member named $m$ is declared or inherited by $C$,
     or one is inherited, but it does not have the required signature%
   }).
+  In this case we also say that $S$ is noSuchMethod forwarded.
 \item
   \textbf{Forced by privacy:}
   There exists a direct or indirect superinterface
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -2379,28 +2382,40 @@ \subsubsection{The Method \code{noSuchMethod}}
...
+\item \textbf{Requested in program:}
+  $C$ has a non-trivial \code{noSuchMethod},
+  the interface of $C$ contains a member signature $S$ named $m$,
+  and $C$ has no concrete member named $m$ that correctly overrides $S$
+  (\commentary{%
+    that is, no member named $m$ is declared or inherited by $C$,
+    or one is inherited, but it does not have the required signature%
+  }).
 \item
   % Inaccessible private methods are not present in the interface of a class,
```
lrhn: Remove this line. They are.

specification/dartLangSpec.tex
```diff
@@ -2379,28 +2382,40 @@ \subsubsection{The Method \code{noSuchMethod}}
 one of the following is true:
 
 \begin{itemize}
-\item $C$ has a non-trivial \code{noSuchMethod},
-  the interface of $C$ contains $m$,
-  and $C$ has no concrete declaration of $m$
-  (\commentary{that is, no member $m$ is declared or inherited by $C$}).
+\item \textbf{Requested in program:}
+  $C$ has a non-trivial \code{noSuchMethod},
+  the interface of $C$ contains a member signature $S$ named $m$,
```
lrhn: "named $m$": Above $m$ was defined as a *member*, not a name.
So ... "with the same name as $m$"?

Also, we have to define the "NSM forwarded" on *interface members" (signatures), not on class members, because the class does not have that member. That's why we are adding it.

So:

For a concrete class $C$ with and an interface member $m$ of the interface of $C$, $m$ is **NSM-forwarded** if:
* $C$ has a non-trivial NSM,
   $C$ does not have a concrete member which is a valid 
   implementation of $m$, and
   the class $C$ has no declaration of a concrete member implementing $m$,
   and nor does any superclass of $C$.
* or $m$ is declared in a library $L2$ and $C$ is declared in library $L1$ and 
   $m$ is not accessible in $L1$, 
   and neither $C$ nor any superclass of $C$ declares a concrete member
   overriding $m$ (*you can only override something which is accessible 
   at the point of the overriding declaration*).

specification/dartLangSpec.tex
eernstg: Indeed, done.

specification/dartLangSpec.tex
```diff
@@ -2379,28 +2382,40 @@ \subsubsection{The Method \code{noSuchMethod}}
...
   % so we need to find a class that can access $m$.
+  \textbf{Forced by privacy:}
   There exists a direct or indirect superinterface
   $D$ of $C$ which is declared in the library $L_2$,
   the interface of $D$ contains $m$
   (\commentary{which implies that $m$ is accessible to $L_2$}),
   $m$ is inaccessible to $L$,
   and no superclass of $C$ has
-  a concrete declaration of $m$ accessible to $L_2$.
+  a concrete member of $m$ accessible to $L_2$.
```
lrhn: ... 
This doesn't work with "member of $m$" instead of "declaration of $m$".

I would say:
> No superclass of $C$ declares a concrete member implementing $m$,

It's still not perfect.


More fundamentally, I guess: What is a "member"?

As I understand it:

A class *declaration* contains member *declarations*, and these can be concrete or abstract. A class declaration induces (at least) a class and an interface (modulo generics as usual).

A class contains members induced by the declarations in the class declaration or inherited from the superclass, or introduced implicitly in other ways like NSM-forwarding (or coming from mixins in some cases, or ... etc).

An interface contains member signatures. Each signature with a private name is also associated with a library and it is only "accessible" inside that same library. There can be more than one signature with the same name, but only if they are private to different libraries.

Class members can implement interface member signatures with the same accessible name, or fail to do so if their function signature is not sufficient.

So, a NSM forwarder added to a class is a class *member* (concrete, but that's a pleonasm), and here we say that such a forwarder would prevent a NSM forwarder in a subclass.
```dart
/// library L2
class C {
   noSuchMethod(i) {}
   // Gets noSuchMethod forwarder member for `foo` interface signature.
   int _foo();
}
abstract class D extends C {
  int _foo([int x]);
}
/// Library L1
class E extends D {
  // Should get noSuchMethod-forwarder for D._foo.
}
```

So, this item does not capture that case unless it does not consider a NSM forwarder as a concrete member (which would probably be an inconsistent point of view). (Or unless it's rewritten as above, so only *declared* concrete members matter).

specification/dartLangSpec.tex
eernstg: Agreed, we need to consider a member signature in the interface of the class, then the relation to the members of the class (there may or may not be one with the specified name).

I've revised the text, PTAL.

eernstg: Note that the notion of being NSM-forwarded is now a property of a name (which is sufficient, along with accessibility for a specific library if the name is private, to uniquely determine the member signature).

specification/dartLangSpec.tex
eernstg: Right, I agree that a declaration is a syntactic entity. It is associated with a static entity which is semantic in the sense that it may have a signature or a type or other properties that are not purely syntactic (that is, not context free, we must do some analysis before we have this data). It is also associated with a dynamic entity (such as a class at run time, which allows us to talk about the value of a type variable at run time, or the value of a static variable, etc).

I also agree that a class interface contains member signatures, and they are specified explicitly already.

I have now defined 'NSM-forwarded in a class $C$ declared in a library $L$' to be a property of a name $m$.

The example you specify was actually omitted, and I have already changed the text accordingly. So we agree about that, too. ;-)

specification/dartLangSpec.tex
```diff
@@ -2373,34 +2376,49 @@ \subsubsection{The Method \code{noSuchMethod}}
...
+
+\LMHash{}%
+It is a compile-time error if the name $m$ is noSuchMethod forwarded
+in a concrete class $C$,
+and a superclass of $C$ has an accessible concrete declaration of $m$
+which is not a noSuchMethod forwarder.
+
+\LMHash{}%
+A noSuchMethod forwarder is a concrete member of $C$
 with the signature taken from the interface of $C$ respectively $D$ above,
```
lrhn: Do we need to say "respectively $D$" because the signature is also in the interface of $C$.
Also, since there might be more than on $D$ satisfying the requirement above for $m$ being privacy-nSM forwarded, we cannot speak about a particular $D$ here.

specification/dartLangSpec.tex
```diff
@@ -2502,6 +2476,45 @@ \subsubsection{The Method \code{noSuchMethod}}
 and hence there is no error in that situation.
 }
 
+\commentary{
+This implies that a \code{noSuchMethod} forwarder has the same
+properties as an explicitly declared concrete member,
+except of course that a \code{noSuchMethod} forwarder
+does not prevent itself from being induced.
```
lrhn: itself -> itself or another nSM forwarder 
?

specification/dartLangSpec.tex
eernstg: Agreed offline.

specification/dartLangSpec.tex
eernstg: Agreed offline.

specification/dartLangSpec.tex
eernstg: Changed to avoid mentioning $D$. Added TODO comment saying that we should change the approach to privacy in general such that we are closer to name-mangling, such that we can actually say precisely what we want to say here.

specification/dartLangSpec.tex

PR #846, opened 2020-02-19
Specify that 'await voidExpression' is allowed

eernstg:
We decided dart-lang/sdk#33415 that `await e` cannot be a compile-time error when the static type of `e` is `void` (that would have been nice, but it was too breaking). This PR updates the language specification to specify this.


lrhn:
LGTM

lrhn:
Any chance we could *change* this for NNBD? We're breaking things anyway ... :)

eernstg:
That would indeed be nice! But breaking more and more things also makes the migration harder, so we do need to justify every bit of it.

PR #854, opened 2020-02-24
Use single quotes throughout the spread spec

charlesboyd:
Replace an extra double quotes with single quotes in the spread collections feature spec so that they match the other quotes.


munificent:
Thanks!

PR #856, opened 2020-02-26
Make yield* non-compositional

eernstg:
PR #830 changed the language specification to use a compositional approach to `yield*` (that is, it allowed things like `yield* <num>[1]` in a synchronous generator returning `Iterable<int>`, and it would then be checked per element that each of them could be yielded).

However, this conflicts with an earlier decision (and with the currently implemented behavior). Cf. https://github.com/dart-lang/sdk/issues/40538#issuecomment-589399667, https://github.com/dart-lang/sdk/issues/40538#issuecomment-589885343.

This PR changes the specification to require that the static type of `e` in `yield* e` is assignable to the declared return type of the enclosing generator function, and also that the dynamic type of the `yield*`'ed object is a subtype of the declared return type.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -15768,8 +15768,6 @@ or it is not a generator.
 Let $T_f$ be the element type of $f$
 (\ref{functions}),
 and let $T$ be the static type of $e$.
-It is a compile-time error if $T$ may not be assigned to
-the declared return type of $f$.
 If $f$ is a synchronous generator,
 it is a compile-time error if $T$ may not be assigned to
 \code{Iterable<$T_f$>}.
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -38,6 +38,8 @@
 %   assignability is enforced per element.
 % - Update whitelist for expressions of type void to include `await e`,
 %   consistent with decision in SDK issue #33415.
+% - Re-adjust `yield*` rules: Per-element type checks are not supported,
+%   the given Iterable/Stream must have a safe element type.
 %
 % 2.6
 % - Specify static analysis of a "callable object" invocation (where
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -15776,8 +15776,7 @@ and it is a compile-time error if $T$ may not be assigned to
 \code{Stream<$T_f$>}.
 
 \LMHash{}%
-If $f$ is a generator and
-$T$ implements \code{Iterable<$U$>} or \code{Stream<$U$>} for some $U$
+If $T$ implements \code{Iterable<$U$>} or \code{Stream<$U$>} for some $U$
 (\ref{interfaceSuperinterfaces})
 then let $T_e$ be $U$.
 %% TODO(eernst): Come nnbd, change 'a top type or null' to \DYNAMIC{} or Never.
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -1386,13 +1386,6 @@ This implies that there is no information about
 the type of elements that the generator will yield.%
 }
 
-Let $f$ be with declared return type $T$.
-If $T$ implements \code{Iterable<$U$>} for some $U$
-(\ref{interfaceSuperinterfaces})
-then the element type of $f$ is $U$.
-
-
-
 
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -15773,6 +15768,8 @@ \subsection{Yield-Each}
 Let $T_f$ be the element type of $f$
 (\ref{functions}),
 and let $T$ be the static type of $e$.
+It is a compile-time error if $T$ may not be assigned to
+the declared return type of $f$.
 If $f$ is a generator and
```
lrhn: If it's not, `yield*` is a compile-time error, so maybe this "If" is not necessary.

specification/dartLangSpec.tex
```diff
@@ -15785,7 +15782,7 @@ \subsection{Yield-Each}
 It is a compile-time error if $T_e$ may not be assigned to $T_f$.
```
lrhn: $T_e$ has not been defined. Do you mean $U$?

specification/dartLangSpec.tex
```diff
@@ -15773,6 +15768,8 @@ \subsection{Yield-Each}
 Let $T_f$ be the element type of $f$
 (\ref{functions}),
 and let $T$ be the static type of $e$.
+It is a compile-time error if $T$ may not be assigned to
+the declared return type of $f$.
 If $f$ is a generator and
 $T$ implements \code{Iterable<$U$>} or \code{Stream<$U$>} for some $U$
 (\ref{interfaceSuperinterfaces})
```
lrhn: Also need to say that:

* if $f$ is a synchronous generator function, it's a compile-time error if $T$ is not assignable to \code{Iterable<$T_f$>}, and
* otherwise $f$ is a synchronous generator function, and it's a compile-time error if $T$ is not assignable to \code{Stream<$T_f$>}.

Otherwise you can statically get away with:
```dart
Object foo() sync* {
  yield* Stream.fromIterable([1, 2]);
}
```
without getting caught.

specification/dartLangSpec.tex
lrhn: Lower case "i" at the beginning? Maybe comma before it.

specification/dartLangSpec.tex
eernstg: The text here is intended to keep working when we get null safety: For instance, it will then prevent `yield* <num>[];` in a synchronous generator returning `Iterable<int>`. Without null safety, the assignability rule allows `yield* e;` where the static type of `e` implements `Iterable<U>` for any `U`, which is consistent with the current implementations.

So I believe this rule will work as intended now and later. Do you see some other location where this property is stated or can be derived? Otherwise this is simply the place where it's specified that it is a compile-time error.

specification/dartLangSpec.tex
eernstg: $T_e$ is defined in lines 15773-15779, and this is line 15785.

specification/dartLangSpec.tex
eernstg: Right! Done.

specification/dartLangSpec.tex
eernstg: Maybe this is a github UI glitch? The line `It is a compile-time error if $T_e$ may not be assigned to $T_f$.` occurs just after a line containing commentary that ends in `.`.

eernstg:
PTAL.

specification/dartLangSpec.tex
lrhn: Those lines define $T_f$, not $T_e$.

lrhn: True, there are lines missing from the view which cannot be expanded.

specification/dartLangSpec.tex
lrhn: It's only the `$f$ is a generator and` part that I'm questioning. If it wasn't a generator function, the `yield*` would already be a compile-time error according to line 15765

specification/dartLangSpec.tex
lrhn: (Unless GitHub is choosing to not show me the correct lines of the correct file. Which it is).

specification/dartLangSpec.tex
eernstg: Right, removed.

PR #861, opened 2020-02-28
Introduced static rules into section Asynchronous For-in

eernstg:
The language specification used to be near-silent on the static analysis of an asynchronous for-in statement. This PR adds rules about that topic.

Edit, Mar 03:

We need to achieve the outcome that a function literal capturing the iteration variable can be evaluated during any or all iterations of the loop body, which means that the iteration variable must be in a fresh scope (considered as a run-time entity) for each execution of the body. The previous text in this section did not specify this requirement at all.

With a synchronous for-in statement we rely on having a local variable in a block statement, which would then ensure that every execution of the body may capture a variable in a fresh run-time scope.

However, this only works because the specification allows us to assume that a local variable declared in a block will be a fresh run-time entity (section 'Blocks' only mentions scopes, no run-time properties). So we are already assuming that it is required that an implementation of Dart must ensure that a captured variable which is declared as a local variable in a block is allocated afresh for each execution of the block. In particular, it is _not_ allowed for an implementation to use the same storage location in the current activation record for state declared in a block that is executed more than once (of course, that's still allowed as an optimization, and a crucial one, whenever there is no observable difference).

This PR requires that the run-time scope is 'fresh' with each execution of the loop body, and supports the intention by commentary. I believe that this improves the level of precision and explicitness that we currently have in locations where the same property is required, and also that it has a sufficiently unambiguous semantics to be tolerable. So I'd recommend that we use this approach for section 'Blocks' as well.

One extra benefit is that it is very concise, and we may use the same approach, e.g., with `\ref{forLoop}`.

Alternatively, we could also extend the implicit reliance on the above mentioned property of blocks, and specify a desugaring step which will introduce an extra level of blocks, adding a local variable declaration (or just an assignment when `$D$` is empty). However, if we do this then we will have to desugar each asynchronous for-in statement _exactly once_ (because the desugaring step produces an output which matches the desugaring rule again), which is a new kind of constraint, or we need to introduce an `await while` statement, or in some other way ensure that this desugaring step will terminate.

We must of course still be able to desugar nested or enclosing loops, and null-shorting expressions in the iterable, and collection literals with `<element>`s, etc, so we will need to specify that this particular desugaring transformation can only be applied once to a given piece of code `$s$`, and then we need to avoid doing it again on any piece of code `$s_1$` which was produced by taking some other desugaring step on `$s$`. Sounds like we'll have to talk about abstract syntax tree nodes with identity in order to be able to say "don't do this twice".

Moreover, we still need to introduce exactly the notion of being a fresh scope at run time that we need for these for-in loops, in order to specify the dynamic semantics of blocks. So it seems reasonable to use it twice, rather than using it once and then introducing a complex desugaring rule just so we avoid using it a second time.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -14793,6 +14793,12 @@ but $e$ evaluates to an instance of a type
 which is not a subtype of \code{Iterable<dynamic>}.
 }
 
+\LMHash{}%
+The scope introduced by \code{\{\,$S$\,\}} around $S$
+is known as the
+\IndexCustom{scope of the for-in statement $s$}{%
+  scope!for-in statement}.
+
 
 \subsubsection{Asynchronous For-in}
 \LMLabel{asynchronousFor-in}
@@ -14800,12 +14806,69 @@ which is not a subtype of \code{Iterable<dynamic>}.
 \LMHash{}%
 A for-in statement may be asynchronous.
 The asynchronous form is designed to iterate over streams.
-An asynchronous for loop is distinguished by the keyword \AWAIT{} immediately preceding the keyword \FOR.
+An asynchronous for loop is distinguished by
+the keyword \AWAIT{} immediately preceding the keyword \FOR.
+
+\LMHash{}%
+It is a compile-time error if an asynchronous for-in statement appears
+inside a synchronous function (\ref{functions}).
+It is a compile-time error if a traditional for loop
+(\ref{forLoop})
+is prefixed by the \AWAIT{} keyword.
+
+\rationale{%
+An asynchronous loop would make no sense within a synchronous function,
+for the same reasons that an await expression makes no sense in
+a synchronous function.%
+}
+
+\LMHash{}%
+Let $D$ be derived from \syntax{<finalConstVarOrType>?}\ and
+let $s$ be an asynchronous for-in statement of the form
+`\code{\AWAIT\,\,\FOR\,\,($D$\,\,\id\,\,\IN\,\,$e$)\,\,$s_b$}'.
+Let $T_e$ be the static type of $e$.
+A compile-time error occurs unless $T_e$ may be assigned to
+\code{Stream<\DYNAMIC>}.
+%% TODO(eernst): Come nnbd, `a top type` --> `\DYNAMIC{}`.
+If $T_e$ is a top type then let $T_{\metavar{elm}}$ be \DYNAMIC.
+Otherwise there exists a type $U$
+such that $T_e$ implements \code{Stream<$U$>},
+in which case $T_{\metavar{elm}}$ is $U$.
+
+\LMHash{}%
+If $D$ is empty:
+Assume that $v$ is a fresh variable with declared type $T_{\metavar{elm}}$.
+%% Several cases here: \id{} could denote a local variable, an instance or
+%% static setter, nothing, or a lot of things causing an error.
+A compile-time error occurs if replacing $s$ by
+the statement `\code{\id\,\,=\,\,$v$;}'
+would give rise to a compile-time error.
+
+\LMHash{}%
+Otherwise, $D$ is not empty:
+Let $V$ be the declared type of \id.
+\commentary{%
+Type inference is assumed to have taken place already
+(\ref{overview}),
+so $V$ may be present explicitly or it may have been inferred.%
+}
+A compile-time error occurs unless $T_{\metavar{elm}}$ may be assigned to $V$.
+
+\LMHash{}%
+In either case
+(\commentary{with an empty or non-empty $D$}),
+a fresh scope, the
+\IndexCustom{scope of the asynchronous for-in statement $s$}{%
+  scope!asynchronous for-in statement},
+is introduced which has the enclosing scope of $s$ as its enclosing scope,
+and which is the current scope for the body $s_b$.
+If $D$ is non-empty it introduces a local variable into
+the scope of $s$ with the name \id{} and the declared type $V$.
 
 \LMHash{}%
 Let $D$ be derived from \syntax{<finalConstVarOrType>?}.
-Execution of a for-in statement, $f$, of the form
-\code{\AWAIT{} \FOR{} ($D$ \id{} \IN{} $e$) $s$}
+Execution of a for-in statement, $s$, of the form
+`\code{\AWAIT\,\,\FOR\,\,($D$\,\,\id\,\,\IN\,\,$e$)\,\,$s_b$}'
 proceeds as follows:
 
 \LMHash{}%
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -14793,71 +14794,151 @@ \subsubsection{For-in}
 which is not a subtype of \code{Iterable<dynamic>}.
 }
 
+\LMHash{}%
+The scope introduced by \code{\{\,$S$\,\}} around $S$
+is known as the
+\IndexCustom{scope of the for-in statement $s$}{%
```
lrhn: Should this  be `$S$` ?

specification/dartLangSpec.tex
```diff
@@ -14793,71 +14794,151 @@ \subsubsection{For-in}
...
+in which case $T_{\metavar{elm}}$ is $U$.
+
+\LMHash{}%
+If $D$ is empty:
+Assume that $v$ is a fresh variable with declared type $T_{\metavar{elm}}$.
+%% Several cases here: \id{} could denote a local variable, an instance or
+%% static setter, nothing, or a lot of things causing an error.
+A compile-time error occurs if replacing $s$ by
+the statement `\code{\id\,\,=\,\,$v$;}'
+would give rise to a compile-time error.
```
lrhn: Too vague. 

For one thing, you at least have to say *which kind of variable* `v` is. Local would be most reasonable. Then you have to say that it must be definitely assigned. 
Then we have to ensure that removing/replacing the loop statement doesn't introduce an error. For example, if the variable `\id` is defined as `late final int id;` and the loop is followed by an `id = 42;` then replacing the loop which *potentially* initializes `id` with an assignment which *definitely* assigns `id` will make the later assignment a compile-time error. That code is possible because the loop might break past the later assignment in the first iteration. Consider the function:
```dart
T first<T>(Stream<T> stream, T defaultValue) {
  late final T result;
  search: {
    await for (result in stream) break search;
    result = defaultValue;
 }
 return result;
}
```
This is "perfectly good" (perhaps a little over-engineered) code. The clause above would disallow it.

So, no removing or replacing code unless it's to something definitely semantically equivalent.
Maybe:

> ```latex
> Replace $S$ with:
> \code{\AWAIT{} \FOR{} ($T_{\metavar{elm}}$ $v$ \IN{} $e$) \{ 
>     \id = $v$; \{ $s_b$; \} 
> \}}
> where $v$ is a fresh variable. It is a compile time error if this would introduce a new compile-time error.
> ```

?
This rewrite should work whether `D` is empty or not.

specification/dartLangSpec.tex
```diff
@@ -14793,71 +14794,151 @@ \subsubsection{For-in}
...
+\commentary{%
+The \code{pause} call can throw,
+although that should never happen for a correctly implemented stream.%
 }
 
 \LMHash{}%
 For each \Index{data event} from $u$,
-the statement $s$ is executed with \id{} bound to the value of the current data event.
+the statement $s_b$ is executed with \id{} bound to
+the value of the current data event.
```
lrhn: That's actually imprecise, going on wrong. There is a *new* variable declared for each iteration, and we need to say that. Somehow.

So, it's more like we evaluate `$D$ \id = $v$` where `$v$` is a fresh variable bound to the value of the current data event in a scope surrounding the one used for `$s_b$`.

So, in general, I guess we can rewrite `await for (D id in e) s` to `await for (tmp in e) { D id = tmp; { s }}`, where `tmp` is a fresh local variable of type `$T_elm$`.

specification/dartLangSpec.tex
```diff
@@ -14866,19 +14947,12 @@ \subsubsection{Asynchronous For-in}
...
+with $e$ as exception object and $st$ as stack trace.
 
 \LMHash{}%
-It is a compile-time error if an asynchronous for-in statement appears inside a synchronous function (\ref{functions}).
-It is a compile-time error if a traditional for loop (\ref{forLoop}) is prefixed by the \AWAIT{} keyword.
-
-\rationale{
-An asynchronous loop would make no sense within a synchronous function, for the same reasons that an await expression makes no sense in a synchronous function.
-}
+When $u$ is done, execution of $s_b$ and $s$ completes normally.
```
lrhn: Just *s*. During execution of *s_b*, there is no way *u* can be done because it just *synchronously* sent a data event, and either *s_b* completes synchronously, or it pauses *u* which prevents it from being done (can't send a "done" event).

specification/dartLangSpec.tex
```diff
@@ -14866,19 +14947,12 @@ \subsubsection{Asynchronous For-in}
...
 where $v$ is a fresh variable referencing the stream subscription $u$.
 If that evaluation throws,
-execution of $f$ throws the same exception object and stack trace.
-Otherwise execution of $f$ throws with $e$ as exception object and $st$ as stack trace.
-
-\LMHash{}%
-When $u$ is done, execution of $f$ completes normally.
+execution of $s_b$ throws the same exception object and stack trace.
+Otherwise execution of $s_b$ throws
+with $e$ as exception object and $st$ as stack trace.
```
lrhn: I can't find the error-event paragraph. Is it just hidden, or should it be here:

> ```latex
> \LMHash{}%
> If $u$ emits an error event with error object \metavar{error} and associated stack trace \metavar{stack},
> then $u$ is cancelled and completed in the same way as if $s_b$ had completed by throwing 
> \metavar{error} and \metavar{stack}.

specification/dartLangSpec.tex
lrhn: Maybe change `is done` to `emits a done event`.

specification/dartLangSpec.tex
eernstg: We use `$s$` to denote a statement in several situations where it gets a name, and never `$S$` (that's mostly a type). So it wouldn't improve consistency to call it `$S$` here.

specification/dartLangSpec.tex
eernstg: We use a `fresh variable` in a similar way in 36 locations, and we define that concept with the added information that it is assumed to be local.

We do not currently include null safety specific information (such as 'this variable is definitely assigned'). I added a TODO comment on the introduction of the concept, such that we will get that part as part of the concept as well.

About the statement replacement: We can fiddle with the rewriting template, but it doesn't seem useful to me to come up with a complex piece of synthetic code if a simpler piece of code would do.

So we might instead be able to say explicitly that the question is whether the  assignment statement itself is an error, and it is just irrelevant whether the actual replacement would create a program that has some errors else where. I tried to do that.

I don't see the need to require that this desugaring operation yields a semantically equivalent program, because the purpose of the desugaring is to give a compact and readable way to determine whether the original code has a compile-time error, there is no intention of running this code.

specification/dartLangSpec.tex
eernstg: Added after `$st$ as stack track` (not after `completes normally`).

specification/dartLangSpec.tex
eernstg: OK, done!

eernstg: Done, no pun intended.

specification/dartLangSpec.tex
eernstg: Now specifying in the dynamic semantics that each iteration uses a fresh variable. This differs from the approach used with a synchronous for-in because it's semantics rather than syntactic (desugaring), but we do introduce the notion of fresh variables in order to specify the semantics of a block statement and such, so it's not a new concept in the spec, just a difference that stands out because for-in is desugared and asynchronous for-in is not. So I believe we can defend both forms, and the discrepancy is a viable choice (even though it would have been a more homogeneous result if we could desugar both forms, or if we would write out the semantics of both forms and avoid the desugaring).

eernstg:
Revised again, now using 'fresh variable' in the dynamic semantics in order to specify that each iteration gets its own iteration variable. PTAL.

Issue #858, opened 2020-02-27
Clarify how SDK versions map to language versions


munificent:
The [language versioning spec](https://github.com/dart-lang/language/blob/master/accepted/future-releases/language-versioning/feature-specification.md) says a language version is inferred from a Dart SDK version in a couple of places:

* When Pub generates a package_config.json file, it uses the packages' SDK constraints to select language versions for each package.

* If a package has no SDK constraint or a library is not part of a package, the library's language version is based on "the current SDK version".

But the spec does not clarify how a three-segment semver version number is converted to a major.minor language version. It should pin that down. The behavior we want (which Pub conveniently already implements) is:

* The language version of an SDK semver version number is the major.minor components of the version number. So SDK versions `1.2.0`, `1.2.3`, `1.2.3+build.4`, and `1.2.3-dev.4` all yield language version `1.2`.

* The language version of a version constraint is the language version of the minimum end of lowest version range in the constraint. So constraints `1.2.3`, `1.2.3 1.4.5`, `>=1.2.3 <2.0.0`, `>1.2.3 <2.0.0`, `>=1.2.3+build.4 <2.0.0`, `>=1.2.3-dev.4 <2.0.0`, and `>=1.2.4 <1.5.0 >=1.6.2 < 2.0.0` are all `1.2`.

PR #863, opened 2020-03-02
Update language text to specify language versions.

lrhn:
Also make it explicit how experiments work and where Pub gets the default language version from.

See issue #858.

### DIFF
--- a/accepted/future-releases/language-versioning/feature-specification.md
+++ b/accepted/future-releases/language-versioning/feature-specification.md
@@ -1,7 +1,7 @@
 # Dart Language Versioning
 
 [lrn@google.com](@lrhn)<br>
-Version: 1.4<br>
+Version: 1.5<br>
 Status: Ready for implementation.
 
 ## Motivation
@@ -17,7 +17,7 @@ We will definitely invest in tools to migrate code from one version of the langu
 
 However, not all language changes will permit a clean automatable migration. For example, introducing non-nullable types requires _adding_ information that was not explicitly present in the original source: is a variable intended to be nullable or not? Static analysis may get us some of the way, but there will always be programs that cannot be analyzed precisely, but where a human programmer would easily see the _meaning_ of the code. So, automatic migration tools can at most be a help&mdash;they cannot be a full solution.
 
-This document specifies an approach to easing the migration of a breaking language change by allowing unmigrated and migrated code to both run in the same program, which allows gradual migration over time while ensuring that a valid program can still run at any stage of the migration.
+This document specifies the chosen approach to easing the migration of a breaking language change by allowing unmigrated and migrated code to both run in the same program, which allows gradual migration over time while ensuring that a valid program can still run at any stage of the migration.
 
 The end goal is still to have all code migrated to the newest language version. This feature is designed to encourage this migration by making migration the easiest path, while still allowing an author to keep some libraries back from being migrated for a while.
 The goal is not to allow code to stay unmigrated indefinitely, only to give authors *reasonable* time to perform the migration, and to migrated a large body of code gradually.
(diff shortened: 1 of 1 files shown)


accepted/future-releases/language-versioning/feature-specification.md
```diff
@@ -26,42 +26,67 @@ The goal of this feature is not to make future or experimental features availabl
...
+
+This feature cannot support language changes that are fundamentally incompatible with earlier language versions, we do need the run-time system to be able to contain both language versions at the same time. For the same reason, this feature does not handle *platform library* changes that are not backwards compatible. The libraries are shared between the different language version libraries in a program, and must work for all of these. (Our tools should still detect use of library features not supported by the minimum required SDK version for a package.)
+
+Versioning uses _language version number_ only. You cannot opt-in to one feature of a new version, but not the rest. You either migrate to that new version, or you do not. This allows us to keep the combinatorial complexity down to being (roughly) linear in the number of supported versions.
```
eernstg:
```suggestion
Versioning uses the language version number only. You cannot opt-in to one feature of a new version, but not the rest. You either migrate to that new version, or you do not. This allows us to keep the combinatorial complexity down to being (roughly) linear in the number of supported versions.
```

Again, removed italic property because this is not (no longer) the defining occurrence of that phrase.

accepted/future-releases/language-versioning/feature-specification.md
```diff
@@ -26,42 +26,67 @@ The goal of this feature is not to make future or experimental features availabl
...
+
+### Experimental Languages
+
+The actual language of the next Dart release is not known for certain until that Dart SDK is finally released. All new features currently being developed are experimental and subject to being changed, postponed, or withdrawn.
+
+During the development of a next Dart release, the SDK exists only as developer releases (for example, the Dart SDK version on bleeding edge is 2.8.0-dev.9.0). There is no Dart 2.8 language yet, but there is a family of experimental Dart 2.8 languages that users can experiment with.
```
eernstg:
```suggestion
During the development of a next Dart release, the SDK exists only as developer releases (for example, the Dart SDK version on bleeding edge is 2.8.0-dev.9.0). There is no Dart 2.8 language version yet, but there is a family of experimental Dart 2.8 languages that users can experiment with.
```

It makes sense to introduce those 'experimental Dart 2.8 languages', and it would be confusing to insist that they are 'versions' of the language. Keeping them separate enables us to always have an associated version number whenever we have a language version.

accepted/future-releases/language-versioning/feature-specification.md
```diff
@@ -26,42 +26,67 @@ The goal of this feature is not to make future or experimental features availabl
...
+
+We switch between these language versions using the `–enable-experiment` flag to the Dart compilers and other tools. The current *base* Dart 2.8 language, the one used with no experiments enabled, is the same as the 2.7 language. A different 2.8 language can be used by enabling one or more experiments. There are, at time of writing, four experiments in various degrees of completion:
```
eernstg:
```suggestion
We switch between these experimental languages using the `–enable-experiment` flag to the Dart compilers and other tools. The current *base* language for experiments under version 2.8, the one used with no experiments enabled, is the same as language version 2.7. A different 2.8 experimental language can be used by enabling one or more experiments. There are, at time of writing, four experiments in various degrees of completion:
```

accepted/future-releases/language-versioning/feature-specification.md
```diff
@@ -26,42 +26,67 @@ The goal of this feature is not to make future or experimental features availabl
...
+### Default Language Version
 
-The `.dart_tool/package_config.json` file is a JSON formatted text file which declares a number of named packages with a root and a package URI root directory for each. The SDK version for a specific package is added as an extra `languageVersion` property on the JSON object for the package.  The package configuration file is generated automatically by the `pub` tool as documented in the file’s specification.
+Each *package* in a program has a default language version associated with it, and each *library* in that package has that language version as its default language version. A library can override the default language version explicitly as described below.
 
-Example package entry:
+The package configuration file,  `.dart_tool/package_config.json`, file is a JSON formatted text file which defines the configuration for a number of named packages using entries of the form:
```
eernstg: Whitespace:

```suggestion
The package configuration file, `.dart_tool/package_config.json`, file is a JSON formatted text file which defines the configuration for a number of named packages using entries of the form:
```

accepted/future-releases/language-versioning/feature-specification.md
```diff
@@ -72,35 +97,33 @@ Example package entry:
...
 
-This means that all Dart libraries in, e.g., `test/` and `bin/` directories of a Pub package are considered as belonging to the package, not just the libraries accessed using `package:` URIs.
+Tools will not have a language default version available for any package until a `package_config.json` file has been generated by running `pub get`. Any tool able to understand Pub package layout *may* derive the information directly from the  `pubspec.yaml` of the current package, rather than rely on an entry in the `package_config.json` file, as long as it derives the same values that Pub would have written into the `package_config.json` file. A lack of a `package_config.json` file is only an issue for *new* packages, or packages with no dependencies, because otherwise the package is mostly unusable until you run `pub get` anyway. A new package is likely to want to be run at the most recent language version, so this is not a major problem. An IDE which reruns `pub get` after each change to `pubspec.yaml` will keep the `package_config.json` file up-to-date.
```
eernstg: Whitespace:

```suggestion
Tools will not have a language default version available for any package until a `package_config.json` file has been generated by running `pub get`. Any tool able to understand Pub package layout *may* derive the information directly from the `pubspec.yaml` of the current package, rather than rely on an entry in the `package_config.json` file, as long as it derives the same values that Pub would have written into the `package_config.json` file. A lack of a `package_config.json` file is only an issue for *new* packages, or packages with no dependencies, because otherwise the package is mostly unusable until you run `pub get` anyway. A new package is likely to want to be run at the most recent language version, so this is not a major problem. An IDE which reruns `pub get` after each change to `pubspec.yaml` will keep the `package_config.json` file up-to-date.
```

accepted/future-releases/language-versioning/feature-specification.md
```diff
@@ -284,4 +325,11 @@ The optimistic upgrade from version 2.5 to 2.9 will go through all the intermedi
 - Allow language version markers above the package default level (for generated code).
 - State that the version used from `pubspec.yaml` is the lower bound, whether inclusive or exclusive, and that the language version is the major/minor versions of that.
 
-1.4:  Update to use new `package_config.json` file.
\ No newline at end of file
+1.4:  Update to use new `package_config.json` file.
+
+1.5: Rewrite and elaboration:
+
+- Make it explicit how the Pub tool choses the package default language version
```
eernstg:
```suggestion
- Make it explicit how the Pub tool choses the package default language version.
```

eernstg (approved):
LGTM, suggesting that we're adjusted a couple of occurrences of 'language' to 'language version' and introduce (explicitly) the notion of a 'language version number'. The effect is that we will consistently talk about each of the "different Darts over time" as a 'language version' that has a 'language version number', and we don't talk about them as "different languages".

Then when we get to `--enable-experiment=...` we switch to talk about 'experimental languages'. We could insist on saying 'version' in some way, but I don't think that's helpful. We avoid implying that there is a 'number' for an experimental language, and when it's explicitly introduced I think it's OK to talk about 'experimental languages' even though we avoid calling each 'language version' a 'language'.

mit-mit:
@lrhn can this be closed?

munificent:
I'm going to go ahead and consider this abandoned. :)

PR #864, opened 2020-03-02
Flow analysis fixes

leafpetersen:
Fixes to the flow analysis spec.


lrhn:
LGTM

scheglov:
Was this landed too fast? 
I expected that [my other concern](https://github.com/dart-lang/sdk/issues/40014#issuecomment-593676061), with using `factor()` and "tested" also will be clarified :-(

leafpetersen:
I can follow up on the remaining concerns with another PR.

Issue #865, opened 2020-03-03
Compile-time error when using assignment initializer with anonymous functions.


hugocastelani:
Seems like Dart analysis gets confused when using assignment initializer with anonymous functions. For instance:
```
class Foo {
  final VoidCallback bar;
  Foo() : bar = () {}; // Expected an identifier.
}
```

When wrapping the function in parenthesis, the error goes away:
```
class Foo {
  final VoidCallback bar;
  Foo() : bar = (() {}); // Fine.
}
```

I'm not sure whether this behavior is expected, but I cannot see any reason why it should be.

lrhn:
The problem here is that the parser has a hard time distinguishing function bodies in the initializer list from a constructor body.

It's not *impossible*, but it's hard enough that Dart has so far not allowed unparenthesized function-expressions in initializer lists. 

The example here is easier because there is no argument, but
```dart
 Foo(x) : bar = (x) { print(x); };
```
requires arbitrary look-ahead to see whether the `{print(x)}` is a function body or a constructor body. (It's probably a function body because of the trailing `;`).

I'm not sure the grammar reflects this restriction. @eernstg?

eernstg:
This restriction has been part of Dart since 2013 (cf. https://github.com/dart-lang/sdk/issues/11509), but the language specification does not mention it, and never did. [This PR](https://github.com/dart-lang/language/pull/866) adds it.

PR #866, opened 2020-03-03
Specify that initializerExpression cannot be a function literal

eernstg:
Cf. https://github.com/dart-lang/sdk/issues/11509, it has been a compile-time error since 2013 for an `<initializerExpression>` (that is, an expression that occurs to the right of `=` in an initializer list) to be a function literal. This PR finally puts that into the language specification.

Given that tools have rejected such function literals for several years, it is not a breaking change, and it requires no updates to implementations.

This resolves issue #865.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -3163,8 +3163,10 @@ There are three kinds of initializers.
 \end{grammar}
 
 \LMHash{}%
-If an \synt{initializerExpression} is a \synt{functionExpression},
-a compile-time error occurs.
+It is a compile-time error if a \synt{functionExpression} occurs
+as a subterm of an \synt{initializerExpression} $e$,
+unless it occurs inside a subterm of $e$ which is of the form
+\syntax{`(' <expression> `)'}.
 
 \rationale{%
 The reason for having this restriction is that
@@ -3188,10 +3190,7 @@ Assuming that this restriction did not exist, consider the following example:
 `\code{C():\,\,\THIS.y\,\,=\,(x)\,\{\ldots\}}'.
 If that snippet of code is followed by `\code{;}' or `\code{,}'
 then `\code{(x)\,\{\ldots\}}' is a function literal,
-otherwise it is the expression \code{(x)} followed by the constructor body.
-If needed, the restriction may be avoided
-by adding a parenthesis around the function literal,
-`\code{C():\,\,\THIS.y\,\,=\,((x)\,\{\ldots\})}'.%
+otherwise it is the expression \code{(x)} followed by the constructor body.%
 }
 
 \LMHash{}%
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -44,6 +44,8 @@
 %   `T` is a function type, plus other similar cases.
 % - Specify actual type arguments passed to a generic function which is invoked
 %   with no type arguments (so it must be a dynamic invocation).
+% - Specify that an <initializerExpression> cannot be a function literal. This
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -3154,11 +3158,44 @@ \subsubsection{Generative Constructors}
...
+`\code{C():\,\,\THIS.y\,\,=\,(x)\,\{\ldots\}}'.
+If that snippet of code is followed by `\code{;}' or `\code{,}'
+then `\code{(x)\,\{\ldots\}}' is a function literal,
+otherwise it is the expression \code{(x)} followed by the constructor body.
+If needed, the restriction may be avoided
+by adding a parenthesis around the function literal,
+`\code{C():\,\,\THIS.y\,\,=\,((x)\,\{\ldots\})}'.%
+}
+
 \LMHash{}%
```
lrhn: I assume the changes below are reflows only.

specification/dartLangSpec.tex
```diff
@@ -3154,11 +3158,44 @@ \subsubsection{Generative Constructors}
 <initializerExpression> ::= <conditionalExpression> | <cascade>
 \end{grammar}
 
+\LMHash{}%
+If an \synt{initializerExpression} is a \synt{functionExpression},
+a compile-time error occurs.
```
lrhn: That's sadly not enough.

We don't allow function expressions at "top level", meaning they also cannot occur as operator operands. They have to be inside parentheses to be allowed.
Examples that are not allowed:
```dart
class C {
  var x, y;
  C() : x = a + () {}, y = () {} + a;
}
```
which could otherwise be made valid by a suitable definition of `a` or an extension method, like:
```dart
extension on void Function() { 
  operator +(void Function() f) => () { this(); f(); }
}
void a() { print("a"); }
``` 
So, the description needs to be more elaborate.

Maybe: 
> It is a compile-time error if an initializerExpression contains a functionExpression which is not inside a parethesizedExpression.

eernstg: Right, done.

specification/dartLangSpec.tex
eernstg: That is true, I generally bring a unit of text (like a subsection) into a "80 chars at most" format when it is edited, trying to maintain the unity of meaning (such that grep will still be able to produce useful results).

eernstg:
Review responses uploaded, PTAL.

specification/dartLangSpec.tex
```diff
@@ -3154,11 +3158,44 @@ \subsubsection{Generative Constructors}
 <initializerExpression> ::= <conditionalExpression> | <cascade>
 \end{grammar}
 
+\LMHash{}%
+It is a compile-time error if a \synt{functionExpression} occurs as a subterm
+(\ref{notation})
+of an \synt{initializerExpression} $e$,
+unless it occurs inside a subterm of $e$ which is of the form
+\syntax{`(' <expression> `)'}.
```
lrhn: HOWEVER (ahem), that's probably *also* not precise enough, now that I got to think about it.

We do allow some function expressions that are not inside parentheses, as long as they are inside *something*.
Being "inside" can be inside a collection literal or string interpolation, or occurring as a function call argument, which are still clearly delimited, but not by a parenthesized expression.

So, maybe we need a definition like:

> A subexpression *e* of another expression *o* is said to be *undelimited* in *o* if:
> * *e* is *o*, or
> * *o* is an operator expression (infix, prefix or suffix operators) 
>     and *e* is undelimited in one of the operand expressions of *o*.

Then it is a compile-time error if a function expression is undelimited in the initializer expression.

PR #897, opened 2020-03-26
Adjust mixed mode rules

eernstg:
Complementing https://dart-review.googlesource.com/c/sdk/+/139817, this PR proposes a few adjustments to the nnbd spec in locations where the description of mixed-mode inheritance differs from recent discussions about the approach and the implementation.

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -6,6 +6,10 @@ Status: Draft
 
 ## CHANGELOG
 
+2020.03.26
+  - Adjust mixed-mode inheritance rules to express a consolidated model
+    where legacy types prevail in some additional cases.
+
 2020.03.05
   - Update grammar for null aware subscript.
   - Fix reversed subtype order in assignability.
@@ -1177,6 +1181,9 @@ class C extends B {
 }
 ```
 
+It is difficult to predict the outcome of migrating `B` in such situations, but
+lints or hints may be used by tools to communicate to developers that `C` may
+need to be changed again when `B` is migrated.
 
 If a class `C` in an opted-in library inherits a member `m` with the same name
 from multiple direct super-interfaces (whether legacy or opted-in), let `T0,
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -992,9 +992,12 @@ arguments) marked as legacy types.
 
 If a class `C` in a legacy library implements the same generic class `I` more
 than once, it is an error if the `LEGACY_ERASURE` of all such super-interfaces
-are not all syntactically equal.  For the purposes of runtime subtyping checks,
-`C` is considered to implement the canonical `LEGACY_ERASURE` of the
-super-interfaces in question.
+are not all syntactically equal.
+
+When `C` implements `I` once, and also when `C` implements `I` more than once
+without error, `C` is considered to implement the canonical `LEGACY_ERASURE` of
+`I`. This determines the outcome of dynamic instance checks applied to instances
+of `C`, as well as static subtype checks on expressions of type `C`.
 
 A member which is defined in a class in a legacy library (whether concrete or
 abstract), is given a signature in which every type is a legacy type.  It is an
@@ -1004,7 +1007,7 @@ rules.
 
 Using the legacy erasure for checking super-interfaces accounts for opted-out
 classes which depend on both opted-in and opted-out versions of the same generic
-interface.
+interface. For example:
 
 ```dart
 //opted in
@@ -1022,8 +1025,7 @@ class C extends A implements B {}
 
 The class `C` is not considered erroneous, despite implementing both `I<int?>`
 and `I<int*>`, since legacy erasure makes both of those interfaces equal.  The
-canonical interface which `C` is chosen to implement for the purposes of runtime
-type checks is `I<int*>`.
+interface which `C` is considered to implement is `I<int*>`.
 
 
 #### Classes defined in legacy libraries as seen from opted-in libraries
@@ -1137,16 +1139,21 @@ arguments) marked as nullable or non-nullable as written.
 If a class `C` in an opted-in library implements the same generic class `I` more
 than once as `I0, .., In`, and at least one of the `Ii` is not syntactically
 equal to the others, then it is an error if `NNBD_TOP_MERGE(S0, ..., Sn)` is not
-defined where `Si` is **NORM(`Ii`)**.  Otherwise, for the purposes of runtime
-subtyping checks, `C` is considered to implement the canonical interface given
-by `NNBD_TOP_MERGE(S0, ..., Sn)`.
+defined where `Si` is **NORM(`Ii`)**.
+
+If `C` implements `I` once, or `C` implements `I` multiple times without error,
+`C` is considered to implement the canonical interface given by
+`NNBD_TOP_MERGE(S0, ..., Sn)`. This determines the outcome of dynamic instance
+checks applied to instances of `C`, as well as static subtype checks on
+expressions of type `C`.
 
 If a class `C` in an opted-in library overrides a member, it is an error if its
 signature is not a subtype of the types of all overriden members from all
-super-interfaces (whether legacy or opted-in).  For the purposes of override
-checking, members which are inherited from opted-in classes through legacy
-classes are still checked against each original declaration at its opted-in
-type.  For example, the following override is considered an error.
+direct super-interfaces (whether legacy or opted-in).  This implies that
+override checks for a member `m` may succeed due to a legacy member signature
+for `m` in a direct super-interface, even in the case where an indirect
+super-interface has a member signature for `m` where the override would be a
+compile-time error. For example:
 
 ```dart
 // opted_in.dart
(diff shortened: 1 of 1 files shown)


accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1137,16 +1143,21 @@ arguments) marked as nullable or non-nullable as written.
 If a class `C` in an opted-in library implements the same generic class `I` more
 than once as `I0, .., In`, and at least one of the `Ii` is not syntactically
 equal to the others, then it is an error if `NNBD_TOP_MERGE(S0, ..., Sn)` is not
-defined where `Si` is **NORM(`Ii`)**.  Otherwise, for the purposes of runtime
-subtyping checks, `C` is considered to implement the canonical interface given
-by `NNBD_TOP_MERGE(S0, ..., Sn)`.
+defined where `Si` is **NORM(`Ii`)**.
+
+If `C` implements `I` once, or `C` implements `I` multiple times without error,
```
johnniwinther: This means that for
```Dart
  // opt-out
  class A implement Foo<FutureOr<Object>> {}
  // opt-in
  class B implements A {} 
```
`A` implements `Foo<FutureOr<Object>>` whereas `B` implements `Foo<Future<Object>>`. Is this intentional?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -992,9 +996,12 @@ arguments) marked as legacy types.
 
 If a class `C` in a legacy library implements the same generic class `I` more
 than once, it is an error if the `LEGACY_ERASURE` of all such super-interfaces
-are not all syntactically equal.  For the purposes of runtime subtyping checks,
-`C` is considered to implement the canonical `LEGACY_ERASURE` of the
-super-interfaces in question.
+are not all syntactically equal.
+
+When `C` implements `I` once, and also when `C` implements `I` more than once
+without error, `C` is considered to implement the canonical `LEGACY_ERASURE` of
```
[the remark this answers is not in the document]
leafpetersen: I think my intention was to say something to the effect of "is considered to implement the canonical signature given by `LEGACY_ERASURE` of the super-interfaces in question".

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: I'm not keen on this change.  My goal here was to only invent a new type if required.  Why this change?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -992,9 +996,12 @@ arguments) marked as legacy types.
 
 If a class `C` in a legacy library implements the same generic class `I` more
 than once, it is an error if the `LEGACY_ERASURE` of all such super-interfaces
-are not all syntactically equal.  For the purposes of runtime subtyping checks,
-`C` is considered to implement the canonical `LEGACY_ERASURE` of the
-super-interfaces in question.
+are not all syntactically equal.
+
+When `C` implements `I` once, and also when `C` implements `I` more than once
```
leafpetersen: So just to be clear, I think this means that if a legacy class `B` has `A<int?>` as a super-interface, then in an opted in library, an assignment of the form `A<int> x = B()` is not considered an error?

accepted/future-releases/nnbd/feature-specification.md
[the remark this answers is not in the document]
eernstg: Using that wording.

accepted/future-releases/nnbd/feature-specification.md
eernstg: For the example, normalization would reduce `Foo<FutureOr<Object*>*>` to `Foo<Object*>`, which eliminates `FutureOr`, and because `Object**` normalizes to `Object*`. So `A` implements `Foo<Object*>`.

And `B` implements `Foo<Object*>` because that's the `LEGACY_ERASURE` of `Foo<Object*>` (actually I think that erasure is `Foo<Object*>*`, but I don't think we want a rule that makes implemented types legacy at the top level, so we may need a variant of `LEGACY_ERASURE` that doesn't add a `*` at the top level, or maybe `LEGACY_ERASURE` just shouldn't add `*` at the top level at all).

But, @leafpetersen, I understand that you view this rule as particularly aggressive in its introduction of legacy typing. I used that wording in order to obtain a property that @johnniwinther mentioned a while back: A legacy class is purely legacy.

There may be different trade-offs here ("keep null-safe types whenever possible", in particular when `I` is only implemented once, versus "make sure that all properties of a legacy class are expressed in terms of legacy types").

@johnniwinther, WDYT?

accepted/future-releases/nnbd/feature-specification.md
eernstg: Right.

This is a loss in static type safety. As I mentioned, it was motivated by the rule that a legacy class should only have legacy properties.

We can of course offer additional help to developers in their quest to detect discrepancies like this (for instance, a lint or a hint could flag such situations as dangerous).

However, `B` could change the actual member signatures and I do not believe that we'd have a rule justifying that `B` would then implement `I` in a different way:

```dart
// Opted in.
abstract class A<X> {
  X get x;
}

abstract class A1 extends A<int> {}

// Opted out.
class B implements A1 {
  int get x => null;
}

// Opted in.
class C implements B {}

void main() {
  A<int> a = C(); // OK!
}
```

So here we trust `C` to implement `A<int>`, even though the overriding implementation of `x` in `B` implies that `x` may actually return null (which means that it is reasonable to say that `C` implements `A<int*>`, but misleading to say that it implements `A<int>`).

So even though it looks like we are preserving more static type information, I suspect that it may not be sufficiently reliable to be very helpful.

accepted/future-releases/nnbd/feature-specification.md
johnniwinther: For me difference is in whether we should perform normalization or not. If there is only one type we don't need to merge and therefore just use the type itself. In this case that'll make `B` implement `Foo<FutureOr<Object*>*>`. If there are more types we need to perform `NNBD_TOP_MERGE` and in this case we perform `NORM` on the input:
```Dart
// opt-out
class A implements Foo<FutureOr<Object>> {}
// opt-in
class B implements Foo<FutureOr<dynamic>> {}
class C implements A, B {}
```
Here `C` implements 
```
NNBD_TOP_MERGE(NORM(Foo<FutureOr<Object*>*>),NORM(Foo<FutureOr<dynamic>>))
  = NNBD_TOP_MERGE(Foo<Object*>,Foo<dynamic>)
  = Foo<Object?>
```

Note that I don't consider the outmost declaration in extends, implements and with clauses to have nullability so `Foo` is just `Foo` regardless of opt-in/out status.

leafpetersen: @eernstg I don't understand your answer.  The issue that @johnniwinther was pointing out and that I was responding to has nothing to do with legacy, as far as I can tell.  My original spec only applied NNBD_TOP_MERGE when there were multiple super-interfaces with the same generic base class.  You add the the line `If `C` implements `I` once,`.  This line is the one that I question.  It implies we normalize all super-interfaces, which is a much higher impact change.

eernstg: The "when `C` implements `I` once" change should only be applied when `C` is legacy, it was a mistake to have it here. Sorry about that.

eernstg: However, I would expect an implementation to be allowed to normalize _any_ term derived from `<type>` as eagerly as it wishes. Some of these terms may have syntax like `FutureOr<X>` where `X` is a type variable, and this may obviously call for normalization at run time, but with `FutureOr<Object?>` I would expect it to be OK for a compiler to normalize it at compile-time.

@johnniwinther, do you prefer to avoid compile-time type normalization except in the case where it is used to resolve a conflict among implemented superinterfaces or inherited member signatures? I guess the point would be that normalization which is eager to some unknown/unspecified degree would create problems in the management of the values of constant expressions? (Some constant expressions suddenly evaluate to the same object, even though they look different).

I think that issue might not arise with the normalization that is applied during the two cases of conflict resolution that are in focus here, because canonicalization would only ever identify instances of the same class, and if we have two instances of the given class `C` then it makes no difference that they implement a normalized version of some superinterface, because that will be true for both of them.

leafpetersen:
Once we settle on the tests, can you file issues for the analyzer and CFE to implement this change?

Also, consider adding an entry to the CHANGELOG at the top?

accepted/future-releases/nnbd/feature-specification.md
johnniwinther: I prefer to only normalize when we need to synthesize a result (e.g. in case of implementing to instantiations of a generic interface).

eernstg: I think there are no further issues here: Normalizing when a synthesized result is needed is necessary, and if we wish to introduce normalization elsewhere then it would be a non-breaking change. So I'll resolve this thread.

eernstg:
Thanks! I updated the CHANGELOG entry (this is line 9 of feature-specification.md, I assume that's what we are talking about). I'll file analyzer/cfe implementation issues when the [test CL](https://dart-review.googlesource.com/c/sdk/+/139817) lands. I see that it is fully LGTM'd, so I'll go ahead and pre-approve the test failures and land it.

@scheglov, you haven't indicated your approval of this PR yet; if you do have further comments, please speak up!

scheglov:
@eernstg LGTM

eernstg:
Thanks!

Issue #896, opened 2020-03-26
Can we call an expression whose type is a type parameter?


eernstg:
Cf. https://github.com/dart-lang/sdk/issues/41218. A compile-time error is [specified](https://github.com/dart-lang/language/blob/35ce3873ec46480b318020828c0d9553e6e595ac/specification/dartLangSpec.tex#L10377) in the language specification for the invocation of an expression of any type `T` which is not `dynamic`, not `Function`, and not a function type. In particular, when `X` is a type parameter whose bound is a function type, it is an error to invoke an expression of type `X`. Example:

```dart
// Situation reported in #41218
class C<X, Y extends void Function(X)> {
  final Y f;
  C(this.f);
  void m(X value) => f(value); // Error in some configurations, but not in most.
}

// Rewrite.
class D<X> {
  final void Function(X) f; // No error, but highly questionable use of `X`.
  D(this.f);
  void m(X value) => f(value); // Unsafe (but very common) use of `X`.
}

// Improved rewrite, come declaration-site variance support.
class E<in X> {
  final void Function(X) f; // Safe.
  E(this.f);
  void m(X value) => f(value); // Safe.
}
```

This error turns out to be reported in dartpad.dev, but not on the command line with `dart`, `dartanalyzer`, nor `dart2js` (version 2.8.0-dev.17.0).

Conceptually, we could say that it is a natural expectation in a language with subtyping to allow every operation on an entity of static type `S` whenever the same operation is allowed on a non-`dynamic` type `T` such that `S <: T`.

However that makes sense for objects and member accesses, because a subtype would generally mean "possibly some additional members, and at least the same, with compatible types". But with a function type `F` there are no other subtypes than `Never` (with null-safety, and `Null` as of today) and other function types, so "additional members" does not count as a justification.

Being able to specify a proper subtype of the given bound in a subtype of class `C` would offer additional affordances:

```dart
class C1<X> implements C<X, Y extends void Function(Object?)> {
  final Y f;
  C(this.f);
  void m(X value) => f(42); // OK, we know that `f` accepts any argument.
}
```

So should we stick with the strict reading of the current specification and get this error implemented everywhere, or should we allow invoking an expression whose type is a type parameter whose bound is a function type? Is there a killer example which will justify that those additional affordances really matter?

Here's a good example why it might be worthwhile to allow it: https://github.com/dart-lang/sdk/issues/41218#issuecomment-604347478.

@munificent, @leafpetersen, @lrhn, WDYT?

[...]

PR #900, opened 2020-03-27
Define T bounded types, to allow f() where f has type X extends Function

eernstg:
Cf. https://github.com/dart-lang/language/issues/896.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -10392,18 +10392,13 @@ Similarly for a
 \IndexCustom{\FUNCTION{} bounded type}{type!function bounded}.
 
 \LMHash{}%
-Similarly, we introduce the notion of a
-\IndexCustom{function-type bounded type}{type!function-type bounded}:
-Every function type $T$
-(\ref{functionTypes})
-is a function-type bounded type,
-with associated function type $T$;
-if $B$ is function-type bounded with associated function type $T$ and
-$X$ is a type variable with bound $B$
-then $X$ is function-type bounded with associated function type $T$;
-finally, if $B$ is function-type bounded with associated function type $T$ and
-$X$ is a type variable
-then $X \& B$ is function-type bounded with associated function type $T$.
+A
+\IndexCustom{function-type bounded type}{type!function-type bounded}
+is a type $T$ which is $T_0$ bounded where $T_0$ is a function type
+(\ref{functionTypes}).
+A function-type bounded type $T$ has an
+\Index{associated function type}
+which is the unique function type $T_0$ such that $T$ is $T_0$ bounded.
 
 \LMHash{}%
 If the static type of $f$ is \DYNAMIC{} bounded or \FUNCTION{} bounded,
@@ -10414,8 +10409,9 @@ Otherwise, it is a compile-time error if the static type of $f$ is not
 function-type bounded.
 
 \LMHash{}%
-Otherwise, the static type of $f$ is function-type bounded
-with some associated function type $F$.
+If no error occurred and the static analysis of $i$ is not complete
+then the static type $T_f$ of $f$ is function-type bounded;
+let $F$ be the associated function type of $T_f$.
 
 \LMHash{}%
 Let $S_0$ be the return type of $F$,
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -4158,7 +4158,7 @@ from one of its superinterfaces
 (\ref{covariantParameters}).
 }
 
-/LMHash{}%
+\LMHash{}%
 For the purpose of performing static checks on ordinary method invocations
 (\ref{ordinaryInvocation})
 and property extractions
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -40,6 +40,8 @@
 %   consistent with decision in SDK issue #33415.
 % - Re-adjust `yield*` rules: Per-element type checks are not supported,
 %   the given Iterable/Stream must have a safe element type.
+% - Clarify that an expression of type `X extends T` can be invoked when
+%   `T` is a function type, plus other similar cases.
 %
 % 2.6
 % - Specify static analysis of a "callable object" invocation (where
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -10372,12 +10374,50 @@ \subsubsection{Binding Actuals to Formals}
...
+finally, if $B$ is $T_0$ bounded and
+$X$ is a type variable
+then $X \& B$ is $T_0$ bounded.
+In particular, a
+\IndexCustom{\DYNAMIC{} bounded type}{type!dynamic bounded}
+is either \DYNAMIC{} itself
+or a type variable whose bound is \DYNAMIC{} bounded,
+or an intersection whose second operand is \DYNAMIC{} bounded.
+Similarly for a
+\IndexCustom{\FUNCTION{} bounded type}{type!function bounded}.
```
lrhn: Do we need `Never` bounded types, or does normalization collapse them all to `Never` anyway?
(Are we not handling `Never` in the spec yet?)

specification/dartLangSpec.tex
```diff
@@ -10372,12 +10374,50 @@ \subsubsection{Binding Actuals to Formals}
...
+
+\LMHash{}%
+Similarly, we introduce the notion of a
+\IndexCustom{function-type bounded type}{type!function-type bounded}:
```
lrhn: How does being "function-type bounded with associated function type T" differ from being a "T bounded type" where T is a function type?

Maybe we can just associate a bound type with every bounded type. If something is `int`-bounded, then its associated bound type is `int`. 
Or is that a problem because then the associated type of `X & Y & T` could be both `T` and `Y & T`?
Can we just say that it's the unique non-intersection, non-type variable type that the type is type-bounded by.

Or we could just say:

> A *function-type bounded type* is a type which is *T* bounded where *T* is a function type.
> A function-type bounded type *T* has an associated function type *F*, 
> which is the unique function type *F* such that the type is an *F* bounded type.

specification/dartLangSpec.tex
```diff
@@ -10372,12 +10374,50 @@ \subsubsection{Binding Actuals to Formals}
...
+In particular, a
+\IndexCustom{\DYNAMIC{} bounded type}{type!dynamic bounded}
+is either \DYNAMIC{} itself
+or a type variable whose bound is \DYNAMIC{} bounded,
+or an intersection whose second operand is \DYNAMIC{} bounded.
+Similarly for a
+\IndexCustom{\FUNCTION{} bounded type}{type!function bounded}.
+
+\LMHash{}%
+Similarly, we introduce the notion of a
```
lrhn: This introduced the term "function-type bounded type" but actually defines "function-type bounded type with associated function-type T". 

Should the first paragraph mention "with associated function-type T"?
Maybe just:
> Every function-type bounded type has an associated function type.

State *something* to that effect, so it's said up front.

specification/dartLangSpec.tex
```diff
@@ -10372,12 +10374,50 @@ \subsubsection{Binding Actuals to Formals}
...
+\LMHash{}%
+If the static type of $f$ is \DYNAMIC{} bounded or \FUNCTION{} bounded,
+no further static checks are performed on the invocation $i$
+(\commentary{apart from separate static checks on subterms like arguments}),
+and the static type of $i$ is \DYNAMIC{}.
+Otherwise, it is a compile-time error if the static type of $f$ is not
+function-type bounded.
+
+\LMHash{}%
+Otherwise, the static type of $f$ is function-type bounded
```
lrhn: Feels redundant with the "Otherwise" above. Maybe just grammatically, but I find it hard to understand.

Consider rewriting the above to;

> It's a compile-time error if the static type of $f$ is not a `dynamic`-bounded, `Function`-bounded 
> or function-type bounded type.

Maybe start the paragraph with that, then handle `dynamic` and `Function`, then this "Otherwise" fits right into the flow.

That avoids the two "Otherwise"s in a row.

specification/dartLangSpec.tex
eernstg: 'Never' doesn't occur in the spec ('Nevertheless' does occur, twice ;). We could add a `TODO` comment about `Never` and the relationship to this topic, but I suspect that the text that we will add about `Never` would include these special cases automatically (for instance, `n.foo` is allowed for all `foo` when `n` has type `Never`, and the expression has type `Never`; that will work for `call` as well).

specification/dartLangSpec.tex
eernstg: Yes, I think that will work. Done.

specification/dartLangSpec.tex
eernstg: With the previous change the notion of the associated function type is defined explicitly.

specification/dartLangSpec.tex
eernstg: Rewritten to avoid the awkward 'Otherwise'.

eernstg:
Thanks for input! The only remaining issue is whether we'd trust the treatment of `Never` to include the cases relevant for this PR, because it will be very inclusive already.

specification/dartLangSpec.tex
```diff
@@ -11267,9 +11303,20 @@ \subsubsection{Ordinary Invocation}
 (\ref{privacy})
 instance member named $m$, unless either:
```
lrhn: Does `X & Foo` have an accessible instance member `foo` if `Foo` has?
(I admit I don't know how we specify the member lookup on type variables).

That is, do we need the *associated bound type* to have the member, or can we say it directly on the type variable type?

lrhn:
I think Never can be handled similarly if necessary. If we allow member invocations on `Never`, then similar phrasing should cover it (and after normalization, I think `Never` is the only `Never` bounded type).

specification/dartLangSpec.tex
eernstg: I believe we have no spec language on the existence of members apart from the notion of 'the interface of a class', which is typically treated as a set of member signatures.

Added a rule to get this property. We would still need the exceptional cases, though (like having a free-for-all `call` on any type which is Function bounded), so I think it's OK to have the current changes plus that new one about interfaces.

eernstg:
Very good, thank you! Landing.

@leafpetersen, please notify me if you're worried about this change, then I'll take another iteration in a new PR.

eernstg:
No work needed: The issue came up because 2.7.2 (dartpad.dev) reports an error when `f` of type `X extends <some function type>` is invoked, but the new test https://dart-review.googlesource.com/c/sdk/+/141249/4/tests/language_2/function_subtype/call_type_variable_test.dart does this, and there are no failures. Also, 2.8.0-dev.18.0 doesn't report the error. So it's all gone soon.

PR #903, opened 2020-03-31
Extend GLB to handle FutureOr

leafpetersen:
This specifies an extension to the lower bound algorithm.  The analyzer has always implemented this as part of its merge algorithm for inference constraints, but not as part of the lower bound computation.  This is a specification change for the NNBD DOWN algorithm to make the merge algorithm and the lower bound algorithm consistent.  See [here](https://github.com/dart-lang/sdk/issues/37439) for discussion.

Proposed tests for this change [here](https://dart-review.googlesource.com/c/sdk/+/141841).

cc @johnniwinther @stefantsov @scheglov @stereotype441 @lrhn @munificent @eernstg 

@scheglov I will file an implementation issue for the analyzer once this has landed.

### DIFF
--- a/resources/type-system/upper-lower-bounds.md
+++ b/resources/type-system/upper-lower-bounds.md
@@ -2,6 +2,11 @@
 
 leafp@google.com
 
+## CHANGELOG
+
+2020.03.30
+  - **CHANGE** Update DOWN algorithm with extensions for FutureOr
+
 This documents the currently implemented upper and lower bound computation,
 modified to account for explicit nullability and the accompanying type system
 changes (including the legacy types).  In the interest of backwards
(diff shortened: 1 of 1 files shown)


resources/type-system/upper-lower-bounds.md
```diff
@@ -304,6 +309,18 @@ follows.
...
 - **DOWN**(`T1`, `T2`) = `T2` if `T2` <: `T1`
+
+- **DOWN**(`FutureOr<T1>`, `FutureOr<T2>`) = `FutureOr<S>`
+  - where `S` is **DOWN**(`T1`, `T2`)
+- **DOWN**(`FutureOr<T1>`, `Future<T2>`) = `Future<S>`
+  - where `S` is **DOWN**(`T1`, `T2`)
+- **DOWN**(`Future<T1>`, `FutureOr<T2>`) = `Future<S>`
+  - where `S` is **DOWN**(`T1`, `T2`)
+- **DOWN**(`FutureOr<T1>`, `T2`) = `S`
+  - where `S` is **DOWN**(`T1`, `T2`)
```
lrhn: Would the underlying rule be that:
 
 **DOWN**(`FutureOr<T1>`, `T2`) is **UP**(**DOWN**(`Future<T1>`, `T2`), **DOWN**(`T1`, `T2`)) ?

(Something is a subtype of a union type if it's a subtype of either type. The greatest common subtype would then be the above, and symmetrically for the second parameter).

I'm not sure I agree with the result of the above rules for:

  **DOWN**(`FutureOr<Future<T>>`, `Future<T>`)

It seems to say that it's `Future`(**DOWN*(`Future<T>`, `T`)), and not `Future<T>`.

leafpetersen: Sure, you could do something like that.  Basically, if you have `A | B` and `C`, then a lower bound is 'D0 | D1` where `D0` is a lower bound for `A` and `C`, and `D1` is a lower bound for `B` and `C`.  The question then is just, what do you do with `D0 | D1` given that we don't have union types?  You can heuristically try to simplify:  check that one is a subtype of another, maybe check that it can be re-written into a `FutureOr<T>` or a `T?` for some `T`.  Then fall back to `Never` as you say?  Or recursively find a lower bound for `D0` and `D1`, but then you've got termination worries to think about.

Overall though, I'm not sure it's worth being smarter here.  As usual, my take is that if you have `FutureOr<Future<int>>` floating around, you have bigger problems.

eernstg: Agreed that [the distributive law](https://proofwiki.org/wiki/Intersection_Distributes_over_Union) and an extensional view of types would directly give us the optimal lower bound, but also that we don't have to be very smart here. After all, nobody complained about the lack of precision so far.

PR #906, opened 2020-04-01
Updated changelog to describe effects of PR #897

eernstg:
Of course, I forgot to push the very last change, so #897 got landed without it. Here it is.


eernstg:
TBR, this is tiny.

PR #907, opened 2020-04-01
Update dynamic invocation type argument rule to use i2b

eernstg:
This was a stale TODO (waiting for instantiation to bound to be specified, but we've had that for a while now). It updates the language specification to specify the behavior that the implementations already have.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -10581,7 +10581,7 @@ instantiation to bound
 (\ref{instantiationToBound})
 on the formal type parameters of $f$.
 A dynamic error occurs if this instantiation to bound fails.
-% If the type argument list is not regular-bounded, an error occurs below.
+% If the type arguments do not satisfy the bounds, an error occurs below.
 Otherwise, if $r \not= s$, a \code{NoSuchMethodError} is thrown.
 
 \LMHash{}%
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -10601,8 +10601,9 @@ $o_i, i \in 1 .. m$,
 and $q_j$ is bound to $o_{m+j}, j \in 1 .. l$.
 All remaining formal parameters of $f$ are bound to their default values.
 
-\commentary{
-All of these remaining parameters are necessarily optional and thus have default values.
+\commentary{%
+All of these remaining parameters are necessarily optional
+and thus have default values.%
 }
 
 \LMHash{}%
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -10574,13 +10574,18 @@ derived from an actual argument part of the form
 is bound to the formal type parameters and formal parameters of $f$ as follows:
 
 \LMHash{}%
-% Passing a wrong number of actual type arguments.
+% Passing a wrong number of actual type arguments: Passing none.
 If $r = 0$ and $s > 0$ then replace the actual type argument list:
-%% TODO[instantiate-to-bound]: The actual type arguments passed here
-%% should be chosen based on the instantiate-to-bound algorithm, but we
-%% cannot yet refer to that because it hasn't yet been specified here.
-let $r$ be $s$ and $t_i = \DYNAMIC{}$ for $i \in 1 .. s$.
-Then, if $r \not= s$, a \code{NoSuchMethodError} is thrown.
+let $r$ be $s$ and let $t_i$ for $i \in 1 .. s$ be the result of
+instantiation to bound
+(\ref{instantiationToBound})
+on the formal type parameters of $f$.
+A dynamic error occurs if this instantiation to bound fails,
+and if the resulting type argument list is not regular-bounded.
+% Passing a wrong number of actual type arguments: Passing some.
+Otherwise, if $r \not= s$, a \code{NoSuchMethodError} is thrown.
+
+\LMHash{}%
 % Passing named arguments to a function with optional positional parameters.
 If $l > 0$ and $n \not= h$, a \code{NoSuchMethodError} is thrown.
 % Passing too few or too many positional arguments.
@@ -17128,13 +17133,14 @@ Any self reference in a type alias,
 either directly or recursively via another type declaration,
 is a compile-time error.
 
-\commentary{
+\commentary{%
 This kind of error may also arise when type arguments have been
 omitted in the program, but are added during static analysis
 via instantiation to bound
 (\ref{instantiationToBound})
-or via type inference
-(\commentary{which will be specified later (\ref{overview})}).
+or via type inference,
+which will be specified later
+(\ref{overview}).%
 }
 
 \commentary{%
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -10576,11 +10576,15 @@ \subsubsection{Binding Actuals to Formals}
...
-%% TODO[instantiate-to-bound]: The actual type arguments passed here
-%% should be chosen based on the instantiate-to-bound algorithm, but we
-%% cannot yet refer to that because it hasn't yet been specified here.
-let $r$ be $s$ and $t_i = \DYNAMIC{}$ for $i \in 1 .. s$.
-Then, if $r \not= s$, a \code{NoSuchMethodError} is thrown.
+Let $r$ be $s$ and let $t_i$ for $i \in 1 .. s$ be the result of
+instantiation to bound
+(\ref{instantiationToBound})
+on the formal type parameters of $f$.
+A dynamic error occurs if this instantiation to bound fails.
```
leafpetersen, line 10583: What does it mean for instantiation to bound to fail?  I took a quick read of the algorithm, and didn't see any failure cases.

eernstg: Here is an example:

```dart
class C<X> {}
typedef F<X extends C<X>> = X Function(X);

void f<X extends F<Y>, Y extends C<Y>>(X x) {
  F<Y> fy = x;
}

void main() {
  (f as dynamic)(1); // We can pass anything, the failure occurs before invocation.
}
```

The i2b algorithm yields the type argument list `<F<C<dynamic>>, C<dynamic>>`, but this is an error because `F<C<dynamic>>` is not well-bounded: It is not regular-bounded because `C<dynamic>` is not a subtype of `C<C<dynamic>>`, and it is not super-bounded because the type argument of `F` is in an invariant position, so the super-boundedness check does not substitute anything, so the resulting type is of course (still) not regular-bounded.

leafpetersen: Sorry, I still don't understand where you are going.  As best I can tell, the i2b algorithm never **fails**.  It always produces a result.  As the comment on line 10584 calls out, the result of the i2b algorithm **may** be subject to further checks (e.g. to check that the result is regular bounded).  But reading the spec again, I still don't see any well defined notion of what it means for i2b to fail.  If you by fail, you mean "produces a non-regular bounded type", then I think that should be defined somewhere.

More generally, I'm also not entirely sure that the comment below is correct - we don't require that the arguments be regular-bounded: just that they are subtypes of the actual bound.  It's not clear to me whether or not i2b can ever produce a non-regular-bounded type which is a subtype of the bound from which it was produced, but in any case, the reason it is an error is because it doesn't satisfy the bound - not because it is not regular-bounded.

eernstg: It's a new situation when i2b is performed on the type arguments passed to a generic function invocation or a constructor invocation, and I guess you can say that i2b then can't fail. The reason why I consider i2b to be capable of failing in all the situations that we could previously have is:

1. The subtyping algorithm does not check and does not rely on bounds being satisfied, so specifying that it is checked that the type arguments of a generic function call / constructor invocation / parameterized class satisfy their bounds does not ensure that the types passed as arguments are well-bounded.

1. We specify that every parameterized type occurring in the program must be well-bounded, or an error occurs; but we specify [separately](https://github.com/dart-lang/language/blob/8299f41a53a9b020da9076490cb912902bd16006/specification/dartLangSpec.tex#L5934) that it is a compile-time error for i2b to produce a type which is not well-bounded, exactly because that's not 'a type that occurs in the program'.

The example I mentioned (with a type alias whose type argument is invariant) is the standard example showing that i2b doesn't necessarily produce a type which is well-bounded.

So my take on this was that we'd still need to keep this extra hook active in order to ensure a check that i2b produces well-bounded types, also when applied to invocations, because the checks on type arguments satisfying their bounds still doesn't do it.

But I admit that the specification of [this error](https://github.com/dart-lang/language/blob/8299f41a53a9b020da9076490cb912902bd16006/specification/dartLangSpec.tex#L5934) needs to be generalized slightly in order to cover invocations.

Alternatively, we could enhance subtype checks to include bounds checking, but I think that's less attractive: Currently, it is obvious that the subtype checks are well-defined even in the situation where we haven't established that all parameterized types are well-bounded, so we can always rely on determining that subtype relationships do or do not exist independently of whether we've talked about well-boundedness.

About 'the argument list is regular-bounded': Of course, we don't require that it consists of type arguments T1..Tk such that Tj is regular-bounded for each j. The intention is that the list as a whole is "regular-bounded", but that's sloppy language, so I adjusted it in the comment.

The actual text below (line 10612, 'It is a dynamic type error if $t_i$ is not a subtype..') does require that each type argument satisfies its bound (using 'actual' to indicate that we must substitute the actual type arguments into the bounds before we check for the subtype relationship).

eernstg:
I've rewritten the controversial part substantially. First, we needed a definition of i2b for generic functions (that's not the same concept as i2b for a raw type). That new definition (in section 'The instantiation to bound algorithm') includes the checks for well-boundedness and bound satisfaction as a static check, records the outcome by marking said function `$f$` as admitting or not admitting 'dynamically chosen type arguments', and then it's simply a dynamic error if we ever try to execute `$f$` without type arguments and it doesn't admit that.

eernstg:
@leafpetersen, do you support landing the updated version?

specification/dartLangSpec.tex
```diff
@@ -5998,6 +5998,27 @@ \subsubsection{The Instantiation to Bound Algorithm}
...
+letting $S_i$ denote the result of instantiation to bound on $B_i$,
+and letting $S_i$ be \DYNAMIC{} when the $i$th bound is omitted.
+
+\LMHash{}%
+Let $f$ be a generic function declaration.
+If instantiation to bound on $f$ yields
+a list of type arguments \List{T}{1}{k} such that,
+for some $j \in 1..k$,
+$T_j$ is or contains a type which is not well-bounded,
+or if \List{T}{1}{k} does not satisfy their bounds,
```
leafpetersen, line 6017: It's a little confusing to say "satisfy their bounds", since the bounds in question are not a property of the type arguments but of the function.  Consider re-phrasing as "does not satisfy the bounds on the formal parameters to `f`" or something similar?

specification/dartLangSpec.tex
```diff
@@ -5998,6 +5998,27 @@ \subsubsection{The Instantiation to Bound Algorithm}
...
+then we say that
+\IndexCustom{$f$ does not admit dynamically chosen type arguments}{%
```
leafpetersen, line 6019: Maybe "does not have default type arguments" instead?  That feels closer to the intent: the type arguments may still be chosen dynamically, they just must be provided explicitly.

specification/dartLangSpec.tex
eernstg: Indeed, done!

specification/dartLangSpec.tex
eernstg: Done.

PR #913, opened 2020-04-03
Specify error for a function returning Never which may return

eernstg:
[Edit, eernstg: I'm closing this PR because one change is already subsumed by existing rules in the language specification, and I'd prefer to avoid the other change (and allow certain programs), and this PR only contained those two changes.]

It seems reasonable to flag a function that may return (like `return;` or `return d;` where `d` has type `dynamic`) if it has return type `Never`. I put this error into and just after the error that catches the situation where these functions (along with other functions whose return type is potentially non-nullable) may complete normally.

Cf. https://github.com/dart-lang/language/issues/914 as well.

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -352,6 +352,9 @@ assigned (see Definite Assignment below).
 It is an error if the body of a method, function, getter, or function expression
 with a potentially non-nullable return type **may complete normally**.
 
+It is an error if the body of a method, function, getter, or function expression
+with return type `Never` **returns with or without an object**.
+
 It is an error if an optional parameter (named or otherwise) with no default
 value has a potentially non-nullable type **except** in the parameter list of an
 abstract method declaration.
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -350,10 +350,11 @@ initializer expression and is not marked `late` is used before it is definitely
 assigned (see Definite Assignment below).
 
 It is an error if the body of a method, function, getter, or function expression
-with a potentially non-nullable return type **may complete normally**.
+with a potentially non-nullable return type **may complete normally** or may
+return without an object.
 
 It is an error if the body of a method, function, getter, or function expression
-with return type `Never` **returns with or without an object**.
+with return type `Never` may return with an object.
 
 It is an error if an optional parameter (named or otherwise) with no default
 value has a potentially non-nullable type **except** in the parameter list of an


accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -350,7 +350,11 @@ initializer expression and is not marked `late` is used before it is definitely
 assigned (see Definite Assignment below).
 
 It is an error if the body of a method, function, getter, or function expression
-with a potentially non-nullable return type **may complete normally**.
+with a potentially non-nullable return type **may complete normally** or may
+return without an object.
```
leafpetersen, line 354: I think it may be worthwhile fixing up the errors and warnings around returns, but I'm not sure this bit here is necessary, nor sufficient.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -350,7 +350,11 @@ initializer expression and is not marked `late` is used before it is definitely
...
+
+It is an error if the body of a method, function, getter, or function expression
+with return type `Never` may return with an object.
```
leafpetersen, line 357: I'm not sure that this is necessary, nor desirable.

```dart
Never bar() => throw "Unreachable";
Never foo() {return bar();}
```

is a good program, and should be supported, I think.

```dart
dynamic bar() => throw "Unreachable";
Never foo() {return bar();}
```

is a bad program, but the reason has nothing to do with `Never`.  There's just an implicit downcast there from `dyanamic` to `Never`.

I think this program:
```dart
Never foo() {return;}
```

does need some further specification.

accepted/future-releases/nnbd/feature-specification.md
eernstg: Right, we could say that the language specification [already](https://github.com/dart-lang/language/blob/897375afe38379589ecfdffd3e7675d124545be7/specification/dartLangSpec.tex#L15434) makes it an error to have `return;` in a function unless it returns `void`, `dynamic`, or `Null`, which already implies `without an object` here; and for an `async` function [it is](https://github.com/dart-lang/language/blob/897375afe38379589ecfdffd3e7675d124545be7/specification/dartLangSpec.tex#L15474) unless `flatten(T)` is `void`, `dynamic`, or `Null`, where `T` is the declared return type.

`return;` is always OK in a generator, and that would remain so.

OK, I'll revert this change: The existing rules do actually cover all these cases already.

accepted/future-releases/nnbd/feature-specification.md
eernstg:
```dart
Never bar() => throw "Unreachable";
Never foo() { return bar(); }
```

This can be defended because it should always be OK to return something of type `T` when the return type is `T`. It does have some of the same weirdness as `void f() { return print("Hello!"); }`, in that it is misleading to read `return ..` and think "a value is returned here". But we allow that exactly because it returns `void` to `void`, so we should probably accept a similar weirdness with `Never` as well.

```dart
dynamic bar() => throw "Unreachable";
Never foo() { return bar(); }
```

However, I'm not convinced that this is a bad program. We just decided that it will be allowed with null-safety enabled to implicitly cast from `dynamic` to any type. So why should a cast from `dynamic` to `Never` be an exception?

I argue in #913 that we would then trust the expression of static type `dynamic` to loop forever, or to throw. But this is no more risky than trusting an expression of static type `dynamic` to evaluate to an object whose interface has any given member, say, with `d.fooBar(42)`.

In any case, if that expression of static type `dynamic` does _not_ live up to the expectation (and it does evaluate to an object), then there will be a dynamic check, and we will get an exception at run time.

```dart
Never foo() { return; }
```

As I argued above (at line 354) that this is already covered by the rules in `\section{Return}` of the language specification, so I don't think we need more to make that an error.

eernstg:
Closing this PR because it is empty. The discussion (especially [here](https://github.com/dart-lang/language/pull/913#discussion_r404092234)) gives further reasons.

Issue #914, opened 2020-04-03
Update return rules for null-safety?


eernstg:
Cf. https://github.com/dart-lang/sdk/issues/41324.

We may wish to revisit the [rules](https://github.com/dart-lang/language/blob/897375afe38379589ecfdffd3e7675d124545be7/specification/dartLangSpec.tex#L15408) about `return e;` because the rules in the language specification mention `Null` specifically in one case, and may need to treat `Never` specifically in other cases).

For instance, `Future<Never> f() async => null as dynamic;` is a compile-time error with the current rules, with null-safety. But this is because `Future<dynamic>` is no longer assignable to `Future<Never>` (so the wrapping with `Future` gives the new assignability a distinction that the old assignability did not have). In contrast, `Never f1() => null as dynamic;` is OK (with null-safety as well as without) because `dynamic` is assignable to `Never`.

In both cases we could argue that an expression of type `dynamic` can be trusted to throw or loop (that is, to not complete, with or without a value) just as well as it could be trusted to return an object of an arbitrary type which is required by the context, or to some type that has a specific member with a specific signature. And we do trust the `dynamic` expression for all those purposes.

So it seems consistent to allow `return e;` when `e` has type `dynamic`, both in a synchronous non-generator function and in an asynchronous non-generator function.

If we do this then we'd need to adjust the language specification to avoid the wrapping in `Future<...>`. This change should be a no-op semantically for now (same meaning), but it should ensure that the rules are still specifying the intended behavior when we introduce null-safety into the language specification.

@munificent, @lrhn, @leafpetersen, WDYT?

leafpetersen:
I'm not sure I see any reason to change any of these.  I do think we should clean up the errors/warnings around return statements as part of this release, since they're in an odd state.  We specified them as warnings to avoid breakage, but then some or all of them got implemented as errors in the analyzer, so currently the analyzer and CFE differ.  We should take this opportunity to make all of them consistent.

eernstg:
The point is that null-safety introduces an anomaly:

```dart
// In a legacy library.
int f1() => 42 as dynamic; // OK.
Future<int> f2() async => 42 as dynamic; // OK.

// In a library where null-safety is enabled.
int f3() => 42 as dynamic; // OK.
Future<int> f4() async => 42 as dynamic; // Error.
```

[...]

PR #930, opened 2020-04-16
Introduced "future value type" of `async` functions; used it to correct return rules

eernstg:
Turns out that the existing rules did not correctly cover the following case:

```dart
import 'dart:async';

Future<Future<int>> g() async {
  return Future<int>.value(42);
}
```

The introduction of the _future value type_ of an `async` function ensures that the situation gets one level of indirection simpler, and I used this simplification to solve the above problem.

This also means that the rules about return statements in `async` functions have been changed such that they are no longer based on wrapping both the type of the expression whose value is used to complete the returned future, as well as the component type of the returned future (that is, the constraint on the possible instances used to complete that future). This is discussed in #914.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -42,6 +42,8 @@
 %   the given Iterable/Stream must have a safe element type.
 % - Clarify that an expression of type `X extends T` can be invoked when
 %   `T` is a function type, plus other similar cases.
+% - Define the 'future value type' of an `async` function. Used for returns,
+%   correcting the spec error reported as language issue #929.
 %
 % 2.6
 % - Specify static analysis of a "callable object" invocation (where
@@ -1366,7 +1368,8 @@ It is a compile-time error if the declared return type of a function marked \cod
 
 \LMHash{}%
 We define the notion of the
-\IndexCustom{element type of a generator}{function!generator!element type}
+\IndexCustom{element type of a generator function}{%
+  function!generator!element type}
 as follows:
 %
 If the function $f$ is a synchronous generator
@@ -1383,12 +1386,30 @@ Otherwise, if the function $f$ is a generator
 then the element type of $f$ is \DYNAMIC.
 
 \commentary{%
-%% TODO(eernst): Come nnbd, change 'a top type' to \DYNAMIC.
+%% TODO(eernst): Come nnbd, change 'a top type' to 'a top type or `Object`'.
 In the latter case the return type is a top type,
 because the declaration of $f$ would otherwise be a compile-time error.
 This implies that there is no information about
 the type of elements that the generator will yield.%
 }
+
+% Note, come nnbd, that the following is not just `flatten(T)`: With return
+% type `Future<int>?` we must get the future value type `int`, not `int?`.
+\LMHash{}%
+We define the notion of the
+\IndexCustom{future value type of an asynchronous non-generator function}{%
+  function!asynchronous non-generator!future value type}
+Let the function $f$ be an asynchronous non-generator.
+If, for some $U$, the declared return type $T$ of $f$ is
+% An error has occurred unless `Future<U> <: $T$` for some `U`, so $T$ is
+% `Future<U>` or `FutureOr<U>` (come nnbd: or `Object`, `Future<U>?`, or
+% `FutureOr<U>?`) for some `U`, or it is a top type.
+\code{Future<$U$>} or \code{FutureOr<$U$>}
+%% TODO(eernst): Come nnbd: `or \code{Future<$U$>?} or \code{FutureOr<$U$>?}`
+then the future value type of $f$ is $U$.
+%
+Otherwise, if $T$ is \VOID{} then the future value type of $f$ is \VOID.
+Otherwise, the future value type of $f$ is \DYNAMIC.
 
 
 \subsection{Function Declarations}
@@ -15468,11 +15489,14 @@ to return an object in a return statement.%
 \LMHash{}%
 \Case{Asynchronous non-generator functions}
 Consider the case where $f$ is an asynchronous non-generator function
-(\ref{functions}).
+(\ref{functions})
+with future value type
+(\ref{functions})
+$T_v$.
 %
 % Returning without an object is only ok for async-"voidy" return types.
 It is a compile-time error if $s$ is \code{\RETURN{};},
-unless \flatten{T}
+unless $T_v$
 (\ref{functionExpressions})
 is \VOID{}, \DYNAMIC{}, or \code{Null}.
 %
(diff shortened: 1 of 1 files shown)


eernstg:
> you're proposing a change to the pre-nnbd return rules? 

I'm proposing that we fix the spec bug which makes the `Future<Future<...>>` example a compile-time error. Note that `dart` and `dart2js` accept and run this program, and the analyzer reports an error where both the return type and the type of the returned expression is obviously wrong (so the associated code in the analyzer must be revisited in any case, and checking "`S` is assignable to `T` or `flatten(S)` is assignable to `T`" could be a very localized change).

With that, I believe that this PR makes the pre-nnbd return rules for the relevant situation say what we intended to say all the time.

Cf. the discussion in #914:

If you both prefer model/option 1 then we should _not_ put it into the language specification at this point, because that is a breaking change:

```dart
import 'dart:async';

Future<num> f() {
  Future f = Future<int>.value(2); // Or `FutureOr f = ..`.
  return f;
}

void main() async {
  print(await f());
}
```

This program works in pre-null-safety Dart, but with model 1 it will be a compile-time error, and I'm not convinced that it is helpful for anybody to backport this breaking change.

leafpetersen:
I don't think changing the pre-nnbd rules is a priority right now.

eernstg:
I expect that the language specification with null-safety will rely on the existing language specification for thousands of details. So fixing a bug there serves both the direct purpose of fixing the bug in pre-nnbd Dart (and I understand that you think this is less important), but it also ensures that the specification is ready for null-safety.

This ensures that the transition to null-safety will be well understood: If we wish to switch from model 2 ('S is assignable to T or flatten(S) is assignable to T') to model 1 ('S is assignable to T or flatten(S) <: T'), then that's what we will say in the nnb spec.

If we don't fix the spec bug now then we'll need to do it later, and it will be more difficult to understand what difference null-safety makes for `async` returns, because it will be a combination of a bug fix and an actual breaking change (2 -> 1, if we do want that).

lrhn:
I don't think that changing the pre-NNBD *behavior* is a priority.
If we change the specification, it should be to match the actual behavior.
We've lived with the current behavior for years now, so changing it is not urgent.

The NNBD behavior is more important because it doesn't work unless modified.
We need to fix that, and ensure that the fix is of good enough quality to keep working for some more years without further changes.

That's still limited by not wanting to add too big changes to NNBD at this point.
(Which is why I don't expect the "don't await futures at all" solution to get through, sadly).

eernstg:
Just wondering, @lrhn, do you agree that the notion of the 'future value type' of an async function (or element type, or whatever we want to call it) clarifies and simplifies the rules overall?

Also, do you agree that `Future<Future<int>> f() async => Future<int>.value(42);` should not be a compile-time error?

> If we change the specification, it should be to match the actual behavior.

I believe this PR describes the existing behavior of the CFE. The analyzer reports an error for the case that I mentioned (that is, `return aFutureOfU` with return type `Future<Future<U>>`, plus various situations that are small variations of this situation).

We could decide by fiat that there is no implementation effort: The CFE is fine, and the analyzer behavior is a known bug (because it is a faithful implementation of a spec rule that didn't have exactly the intended properties). So we just decide that we don't intend to fix it before null-safety is introduced.

> The NNBD behavior is more important because it doesn't work unless modified

We can easily keep '`S` is assignable to `T` or flatten(S) is assignable to `T`', or change it to '`S` is assignable to `T` or `flatten(S) <: T`'.

With that, and the rest of the specification switched to null-safety, I do believe that we have a working model.

For the dynamic semantics, I don't see why we shouldn't be able to keep the current specification, if we prefer that.

I proposed the 'extended static await' (xsa) rules in order to make all properties, and in particular the presence/absence of await, statically known whenever possible.

I checked that it is a non-breaking change (except for that very marginal case where the returned expression `e` has static type `T` and evaluates to an object that has type `T` _as well as_ type `Future<T>`; in that case the current semantics will await the future, and the xsa semantics will use the object directly).  Other than that, the difference between the current semantics and the xsa semantics is only that the current semantics will incur a dynamic error in some cases where the xsa semantics continues execution.

So the xsa rules are a very-very-nearly non-breaking change, and I find the static nature of these rules helpful, in order to be able to tell developers what's going on.

If we use `flatten(S) <: T` then I believe that the two semantics coincide apart from the weird `T & Future<T>` case (the extras allowed by the 'extended static await' semantics are then compile-time errors).

I would still prefer to use the xsa rules because they are more static (and I think they make programs more predictable), but the difference is very little at that point.

lrhn:
> Just wondering, @lrhn, do you agree that the notion of the 'future value type' of an async function (or element type, or whatever we want to call it) clarifies and simplifies the rules overall?

That's a good move, especially since it's not just *flatten* of the return type, as I initially assumed.

> Also, do you agree that Future<Future<int>> f() async => Future<int>.value(42); should not be a compile-time error?

Yes. That's just wrong. That said, I don't particularly care whether we fix it in pre-NNBD code, as long as we fix it for post-NNBD code. We've lived with the pre-NNBD behavior for years now, and haven't gotten any serious complaints. Risking a breaking change is not worth it. So ...

> We could decide by fiat that there is no implementation effort: The CFE is fine, and the analyzer behavior is a known bug (because it is a faithful implementation of a spec rule that didn't have exactly the intended properties). So we just decide that we don't intend to fix it before null-safety is introduced.

Yes to that. The analyzer can fix the bug eventually, if they want to, but there is no urgency, but we'd have to specify the behavior they should fix it to (and not as "what the CFE does").

So, post-NNBD. I think I prefer "*flatten*(*S*) &lt;: *T*" to "*flatten*(*S*) assignable to *T*".

I prefer just "*S* assignable to *T*" even more, but fear it's too breaking (would require changes to the migration tool, and it's probably too late now, because it would require re-migrating already migrated code).

The *extended static await* rule is interesting, though.

> 1. If `S <: T`: Desugar to `_return r`.
> 2. If `S` is `dynamic`: Desugar to `_return (r is Future<T>) ? await r : (r as T)`.
> 3. If `S` implements `Future<U>`, `U` assignable to `T`: Desugar to `_return await r`, resp. `_return (await r) as T`.
> 4. If `S` is `FutureOr<U>`, `U` assignable to `T`: Desugar to `if (r is Future<T>) _return await r; else _return r;`, resp. the same thing with `as T` in the else branch.

The tricky case is:
```dart
Future<Future<int>> foo() async {
  Future<dynamic> x = Future<int>.value(42);
  return x;
}
```
Here the static rules would use item 3. and await the `Future<dynamic>`, get an `int` and throw because it's not a `Future<int>`. Had it not awaited, the code would have worked.

That code currently works in our implementations, and it's allowed by the run-time semantics, so that would be the breaking change.

I do like that the only implicit down-casts happen when the original static type contains `dynamic`.
I'm slightly iffy about the difference between `Future<dynamic>` and `FutureOr<dynamic>`.

The former will await any future, then down-cast the result to `T` . The latter only awaits `Future<T>`s, then tries to downcast other futures to `T`, so an actual `Future<dynamic>` would not be awaited even though it's not a `T`. That's an asymmetry which irks me.

How about these "generalized static await" rules (to give them a name):
* *S* <: *T* &mapsto; `_return r;`
* *S* is `dynamic` &mapsto; `_return r as T;`
* *S* implements <code>Future&lt;*T*&gt;</code> &mapsto; `_return await r;`
* *S* implements <code>Future&lt;dynamic&gt;</code> &mapsto; `_return (await r) as T;`
* *S* is <code>FutureOr&lt;*U*&gt;</code> and *U* <: *T* &mapsto; `_return r is Future<U> ? await r : r;`
* *S* is <code>FutureOr&lt;*dynamic*&gt;</code> `_return (r is Future<dynamic> ? await r : r) as T;` 

I like that the type we are checking the `FutureOr<U>`/`FutureOr<dynamic>` against is one of the parts of the union type. This guarantees that we are actually deconstructing the union type, and not accidentally cutting along some other division line and getting accidental matches.
(If `Future<T>` allows part of `U`, then deconstructing `FutureOr<U>` by checking `is Future<T>` may potentially await some `U`s that are not `Future<U>`s. It still works type-wise, but it feels wrong).

I think the rules can be written shorter as:

* *S* assignable to *T* &mapsto; `_return r as T;`
* *R* = *flatten*(*S*) &ne; *S* and *R* assignable to *T*  &mapsto; <code>_return (r is Future<*R*> ? await r : r) as T;</code>

with some of the tests and casts being redundant in some cases.

lrhn:
Initial steps seem fine.

The one thing I worry about wrt. NNBD semantics is that we change the run-time semantics in a way that cause new run-time errors. That is, the same programs are still accepted, but they have slightly different semantics in a few edge cases. That's still going to hit some people, and there won't be any good static help.
Statically refusing something that will be a run-time error is good, though. As always. :)

eernstg:
We agreed to perform all the changes in null-safety world, so there will not be any changes to the language specification at this point (it happens in the nnbd spec). So I'll close this PR without landing it.

PR #936, opened 2020-04-22
Remove passive aggressive language from the spec

mraleph:
Nobody should be shamed for asking. Asking is the primary way to learn.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -5104,7 +5104,7 @@ even though \THIS{} is not available.
 }
 
 \commentary{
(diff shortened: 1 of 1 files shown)


mraleph:
@eernstg PTAL

mateusfccp:
I disagree. You ask when you don't have quick and easy access on one subject. When the answer is a "Google" of distance, you should simply Google and get your answer.

Unsurprinsingly, typing "F-bounded quantification" returns a lot of results:
https://lmgtfy.com/?q=f-bounded+quantification

PR #940, opened 2020-04-24
Add rule about static check on `e1 == e2`

eernstg:
Thanks to @johnniwinther for raising this. The null-safety specification does not seem to specify how to perform static checking of an expression of the form `e1 == e2`. This expression needs special attention because it amounts to some checks for null, possibly followed by an invocation of `operator ==`, and hence the static analysis needs to allow null as the value of `e1` as well as the value of `e2`.

It is possible to declare an overriding `operator ==` with a covariant parameter, and we should preserve the property that this gives rise to a stronger constraint on the expression. Both `e1` and `e2` can still be null, but when `e1` is not null and the operator will be invoked, `e2` should be assignable to `T?` where `T` is the (covariant) parameter type of `operator ==` in the interface of the receiver.

This PR adds a paragraph to section 'Expression Typing' to that effect.

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -503,11 +503,13 @@ Consider an expression `e` of the form `e1 == e2` where the static type of
 formal parameter of `operator ==` in the interface of **NonNull**(`T1`).
 It is a compile-time error unless `T2` is assignable to `S?`.
 
-_In particular, even if the static type of `e1` is potentially nullable,
-the parameter type of the `operator ==` of the corresponding non-null type
-is taken into account. Similarly, it is not a compile-time error for the
-static type of `e2` to be potentially nullable, even when the parameter type
-of said `operator ==` is non-nullable (which is a very common case)._
+_Even if the static type of `e1` is potentially nullable, the parameter type
+of the `operator ==` of the corresponding non-null type is taken into account,
+because that instance method will not be invoked when `e1` is null. Similarly,
+it is not a compile-time error for the static type of `e2` to be potentially
+nullable, even when the parameter type of said `operator ==` is non-nullable.
+This is again safe, because the instance method will not be invoked when `e2`
+is null._
 
 In legacy mode, an override of `operator ==` with no explicit parameter type
 inherits the parameter type of the overridden method if any override of
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -6,7 +6,7 @@ Status: Draft
 
 ## CHANGELOG
 
-2020.04.24
+2020.04.30
   - Specify static analysis of `e1 == e2`.
 
 2020.04.20
@@ -489,6 +489,10 @@ on `Object`, regardless of the type of the member as declared on `T` (note that
 the type as declared on `T` must be a subtype of the type on `Object`, and so
 choosing the `Object` type is a sound choice.  The opposite choice is not
 sound).
+
+_Note that evaluation of an expression `e` of the form `e1 == e2` is not an
+invocation of `operator ==`, it includes special treatment of null. The
+precise rules are specified later in this section._
 
 Calling a method (including an operator) or getter on a receiver of static type
 `Never` is treated by static analysis as producing a result of type `Never`.
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -495,10 +495,16 @@ of a function call produces a result of type `Never`.
 
 The static type of a `throw e` expression is `Never`.
 
-In legacy mode, an override of operator== with no explicit parameter type
+An expression of the form `e1 == e2` where the static type of `e1` and `e2` is
+`T1` respectively `T2` is type checked as an invocation of the instance method
+`operator ==` on **NonNull**(`T1`) with an actual argument of type
+**NonNull**(`T2`).
+
+In legacy mode, an override of `operator ==` with no explicit parameter type
 inherits the parameter type of the overridden method if any override of
-operator== between the overriding method and Object.== has an explicit parameter
-type.  Otherwise, the parameter type of the overriding method is dynamic.
+`operator ==` between the overriding method and `Object.==` has an explicit
+parameter type.  Otherwise, the parameter type of the overriding method is
+`dynamic`.
 
 Top level variable and local function inference is performed
 as
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -498,20 +498,16 @@ of a function call produces a result of type `Never`.
 
 The static type of a `throw e` expression is `Never`.
 
-An expression `e` of the form `e1 == e2` is treated as
-`let v1 = e1, v2 = e2 in v1 != null ? (v2 != null ? v1.==(v2) : false) : (v2 !=
-null ? false : true)`,
-where `v1.==(v2)` denotes an invocation of the instance method `operator ==`
-on `v1`, and the `let` construct is defined in the section 'Null aware
-operator'.
+Consider an expression `e` of the form `e1 == e2` where the static type of
+`e1` is `T1` and the static type of `e2` is `T2`. Let `S` be the type of the
+formal parameter of `operator ==` in the interface of **NonNull**(`T1`).
+It is a compile-time error unless `T2` is assignable to `S?`.
 
 _In particular, even if the static type of `e1` is potentially nullable,
 the parameter type of the `operator ==` of the corresponding non-null type
 is taken into account. Similarly, it is not a compile-time error for the
-static type of `e2` to be potentially nullable even when `operator ==` is
-statically known to have a parameter type which is non-nullable. Of course,
-implementations may eliminate the parts of this expression that are known to
-be unreachable, e.g., if the static type of `e1` is strictly non-nullable._
+static type of `e2` to be potentially nullable, even when the parameter type
+of said `operator ==` is non-nullable (which is a very common case)._
 
 In legacy mode, an override of `operator ==` with no explicit parameter type
 inherits the parameter type of the overridden method if any override of
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -498,10 +498,20 @@ of a function call produces a result of type `Never`.
 
 The static type of a `throw e` expression is `Never`.
 
-An expression of the form `e1 == e2` where the static type of `e1` and `e2` is
-`T1` respectively `T2` is type checked as an invocation of the instance method
-`operator ==` on **NonNull**(`T1`) with an actual argument of type
-**NonNull**(`T2`).
+An expression `e` of the form `e1 == e2` is treated as
+`let v1 = e1, v2 = e2 in v1 != null ? (v2 != null ? v1.==(v2) : false) : (v2 !=
+null ? false : true)`,
+where `v1.==(v2)` denotes an invocation of the instance method `operator ==`
+on `v1`, and the `let` construct is defined in the section 'Null aware
+operator'.
+
+_In particular, even if the static type of `e1` is potentially nullable,
+the parameter type of the `operator ==` of the corresponding non-null type
+is taken into account. Similarly, it is not a compile-time error for the
+static type of `e2` to be potentially nullable even when `operator ==` is
+statically known to have a parameter type which is non-nullable. Of course,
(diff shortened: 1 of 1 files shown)


accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -495,10 +498,16 @@ of a function call produces a result of type `Never`.
 
 The static type of a `throw e` expression is `Never`.
 
-In legacy mode, an override of operator== with no explicit parameter type
+An expression of the form `e1 == e2` where the static type of `e1` and `e2` is
```
leafpetersen, line 501: I don't think this does what we want.  For a number of types `T`, **NonNull**(`T`) is nullable (e.g. `dynamic`, `Function`, or `FutureOr<S>` where `S` is nullable).  This formulation would not allow that invocation.  I think the `**NonNull**(`T1`) part is ok, but I think you need to use something like `T2` assignable to `S?` where `S` is the parameter type of `operator ==` on `**NonNull**(`T`).  Does that make sense?

eernstg: Of course! I changed the approach such that it uses desugaring to specify the expression in general (for the static analysis as well as the dynamic semantics), and used the `let` construct which is introduced later in the nnbd spec. The resulting expression would then rely on promotion to ensure that it is statically known that the `operator ==` is only invoked when both operands are non-null. How do you like that?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -495,10 +498,26 @@ of a function call produces a result of type `Never`.
 
 The static type of a `throw e` expression is `Never`.
 
-In legacy mode, an override of operator== with no explicit parameter type
+An expression `e` of the form `e1 == e2` is treated as
+`let v1 = e1, v2 = e2 in v1 != null ? (v2 != null ? v1.==(v2) : false) : (v2 !=
```
leafpetersen, line 502: I don't think this de-sugaring captures what we want - as written, it implies that:

```
int x;
FutureOr<int?> y;
x == y;
```
is an error.  You clarify further down in the commentary, but I think the result is confusing.  I'd prefer to specify this directly - I think it's a lot clearer.  That is, say that:

Given `e1 == e2` of static type `T1` and `T2` respectively:
  -  Let `S` be the declared type of the parameter to the `operator ==` method on `T1` as defined elsewhere
  - It is an error if `T1` is not assignable to `S?`.
    - Note specifically that it is not necessarily an error if `T1` is not assignable to `S`.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -495,10 +498,26 @@ of a function call produces a result of type `Never`.
...
+null ? false : true)`,
+where `v1.==(v2)` denotes an invocation of the instance method `operator ==`
+on `v1`, and the `let` construct is defined in the section 'Null aware
+operator'.
+
+_In particular, even if the static type of `e1` is potentially nullable,
```
leafpetersen, line 508: This disagrees with what I've written above about the types of the `Object` members when invoked on a nullable type, and I'm not sure I see a need to make an exception here.

accepted/future-releases/nnbd/feature-specification.md
eernstg: Amazingly, given a promotable variable `v`, `v != null ? ../* v isn't null */.. : ....` will obviously make the claim in the comment sound, but since null promotion relies completely on **NonNull**, it is actually not always known that `v` is non-null at that point.

The test you mention will drop a little bit of type information, though:

```dart
class A {
  operator ==(covariant A other) => ...;
  ...
}

void main() {
  A a = A();
  A? aq = a;
  a == 1; // Compile-time error: `1` does not have type `A`.
  aq == 2; // OK.
}
```

We _do_ know that also the second test will fail (based on the statically known type of the receiver in the situation where the `operator ==` is actually invoked), but the parameter type will be `Object` because the type of `aq` is nullable. So we could rely on 'Let `S` be the declared type of the parameter to the `operator ==` method on **NonNull**(`T1`)' to include this extra information, at least for the cases where **NonNull** is able to eliminate the null.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -495,10 +498,24 @@ of a function call produces a result of type `Never`.
 
 The static type of a `throw e` expression is `Never`.
 
-In legacy mode, an override of operator== with no explicit parameter type
+Consider an expression `e` of the form `e1 == e2` where the static type of
+`e1` is `T1` and the static type of `e2` is `T2`. Let `S` be the type of the
+formal parameter of `operator ==` in the interface of **NonNull**(`T1`).
```
leafpetersen, line 503: This contradicts the text beginning on line 485.  I'm ok with making this change (to treat == specially), but I really don't want to leave this kind of trap for readers of the spec.  The text around line 485 should call out that there is an exception specified below (e.g. "With the exception of operator== which is specified separately below, the type used for static analysis.... ").

accepted/future-releases/nnbd/feature-specification.md
eernstg: Resolving -- this was rewritten completely.

accepted/future-releases/nnbd/feature-specification.md
eernstg: I think there is no real conflict here: Evaluation of `e1 == e2` is not an instance member invocation, and there is no way we can get away with ignoring the difference. So the rules about instance member invocations do not apply directly for `e1 == e2` anyway.

In particular, `Object.==` has a formal parameter of type `Object`, so we can't explain why `x == null` is allowed unless we make it explicit that this is different from an invocation of `operator ==` on the value of `x` (hence "assignable to `S?`", not ".. `S`").

Using **NonNull**(`T1`) as the receiver type when we look up the formal parameter type of `operator ==` is also justified by the special null pre-processing in the semantics.

accepted/future-releases/nnbd/feature-specification.md
eernstg: I added a commentary paragraph just after line 485 to make it explicit that `e1 == e2` is not just an invocation of the instance member `operator ==`, and that the precise rules are specified later in the same section.

eernstg:
I'll write some tests now.

eernstg:
Tests added in https://dart-review.googlesource.com/c/sdk/+/145582.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -495,10 +502,24 @@ of a function call produces a result of type `Never`.
...
+It is a compile-time error unless `T2` is assignable to `S?`.
```
lrhn, line 508: I think this is a mistake, and we should not do it.

No-one should create an operator == which takes anything except `Object`. A proper implementation must gracefully accept any object as argument and *must not throw*.

We should not support giving the operator method a parameter other than `Object`, and we should not even suggest that it's in any way acceptable.

I'd say that:

> Consider and expression `e` of the form `e1 == e2`. The static type of `e` is `bool`.
> It is a compile-time error if the static type of `e1` or `e2` is `void` or `void*`.

That's all we need.

The semantics should not be worrying about what type `e1` actually has, it should just cast it and `e2` to `Object`, then call `operator==`  on that. If that throws, someone made a mistake and they should fix it.

I would support a hint that warned you if you use anything but `Object` as the parameter type of `operator==`.

leafpetersen: I don't care strongly about this one way or the other.  I think though that you are proposing more than just undoing this change: currently, you can covariantly override the parameter to == on class A and have that respected.  This change is just around whether == on `A?` is treated the same as == on `A`, or differently.  While I don't care much, I don't see a great argument for respecting it for == on A and not on `A?`.  I think you're advocating for never respecting it.  I'm amenable to discussing, maybe open a language issue on it?

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: There may be no technical conflict, but it's confusing.  I think with the additional commentary it's fine though, thanks!

accepted/future-releases/nnbd/feature-specification.md
lrhn: The current spec says only:

> The static type of an equality expression is \code{bool}.

There is no concern for the declaration of `operator==` on the first operand. That's (among other things) what this change is changing, and what I want undone.

If the current implementations are doing something different, they are wrong. That seems to apply to the analyzer, the CFE does not complain (it just throws at run-time like it should for any mal-called covariant parameter).

That is, you *cannot* have your covariant override respected in general, only in the analyzer, and that is a bug according to the specification.

I'm opposing changing the specification to match the analyzer. I'd prefer to keep it matching the CFE.

I'll open an issue for discussing this.

eernstg:
@lrhn, your strong preference for weakening the static analysis of equality expressions seems to be rooted in a desire to teach developers that they should not declare certain signatures on `operator ==` that the language does otherwise support. That's quite opinionated, and we have opted for a more pragmatic approach in several other cases (e.g., allowing an extension `on Never` even though it's likely to be a mistake). But, more importantly, it is not effective.

I think it would be more fair to declare outright that it is a compile-time error for the formal parameter of an `operator ==` to have a type `T` unless `Object <: T`.

This means that people who want to use that kind of signature to get a static check on equality expressions will know up front that this is not possible, so that particular software design is prevented. Alternatively, they'd discover it post-hoc because the expected compile-time error for `myCheckedEqualityThing == someUnrelatedThing` simply does not occur, which is a much less safe and less actionable situation.

Of course, I'd prefer that we use the strongest possible static analysis which is sound, and we do actually have the guarantee that the value of the left operand is non-null in the case where `operator ==` is invoked. So we would not have the above-mentioned compile-time error, and covariant parameters would be supported just like they are in other parts of the language.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: > I'm opposing changing the specification to match the analyzer. I'd prefer to keep it matching the CFE.

I don't understand where you're getting this.  The following code is rejected by both the analyzer and the CFE:

```dart
class A {
  operator ==(covariant int x) => false;
}

void main() {
  var x = new A();
  x == "hello";
}
```

lrhn: My impression of CFE comes from checking in DartPad and locally on a 2.7.2 VM. If the CFE has changed, it is a recent change.

lrhn: I've been convinced that this isn't that bad, and we are already doing it. (Which kindof proves that it isn't bad).

lrhn, line 508: Should we state explicitly that the return type of S.operator== is irrelevant and the static type of `e1==e2` is always bool?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -495,10 +502,24 @@ of a function call produces a result of type `Never`.
 
 The static type of a `throw e` expression is `Never`.
 
-In legacy mode, an override of operator== with no explicit parameter type
+Consider an expression `e` of the form `e1 == e2` where the static type of
+`e1` is `T1` and the static type of `e2` is `T2`. Let `S` be the type of the
+formal parameter of `operator ==` in the interface of **NonNull**(`T1`).
```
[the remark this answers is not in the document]
leafpetersen: Yes.  `Never.==` has type `Never`, and applying it to anything produces a value of type `Never`.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: I think it falls out naturally, but it doesn't hurt to be explicit.  Also, we should specify whether `Never x = throw "error"; var b = x == 3;` infers `Never` for ` b` or `bool`.

lrhn: I'd say that it infers `bool` because the static type of `e1 == e2` is always `bool`. Much simpler that way.

I don't want to special-case `Never`. It's not worth that much effort. Whatever we do should be general and apply to any return type of `operator==` and not assume that it's just `bool` or `Never` and that `Never` is special in known ways.

The run-time behavior is to either return a boolean if either operand is `null`, or return the result of calling `operator==`, which is (currently) restricted to subtypes of `bool`, which currently means `bool` or `Never`.
So, the correct static type would be LUB(`bool`, return-type-of-`operator==`). That's currently `bool`.

I know that `Never` is the only subtype of `bool`, so allowing `x==3` to have any type other than `bool` would mean it's `Never`, and `x == 3` will definitely never evaluate to a value (because `x` won't). But that's a lot of special knowledge.

PR #941, opened 2020-04-27
Specify null-safety `async` returns

eernstg:
This PR on the nnbd spec introduces the requirement "`S` assignable to `T_v` or` flatten(S) <: T_v`" (aka model 1) for a return statement in an `async` function, where `S` is the static type of the returned expression, and `T_v` is the future value type of the function.

This PR also specifies the dynamic semantics, using an approach where the presence/absence of an `await` step is statically determined whenever possible.

The new dynamic semantics is the same as the pre-null-safety dynamic semantics (that is, the one which is currently specified in `dartLangSpec.tex`), except for the case where the returned object dynamic type is a subtype of as well `T_v` as `Future<T_v>`, and the returned object static type is not `dynamic`, but it is a subtype of `T_v`. This case is expected to be very rare.

@lrhn, I made some changes to the rules that you proposed: (1) The case where `S` is `dynamic` is now first: This ensures that we _will_ await a `Future<T_v>`, even in the case where `T_v` is a top type. If we check `S <: T_v` first then that case will apply, and we wouldn't await the future. The typical case is `f() async => Future.value(1) as dynamic;`, and I think it is an unnecessary breaking change if `f` would stop awaiting the future and instead use it to complete the returned `Future<dynamic>`.

(2) I removed the case for `S implements Future<dynamic>`, because that is a compile-time error with model 1 (because otherwise an earlier case would apply). (3) I removed the case for `S is FutureOr<dynamic>` for the same reason.

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -296,7 +296,7 @@ the last case will only be applied when `S` is a non-`void` top type._
 
 The static analysis of return statements is changed in the following
 way, where `$T$` is the declared return type and `$S$` is the static type of
-the expression `e`:
+the expression `e`.
 
 At [this location](https://github.com/dart-lang/language/blob/65b8267be0ebb9b3f0849e2061e6132021a4827d/specification/dartLangSpec.tex#L15477)
 about synchronous non-generator functions, the text is changed as follows:
@@ -338,8 +338,7 @@ to return a future when the future value type is a suitable future; for
 instance, we can have `return Future<int>.value(42)` in an `async` function
 with declared return type `Future<Future<int>>`. Conversely, it is no longer
 allowed to return a `Future<dynamic>` or `FutureOr<dynamic>` when the future
-value type is `Future<U>` for some `U` which is not a top type, because it is
-no longer enough that `dynamic` is assignable to `U`._
+value type is `Future<U>` for some `U` which is not a top type._
 
 The dynamic semantics specified at
 [this location](https://github.com/dart-lang/language/blob/65b8267be0ebb9b3f0849e2061e6132021a4827d/specification/dartLangSpec.tex#L15597)
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -271,19 +271,27 @@ The **flatten** function is modified as follows:
 and for all `R`, if `T <: Future<R>` then `S <: R`; then **flatten**(`T`) = `S`
   - otherwise **flatten**(`T`) = `T`
 
-### The future value type of an asynchronous generator
+### The future value type of an asynchronous non-generator function
 
 _We specify a concept which corresponds to the type of object which is
 delivered by an `async` function, that is, the type of object which is used to
 complete the future which is returned from such a function when it suspends
 for the first time._
 
-Let _f_ be an asynchronous non-generator function. We define the notion of the
-**future value type** of _f_ as follows: If, for some `U`, the declared return
-type `T` of _f_ is `Future<U>`, `FutureOr<U>`, `Future<U>?`, or `FutureOr<U>?`,
-the future value type of _f_ is `U`.
-Otherwise, if `T` is `void` then the future value type of _f_ is `void`.
-Otherwise, the future value type of _f_ is `Object?`.
+Let _f_ be an asynchronous non-generator function with declared return type
+`T`. Then the **future value type** of _f_ is **futureValueType**(`T`).
+The function **futureValueType** is defined as follows:
+
+- **futureValueType**(`S?`) = **futureValueType**(`S`), for all `S`.
+- **futureValueType**(`S*`) = **futureValueType**(`S`), for all `S`.
+- **futureValueType**(`Future<S>`) = `S`, for all `S`.
+- **futureValueType**(`FutureOr<S>`) = `S`, for all `S`.
+- **futureValueType**(`void`) = `void`.
+- Otherwise, for all `S`, **futureValueType**(`S`) = `Object?`.
+
+_Note that it is a compile-time error unless the return type of an asynchronous
+non-generator function is a supertype of `Future<Never>`, which means that
+the last case will only be applied when `S` is a non-`void` top type._
 
 ### Return statements
 
@@ -297,7 +305,7 @@ about synchronous non-generator functions, the text is changed as follows:
 ```
 It is a compile-time error if $s$ is \code{\RETURN{} $e$;},
 $T$ is neither \VOID{} nor \DYNAMIC,
-and $S$ is \VOID.
+and $S$ is \VOID, \code{\VOID?}, or \code{\VOID*}.
 ```
 
 _Comparing to Dart before null-safety, this means that it is no longer allowed
@@ -310,7 +318,7 @@ the text is changed as follows:
 ```
 It is a compile-time error if $s$ is \code{\RETURN{} $e$;},
 $T_v$ is neither \VOID{} nor \DYNAMIC,
-and \flatten{S} is \VOID{}.
+and \flatten{S} is \VOID, \code{\VOID?}, or \code{\VOID*}.
 ```
 
 _Comparing to Dart before null-safety, this means that it is no longer allowed
@@ -349,51 +357,54 @@ Then the return statement $s$ completes returning $o$
 (\ref{statementCompletion}).
 
 \commentary{%
-The case where the evaluation of $e$ throws is covered by
-the general rule which propagates the throwing completion,
-in this case to the call site.%
+The case where the evaluation of $e$ throws is covered by the general rule
+which propagates the throwing completion from $e$ to $s$ to the function body.%
 }
 
 When $f$ is an asynchronous non-generator with future value type $T_v$
 (\ref{functions}),
-the return statement $s$ is desugared at compile time as described below.
-The resulting code uses \code{\_return $e'$},
-which is a statement that evaluates $e'$ to an object
-$o'$, completes the returned future of $f$ with $o'$,
-and finally returns without an object
-(\ref{statementCompletion}),
-thus terminating the execution of $f$.
-Alternatively, if the evaluation of $e'$ throws
-an object \metavar{error} and stack trace \metavar{stackTrace},
-the returned future of $f$ is error-completed
-with \metavar{error} and \metavar{stackTrace},
-and finally the statement returns without an object.
-The desugaring proceeds as follows,
-where $\Gamma$ is the typing environment of $s$,
-and the first item whose requirement is satisfied is used:
+the return statement $s$ is desugared at compile time.
+In the definition, the first item whose requirement is satisfied is used,
+and $\Gamma$ denotes the typing environment of $s$:
 
 \begin{itemize}
 \item% 1
   When $S$ is \DYNAMIC:
-  \code{\VAR r = $e$; \_return r \IS{} Future<$T_v$> ? \AWAIT{} r : (r \AS{} $T_v$);}
+  \code{\VAR{} r = $e$; \RETURN{} r \IS{} Future<$T_v$> ? \AWAIT{} r : (r \AS{} $T_v$);}
 \item% 2
-  When \SubtypeStd{S}{T_v}: \code{\_return $e$;}
+  When $\flatten{S} \not= S$ and \SubtypeStd{\flatten{S}}{T_v}:
+  % flatten(S) != S implies that `Future` or `FutureOr` occurs.
+  \begin{itemize}
+  \item% 2a
+    When \SubtypeStd{S}{\code{Future<$T_v$>}} or
+    $S$ is of the form \code{$S_1$?} such that
+    $S_1$ implements \code{Future<$U$>} and \SubtypeStd{\code{$U$}?}{T_v}:
+    \code{\RETURN{} \AWAIT{} $e$;}
+  \item% 2b
+    When $S$ is of the form \code{FutureOr<$U$>} where \SubtypeStd{U}{T_v},
+    or $S$ is of the form \code{FutureOr<$U$>?}
+    where \SubtypeStd{\code{$U$?}}{T_v}:
+    \code{\VAR{} r = $e$; \RETURN{} r is Future<$T_v$> ? \AWAIT{} r : r;}
+  \end{itemize}
 \item% 3
-  When \SubtypeStd{S}{\code{Future<$T_v$>}}: \code{\_return \AWAIT{} $e$;}
-\item% 4
-  When $S$ is of the form \code{FutureOr<$U$>} where \SubtypeStd{U}{T_v}:
-  \code{\VAR r = $e$; \_return r is Future<$T_v$> ? \AWAIT{} r : r;}
+  When \SubtypeStd{S}{T_v}: \code{\RETURN{} $e$;}
 \end{itemize}
 
 \commentary{%
 Note that one of the above cases apply, or a compile-time error has occurred.%
 % It is required at compile-time that S is assignable to T_v, or that
-% flatten(S) <: T_v. If S is assignable to T_v then item 1 or 2 will apply.
-% Otherwise flatten(S) <: T_v. We know that flatten(S) != S, or S would be
-% assignable to T_v (and item 1 or 2 would apply). So S is of the form
-% FutureOr<S1> where S1 <: T_v; or S implements Future<S1> and
-% flatten(S) == S1. In the latter case item 3 applies, and in the former case
-% item 4 applies.
+% flatten(S) <: T_v.
+%
+% If S is assignable to T_v then at least item 1 or 3 applies.
+%
+% Otherwise flatten(S) <: T_v. If flatten(S) == S, then S <: T_v, which is the
+% previous case (and 1 or 3 applies).
+%
+% So S != flatten(S) <: T_v. In this situation there will be an item
+% under item 2 that applies: S is of the form FutureOr<S1> where S1 <: T_v; or
+% S is of the form FutureOr<S1>? where S1? <: T_v; or S implements Future<S1>
+% and flatten(S) == S1; or S is S1? such that S1 implements Future<S2>, and
+% S2? <: T_v. In each case, subitem 2a or 3b applies.
 }
 
 % The new semantics preserves the pre-null-safety semantics with the following
@@ -409,18 +420,17 @@ Note that one of the above cases apply, or a compile-time error has occurred.%
 % other type of object (e.g., we don't await a `Future<dynamic>`), and always
 % guard an object which is not awaited with a dynamic type check.
 %
-% item 2: Breaking change when the value of $e$ has type `T_v & Future<T_v>`,
+% item 2: New semantics is identical to old semantics.
+%
+% item 3: Breaking change when the value of $e$ has type `Future<T_v> & T_v`,
 % which is expected to be very rare. New semantics will then use said value,
 % old semantics would await it. Remaining cases: identical behavior.
-%
-% item 3 and 4: New semantics is identical to old semantics.
 
 \commentary{%
-There cases where $f$ is a generator cannot occur,
+The cases where $f$ is a generator cannot occur,
 because in that case $s$ is a compile-time error.%
 }
 ```
-
 
 ### Static errors
 #### Nullability definitions
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -348,41 +348,78 @@ the actual return type of $f$
 Then the return statement $s$ completes returning $o$
 (\ref{statementCompletion}).
 
-When $f$ is an asynchronous non-generator with future value type $T_v$,
-the return statement $s$ is desugared as follows at compile time,
-where \code{\SUSPEND{} $e$} is a statement that evaluates $e$ to an object
-$o$, then completes the returned future if $f$ with $o$, and finally suspends
-the execution of $f$. The first item whose requirement is satisfied is used:
+\commentary{%
+The case where the evaluation of $e$ throws is covered by
+the general rule which propagates the throwing completion,
+in this case to the call site.%
+}
+
+When $f$ is an asynchronous non-generator with future value type $T_v$
+(\ref{functions}),
+the return statement $s$ is desugared at compile time as described below.
+The resulting code uses \code{\_return $e'$},
+which is a statement that evaluates $e'$ to an object
+$o'$, completes the returned future of $f$ with $o'$,
+and finally returns without an object
+(\ref{statementCompletion}),
+thus terminating the execution of $f$.
+Alternatively, if the evaluation of $e'$ throws
+an object \metavar{error} and stack trace \metavar{stackTrace},
+the returned future of $f$ is error-completed
+with \metavar{error} and \metavar{stackTrace},
+and finally the statement returns without an object.
+The desugaring proceeds as follows,
+where $\Gamma$ is the typing environment of $s$,
+and the first item whose requirement is satisfied is used:
 
 \begin{itemize}
-\item When $S$ is a subtype of $T_v$: \code{\SUSPEND{} $e$;}.
-\item When $S$ is \DYNAMIC{}:
-  \code{\VAR r = $e$; \SUSPEND{} r \IS{} Future<$T_v$> ? \AWAIT{} r : (r \AS{} $T_v$)}.
-\item When $
-
+\item% 1
+  When $S$ is \DYNAMIC:
+  \code{\VAR r = $e$; \_return r \IS{} Future<$T_v$> ? \AWAIT{} r : (r \AS{} $T_v$);}
+\item% 2
+  When \SubtypeStd{S}{T_v}: \code{\_return $e$;}
+\item% 3
+  When \SubtypeStd{S}{\code{Future<$T_v$>}}: \code{\_return \AWAIT{} $e$;}
+\item% 4
+  When $S$ is of the form \code{FutureOr<$U$>} where \SubtypeStd{U}{T_v}:
+  \code{\VAR r = $e$; \_return r is Future<$T_v$> ? \AWAIT{} r : r;}
 \end{itemize}
 
+\commentary{%
+Note that one of the above cases apply, or a compile-time error has occurred.%
+% It is required at compile-time that S is assignable to T_v, or that
+% flatten(S) <: T_v. If S is assignable to T_v then item 1 or 2 will apply.
+% Otherwise flatten(S) <: T_v. We know that flatten(S) != S, or S would be
+% assignable to T_v (and item 1 or 2 would apply). So S is of the form
+% FutureOr<S1> where S1 <: T_v; or S implements Future<S1> and
+% flatten(S) == S1. In the latter case item 3 applies, and in the former case
+% item 4 applies.
+}
 
+% The new semantics preserves the pre-null-safety semantics with the following
+% exceptions:
+%
+% The old semantics tests for `is Future<flatten(T)>` where `T` is the declared
+% return type; the new semantics tests for `is Future<T_v>` where `T_v` is the
+% future value type. They differ only for null-safety specific types of the
+% forms `Future<U>?`, `FutureOr<U>?`, or `Object`. So the use of 'future value
+% type' causes no difference in behavior.
+%
+% item 1: Identical behavior, we always await a `Future<T_v>`, never await any
+% other type of object (e.g., we don't await a `Future<dynamic>`), and always
+% guard an object which is not awaited with a dynamic type check.
+%
+% item 2: Breaking change when the value of $e$ has type `T_v & Future<T_v>`,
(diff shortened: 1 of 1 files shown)


accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +268,157 @@ The **flatten** function is modified as follows:
...
+\end{itemize}
+
+\commentary{%
+Note that one of the above cases apply, or a compile-time error has occurred.%
+% It is required at compile-time that S is assignable to T_v, or that
+% flatten(S) <: T_v. If S is assignable to T_v then item 1 or 2 will apply.
+% Otherwise flatten(S) <: T_v. We know that flatten(S) != S, or S would be
+% assignable to T_v (and item 1 or 2 would apply). So S is of the form
+% FutureOr<S1> where S1 <: T_v; or S implements Future<S1> and
+% flatten(S) == S1. In the latter case item 3 applies, and in the former case
```
lrhn, line 392: We do not know that those are the only forms (any more). As mentioned above, we need to account for `?` types.

So, let's be extra careful about this case analysis.

It's still *correct* to simply have a 3rd item of the form:

  3. If *U* = *flatten*(*S*), *U* <: *T<sub>v</sub>*: `var r = e; return r is Future<U> ? await r : r;`.

The non-await branch might be unreachable if *S* is a non-FutureOr type, but that can just be an implementation optimization.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +268,157 @@ The **flatten** function is modified as follows:
...
+% guard an object which is not awaited with a dynamic type check.
+%
+% item 2: Breaking change when the value of $e$ has type `T_v & Future<T_v>`,
+% which is expected to be very rare. New semantics will then use said value,
+% old semantics would await it. Remaining cases: identical behavior.
+%
+% item 3 and 4: New semantics is identical to old semantics.
+
+\commentary{%
+There cases where $f$ is a generator cannot occur,
```
lrhn, line 416: "There" -> "The"

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +268,157 @@ The **flatten** function is modified as follows:
 and for all `R`, if `T <: Future<R>` then `S <: R`; then **flatten**(`T`) = `S`
   - otherwise **flatten**(`T`) = `T`
 
+### The future value type of an asynchronous generator
```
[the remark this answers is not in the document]
eernstg: Done.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +268,157 @@ The **flatten** function is modified as follows:
...
+
+_We specify a concept which corresponds to the type of object which is
+delivered by an `async` function, that is, the type of object which is used to
+complete the future which is returned from such a function when it suspends
+for the first time._
+
+Let _f_ be an asynchronous non-generator function. We define the notion of the
+**future value type** of _f_ as follows: If, for some `U`, the declared return
+type `T` of _f_ is `Future<U>`, `FutureOr<U>`, `Future<U>?`, or `FutureOr<U>?`,
+the future value type of _f_ is `U`.
```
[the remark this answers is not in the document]
eernstg: I believe we have a split here:

Normalization is applied in just two specific situations according to the nnbd **spec** (mixed inheritance conflict mitigation and reified `Type` equality).

However, the **implementations** (certainly the CFE, probably the analyzer as well) will apply normalization in several very common situations in addition to the specified situations: The nullability of a type is represented as the value of a field (not as an operator on a type, yielding a wrapped object that represents the type, and a wrapper object that represents the `*` or `?`). One consequence of this representation choice is that it is impossible for a CFE representation of a type to model `T**` for any `T`; when we add the second `*`, it amounts to setting the nullability field to 'legacy' (I think it's called `indefinite`), and setting that field the second time is a normalization, because we can't see that it was done twice. Similarly, if the nullability was 'nullable' (`?`) and it is made 'legacy' it will just remain 'nullable'.

This means that normalization is used very sparingly in the spec, but it is also used very frequently in certain situations. Luckily, those very frequent non-specified normalizations are probably benign. I do not think that the "heavier" normalizations like `FutureOr<dynamic>` --> `dynamic` will ever occur except in the specified situations.

This means that in practice we will never see `Future<int>??` and such.

However, I'll reword the text in such a way that it _will_ work with the spec model (where `T**` can exist). Definitely that would be the safe approach right now.

That said, we probably don't have to insist on aligning the implementations and the spec in this respect, as long as we are aware of the fact that we may predict a type like `T**` based on the spec, and the implementations will insist that it's `T*`. It is going to be tricky to observe the difference, because normalization will never introduce nor eliminate any subtyping relationships (we always have `T <: NORM(T) <: T`).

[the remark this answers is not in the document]
lrhn: So the spec needs to specify what *flatten*(FutureOr<int?*?>*?*)` is too, and everywhere we say "of the form ..." we need to account or any number of modifiers afterwards?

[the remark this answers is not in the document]
eernstg: Alternatively, we could have more than one kind of normalization: The "small scale" normalization that just replaces `**` by `*` and so on everywhere and the "large scale" that replaces actual type structure (like `FutureOr<T>` --> `T` for some `T`), which is only applied.

We could also try to argue that the difference is not observable (except for some unspecified cases like `print(something.runtimeType)`). The subtype rules will make most of the distinctions go away, because of the `T <: NORM(T) <: T` property, so we can only see the difference when there is no subtyping step involved, e.g., in declarations of the superinterfaces of a class.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +268,157 @@ The **flatten** function is modified as follows:
...
+to "return void to null" in a regular function._
+
+At [this location](https://github.com/dart-lang/language/blob/65b8267be0ebb9b3f0849e2061e6132021a4827d/specification/dartLangSpec.tex#L15525)
+about an asynchronous non-generator function with future value type `$T_v$`,
+the text is changed as follows:
+
+```
+It is a compile-time error if $s$ is \code{\RETURN{} $e$;},
+$T_v$ is neither \VOID{} nor \DYNAMIC,
+and \flatten{S} is \VOID{}.
```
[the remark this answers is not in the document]
eernstg: I know that @johnniwinther considers normalization "everywhere" to be disruptive, so that's already an important thing to take into account. I think one of the reasons for this is performance, but it's better that @johnniwinther comments on this as he sees fit.

This means that it may be a good choice to apply normalization rather sparsely. An example could be that some people want to use the type `FutureOr<void>` to indicate that a value may be a future whose completion should be ignored, or it might be "something else" which should be ignored. Of course, that type is questionable in its own right, but we decided to allow it, and it does get special treatment: It is allowed to perform a type test on an expression with that type, so it is not treated exactly like an expression of type `void`.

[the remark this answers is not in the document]
eernstg: I changed the rule to raise the error when `$S$` is `void`, `void?`, or `void*`.

It is possible that we won't mention anything involving legacy types in the language specification, but I think it makes sense to include `void*` here, and then it is easy to delete it when the language specification is updated, if that's what we do.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +268,157 @@ The **flatten** function is modified as follows:
...
+The resulting code uses \code{\_return $e'$},
+which is a statement that evaluates $e'$ to an object
+$o'$, completes the returned future of $f$ with $o'$,
+and finally returns without an object
+(\ref{statementCompletion}),
+thus terminating the execution of $f$.
+Alternatively, if the evaluation of $e'$ throws
+an object \metavar{error} and stack trace \metavar{stackTrace},
+the returned future of $f$ is error-completed
+with \metavar{error} and \metavar{stackTrace},
```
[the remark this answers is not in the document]
eernstg: Indeed, I forgot that the "post-processing" that will complete the returned future and handle suspension is done in section 'Function Invocation'.

This is actually one of the main improvements that was obtained by the spec refactoring that you performed a while back, where statement completions were introduced. So it's all so nice and simple now! ;-)

Done.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +268,157 @@ The **flatten** function is modified as follows:
...
+\begin{itemize}
+\item% 1
+  When $S$ is \DYNAMIC:
+  \code{\VAR r = $e$; \_return r \IS{} Future<$T_v$> ? \AWAIT{} r : (r \AS{} $T_v$);}
+\item% 2
+  When \SubtypeStd{S}{T_v}: \code{\_return $e$;}
+\item% 3
+  When \SubtypeStd{S}{\code{Future<$T_v$>}}: \code{\_return \AWAIT{} $e$;}
+\item% 4
+  When $S$ is of the form \code{FutureOr<$U$>} where \SubtypeStd{U}{T_v}:
```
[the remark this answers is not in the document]
eernstg: Right, those `?` cases were missing. Added them.

PS: I swapped item 2 and 3, because this reduces the set of situations where there is a breaking change to the semantics. We have the following breaking change example which may or may not be worrying:

```dart
Future<Object?> f() async => Future.value(1) as Object?;
```

This function will await the future with the old semantics, but it will complete the returned future with the future with the new semantics.

Note that the example needs all the twists: `T_v` must be a top type such that `R <: Future<T_v> & T_v` reduces to `R <: Future<Object?>`, where `R` is the dynamic type of the value of `$e$`. And the static type of `$e$` cannot be `Future<..>`, because then item 2 applies (and we _will_ await the future, which is the old semantics).

Other than that, only `class MyWeirdo implements C1, Future<C2> {}` where `C2 <: C1` will give rise to breaking changes to the semantics.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +268,157 @@ The **flatten** function is modified as follows:
...
+  \code{\VAR r = $e$; \_return r is Future<$T_v$> ? \AWAIT{} r : r;}
```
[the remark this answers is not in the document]
eernstg: The reason why I changed it to `Future<T_v>` is that this is obviously not changing the semantics (compared to today).

So it would be good to know which cases would make a difference, and whether the beauty is worth the breakage. Do you have an example where the beauty is evident? ;-)

accepted/future-releases/nnbd/feature-specification.md
eernstg: I changed item 2 to cover the cases, PTAL there.

accepted/future-releases/nnbd/feature-specification.md
eernstg: Done.

eernstg:
Rethinking this a bit:

The purpose of switching to a dynamic semantics built on "static await" criteria would be that programs are more comprehensible when it is specified directly that an await operation will definitely occur, or will definitely not occur.

However, the rules turn out to be so complex that it might not help the reader of code in practice. So maybe we should simply keep the old semantics? It's somewhat quirky, and they do make it less predictable when there will be an await operation, but the old semantics _is_ considerably simpler...

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +268,167 @@ The **flatten** function is modified as follows:
...
+% S2? <: T_v. In each case, subitem 2a or 3b applies.
+}
+
+% The new semantics preserves the pre-null-safety semantics with the following
+% exceptions:
+%
+% The old semantics tests for `is Future<flatten(T)>` where `T` is the declared
+% return type; the new semantics tests for `is Future<T_v>` where `T_v` is the
+% future value type. They differ only for null-safety specific types of the
+% forms `Future<U>?`, `FutureOr<U>?`, or `Object`. So the use of 'future value
```
leafpetersen, line 413: Actually this does change the behavior right?  I think this is a problem with the current semantics.

https://github.com/dart-lang/language/issues/944

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +268,167 @@ The **flatten** function is modified as follows:
...
+which propagates the throwing completion from $e$ to $s$ to the function body.%
+}
+
+When $f$ is an asynchronous non-generator with future value type $T_v$
+(\ref{functions}),
+the return statement $s$ is desugared at compile time.
+In the definition, the first item whose requirement is satisfied is used,
+and $\Gamma$ denotes the typing environment of $s$:
+
+\begin{itemize}
```
leafpetersen, line 367: As an aside - I don't think we should be writing extensive latex in these markdown documents, it's unreadable.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +268,167 @@ The **flatten** function is modified as follows:
...
+% forms `Future<U>?`, `FutureOr<U>?`, or `Object`. So the use of 'future value
+% type' causes no difference in behavior.
+%
+% item 1: Identical behavior, we always await a `Future<T_v>`, never await any
+% other type of object (e.g., we don't await a `Future<dynamic>`), and always
+% guard an object which is not awaited with a dynamic type check.
+%
+% item 2: New semantics is identical to old semantics.
+%
+% item 3: Breaking change when the value of $e$ has type `Future<T_v> & T_v`,
```
leafpetersen, line 422: I don't think this is right.  I think this is a much bigger change than implied here.  In particular, I believe that this proposal changes the following program from printing `int` to printing `Future<int>`:

```
Object foo() async {
  Object f = Future<Object>.value(3);
  return f;
}
void main() async {
  var f = foo();
  print((await f).runtimeType);
}
```

leafpetersen:
As I mentioned in my review, I think we may need to use future value type in the dynamic semantics for reasons that I filed an issue about (unless I'm confusing myself in the late hour).  I'm not convinced that the rest of the change to the dynamic semantics pulls its weight though.  I think it's more breaking than advertised, and I don't see that it really makes anything clearer (if anything, less so).

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +268,167 @@ The **flatten** function is modified as follows:
...
+way, where `$T$` is the declared return type and `$S$` is the static type of
+the expression `e`:
+
+At [this location](https://github.com/dart-lang/language/blob/65b8267be0ebb9b3f0849e2061e6132021a4827d/specification/dartLangSpec.tex#L15477)
+about synchronous non-generator functions, the text is changed as follows:
+
+```
+It is a compile-time error if $s$ is \code{\RETURN{} $e$;},
+$T$ is neither \VOID{} nor \DYNAMIC,
+and $S$ is \VOID, \code{\VOID?}, or \code{\VOID*}.
```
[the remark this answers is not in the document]
eernstg: True. Removed `void?`. It is also inconsistent to keep `void*`, but it seems silly to omit it. For instance, the `void*` would be an error in the corresponding situation in the legacy library where that type came from.

But then we ought to add a general remark that it is an error for an expression to have type `void*`, unless it occurs in one of the locations which are whitelisted for void expressions.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +268,167 @@ The **flatten** function is modified as follows:
 and for all `R`, if `T <: Future<R>` then `S <: R`; then **flatten**(`T`) = `S`
   - otherwise **flatten**(`T`) = `T`
 
+### The future value type of an asynchronous non-generator function
+
+_We specify a concept which corresponds to the type of object which is
```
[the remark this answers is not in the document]
eernstg: Done.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +268,167 @@ The **flatten** function is modified as follows:
...
+to "return void to null" in a regular function._
+
+At [this location](https://github.com/dart-lang/language/blob/65b8267be0ebb9b3f0849e2061e6132021a4827d/specification/dartLangSpec.tex#L15525)
+about an asynchronous non-generator function with future value type `$T_v$`,
+the text is changed as follows:
+
+```
+It is a compile-time error if $s$ is \code{\RETURN{} $e$;},
+$T_v$ is neither \VOID{} nor \DYNAMIC,
+and \flatten{S} is \VOID, \code{\VOID?}, or \code{\VOID*}.
```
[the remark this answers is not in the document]
eernstg: Done.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +268,167 @@ The **flatten** function is modified as follows:
...
+_Comparing to Dart before null-safety, this means that it is no longer allowed
+to "return void to null" in an `async` function, nor to "return a void future
+to null"._
+
+The next sentence is changed as follows:
+
+```
+It is a compile-time error if $s$ is \code{\RETURN{} $e$;},
+\flatten{S} is not \VOID,
+$S$ is not assignable to $T$,
```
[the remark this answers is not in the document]
eernstg: Right, thanks! Done.

accepted/future-releases/nnbd/feature-specification.md
eernstg: Right, we want to fix the soundness issue that would arise if we were to use the old semantics in a null-safety context, but the issue raised in #944 only exists when null-safety features are present. Adjusted the comment.

accepted/future-releases/nnbd/feature-specification.md
eernstg: I wanted to make it obvious what the changes are in the case where the proposal is a concrete language specification update.

Do you prefer that we don't aim at concrete language specification updates at all (in which case there would be a need to write more context), or perhaps that we write updates for the language specification, but using pure markdown to indicate an approximation of the update?

accepted/future-releases/nnbd/feature-specification.md
eernstg: Yes, I gave the following concrete example in an earlier comment:

```dart
Future<Object?> f() async => Future.value(1) as Object?;
```

This will complete the returned future with `1` with the old semantics and it will complete it with the future with the new semantics.

eernstg:
> I don't see that it really makes anything clearer

I agree, and I'm also worried about the breakage. So I'd suggest that we use the existing dynamic semantics, except for the 'future value type' soundness fix.

eernstg:
I uploaded another commit where we use the old semantics, only adjusted to have the 'future value type' correction.

lrhn:
My reason for preferring a static approach is not (just) simplicity, but lack of run-time overhead for type-checking - at least when it's possible.

So, the static rule for `return e;` in an `async` function is:

> Let *T* be the future value type of the function, let *S* be the static type of *e* with context type `FutureOr<T>`.
> It's a static type error unless *S* is assignable to *T* or *flatten*(*S*) is a subtype of *T*.

and the dynamic semantics are:

> 1. Evaluate *e* to a value *o*.
> 2. If *o* implements `Future<X>`, *X* <: *T*, let *r* be the object resulting from awaiting *o*.
> 3. Otherwise let *r* be *o*.
> 4. Then `return e;` returns with the value *r*.

Those two match up, but I'd still prefer to change 2. to:

> 2. If *o* implements <code>Future<*X*></code>, *X* <: *flatten*(*S*), let *r* be the object resulting from awaiting *o*.

The only difference would occur when *S* <: <code>Future<*T*></code> and *flatten*(*S*) is not a subtype of *T*. That ... cannot happen? If it can't, then the change is safe and it matches the static check better. If it can happen, ... I'd like to see how :)

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: My goals are:
  - As much as possible, we have a single point of truth for the *changes* involved for a new language feature
  - Have the changes specified in a readable format

The first is why I really don't want to land some changes in the formal spec and some in the feature spec. 
  -  We could do away with feature specs, but I do find it useful having a single point of truth for all of the *new* things involved in a language feature.  
  - We could land changes in the formal spec, but then we'd have to point to long series of diffs.
  - We could maintain a CL with the accumulated changes to the spec, but then people have to read it in unformatted latex, and also it's tricky to see what the workflow is to land changes.

I don't claim to have a perfect solution here.  What I've been doing is using the feature spec, written in markdown, as a readable source of truth, with the intention that we then land, at some point all of the changes implied by the feature spec in the formal spec.

I'm open to other workflows.  One thing we could consider is combining them: maintain a CL with the latex changes, which we aim to be fully comprehensive, and have the feature spec be an abbreviated summary with pointers to the full language?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +270,113 @@ The **flatten** function is modified as follows:
...
+to return a future when the future value type is a suitable future; for
+instance, we can have `return Future<int>.value(42)` in an `async` function
+with declared return type `Future<Future<int>>`. Conversely, it is no longer
+allowed to return a `Future<dynamic>` or `FutureOr<dynamic>` when the future
+value type is `Future<U>` for some `U` which is not a top type._
+
+The dynamic semantics specified at
+[this location](https://github.com/dart-lang/language/blob/65b8267be0ebb9b3f0849e2061e6132021a4827d/specification/dartLangSpec.tex#L15597)
+is changed as follows, where `$f$` is the enclosing function with declared
+return type `$T$`, and `$e$` is the returned expression with static type `$S$`:
```
leafpetersen, line 345: You don't mention `S` below, so eliminate it?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +270,113 @@ The **flatten** function is modified as follows:
...
+- **futureValueType**(`S?`) = **futureValueType**(`S`), for all `S`.
+- **futureValueType**(`S*`) = **futureValueType**(`S`), for all `S`.
+- **futureValueType**(`Future<S>`) = `S`, for all `S`.
+- **futureValueType**(`FutureOr<S>`) = `S`, for all `S`.
+- **futureValueType**(`void`) = `void`.
+- Otherwise, for all `S`, **futureValueType**(`S`) = `Object?`.
+
+_Note that it is a compile-time error unless the return type of an asynchronous
+non-generator function is a supertype of `Future<Never>`, which means that
+the last case will only be applied when `S` is a non-`void` top type._
```
lrhn, line 292: ... or `Object`.

(The one non-top-type supertype of Future that we now have)

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +270,113 @@ The **flatten** function is modified as follows:
...
+way, where `$T$` is the declared return type and `$S$` is the static type of
+the expression `e`.
+
+At [this location](https://github.com/dart-lang/language/blob/65b8267be0ebb9b3f0849e2061e6132021a4827d/specification/dartLangSpec.tex#L15477)
+about synchronous non-generator functions, the text is changed as follows:
+
+```
+It is a compile-time error if $s$ is \code{\RETURN{} $e$;},
+$T$ is neither \VOID{} nor \DYNAMIC,
+and $S$ is \VOID{} or \code{\VOID*}.
```
lrhn, line 306: But not `void?` or `void??` or ...?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +270,113 @@ The **flatten** function is modified as follows:
...
+```
+
+_Comparing to Dart before null-safety, this means that it is no longer allowed
+to "return void to null" in a regular function._
```
lrhn, line 310: Have to give an example. I don't understand what "return void to null" means. Or just write out "cannot return a `void` expression if the return type is `null`.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +270,113 @@ The **flatten** function is modified as follows:
...
+The static analysis of return statements is changed in the following
+way, where `$T$` is the declared return type and `$S$` is the static type of
+the expression `e`.
+
+At [this location](https://github.com/dart-lang/language/blob/65b8267be0ebb9b3f0849e2061e6132021a4827d/specification/dartLangSpec.tex#L15477)
+about synchronous non-generator functions, the text is changed as follows:
+
+```
+It is a compile-time error if $s$ is \code{\RETURN{} $e$;},
+$T$ is neither \VOID{} nor \DYNAMIC,
```
lrhn, line 305: Why is this using *T* and not *futureReturnValue*(*T*) ?

If I have:
```dart
Future<void> foo() async {
  return print("ok");
}
```
why is that not OK? 
(I'm fine with it, I'd disallow `return e` if `T_v` is `void` no matter what type `e` has, but it does feel weird that return types `Future<dynamic>` and `dynamic` cause different behavior in the return statement).

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -263,6 +270,113 @@ The **flatten** function is modified as follows:
...
+to "return void to null" in a regular function._
+
+At [this location](https://github.com/dart-lang/language/blob/65b8267be0ebb9b3f0849e2061e6132021a4827d/specification/dartLangSpec.tex#L15525)
+about an asynchronous non-generator function with future value type `$T_v$`,
+the text is changed as follows:
+
+```
+It is a compile-time error if $s$ is \code{\RETURN{} $e$;},
+$T_v$ is neither \VOID{} nor \DYNAMIC,
+and \flatten{S} is \VOID{} or \code{\VOID*}.
```
lrhn, line 319: But not `void?` ?

accepted/future-releases/nnbd/feature-specification.md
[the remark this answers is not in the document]
johnniwinther: The main problem with normalization is when it leaks into the types that we present to the user. If the user has written a type one way but we report problems pertaining to that type in another way it is very confusing. Like

```Dart
Error: Bar isn't a valid type:
  Foo foo = ....
  ^^^
```
For that reason I'd like to avoid normalization in cases where it is not needed for the correct result, i.e. in cases where the user provided type suffices, and only use normalization why we need to synthesize a result.

This is for instance the strategy we use when computing the canonical supertypes:
```Dart
class A<T> {}
class B implements A<FutureOr<void>> {} // B implements A<FutureOr<void>>
class C implements A<dynamic> {} // C implements A<dynamic>
class D implements B, C {} // D implements A<Object?>
```
where we only apply `NNBD_TOP_MERGE(NORM(A<FutureOr<void>>),NORM(A<dynamic>)) = A<Object?>` because we need to for a well-defined supertype relation.

accepted/future-releases/nnbd/feature-specification.md
eernstg: Done.

accepted/future-releases/nnbd/feature-specification.md
eernstg: Exactly. I used to have `void?` as well, but Leaf argued that this would be inconsistent, and I agree.

accepted/future-releases/nnbd/feature-specification.md
eernstg: Done.

accepted/future-releases/nnbd/feature-specification.md
eernstg: Done.

accepted/future-releases/nnbd/feature-specification.md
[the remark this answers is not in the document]
eernstg: The futureValueType function was introduced and rules rewritten to handle non-normalized types.

accepted/future-releases/nnbd/feature-specification.md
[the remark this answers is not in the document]
eernstg: Newest rules do not have this case.

accepted/future-releases/nnbd/feature-specification.md
eernstg: This issue was eliminated by going back to rules that are much more like the current rules.

accepted/future-releases/nnbd/feature-specification.md
eernstg: I'll get some more feedback outside this PR, and rewrite as needed to markdown in a separate PR.

eernstg: I think the overall readability is lowered when the null-safety spec and similar documents contain updated snippets of text from the language specification along with permalinks to the existing text. On the other hand, the link plus the updated text is unambiguous.

I created language issue #947 to clarify this.

eernstg:
@lrhn wrote:
> _S_ <: <code>Future<_T_></code> and _flatten_(S) is not a subtype of _T_.

If `S <: Future<T>` then `S` is `Never` or `S` implements `Future<U>` for some `U` such that `U <: T`, and then **flatten**(`S`) is equal to `Never` or `U`, so said subtype relation is guaranteed to hold (and we don't even need to invoke the non-evaluability of an expression of type `Never`). How lucky. ;-)

PR #948, opened 2020-04-30
Update on top of #941: Some review responses got lost, and `flatten(S)` must be changed in some locations

eernstg:
This PR performs those missing changes discussed in reviews. Also, the changes to several occurrences of `flatten(T)` to `$T_v$` are required for the same reasons that caused the change in the "`S` is assignable to `Tv` or `flatten(S) <: Tv` case.

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -326,11 +326,11 @@ is \VOID, \DYNAMIC, or \code{Null}.
 It is a compile-time error if $s$ is \code{\RETURN{} $e$;},
 $T_v$ is \VOID,
 and \flatten{S} is neither \VOID, \code{\VOID*}, nor \DYNAMIC.
-
+%
 It is a compile-time error if $s$ is \code{\RETURN{} $e$;},
 $T_v$ is neither \VOID{} nor \DYNAMIC,
 and \flatten{S} is \VOID{} or \code{\VOID*}.
-
+%
 It is a compile-time error if $s$ is \code{\RETURN{} $e$;},
 \flatten{S} is not \VOID{} nor \code{\VOID*},
 $S$ is not assignable to $T_v$,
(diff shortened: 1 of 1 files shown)


accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -310,42 +311,51 @@ and $S$ is \VOID{} or \code{\VOID*}.
...
 
 ```
+It is a compile-time error if $s$ is \code{\RETURN{};},
+unless $T_v$
+is \VOID, \DYNAMIC, or \code{Null}.
+%
+It is a compile-time error if $s$ is \code{\RETURN{} $e$;},
+$T_v$ is \VOID,
+and \flatten{S} is neither \VOID, \DYNAMIC, \code{Null},
+\code{\VOID*}, \code{\DYNAMIC*}, nor \code{Null*}.
```
leafpetersen, line 329: This really feels at best pointless to me.  There are hundreds of places in the spec where we talk about `dynamic`, `Null` and `void`, all of which, unless otherwise called out, are intended to apply to `dynamic*`, `Null*` and `void*`.  Likewise for `int` and `int*`.  Shall we go through the spec and point out that every comment about something of type `int` applies equally to something of type `int*`?  I think the intent is clear, and calling it out in some places simply makes things inconsistent and confusing.

Also, we certainly don't intend to put this text into the formal spec as far as I know, since we don't intend to specify the legacy support.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -310,42 +311,51 @@ and $S$ is \VOID{} or \code{\VOID*}.
...
-```
-
-_Comparing to Dart before null-safety, this means that it is no longer allowed
-to "return void to null" in an `async` function, nor to "return a void future
-to null"._
-
-The next sentence is changed as follows:
-
-```
+%
```
leafpetersen, line 334: What if `T_v` is `\DYNAMIC*` or `\VOID*`?  See what I mean?  There's no end to it.

accepted/future-releases/nnbd/feature-specification.md
eernstg: Very good, removed the references to `*` types in the text about returns.

accepted/future-releases/nnbd/feature-specification.md
eernstg: Indeed. ;-)

Issue #677, opened 2019-11-11
Implementations allowing `Ext(c)?.foo()`.


lrhn:
See https://github.com/dart-lang/sdk/issues/39325 for context.

The specification of extension methods states that `Ext(c)?.foo` is disallowed because it treats `Ext(c)` as a value in order to check whether it's `null`.

The language implementations (front-end *and* analyzer) accepts it as implicitly meaning `(tmp = c) == null ? null : Ext(tmp).foo()`. That is, they treat `Ext(c)` as having the same *value* as `c`, just with different "members", and applies the `?.` to that.

That behavior is not unreasonable. It allows users to avoid applying extensions to `null` inline (without a separate `x != null ? ...`), even though all extensions currently accept `null`.
If we ever move to "extension types" where you can write `var tmp = Ext(c); tmp.foo();`, it would still make sense to allow `tmp?.foo()` because at that point `Ext(c)` *will* have a value: the same value as `c` but with a different *static only* type. Checking the actual value for being `null` makes sense then, and allowing `Ext(c)?.foo()` is consistent with that view (evaluate `c` to a value, statically cast it to the extension type, then do `?.foo()` on that value which might be `null`).

So, should we keep the currently implemented behavior and amend the specification?

eernstg:
We [decided](https://github.com/dart-lang/language/issues/175#issuecomment-455108385) that `Object` methods should be callable on receivers whose type is nullable (that's `Object?` methods, of course, which is the reason why this is sound). So `x.toString()` is OK for a nullable `x`.

The case for extension methods is even stronger: It has been one of our standard examples for extension methods that it can be `on` a nullable type, and this can be used to give special treatment to the case where the receiver is null.

So `x.foo()` should surely be allowed in the case where `x` has a nullable (or potentially nullable) type `T` which does match the `on` clause of the extension that declares `foo`.

In general, the nnbd feature spec [says](https://github.com/dart-lang/language/blob/master/accepted/future-releases/nnbd/feature-specification.md#static-errors) that 'it is a warning to use a null aware operator (?., ?.., ??, ??=, or ...?) on a non-nullable receiver'.

We could extend this to say that it is then also a warning (or similar) to use `x?.m` when `m` is a member of `Object?`, because `x.m` would be OK, and similarly for extensions. But we don't do that currently.

This means that we currently support _avoiding_ to execute an `Object?` member using `?.`, e.g., `x?.toString()` where the type of `x` is potentially nullable.

[...]

PR #953, opened 2020-05-06
Specify null-aware explicit extension method invocation

eernstg:
This PR introduces text corresponding to the decision in issue #677.

This update uses `let`, which is a non-standard device in Dart specifications. It does seem to serve the purpose, though, and we can change it when this feature specification gets integrated into the language specification. In any case, an obvious alternative is to spell out the semantics of the desugared form using a fresh variable.

### DIFF
--- a/accepted/2.6/static-extension-members/feature-specification.md
+++ b/accepted/2.6/static-extension-members/feature-specification.md
@@ -40,7 +40,7 @@ A static extension of a type is declared using syntax like:
 extension MyFancyList<T> on List<T> {
   int get doubleLength => this.length * 2;
   List<T> operator-() => this.reversed.toList();
-  List<List<T>> split(int at) => 
+  List<List<T>> split(int at) =>
       <List<T>>[this.sublist(0, at), this.sublist(at)];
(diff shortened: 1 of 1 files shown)


accepted/2.6/static-extension-members/feature-specification.md
```diff
@@ -181,7 +181,7 @@ The static type of a member invocation on an extension application is the return
 
 Composite member invocations, like composite assignment `e.id += 2` or increment `e.id++`, are defined in terms of two individual member invocations (always one get and one set operation). If the target expression of a composite member invocation is an extension application, we need to recognize and handle it specially.
 
-A composite assignment of the form `e1.id += 2` is equivalent to `e1.id = e1.id + 2` except that `e1` is only evaluated once, and the value is used twice. 
+A composite assignment of the form `e1.id += 2` is equivalent to `e1.id = e1.id + 2` except that `e1` is only evaluated once, and the value is used twice.
```
lrhn, line 184: The `e1[e2]=e3` case below has one `e` which should be `e1`.

PR #957, opened 2020-05-09
Specify constraint solving for null safe types

leafpetersen:
Specification for the constraint solving algorithm used during inference, accounting for null safe types.  In addition to changes for null safety, this PR includes several fixes for outstanding known issues.  Specifically:
  - This fixes the issue discovered by @stefantsov and Kallen in the fall in which unknown types were handled incorrectly in the constraint merging algorithm (see the example added under the "Upper bound" section).
  - This fixes a longstanding bug in which inference may cause type variables to escape their scope (see the example in the "Closure of type constraints" section, which currently crashes the CFE).
  - This resolves the issue of how to compute to the greatest and least closures of invariant types correctly.

I have made several adjustments to the ordering of the core inference algorithm to account for the interaction between the two union types (`FutureOr` and `T?`).  I will work on tests next week to validate that these changes produce the results that we want.

Comments and feedback welcome.

cc @johnniwinther @scheglov @stereotype441 @stefantsov @eernstg @lrhn @munificent


resources/type-system/inference.md
```diff
@@ -474,42 +462,78 @@ replaced with `Null`, and every covariant occurrence of `Ti` replaced with
   from any of the `Bi`:
```
scheglov, line 462: I don't understand the "from any of the `Bi`" part here and above.

There are descriptions, which I would understand, don't know which of them are right.
1. "`L` contains any of `Xi`"
2. "any of `Bi` references a type variable from `L`"

resources/type-system/inference.md
```diff
@@ -474,42 +462,78 @@ replaced with `Null`, and every covariant occurrence of `Ti` replaced with
...
-    - `U` is the greatest closure of `T` with respect to `L` 
-    - `Ui` is the least closure of `Ti` with respect to `L` 
+    - `U` is the greatest closure of `T` with respect to `L`
+    - `Ui` is the least closure of `Ti` with respect to `L`
     - with the usual capture avoiding requirement that the `Xi` do not appear in
   `L`.
+- if `S` is `T Function<X0 extends B0, ...., Xk extends Bk>(T0 x0, ...., Tn xn,
+    [Tn+1 xn+1, ..., Tm xm])` or `T Function<X0 extends B0, ...., Xk extends Bk>(T0 x0, ...., Tn xn,
+{Tn+1 xn+1, ..., Tm xm})`and `L` contains any free type variables
+  from any of the `Bi`:
```
scheglov, line 494: Does this rule subsume the first (already existing) rule for function types above, also talking about "contains any free type variables"? It seems that the existing rule was not updated for null safety and still uses `Null` and `Object`.

resources/type-system/inference.md
```diff
@@ -520,23 +544,22 @@ Tt` for some type `Tb` which satisfies schema `Pb`, and some type `Tt` which
...
 Note that the closure of a type constraint implies the original constraint: that
 is, any solution to the original constraint that is closed with respect to `L`,
 is a solution to the new constraint.
 
 The motivation for these operations is that constraint generation may produce a
 constraint on a type variable from an outer scope (say `S`) that refers to a
-type variable from an inner scope (say `T`).  For example, ` <T>(S) -> int <:
-<T>(List<T> -> int ` constrains `List<T>` to be a subtype of `S`.  But this
+type variable from an inner scope (say `T`).  For example, ` <T>(T) -> List<T> <:
+<T>(T -> S) ` constrains `List<T>` to be a subtype of `S`.  But this
```
scheglov, line 562: Should it be `<T>(T) -> List<T> <: <T>(T) -> S`?

resources/type-system/inference.md
```diff
@@ -545,6 +568,19 @@ closure of `List<T>` (i.e. `List<Object>`) as the new supertype constraint on
 `S` results in the constraint `List<Object> <: S`, which implies the original
```
scheglov, line 568: Do you mean `Object?` here and above?

resources/type-system/inference.md
```diff
@@ -407,48 +409,34 @@ The invariant occurrences of a type `T` in another type `S` are:
...
+replaced with `Never`, and every contravariant occurrence of `Xi` replaced with
+`Object?`.  The invariant occurrences are treated as described explicitly below.
 
 We define the greatest closure of a type `M` with respect to a set of type
-variables `T0, ..., Tn` to be `M` with every contravariant occurrence of `Ti`
-replaced with `Null`, and every covariant occurrence of `Ti` replaced with
-`Object`.
+variables `X0, ..., Xn` to be `M` with every contravariant occurrence of `Xi`
+replaced with `Never`, and every covariant occurrence of `Xi` replaced with
+`Object?`. The invariant variables are treated as described explicitly below.
```
eernstg, line 426: Would that be `invariant occurrences`?

resources/type-system/inference.md
```diff
@@ -474,42 +462,78 @@ replaced with `Null`, and every covariant occurrence of `Ti` replaced with
...
+    - `U` is the greatest closure of `T` with respect to `L`
+    - `Ui` is the least closure of `Ti` with respect to `L`
     - with the usual capture avoiding requirement that the `Xi` do not appear in
   `L`.
+- if `S` is `T Function<X0 extends B0, ...., Xk extends Bk>(T0 x0, ...., Tn xn,
+    [Tn+1 xn+1, ..., Tm xm])` or `T Function<X0 extends B0, ...., Xk extends Bk>(T0 x0, ...., Tn xn,
+{Tn+1 xn+1, ..., Tm xm})`and `L` contains any free type variables
+  from any of the `Bi`:
+  - The least closure of `S` with respect to `L` is `Never`
+  - The greatest closure of `S` with respect to `L` is `Object?`
```
eernstg, line 496: I guess this could be `Function`.

resources/type-system/inference.md
eernstg: Right, line 455-460 looks like it is now subsumed by 491-496 (so it doesn't matter that 455-460 isn't updated if it's going away).

resources/type-system/inference.md
```diff
@@ -698,15 +770,32 @@ to `L` under constraints `C`.
...
+  - If `B0i` is a subtype match for `B1i` with constraint set `Ci0`
+  - And `B1i` is a subtype match for `B0i` with constraint set `Ci1`
+  - And `Ci2` is `Ci0 + Ci1`
+  - And `Z0...Zn` are fresh variables with bounds `B20, ..., B2n`
+    - Where `B2i` is `B0i[Z0/T0, ..., Zn/Tn]` if `P` is a type schema
+    - Or `B2i` is `B1i[Z0/S0, ..., Zn/Sn]` if `Q` is a type schema
+      - In other words, we choose the bounds for the fresh variables from
+        whichever of the two generic function types is a type schema and does
+        not contain any variables from `L`.
+  - And `F0[Z0/T0, ..., Zn/Tn]` is a subtype match for `F0[Z0/S0, ..., Zn/Sn]`
```
scheglov, line 786: Do you mean `F1[Z0/S1, ..., Zn/Sn]` as the second function type?

resources/type-system/inference.md
```diff
@@ -698,15 +770,32 @@ to `L` under constraints `C`.
...
-  - If `F0[Z0/T0, ..., Zn/Tn]` is a subtype match for `F0[Z0/S0, ..., Zn/Sn]`
-with respect to `L` under constraints `C`, where each `Zi` is a fresh type
-variable with bound `Bi`.
-  - And `Cl` is `C` with each constraint replaced with its closure with respect
+
+- A generic function type `<T0 extends B00, ..., Tn extends B0n>F0` is a subtype
+match for a generic function type `<S0 extends B10, ..., Sn extends B1n>F1` with
+respect to `L` under constraint set `C2`
+  - If `B0i` is a subtype match for `B1i` with constraint set `Ci0`
+  - And `B1i` is a subtype match for `B0i` with constraint set `Ci1`
```
scheglov, line 778: What are examples of non-equal `B0` and `B1` when `B0 <# B1` and `B1 <# B0`?
Only top types `dynamic`, `Object?`, and `void`?

scheglov:
In my implementation `int*` does not match `num*`.

1. `P = int*`, `Q = num*`.
2. We apply "If P is a legacy type P0*" and match `int <# num*`.
3. We apply "If Q is a legacy type Q0*" and match `int <# num?`.
4. We apply "If `Q` is `Q0?`":
4.1. `int` is not `P0?`, so we don't apply "If P is P0?".
4.2. We apply "Or if P is a subtype match for Q0", and `int <# num` is true, but there are no constraints, so the second part "under non-empty constraint set C" **is false**. So, we continue.
4.3. "Or if P is a subtype match for Null" does not match.
5. "If P is C<M0, ..., Mk> and Q is C<N0, ..., Nk>" directly, or via implemented interface does not apply, because `P = int` but `Q = num?` - has nullable suffix, so interface type matching immediately returns `false`.

I landed initial implementation of subtype matching in a previous CL, and slightly patched it to handle implemented interfaces in legacy libraries better [in this CL](https://dart-review.googlesource.com/c/sdk/+/148946). It also contains the "solo" test for `int* <# num*` issue.

Is there a mistake in the way I apply the rules listed above?
Or is this a hole in the rules?

resources/type-system/inference.md
leafpetersen: Right, done.

resources/type-system/inference.md
leafpetersen: The latter.  Rephrased to be clearer.

resources/type-system/inference.md
leafpetersen: Right, sorry, missed deleting the old definition.  Fixed.

resources/type-system/inference.md
leafpetersen: True, probably better.  cc @scheglov note the change.

resources/type-system/inference.md
leafpetersen: Yes, fixed.

resources/type-system/inference.md
leafpetersen: Right, fixed.

resources/type-system/inference.md
leafpetersen: `void Function<T extends S>()` is a subtype match for `void Function<T extends int>()` with respect to `{S}` under constraint set `{int <: S <: int}`.

resources/type-system/inference.md
leafpetersen: Indeed I do.  Thanks for the close reading.

leafpetersen:
> In my implementation `int*` does not match `num*`.

@scheglov the rules are incorrectly stated.  I'll upload a fix, but basically, the intention with both `Q <# FutureOr<T>` and  `Q <# T? ` is that you:
  - Try to match `Q` against `Future<T>` (or `T` respectively)
  - If that succeeds, the match *will succeed*.  But if it succeeded with no constraints:
     - Try to match `Q` against `T` (or `Null` respectively)
     - If that succeeds, return the constraints (if any it generated)
     - Otherwise return success, but with no constraints
  - If the first match failed, try to match `Q` against `T` (or `Null` respectively) 

Does that make sense?

scheglov:
> Does that make sense?

Yes, thank you.

I think I found a bug in analyzer and CFE inference, which failed to find an error in user code internally.  But new rules find it.

```dart
// @dart = 2.7
import 'dart:async';

class A {}

class B {
  Future<A> foo() {
    return _waitOn(null);
  }

  FutureOr<T> _waitOn<T>(FutureOr<T> value) => throw 0;
}
```

We match `Future<T> <# Future<A>`, and get two separate constraints `_ <: T <: int` and `_ <: T <: Future<int>`. They merge into `_ <: T <: Never`.

resources/type-system/inference.md
```diff
@@ -464,52 +452,80 @@ replaced with `Null`, and every covariant occurrence of `Ti` replaced with
...
+
+```
+class C<X> {
+  C(void Function(X) x);
+}
+T check<T>(C<List<T>> f) {
+  return null as T;
+}
+void test() {
+  var x = check(C((List<int> x) {})); // Should inver `int` for `T`
```
stefantsov, line 525: Small nit: "inver" ==> "infer".

eernstg:
@scheglov, to ensure progress: I have no reason to request further changes, but I assume that you will mark this PR as approved at some point.

leafpetersen:
> I think I found a bug in analyzer and CFE inference, which failed to find an error in user code internally. But new rules find it.

@scheglov can you clarify what the previous behavior and the current behavior is?  I'm surprised that this example changes.  The result that I would expect for the example that you list above is `_ <: T <: Never` after the constraints are merged.

resources/type-system/inference.md
leafpetersen: Done.

scheglov:
> > I think I found a bug in analyzer and CFE inference, which failed to find an error in user code internally. But new rules find it.
> 
> @scheglov can you clarify what the previous behavior and the current behavior is? I'm surprised that this example changes. The result that I would expect for the example that you list above is `_ <: T <: Never` after the constraints are merged.

Right, after the constraints are merged.

The difference is that the inference spec now says that the constraints are merged after collecting, before they are used. But the previous implementation used to collect individual constraints and use them without merging first, just computing upper / lower bound and choosing the candidate inferred type, which is then checked to satisfy every constraint.

Old: `[match][t1: FutureOr<U*>*][t2: Future<A*>*][constraints: {U: ['U' must extend 'A*', 'U' must extend 'Future<A*>*']}]`.
New: `[match][t1: FutureOr<U*>*][t2: Future<A*>*][constraints: {U: ['U' must extend 'Never']}]`
In both cases the downward inferred type is `Null*`.
But in the old case it satisfies each individual constraint.
And in the new - not.

scheglov:
I looked some more on this example.
```dart
// @dart = 2.7
import 'dart:async';

class A {}

class B {
  Future<A> foo() {
    return _waitOn(null);
  }

  FutureOr<T> _waitOn<T>(FutureOr<T> value) => throw 0;
}
```

1. With these new rules individual constraints that we gather are:
```
_ <: T <: A*
_ <: T <: Future<A*>
```
  Note the missing `*` after the `Future<A*>`.
  We get them via the following matching:
```dart
FutureOr<T*>* <# Future<A*>*
FutureOr<T*> <# Future<A*>*
FutureOr<T*> <# Future<A*>?
FutureOr<T*> <# Future<A*>
Future<T*> <# Future<A*>
T* <# A*
T <# A*

T* <# Future<A*>
T <# Future<A*>
```
Then `DOWN(A*, Future<A*>)` gives us `Never`.

2. When we choose type from constraints during downward inference, we do get `Never`, but we immediately convert it into a legacy type (we are in a legacy library), which is `Null*`.

PR #968, opened 2020-05-17
Fix typo

mnordine:


mateusfccp:
Are you sure it is a typo. Isn't it the `iff` as in `if and only if`?

mnordine:
Not sure, that could be the case.

mnordine:
I've seen it in other places in the doc, you're probably right, I'll close it.

PR #969, opened 2020-05-18
Add feature specification of external/abstract variables

eernstg:
Tests and Dart.g update in https://dart-review.googlesource.com/c/sdk/+/148324.

### DIFF
--- a/accepted/future-releases/abstract-external-fields/feature-specification.md
+++ b/accepted/future-releases/abstract-external-fields/feature-specification.md
@@ -1,0 +1,171 @@
+# Abstract Fields and External Fields
+
+Authors: lrn@google.com, eernst@google.com<br>
+Version: 1.0
+
+## Background and Motivation
+
+Dart allows abstract instance methods, getters and setters to be declared in
+classes and mixins. It allows external functions, methods, getters, setters and
+constructors to be declared as top-level, static or instance declarations.
+
+The syntax of an abstract member is simply a declaration with no body. The
+syntax for an external declaration prefixes the declaration with the modifier
+`external`.
+
+There are two use-cases not supported by this:
+
+1. Easily declaring an abstract "field" (a pair of an abstract getter and an
+   abstract setter) inside an "interface" class. Currently that requires two
+   declarations: `int get foo;set foo(int _);`. Users often mistakenly write
+   `int foo;` which works until someone *extends* the class instead of
+   implementing it. Then they get an extra field on every instance of the
+   class.
+2. Easily declaring an external "field". This has come up in `dart:ffi` where
+   users write Dart classes as a way to describe native memory layout. All
+   instances of the interface will be backed by native (external)
+   code. Currently `dart:ffi` users simply declare the field as `int foo;`.
+
+Using a non-abstract field declaration has "worked" for both situations until
+Null Safety. It's not perfect, but it isn't *wrong*.
+
+With Null Safety that approach no longer works. Declaring a field as `int foo;`
+means that it must be initialized by a constructor since non-nullable fields
+cannot be left uninitialized.
+
+So, to provide a migration path for these uses, preferably a *better* one than
+the current pre-Null Safety state, we introduce *abstract and external field
+declarations*.
+
+## Proposal
+
+An instance variable can be declared abstract by prefixing it with the reserved
+word `abstract`.
+
+Currently variable declarations inside a class can have the "modifiers":
+
+* `static` &mdash; makes the variable a class variable instead of an instance variable.
+* `const` &mdash; makes the variable a compile-time constant.
+* `covariant`&mdash; makes the implicit setter's parameter covariant.
+* `late` &mdash; makes the variable "late initialized", allowing it to stay uninitialized after creation.
+* `final` &mdash; makes there be no setter, unless the variable is late and has no initializer.
+* `var` &mdash; makes the variable non-final.
+* a type &mdash; the type of the variable.
+* an initializer expression &mdash; initializes the variable to a specific value on creation or first read.
+
+Not all combinations are allowed.
+
+A `var` cannot be combined with `final` or a type.
+
+The `final` and `covariant` are usually mutually exclusive, because `covariant`
+modifies a setter and `final` prevents a setter from existing (unless the
+variable is `late` and has no initializer).
+
+The `covariant` modifier only applies to instance members, so it's incompatible
+with `static`.
+
+A `const` variable must be `static`, non-`late`, non-`final` (but implicitly
+counts as final) and must have an initializer.
+
+If the modifiers occur, they must occur in the order listed here.
+
+To this we add the modifiers `abstract` and `external`, making the possible
+modifiers:
+
+* `external`
+* `static`
+* `const`
+* `abstract`
+* `covariant`
+* `late`
+* `final`
+* `var`
+* a type.
+
+* initializer expression.
+
+### Abstract Fields
+
+An abstract instance variable is an instance variable with the `abstract`
+modifier. That means a non-static variable declared inside a class or mixin
+declaration.
+
+It *must not* be `external`, `static` or `const`, so the declaration starts with
+`abstract`. It must not be `late` or have an initializer expression.
+
+That makes it a non-external instance variable with no initializer and not
+`late`, which means that it can be represented by an abstract getter and, if not
+`final`, an abstract setter.
+
+Being `late` or `external` or having an initializer are *implementation*
+details. An abstract variable (getter/setter) declaration must not have any
+such. It can be `covariant` if it's not `final`.
+
+The abstract instance variable declaration with type *T* means exactly the same
+as an abstract getter declaration with the same name and return type *T* , and,
+if not `final`, an abstract setter declaration with the same name and parameter
+type *T*. As such, the abstract instance variable introduces only members into
+the class interface, there is no implementation.
+
+Examples:
+
+```dart
+abstract int x; &mapsto; int get x; set x(int x);
+abstract final int x; &mapsto; int get x;
+abstract covariant int x; &mapsto; int get x; set x(covariant int x);
+abstract var x; &mapsto; get x; set x(x); // May inherit types.
+
+```
+
+A declaration of `abstract var x;` or `abstract final x;`, with no declared type, may inherit types from the super-interfaces of the class.
+
+As a special rule for `abstract var x`, If the super-interfaces of the class has only a getter or only a setter, then that type is inherited by the abstract variable declaration and is used for both the implicit getter and setter.
+
+Any metadata on the abstract instance variable declaration applies to both the setter and the getter.
+
+### External Fields
+
+An external field is a variable declaration starting with `external`.
+
+An external variable declaration cannot be `const`, `abstract` or `late`, or
+have an initializer expression. All of these are implementation choices (or
+documented lack of it, for `abstract`), and external fields are considered
+implemented externally, so all implementation details must be kept out of the
+Dart declaration.
+
+Both top-level and class-level variable declarations can be `external` and
+class-level declarations can be `static`. It can be `covariant` only if it's an
+instance variable (not static, not top-level). Local variables cannot be
+external.
+
+An external variable declaration is completely equivalent to an external getter
+declaration and, if not `final`, an external setter declaration, both with the
+same type and name.
+
+This means that external instance variables need not (and must not) be
+initialized by constructors. They are treated exactly as an instance
+getter/setter pair for all static analysis purposes. The implementation will be
+provided by the run-time system somehow.
+
+A instance variable declaration of `external var x;` or `external final x;`,
+with no declared type, may inherit types from the super-interfaces of the
+class.
+
+As a special rule for `external var x`, If the super-interfaces of the class has
+only a getter or only a setter, then that type is inherited by the abstract
+variable declaration and is used for both the implicit getter and setter.
+
+Any metadata on the abstract instance variable declaration applies to both the
+setter and the getter.
+
+## Summary
+
+We allow `abstract` to prefix an instance variable declaration which is not
+`late` and has no initializer.
+
+We allow `external` to prefix a top-level, static or instance variable
+declaration which is not `abstract`, `const` or `late`, and has no initializer.
+
+Both are equivalent to an abstract, respectively external, getter and (unless
+`final`) setter declaration which has the same name, type (as return or
+parameter type), and metadata.
--- a/accepted/future-releases/abstract-external-fields/feature-specification.md
+++ b/accepted/future-releases/abstract-external-fields/feature-specification.md
@@ -17,16 +17,9 @@ or a library namespace), also allowing statically checked usage of
 implementations that are not known statically; the declarations are
 associated with implementations via an implementation specific mechanism.
 
-However, variables do not seem to be compatible with these modifiers. A
-variable is associated with a storage location and an implicitly induced
-setter and/or getter. This amounts to an implementation, so variables seem
-to be inherently non-abstract and non-external.
-
-However, we may focus on the implicitly induced setter and/or getter for a
-given instance variable declaration, in which case it is simply considered
-to be a more concise notation for said accessors. The point is that the
-setter and/or getter is added to the class interface, and they are intended
-to be implemented by subtypes.
+Previously, Dart supported neither abstract nor external variables of any
+kind. Such constructs are useful if viewed as declaring a pair consisting
+of a setter and a getter.
 
 As a workaround, it is possible to use a concrete instance variable as an
 "abstract instance variable" by simply ignoring the implementation. The
@@ -35,11 +28,13 @@ getter to the interface of the enclosing class (let us call it `C`). Any
 concrete subtypes (`class D implements C ..`) will be required to implement
 said accessors.
 
-However, _subclasses_ of `C` will not be required to implement the
-accessors, they will tacity inherit the implementation of the concrete
-field, which may be a bug. Also, if the subclasses of `C` _do_ implement
-the accessors, the storage reserved for the concrete variable will be a
-space leak (it's simply unused).
+However, if `D` is a _subclass_ of `C` then `D` is not required to
+implement the accessors. Instead, `D` tacity inherits the implementation of
+the concrete field. This may be a bug, because `C` is designed for having
+these accessors overridden to do specific things that the concrete variable
+will not do.  Also, if `D` _does_ override the accessors with a getter and
+setter as intended, the storage reserved for the concrete instance variable
+will be a space leak (it's simply unused).
 
 Finally, if null-safety is enabled then a concrete instance variable
 declaration (say, `int foo;`) is a compile-time error if the type of the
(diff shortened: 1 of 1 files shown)


accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,205 @@
...
+  int get i2;
+  void set i2(int _);
+  dynamic get x;
+  void set x(dynamic _);
+  int get fi;
+  dynamic get fx;
+  num get cn;
+  void cn(covariant num _);
+  dynamic get cx;
+  void set cx(covariant dynamic _);
```
lrhn, line 143: I'd rewrite it to:
```dart
   void set cx(covariant _)
```
and then let inference handle the `var`->`dynamic` conversion.
That is, copy the abstract declaration's "type", but erasing `var` (because it'll need to be removed as return type anyway).

Same everywhere, don't convert to `dynamic` as part of the rewrite. Just rewrite, and then apply inference or other type resolutions afterwards.

accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,205 @@
...
+
+
+## Feature Specification
+
+This section specifies abstract and external variables.
+
+
+### Syntax
+
+The grammar is modified as follows:
```
lrhn, line 83: Need to include `late` as well. This is happening post-NNBD.

accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,205 @@
+# Abstract Fields and External Fields
+
+Authors: lrn@google.com, eernst@google.com<br>
+Version: 1.0
+
+## Background and Motivation
+
+Dart allows abstract instance methods, getters and setters to be declared
+in classes and mixins.  It allows external functions, methods, getters, and
```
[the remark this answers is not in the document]
eernstg: I did that for many years, because various pieces of software wanted it (and Leslie Lamport said so ;-), then dropped it because some other pieces of software seemed to be more happy about a single space.

At this point I noted that the double space is used a lot (in at least 458 locations in the language repo) in `*.md` files, so I thought that the double space would be least disruptive in this context.

(Religious wars may proceed on demand ;-)

accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,205 @@
...
+or a library namespace), also allowing statically checked usage of
+implementations that are not known statically; the declarations are
+associated with implementations via an implementation specific mechanism.
+
+A field is associated with a storage location and an implicitly induced
+setter and/or getter which amounts to an implementation, so fields seem to
+be inherently non-abstract and non-external. However, we can identify the
+field with its accessors:
+
+1. An "abstract instance field" can be a pair of an abstract instance
```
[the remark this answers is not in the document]
eernstg: This is the background and motivation section, and this particular part is about the design and the underlying ideas, so it doesn't actually define anything. I've adjusted the wording to make this more obvious.

accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,205 @@
...
+   getter and an abstract instance setter.  This can be useful, because one
+   declaration replaces two declarations containing some redundant
+   elements: `int get foo;s et foo(int _);`.  Developers could otherwise
```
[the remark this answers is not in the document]
eernstg: Control-t is my friend. Perhaps one more time than intended. Done. ;)

accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,205 @@
...
+   use a concrete field, `int foo;`, but this may cause a waste of space if
+   someone _extends_ the class instead of implementing it, because the
+   intention is to override the getter and setter, and never use the
+   storage.
```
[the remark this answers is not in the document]
eernstg: Indeed, done.

accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,205 @@
...
+
+2. Similarly, an "external field" can be a pair of an external getter and
```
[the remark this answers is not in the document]
eernstg: The same adjustments were made here.

accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,205 @@
...
+   users write Dart classes in order to specify a native memory layout.
+   All instances of the interface will be backed by native (external)
+   code. Currently, `dart:ffi` developers use a regular field declaration,
+   `int foo;`, but what is actually needed is an external setter and/or
+   getter.
+
+An abstract or external field declaration replaces a setter/getter pair or
+a getter.  In the former case it helps avoid redundant specification of the
+name and type, which is more concise and less error-prone.  In the latter
+case the declaration may actually be more verbose (`abstract final int foo;`
```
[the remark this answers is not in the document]
eernstg: Good point. Many elements rewritten.

accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,205 @@
...
+type.
+
+To provide a migration path for these use cases, we introduce _abstract and
+external field declarations_.
+
+
+## Feature Specification
+
+This section specifies abstract and external variables.
+
```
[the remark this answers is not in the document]
eernstg: Done.

accepted/future-releases/abstract-external-fields/feature-specification.md
eernstg: It doesn't make sense to include `late` in a grammar rule for a construct that cannot contain that modifier. Or do you wish to include `late abstract` or `late external` variables after all?

accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,205 @@
...
+are specified in terms of a small program transformation that eliminates
+them.  This fully determines the further static analysis (including errors
+and warnings), and the dynamic semantics.  The transformations are as
+follows:
+
+An abstract instance variable declaration _D_ is treated as an abstract setter
+declaration and/or an abstract getter declaration. The setter is included if
+and only if _D_ is non-final. The return type of the getter and the parameter
+type of the setter, if present, is the type of _D_ (*which may be declared
+explicitly, obtained by override inference, or defaulted to `dynamic`*).  The
```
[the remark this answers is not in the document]
eernstg: > What happens..

I see no reason to introduce new rules for the declarations introduced by desugaring, we should simply use the [existing rules](https://github.com/dart-lang/language/blob/master/resources/type-system/inference.md#instance-field-getter-and-setter-override-inference).

accepted/future-releases/abstract-external-fields/feature-specification.md
eernstg: The reason why I made `dynamic` explicit is that this desugaring step relies on the result of inference. For instance, if the superclass has `var x = 2;` as source code then we need to ensure that inference has been performed on the declaration in order to either make it `int x = 2;` or, for instance `Object? x = 2;`, depending on both the initializer in the declaration itself, and the declarations of `x` in the superinterfaces of the superclass.

I believe that the form that you mention will never exist, so it seems less than helpful to try to invent it, and maybe getting it wrong in some sense..

eernstg:
Review response: Substantial rewrites performed, some issues not resolved. PTAL.

accepted/future-releases/abstract-external-fields/feature-specification.md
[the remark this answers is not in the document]
eernstg: (Deleted the double spaces, btw.)

accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,221 @@
...
+
+### Syntax
+
+The grammar is modified as follows:
+
+```
+<topLevelDefinition> ::=
+  ... |
+  // New alternative.
+  'external' ('final' <type>? | <varOrType>) <identifierList> ';'
```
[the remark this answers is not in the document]
eernstg: I can't see how that could be the case, e.g., `<identifierList>` cannot even derive the empty string, so it doesn't even include `external;`. So I need more hints. ;-)

accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,221 @@
...
+to be implemented by subtypes.
+
+As a workaround, it is possible to use a concrete instance variable as an
+"abstract instance variable" by simply ignoring the implementation. The
+concrete variable declaration will add the implicitly induced setter and/or
+getter to the interface of the enclosing class (let us call it `C`). Any
+concrete subtypes (`class D implements C ..`) will be required to implement
+said accessors.
+
+However, _subclasses_ of `C` will not be required to implement the
```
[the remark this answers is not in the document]
eernstg: I deobfuscated several parts of the sentence.

accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,221 @@
...
+it allows external constructors to be declared in classes.
+
+Abstract declarations add members to a class interface, allowing statically
+checked usage of different implementations provided in subclasses.
+External declarations add members to a namespace (such as a class interface
+or a library namespace), also allowing statically checked usage of
+implementations that are not known statically; the declarations are
+associated with implementations via an implementation specific mechanism.
+
+However, variables do not seem to be compatible with these modifiers. A
```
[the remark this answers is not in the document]
eernstg: Done.

accepted/future-releases/abstract-external-fields/feature-specification.md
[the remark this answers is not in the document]
lrhn: I fixed the source to make the question visible :)

I think the answer is that the current specification doesn't have any `finalVarOrType` production. There is a `varOrType` and a `finalConstVarOrType`, but nothing in between.
I'm not even sure `varOrType` is used anywhere except in the `finalConstVarOrType`.

So, maybe use `finalConstVarOrType` here as well, and say that it's a compile-time error if a declaration is both `const` and `external`.

accepted/future-releases/abstract-external-fields/feature-specification.md
lrhn: ACK, it won't apply to the new alternative because it's mutually exclusive with both `abstract` and `external`.

accepted/future-releases/abstract-external-fields/feature-specification.md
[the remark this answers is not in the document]
lrhn: I am firm believer that the background and motivation should not talk about the *new* thing at all.

First describe the status quo, and why it has problems (Background and Motivation). Preferably briefly, since it can easily be known to the reader already.

Then describe the proposed change. Be precise and specific. This is specification.

Then elaborate on consequences, alternatives and justifications for choices.
If you need to justify the design, do it here, not where it's being specified. That might cause readers to start wondering about alternatives that won't happen anyway.

accepted/future-releases/abstract-external-fields/feature-specification.md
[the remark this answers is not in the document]
eernstg: OK, having seen the whole question it makes sense, and it is true that `<finalVarOrType>` could be used as Leaf said. I introduced that non-terminal when I discovered that it was needed multiple times, but forgot the first occurrence.

Done.

accepted/future-releases/abstract-external-fields/feature-specification.md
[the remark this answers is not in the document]
eernstg: > should not talk about the _new_ thing at all.

That's indeed a nice principle. This particular section was written by you, and I updated it, but the mixture of old and new was there all the time. Anyway, I like the structure you suggest, and I've revised the text to separate the two.

accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,240 @@
...
+an abstract instance variable is inconvenient, error prone, and confusing
+for a reader of the code, and for a writer of a subclass, and similar
+problems exist when a regular variable is used as a replacement for an
+external getter and possibly setter.
+
+
+## Design Idea
+
+In response to the issues above, this document introduces abstract and
+external variables.
```
lrhn, line 81: Maybe: ... and external variable declarations.
Since they don't actually introduce a *variable*.

We're not being consistent about this, so maybe not important.
(Personally, I don't think there is such a thing as an "abstract getter". There is an abstract getter declaration which introduces a getter signature and no getter implementation. There is no "getter" which is abstract in that.)

accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,240 @@
...
+
+The basic idea is that an abstract variable is syntactic sugar for an
+abstract getter and possibly an abstract setter, and similarly for an
+external variable, in both cases such that there is no storage and no
+implementation of the getter and possible setter, and they have the same
+signatures as the ones that would be induced implicitly by a concrete
+variable declaration.
```
lrhn, line 88: Too long sentence. :) Give people a chance to breathe along the way. Maybe:

> The basic idea is that an abstract variable is syntactic sugar for an abstract getter and possibly an abstract setter.
> Similarly, an external variable is syntactic sugar for an external getter and possibly an external setter.
> In both cases there is no storage and no implementation backing the getter and possible setter, 
> they merely have the same interface getter and setter signatures as would be induced by a concrete variable declaration.

accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,240 @@
...
+
+The next section is the normative text that specifies the syntax and
+semantics that realize this idea.
```
lrhn, line 91: that -> which
(it's a descriptive clause, not a restrictive one).

accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,240 @@
...
+
+
+## Feature Specification
+
+An _abstract instance variable declaration_ is an instance variable
+declaration prefixed by the modifier `abstract`. It must not be late, and
```
lrhn, line 97: It --> The declaration

Be careful about using an unqualified "it". The previous sentence contains two noun phrases, and it's not obvious that "it" binds to the latter.

accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,240 @@
...
+it cannot have an initializer expression.
+
+An _external variable declaration_ is a non-local, non-parameter variable
+declaration prefixed by the modifier `external`. It must not be abstract,
+const, or late, and it cannot have an initializer expression.
+
+The syntax and behavior of these constructs is specified in the following
```
lrhn, line 104: is -> are

accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,240 @@
...
+      <identifierList> |
+  // New alternative.
+  'abstract' (<finalVarOrType> | 'covariant' <varOrType>) <identifierList>
+```
+
+
+### Static Analysis
+
+The features specified in this document are syntactic sugar, that is, they
+are specified in terms of a small program transformation that eliminates
```
lrhn, line 135: eliminates them -> reduces them to a different syntactic construct.

(there's still something left afterwards, so it's not *eliminiated*.)

accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,240 @@
...
+  // Desugared meaning of external static variables.
+  external static int get i1;
+  external static void set i1(int _); // ...
+  external static dynamic get fx;
+}
+```
+
+Metadata on an abstract or external variable declaration _D_ is associated
+with both the getter and the setter, if present, that arise by desugaring
+of _D_.
```
lrhn, line 218: Maybe add ", just as for a concrete variable declaration" (so readers won't think this is special for abstract ones).

lrhn:
Late comments :)

accepted/future-releases/abstract-external-fields/feature-specification.md
```diff
@@ -0,0 +1,216 @@
...
+it allows external constructors to be declared in classes.
+
+Abstract declarations add members to a class interface, allowing statically
+checked usage of different implementations provided in subclasses.
+External declarations add members to a namespace (such as a class interface
+or a library namespace), also allowing statically checked usage of
+implementations that are not known statically; the declarations are
+associated with implementations via an implementation specific mechanism.
+
+Previously, Dart supported neither abstract nor external variables of any
```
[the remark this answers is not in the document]
eernstg: Done.

accepted/future-releases/abstract-external-fields/feature-specification.md
eernstg: Done in new PR on branch spec_external_abstract_fields_jun20.

accepted/future-releases/abstract-external-fields/feature-specification.md
eernstg: Yep, done in the new PR.

accepted/future-releases/abstract-external-fields/feature-specification.md
eernstg: Done in new PR.

accepted/future-releases/abstract-external-fields/feature-specification.md
eernstg: Done in new PR.

accepted/future-releases/abstract-external-fields/feature-specification.md
eernstg: Done in new PR.

accepted/future-releases/abstract-external-fields/feature-specification.md
eernstg: Done in new PR.

accepted/future-releases/abstract-external-fields/feature-specification.md
eernstg: But that is actually not true. At least, according to 'dart:mirrors' and reflectable, metadata is associated with concrete declarations, in this case the variable declaration, and implicitly induced getters/setters do not have the same metadata.

But with external/abstract variables there is no such variable, so the metadata needs to be on the getter/setter directly. Luckily, canonicalization implies that it is benign to copy the metadata syntax, because that syntax ends up in the same binding environment and will denote the same objects on the accessors as it did on the variable which was eliminated.

eernstg:
Late fixes in new PR, #1032.

PR #975, opened 2020-05-19
Change method override inference to use combined member signature

eernstg:


resources/type-system/inference.md
```diff
@@ -150,22 +153,27 @@ error if there is such an ordering.  Note that method override inference is
...
+
+A compile-time error occurs if `s` does not exist.  *E.g., one
+superinterface could have signature `void m([int])` and another one could
+have signature `void m(num)`, such that none of them is most specific.
+There may still exist a valid override of both (e.g., `void m([num])`).  In
+this situation `C.m` can be declared with a complete signature, it just
+cannot use override inference.*
+
+If there is no corresponding parameter in `s` for a parameter of the
+declaration of `m` in `C`, it is treated as `dynamic` (*e.g., this occurs
```
eernstg, line 174: Maybe `Object?` rather than `dynamic`?

lrhn: Preferably, if possible.
Means that all existing code which writes nothing and inherits nothing must be migrated to an explicit `dynamic`. That's a good thing.

resources/type-system/inference.md
```diff
@@ -150,22 +153,27 @@ error if there is such an ordering.  Note that method override inference is
...
-is being inferred for the overriding method, it is an error.  If there is no
-corresponding parameter position in the overridden method to infer from and the
-signatures are compatible, it is treated as dynamic (e.g. overriding a one
-parameter method with a method that takes a second optional parameter).  Note:
-if there is no corresponding parameter position in the overridden method to
-infer from and the signatures are incompatible (e.g. overriding a one parameter
-method with a method that takes a second non-optional parameter), the inference
-result is not defined and tools are free to either emit an error, or to defer
-the error to override checking.
+A method `m` of a class `C` which is subject to override inference is
```
leafpetersen, line 159: Maybe "A method `m` of a class `C` is subject to override inference if it"

resources/type-system/inference.md
leafpetersen: I think we probably should land this specifying `dynamic`.  We can discuss a global no implicit dynamic change, but I honestly think it's too much to bite off this late in the game.

resources/type-system/inference.md
```diff
@@ -150,22 +153,27 @@ error if there is such an ordering.  Note that method override inference is
...
+this situation `C.m` can be declared with a complete signature, it just
+cannot use override inference.*
+
+If there is no corresponding parameter in `s` for a parameter of the
+declaration of `m` in `C`, it is treated as `dynamic` (*e.g., this occurs
+when overriding a one parameter method with a method that takes a second
+optional parameter*).
 
 
 ##### Instance field, getter, and setter override inference
```
leafpetersen, line 179: Should this section be updated?  I think it should, right?  There is a notion of "combined member signature" for getters/setters/fields as well, right?

leafpetersen:
What's the status of this?

resources/type-system/inference.md
eernstg: Done.

resources/type-system/inference.md
eernstg: Keeping `dynamic`.

resources/type-system/inference.md
eernstg: Done.

eernstg:
Updated the getters/setters section, PTAL.

Issue #946, opened 2020-04-30
Finalize details of initialization based promotion.


leafpetersen:
The [flow analysis specification](https://github.com/dart-lang/language/blob/master/resources/type-system/flow-analysis.md), based on discussion [here](https://github.com/dart-lang/language/issues/440) includes support for promotion based on initialization assignments.  Specifically: 
  - if a value of type T is assigned to a variable x
  - x is definitely unassigned and not definitely assigned
  - x was declared without an explicit type, and with no initialization expression
  - then x may be promoted to `T` subject to the usual restrictions around capture etc.

In discussion, the language team still feels that this is a valuable feature.  In discussion, several related questions were raised.

1) Is it an error to use such a variable when it is or may not be initialized?
    - Is it an error to use it when it is not definitely assigned?
    - Is it an error to use it when it is definitely unassigned?
2) If 1) is not an error, then what type does the variable have at the pre-initialization use point?
    - dynamic  
    - Object?
    - Null
    - Something else
3) May such a variable be demoted from its promoted type, and if so, to what type is it demoted?
    - dynamic
    - Object?
4) Is it an error to initialize such a variable at two unrelated types?
5) If we choose to make 1) an error, should such an error apply to all variables declared with no initializer, regardless of nullability and type?

Examples:

```dart
void test1And2(bool b) {
   var x;
   if (b) {
     // `x` is definitely unassigned.  Is this an error?  If not, what type do we treat x as having.
     var l = x; 
     x = 3; // Promotes x to int
   }
    // `x` is not definitely unassigned, but it is not definitely assigned.  Is this an error?
   // If not, what type does 'x' have.
   var a = x;
}
void test3(bool b) {
    var x;
    if (b) {
      x = 3; // Promotes x to int
   }
   // What is the type of x here?  Demote to dynamic?  Demote to Object??
  var y;
  y = 3; // y is promoted to int;
  y = "hello";  // Is this legal?  Does it demote y to dynamic?  To Object?
}

void test4(bool b)  {
    var x;
    if (b) {
       x = 3;
       x.isEven; // Ok, x is promoted to int.
    } else { 
       x = "hello";
       x.length;  // Ok, x is promoted to String
   }
   // Is there an error at the join point?  Or do we demote?
   var l = x; // If we demote, to what type do we demote?  dynamic?  Object??
    x.arglbargle;  // Legal if we treat this as demotion to dynamic, do we want this?
}

void test5(bool b) {
    String? x;
    if (b) {
[...]

PR #991, opened 2020-05-28
Initial proposal for initialization based promotion

leafpetersen:
Here's a stake in the ground along the lines I outlined [here](https://github.com/dart-lang/language/issues/946#issuecomment-630593240).   Thoughts, comments?

cc @munificent @eernstg @lrhn @johnniwinther @scheglov @stereotype441

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -638,6 +638,10 @@ It is a compile time error to read a local variable when the variable is
 **definitely unassigned** unless the variable is non-final and has nullable
 type.  This includes variables marked `late` and/or `final`.
 
+It is a compile time error to read a local variable when the variable is not
+**definitely assigned** unless the variable is non-final and has nullable type
+or is `late`.
+
 The erors specified above are summarized in the following table (using `int` as
 an example non-nullable type).  A variable which has an initializer (explicit or
 implicit) is always considered definitely assigned, and is never considered
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -6,6 +6,9 @@ Status: Draft
 
 ## CHANGELOG
 
+2020.05.18
+  - **CHANGE** Changes to definite assignment for local variables without types.
+
 2020.05.14
   - **CHANGE** Strong mode is auto-opted in when the "main" file is opted in.
   - **CHANGE** Specify weak mode/strong mode flag.
@@ -481,25 +484,24 @@ fields on `Object`.
 It is an error to call an expression whose type is potentially nullable and not
 `dynamic`.
 
-It is an error if a top level variable or static variable with a non-nullable type
-has no initializer expression unless the variable is marked with the `late` modifier.
+It is an error if a top level variable or static variable with a non-nullable
+type has no initializer expression unless the variable is marked with the `late`
+modifier.
 
-It is an error if a class declaration declares an instance variable with a potentially
-non-nullable type and no initializer expression, and the class has a generative constructor
-where the variable is not initialized via an initializing formal or an initializer list entry,
-unless the variable is marked with the `late` modifier.
+It is an error if a class declaration declares an instance variable with a
+potentially non-nullable type and no initializer expression, and the class has a
+generative constructor where the variable is not initialized via an initializing
+formal or an initializer list entry, unless the variable is marked with the
+`late` modifier.
 
-It is an error if a mixin declaration or a class declaration with no generative constructors
-declares an instance variable with a potentially non-nullable type and no initializer expression
-unless the variable is marked with the `late` modifier.
+It is an error if a mixin declaration or a class declaration with no generative
+constructors declares an instance variable with a potentially non-nullable type
+and no initializer expression unless the variable is marked with the `late`
+modifier.
 
-It is an error to derive a mixin from a class declaration which contains
-an instance variable with a potentially non-nullable type and no initializer expression
-unless the variable is marked with the `late` modifier.
-
-It is an error if a potentially non-nullable local variable which has no
-initializer expression and is not marked `late` is used before it is definitely
-assigned (see Definite Assignment below).
+It is an error to derive a mixin from a class declaration which contains an
+instance variable with a potentially non-nullable type and no initializer
+expression unless the variable is marked with the `late` modifier.
 
 It is an error if the body of a method, function, getter, or function expression
 with a potentially non-nullable return type **may complete normally**.
@@ -553,18 +555,9 @@ expression.
 It is an error for a class with a `const` constructor to have a `late final`
 instance variable.
 
-It is not a compile time error to write to a `final` variable if that variable
-is declared `late` and does not have an initializer.
-
-It is a compile time error to assign a value to a local variable marked `late`
-and `final` when the variable is **definitely assigned**.  This includes all
-forms of assignments, including assignments via the composite assignment
-operators as well as pre and post-fix operators.
-
-It is a compile time error to read a local variable marked `late` when the
-variable is **definitely unassigned**. This includes all forms of reads,
-including implicit reads via the composite assignment operators as well as pre
-and post-fix operators.
+It is not a compile time error to write to a `final` top-level variable field
+(static or instance) if that variable is declared `late` and does not have an
+initializer.  For local variables, see the section below.
 
 It is an error if the object being iterated over by a `for-in` loop has a static
 type which is not `dynamic`, and is not a subtype of `Iterable<dynamic>`.
@@ -612,6 +605,198 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
 
 It is a warning to use the null check operator (`!`) on an expression of type
 `T` if `T` is **strictly non-nullable** .
+
+### Local variables and definite (un)assignment.
+
+As part of the null safety release, errors for local variables are specified to
+take into account **definite assignment** and **definite unassignment** (see
+the secontion on Definite Assignment below).
+
+In all cases in this section, errors that are described as occurring on reads of
+a variable are intended to apply to all form of reads, including indirectly as
+part of composite assignment operators, as well as via pre and post-fix
+operators.  Similarly, errors that are described as occurring on writes of a
+variable are intended to apply to all form of writes.
+
+It is a compile time error to assign a value to a local variable marked `final`
+when the variable is **definitely assigned**.  This includes variables that are
+also marked `late`.  A variable is always considered definitely assigned if it
+has an explicit initializer, or an implicit initializer as part of a larger
+construct (e.g. the loop variable in a `for in` construct).
+
+It is not a compile time error to write to a `final` local variable if that
+variable is declared `late`, does not have an initializer, and is not
+**definitely assigned**.
+
+It is not a compile time error to write to a `final` local variable if that
+variable is **definitely unassigned** (and hence does not have an initializer).
+
+It is a compile time error to read a local variable when the variable is
+**definitely unassigned** unless the variable is non-final and has nullable
+type, either explicitly provided or inferred directly from the initalizer.  This
+includes variables marked `late` and/or `final`.
+
+The erors specified above are summarized in the following table.  A variable
+which has an initializer (explicit or implicit) is always considered definitely
+assigned, and is never considered definitely unassigned.
+
+
+Read Errors:
+
+| Declaration form  | Def. Assigned | Not Def. Assigned | Def. Unassigned |
+| -------- ---------| ------------- | ----------------- | --------------- |
+| var x;            | Ok            | Error             | Error           |
+| final x;          | Ok            | Error             | Error           |
+| int x;            | Ok            | Error             | Error           |
+| int? x;           | Ok            | Ok                | Ok              |
+| final int x;      | Ok            | Error             | Error           |
+| final int? x;     | Ok            | Error             | Error           |
+| late var x;       | Ok            | Ok                | Error           |
+| late final x;     | Ok            | Ok                | Error           |
+| late T x;         | Ok            | Ok                | Error           |
+| late final T x;   | Ok            | Ok                | Error           |
+
+Write Errors:
+
+| Declaration form  | Def. Assigned | Not Def. UnAssigned | Def. Unassigned |
+| -------- ---------| ------------- | ------------------- | --------------- |
+| var x;            | Ok            | Ok                  | Ok              |
+| final x;          | Error         | Error               | Ok              |
+| int x;            | Ok            | Ok                  | Ok              |
+| int? x;           | Ok            | Ok                  | Ok              |
+| final int x;      | Error         | Error               | Ok              |
+| final int? x;     | Error         | Error               | Ok              |
+| late var x;       | Ok            | Ok                  | Ok              |
+| late final x;     | Error         | Ok                  | Ok              |
+| late T x;         | Ok            | Ok                  | Ok              |
+| late final T x;   | Error         | Ok                  | Ok              |
+
+### Local variables and initialization based inference
+
+Local variables with explicitly written types are given the declared types as
+written.  The declared type of the variable is considered a "type of interest"
+in the sense defined in the flow analysis specification.  If the variable has an
+initializer (explicit or implicit) and is not `final`, then the declaration is
+treated as an assignment for the purposes of promotion.
+
+*Treating the declared type of the variable as a "type of interest" implies that
+if the variable is a nullable type, then its non-nullable version is also a type
+of interest.  Treating the initialization as an assignment for the purposes of
+promotion means that initializing a mutable variable declared at type `T?` with
+a value of non-nullable type `T` immediately promotes the variable to the
+non-nullable type.*
+
+```dart
+void test() {
+  int? x = 3; // x is declared at `int?`
+  x.isEven; // Valid, x has been promoted to `int`
+  x = null; // Valid, demotes to the declared type.
+}
+```
+
+Local variables with no explicitly written type but with an initializer are
+given an inferred declared type equal to the type of their initializer.  In the
+case that the type of the initializer is a promoted type variable `X & T`, the
+inferred type of the variable shall be `X`.  However, such a variable shall be
+treated as immediately promoted to `X & T`.
+
+In the null safety release, local variables with no explicitly written type and
+no initializer are subject to further promotion based inference.  The variables
+under consideration here are those declared as `var x;`, `final x;`, `late var
+x;`, or `late final x;`.  We refer to such variables as **untyped variables**.
+
+An **initializing assignment** is an assignment to an **untyped variable** when
+that variable is in the **definitely unassigned** state.  Such an assignment on
+a path shall cause the variable in question to be treated by the flow analysis
+on that path as if it were declared with the assigned expression as its
+initializing expression, up until the next join point in the program.  We refer
+to this inferred type as the **initialization inferred type** of the variable.
+The **initialization inferred type** of the variable is always treated as a
+**type of interest** for the purposes of promotion (including its non-nullable
+version if applicable).
+
+```dart
+void test() {
+  var x; // x is an untyped variable
+  x = 3; // x is given the initialization inferred type of `int`
+}
+```
+
+At a join point in the program, it is an error if an **untyped variable** has
+been given an **initialization inferred type** on one of the paths to the join
+point and not another, unless the variable is declared as `late`.
+
+```dart
+void test(bool b) {
+  var x; // x is an untyped variable
+  if (b) {
+    x = 3;
+  } // Error, x has not been given an inferred type on all paths
+
+  late var y; // y is a late untyped variable
+  if (b) {
+    y = 3;
+  } // Ok, y has inferred type `int`
+}
+```
+
+At a join point in the program, if an **untyped variable** has been given an
+**initialization inferred type** on more than one of the paths to the join
+point, the variable is treated after the join point as having the
+**initialization inferred type** equal to the upper bound of the
+**initialization inferred types** from the paths into the join point.  The
+**initialization inferred type** so computed is treated as a **type of
+interest** for the variable in question (including its non-nullable version if
+applicable).  The promoted type chains for the variable on the respective paths
+are intersected as usual.
+
+```dart
+void test(bool b) {
+  var x; // x is an untyped variable
+  if (b) {
+    x = 3;
+  } else {
+    x = 3.0;
+  } // x is has type num = UP(int, double)
+
+  late var y; // y is a late untyped variable
+  if (b) {
+    if (b) {
+      y = 3;
+    } else {
+      y = 3.0;
+    } // y has type num
+    y = 3.0; // y is promoted to double, since double is a type of interest
+    y = (3 as num); y is demoted to num.
+  } // y still has type num
+  y = "hello";  // Error, not an initializating write.
+}
+```
+
+It is an error to read an **untyped variable** when it has no **initialization
+inferred type**.
+
+
+```dart
+void test(bool b) {
+  var x; // x is an untyped variable
+  print(x); // Error, no initialization inferred type
+
+  late var y; // y is a late untyped variable
+  var f = () => print(y); // Error, y has no initialization inferred type
+
+
+  var z;
+  do {
+    // z is potentially assigned.
+    if (b) {
+      z = 42;
+    } else {
+      z = 37;
+    }
+  } while (something);
+  print(z); // Error, z has not initialization inferred type
+```
 
 ### Expression typing
 
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -558,9 +558,9 @@ expression.
 It is an error for a class with a `const` constructor to have a `late final`
 instance variable.
 
-It is not a compile time error to write to a `final` top-level variable field
-(static or instance) if that variable is declared `late` and does not have an
-initializer.  For local variables, see the section below.
+It is not a compile time error to write to a `final` non-local variable or field
+if that variable or field is declared `late` and does not have an initializer.
+For local variables, see the section below.
 
 It is an error if the object being iterated over by a `for-in` loop has a static
 type which is not `dynamic`, and is not a subtype of `Iterable<dynamic>`.
@@ -622,32 +622,34 @@ operators.  Similarly, errors that are described as occurring on writes of a
 variable are intended to apply to all form of writes.
 
 It is a compile time error to assign a value to a local variable marked `final`
-when the variable is **definitely assigned**.  This includes variables that are
-also marked `late`.  A variable is always considered definitely assigned if it
-has an explicit initializer, or an implicit initializer as part of a larger
-construct (e.g. the loop variable in a `for in` construct).
+unless the variable is **definitely unassigned**, or the variable is marked
+`late`. Conversely, it is not a compile time error to write to a `final` local
+variable if that variable is **definitely unassigned**.
 
-It is not a compile time error to write to a `final` local variable if that
-variable is declared `late`, does not have an initializer, and is not
-**definitely assigned**.
+It is a compile time error to assign a value to a local variable marked `final`
+and `late` if the variable is **definitely assigned**.  Conversely, it is not a
+compile time error to write to a `final` local variable if that variable is
+declared `late`, and is not **definitely assigned**.
 
-It is not a compile time error to write to a `final` local variable if that
-variable is **definitely unassigned** (and hence does not have an initializer).
+*Note that a variable is always considered definitely assigned and not
+definitely unassigned if it has an explicit initializer, or an implicit
+initializer as part of a larger construct (e.g. the loop variable in a `for in`
+construct).
 
 It is a compile time error to read a local variable when the variable is
 **definitely unassigned** unless the variable is non-final and has nullable
-type, either explicitly provided or inferred directly from the initalizer.  This
-includes variables marked `late` and/or `final`.
+type.  This includes variables marked `late` and/or `final`.
 
-The erors specified above are summarized in the following table.  A variable
-which has an initializer (explicit or implicit) is always considered definitely
-assigned, and is never considered definitely unassigned.
+The erors specified above are summarized in the following table (using `int` as
+an example non-nullable type).  A variable which has an initializer (explicit or
+implicit) is always considered definitely assigned, and is never considered
+definitely unassigned.
 
 
 Read Errors:
 
-| Declaration form  | Def. Assigned | Not Def. Assigned | Def. Unassigned |
-| -------- ---------| ------------- | ----------------- | --------------- |
+| Declaration form  | Def. Assigned | Neither           | Def. Unassigned |
+| ----------------- | ------------- | ----------------- | --------------- |
 | var x;            | Ok            | Error             | Error           |
 | final x;          | Ok            | Error             | Error           |
 | int x;            | Ok            | Error             | Error           |
@@ -656,13 +658,13 @@ Read Errors:
 | final int? x;     | Ok            | Error             | Error           |
 | late var x;       | Ok            | Ok                | Error           |
 | late final x;     | Ok            | Ok                | Error           |
-| late T x;         | Ok            | Ok                | Error           |
-| late final T x;   | Ok            | Ok                | Error           |
+| late int x;       | Ok            | Ok                | Error           |
+| late final int x; | Ok            | Ok                | Error           |
 
 Write Errors:
 
-| Declaration form  | Def. Assigned | Not Def. UnAssigned | Def. Unassigned |
-| -------- ---------| ------------- | ------------------- | --------------- |
+| Declaration form  | Def. Assigned | Neither             | Def. Unassigned |
+| ----------------- | ------------- | ------------------- | --------------- |
 | var x;            | Ok            | Ok                  | Ok              |
 | final x;          | Error         | Error               | Ok              |
 | int x;            | Ok            | Ok                  | Ok              |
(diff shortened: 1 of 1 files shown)


accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -613,6 +606,198 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
 It is a warning to use the null check operator (`!`) on an expression of type
 `T` if `T` is **strictly non-nullable** .
 
+### Local variables and definite (un)assignment.
+
+As part of the null safety release, errors for local variables are specified to
+take into account **definite assignment** and **definite unassignment** (see
+the secontion on Definite Assignment below).
```
johnniwinther, line 613: section

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -613,6 +606,198 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+**initialization inferred type** equal to the upper bound of the
+**initialization inferred types** from the paths into the join point.  The
+**initialization inferred type** so computed is treated as a **type of
+interest** for the variable in question (including its non-nullable version if
+applicable).  The promoted type chains for the variable on the respective paths
+are intersected as usual.
+
+```dart
+void test(bool b) {
+  var x; // x is an untyped variable
```
johnniwinther, line 755: So what is the "declared type" of [x] and [y] (below), i.e. the type we put on the variable declaration for backends to see? Is it `dynamic`?

lrhn: If tools expect all variables to have a declared type, then we might want to give them one, even though this variable actually has *no* declared type. (It might also make things easier for ourselves if we can consistently refer to the declared type of any variable).

It shouldn't matter what the type is, because it should not be exposed in any way. A RHS occurrence of `x` is a compile-time error by itself, so nobody needs to ask about its type.
A LHS occurrence of `x` doesn't need a type. Any type would be consistent.


I'd go for either `void` or `Never`, with `void` being the most consistent (because we can promote it to something else), but `Never` being totally true as well.

leafpetersen: @johnniwinther Good question.  I think there are two answers I can see using.  

I think the best answer is as follows:
  - If we insert an artificial exist node and join all variables on the exit node, then every variable which is given a initialization inferred type is eventually given a maximal initialization inferred type which is a super type of all of its initialization inferred types.  We treat that type as the declared type of the variable, and each initializing assignment is a promotion.
  - For variables which are never given an initialization inferred type (and hence must never be read), we give them the type `Object?` (or `Never`?).

The second best answer is that we just give all such variables the initial type `Object?`.  I think this is inferior for the purposes of error reporting and documentation in the analyzer, not sure how much difference it makes in CFE.

WDYT?

leafpetersen: Actually, thinking further about the above, it does raise a more general question.  Should this program be legal?

```dart
void test() {
  var x;
  if (something) {
    x = 3;
  } else if (somethingelse) {
    x = 3.0;
  } else {
     return;
  }
 print(x);
}
```

That is, should we allow variables to not be initialized on joins from unreachable paths?

lrhn: Yes.

The unreachable path should not have any effect on the join (its not joining into the flow at all). If it does, then we have an issue with our flow analysis.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: Done.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: I have updated this to take reachability into account, and to specify a documentation type for untyped variables.  @johnniwinther @lrhn PTAL

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,245 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+| late final x;     | Error         | Ok                  | Ok              |
+| late T x;         | Ok            | Ok                  | Ok              |
+| late final T x;   | Error         | Ok                  | Ok              |
+
+### Local variables and initialization based inference
+
+Local variables with explicitly written types are given the declared types as
+written.  The declared type of the variable is considered a "type of interest"
+in the sense defined in the flow analysis specification.  If the variable has an
+initializer (explicit or implicit) and is not `final`, then the declaration is
```
leafpetersen, line 682: @munificent @lrhn @eernstg @johnniwinther should we maintain the restriction to final here, or do the promotion for all variables?  @munificent argues that it is more uniform to treat all variables the same here.  My reasoning was that for a final variable, it is useless to declare it as nullable and then assign a non-nullable variable, since you can never re-assign it to null: you should just declare it with the non-nullable version.  But I don't feel that strongly about it.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,245 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+  if (b) {
+    y = 3;
+  } // Ok, y has inferred type `int`
+
+  var z;
+  if (b) {
+     z = 3;
+  } else {
+    return;
+  }  // No error: z has been given a type on all reachable paths into the join
```
lrhn, line 752: Maybe say: 
  z has inferred type `int`
for symmetry with `y`.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,245 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+
+At a join point in the program, if an **untyped variable** has been given an
+**initialization inferred type** on more than one of the reachable paths to the
+join point, the variable is treated after the join point as having the
+**initialization inferred type** equal to the upper bound of the
+**initialization inferred types** from the reachable paths into the join point.
+The **initialization inferred type** so computed is treated as a **type of
+interest** for the variable in question (including its non-nullable version if
+applicable).  The promoted type chains for the variable on the respective paths
+are intersected as usual.
```
lrhn, line 764: I can't read from this what the following program does:
```dart
int? v = ...;
var x;
if (test) {
  x = v;  // initializing assingment, type int?.
  if (x == null) x = 42;
  // x has type int.
} else {
  x = 0;
}
// What is type of x here?

That is, *what are* the chains in question. Is the initialization inferred type part of the chain or not?

File: accepted/future-releases/nnbd/feature-specification.md [hunk 6]
```diff
@@ -616,6 +609,245 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+  } // x is has type num = UP(int, double)
+
+  late var y; // y is a late untyped variable
+  if (b) {
+    if (b) {
+      y = 3;
+    } else {
+      y = 3.0;
+    } // y has type num
+    y = 3.0; // y is promoted to double, since double is a type of interest
```
lrhn, line 782: (We union the types of interest from all branches at a join point?)

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,245 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+    } else {
+      z = 37;
+    }
+  } while (something);
+  print(z); // Error, z has no initialization inferred type
+```
+
+Note that as a consequence of the above it is impossible, in an error-free
+program, to observe the static type or contents of an untyped variable before it
+has been initialized and given an inferred type.  From this standpoint it may be
```
lrhn, line 825: ... unless it's `late` and only potentially assigned. Or is that covered by "error-fee"?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,245 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+
+```dart
+void test(b) {
+  var x; // x is an untyped variable, documented as having type num
+  var z; // z is an untyped variable, documented as having type Never
+  if (b) {
+    x = 3.0; // x is inferred to double
+    return;
+  } else if (b) {
+    return;  // Note, x is has inferred type
```
lrhn, line 843: Comment doesn't pass. I expect "has no"?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,245 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+
+An **initializing assignment** is an assignment to an **untyped variable** when
+that variable is in the **definitely unassigned** state.  Such an assignment on
+a path shall cause the variable in question to be treated by the flow analysis
+on that path as if it were declared with the assigned expression as its
+initializing expression, up until the next join point in the program.  We refer
+to this inferred type as the **initialization inferred type** of the variable.
+The **initialization inferred type** of the variable is always treated as a
+**type of interest** for the purposes of promotion (including its non-nullable
+version if applicable).
```
eernstg, line 719: So
```dart
num n = ...;
late var x;
if (b) { x = 1; x = n; }
```
 is an error (because we can't demote beyond the declared type), but 
```dart
num n = ...; 
late var x; 
if (b) x = 1; 
if (b) x = n;
```
is OK?

The fact that `x` can have many different types in a handful of lines of code, and some assignments will give it a declared type, for a while, and others will give a regular promotion, or a demotion, and so on, makes me think that the notion of initializing assignments may well be counter-productive.

It is very far from "just another syntax for `?:`". I'm tempted to think that when questions come up (like "why does `x` have type `T` here?"), the sane advice will almost always be to eliminate this mechanism to some extent, maybe by initializing `x` immediately, or giving it an explicit type annotation.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -556,18 +558,9 @@ expression.
...
-and `final` when the variable is **definitely assigned**.  This includes all
-forms of assignments, including assignments via the composite assignment
-operators as well as pre and post-fix operators.
-
-It is a compile time error to read a local variable marked `late` when the
-variable is **definitely unassigned**. This includes all forms of reads,
-including implicit reads via the composite assignment operators as well as pre
-and post-fix operators.
+It is not a compile time error to write to a `final` top-level variable field
+(static or instance) if that variable is declared `late` and does not have an
```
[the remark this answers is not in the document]
leafpetersen: Fixed.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,245 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+has an explicit initializer, or an implicit initializer as part of a larger
+construct (e.g. the loop variable in a `for in` construct).
+
+It is not a compile time error to write to a `final` local variable if that
+variable is declared `late`, does not have an initializer, and is not
+**definitely assigned**.
+
+It is not a compile time error to write to a `final` local variable if that
+variable is **definitely unassigned** (and hence does not have an initializer).
+
```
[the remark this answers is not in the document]
leafpetersen: It is already specified that writing to a `final` variable is a compile time error, with certain exceptions (e.g. there is setter with the right name in scope).  This spec was intended to add to the list of exceptions, without removing the original error.  I can make this explicit though.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,245 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+It is a compile time error to read a local variable when the variable is
+**definitely unassigned** unless the variable is non-final and has nullable
+type, either explicitly provided or inferred directly from the initalizer.  This
```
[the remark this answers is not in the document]
leafpetersen: Deleted the offending phrase.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,245 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+includes variables marked `late` and/or `final`.
+
+The erors specified above are summarized in the following table.  A variable
+which has an initializer (explicit or implicit) is always considered definitely
+assigned, and is never considered definitely unassigned.
```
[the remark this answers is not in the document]
leafpetersen: Ack.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,245 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+
+
+Read Errors:
+
+| Declaration form  | Def. Assigned | Not Def. Assigned | Def. Unassigned |
```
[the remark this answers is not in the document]
leafpetersen: Changed to `Neither`.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,245 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+| -------- ---------| ------------- | ----------------- | --------------- |
+| var x;            | Ok            | Error             | Error           |
+| final x;          | Ok            | Error             | Error           |
+| int x;            | Ok            | Error             | Error           |
+| int? x;           | Ok            | Ok                | Ok              |
+| final int x;      | Ok            | Error             | Error           |
+| final int? x;     | Ok            | Error             | Error           |
+| late var x;       | Ok            | Ok                | Error           |
+| late final x;     | Ok            | Ok                | Error           |
+| late T x;         | Ok            | Ok                | Error           |
```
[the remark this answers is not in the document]
leafpetersen: should have been `int`.  Fixed, and noted that `int` is just an example.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,245 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+| late final T x;   | Ok            | Ok                | Error           |
+
+Write Errors:
+
+| Declaration form  | Def. Assigned | Not Def. UnAssigned | Def. Unassigned |
+| -------- ---------| ------------- | ------------------- | --------------- |
```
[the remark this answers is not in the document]
leafpetersen: Fixed.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,245 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+}
+```
+
+Local variables with no explicitly written type but with an initializer are
+given an inferred declared type equal to the type of their initializer.  In the
+case that the type of the initializer is a promoted type variable `X & T`, the
+inferred type of the variable shall be `X`.  However, such a variable shall be
+treated as immediately promoted to `X & T`.
+
+In the null safety release, local variables with no explicitly written type and
```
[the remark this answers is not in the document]
leafpetersen: Eliminated the throat clearing.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: Maybe.  I don't know how to answer that question without experience with it.  I agree that the example that you give above is a bit weird.  But it's at the intersection of a number of features: a late untyped variable, that is re-assigned when it is in a potentially assigned state.  I'm not willing to throw out all of assignment based promotion because `late` behaves a little oddly in some cases.  In the non-late case, we don't have this anomaly.  We could consider doing something different with late, but so var, this is the best approach that I've found.

leafpetersen: > I'm tempted to think that when questions come up (like "why does `x` have type `T` here?"), the sane advice will almost always be to eliminate this mechanism to some extent, maybe by initializing `x` immediately, or giving it an explicit type annotation.

Specifically to this, I'm not sure I agree.  The sane advice will be to say "because inference, is it causing a problem?", and as far as I can tell the answer to the latter will almost always be "no".  When do you see the answer being "yes"?

What I have proposed has the nice property that it either tells you it can't infer a type, or it infers a type that I think will usually be what you want.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: Not sure I agree - the comment is emphasizing the salient point of the example.  I'll rephrase to be closer, but I don't want to lose the point.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: As specified here, the IIT is not part of the chain, and the type of `x` after the final join in your example will be `int?`.  It is probably reasonable to add the IIT to the chain at some point, albeit a bit weird (usually we have the elements of the chain are proper subtypes of the declared type).  Note that this won't fix all of the examples that you were interested in.  The final IIT for this example will still be `int?`, it's just that it will have been promoted to `null` after the join.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: yes.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: I'm using initialized loosely here. I'll make it more precise.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: yes, fixed.

accepted/future-releases/nnbd/feature-specification.md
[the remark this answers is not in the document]
lrhn: Exceptions should generally be very precise about what they are exceptions to. Otherwise readers might consider them exceptions to other things as well.

I worry someone might read an "It is not a compile time error to write to a `final` local variable if that variable is **definitely unassigned**" to mean that it's also not an error if the types are wrong.

It does take some amount of disinginuity to get there, but there's always one :)

That's why I want to be specific about which compile-time error it isn't, and why I prefer not to have negative definitions at all, so exceptions are written right at the point that they are exceptions to. (So don't say "A is invalid" in one place and then say "A is valid anyway in some cases" in another place, someone is bound to miss the exception).

> It's a compile-time error to write to a local final non-late variable that is not definitely unassigned.
> It's a compile-time error to write to a local final late variable that is definitely assigned.

(Those are the only variables you write to directly outside of initializer lists, all other "variables" are getters and setters, and there it's a compile-time error if there is no setter in scope).


All in all, this is probably fine, as long as the reader is not trying to misunderstand it.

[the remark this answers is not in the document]
leafpetersen: @lrhn agreed, my intent is to be clear.  I did rewrite this based on your original comment, and I'm not sure if your last comment was intended to apply to the changed version or the original.  Would you mind taking a look at the current version and letting me know where you still see things that might be unclear?  I will do the same, but I may be too close to the proposal to do so effectively.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: > the sane advice will almost always be to eliminate this mechanism to some extent, maybe by initializing `x` immediately, or giving it an explicit type annotation.

One more comment.  What you're describing is the sane advice *now*.  That is, if you write `var x;` now, you get no static checking at all.  So if your worst case scenario is true (I don't think it is) and out advice becomes "don't use var x;", then we are in no worse of a position from a language standpoint than we are now.

That said, I can see this as an argument for my original proposal, which did not use LUB, and which consequently had the property that each of these variables took on at most one type.  @lrhn @munificent thoughts on this?  I can write up a variant that does not use LUB to see what it would look like concretely.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: I have uploaded a patch specifying this, take a look.

accepted/future-releases/nnbd/feature-specification.md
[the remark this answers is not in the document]
lrhn: The new wording is clearer and seems to cover everything.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,267 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+| final x;          | Ok            | Error             | Error           |
+| int x;            | Ok            | Error             | Error           |
+| int? x;           | Ok            | Ok                | Ok              |
+| final int x;      | Ok            | Error             | Error           |
+| final int? x;     | Ok            | Error             | Error           |
+| late var x;       | Ok            | Ok                | Error           |
+| late final x;     | Ok            | Ok                | Error           |
+| late int x;       | Ok            | Ok                | Error           |
+| late final int x; | Ok            | Ok                | Error           |
+
```
lrhn, line 663: Should we cover potentially non-nulllable types as well? It is it sufficient to say that they are treated exactly the same as non-nullable types in this regard?

accepted/future-releases/nnbd/feature-specification.md
lrhn:
```dart
num n = ...; 
late var x; 
if (b) x = 1; 
if (b) x = n;
```
is actually not OK.
You avoid the error at the join-point after the first `if` because the variable is late, but at the second assignment to `x`, it is no longer definitely unassigned, so it is not a initializing assignment.
The type of `x` at that point is `int`, so the `num` assignment is an error.

accepted/future-releases/nnbd/feature-specification.md
lrhn: LGTM.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -484,25 +487,24 @@ fields on `Object`.
...
+modifier.
 
-It is an error if a mixin declaration or a class declaration with no generative constructors
-declares an instance variable with a potentially non-nullable type and no initializer expression
-unless the variable is marked with the `late` modifier.
+It is an error if a class declaration declares an instance variable with a
+potentially non-nullable type and no initializer expression, and the class has a
+generative constructor where the variable is not initialized via an initializing
+formal or an initializer list entry, unless the variable is marked with the
+`late` modifier.
```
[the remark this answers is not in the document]
leafpetersen: This is a whitespace only change to text you added in https://github.com/dart-lang/language/pull/811 .  Maybe follow up separately if we want changes here?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -484,25 +487,24 @@ fields on `Object`.
...
 `dynamic`.
 
-It is an error if a top level variable or static variable with a non-nullable type
-has no initializer expression unless the variable is marked with the `late` modifier.
-
-It is an error if a class declaration declares an instance variable with a potentially
-non-nullable type and no initializer expression, and the class has a generative constructor
-where the variable is not initialized via an initializing formal or an initializer list entry,
-unless the variable is marked with the `late` modifier.
+It is an error if a top level variable or static variable with a non-nullable
```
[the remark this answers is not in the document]
leafpetersen: Done.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -484,25 +487,24 @@ fields on `Object`.
...
+constructors declares an instance variable with a potentially non-nullable type
+and no initializer expression unless the variable is marked with the `late`
+modifier.
 
-It is an error if a potentially non-nullable local variable which has no
-initializer expression and is not marked `late` is used before it is definitely
-assigned (see Definite Assignment below).
+It is an error to derive a mixin from a class declaration which contains an
+instance variable with a potentially non-nullable type and no initializer
+expression unless the variable is marked with the `late` modifier.
```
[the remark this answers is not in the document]
leafpetersen: Again this is text you added here https://github.com/dart-lang/language/pull/811 .  Can you follow up separately, I don't really have any context on that pull request and would rather not get side tracked in this PR.  Sorry for the noise in this PR - I wrapped some overlong lines and github is completely butchering the diff.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,267 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+the section on Definite Assignment below).
+
+In all cases in this section, errors that are described as occurring on reads of
+a variable are intended to apply to all form of reads, including indirectly as
+part of composite assignment operators, as well as via pre and post-fix
+operators.  Similarly, errors that are described as occurring on writes of a
+variable are intended to apply to all form of writes.
+
+It is a compile time error to assign a value to a local variable marked `final`
+unless the variable is **definitely unassigned**, or the variable is marked
```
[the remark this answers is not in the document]
leafpetersen: Done.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,267 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+`late`. Conversely, it is not a compile time error to write to a `final` local
+variable if that variable is **definitely unassigned**.
```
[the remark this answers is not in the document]
leafpetersen: Done.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,267 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+
+It is a compile time error to assign a value to a local variable marked `final`
+and `late` if the variable is **definitely assigned**.  Conversely, it is not a
+compile time error to write to a `final` local variable if that variable is
+declared `late`, and is not **definitely assigned**.
```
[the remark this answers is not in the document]
leafpetersen: Done.

accepted/future-releases/nnbd/feature-specification.md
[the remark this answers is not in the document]
leafpetersen: Great, thanks!

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: @eernstg 

> How about `Null`? Do we actually want `int? i = null;` to promote `i` to `Null` immediately, and even `int? i;`? It's kind of nice that we prevent method invocations that would fail because the value is definitely null, but I suspect that it would be impractical in many other ways, e.g., because `int` isn't a subtype of `Null` so we can't proceed to promote to `int`.

I'm not sure what you're asking here.  As currently specified and implemented, if `int?` is a type of interest, then `int` is also type of interest, but `Null` is not.  Are you proposing that we make it so?

> So it does make sense to promote the type of a final variable even though it doesn't get a new value. So I'd prefer to include final variables here.

I'm don't really see the connection between these two.  The argument against promoting the final variable on initialization is not that it can't be promoted, but that it can never be demoted, so declaring the looser type is pointless.




> For the point about being nullable: The static type of the initializing expression (or of the right hand side of an initializing assignment) may be nullable, and it still makes sense to test and detect that the value is non-null, even though it can't get a new value.

I don't understand what you're trying to say here, can you give an example?

leafpetersen: > Maybe define 'typed' and 'untyped' variables first, to make it a bit more explicit that we have text about the former first, and further down we'll have text about the latter—maybe we could even have a subsubsubtitle saying "Typed Variables" and "Untyped Variables".

I'm not convinced this will make things clearer.  Defining what a "typed variable" is explicitly is a bit painful, unless define it as "not an untyped variable".  The latter is clear and precise, but requires untyped variables to be introduced first.  Given that I don't actually need a term for "typed variable", I'm not sure why it helps things to introduce it as a concept.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: @lrhn Ah, right, good catch.

accepted/future-releases/nnbd/feature-specification.md
eernstg: >> How about `Null`? ..

> I'm not sure what you're asking here. As currently specified and
> implemented, if `int?` is a type of interest, then `int` is also type of
> interest, but `Null` is not. Are you proposing that we make it so?

My reading of the following suggests that the variable should be promoted to have type `Null` immediately at the declaration:

> If the variable has an initializer (explicit or implicit) and is not `final`,
> then the declaration is treated as an assignment for the purposes
> of promotion.

This would occur even for `int? x;`, where the initializer is an implicit `null`. But are you saying that `Null` is not a type of interest even in the case where the variable has been promoted to have type `Null`?:

>  if `int?` is a type of interest, then `int` is also type of interest, but `Null` is not.

In any case, my point was that the variable was promoted to have type `Null`, not whether `Null` is a type of interest. I was just noting that, apparently, variables of nullable type with no non-null initializer would have type `Null` until they are demoted to some other type. 

This might actually be fine, but it seemed somewhat surprising to me. For instance, it could be confusing for developers if they get "`Null` does not have a member `isEven`" rather than "`isEven` cannot be invoked when the receiver has the nullable type `int?`".

----

> declaring the looser type is pointless.

Right, I was just thinking that "pointless" is usually serious enough for a lint, but not for giving a situation a special treatment in the language. So I thought it would make sense to perform the promotion of a final variable based on its initializer (that is: remove the exception for `final`), and then leave it to a lint to tell the developer that the declared type is too loose and will basically be ignored.

>> For the point about being nullable ..

Same topic, no need to go deeper with that.

----

>> Maybe define 'typed' and 'untyped' variables first

This was just a readability issue. The text talks about typed variables for 3 paragraphs and a code snippet, and then about untyped variables basically to the end of the section (which is more like multiple "pages"). I thought it would be helpful to make this overall structure visible in the presentation, rather than tucking away the definition of 'untyped variable' at the end of the 5th paragraph. There's no need to make 'typed variable' anything other than "not 'untyped variable'". The section could start with this:

```
An **untyped variable** is a variable declared as `var x;`, `final x;`, `late var x;`,
or `late final x;`, and a **typed variable** is a variable which is not untyped.
We discuss typed variables first, and then untyped ones.
```

accepted/future-releases/nnbd/feature-specification.md
eernstg: > so it is not a initializing assignment.

Sure, I constructed the example such that `x = 1` is an initializing assignment and `x = n` is not.

But I got the computation of the pseudo-declared type of `x` wrong: `int` is only valid until the join, but at the join we compute a new pseudo-declared type for `x` (we have the LUB and the "must be the same type" models), and non-assigning branches (which are allowed for a late variable) do not contribute, so it's still `int` when we reach the 2nd `if`.

For instance, if the first `if` is 50 lines of code, and at some point in time it doesn't assign to `x`; then we modify the code and add `x = 1;` somewhere in the first `if`; and then the assignment `x = n` in the second `if` is suddenly treated very differently (because it is no longer an initializing assignment, and in this case it becomes an error).

Do you seriously think that there is any other sane advice for practical Dart development than "go back and write the type of `x` explicitly!"?

----

> That said, I can see this as an argument for my original proposal, which did not use LUB

I think that's at least somewhat more maintainable, and I find it safer to avoid using LUB as far as possible in any case.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: > > > How about `Null`? ..
> 
> > I'm not sure what you're asking here. As currently specified and
> > implemented, if `int?` is a type of interest, then `int` is also type of
> > interest, but `Null` is not. Are you proposing that we make it so?
> 
> My reading of the following suggests that the variable should be promoted to have type `Null` immediately at the declaration:

@eernstg I may be missing something, but I'm not following you at all.  Let me try to walk through this as I understand it, and you can tell me where you're seeing something different.

> 
> > If the variable has an initializer (explicit or implicit) and is not `final`,
> > then the declaration is treated as an assignment for the purposes
> > of promotion.
> 
> This would occur even for `int? x;`, where the initializer is an implicit `null`. But are you saying that `Null` is not a type of interest even in the case where the variable has been promoted to have type `Null`?:

I don't understand where you're getting to the point that the variable has been promoted to `Null`.  We do not promote on all assignments - only assignments where the RHS has a type of interest.

> 
> > if `int?` is a type of interest, then `int` is also type of interest, but `Null` is not.
> 
> In any case, my point was that the variable was promoted to have type `Null`, not whether `Null` is a type of interest. I was just noting that, apparently, variables of nullable type with no non-null initializer would have type `Null` until they are demoted to some other type.

Right.  So where did the variable get promoted?  The sentences you quite imply that `int? x = null` makes `int` and `int?` a type of interest, and treats `int? x = null` as if it were `int? x; x = null` for the purposes of promotion.  The assignment `x = null` causes the question to be asked "is `Null` a type of interest".  The answer is "no", so no promotion happens.  So again, where are you seeing a promotion to `Null` happening?

accepted/future-releases/nnbd/feature-specification.md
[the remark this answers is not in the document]
lrhn: ACK

accepted/future-releases/nnbd/feature-specification.md
eernstg: Right, I thought that `int? x = null` would make `= null` an initializing assignment, but I can see that this is explicitly restricted to untyped valiables. So it's a regular assignment after all, so I agree that it won't promote.

This leaves only one topic alive in this thread: 

> should we maintain the restriction to final

accepted/future-releases/nnbd/feature-specification.md
eernstg: The main point I was making is that there is a substantial complexity cost that we're imposing on developers, and I think that's been covered in various ways now. So I'll resolve this thread.

leafpetersen:
A different proposal was accepted and landed.

PR #997, opened 2020-05-29
Adjust futureValueType{,Schema} to propagate `dynamic`

eernstg:
Here is yet another twist with the static checks on return statements: The new rules are treating the return type `dynamic` (along with `dynamic?` and `dynamic*`) the same as `Object?`, which causes a substantial amount of breakage. For instance:

```dart
f() async {
  return; // Error!
}
```

I think we do not want this breakage, so this PR adds one extra alternative to the definition of `futureValueType` and `futureValueTypeSchema` such that an `async` function returning `dynamic` (and its variants) will have future value type `dynamic`.

Konstantin has implemented this addition, and it removes the breakage as expected.

So do you agree that we should do this, @lrhn, @leafpetersen, @munificent?

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -298,12 +298,12 @@ The function **futureValueType** is defined as follows:
 - **futureValueType**(`Future<S>`) = `S`, for all `S`.
 - **futureValueType**(`FutureOr<S>`) = `S`, for all `S`.
 - **futureValueType**(`void`) = `void`.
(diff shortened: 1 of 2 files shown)


lrhn:
Agree. We should not be changing things to be breaking unless necessary.
I expected the "future value type" of `void` to be `void` and of `dynamic` to be `dynamic.
It's only `Object` and `Object?` which are both `Object?`.

eernstg:
Cf. https://github.com/dart-lang/sdk/issues/42133, https://github.com/dart-lang/sdk/issues/42134.

PR #1000, opened 2020-05-30
Simple definite assignment

leafpetersen:
This is an alternative approach to definite assignment based promotion derived as a simplification of the definite assignment promotion proposal [here](https://github.com/dart-lang/language/pull/991).  It eliminates the use of upper bound at join points, thereby enforcing the property that an untyped variable is never assigned more than one inferred type, except on paths that never join.

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -638,6 +638,10 @@ It is a compile time error to read a local variable when the variable is
 **definitely unassigned** unless the variable is non-final and has nullable
 type.  This includes variables marked `late` and/or `final`.
 
+It is a compile time error to read a local variable when the variable is not
+**definitely assigned** unless the variable is non-final and has nullable type
+or is `late`.
+
 The erors specified above are summarized in the following table (using `int` as
 an example non-nullable type).  A variable which has an initializer (explicit or
 implicit) is always considered definitely assigned, and is never considered
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -6,6 +6,9 @@ Status: Draft
 
 ## CHANGELOG
 
+2020.05.18
+  - **CHANGE** Changes to definite assignment for local variables without types.
+
 2020.05.14
   - **CHANGE** Strong mode is auto-opted in when the "main" file is opted in.
   - **CHANGE** Specify weak mode/strong mode flag.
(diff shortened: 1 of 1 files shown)


accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,258 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+given an inferred type.  From this standpoint it may be thought of as having no
+type.  However, IDEs and tools may wish to report a static type for such a
+variable for the purposes of documentation and error reporting.  For such
+purposes, tools shall treat an untyped variable as if it were declared with a
+type equal to the upper bound of all of the **initialization inferred types**
+that it is ascribed in the method, plus `Never`.  The latter implies that a
+variable which is ascribed no **initialization inferred types** on any path
+shall be treated as having declared type `Never`.  Note that despite the
+requirement that the **initialization inferred types** be equal at join points,
+it is possible for a variable to take on different **initialization inferred
```
leafpetersen, line 845: We could alternatively specify it to be an error if a variable ever takes on different initialization inferred types.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,258 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+```dart
+void test() {
+  int? x = 3; // x is declared at `int?`
+  x.isEven; // Valid, x has been promoted to `int`
+  x = null; // Valid, demotes to the declared type.
+}
+```
+
+Local variables with no explicitly written type but with an initializer are
+given an inferred declared type equal to the type of their initializer.  In the
```
scheglov, line 703: Is this "inferred declared type" considered a "type of interest"?

From the description of **initializing assignment**, where we say that it is "treated ... as declared with the assigned expression as its initializing expression", and then we say that it is a **type of interest**, it seems that the answer is "yes". But I'd like to know for sure.

accepted/future-releases/nnbd/feature-specification.md
lrhn: This would still allow a single type to be used, but anything with subtypes needs to be declared. Initializing to `0` and `1.0` would be invalid, not `null` (and avoid the LUB).
If you initialize with two different Widget subclasses, you need to say Widget on the variable.

It does mean that `var x; if (test) { x = e1; } else {x = e2;}` doesn't work like `var x = test ? e1 : e2;`, but there are so many ways it can not do that, and I'm not particularly fond of our LUB to begin with.

It could probably work. And it's a safe initial  behavior, we can always improve later if we make it an error now.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,258 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+  } else {
+    return;
+  }  // Ok, z has inferred type `int` even though it is unassigned on one path.
+}
+```
+
+At a join point in the program, if an **untyped variable** has been given an
+**initialization inferred type** on more than one of the reachable paths to the
+join point and all of such **initialization inferred types** are syntactically
+equal, the variable is treated after the join point as having that
```
lrhn, line 761: Syntactically equal before or after normalizaion?

lrhn:
Agree. Let's do this.

leafpetersen:
Closing this, we took a different approach.

PR #1003, opened 2020-06-03
Batched small updates

leafpetersen:
Several small updates to questions and clarifications that have been resolved in the issue tracker but not moved to the spec.

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -1228,12 +1228,12 @@ appear in reified positions.
 ### Typedefs defined in legacy libraries used in opted-in libraries
 
 A typedef which is define in a legacy library and used in an opted-in library is
-treated defining a function type all of the components of which are treated as
-legacy, but which is treated as non-nullable at the top level.  Hence given the
-following program, it is an error to assign a nullable value to a variable of
-type `F` in an opted-in library, but any function which is compatible with a
-legacy function of type `int* Function(int*)` may be assigned to such a
-variable.
+treated as defining a function type, all of the components of which are
+legacy. The function type itself is treated as non-nullable (and not legacy) at
+the top level.  Hence given the following program, it is an error to assign a
+nullable value to a variable of type `F` in an opted-in library, but any
+function which is compatible with a legacy function of type `int*
+Function(int*)` may be assigned to such a variable.
 
 ```dart
 // Opted-out library "opted_out.dart".
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -6,6 +6,11 @@ Status: Draft
 
 ## CHANGELOG
 
+2020..06.02
+  - Fix the diff to the spec for potentially constant instance checks
+  - Specify that extensions do not apply to values of type `Never`
+  - Specify the treatment of typedefs from legacy libraries
+
 2020.05.20
   - Turn new references to `CastError` into being dynamic type errors.
 
@@ -702,8 +707,12 @@ class G<T> {
 
 For the purposes of extension method resolution, there is no special treatment
 of nullable types with respect to what members are considered accessible.  That
-is, the only members of a nullable tyhpe that are considered accessible
+is, the only members of a nullable type that are considered accessible
 (and hence which take precedence over extensions) are the members on `Object`.
+
+For the purposes of extension method resolution, the type `Never` is considered
+to implement all members, and hence no extension may apply to an expression of
+type `Never`.
 
 ### Assignability
 
@@ -756,7 +765,8 @@ We change the following specification text:
 to
 
 ```
-\item An expression of the form \code{$e$\,\,as\,\,$T$} is potentially constant
+\item An expression of the form \code{$e$\,\,as\,\,$T$} or
+  \code{$e$\,\,is\,\,$T$} is potentially constant
   if $e$ is a potentially constant expression
   and $T$ is a potentially constant type expression,
   and it is further constant if $e$ is constant.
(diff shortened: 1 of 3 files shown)


resources/type-system/flow-analysis.md
```diff
@@ -432,36 +435,22 @@ Definitions:
...
+  - Inference for `E` is done as usual, using `S` as the downwards typing
+    context.  All reference to `x` within `E` are treated as having type `S` as
+    usual.
+  - Let `T` be the resulting inferred type of `E`.
+  - If `T` is assignable to `S` then no further work is required.
+  - Otherwise, the assignment is treated as an assignment of an expression of
+    type `T` to a variable of type `S`, that is:
+    - If the demotion policy applies, treat the assignment as a demoting
+      assignment which demotes `x` to `T`.
+    - Otherwise, it is an error, unless `T` is dynamic, in which case it is an
```
lrhn, line 452: `dynamic` is assignable to `S`, so it wouldn't reach here.

What type is the implicit downcast to? I'm assuming `S`.

resources/type-system/flow-analysis.md
```diff
@@ -432,36 +435,22 @@ Definitions:
...
+      implicit downcast.
```
lrhn, line 453: Would it rather be:

> * If `T` is a subtype of `S`, then if `T` is a type of interest for `x`, `x` is promoted to `T`, 
>   and if it is not a type of interest, nothing further happens to the type of `x`. *(This includes `T` being `S`.)*
> * If the demotion policy applies, treat the assignment as a demoting assignment which demotes `x` to `T`.
> * Otherwise if `T` is `dynamic`, then perform an implicit downcast from `T` to `S` on the value of `E`.
> * Otherwise the assignment is a compile-time error.


(I assume that context-type based conversions like callable-object tear-off has happened during inference of `E`).

resources/type-system/inference.md
```diff
@@ -212,15 +216,22 @@ unlike normal fields, the initializer for a `late` field may reference `this`.
 
 Function literals which are inferred in an empty typing context (see below) are
 inferred using the declared type for all of their parameters.  If a parameter
-has no declared type, it is treated as if it was declared with type `dynamic`.
+has no declared type, it is treated as if it was declared with type `dynamic.
```
lrhn, line 219: Missing a final \` after `` `dynamic``

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1215,6 +1225,31 @@ all occurrences of `S*` in `T` shall be replaced with `S`.  As a result, legacy
 types will never appear as type annotations in opted-in libraries, nor will they
 appear in reified positions.
 
+### Typedefs defined in legacy libraries used in opted-in libraries
+
+A typedef which is define in a legacy library and used in an opted-in library is
+treated defining a function type all of the components of which are treated as
```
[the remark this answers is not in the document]
leafpetersen: Done.

resources/type-system/flow-analysis.md
```diff
@@ -432,36 +435,22 @@ Definitions:
...
+#### Interactions between downwards inference and promotion
+
+Given an assignment (or composite assignment) `x = E` where `x` has current type
+`S` (possibly the result of promotion), inference and promotion interact as
+follows.
+  - Inference for `E` is done as usual, using `S` as the downwards typing
+    context.  All reference to `x` within `E` are treated as having type `S` as
+    usual.
+  - Let `T` be the resulting inferred type of `E`.
+  - If `T` is assignable to `S` then no further work is required.
```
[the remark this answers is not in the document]
leafpetersen: Simplified this to just say that the usual promotion/demotion/error rules for an assignment of an expression of type `T` to a variable of current type `S` apply.

resources/type-system/flow-analysis.md
leafpetersen: True, and yes.

resources/type-system/flow-analysis.md
leafpetersen: Rather than re-specify all of the promotion/demotion policies, I've just changed this to say "at this point, treat it like any other assignment of something of type `T` to a variable of type `S`."  I don't think we need any further text here do we?  The point is just that we use the current type of `x` for downwards inference, but still consider the assignment as a potential promotion/demotion point.

resources/type-system/inference.md
leafpetersen: Done.

leafpetersen:
Ok I addressed comments, landing this.  I'll follow up if more comments come in on the revised text.

cc @scheglov @johnniwinther for visibility.  I believe all of the changes here are things we already clarified in issues and are most likely already implemented.  I am adding some quick tests to verify.

leafpetersen:
[Tests](https://dart-review.googlesource.com/c/sdk/+/150160) out for review.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -6,6 +6,11 @@ Status: Draft
 
 ## CHANGELOG
 
+2020..06.02
```
yzhong52, line 9: FYI: double dots here.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1215,6 +1225,31 @@ all occurrences of `S*` in `T` shall be replaced with `S`.  As a result, legacy
 types will never appear as type annotations in opted-in libraries, nor will they
 appear in reified positions.
 
+### Typedefs defined in legacy libraries used in opted-in libraries
+
+A typedef which is define in a legacy library and used in an opted-in library is
```
johnniwinther, line 1230: defined

accepted/future-releases/nnbd/feature-specification.md
eernstg: Fixed in 8c3bf5e2ba6b6f38dbf66fd7a39e2f84a3463eba.

PR #1009, opened 2020-06-06
Nullable definite assignment

leafpetersen:
This a variant of the simple definite assignment proposal which requires all initializing assignment to assign values of the same type except that `Null` is also allowed, in which case the variable is made nullable.  It also allows variables declared `var` to not be assigned on every path, in which case the type is also made nullable.

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -639,20 +639,22 @@ It is a compile time error to read a local variable when the variable is
 type.  This includes variables marked `late` and/or `final`.
 
 It is a compile time error to read a local variable when the variable is not
-**definitely assigned** unless the variable is non-final and has nullable type
-or is `late`.
+**definitely assigned** unless the variable is non-final and has nullable type,
+or is `late`, or is declared using `var` with no type and no initializer.
 
 The erors specified above are summarized in the following table (using `int` as
 an example non-nullable type).  A variable which has an initializer (explicit or
 implicit) is always considered definitely assigned, and is never considered
-definitely unassigned.
+definitely unassigned.  Note that the `var x` form may only be read in the "not
+definitely assigned" state when it has been subject to initialization promotion
+as described below.
 
 
 Read Errors:
 
 | Declaration form  | Def. Assigned | Neither           | Def. Unassigned |
 | ----------------- | ------------- | ----------------- | --------------- |
-| var x;            | Ok            | Error             | Error           |
+| var x;            | Ok            | Ok/Error          | Error           |
 | final x;          | Ok            | Error             | Error           |
 | int x;            | Ok            | Error             | Error           |
 | int? x;           | Ok            | Ok                | Ok              |
@@ -707,6 +709,8 @@ case that the type of the initializer is a promoted type variable `X & T`, the
 inferred type of the variable shall be `X`.  However, such a variable shall be
 treated as immediately promoted to `X & T`.
 
+#### Untyped variables
+
 Local variables with no explicitly written type and no initializer are subject
 to further promotion based inference.  The variables under consideration here
 are those declared as `var x;`, `final x;`, `late var x;`, or `late final x;`.
@@ -732,13 +736,13 @@ void test() {
 At a join point in the program, it is an error if an **untyped variable** has
 been given an **initialization inferred type** on one of the reachable paths to
 the join point and not on another reachable path, unless the variable is
-declared as `late`.  Note that it is valid for a variable to have no
-**initialization inferred type** on a path into a join point which is known to
-be unreachable.
+declared as `late`, or using `var`.  Note that it is valid for a variable to
+have no **initialization inferred type** on a path into a join point when the
+join point is known to be unreachable along that path.
 
 ```dart
 void test(bool b) {
-  var x; // x is an untyped variable
+  final x; // x is an untyped variable
   if (b) {
     x = 3;
   } // Error, x has not been given an inferred type on all paths
@@ -748,7 +752,7 @@ void test(bool b) {
     y = 3;
   } // Ok, y has inferred type `int`
 
-  var z;
+  final z;
   if (b) {
      z = 3;
   } else {
@@ -758,53 +762,114 @@ void test(bool b) {
 ```
 
 At a join point in the program, if an **untyped variable** has been given an
-**initialization inferred type** on more than one of the reachable paths to the
-join point and all of such **initialization inferred types** are syntactically
-equal, the variable is treated after the join point as having that
-**initialization inferred type**.  The promoted type chains for the variable on
-the respective paths are then intersected as usual.
+**initialization inferred type** on all of the reachable paths into the join and
+at least one of the paths is reachable, then the **initialization inferred
+type** for the variable after the join point is computed as follows:
+  - If the **initialization inferred type** on each reachable path is the same
+    type `T` (using syntactic equality), then the **initialization inferred
+    type** after the join is `T`.
+  - If the **initialization inferred type** on one or more reachable paths is
+    the same type `T` (using syntactic equality), and the **initialization
+    inferred type** on all other reachable paths is `Null`, then the
+    **initialization inferred type** after the join is `T?`.
+  - Otherwise, it is an error.
+
+In other words, if an **untyped** variable of any kind is assigned on all paths
+into the join, it must be assigned either a value of the same type, or a value
+of type `Null`, on all paths; and if the variable is assigned a value of type
+`Null` on any path, then the variable is treated as having nullable type.
 
 At a join point in the program, if an **untyped variable** has been given an
-**initialization inferred type** on more than one of the reachable paths to the
-join point and any two such **initialization inferred types** are not
-syntactically equal, it is an error.
+**initialization inferred type** on some but not all of the reachable paths into
+the join, then the **initialization inferred type** for the variable after the
+join point is computed as follows:
+  - If the variable is declared `final` and not `late` it is an error, and there
+    is no **initialization inferred type** after the join.
+  - If the variable is declared `late` and the **initialization inferred type**
+    on each path which has one is the same type `T` (using syntactic equality),
+    then the **initialization inferred type** after the join is `T`.
+  - If the variable is declared `late` and/or `var` and the **initialization
+    inferred type** on one or more paths is the same type `T` (using syntactic
+    equality), and the **initialization inferred type** on one or more paths is
+    `Null`, then the **initialization inferred type** after the join is `T?`.
+  - If the variable is declared `var` and not `late` and the **initialization
+    inferred type** on one or more paths is the same type `T` (using syntactic
+    equality), and the **initialization inferred type** on all other paths (if
+    any) is `Null`, then the **initialization inferred type** after the join is
+    `T?`.
+  - Otherwise it is an error.
+
+In other words, a non-`late`, `final` **untyped** variable must be assigned a
+value on every path, but a variable declared `late` and/or `var` may be assigned
+a value on only some of the paths into a join.  In the latter case, the type is
+made nullable if the variable is declared as non-`late`, and the variable is
+treated as implicitly null initialized.  Variables declared `late` are only made
+nullable if `Null` is explicitly assigned on one more paths into the join, and
+there is no implicit null initialization.  In all cases, the values assigned
+must all have the same type, or type `Null`.
 
 
 ```dart
 void test(bool b) {
-  var x; // x is an untyped variable
-  if (b) {
-    x = 3;
-  } else {
-    x = 3.0;
-  } // Error: x has two different types on different paths
-
-  late var y; // y is a late untyped variable
-  if (b) {
+  { // late and/or final the same
+    var x; // x is an untyped variable
     if (b) {
-      y = 3;
+      x = 3;
     } else {
-      y = 3.0;
-    } // Error: y has two different types on different paths
+      x = 3.0;
+    } // Error: x has two different types on different paths
   }
 
- var z; // z is an untyped variable
- if (b) {
-   z = 3; // z is promoted to int
- } else {
-   z = 3.0; // z is promoted to double
-   return;
- } // No error.  z is inferred as int on the only reachable path
- z.isEven; // No error, only one reachable path
+  { // late and/or final the same
+    var x; // x is an untyped variable
+    if (b) {
+      x = 3;
+    } else {
+      x = null;
+    } // x has type int?
+  }
 
- var u; // u is an untyped variable
- if (b) {
-   u = 0 as num; // u is promoted to num
-   if (u is! int) u = 1;
-   // u has inferred type num, but is promoted to int
- } else {
-   u = 3;
- } // Error: u is inferred to different types on different paths
+  {
+    var x; // x is an untyped variable
+    if (b) {
+      x = 3;
+    }
+    // x has type int?
+  }
+
+  {
+    final x; // x is an untyped variable
+    if (b) {
+      x = 3;
+    }
+    // Error: not assigned on all paths
+  }
+
+  { // late final the same
+    late x; // x is an untyped variable
+    if (b) {
+      x = 3;
+    }
+    // x has type int
+  }
+
+  var z; // z is an untyped variable
+  if (b) {
+    z = 3; // z is promoted to int
+   } else {
+    z = 3.0; // z is promoted to double
+    return;
+  } // No error.  z is inferred as int on the only reachable path
+  z.isEven; // No error, only one reachable path
+
+  var u; // u is an untyped variable
+  if (b) {
+    u = 0 as num; // u is promoted to num
+    if (u is! int) u = 1;
+    // u has inferred type num, but is promoted to int
+  } else {
+    u = 3;
+  } // Error: u is inferred to different types on different paths
 }
 ```
 
@@ -814,23 +879,39 @@ inferred type**.
 
 ```dart
 void test(bool b) {
-  var x; // x is an untyped variable
-  print(x); // Error, no initialization inferred type
+  {
+    var x; // x is an untyped variable
+    print(x); // Error, no initialization inferred type
+  }
 
-  late var y; // y is a late untyped variable
-  var f = () => print(y); // Error, y has no initialization inferred type
+  {
+    late var x; // x is an untyped variable
+    var f = () => print(x); // Error, x has no initialization inferred type
+  }
 
-
-  var z;
-  do {
-    // z is potentially assigned.
+  {
+    var x; // x is an untyped variable
     if (b) {
-      z = 42;
+      print(x); // Error: x has no initialization inferred type.
+      x = 3;
     } else {
-      z = 37;
-    }
-  } while (something);
-  print(z); // Error, z has no initialization inferred type
+      x = 4;
+    } // x is inferred as type int
+    print(x); // No error here, since x has been inferred at this point
+  }
+
+  {
+    var z;
+    do {
+      // z is potentially assigned.
+      if (b) {
+        z = 42;
+      } else {
+        z = 37;
+      }
+    } while (something);
+    print(z); // Error, z has no initialization inferred type
+  }
 ```
 
 As a consequence of the above it is impossible, in an error-free program, to
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -612,8 +612,10 @@ It is a warning to use the null check operator (`!`) on an expression of type
 ### Local variables and definite (un)assignment.
 
 As part of the null safety release, errors for local variables are specified to
-take into account **definite assignment** and **definite unassignment** (see
-the section on Definite Assignment below).
+take into account **definite assignment** and **definite unassignment** (see the
+section on Definite Assignment below).  We say that a variable is **potentially
+assigned** if it is not **definitely unassigned**, and that a variable is
+**potentially unassigned** if it is not **definitely assigned**.
 
 In all cases in this section, errors that are described as occurring on reads of
 a variable are intended to apply to all form of reads, including indirectly as
@@ -622,35 +624,37 @@ operators.  Similarly, errors that are described as occurring on writes of a
 variable are intended to apply to all form of writes.
 
 It is a compile time error to assign a value to a `final`, non-`late` local
-variable which is not **definitely unassigned**.  Thus, it is *not* an error to
+variable which is **potentially assigned**.  Thus, it is *not* an error to
 assign to a **definitely unassigned** `final` local variable.
 
 It is a compile time error to assign a value to a `final`, `late` local variable
 if it is **definitely assigned**. Thus, it is *not* an error to assign to a
-*not* **definitely assigned** `final`, `late` local variable.
+**potentially unassigned** `final`, `late` local variable.
 
-*Note that a variable is always considered definitely assigned and not
-definitely unassigned if it has an explicit initializer, or an implicit
+*Note that a variable is always considered **definitely assigned** and not
+**definitely unassigned** if it has an explicit initializer, or an implicit
 initializer as part of a larger construct (e.g. the loop variable in a `for in`
-construct).
+construct).*
 
 It is a compile time error to read a local variable when the variable is
 **definitely unassigned** unless the variable is non-final and has nullable
-type.  This includes variables marked `late` and/or `final`.
+type.
 
-It is a compile time error to read a local variable when the variable is not
-**definitely assigned** unless the variable is non-final and has nullable type,
-or is `late`, or is declared using `var` with no type and no initializer.
+It is a compile time error to read a local variable when the variable is
+**potentially unassigned** unless the variable is non-final and has nullable
+type, or is `late`, or is declared using `var` with no type and no initializer.
 
-The erors specified above are summarized in the following table (using `int` as
-an example non-nullable type).  A variable which has an initializer (explicit or
+The errors specified above are summarized in the following table, where `int` is
+used as an example of an arbitrary **potentially non-nullable** type, `int?` is
+used as an example of an arbitrary **nullable** type, and `T` is used to stand
+for a type of any nullability.  A variable which has an initializer (explicit or
 implicit) is always considered definitely assigned, and is never considered
-definitely unassigned.  Note that the `var x` form may only be read in the "not
-definitely assigned" state when it has been subject to initialization promotion
-as described below.
+definitely unassigned.  Note that the `var x` form may only be read in the
+**potentially unassigned** state when it has been subject to initialization
+promotion as described below.
 
 
-Read Errors:
+Read Behavior:
 
 | Declaration form  | Def. Assigned | Neither           | Def. Unassigned |
 | ----------------- | ------------- | ----------------- | --------------- |
@@ -658,14 +662,13 @@ Read Errors:
 | final x;          | Ok            | Error             | Error           |
 | int x;            | Ok            | Error             | Error           |
 | int? x;           | Ok            | Ok                | Ok              |
-| final int x;      | Ok            | Error             | Error           |
-| final int? x;     | Ok            | Error             | Error           |
+| final T x;        | Ok            | Error             | Error           |
 | late var x;       | Ok            | Ok                | Error           |
 | late final x;     | Ok            | Ok                | Error           |
-| late int x;       | Ok            | Ok                | Error           |
-| late final int x; | Ok            | Ok                | Error           |
+| late T x;         | Ok            | Ok                | Error           |
+| late final T x;   | Ok            | Ok                | Error           |
 
-Write Errors:
+Write Behavior:
 
 | Declaration form  | Def. Assigned | Neither             | Def. Unassigned |
 | ----------------- | ------------- | ------------------- | --------------- |
@@ -673,12 +676,11 @@ Write Errors:
 | final x;          | Error         | Error               | Ok              |
 | int x;            | Ok            | Ok                  | Ok              |
 | int? x;           | Ok            | Ok                  | Ok              |
-| final int x;      | Error         | Error               | Ok              |
-| final int? x;     | Error         | Error               | Ok              |
+| final T x;        | Error         | Error               | Ok              |
 | late var x;       | Ok            | Ok                  | Ok              |
 | late final x;     | Error         | Ok                  | Ok              |
-| late int x;       | Ok            | Ok                  | Ok              |
-| late final int x; | Error         | Ok                  | Ok              |
+| late T x;         | Ok            | Ok                  | Ok              |
+| late final T x;   | Error         | Ok                  | Ok              |
 
 ### Local variables and initialization based inference
 
@@ -704,10 +706,12 @@ void test() {
 ```
 
 Local variables with no explicitly written type but with an initializer are
-given an inferred declared type equal to the type of their initializer.  In the
-case that the type of the initializer is a promoted type variable `X & T`, the
-inferred type of the variable shall be `X`.  However, such a variable shall be
-treated as immediately promoted to `X & T`.
+given an inferred type equal to the type of their initializer.  The inferred
+type of the variable is considered a "type of interest" in the sense defined in
+the flow analysis specification.  In the case that the type of the initializer
+is a promoted type variable `X & T`, the inferred type of the variable shall be
+`X`.  However, such a variable shall be treated as immediately promoted to `X &
+T`.
 
 #### Untyped variables
 
@@ -761,6 +765,18 @@ void test(bool b) {
 }
 ```
 
+At join points, some adjustment to the **initialization inferred type** is done
+to account for nullability.  Informally, a non-`late`, `final` **untyped**
+variable must be assigned a value on every path, but a variable declared `late`
+and/or `var` may be assigned a value on only some of the paths into a join.  In
+the latter case, the type is made nullable if the variable is declared as
+non-`late`, and the variable is treated as implicitly null initialized.
+Variables declared `late` are only made nullable if `Null` is explicitly
+assigned on one more paths into the join, and there is no implicit null
+initialization.  In all cases, the values assigned must all have the same type,
+or type `Null`.  More precisely, the **initialization inferred type** of a
+variable after a join point is specified as follows.
+
 At a join point in the program, if an **untyped variable** has been given an
 **initialization inferred type** on all of the reachable paths into the join and
 at least one of the paths is reachable, then the **initialization inferred
@@ -785,28 +801,14 @@ the join, then the **initialization inferred type** for the variable after the
 join point is computed as follows:
   - If the variable is declared `final` and not `late` it is an error, and there
     is no **initialization inferred type** after the join.
-  - If the variable is declared `late` and the **initialization inferred type**
-    on each path which has one is the same type `T` (using syntactic equality),
-    then the **initialization inferred type** after the join is `T`.
-  - If the variable is declared `late` and/or `var` and the **initialization
-    inferred type** on one or more paths is the same type `T` (using syntactic
-    equality), and the **initialization inferred type** on one or more paths is
-    `Null`, then the **initialization inferred type** after the join is `T?`.
-  - If the variable is declared `var` and not `late` and the **initialization
-    inferred type** on one or more paths is the same type `T` (using syntactic
-    equality), and the **initialization inferred type** on all other paths (if
-    any) is `Null`, then the **initialization inferred type** after the join is
-    `T?`.
+  - Otherwise, if the variable is declared `late` and the **initialization
+    inferred type** on each path which has one is the same type `T` (using
+    syntactic equality), then the **initialization inferred type** after the
+    join is `T`.
+  - Otherwise, if the **initialization inferred type** on each path into the
+    `join` which has one is either `T`, `T?`, or `Null` for some type `T`, the
+    **initialization inferred type** after the join is `T?`.
   - Otherwise it is an error.
-
-In other words, a non-`late`, `final` **untyped** variable must be assigned a
-value on every path, but a variable declared `late` and/or `var` may be assigned
-a value on only some of the paths into a join.  In the latter case, the type is
-made nullable if the variable is declared as non-`late`, and the variable is
-treated as implicitly null initialized.  Variables declared `late` are only made
-nullable if `Null` is explicitly assigned on one more paths into the join, and
-there is no implicit null initialization.  In all cases, the values assigned
-must all have the same type, or type `Null`.
 
 
 ```dart
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -721,14 +721,17 @@ are those declared as `var x;`, `final x;`, `late var x;`, or `late final x;`.
 We refer to such variables as **untyped variables**.
 
 An **initializing assignment** is an assignment to an **untyped variable** when
-that variable is in the **definitely unassigned** state, and the variable would
-have been promoted if it had been declared with type `Object?` and the type of
-the assigned expression were a type of interest for the variable.  Such an
-assignment on a path shall cause the variable in question to be treated by the
-flow analysis on that path as if it were declared with the assigned expression
-as its initializing expression, up until the next join point in the program.  We
-refer to this inferred type as the **initialization inferred type** of the
-variable.  The **initialization inferred type** of the variable is always
+it is *promotable via initialization* as defined in
+the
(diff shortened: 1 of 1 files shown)


accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,341 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+at least one of the paths is reachable, then the **initialization inferred
+type** for the variable after the join point is computed as follows:
+  - If the **initialization inferred type** on each reachable path is the same
+    type `T` (using syntactic equality), then the **initialization inferred
+    type** after the join is `T`.
+  - If the **initialization inferred type** on one or more reachable paths is
+    the same type `T` (using syntactic equality), and the **initialization
+    inferred type** on all other reachable paths is `Null`, then the
+    **initialization inferred type** after the join is `T?`.
+  - Otherwise, it is an error.
```
lrhn, line 775: How about the situation where three branches has the types `int`, `int?` and `Null`?
Or just two branches having the types `int` and `int?`. Which can happen with consecutive joins like;
```dart
var x;
if (test1) {
  x = 1;
} else if (test2) {
  x = null;
} else {
  x = 0;
}
```
Here the join points would first combine `x=null` and `x=0` to `int?`, then fail to join that with `x=1`.

So, reasonable code which should probably be accepted, because otherwise we can't explain this rule to users.

Maybe:

> - Otherwise, if there exists a type *T* such that **initialization inferred type** on all paths are either the same as *T* or *T*? (using syntactic equality) or it is `Null`, then the **initialization inferred type** after the join is *T*?.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,341 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+      x = null;
+    } // x has type int?
+  }
+
+  {
+    var x; // x is an untyped variable
+    if (b) {
+      x = 3;
+    }
+    // x has type int?
```
munificent, line 837: I think this will *not* be what users want most of the time. It's probably a mistake that they didn't initialize it. But the idea is that you'll get a nullable type and then you'll get some other downstream error when you try to use that, so erroneous uses of this should still be caught. And then it *is* what you want, it lets you. Is that the idea?

If so, SGTM.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,341 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+  z.isEven; // No error, only one reachable path
+
+  var u; // u is an untyped variable
+  if (b) {
+    u = 0 as num; // u is promoted to num
+    if (u is! int) u = 1;
+    // u has inferred type num, but is promoted to int
+  } else {
+    u = 3;
+  } // Error: u is inferred to different types on different paths
```
munificent, line 872: Subtle! This will trip people up when they run into it. Is it a crazy idea to use the promoted type at a join point instead of just the inferred type?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,341 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+type.  However, IDEs and tools may wish to report a static type for such a
+variable for the purposes of documentation and error reporting.  For such
+purposes, tools shall treat an untyped variable as if it were declared with a
+type equal to the upper bound of all of the **initialization inferred types**
+that it is ascribed in the method, plus `Never`.  The latter implies that a
+variable which is ascribed no **initialization inferred types** on any path
+shall be treated as having declared type `Never`.  Note that despite the
+requirement that the **initialization inferred types** be equal at join points,
+it is possible for a variable to take on different **initialization inferred
+types** on paths that never join before exiting.
```
munificent, line 929: So then a stranger example is:

```dart
test(c) {
  var x;
  if (c) {
    x = "a string";
    return;
  } else {
    x = 123;
    return;
  }
}
```

So you can create a variable and use it as totally arbitrary unrelated types as long as you escape before those paths join. It's a little strange to me that the above example is allowed but this is not:

```dart
test(c) {
  var x;
  if (c) {
    x = "a string";
  } else {
    x = 123;
  }
}
```

With `final`, we don't make it an error to join two paths where one is definitely assigned and the other is not. We just leave the variable in a half-assigned unusable state and then make it an error to touch it. But here we say the join itself is the point of error.

Should we pick one or the other strategy and have a more consistent model?

I lean slightly towards making the use of the variable the error and not the join itself. Pragmatically, that allows more programs that don't do anything harmful. And I think it draws the user's attention to a concrete entity — a variable they declared that they are using — instead of to a somewhat nebulous control flow concept.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,341 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+
+In all cases in this section, errors that are described as occurring on reads of
+a variable are intended to apply to all form of reads, including indirectly as
+part of composite assignment operators, as well as via pre and post-fix
+operators.  Similarly, errors that are described as occurring on writes of a
+variable are intended to apply to all form of writes.
+
+It is a compile time error to assign a value to a `final`, non-`late` local
+variable which is not **definitely unassigned**.  Thus, it is *not* an error to
+assign to a **definitely unassigned** `final` local variable.
```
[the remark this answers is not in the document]
lrhn: We've previously used "potentially assigned" and "potentially unassigned" for "not definitely unassigned" and "not definitely assigned", respectively.

accepted/future-releases/nnbd/feature-specification.md
lrhn: Alternatively, we can just make it have *no* inferred initialization type, and make it an error to read or write a variable with no type. The variable will still be potentially assigned so no further initialization can happen.
Then
```dart
foo() {
  var x;
  if (test) {
    x = 4;
    something(x);
  } else if (test2) {
    x = 4.0;
    somethingElse(x);
  } else {
    throw "whatever";
  }
  return;
}
```
would be valid because even though there is no type for `x` after the if-chain, there is also no use of it.

it's different because we have to let the program survive in a state where a variable has no type, rather than make it an error at the join point, which is why we have to explicitly call out that the variable is unusable afterwards.

accepted/future-releases/nnbd/feature-specification.md
lrhn: After seeing these examples, I agree that it's not the join that's bad, it's the use after join.
It's not just reading the variable which is iffy afterwards, you also shouldn't assign to the variable (because it has no type).

It still means that we need to introduce a *state* after the join which can represent the "unusable" variable.
Maybe just:
* Variable is untyped and definitely unassigned: You can assign to it, and it's an initializing assignment. After the assignment it's not untyped any more, it has an inferred initialization type.
* Variable is untyped and potentially assigned: You cannot read or write the variable. Not even if it's late. You can't lose "potentially assigned" by joining.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,341 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+
+It is a compile time error to assign a value to a `final`, `late` local variable
+if it is **definitely assigned**. Thus, it is *not* an error to assign to a
+*not* **definitely assigned** `final`, `late` local variable.
+
+*Note that a variable is always considered definitely assigned and not
+definitely unassigned if it has an explicit initializer, or an implicit
+initializer as part of a larger construct (e.g. the loop variable in a `for in`
+construct).
```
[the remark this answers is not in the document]
leafpetersen: done

accepted/future-releases/nnbd/feature-specification.md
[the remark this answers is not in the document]
leafpetersen: Done.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,341 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+
+It is a compile time error to read a local variable when the variable is
+**definitely unassigned** unless the variable is non-final and has nullable
+type.  This includes variables marked `late` and/or `final`.
```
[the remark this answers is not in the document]
leafpetersen: done.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,341 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+
+It is a compile time error to read a local variable when the variable is not
+**definitely assigned** unless the variable is non-final and has nullable type,
+or is `late`, or is declared using `var` with no type and no initializer.
+
+The erors specified above are summarized in the following table (using `int` as
```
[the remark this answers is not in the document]
leafpetersen: done

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,341 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+an example non-nullable type).  A variable which has an initializer (explicit or
```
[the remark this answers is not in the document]
leafpetersen: done.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,341 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+implicit) is always considered definitely assigned, and is never considered
+definitely unassigned.  Note that the `var x` form may only be read in the "not
+definitely assigned" state when it has been subject to initialization promotion
+as described below.
+
+
+Read Errors:
```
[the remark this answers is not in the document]
leafpetersen: done

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,341 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+```dart
+void test() {
+  int? x = 3; // x is declared at `int?`
+  x.isEven; // Valid, x has been promoted to `int`
+  x = null; // Valid, demotes to the declared type.
+}
+```
+
+Local variables with no explicitly written type but with an initializer are
+given an inferred declared type equal to the type of their initializer.  In the
```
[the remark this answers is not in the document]
leafpetersen: done

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,341 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+case that the type of the initializer is a promoted type variable `X & T`, the
+inferred type of the variable shall be `X`.  However, such a variable shall be
+treated as immediately promoted to `X & T`.
```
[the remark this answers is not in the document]
leafpetersen: Eliminated all uses of declared.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,341 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+version if applicable).
+
+```dart
+void test() {
+  var x; // x is an untyped variable
+  x = 3; // x is given the initialization inferred type of `int`
+}
+```
+
+At a join point in the program, it is an error if an **untyped variable** has
```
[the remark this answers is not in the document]
leafpetersen: I find your rewording very confusing, because the binding of the `non` part is ambiguous.  I think you mean "non-`late`, and `final`,", which is correct but awkward when written out that way, but as written it can also be read "non-('late` and `final`) which is incorrect.  

To your second question, no.  You can have variables that are definitely assigned but have no initialization inferred types.  That's one of the key points of all three of these proposals: that they separate the notion of having an initialization inferred type from being definitely assigned.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,341 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+been given an **initialization inferred type** on one of the reachable paths to
+the join point and not on another reachable path, unless the variable is
+declared as `late`, or using `var`.  Note that it is valid for a variable to
```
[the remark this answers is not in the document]
leafpetersen: > This makes it an error for an _untyped_ `final`, non-`late` variable to be potentially both assigned and potentially unassigned.

Sorry, I don't follow.  Nothing you quote says anything about definite assignment, and I think your statement is not true.  This program is not an error in my proposal:

```dart
final x;
if (b) {
  while(b) {
    x = 3;
  }
  // x is potentially assigned and potentially unassigned
} else {
  while(b) {
    x = 3;
  }
  // x is potentially assigned, and potentially unassigned
} // not an error, x has no initialization inferred type on any path
```

I'm not sure if that's the example you had in mind?  But it's not an error.  In general, talking about definite assignment here is not useful: the only place that definite assignment/unassignment comes into this section is in the definition of an initializing assignment.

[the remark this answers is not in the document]
leafpetersen: > ```dart
> foo() {
>   final x;
>   if (c) {
>     x = value;
>     print(x);
>   }
> }
> ```

No, my proposal makes this an error.  See the first test below.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,341 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+  final x; // x is an untyped variable
+  if (b) {
+    x = 3;
+  } // Error, x has not been given an inferred type on all paths
+
+  late var y; // y is a late untyped variable
+  if (b) {
+    y = 3;
+  } // Ok, y has inferred type `int`
+
```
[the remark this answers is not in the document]
leafpetersen: Yes.  One path into the join has an initializing assignment, given an IIT of `int`.  At the join point, the other path has no IIT, so we make the IIT after the join `int?` and the assignment is valid.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: Good point, I'll take a crack at making this work.  Definitely starts getting complicated here though.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,341 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+join point is computed as follows:
+  - If the variable is declared `final` and not `late` it is an error, and there
+    is no **initialization inferred type** after the join.
+  - If the variable is declared `late` and the **initialization inferred type**
+    on each path which has one is the same type `T` (using syntactic equality),
+    then the **initialization inferred type** after the join is `T`.
+  - If the variable is declared `late` and/or `var` and the **initialization
+    inferred type** on one or more paths is the same type `T` (using syntactic
+    equality), and the **initialization inferred type** on one or more paths is
+    `Null`, then the **initialization inferred type** after the join is `T?`.
```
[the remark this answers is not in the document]
leafpetersen: done.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,341 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+  - If the variable is declared `var` and not `late` and the **initialization
+    inferred type** on one or more paths is the same type `T` (using syntactic
+    equality), and the **initialization inferred type** on all other paths (if
+    any) is `Null`, then the **initialization inferred type** after the join is
+    `T?`.
```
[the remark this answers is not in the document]
leafpetersen: No, but rephrased anyway.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,341 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+  - Otherwise it is an error.
+
+In other words, a non-`late`, `final` **untyped** variable must be assigned a
+value on every path, but a variable declared `late` and/or `var` may be assigned
+a value on only some of the paths into a join.  In the latter case, the type is
+made nullable if the variable is declared as non-`late`, and the variable is
+treated as implicitly null initialized.  Variables declared `late` are only made
+nullable if `Null` is explicitly assigned on one more paths into the join, and
+there is no implicit null initialization.  In all cases, the values assigned
+must all have the same type, or type `Null`.
```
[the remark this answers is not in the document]
leafpetersen: done.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: Yes.  I mean, it's easy to not allow this at all - I can say that you *must* write a value on every path to the join, so you'd have to write this with an `x=null` in an else branch.  But the idea was to allow this, since anything else feels a little verbose.  The one place I can see this not catching legit errors is if you're using a nullable API, but you didn't actually want to pass `null`.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: > This will trip people up when they run into it. Is it a crazy idea to use the promoted type at a join point instead of just the inferred type?

Not sure, I can think about it.

leafpetersen: > Alternatively, we can just make it have _no_ inferred initialization type, and make it an error to read or write a variable with no type.

It is already an error to read a variable with no IIT.  We can make it an error to write, or not.

The general idea of making the IIT non-monotonic feels icky, but maybe would work out fine.  You'd have a variable which was in a usable state for some period of time, and then becomes unusable.  Feels weird, but probably doable.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: > With `final`, we don't make it an error to join two paths where one is definitely assigned and the other is not.

One of us is confused.  Again, you need to stop talking about definite assignment, but assuming you mean "join two paths where one has an IIT and one does not", I believe my spec is intended to make this an error:

```
  - If the variable is declared `final` and not `late` it is an error, and there
    is no **initialization inferred type** after the join.
```

leafpetersen: > After seeing these examples, I agree that it's not the join that's bad, it's the use after join.
> I

I can probably work something out around this.

leafpetersen: > It still means that we need to introduce a _state_ after the join which can represent the "unusable" variable.

Thinking about this a bit, an unpleasant consequence of this is that we need to take this state into account in joins.  The following program shouldn't be legal:

```
var x;
if (b) {
  if (b) { 
    x = 3.0;
  } else {
   x = 3;
  } // x is joined, and loses its IIT
} else {
  x = "hello";
} // x has not IIT on one path, and an IIT of `String` on another.
  // As currently specified, we would infer `String?` which is wrongo.
  // So we need to track the fact that `x` had an IIT at some point
  // in the past.  It might be sufficient to use "potentially assigned" as 
  // a proxy for this.  Complicated....

File: accepted/future-releases/nnbd/feature-specification.md (hunk 3 above)
Comment (leafpetersen) [reply accepted/future-releases/nnbd/feature-specification.md:874]: > Subtle! This will trip people up when they run into it. Is it a crazy idea to use the promoted type at a join point instead of just the inferred type?

Where this starts getting complicated is deciding what combinations you allow.  If I infer a variable as `Object` on one path, then promoted it from there to `num`, and then to `int`, what joins do I allow?  Can I join it with another path that inferred `Object` (demoting to `Object`)?  Can I join it with another path that inferred `num` (demoting to `num`)?  What about another path that inferred `num?`?  Do I have to look at every pair of types in the promotion chain, with their nullable + non-nullable versions?  And what about when there are multiple paths into the join?  I'll think about whether there's a clean story here, but it feels really icky on first glance.

File: accepted/future-releases/nnbd/feature-specification.md (hunk 4 above)
Comment (leafpetersen) [reply accepted/future-releases/nnbd/feature-specification.md:931]: > I lean slightly towards making the use of the variable the error and not the join itself. Pragmatically, that allows more programs that don't do anything harmful. And I think it draws the user's attention to a concrete entity — a variable they declared that they are using — instead of to a somewhat nebulous control flow concept.

Just a comment, as I run through updating the test cases.  This isn't how it ends up feeling to me.  When the error is at the join point, there's a clear error you can report: I can't infer a type for this variable.  When the error is at the use point, it feels weirder.  What do you say?  "This variable can't be used because I couldn't infer a type" is the best I could get.  You can say "Type can't be inferred for a variable", but that's going to be confusing because I had code that worked until I added a use of a variable.  The user says "you could infer it before, why not now?".

Comment (leafpetersen) [reply accepted/future-releases/nnbd/feature-specification.md:968]: I added a patch set implementing this.  You can see the changes from that commit only here: https://github.com/dart-lang/language/pull/1009/commits/ee0d27dc1038e021d72b4cb59df573ba46e0bd44

File: accepted/future-releases/nnbd/feature-specification.md (hunk 13 above)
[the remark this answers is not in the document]
Comment (lrhn) [reply accepted/future-releases/nnbd/feature-specification.md]: English is weirdly ambiguous sometimes! In Danish I'd write `non-late and -final` if I wanted `non` to bind to both, and `non-late and final` if I didn't.

Apart from that, I like using "assigned" rather than having an inferred type. Then it's just a matter of permuting the conjunctive phrases until it is readable.

> If an untyped variable is `final` and non-`late`, it is an error if, at a join point, the variable is potentially assigned on at least one reachable path, and potentially unassigned on at least one other reachable path.

or

> It is an error if an untyped variable is declared `final` and not declared `late` and there are two reachable paths leading to a join point where the variable is potentially assigned on one path and potentially unassigned on the other.

or ...

File: accepted/future-releases/nnbd/feature-specification.md (hunk 4 above)
Comment (eernstg) [reply accepted/future-releases/nnbd/feature-specification.md:968]: Considering the complexity of working with code that relies on these rules, I agree that reporting the error at the join is preferable.

Otherwise we'll have a situation where any number of variables could have a broken type for any number of lines, and you can't readily see how far back you need to go to find (1) the join where the problem can first be detected, and (2) the predecessors in the flow graph where some unintended IIT caused the problem. The developer needs to find that second location, and we shouldn't make that task intractable.

File: accepted/future-releases/nnbd/feature-specification.md [hunk 19]
```diff
@@ -616,6 +609,442 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+    x = 3; // x is inferred to int
+    return;
+  } else if (b) {
+    return;  // Note, x has no inferred type on this path
+  } else {
+    // x not assigned
+  } // x has type `int?`
+}
+```
+
```
lrhn, line 1047: Do we need to say that when the specification refers to the *declared type* of a variable, an untyped variable uses its IIT instead, if it has one (and it's probably an error if it doesn't have one).

lrhn:
Looks very good now. I tried to find holes in the algorithm, but couldn't.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,442 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+      // u is inferred as int
+    } Error: u has no inferred type
+  }
+}
+```
+
+It is an error to read an **untyped variable** when it has no **initialization
+inferred type**.
+
+It is an error to write to an **untyped variable** when it has no
```
stereotype441, line 982: Do we need this?  In all the cases I can think of where an untyped variable is written to and there's no initialization inferred type, either the write is an initializing assignment or there is already some other error.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,442 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+    print(z); // Error, z has no initialization inferred type
+  }
+```
+
+Notice that the above definitions never discard an **initialization inferred
+type** from any path, and always ensure that the **initialization inferred
+type** is consistent across joins.  Any variable which is given an
+**initialization inferred type** in this way may be treated by IDEs and tools as
+having this type for the purposes of documentation and error messages. An
+untyped cariable which is never read nor written may have no **initialization
```
stereotype441, line 1029: cariable -> variable

accepted/future-releases/nnbd/feature-specification.md
stereotype441: +1.

Side note: I think the spec has similar problems with static and instance variables.  E.g. the "variables" section of the spec says:
> Let v be variable declared in an initializing variable declaration, and let e be the associated initializing expression.  It is a compile-time error if the static type of e is not assignable to the declared type of v.

But if v is an instance variable, and the declaration of v doesn't include a type, and the type is inferred from a base class, then we want this rule to apply to the inferred type of v.

I wonder if it would be better to replace all instances of the term "declared type" in the spec with some other term, and define that term explicitly?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,442 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
 It is a warning to use the null check operator (`!`) on an expression of type
 `T` if `T` is **strictly non-nullable** .
 
+### Local variables and definite (un)assignment.
+
+As part of the null safety release, errors for local variables are specified to
+take into account **definite assignment** and **definite unassignment** (see the
+section on Definite Assignment below).  We say that a variable is **potentially
+assigned** if it is not **definitely unassigned**, and that a variable is
+**potentially unassigned** if it is not **definitely assigned**.
```
munificent, line 618: I was very confused about the two "potential" states for a while because I couldn't tell how they differed. That's because I assumed that these states were mutually exclusive. But on further reader, I guess that's not the intent? Am I correct that these mean:

* Definitely assigned: All paths assign the variable.
* Definitely assigned: No path assigns the variable.
* Potentially assigned: There exists at least one path which assigns the variable. There may be paths that do not, or it may be that all paths do. So every definitely assigned variable is also potentially assigned.
* Potentially unassigned: There exists at least one path which does not assign the variable. There may be paths that do, or it may be that no paths do. So every definitely unassigned variable is also potentially unassigned.

If the "potentially" states are intended to be exclusive of the "definitely" states, then I think that implies that there is only one potentially state:

* Either: There exists at least one path which assigns (to exclude it from the definitely unassigned state) and at least one path which does not assign (to exclude it from the definitely assigned state).

Am I understanding this correctly?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,442 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+initializer as part of a larger construct (e.g. the loop variable in a `for in`
+construct).*
+
+It is a compile time error to read a local variable when the variable is
+**definitely unassigned** unless the variable is non-final and has nullable
+type.
+
+It is a compile time error to read a local variable when the variable is
+**potentially unassigned** unless the variable is non-final and has nullable
+type, or is `late`, or is declared using `var` with no type and no initializer.
```
munificent, line 645: You can drop the "and no initializer" because it the variable had an initializer, it would be definitely assigned and not potentially unassigned.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,442 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+
+*Treating the declared type of the variable as a "type of interest" implies that
+if the variable is a nullable type, then its non-nullable version is also a type
+of interest.  Treating the initialization as an assignment for the purposes of
+promotion means that initializing a mutable variable declared at type `T?` with
+a value of non-nullable type `T` immediately promotes the variable to the
+non-nullable type.*
+
+```dart
+void test() {
```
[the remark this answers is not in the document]
munificent: I think Leaf is trying to show an example of the last sentence — the initializer itself promotes to the non-nullable type. So these two would be equivalent:

```dart
void test1() {
  int? i = 1; // Promote on initialization.
  i.isEven;
}

void test2() {
  int? i;
  i = 1; // Promote on assignment.
  i.isEven;
}
```

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,442 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+**initialization inferred type** on all of the *reachable* paths into the join,
+then the **initialization inferred type** for the variable after the join point
+is computed as follows:
+  - If the **initialization inferred type** on each path which has one
+    (reachable or not) is the same type `T` (using syntactic equality), then the
+    **initialization inferred type** after the join is `T`.
+  - Otherwise, if the **initialization inferred type** on each path which has
+    one (reachable or not) is one of `T`, `T?` or `Null` for some type `T`
+    (using syntactic equality), then the **initialization inferred type** after
+    the join is `T?`.
```
munificent, line 807: This case would match if the IIT on all paths is `Null`, but in that case, you don't know what T should be:

```dart
var x;
if (c) {
  x = null;
} else {
  x = null;
}
// what is x here?
```

This is a corner where my proposal gives you an answer. It says `x` is in the potentially assigned `Null` type state, which means it can be used as `null` but can still have a later type inferred:

```dart
var x; // Unassigned+Untyped
if (c) {
  x = null; // Assigned+Null
} else {
  x = null; // Assigned+Null
} // Join = Assigned+Null
print(x); // OK, null.
x = 123; // OK, Assigned+Typed(int?)
x = "str"; // Error, wrong type.
```

(Whether the answer my proposal provides is a *good* answer is an open question. :) )

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,442 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+**initialization inferred type** on at least one *reachable* path into the join,
+then the **initialization inferred type** for the variable after the join point
+is computed as follows:
+  - If the variable is declared `late` and the **initialization inferred type**
+    on each path (reachable or not) which has one is the same type `T` (using
+    syntactic equality), then the **initialization inferred type** after the
+    join is `T`.
+  - Otherwise, if the **initialization inferred type** on each path into the
+    `join` which has one is either `T`, `T?`, or `Null` for some type `T`, the
+    **initialization inferred type** after the join is `T?`.
```
munificent, line 829: Likewise:

```dart
var x;
if (c) x = null;
// what is x here?
```

And according to what I proposed:

```dart
var x; // Unassigned+Untyped
if (c) x = null; // Assigned+Null
// Join: Either+Null
print(x); // OK, potentially assigned some unknown nullable type, treat as Null for now.
x = 123; // OK, now Assigned+Typed(int?)
x = "str"; // Error, wrong type.
```

accepted/future-releases/nnbd/feature-specification.md
munificent: I keep thinking about this case and I still worry that we're implicitly giving a user a nullable type that they didn't request. I wish we could instead make this case an error and tell them to solve it by writing:

```dart
var x = null; // User opts in to null.
if (b) {
  x = 3;
}
```

This is just a hunch, though, and I could be wrong. Here's the evidence I have in mind:

* Users often tell us that they want more "missing return" errors even in functions that can return `null`. Many seem to want this to be an error:

    ```dart
    Object foo(bool c) { // Object? with NNBD.
      if (c) return "thing";
    }
    ```
    
    And instead want to have to write:

    ```dart
    Object foo(bool c) { // Object? with NNBD.
      if (c) return "thing";
      return null;
    }
    ```

* Dart requires final fields to be definitely initialized in the constructor initialization list, even when `null` is a valid value. Users don't seem to mind having to put an explicit `= null` there. This may be confounded by them having different preferences between final and non-final fields, but I'm not sure about that.

* I've definitely had a number of users over the years tell me they are surprised by the Effective Dart rule that tells them to rely on default initialization to `null`. They felt it should have been made explicit.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -484,25 +487,24 @@ fields on `Object`.
...
+constructors declares an instance variable with a potentially non-nullable type
+and no initializer expression unless the variable is marked with the `late`
+modifier.
 
-It is an error if a potentially non-nullable local variable which has no
-initializer expression and is not marked `late` is used before it is definitely
-assigned (see Definite Assignment below).
+It is an error to derive a mixin from a class declaration which contains an
+instance variable with a potentially non-nullable type and no initializer
+expression unless the variable is marked with the `late` modifier.
```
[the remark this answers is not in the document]
leafpetersen: Didn't we already do this once?  This is your text, not mine.  I'm just touched the whitespace and github diffs suck.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -556,18 +558,9 @@ expression.
...
-forms of assignments, including assignments via the composite assignment
-operators as well as pre and post-fix operators.
-
-It is a compile time error to read a local variable marked `late` when the
-variable is **definitely unassigned**. This includes all forms of reads,
-including implicit reads via the composite assignment operators as well as pre
-and post-fix operators.
+It is not a compile time error to write to a `final` non-local variable or field
+if that variable or field is declared `late` and does not have an initializer.
+For local variables, see the section below.
```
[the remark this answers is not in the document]
leafpetersen: This isn't new text.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: Yes, these are not exclusive.  I introduced these terms in response to feedback that just using "not definitely assigned" (i.e. could be "Either" or could be "Definitely unassigned") was confusing.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,442 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+| late var x;       | Ok            | Ok                  | Ok              |
+| late final x;     | Error         | Ok                  | Ok              |
+| late T x;         | Ok            | Ok                  | Ok              |
+| late final T x;   | Error         | Ok                  | Ok              |
+
+### Local variables and initialization based inference
+
+Local variables with explicitly written types are given the declared types as
+written.  The declared type of the variable is considered a "type of interest"
+in the sense defined in the flow analysis specification.  If the variable has an
```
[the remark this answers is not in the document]
leafpetersen: Yes.  To be clear, my intention is that the flow analysis spec be definitive.  But the combination of the flow analysis spec not being generally readable + having bugs at the moment drove me to write a textual description in the feature spec.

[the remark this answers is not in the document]
leafpetersen: And yes, I intend there be follow up to land this in the flow analysis spec.

accepted/future-releases/nnbd/feature-specification.md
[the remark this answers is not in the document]
leafpetersen: Yes, lines 690 + 693 are intended to make this code non-erroneous.  @stereotype441 if you don't think I've correctly captured that we should discuss further.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,442 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+```
+
+At join points, some adjustment to the **initialization inferred type** is done
+to account for nullability.  Informally, a non-`late`, `final` **untyped**
+variable must be assigned a value on every path, but a variable declared `late`
+and/or `var` may be assigned a value on only some of the paths into a join.  In
+the latter case, the type is made nullable if the variable is declared as
+non-`late`, and the variable is treated as implicitly null initialized.
+Variables declared `late` are only made nullable if `Null` is explicitly
+assigned on one more paths into the join, and there is no implicit null
```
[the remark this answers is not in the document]
leafpetersen: None of this is normative.  I can see that the "and there is no" is badly worded though, I will follow up if this PR stays live.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -616,6 +609,442 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+into the join (in which case it does not need to be made implicitly nullable),
+or that it has at least one reachable path with no **initialization inferred
+type** (in which case it must be made implicitly nullable if it is non-`late`).
+
+At a join point in the program, if an **untyped variable** has an
+**initialization inferred type** on all of the *reachable* paths into the join,
+then the **initialization inferred type** for the variable after the join point
+is computed as follows:
+  - If the **initialization inferred type** on each path which has one
+    (reachable or not) is the same type `T` (using syntactic equality), then the
```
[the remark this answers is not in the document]
leafpetersen: We can use **NORM** but I'm not sure we really want to.   To give a precise accounting of what the analyzer and the CFE are doing, we need a semi-normal form which collapses multiple `??` as appropriate.  This is already implicitly assumed all over the places, so I'm inclined to just assume that here.  But maybe **NORM** is reasonable?  I'll think about it.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: As discussed, `Null` is a solution for `T` in `T` or `T?` or `Null`, and so this proposal also gives `Null` as the solution here.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: Also `Null`.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: It depends on the exact details of the proposal, but for this variant at least, yes:

var x;
if (true) {
  while(b) {
    x = 3; // Not an initializing write
  }
  // x is still untyped
} else {
  x = "hello"; // x inferred to String
} // x is untyped on one branch, `String` on another, so infer `String?`
x?.length; // x is a non-null int, what does this mean?
```

File: accepted/future-releases/nnbd/feature-specification.md [hunk 32]
```diff
@@ -616,6 +609,442 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+assign to a **definitely unassigned** `final` local variable.
+
+It is a compile time error to assign a value to a `final`, `late` local variable
+if it is **definitely assigned**. Thus, it is *not* an error to assign to a
+**potentially unassigned** `final`, `late` local variable.
+
+*Note that a variable is always considered **definitely assigned** and not
+**definitely unassigned** if it has an explicit initializer, or an implicit
+initializer as part of a larger construct (e.g. the loop variable in a `for in`
+construct).*
```
eernstg, line 637: This is yet another place where model 3 (https://github.com/dart-lang/language/issues/946#issuecomment-649229421) would require an exception: With `var x = null;`, `x` would be considered definitely unassigned, or no assignment after the declaration could be considered an initializing assignment.

accepted/future-releases/nnbd/feature-specification.md
eernstg: I think those terms are helpful for the overall readability.

accepted/future-releases/nnbd/feature-specification.md
lrhn: Yes, I'm thinking we'll need to have intiailziation/typing separate from assignment.

A variable is either typed, untyped (with nullable hint or not), or initialization-typed (or maybe also invalid-typed if we want to continue after incompatible joins).

If it's untyped and assigned to, then it becomes initialization-typed. We can join initialization-typed paths for variables, maybe untyped ones too (if we want to).

leafpetersen:
Abandoning this, we didn't take this approach.

PR #1046, opened 2020-06-24
More spec cleanup

mit-mit:
Follow-up to https://github.com/dart-lang/language/pull/1043


mit-mit:
Landing this to fix broken link

PR #1053, opened 2020-06-29
Fix flow analysis handling of write-captured variables in loops and try-blocks.

stereotype441:
This change ensures that we properly reject the following unsound program:

```dart
f(Object x) {
  void Function()? g;
  while (true) {
    if (x is int) {
      g?.call();
      print(x.isEven); // UNSOUND
    }
    g = () {
      x = 'bad';
    }
  }
}
```

(The program is unsound because the second time through the loop, a
closure exists that writes to `x`, therefore after (A) there is no
guarantee that `x` is still an `int`.)

Note that this bug has already been fixed in the implementation (see
https://github.com/dart-lang/sdk/commit/88b76997c49b6a86bec2ef5afc6e97b329e5a2eb).

### DIFF
--- a/resources/type-system/flow-analysis.md
+++ b/resources/type-system/flow-analysis.md
@@ -9,6 +9,11 @@ https://docs.google.com/document/d/11Xs0b4bzH6DwDlcJMUcbx4BpvEKGz8MVuJWEfo_mirE/
 
 ## CHANGELOG
 
+2020.06.29
+  - Fix handling of variables that are write captured in loops, switch
+    statements, and try-blocks (such variables should be conservatively assumed
+    to be captured).
+
 2020.06.02
   - Specify the interaction between downwards inference and promotion/demotion.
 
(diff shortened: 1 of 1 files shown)


resources/type-system/flow-analysis.md
```diff
@@ -329,12 +338,17 @@ We also make use of the following auxiliary functions:
...
+  approximation of the flow model that could result from joining `M` with a
+  model in which variables in `written` might have been written to and variables
+  in `captured` might have been write-captured.  It is defined as
+  `FlowModel(r, VI1)` where `M` is `FlowModel(r, VI0)` and `VI1` is the map
   such that:
     - `VI0` maps `v` to `VM0 = VariableModel(d0, p0, s0, a0, u0, c0)`
-    - If `S` contains `v` then `VI1` maps `v` to `VariableModel(d0, [], s0, a0, u0, c0)`
+    - If `captured` contains `v` then `VI1` maps `v` to
+      `VariableModel(d0, [], s0, a0, u0, true)`
+    - Otherwise if `written` contains `v` then `VI1` maps `c` to
```
leafpetersen, line 350: should be "`VI1` maps `v` to"?

Issue #1059, opened 2020-07-02
Generators returning `void` is no longer an error


eernstg:
Consider the following program:

```dart
void h() sync* { }
main() => h();
```

This program causes a compile-time error with the CFE, but no errors in the analyzer. The reported error is "Functions marked 'sync*' can't have return type 'void'", which seems to be a useful rule, because it is difficult to make the case that a generator returning `void` is useful.

In [generalized-void.md](https://github.com/dart-lang/sdk/blob/master/docs/language/informal/generalized-void.md) we had the following rule:

> It is a compile-time error if a function marked `async*`
> or `sync*` has return type void.

The behavior of the CFE is consistent with this rule.

My interpretation is that this rule should have been included when 'generalized-void.md' was integrated into the language specification, but that didn't happen.

Given that the CFE is already enforcing this rule, it should not be a breaking change to re-introduce it, so I'd recommend that we adjust the specification to have this rule, and ask for an implementation of the rule in the analyzer.

@leafpetersen, @munificent, @lrhn, @natebosch, @stereotype441, @jakemac53, WDYT?

eernstg:
Created PR https://github.com/dart-lang/language/pull/1060; when I have feedback here, I'll either delete that PR (if the feedback is negative) or request a review of that PR (if it is positive).

lrhn:
Seems reasonable to disallow `void` generators. 

Unlike `async` functions, a generator has no effect if its return value is ignored, so the entire function body could be replaced by `{}` without changing behavior. We don't make dead code an error, but in this case I'm willing to make an exception.

stereotype441:
SGTM

eernstg:
Thanks! -- got the feedback, landed #1060.

eernstg:
@johnniwinther, just FYI: The CFE reports a compile-time error for a generator function returning `void` (e.g., `Functions marked 'sync*' can't have return type 'void'.`). The spec has just been updated to specify this behavior. So if there are any issues requesting that the CFE should stop reporting this error (because that's what the spec used to imply), those issues are now obsolete.

natebosch:
Do we have an issue filed to add this error in the analyzer?

stereotype441:
> Do we have an issue filed to add this error in the analyzer?

Not that I'm aware of.  Filing one now

stereotype441:
> > Do we have an issue filed to add this error in the analyzer?
> 
> Not that I'm aware of. Filing one now

[...]

PR #1060, opened 2020-07-02
Add rule about generators and return type void

eernstg:
Cf. https://github.com/dart-lang/language/issues/1059, this PR adds the rule that it is a compile-time error for a generator function to have return type `void`.

If and when this lands, update https://github.com/dart-lang/sdk/issues/32192.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -1362,9 +1362,17 @@ This situation is very unusual so it is not worth making an exception to the gen
 }
 
 \LMHash{}%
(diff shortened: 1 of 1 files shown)


leafpetersen:
cc @scheglov .  Tests and implementation issues to follow.

srawlins:
Analyzer implementation issue is dart-lang/sdk#32192; CFE already implements this.

scheglov:
Thank you, https://dart-review.googlesource.com/c/sdk/+/153953 should fix it.

PR #1061, opened 2020-07-02
Add implementation plan for external and abstract variables

eernstg:


eernstg:
The FFI team want to use the external/abstract variable feature, and we have discussed when to do it as well as how long it would take. This PR adds an implementation plan for this feature. PTAL and say whether you think it would work.

It does not include any specific timeline, that part is to be settled separately such that it fits into everyone's plans.

accepted/future-releases/abstract-external-fields/implementation-plan.md
```diff
@@ -0,0 +1,153 @@
...
+* [Tracking issue](https://github.com/dart-lang/sdk/issues/42560)
+* [Feature specification](https://github.com/dart-lang/language/blob/master/accepted/future-releases/abstract-external-fields/feature-specification.md)
+
+## Phase 0 (Preliminary steps)
+
+### "external-abstract-variables" Experimental flag
+
+The implementation of this feature is hidden behind an
+[experiment flag][].
+The flag `--enable-experiment=external-abstract-variables`
```
johnniwinther, line 16: Is there a connection between the non-nullable flag and this flag (will one imply the other)?

dart:ffi uses external fields to support nnbd so they needs together.

eernstg: Yes, the need gets acute with null-safety, and it's much more like a convenience for legacy code. For instance, `int x;` is an error with null-safety, so there's a need to initialize `x` (even though that's just useless noise, as seen from the perspective that `x` only exists in order to induce a getter and a setter).

So the intention is that this feature only exists when null-safety is enabled. I assume it's helpful to make this strict (such that there's no need to support the feature at all in legacy mode), so I've added that to the document.

johnniwinther: The current CFE support for external fields is guarded by `--enable-experiment=non-nullable`.

eernstg:
@leafpetersen, do you wish to request any changes here?

accepted/future-releases/abstract-external-fields/implementation-plan.md
```diff
@@ -0,0 +1,157 @@
...
+
+### Grok
+
+Update Grok to handle the new AST.
+
+### Analysis server
+
+No additional work is expected for the analysis server.
+
+### Atom plug-in
```
devoncarew, line 99: This is no longer current. Github's syntax highlighting is now driven from here: https://github.com/dart-lang/dart-syntax-highlight/tree/master/grammars

I suggest we rename this item to 'update github syntax highlighting' and include a reference to the file I linked to. This is the same grammar that drives VS Code as well. I don't think we need a separate item in this impl. plan for VS Code, but the PR that updates the grammar should cc Danny Tuppany (or open an issue on the Dart-Code/Dart-Code repo. He uses the same file, but a copy of it, so will need to know to update his copy.

accepted/future-releases/abstract-external-fields/implementation-plan.md
```diff
@@ -0,0 +1,157 @@
...
+
+The [Dart Atom plug-in][atom] has a grammar for syntax highlighting
+Dart code in Atom. This same grammar is also used for syntax highlighting
+on GitHub. Update this to handle the new syntax.
+
+[atom]: https://github.com/dart-atom/dart
+
+### VS Code
```
devoncarew, line 107: This section can be removed (or, re-worded to reference the above section).

accepted/future-releases/abstract-external-fields/implementation-plan.md
```diff
@@ -0,0 +1,157 @@
...
+
+Update the IntelliJ parser to support the new syntax.
+
+### Grok
+
+Update Grok to handle the new AST.
+
+### Analysis server
+
+No additional work is expected for the analysis server.
```
devoncarew, line 97: 'validate that `external` shows up in code completion'?

accepted/future-releases/abstract-external-fields/implementation-plan.md
```diff
@@ -0,0 +1,157 @@
+# Implementation Plan for "External and Abstract Variables"
```
devoncarew, line 1: I don't see an entry for dartdoc in here; is there expected to be no change?

accepted/future-releases/abstract-external-fields/implementation-plan.md
```diff
@@ -0,0 +1,157 @@
...
+Define and implement formatting rules for the new syntax.
+Add formatting tests.
+
+### Formal specification
+
+Add a formal specification of this feature to the language specification.
+
+### Cider
+
+Validate Cider support.
```
devoncarew, line 85: FYI, Cider syntax highlighting uses codemirror (as does DartPad); https://github.com/codemirror/CodeMirror/blob/master/mode/dart/dart.js.

accepted/future-releases/abstract-external-fields/implementation-plan.md
eernstg: Thanks! Done.

accepted/future-releases/abstract-external-fields/implementation-plan.md
eernstg: Reworded as suggested.

accepted/future-releases/abstract-external-fields/implementation-plan.md
eernstg: Thanks, done!

accepted/future-releases/abstract-external-fields/implementation-plan.md
eernstg: No, it's likely that there will be some work. For instance, `///` can be placed on a declaration using new syntax. Added a section.

accepted/future-releases/abstract-external-fields/implementation-plan.md
eernstg: Thanks again for all this information! Included this, too.

accepted/future-releases/abstract-external-fields/implementation-plan.md
leafpetersen: I think we should consider just keeping this under the non-nullable flag and committing to ship it as part of the null safety release given the current timing.  cc @mit-mit @devoncarew for thoughts on this.

leafpetersen:
cc @franklinyow

accepted/future-releases/abstract-external-fields/implementation-plan.md
devoncarew: I have no concerns with that -

eernstg: It sounds like we're leaning toward reusing `--enable-experiment=non-nullable` for this feature, to ensure that it is inseparable from null-safety.

eernstg:
This PR now includes templates/implementation_meta_issue.txt with updates based on Devon's input.

templates/implementation_meta_issue.txt
```diff
@@ -20,16 +20,15 @@ The implementation plan for this issue is <link to implementation plan>.
...
-- [ ] Issue for implementation in Cider:
 - [ ] Issue for implementation in Dartfmt:
 - [ ] Issue for implementation in Dartdoc:
-- [ ] Issue for dart-atom (github support):
 - [ ] Issue for implementation in build systems and Angular compiler:
 - [ ] Issue for VM Service support:
 - [ ] Issue for web debugging support:
-- [ ] Issue for [codemirror support](https://github.com/codemirror/CodeMirror/) (dartpad):
+- [ ] Issue for [codemirror support](https://github.com/codemirror/CodeMirror/) (Cider, dartpad):
 - [ ] Issue for [highlight.js support](https://github.com/highlightjs/highlight.js) (api.dart.dev and api.pub.dev):
```
devoncarew, line 32: nit: dartdoc also uses highlight.js

eernstg: Thanks, added that.

eernstg:
I've changed the experiment flag to be 'non-nullable', which basically means that this feature is considered to be part of the null-safety feature bundle.

@leafpetersen, @devoncarew, @johnniwinther, @stereotype441, are you happy to have this landed now?

accepted/future-releases/abstract-external-fields/implementation-plan.md
eernstg: Changed the text to use 'non-nullable'.

stereotype441:
> @leafpetersen, @devoncarew, @johnniwinther, @stereotype441, are you happy to have this landed now?

lgtm!

PR #1065, opened 2020-07-03
Update to new number operation rules

lrhn:
After writing tests, I gave the rules a serious look.

They were both too complex and too imprecise.

I changed them so that `x + y` will take the type of `x` where possible (not `int + double`, but `int + int`, `double + num` etc.)
This ensures that a variable which has a type variable (possibly promoted) as type will be safe to do `x += 1` and `x++` because such an operation *always* has the same type as `x` (as long as `x` is a non-`Never` subtype of `num`).

That makes the rules asymmetric, but I think most users will readily accept that the type of the LHS is more important than the RHS. It is a method call after all.
(Most users will also not have type variables that are subtypes of `num` or `int` at all).

### DIFF
--- a/accepted/future-releases/nnbd/number-operation-typing.md
+++ b/accepted/future-releases/nnbd/number-operation-typing.md
@@ -73,21 +73,21 @@ We extend the special-casing rules of `+`, `-`, `*` and `%` to also cover calls 
 Let `e` be an expression of one of the forms `e1 + e2`, `e1 - e2`, `e1 * e2`, `e1 % e2` or `e1.remainder(e2)`, where the static type of `e1` is a non-`Never` type *T* where *T* <: `num` and the static type of `e2` is *S*. Then:
 
 * If *T* <: `double` , then the static type of `e` is *T*.
-* Otherwise if *S* is a non-`Never` subtype of `double` then the static type of `e` is `double`.
+* Otherwise, if *S* is a non-`Never` subtype of `double` then the static type of `e` is `double`.
 * Otherwise, if *S* is a non-`Never` subtype of  `int` then the static type of `e` is *T*.
-* Otherwise if is a non-`Never` subtype of *T* , then the static type of `e` is *T*.
+* Otherwise, if *S* is a non-`Never` subtype of *T*  then the static type of `e` is *T*.
 * Otherwise the static type of *e* is `num`.
 
 We also special-case the `clamp` method.
 
-Let `e` be a normal invocation of the form `e1.clamp(e2, e3)`, where the static types of `e1`, `e2` and `e3` are *T*<sub>1</sub>, *T*<sub>2</sub> and *T*<sub>3</sub> respectively, and where  *T*<sub>1 is a non-`Never` subtype of `num`. Let *R* be LUB(*T*<sub>1</sub>, *T*<sub>2</sub>, *T*<sub>3</sub>). Then:
+Let `e` be a normal invocation of the form `e1.clamp(e2, e3)`, where the static types of `e1`, `e2` and `e3` are *T*<sub>1</sub>, *T*<sub>2</sub> and *T*<sub>3</sub> respectively, and where  *T*<sub>1</sub> is a non-`Never` subtype of `num`. Let *R* be LUB(*T*<sub>1</sub>, *T*<sub>2</sub>, *T*<sub>3</sub>). Then:
 
-* If *R* <: `num`, the static type of`e` as *R*.
+* If *R* <: `num`, the static type of `e` as *R*.
 * Otherwise the static type of `e` is `num`.
 
 With these typing rules, we cover all the instance members of `int` which has a return type of `num`, and we ensure that a using operands with the *same* type gives a result of that type, even if that type is a type variable (like `X extends int`) or promoted type variable (like `X & int`).
 
-There are no special rules for `/` and `~/` because their return type is not `num`, and the return value's type is independent of the argument types. A `/` operation always returns a `double` and a `~/` operation always returns an `int`.ma
+There are no special rules for `/` and `~/` because their return type is not `num`, and the return value's type is independent of the argument types. A `/` operation always returns a `double` and a `~/` operation always returns an `int`.
 
 The rules for the binary operators can be summarized (non-normatively) as:
 
@@ -112,15 +112,16 @@ If `e` is an expression of the form  `e1 + e2`, `e1 - e2`, `e1 * e2`, `e1 % e2` 
 
 * If *C* <: `int` and *T* <: `int`, then the context type of `e2` is `int`. 
 * If *C* <: `double` and *T* is not a subtype of `double`, then the context type of `e2` is `double`.
+* Otherwise, the context type of `e2` is `num`.
 
-It is a compile-time error if the static type of `e2` is not a subtype of `num`. *(It is not necessarily a compile-time error if the static type of `e2` is not a subtype of _C_.)*
+*(It is not necessarily a compile-time error if the static type of `e2` is not a subtype of _C_, but it is still a compile-time error if the static type of `e2` is not assignable to the actual parameter type,`num`.)*
 
 If `e` is an expression of the form `e1.clamp(e2, e3)` where *C* is the context type of `e` and *T* is the static type of `e1` where both *T* is a non-`Never` subtype of `num`, then:
 
 * If *C* is a non-`Never` subtype of `num` then  the context type of `e2` and `e3` is *C*.
 * Otherwise the context type of `e2` an `e3` is `num`
 
-It is a compile-time error if the static type of `e2` or `e3` is not a subtype of `num`.  *(It is not necessarily a compile-time error if the static type of `e2` or `e3` is not a subtype of _C_.)*
+*(It is not necessarily a compile-time error if the static type of `e2` or `e3` is not a subtype of _C_, but it is still a compile-time error if the static type of `e2` or `e3` is not assignable to the actual parameter type,`num`.)*
 
 (This does emphasize the inherent non-symmetry of Dart operators: The first operand is a receiver which is always evaluated with no type context, and is then used to resolve the operator method against, and the second operand is an argument to that method. We need to fully resolve the first operand and the operator before we can even begin with the second operand.)
 
--- a/accepted/future-releases/nnbd/number-operation-typing.md
+++ b/accepted/future-releases/nnbd/number-operation-typing.md
@@ -10,7 +10,7 @@ That's not something the general Dart type system can capture, so Dart special-c
 
 The special-cased typing works because the language enforces that the only subclasses of `num` is `int` and `double`, and those cannot have subclasses at all. Even though operators are otherwise virtual, the only two `+` operations that can possibly be in play for `numValue + something`  are the `int.operator+` and `double.operator+`  which have known behavior compatible with the special typing, and `intValue + something` is known to call the `int.operator+` method. (This is also the reason those operators can be used in constant expressions: Their behavior is known at compile-time.).
 
-The current rules do not cover all integer operations, or any double operations, and this means that a few operations have the default type of `num` inherited from the `num` interface. That has so far not been a serious problem since Dart has had *implicit downcasts* which allows `num` to be assigned to both `int` and `double`. Some static checking my be lacking, but it still runs.
+The original Dart 2.0 rules do not cover all integer operations, or any double operations, and this means that a few operations have the default type of `num` inherited from the `num` interface. That has so far not been a serious problem since Dart has had *implicit downcasts* which allows `num` to be assigned to both `int` and `double`. Some static checking my be lacking, but it still runs.
 
 With Null Safety, we remove implicit downcasts from the language. This causes some existing, functioning code to become invalid, and currently the only workaround is adding an explicit cast.
 
@@ -24,11 +24,11 @@ See [language#971][], [language#597][], [sdk#41559][], [sdk#39652][], [sdk#32645
 
 The special-case typing rules only apply to arithmetic *operators* (`+`, `-`, `*`, `%`). They do not apply to `int.remainder`, even though it is otherwise equivalent to `%`,and they do not apply to `num.clamp`. These are the two remaining members of `int` which has a return type of `num`.
 
-This has caused issues before, but now those issues become compile-time errors because of the lack of [impliciit downcasts][sdk#39652].
+This has caused issues before, but now those issues become compile-time errors because of the lack of [implicit downcasts][sdk#39652].
 
 ### Rules do not work with type variables
 
-An operation like `T add<T extends num>(T a, T b) => a + b;` becomes invalid with Null Safety because the type of `a + b` is num, not `T`, and the implicit down-cast from `num` to `T` is no longer valid.
+An operation like `T add<T extends num>(T a, T b) => a + b;` becomes invalid with Null Safety because the type of `a + b` is `num`, not `T`, and the implicit down-cast from `num` to `T` is no longer valid.
 
 ### No special case for `num op double`
 
@@ -43,7 +43,7 @@ expectsDouble(y); // Used to be implicit downcast, now a compile-time error.
 
 ### Inference doesn't know special rules
 
-Code like `int n = …; double y = n * 2;` is a compile-time error. The type context for `2` is `num`. The author expected `2` to be a double literal, and it's possible to recognize that *making* it a double literal would make the code correct. Not all instances of this problem are as obvious as this one, for example: `double y = n * await Future(() => 2.0);` will currently infer the future to be a `Future<num>` independently of the operation.
+Code like `int n = …; double y = n * 2;` is a compile-time error. The type context for `2` is `num`. The author expected `2` to be a double literal, and it's possible to recognize that *making* it a double literal would make the code correct. Not all instances of this problem are as obvious as this one, for example: `double y = n * await Future(() => 2.0);` will infer the future to be a `Future<num>` independently of the operation.
 
 ## Solution goal
 
@@ -60,7 +60,7 @@ double lerp(num start, num end, double y) => start + (end - start) * y;
 T add<T extends num>(T a, T b) => a + b;
 ```
 
-because any attempt to plug in actual values will give sound results.
+because any attempt to plug in actual values will give sound results. They will expect `x += 1;` to work no matter which numeric type `x` has.
 
 Users also understand that `clamp` will return either the receiver or one of the arguments. If those all have the same type, the result will have that type.
 
@@ -70,7 +70,7 @@ Users also understand that `clamp` will return either the receiver or one of the
 
 We extend the special-casing rules of `+`, `-`, `*` and `%` to also cover calls of the `remainder` method, and to also work with type parameters which extend `num` , `int` or `double`. Finally, if the second operand is a `double` and the first is a `num`, the result is guaranteed to be a `double`.
 
-Let `e` be an expression of one of the forms `e1 + e2`, `e1 - e2`, `e1 * e2`, `e1 % e2` or `e1.remainder(e2)`, where the static type of `e1` is a non-`Never` type *T* where *T* <: `num` and the static type of `e2` is *S*. Then:
+Let `e` be an expression of one of the forms `e1 + e2`, `e1 - e2`, `e1 * e2`, `e1 % e2` or `e1.remainder(e2)`, where the static type of `e1` is a non-`Never` type *T* and *T* <: `num`, and where the static type of `e2` is *S* and *S* is assignable to `num`. Then:
 
 * If *T* <: `double` , then the static type of `e` is *T*.
 * Otherwise, if *S* is a non-`Never` subtype of `double` then the static type of `e` is `double`.
@@ -80,7 +80,7 @@ Let `e` be an expression of one of the forms `e1 + e2`, `e1 - e2`, `e1 * e2`, `e
 
 We also special-case the `clamp` method.
 
-Let `e` be a normal invocation of the form `e1.clamp(e2, e3)`, where the static types of `e1`, `e2` and `e3` are *T*<sub>1</sub>, *T*<sub>2</sub> and *T*<sub>3</sub> respectively, and where  *T*<sub>1</sub> is a non-`Never` subtype of `num`. Let *R* be LUB(*T*<sub>1</sub>, *T*<sub>2</sub>, *T*<sub>3</sub>). Then:
+Let `e` be a normal invocation of the form `e1.clamp(e2, e3)`, where the static types of `e1`, `e2` and `e3` are *T*<sub>1</sub>, *T*<sub>2</sub> and *T*<sub>3</sub> respectively, and where  *T*<sub>1</sub>, *T*<sub>2</sub>, and *T*<sub>3</sub> are all non-`Never` subtypes of `num`. Let *R* be LUB(*T*<sub>1</sub>, *T*<sub>2</sub>, *T*<sub>3</sub>). Then:
 
 * If *R* <: `num`, the static type of `e` as *R*.
 * Otherwise the static type of `e` is `num`.
@@ -123,7 +123,7 @@ If `e` is an expression of the form `e1.clamp(e2, e3)` where *C* is the context 
 
 *(It is not necessarily a compile-time error if the static type of `e2` or `e3` is not a subtype of _C_, but it is still a compile-time error if the static type of `e2` or `e3` is not assignable to the actual parameter type,`num`.)*
 
-(This does emphasize the inherent non-symmetry of Dart operators: The first operand is a receiver which is always evaluated with no type context, and is then used to resolve the operator method against, and the second operand is an argument to that method. We need to fully resolve the first operand and the operator before we can even begin with the second operand.)
+These rules emphasize the inherent non-symmetry of Dart operators: The first operand is a receiver which is always evaluated with no type context, and is then used to resolve the operator method against, and the second operand is an argument to that method. We need to fully resolve the first operand and the operator before we can even begin with the second operand.
 
 For the binary operators, the context type of the second operand, based on the first operand and the context type of the entire operation, can be summarized (non-normatively) as:
 
@@ -149,17 +149,13 @@ The static type of `lhs += e` is the static type of `lhs + e`.
 
 In general the static type of `x + integer` is the same as the static type of `x` when the type is a subtype of `num`, exactly to allow compound assignment to work no matter which number type is on the left-hand side. 
 
-*It's possible to have a setter with a less specific argument type than the type of the corresponding getter, like `set foo(num x); int get x;`. That is not a problem for these rules, it merely means that for `x += e`, equivalent to `x = x + e`, the context type might not be the same as the type of  reading `x`. The typing and context rules are compatible with this.*
+*It's possible to have a setter with a less specific argument type than the type of the corresponding getter, like `set foo(num x); int get foo;`. That is not a problem for these rules, it merely means that for `foo += e`, equivalent to `foo = foo + e`, the context type might not be the same as the type of  reading `x`. The typing and context rules are compatible with this.*
 
 ### Increment/Decrement Operators
 
 The prefix and suffix increment and decrement operators (`++x`, `x++`, `--x` and `x--`) are roughly equivalent to expressions containing `x + 1` or `x - 1`, and assignments of that back to `x` (the only difference is whether the result value is the value of `x` before or after the assignment). 
 
-If `e` is an assignable expression with a static type *T* which is a non-`Never` subtype of `num`, then the static type of `e++`, `e--`, `++e` and `--e` is always *T*.
-
-The value of of `x++` and `x--` is the value of `x`, so that follows directly.
-
-The value of `++x` and `--x` is the value of `x + (1 as int)` or `x - (1 as int)`, which is the same as the type of `x` for any subtype of `num`.
+*If `e` is an assignable expression with a static type T which is a non-`Never` subtype of `num`, then the static type of `e++`, `e--`, `++e` and `--e` is always T. This follows from the current semantics where the type of `e++` and `e--` is the static type of `e`, and `++e` and `--e` are equivalent to `e += 1` or `e -= 1`.*
 
 ## Summary
 
--- a/accepted/future-releases/nnbd/number-operation-typing.md
+++ b/accepted/future-releases/nnbd/number-operation-typing.md
@@ -75,7 +75,7 @@ Let `e` be an expression of one of the forms `e1 + e2`, `e1 - e2`, `e1 * e2`, `e
 * If *T* <: `double` , then the static type of `e` is *T*.
 * Otherwise if *S* is a non-`Never` subtype of `double` then the static type of `e` is `double`.
 * Otherwise, if *S* is a non-`Never` subtype of  `int` then the static type of `e` is *T*.
-* Otherwise if *S* <: *T* and *T* <: *S*, then the static type of `e` is *T*.
+* Otherwise if is a non-`Never` subtype of *T* , then the static type of `e` is *T*.
 * Otherwise the static type of *e* is `num`.
 
 We also special-case the `clamp` method.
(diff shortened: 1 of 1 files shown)


accepted/future-releases/nnbd/number-operation-typing.md
```diff
@@ -62,47 +62,76 @@ T add<T extends num>(T a, T b) => a + b;
...
+> If the context type of `e1.clamp(e2, e3)` is a non-`Never` subtype of `num`, *T*, then the context types of `e2` and `e3` are also *T*. Otherwise the context types of `e2` and `e3` are `num`.
 
 (This does emphasize the inherent non-symmetry of Dart operators: The first operand is a receiver which is always evaluated with no type context, and is then used to resolve the operator method against, and the second operand is an argument to that method. We need to fully resolve the first operand and the operator before we can even begin with the second operand.)
 
+Tabulated, the context type of the second operand, based on the first operand and the context type of the entire operation, is:
+
+| op1 \\ ctx  | int             | int+            | double | double+         | num  | num+            |
+| :---------: | --------------- | --------------- | ------ | --------------- | ---- | --------------- |
+|   **int**   | int             | int<sup>*</sup> | double | num<sup>*</sup> | num  | num<sup>*</sup> |
+|  **int+**   | int             | int             | double | num<sup>*</sup> | num  | num<sup>*</sup> |
```
eernstg, line 127: Column 3 (context `int+`) should be `int<sup>*</sup>` as well, I think: `e1 + e2` will have type `int`, not anything denoted by `int+`.

accepted/future-releases/nnbd/number-operation-typing.md
```diff
@@ -62,47 +62,76 @@ T add<T extends num>(T a, T b) => a + b;
...
+| **double**  | num<sup>*</sup> | num<sup>*</sup> | num    | num<sup>*</sup> | num  | num<sup>*</sup> |
+| **double+** | num<sup>*</sup> | num<sup>*</sup> | num    | num             | num  | num<sup>*</sup> |
+|   **num**   | num<sup>*</sup> | num<sup>*</sup> | double | num<sup>*</sup> | num  | num<sup>*</sup> |
+|  **num+**   | num<sup>*</sup> | num<sup>*</sup> | double | num<sup>*</sup> | num  | num             |
```
eernstg, line 131: Of course, all those `T+` types must denote the _same_ type when they occur as the type of the receiver as well as the context type. However, in the case where both the receiver static type and the context type is some `U <: num`, it would actually be sound to give `e` the static type `U`. So I guess that would be expressed as `num+` in the last column in line 131.

It would also be satisfied by the rules if we return to the two subtype based rules: `e1` has type `U`, the context type is `U`, and `e2` is subject to type inference with context type `U`, which succeeds and yields an `e2` of type `U`.

accepted/future-releases/nnbd/number-operation-typing.md
```diff
@@ -111,11 +140,13 @@ These extensions also carry over to the compound assignment operators and increm
 
 An `lhs += e` expression is roughly equivalent to `lhs = lhs + e` except that subexpressions of `lhs` are only evaluated once. The static typing rules that apply to `lhs + e` also applies to `lhs += e`, and similarly for the other binary operators that are special cased.
 
-That means that `int x = 0; x += 1;` works because `x += 2` is equivalent to `x =  x + 2`, and `x + 2` has type `int`.
+That means that `int x = 0; x += 2;` works because `x += 2` is equivalent to `x =  x + 2`, and `x + 2` has type `int`.
```
eernstg, line 143: Maybe cover the case `double d = 0; d += 2;` as well. `d += 2.0` will work fine, but with `2` we'd get context type `num` and `2` will have static type `int`, and `d = d + 2` has no errors.

But will it ever be a problem that `2` is an `int` here, when it will often have type `double` as soon as it gets within a few inches of anything double?

accepted/future-releases/nnbd/number-operation-typing.md
```diff
@@ -62,47 +62,76 @@ T add<T extends num>(T a, T b) => a + b;
...
 We also special-case the `clamp` method:
 
 > For a normal invocation `e` of the form `e1.clamp(e2, e3)`, where the static types of `e1`, `e2` and `e3` are *T*<sub>1</sub>, *T*<sub>2</sub> and *T*<sub>3</sub> respectively, which are all non-`Never` subtypes of `num`:
 >
-> * If any of *T*<sub>1</sub>, *T*<sub>2</sub> or *T*<sub>3</sub> is a supertype of the other two, the first such is the static type of `e`.
+> * If any of *T*<sub>1</sub>, *T*<sub>2</sub> or *T*<sub>3</sub> is a supertype of the other two, the first such (in source order of the corresponding operands) is the static type of `e`.
 > * Otherwise, if all of *T*<sub>1</sub>, *T*<sub>2</sub> and *T*<sub>3</sub> are subtypes of `int`, the static type of `e` is `int`.
 > * Otherwise if all of *T*<sub>1</sub>, *T*<sub>2</sub> and *T*<sub>3</sub> are subtypes of `double`, the static type of `e` is `double`.
-> * Otherwise the static type of `e` is `num`.
+> * Otherwise the static type of `e` is `num`.
```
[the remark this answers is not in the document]
lrhn: It should be safe. Are we sure it will always be a subtype of `num`?

accepted/future-releases/nnbd/number-operation-typing.md
```diff
@@ -62,47 +62,76 @@ T add<T extends num>(T a, T b) => a + b;
...
+> For an expression `e` of one of the forms `e1 + e2`, `e1 - e2`, `e1 * e2`, `e1 % e2` or `e1.remainder(e2)`, where the static type of `e1` is a non-`Never` type *T* where *T* <: `num` and the static type of `e2` is a non-`Never` type *S* where *S* <: `num`:
 >
-> * If S <: T then the static type of `e` is *T*. 
-> * Otherwise If *T* <: *S* then the static type of `e` is *S*.
-> * Otherwise if *T* <: `int` and *S* <: `int` then the static type of `e` is `int`.
-> * Otherwise if *T* <: `double` or *S* <: *double* then the static type of *e* is `double`.
+> * If *T* <: `int` and *S* <: `int` then the static type of `e` is *T*.
+> * Otherwise if *T* <: `double` , then the static type of `e` is *T*.
+> * Otherwise if *S* <: `double` then the static type of `e` is `double`.
+> * Otherwise if *T* is not a subtype of `int` or `double`, then the static type of `e` is *T*.
```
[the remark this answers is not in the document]
lrhn: Ack, so I though of that the first time around.

But LUB is not what I want here, because that doesn't work.

Say x has type *T* = `X&int` and y has type *S* = `int`, then `x += y` should be valid. That only works if we take the type of *T* for the int+int case and the double+anything case.

I guess the rule should be:

* If *S* is a non-`Never` subtype of `int`, then the static type of `e` is *T*.  *(Anything + int is the first thing again, true and sound for all of `int`, `double` and `num`)*.
* Otherwise, if `T` is a non-`Never` subtype of `double`, then the static type of `e` is *T*. *(`Double` + anything is the same double again)*
* Otherwise, if *S* is a non-`Never` subtype of `double`, then the static type of `e` is `double`. *(Anything + `double` is `double`).
* Otherwise the static type of `e` is `num`.

That also fits better with how I'm thinking about it.

accepted/future-releases/nnbd/number-operation-typing.md
```diff
@@ -62,47 +62,76 @@ T add<T extends num>(T a, T b) => a + b;
...
+
+
+| op1 \\ op2  | int     | int+    | double  | double+ | num     | num+    |
+| :---------: | ------- | ------- | ------- | ------- | ------- | ------- |
+|   **int**   | int     | int     | double  | double  | num     | num     |
+|  **int+**   | int+    | int+    | double  | double  | num     | num     |
+| **double**  | double  | double  | double  | double  | double  | double  |
+| **double+** | double+ | double+ | double+ | double+ | double+ | double+ |
+|   **num**   | num     | num     | double  | double  | num     | num     |
+|  **num+**   | num+    | num+    | double  | double  | num+    | num+    |
```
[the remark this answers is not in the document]
lrhn: ACK. With the rewrite above, the table would be:

| op1 \\ op2  | int     | int+    | double  | double+ | num     | num+    |
| :---------: | ------- | ------- | ------- | ------- | ------- | ------- |
|   **int**   | int     | int     | double  | double  | num     | num     |
|  **int+**   | int+    | int+    | double  | double  | num     | num     |
| **double**  | double  | double  | double  | double  | double  | double  |
| **double+** | double+ | double+ | double+ | double+ | double+ | double+ |
|   **num**   | num     | num     | double  | double  | num     | num     |
|  **num+**   | num+    | num+    | double  | double  | num    | num/num+    |

(Only change is the two last `num`'s in the final row which are no longer `num+`. The `num/num+` covers the case where the two operands have the *same* type which extends num.).

accepted/future-releases/nnbd/number-operation-typing.md
```diff
@@ -62,47 +62,76 @@ T add<T extends num>(T a, T b) => a + b;
...
 
-> If the context type of `e1.clamp(e2, e3)` is a non-`Never` subtype of `num`, *T*, and the static type of `e1` is a non-`Never` subtype of *T*, then the context types of `e2` and `e3` are both *T*.<br>Otherwise, if the context type of the invocation is `int` and the static type of `e1` is a non-`Never` subtype of `int`, then the context types of `e2` and `e3` are both `int`, and if the context type of the invocation is `double` and the static type of `e1` is a non-`Never` subtype of `double`, then the context types of `e2` and `e3` are both `double`. Otherwise the context types of `e2` and `e3` are `num`.
+> If the context type of `e1.clamp(e2, e3)` is a non-`Never` subtype of `num`, *T*, then the context types of `e2` and `e3` are also *T*. Otherwise the context types of `e2` and `e3` are `num`.
```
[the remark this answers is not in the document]
lrhn: ACK, true. Cut that down a little too far.

accepted/future-releases/nnbd/number-operation-typing.md
lrhn: if `e1` and `e2` have the same type, then the result has that type.

So, if both `e1` and the context type are the same a proper subtype of `int`, then so must `e2` be. So, it should be `int+`.

accepted/future-releases/nnbd/number-operation-typing.md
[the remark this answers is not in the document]
eernstg: The new rules make sense to me!

accepted/future-releases/nnbd/number-operation-typing.md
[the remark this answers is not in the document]
eernstg: The signature of `clamp` is `num clamp(num, num)` in all of `int`, `double`, `num`, so the invocation is an error unless *T2*, *T3* is assignable to `num`, and all these rules only apply if *T1* is a subtype of `num`.

So the interesting cases are when some of *T2*, *T3* are dynamic. In that case the actual arguments must still pass `arg as num`.

So it would probably be most helpful to say `min(num, LUB(T1, T2, T3))`, and that would be sound.

accepted/future-releases/nnbd/number-operation-typing.md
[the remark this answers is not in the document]
eernstg: I think these tables should be non-normative (and given that feature specs usually don't make that distinction very strictly, we should probably say it in words), because they involve the subtleties around `int+` and `int+` being, say, `X extends int` and `Y & int` and similar cases where it isn't just the same type. The remaining text should be complete anyway.

accepted/future-releases/nnbd/number-operation-typing.md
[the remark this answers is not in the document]
lrhn: It does miss the part where "if *S* and *T* is the same type, there result is *T*." which allows `T add<T extends num>(T a, T b) => a + b;` to work.
Maybe it needs to be phrased as:

> * Otherwise, If *S* <: *T* and *T* <: *S*, the static type of `e` is *T*.
> * Otherwise, the static type of `e` is `num`.

accepted/future-releases/nnbd/number-operation-typing.md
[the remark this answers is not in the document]
lrhn: What if the static type of *T2* and *T3* are `T extends Comparable<num>&double` and `T extends Comparable<num>&int` respectively. Is the LUB of that `Comparable<num>&num` or just `Comparable<num>`?

accepted/future-releases/nnbd/number-operation-typing.md
eernstg: OK, it is actually sound: We could for instance have a receiver whose static type is a type variable `X` and dynamic type _S_ is "JavaScript number" which is a proper subtype of `int`, but the result would also be of type _S_. So even if the actual value of `X` could somehow be _S_ it is still sound. Similarly, the treatment of `Never` is also sound.

accepted/future-releases/nnbd/number-operation-typing.md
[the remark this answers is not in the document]
eernstg: How would that be possible? The bound of a type variable is declared, and there is no way to denote those `&` types.

accepted/future-releases/nnbd/number-operation-typing.md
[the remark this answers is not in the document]
eernstg: Yep!

accepted/future-releases/nnbd/number-operation-typing.md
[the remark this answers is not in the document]
lrhn:
```dart
void foo<C extends Comparable<num>>(C x, C y) {
  num z = 0;
  if (x is int && y is double) {
    print([x, y].runtimeType);  // prints "List<Comparable<num>>"
    print([x, y, z].runtimeType);  // prints "List<Comparable<num>>"
    print([z.clamp(x, y)].runtimeType);  // prints "List<num>"
  }
}
main() {
  foo<Comparable<num>>(1, 1.5);
}
```

Here the static type of `x` is `C extends Comparable<num> & int`, and the static type of `y` is `C extends Comparable<num> & double`. The output suggests that the LIUB of those (and `num`) is actually `Comparable<num>`.

The static type of `z.clamp(x, y)` should still be `num`.
I'll handle this by saying that if the LUB is not <: `num`, the static type is `num`. Then the static type is still sound, even when the inputs are outrageous. (Won't happen in practice, but it should still be sound).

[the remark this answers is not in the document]
eernstg: Ah, the notation `C extends Comparable<num> & int` is new, we'd usually write `C & int`.

> I'll handle this by saying that if the LUB is not <: num, the static type is num.

Sounds good, that would be the same thing that I intended when I wrote `min(num, LUB(T1, T2, T3))`.

accepted/future-releases/nnbd/number-operation-typing.md
lrhn: Probably not a problem in practice.

The only situation I can see is:

```dart
double x = something;
x += 147573952589676410000;  // 2^67
```
This will fail to compile because the context is `num`. It could compile if we made the context `double`.
It would only make a difference for literals that are actually valid double values, and an optimizing compiler can do that without any rules.

So, we could make the context type of `e2` in `dbl + e2` be `double` because we know that the first operation on the value will be `toDouble()`. 

It would affect an integer literal: `dbl + 2` would become `dbl + 2.0`, but that would happen anyway.
And it would affect `dbl + await Future.value(2);` which would ... possibly ... make that `await Future.value(2.0)`, but again no real value.

I don't think it would change anything in practice, so let's not go there.

eernstg: Makes sense.

eernstg:
@lrhn, I think you are not waiting for input from me on this one. True?

lrhn:
@eernst Not unless you have some new insights or comments on the latest change.

I am waiting to see if @stereotype441 or @johnniwinther have comments.

accepted/future-releases/nnbd/number-operation-typing.md
```diff
@@ -62,47 +62,80 @@ T add<T extends num>(T a, T b) => a + b;
...
 
 We extend type inference to take the special number rules into account.
 
-> For  `e1 + e2`, `e1 - e2`, `e1 * e2`, `e1 % e2` or `e1.remainder(e2)` where the static type of  `e1` is a non-`Never` subtype of `int`, if the context type of the entire expression is `int`, then the context type of `e2` is `int`. If the static type of `e1` is a non-`Never` subtype of `num` that is *not* a subtype of `double` and the context type of the entire expression is `double`, then the context type of `e2` is `double`.
+If `e` is an expression of the form  `e1 + e2`, `e1 - e2`, `e1 * e2`, `e1 % e2` or `e1.remainder(e2)`, where *C* is the context type of `e` and *T* is the static type of `e1`, and where both *C*  and *T* are non-`Never` subtypes of `num`, then:
+
+* If *C* <: `int` and *T* <: `int`, then the context type of `e2` is `int`. 
+* If *C* <: `double` and *T* is not a subtype of `double`, then the context type of `e2` is `double`.
+
+It is a compile-time error if the static type of `e2` is not a subtype of `num`. *(It is not necessarily a compile-time error if the static type of `e2` is not a subtype of _C_.)*
```
stereotype441, line 116: Is this normative?  It seems like it's unnecessary, because it follows from the normal rules for analyzing method arguments.

accepted/future-releases/nnbd/number-operation-typing.md
```diff
@@ -62,47 +62,80 @@ T add<T extends num>(T a, T b) => a + b;
...
 
-> If the context type of `e1.clamp(e2, e3)` is a non-`Never` subtype of `num`, *T*, and the static type of `e1` is a non-`Never` subtype of *T*, then the context types of `e2` and `e3` are both *T*.<br>Otherwise, if the context type of the invocation is `int` and the static type of `e1` is a non-`Never` subtype of `int`, then the context types of `e2` and `e3` are both `int`, and if the context type of the invocation is `double` and the static type of `e1` is a non-`Never` subtype of `double`, then the context types of `e2` and `e3` are both `double`. Otherwise the context types of `e2` and `e3` are `num`.
+If `e` is an expression of the form `e1.clamp(e2, e3)` where *C* is the context type of `e` and *T* is the static type of `e1` where both *T* is a non-`Never` subtype of `num`, then:
+
+* If *C* is a non-`Never` subtype of `num` then  the context type of `e2` and `e3` is *C*.
+* Otherwise the context type of `e2` an `e3` is `num`
+
+It is a compile-time error if the static type of `e2` or `e3` is not a subtype of `num`.  *(It is not necessarily a compile-time error if the static type of `e2` or `e3` is not a subtype of _C_.)*
```
stereotype441, line 123: As above, is this normative?  It seems unnecessary.

accepted/future-releases/nnbd/number-operation-typing.md
lrhn: It can be either normative or not, since it states the same thing as the default rules. Except that it forgot about `dynamic`, so it's bad.
Let's remove the compile-time error requirement here, and just non-normatively state that the normal rules apply for assignability to the parameter type.

The question is what effect the context type has here.
It's not a requirement, just a hint.
What if the operand is `dynamic`, is it implicitly downcast to `num` or to *C*? 

It probably should be downcast to *C* because otherwise;

```dart
dynamic x = 1;
double x = 1 + x;  
```
would be invalid. `1+x` has type `num` unless `x` has type `double`, but with a context type of `double` on `x`, if we implicitly downcast it to `double`, then the code is allowed, and it will throw at the correct point.

Likewise:
```dart
dynamic x = 1;
int z = 1 + x;  // invalid unless we downcast `x` to `int`, then it is valid.
```

So, do we want to downcast `dynamic` to the context type. Is that what happens automatically?

If so, do we want to make it a compile-time error for:
```dart
int x = 1 + e;
```
if the static type of `e` is not `int`? Or just rely on the static type of `1 + e` not being `int` to get the error? The former would likely give the more precise error message.

WDYT? How much effect should the context type have:
* None, it's just a hint, not satisfying it won't matter.
* Cause downcast of `dynamic` to *C*.
* Cause compile-time error if static type is not *C*.
* Both of the above.

?

eernstg: > So, do we want to downcast `dynamic` to the context type.

I think we'd usually say that inference uses a context type, and the type of `x` is still `dynamic` when inference is complete. Then we compile `1 + x` to an invocation of `operator +` where it is checked dynamically that the value of `x` has type `num`. This is the standard treatment of expressions of type `dynamic` passed as actual arguments (or assigned to variables, etc.), so there's no need to re-specify it here.

It is possible for inference to fail, and we do report this as a compile-time error in some cases:

```dart
class A<X extends A<X>> {}
void main() => A();
```

If inference succeeds, I believe we can consider subsequent steps as independent of inference: Say, if `e` becomes `e1` during inference, then we just check the normal rules (like assignability) on `e1`, and it doesn't matter that the source code is actually `e` and `e1` is obtained by inference. A downcast in an initialization or assignment would be one of those subsequent steps, and we don't need to re-specify them.

In any case, I think the newest wording (where the paragraph shown as starting in line 116 above is all-commentary) is fine.

lrhn: ACK, so what we would need in order to support `int x = 1 + dyn;` is to say:

* If *T* <: `int` and *C* <: `int` and *S* is `dynamic`, then the static type of `e` is `int` and it's a *run-time error* if `e2` evaluates to a value which is not an `int`.
* If *C* <: `double` and *T* not <: `double`, then the static type of `e` is `double` and it's a run-time error if `e2` evaluates to a value which is not a `double`.

It would require combining the context and typing rules because then the typing rules depend on the context too.

We can do that. Should we?

stereotype441: I want `int x = 1 + dyn;` to be a compile-time error in opted-in code.  It's an implicit downcast from `num` to `int`, and we don't allow implicit downcasts in opted-in code.  In my mind it's no different from something like this:
```dart
class C {
  Object m() => ...;
}
main() {
  C c = C().m(); // Error: implicit downcast from `Object` to `C`
}
```
Similarly, for `double x = 1 + dyn;`, it should be a compile time error, because it's an implicit cast from `num` to `double`.

I gues this means that the user won't be able to do `int x = ...; x += dyn;`, because that desugars to `x = x + dyn;`, and the static type of `x + dyn` is `num`.  I'm ok with that.

lrhn: I can see that this is compatible with how we treat `dynamic` otherwise.
The static type of the expression is `dynamic`, and that's assignable to any type. The fact that an argument is `dynamic` doesn't change the return type of a function call (unless inference is involved).

However, it is a little curious to me that we insert an implicit downcast, but we don't actually change the type of the expression. For `int foo(int x) => x; dynamic y = 1; foo(y);`, we say that the last invocation is equivalent to the explicit downcast `foo(y as int)`, but it isn't because the latter changes the static type of the expression.

We could instead say that the implicit cast inserted by the type system actually *changes* the static type of the expression, as if a real `as contextType` was inserted.

In that case, `int x = 1 + dyn;` would be equivalent to `int x = 1 + (dyn as int);`, and that code would work.

I hadn't though of `int x = ...; x += dyn;`, but if anything, that makes me want it to work. :)

stereotype441: Ok, I see where you are coming from; you're right that `int x = 1 + (dyn as int);` would work, and it does seem reasonable to make `int x = 1 + dyn;` behave equivalently to that, because the only values for `dyn` that wouldn't result in a runtime error are ints.  That does a lot to ease my concerns.

I still feel a bit weird about it, though, and I can't decide if I'm just being stubborn or if there is a real user trap here.  Perhaps the crux of my concern is that in general, it's not safe to assume that a real `as contextType` gets inserted when there is a dynamic downcast, because when assigning to a promoted variable, we use a context type that's more specific than what's required, e.g.:
```dart
T f<T>(List<T> values) => values.first;
test(List<num> x) {
  dynamic y = <num>[];
  if (x is List<int>) {
    x = []; // Downward context is `List<int>` so this gets inferred as `<int>[]`
    x = y; // Ok; assignment de-promotes
  }
}
```

eernstg: I like the consistency of @stereotype441's approach: We use the special-cased context type to make `e2` an `int` when type inference can achieve that, but we don't try to make `e2` an `int` using any other means. When type inference does make `e2` an `int`, the developer can _trust_ this outcome.

This makes it natural for `1 + dyn` to apply the pure, signature based type check at `num` (turning it into `1 + (dyn as num)`), and not adding an implicit downcast all the way down to `int`. Consequently, as @stereotype441 mentioned, the initialization of `int x` has a further downcast from `num` to `int`, and that's an error with null-safety.

So it may seem tempting to "make it work in two more cases", but I think it's more important to keep the distinction between the safe steps (where type inference actually yields an expression that safely has the required type) and unsafe steps (where we implicitly add a dynamic type check). I think it's a better quality of service to ask developers to make those downcasts explicit.

accepted/future-releases/nnbd/number-operation-typing.md
```diff
@@ -62,47 +62,81 @@ T add<T extends num>(T a, T b) => a + b;
...
 
 ## Solution
 
 ### Extended Rules
 
-We extend the special-casing rules of `+`, `-`, `*` and `%` to also cover calls of the `remainder` method, and to also work with type parameters which extend `num` , `int` or `double`. Finally, if the second operand is a `double` and the first is a `num`, the result is guaranteed to be a `double`.
+We extend the special-casing rules of `+`, `-`, `*` and `%` to also cover calls of the `remainder` method, and to also work with type parameters which extend `num` , `int` or `double`. Finally, if the second operand is a `double` and the first is a `num`, the result is guaranteed to be a `double`.
 
-That is:
+Let `e` be an expression of one of the forms `e1 + e2`, `e1 - e2`, `e1 * e2`, `e1 % e2` or `e1.remainder(e2)`, where the static type of `e1` is a non-`Never` type *T* where *T* <: `num` and the static type of `e2` is *S*. Then:
```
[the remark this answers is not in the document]
lrhn: I actually want `T&double + dynamic` to have type `T&double`.  I can't do that if I pre-constrain *S* to being a subtype of `num`.

accepted/future-releases/nnbd/number-operation-typing.md
leafpetersen: While I think overall the experience of inserting casts to the context type as deep as possible is probably better, I'm not sure it's a good idea to do so here when we don't do so in other places.  e.g. this code which is accepted and wouldn't be if we "pushed the downcast down":

```dart
void main() {
  dynamic d = "world";
  int x = (true) ? "hello" : d;
}
```

accepted/future-releases/nnbd/number-operation-typing.md
```diff
@@ -1,22 +1,22 @@
...
 
-The current rules do not cover all integer operations, or any double operations, and this means that a few operations have the default type of `num` inherited from the `num` interface. That has so far not been a serious problem since Dart has had *implicit downcasts* which allows `num` to be assigned to both `int` and `double`. Some static checking my be lacking, but it still runs.
+The current rules do not cover all integer operations, or any double operations, and this means that a few operations have the default type of `num` inherited from the `num` interface. That has so far not been a serious problem since Dart has had *implicit downcasts* which allows `num` to be assigned to both `int` and `double`. Some static checking my be lacking, but it still runs.
```
leafpetersen, line 13: "Current" is confusing, since once this lands, these become the current rules.  Consider "Previous to this update, the rules did not cover..."?

accepted/future-releases/nnbd/number-operation-typing.md
leafpetersen: As a counter-point to my own commentary, if we view this as an ad hoc use of overloading, then there are two things happening here:

1) A static overload resolution process which decides which version of (e.g.) `operator +` is being called
2) Normal inference + type checking.  

In this model, it makes sense that the cast would be on the second argument, since once you have determined the correct overload, the *type* of the overload is what really defines the context type for the argument, and just as with any other method invocation, we insert a cast to the parameter type if the inferred static type of the argument is not a subtype of the inferred (overload resolved) parameter type.

accepted/future-releases/nnbd/number-operation-typing.md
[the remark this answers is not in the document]
eernstg: Which is exactly the reason why I said "assignable to" and not "subtype of".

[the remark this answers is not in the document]
lrhn: Ah. :)

I'm still not sure why it's error prone to not make requirements on the type of the operand before deciding that it's a number operation based on the receiver. 

If this is a number operation (the receiver is a non-`Never` subtype of `num`, then we use these rules for the static type of the operation instead of using the normal rule based on the declaration of `+`, which would give `num` or `double`).

The second operand must still be assignable to num, otherwise it's a compile-time error, for all the normal reasons.

accepted/future-releases/nnbd/number-operation-typing.md
lrhn: I agree that I'm thinking of this in terms of overloading.

Now, if we ever get actual overloading, we might want these rules to fit into that.

If so, what we are seeing here is overloading based on context type (return type) and receiver. That might be a little far to go. Java does not have return type based overloading (they could, the [byte code allows methods that differ only in return type](https://community.oracle.com/docs/DOC-983207), but they don't base resolution on return type).

With only receiver+parameter based overloading, we can't choose between `double +(int, double)` and `int +(int, int)` when faced with `int+dynamic`.

(Then again, without self types, we can't handle the `Self operator+(int)`, which the rules here also specify, so we are already stretching overloading in order to make `x += 1` work for a any numeric type).

accepted/future-releases/nnbd/number-operation-typing.md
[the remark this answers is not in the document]
eernstg: Maybe it should be made explicit that the usual static checks apply.

Given that the type of `e1 + e2` does not follow the normal rules, it is not obvious that the static checks that we normally apply to a member invocation are applicable to `e1 + e2`. And we can't just say "Consider a program with no compile-time errors" because the new special case rules about numbers are needed in order to eliminate some errors that would otherwise exist. 

If you say "where *S* is assignable to `num`" then we'd explicitly say that this whole set of special rules don't apply otherwise, and that would make those other cases an error according to all the normal rules whenever appropriate.

Without that, I think we're saying (line 75) that `42.0 + "Hello!";` is correct and has static type `double`.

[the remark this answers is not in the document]
lrhn: Reasonable. Let me try to rephrase it.

accepted/future-releases/nnbd/number-operation-typing.md
```diff
@@ -111,11 +145,21 @@ These extensions also carry over to the compound assignment operators and increm
...
+
+*It's possible to have a setter with a less specific argument type than the type of the corresponding getter, like `set foo(num x); int get x;`. That is not a problem for these rules, it merely means that for `x += e`, equivalent to `x = x + e`, the context type might not be the same as the type of  reading `x`. The typing and context rules are compatible with this.*
 
 ### Increment/Decrement Operators
 
-The prefix and suffix increment and decrement operators (`++x`, `x++`, `--x` and `x--`) are roughly equivalent to expressions containing `x + 1` or `x - 1`. The static typing rules that apply to `+` also apply to `--` and `++`, with the caveat that the `1` is not treated as a number literal that can potentially be a `double`, even if we apply a context type of `double` to the position. The `1` is *always* an integer literal.
+The prefix and suffix increment and decrement operators (`++x`, `x++`, `--x` and `x--`) are roughly equivalent to expressions containing `x + 1` or `x - 1`, and assignments of that back to `x` (the only difference is whether the result value is the value of `x` before or after the assignment). 
+
```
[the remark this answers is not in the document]
lrhn: Correct. The current semantic state that the type of `e++` and `e--` is the static type of `e`, and that `++e` and `--e` are equivalent to `e += 1` and `e -= 1`. That's still correct, no need for changes here.

accepted/future-releases/nnbd/number-operation-typing.md
lrhn: Decision is that we do not make special rules for dynamic.

PR #1070, opened 2020-07-06
Clarify static behavior of `super == other` under NNBD.

stereotype441:
See https://github.com/dart-lang/sdk/issues/42603.

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -650,6 +650,11 @@ Consider an expression `e` of the form `e1 == e2` where the static type of
 formal parameter of `operator ==` in the interface of **NonNull**(`T1`).
 It is a compile-time error unless `T2` is assignable to `S?`.
 
+Similarly, consider an expression `e` of the form `super == e2`, inside a class
+`C`, where the static type of `e2` is `T2`.  Let `S` be the type of the formal
+parameter of `operator ==` in the super-interface of `C`.  It is a compile-time
+error unless `T2` is assignable to `S?`.
+
 _Even if the static type of `e1` is potentially nullable, the parameter type
 of the `operator ==` of the corresponding non-null type is taken into account,
 because that instance method will not be invoked when `e1` is null. Similarly,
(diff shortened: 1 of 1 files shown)


accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -650,6 +650,11 @@ Consider an expression `e` of the form `e1 == e2` where the static type of
 formal parameter of `operator ==` in the interface of **NonNull**(`T1`).
 It is a compile-time error unless `T2` is assignable to `S?`.
 
+Similarly, consider an expression `e` of the form `super == e2`, inside a class
+`C`, where the static type of `e2` is `T2`.  Let `S` be the type of the formal
+parameter of `operator ==` in the super-interface of `C`.  It is a compile-time
+error unless `T2` is assignable to `S?`.
```
eernstg, line 656: I agree with Leaf that it would be nice to combine the two cases if possible, but I suspect that we have to keep them separate because the lookup for a superinvocation is somewhat delicate.

The point is that superinvocations must ignore `implements` links, transitively, and only search through the superclass chain for declarations, and it must ignore all abstract declarations in the superclass chain. This kind of search is what the language specification denotes by a 'method lookup', so we need to mention that concept in order to obtain the correct member signature. For instance, the superinterface could have a `bool operator ==(covariant SomeType)` signature, but the declaration which is actually invoked could still be `bool operator ==(Object)`.

Maybe we could phrase it as follows:

Consider an expression `e` of the form `super == e2` that occurs in a class
whose superclass is `C`, where the static type of `e2` is `T2`. Assume that
method lookup for `operator ==` in `C` yields a concrete declaration where
the formal parameter has type `S`. It is a compile-time error unless `T2` is
assignable to `S?`.

stereotype441: @eernstg I like your approach.  What would you think of this minor rephrasing to the parenthetical comment:

> Consider an expression `e` of the form `super == e2` that occurs in a class
> whose superclass is `C`, where the static type of `e2` is `T2`. Let `S` be the
> formal parameter type of the concrete declaration of `operator ==` found by
> method lookup in `C` (_if that search succeeds, otherwise it is a compile-time error_).
> It is a compile-time error unless `T2` is assignable to `S?`.

eernstg: LGTM, and reads more naturally.

stereotype441: Ok, wording updated.  @leafpetersen are you ok with this?

leafpetersen: LGTM

Issue #1067, opened 2020-07-06
permit eager loading of deferred imports


mraleph:
Current spec describes deferred imports in a way which mandates issuing dynamic errors if deferred prefix is accessed before the `loadLibrary` is called.

This implies code size penalty even if no libraries are actually deferred, at minimum code needs to be generated to check the status of the library prefix and state of the library prefix needs to be stored somewhere. This accounts for as much as 0.4% on large internal apps - and is essentially completely dead payload.

I propose to amend the spec with something like the following paragraph: 

> An implementation is free to treat _all_ deferred imports in the program as if they were eagerly loaded i.e. `loadLibrary` was invoked on each and every single one of deferred prefixes in the program before execution of the main begins.

Note: this all or nothing mode - either all deferred imports are eagerly loaded or none of them, which makes it predictable and understandable. 

This is more or less how Dart VM implementation behaved since the switch to Kernel based pipeline - so amending the spec simply ratifies Dart VM behaviour. 

/cc @lrhn @leafpetersen @rmacnak-google

eernstg:
I do not have strong opinions on this, but [here](https://github.com/dart-lang/language/pull/1071)'s a PR that proposes how we could say it.

lrhn:
I'd state it as:

> A deferred library import with prefix *p* may be loaded lazily at the compiler's discretion. 
> Accessing any declaration through the prefix *p* other than *p.loadLibrary* causes a run-time error
> if evaluated before the deferred import has been loaded.
> Invoking <code>*p*.loadLibrary</code> will cause the import to be loaded, 
> and the returned future completes when the library has been loaded.
> *(A compiler may choose to eagerly load the deferred library, and if so, accessing the imported 
> declarations will not cause a run-time error even though <code>*p*.loadLibrary</code> has not 
> been invoked yet.)*

I don't want to have "all or none" requirements. Either all deferred imports require a `loadLibrary`, or they don't. If they don't, I see no reason to make any further requirements.

A deferred library is like a global `late` variable. It's up to the user to use it correctly, we can't help them because that would require global flow analysis.

leafpetersen:
cc @rakudrama @sigmundch

sigmundch:
I feel that making optional would defeat the purpose of this check - we wanted it so that you can have consistent semantics regardless of whether code was actually deferred.

DDC doesn't have this check today but we do want to add it: we want to prevent users from finding in production that they accidentally accessed something before it was available.

[...]

PR #1071, opened 2020-07-07
Added spec language that deferred imports can be eager

eernstg:
Proposal for how we could allow deferred imports to be handled eagerly. Awaiting a decision in https://github.com/dart-lang/language/issues/1067.


eernstg:
[Proposal](https://github.com/dart-lang/language/issues/1067) dropped, closing this PR.

PR #1079, opened 2020-07-09
Spec null-safety combined member signature and spread type

eernstg:
To maintain the result of discussions: This PR adds a specification of how to compute the combined member signature with null-safety, and how to compute various associated types for a `<spreadElement>` with null-safety.

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -687,9 +687,12 @@ Dart.
 
 In a collection literal in Dart before null-safety, the inferred element
 type of a spread element of the form `...?e` where `e` has static type
-`Null` is `Null`, and so are the inferred key type and value type. With
-null-safety, when the static type of `e` is `Null` or a subtype thereof, the
-inferred element, key, and value type is `Never`
+`Null` is `Null`, and so are the inferred key type and value type.
+
+With null-safety, when the static type of `e` is `Null` or a potentially
+nullable subtype thereof, the inferred element, key, and value type
+of `...?e` is `Never`, and so is the element, key, and value type of
+`...e` where the static type of `e` is a subtype of `Never`.
 (which corresponds to 6 changes in
 [this section](https://github.com/dart-lang/language/blob/9e12517922c1f0021aead2af163c3b502497f312/specification/dartLangSpec.tex#L8433)).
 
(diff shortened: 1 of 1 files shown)


eernstg:
Tests for null spreads are [here](https://dart-review.googlesource.com/c/sdk/+/154693).

eernstg:
Tightened the rule to cover `...?e` where `e` has types like `Never` and `X extends Null` as well.

eernstg:
@leafpetersen, do you maintain the approval with the changes of spread element typing?

munificent:
This LGTM too. Just to confirm, this aligns the spec with https://github.com/dart-lang/sdk/issues/42304, right?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -679,6 +683,19 @@ If no type is specified in a catch clause, then the default type of the error
 variable is `Object`, instead of `dynamic` as was the case in pre-null safe
 Dart.
 
+#### Spread element typing
+
+In a collection literal in Dart before null-safety, the inferred element
+type of a spread element of the form `...?e` where `e` has static type
+`Null` is `Null`, and so are the inferred key type and value type.
+
+With null-safety, when the static type of `e` is `Null` or a potentially
```
leafpetersen, line 692: First off, I'm confused by the phrasing here.  I think you're trying to cover two cases: `...?e` where `e` has certain types, and `...e` where `e` has certain types?  As a textual thing, I'd suggest to split these out into two separate paragraphs for clarity, even at the expense of some redundancy.

Second, I'm not sure this covers all the cases.  The first sentence covers `...?e` where `e` has type `Null`, or `X extends Null` or `X & Null`, and I think nothing else.  The second sentence covers `...e` when `e` has type `Never` or `X extends Never` or `X & Never`.  I think the second syntax with the first set of types is an error?  But shouldn't the second set of types with the first syntax also be valid and have type `Never`?  I think it's specified as a warning, but not an error, right?

eernstg:
> this aligns the spec with dart-lang/sdk#42304, right?

Yes, with `Never` rather than `Null` the situations in that issue are covered.

accepted/future-releases/nnbd/feature-specification.md
eernstg: Indeed, fixed that.

Issue #1012, opened 2020-06-08
Change most specified warnings to errors


eernstg:
Cf. https://github.com/dart-lang/sdk/issues/36127 and https://github.com/dart-lang/sdk/issues/36126, the language specification specifies a few compile-time warnings:

- It is a static warning if the return type of a user-declared operator `[]=` is explicitly declared and not `void`. (E)
- It is a static warning if a setter declares a return type other than `void`. (E)
- It is a static warning if a class has a setter named `v=` with argument type T and a getter named `v` with return type S, and S may not be assigned to T. (E)
- It is a static warning if an overriding method declaration declares a different default value for a parameter. (W)
- A switch statement may yield static warnings for various non-exhaustive lists of cases. (W)
- It is a static warning to import two different libraries with the same name unless their name is the empty string. (E)

However, the treatment of these warnings has been inconsistent for a long time.

The situations marked '(E)' are currently flagged as errors by the analyzer, and '(W)' is flagged as a warning. Given that the analyzer has flagged these situations as errors for a long time, we have the option to specify them as errors without breaking existing code.

This issue serves to discuss and clarify this situation.

Here is a concrete proposal: Change the specification to specify as errors those situations that are currently flagged as errors, as indicated above.

The treatment of the getter/setter type conflict is specified for null-safety, so it will be an error soon anyway, and similarly for the switch cases.

The examples in https://github.com/dart-lang/sdk/issues/36127 would be adjusted and added as language tests.

@lrhn, @munificent, @leafpetersen, WDYT?

lrhn:
I agree with making the `(E)` ones errors.

I still want to make the "different default value" not a warning at all. If it's not an error, then it's not a *problem*. People who wants the warning can still get it from the analyzer as a lint, if they want to, but then they also have a way avoid it. They can't avoid spec-mandated warnings.

The switch warning on incomplete enums is probably fine as a warning (but I'd prefer to make it non-spec mandated and leave it up to the analyzer, and try to avoid *warnings* in the spec).

munificent:
> I agree with making the `(E)` ones errors.

Me too.

eernstg:
@leafpetersen, @natebosch, @jakemac53, @stereotype441, do you wish to speak up against making the '(E)' cases errors in the spec?

stereotype441:
Making the '(E)' cases errors SGTM.

leafpetersen:
LGTM.

eernstg:
[...]

PR #1083, opened 2020-07-10
Changed warnings to errors, cf. language issue #1012

eernstg:
Addendum to #1012: It turns out that there is support in the language team for deleting the rule about `library` name clashes in imports, rather than changing it from a warning to an error. Hence, this PR deletes that rule.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -29,6 +29,7 @@
 % 2.8
 % - Change several warnings to compile-time errors, matching the actual
 %   behavior of tools.
+% - Eliminate error for library name conflicts.
 %
 % 2.7
 % - Rename non-terminals `<...Definition>` to `<...Declaration>` (e.g., it is
@@ -16398,18 +16399,7 @@ The use of a class or interface, and of its members, is separate from its declar
 The usage and declaration may occur in widely separated places in the code, and may in fact be authored by different people or organizations.
 It is important that errors are given at the offending declaration so that the party that receives the error can respond to it a meaningful way.
 
-In contrast a library comprises both imports and their usage; the library is under the control of a single party and so any problem stemming from the import can be resolved even if it is reported at the use site.
-
-%On a related note, the provenance of the conflicting elements is not considered. An element that is imported via distinct paths may conflict with itself. This avoids variants of the well known "diamond" problem.
-}
-
-\LMHash{}%
-It is a compile-time error to import two different libraries with the same name
-unless their name is the empty string.
-
-\commentary{
-A widely disseminated library should be given a name that will not conflict with other such libraries.
-The preferred mechanism for this is using pub, the Dart package manager, which provides a global namespace for libraries, and conventions that leverage that namespace.
+In contrast a library comprises both imports and their usage; the library is under the control of a single party and so any problem stemming from the import can be resolved even if it is reported at the use site.%
 }
 
 \commentary{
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -2786,8 +2790,8 @@ \subsection{Setters}
 }
 
 \LMHash{}%
-It is a static warning if a setter declares a return type other than \VOID{}.
-It is a static warning if a class has
+It is a compile-time error if a setter declares a return type other than \VOID.
```
lrhn, line 2793: The `[]=` text says "explicitly declared" and this just says "declared".

Consider making the phrasing consistent, so readers won't think there is a meaningful difference.

specification/dartLangSpec.tex
```diff
@@ -2786,8 +2790,8 @@ \subsection{Setters}
...
+It is a compile-time error if a class has
 a setter named \code{$v$=} with argument type $T$ and
 a getter named $v$ with return type $S$,
 and $S$ may not be assigned to $T$.
```
lrhn, line 2797: So we do allow a `dynamic` getter and a non-top-type setter. Yey?

specification/dartLangSpec.tex
```diff
@@ -16401,7 +16405,8 @@ \subsection{Imports}
 }
 
 \LMHash{}%
-It is a static warning to import two different libraries with the same name unless their name is the empty string.
+It is a compile-time error to import two different libraries with the same name
```
lrhn, line 16408: "to import" suggests that this only matters for a single library, not transitively?

specification/dartLangSpec.tex
eernstg: Adjusted the text about `[]=` --- we don't have any other occurrence of 'explicitly declare{s,d}' in the language specification outside commentary.

specification/dartLangSpec.tex
eernstg: Yes, this must be an oversight: The rule doesn't apply to non-class getters and setters. But the null-safety rules cover all setter/getter pairs, so we would presumably just leave legacy code as it is and get the cleanup with null-safety (rather than starting a breaking change process with legacy code).

Also the old rules allow a `dynamic` getter and any setter parameter type, but that's not a surprise: The nnbd feature spec explicitly says that it is _now_ a subtype rule, not an assignability rule.

specification/dartLangSpec.tex
eernstg: That is actually the implemented behavior.

This rule would be motivated by the ability to detect that the same library has been imported twice in ways that makes the execution consider them as distinct (all the way from slightly different URIs to an actual copy of the file).

So it's pretty much a useless rule when it is not checked transitively.

Also, it's not likely to matter much to anybody, because the use of `library` at all has had very little support for several years.

So it's hardly worth the trouble to fix it up (and handle breakage wherever it might arise).

Maybe we should stick to the plan and just change the spec to make it an error, such that the spec follows the actual behavior of the analyzer?

eernstg: Created https://github.com/dart-lang/language/issues/1084, suggesting that we delete this rule. It's impressively useless.

eernstg: Resolving this discussion, it's handled in #1084.

eernstg:
The latest commit on this PR removes the library name clash error.

PR #1091, opened 2020-07-14
Final definite assignment

leafpetersen:
Final variant/proposal for definite assignment based inference.

This variant requires explicit initialization on all branches, but allows adjustment for Null.


leafpetersen:
cc @lrhn @eernstg @jakemac53 @natebosch @munificent @stereotype441

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -621,6 +614,488 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+    var z;
+    do {
+      // z is potentially assigned.
+      if (b) {
+        z = 42;
+      } else {
+        z = 37;
+      }
+    } while (something);
+    print(z); // Error, type cannot be inferred for z
```
eernstg, line 1071: The `do` loop is actually a bit more analyzable than a `while` loop because it is guaranteed to run at least once. So if the body considered without the back-edge gives rise to a typing, it should be sound to use that. The fact that it may repeat will still need to be taken into account, of course, such that we'll get an error if the loop body is writing to a `final` variable. But that might just be an extra run of the flow analysis on the loop body in a state where `z` is considered potentially assigned, checking for those "don't reassign" constraints.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -621,6 +614,488 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+  var x; // x is an untyped variable, documented as having type int?
+  var z; // z is an untyped variable, documented as having type Object?
+  if (b) {
+    x = 3; // x is inferred to int
+    return;
+  } else if (b) {
+    return;  // Note, x has no inferred type on this path
+  } else {
+    // x not assigned
+  } // x has type `int?`
```
eernstg, line 1095: Why don't we get an error: `x` is non-`late`, cf. line 862?

resources/type-system/flow-analysis.md
```diff
@@ -220,13 +226,36 @@ regardless of their static types.
 
 We also make use of the following auxiliary functions:
 
-- `joinV(VM1, VM2)`, where `VM1` and `VM2` are variable models, represents the
-  union of two variable models, defined as follows:
+- `joinDeclaredType(d1, d2)` is defined as:
+  - if `d1` and `d2` are declared types (the variable was declared with a type,
+    or has a type inferred from its initializer), then `d3`.
```
eernstg, line 231: Unfinished sentence?

resources/type-system/flow-analysis.md
```diff
@@ -220,13 +226,36 @@ regardless of their static types.
...
+  - Otherwise, `d1` and `d2` are **initialization inferred types** and are
+    merged as follows:
+    - If `d1` and `d2` are both missing, then `d3` is missing.
+    - Otherwise if `r1` and `r2` are both true and either `d1` or `d2` is
+      missing, and the variable in question was not declared `late` then it is
+      an error.
+    - Otherwise if `d1` and `d2` are equal then `d3` is `d1`, and is final if
```
eernstg, line 238: Confusing, maybe 'is `d1` (which may or may not be final)'. I assume that `equal` implies they are both final, or none of them is final.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -621,6 +614,488 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+  var x;
+  x = (3 as num); // x is given the provisional initialization inferred type of `num`
+  if (x is! int) return;
+  while(b) { // x is de-promoted to num, and num is made the final initialization inferred type
+    x = 3;
+  }
+}
+```
+
+On joins with back edges in the flow analysis, **untyped variables** are
```
stereotype441, line 806: Note: the flow analysis spec calls these joins "conservative joins" rather than "joins with back edges".  The rationale is that we have one circumstance where we do conservative joins even though there is no back edge: when analyzing `try` statements.  E.g., in the code below, `x` is promoted to `int` at both the top and bottom of the `try` block, but there is some point during the `try` block where it's not promoted, so we need to demote it during processing of the `catch`.  That's implemented using "conservative join" even though there's no back edge involved.
```dart
Object x = ...;
if (x is! int) return; // Promotes to int
try {
  x = ...; // De-promotes
  ...; // Might throw exception
  if (x is! int) return;
  ...; // Might throw exception
} catch (_) {
  x.isEven; // Error: x not known to be `int`
}
```
I think in all the cases in this doc where you talk about "joins with back edges", the same logic will apply to these sorts of try/catch scenarios.  So you might consider rephrasing some of the text in this doc to use "conservative join" terminology rather than talking about back edges.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -621,6 +614,488 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+**initialization inferred type** after the join.
+
+If neither of the above cases apply, then there is at least one path (reachable
+or not) into the join which has an **initialization inferred type**, and the
+**initialization inferred type**  for the variable after the join point
+is computed as follows:
+  - If any path into the join (reachable or not) has a **final initialization
+    inferred type** `T`, then:
+    - It is an error if any other path (reachable or not) has a **final
+      initialization type** `S` and `S` is not equal to `T`.
```
stereotype441, line 877: Is "equal" defined in terms of normalization?  E.g. is `FutureOr<int?>` "equal" to `FutureOr<int?>?` for purposes of this check?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -621,6 +614,488 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+    - It is an error if any other path (reachable or not) has the **provisional
+      initialization type** `Null` and `T` is not nullable.
+    - It is an error if any other path (reachable or not) has a **provisional
+      initialization type** `S` such that `S` is not `Null`, and neither
+      `S` nor `S?` are equal to `T`.
```
stereotype441, line 882: Same issue with the definition of "equal" here

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -621,6 +614,488 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+    - Otherwise, the **final initialization type** after the join is `T`.
+  - Otherwise, if the **provisional initialization inferred type** on each path
+    which has one (reachable or not) is the same type `T` (using syntactic
+    equality), then the **provisional initialization inferred type** after the
+    join is `T`.
+  - Otherwise, if the **provisional initialization inferred type** on each path
+    which has one (reachable or not) is one of `T`, `T?` or `Null` for some type
+    `T` (using syntactic equality), then the **provisional initialization
```
stereotype441, line 890: I'm concerned about using syntactic equality here rather than normalized equality.  Does that mean that if the provisional initialization inferred type on one of the paths is `Never?`, then this rule doesn't apply?  That seems bad.

accepted/future-releases/nnbd/feature-specification.md
stereotype441: Personally I would rather not complicate flow analysis further by adding a special rule for `do` loops, especially if that means that we have to run flow analysis twice on the loop body.  It was an explicit design goal of flow analysis to make sure we didn't have to run it twice, since the CFE and analyzer resolution logic that it's fitting into is designed in such a way that doesn't permit running twice.  IMHO, `do` loops are rare enough in practice that the extra complexity we would take on by making this change wouldn't carry enough user benefit to justify it.

resources/type-system/flow-analysis.md
```diff
@@ -220,13 +226,36 @@ regardless of their static types.
...
+
+- `joinV(VM1, r1, VM2, r2)`, where `VM1` and `VM2` are variable models and `ri`
+  indicates whether the path modeled by `VMi` is reachable, represents the union
+  of two variable models, defined as follows:
   - If `VM1 = VariableModel(d1, p1, s1, a1, u1, c1)` and
   - If `VM2 = VariableModel(d2, p2, s2, a2, u2, c2)` then
   - `VM3 = VariableModel(d3, p3, s3, a3, u3, c3)` where
-   - `d3 = d1 = d2`
-     - Note that all models must agree on the declared type of a variable
+   - `d3 = joinDeclaredType(d1, r1, d2, r2)`
```
stereotype441, line 258: The definition of `joinDeclaredType` above only takes two arguments

resources/type-system/flow-analysis.md
```diff
@@ -284,22 +314,23 @@ We also make use of the following auxiliary functions:
...
+  `rb` (`rf` respectively) indicates whether the path modeled by `VMB` (`VMF`
+  respectively) is reachable, and `b` is a boolean indicating whether the
+  variable is written in the finally block, represents the composition of two
+  variable models through a try/finally and is defined as follows:
   - If `VMB = VariableModel(d1, p1, s1, a1, u1, c1)` and
   - If `VMF = VariableModel(d2, p2, s2, a2, u2, c2)` then
   - `VM3 = VariableModel(d3, p3, s3, a3, u3, c3)` where
-   - `d3 = d1 = d2`
-     - Note that all models must agree on the declared type of a variable
+   - `d3 = joinDeclaredType(d1, r1, d2, r2)`
```
stereotype441, line 325: Again, `joinDeclaredType` only takes two arguments

resources/type-system/flow-analysis.md
```diff
@@ -433,7 +468,7 @@ Definitions:
       - Else if `S` is `X extends R` then let `T1` = `X & T`
       - Else If `S` is `X & R` then let `T1` = `X & T`
       - Else `x` is not promotable (shouldn't happen since we checked above)
-      - Let `VM2 = VariableModel(declared, T1::promoted, T1::tested, assigned,
+      - Let `VM2 = VariableModel(declared, T1::promoted, S::tested, assigned,
```
stereotype441, line 471: Is this a deliberate change?  It seems wrong--we want to add `T1` to the "tested" set, since it's (effectively) the type we're testing against.  `S` is the *old* type, so adding it to "tested" doesn't gain us anything.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -489,25 +492,24 @@ fields on `Object`.
...
+constructors declares an instance variable with a potentially non-nullable type
+and no initializer expression unless the variable is marked with the `late`
+modifier.
 
-It is an error if a potentially non-nullable local variable which has no
-initializer expression and is not marked `late` is used before it is definitely
-assigned (see Definite Assignment below).
+It is an error to derive a mixin from a class declaration which contains an
+instance variable with a potentially non-nullable type and no initializer
+expression unless the variable is marked with the `late` modifier.
```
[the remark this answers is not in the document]
leafpetersen: For the nth time now.... :)  This is text that @lrhn  added in February, and you reviewed.  I have no context on it whatsoever - I just reformatted it.  You both have called it out multiple times now, so I believe you that there is some issue that should be resolved with this, but again, I have **no context whatsoever** on why this was added and what is missing, so maybe you guys should follow up separately?

resources/type-system/flow-analysis.md
leafpetersen: Yes, `S` was wrong.  It's not clear to me that `T1` is right though, is it?  Or should it be `T`?

leafpetersen:
Removed initialization based promotion, restored `var x` to dynamic, and specified the behavior of initialization to null.  PTAL.

resources/type-system/flow-analysis.md
stereotype441: Good point.  The current implementation uses `T`, so let's change it to that.  I'm happy to do that as a separate PR if you want (since at this point it's unrelated to this change), or you can go ahead and do it now--your call.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -621,6 +614,109 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+
+As part of the null safety release, errors for local variables are specified to
+take into account **definite assignment** and **definite unassignment** (see the
+section on Definite Assignment below).  We say that a variable is **potentially
+assigned** if it is not **definitely unassigned**, and that a variable is
+**potentially unassigned** if it is not **definitely assigned**.
+
+In all cases in this section, errors that are described as occurring on reads of
+a variable are intended to apply to all form of reads, including indirectly as
+part of composite assignment operators, as well as via pre and post-fix
```
eernstg, line 627: The standard phrase for `+=` and such is 'compound' assignment operator.
```suggestion
part of compound assignment operators, as well as via pre and post-fix
```

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -621,6 +614,109 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+operators.  Similarly, errors that are described as occurring on writes of a
+variable are intended to apply to all form of writes.
+
+It is a compile time error to assign a value to a `final`, non-`late` local
+variable which is **potentially assigned**.  Thus, it is *not* an error to
+assign to a **definitely unassigned** `final` local variable.
+
+It is a compile time error to assign a value to a `final`, `late` local variable
+if it is **definitely assigned**. Thus, it is *not* an error to assign to a
```
eernstg, line 636: It may cause a dynamic error, so maybe we'd spell it out:
```suggestion
if it is **definitely assigned**. Thus, it is *not* a compile-time error to assign to a
```

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -621,6 +614,109 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+```dart
+void test() {
+  int? x = 3; // x is declared at `int?`
+  x.isEven; // Valid, x has been promoted to `int`
+  x = null; // Valid, demotes to the declared type.
+}
+```
+
+Local variables with no explicitly written type but with an initializer are
+given an inferred type equal to the type of their initializer, unless the that
```
eernstg, line 712:
```suggestion
given an inferred type equal to the type of their initializer, unless that
```

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -561,18 +563,9 @@ expression.
...
-It is a compile time error to assign a value to a local variable marked `late`
-and `final` when the variable is **definitely assigned**.  This includes all
-forms of assignments, including assignments via the composite assignment
-operators as well as pre and post-fix operators.
-
-It is a compile time error to read a local variable marked `late` when the
-variable is **definitely unassigned**. This includes all forms of reads,
-including implicit reads via the composite assignment operators as well as pre
-and post-fix operators.
+It is not a compile time error to write to a `final` non-local variable or field
```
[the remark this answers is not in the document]
leafpetersen: I'll change it to instance variable.  Just using "non-local variable" to mean field is too confusing to me - fields aren't variables semantically (one way to see this is to note that they don't admit alpha-variance).  "Instance variable" is consequently an odd term, but at least it's unambiguous.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: Done.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: Done.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: Done.

resources/type-system/flow-analysis.md
leafpetersen: Since I backed out the rest of the changes to this doc, how about you add this change to one of your updates?

stereotype441: Ok, will do.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -621,6 +614,109 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
...
+### Local variables and inference
+
+Local variables with explicitly written types are given the declared types as
+written.  The declared type of the variable is considered a "type of interest"
+in the sense defined in the flow analysis specification.  If the variable has an
+initializer (explicit or implicit) and is not `final`, then the declaration is
+treated as an assignment for the purposes of promotion.
+
+*Treating the declared type of the variable as a "type of interest" implies that
+if the variable is a nullable type, then its non-nullable version is also a type
```
munificent, line 697: Maybe:

"is a nullable type..." -> "has a nullable type, then the type's non-nullable version is also a type"

For what it's worth, in the doc I wrote for null safety, I use "underlying type" to refer to `T` in `T?`. Both "version" and "base" are pretty overloaded words, so that seemed fairly clear to me.

leafpetersen: Done.

accepted/future-releases/nnbd/feature-specification.md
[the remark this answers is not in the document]
eernstg: I haven't been worried about the term 'instance variable' (and the language specification is using it 70 times vs. just two occurrences of 'field' in normative text), but it sounds like we'd consider 'instance variable' acceptable for now.

However, I don't understand why this rule would not be applicable to a library variable or a class variable. As far as I can see, 'non-local variable' covers exactly the three cases we need, and 'instance variable' covers only one of them.

resources/type-system/flow-analysis.md
stereotype441: https://github.com/dart-lang/language/pull/1116

leafpetersen:
Tests out for review here: https://dart-review.googlesource.com/c/sdk/+/157881

leafpetersen:
Tests for the initializer promotion related behavior out for review here: https://dart-review.googlesource.com/c/sdk/+/163663

Issue #1063, opened 2020-07-03
Include normal completion during return type inference?


eernstg:
[Edit: A summary of the issues brought up can be found [here](https://github.com/dart-lang/language/issues/1063#issuecomment-654889550).]

Cf. https://github.com/dart-lang/sdk/issues/42568, the analyzer and the CFE disagree on some programs with respect to inference of the return type of a function literal.

The difference in behavior fits the following understanding: The analyzer takes normal completion into account (which means that the 'actual returned type' is considered to be nullable when the body of the function can complete normally), but the CFE (and the [current rules](https://github.com/dart-lang/language/blob/master/resources/type-system/inference.md#function-literal-return-type-inference)) do not include that step.

[Correction, July 6th: The current rules _do_ include a step where normal completion includes `Null` in the return type.]

It may seem reasonable to include this step, because it allows function literals to complete normally when that's convenient, and the result is simply that the return type will be nullable, which is sound.

However, we do not otherwise equip normal completion with such powers:

```dart
void main() {
  var b = true;
  f() {
    if (b) return 1;
  }
  g() {
    if (b) return 1;
    if (!b) return null;
  }
}
```

Here, `f` is an error because the inferred return type is `int`, and this makes normal completion of the body an error. However, `g` is accepted, because `return null;` makes the inferred return type nullable, in which case it is not an error to reach the end.

So I'd recommend that we keep the current rules and ask the analyzer to report this error. [Correction July 6th: The current rules are actually a bit confusing, because the function literal gets a nullable return type if the body can complete normally, but a local function seems to omit that step.]

(We have otherwise considered new errors from the analyzer to be a breaking change, but it seems likely that the CFE is used in so many contexts today that it is very difficult to write Dart software which is being used for anything and which avoids seeing an error from the CFE. In other words, it seems reasonable to add the rule that it is also _not a breaking change_ to report a new error in the analyzer if the CFE already reports that error.)

@lrhn, @leafpetersen, @natebosch, @stereotype441, @jakemac53, @munificent, WDYT?

stereotype441:
> (We have otherwise considered new errors from the analyzer to be a breaking change, but it seems likely that the CFE is used in so many contexts today that it is very difficult to write Dart software which is being used for anything and which avoids seeing an error from the CFE. In other words, it seems reasonable to add the rule that it is also _not a breaking change_ to report a new error in the analyzer if the CFE already reports that error.)

[...]

PR #1092, opened 2020-07-14
Make function literal return type inference infer void from context

eernstg:
Cf. https://github.com/dart-lang/language/issues/1063, in particular https://github.com/dart-lang/language/issues/1063#issuecomment-658078704.

This PR changes function literal return type inference to infer `void` from context (rather than using the bottom-up computed actual returned type when the context type is `void` or, for an `async` function, `Future<void>` and similar types like `FutureOr<void>?`). This means that the usual diagnostics emitted for void functions will now also be emitted for function literals passed where the context type has return type `void`/`Future<void>`.

### DIFF
--- a/resources/type-system/inference.md
+++ b/resources/type-system/inference.md
@@ -6,6 +6,9 @@ Status: Draft
 
 ## CHANGELOG
 
+2020.07.14:
+  - Infer return type `void` from context with function literals.
+
 2020.06.04:
   - Make conflict resolution for override inference explicit.
 
@@ -318,9 +321,11 @@ The **actual returned type** of the function literal is the value of `T` after
 all `return` and `yield` statements in the block body have been considered.
 
 Let `T` be the **actual returned type** of a function literal as computed above.
-Let `R` be the greatest closure of the typing context `K` as computed above.  If
-`T <: R` then let `S` be `T`.  Otherwise, let `S` be `R`.  The inferred return
-type of the function literal is then defined as follows:
+Let `R` be the greatest closure of the typing context `K` as computed above.
+If `R` is `void`, or the function literal is marked `async` and `R` is
+`FutureOr<void>`, let `S` be `void`. Otherwise, if `T <: R` then let `S` be
+`T`.  Otherwise, let `S` be `R`.  The inferred return type of the function
+literal is then defined as follows:
   - If the function literal is marked `async` then the inferred return type is
     `Future<flatten(S)>`.
   - If the function literal is marked `async*` then the inferred return type is


scheglov:
When I run [my analyzer CL](https://dart-review.googlesource.com/c/sdk/+/154620), this breaks many `language/` tests.
Do you plan updating them?
Or is this a bug in my implementation?

Note, that there are also failures in Flutter.
If you think that these are expected, I will update Flutter.

leafpetersen:
@scheglov This was intended to be a null safety only change, but in looking at this PR again, I notice it wasn't specified as such.  @eernstg we may want to clarify that.  I think that may be the cause of the issues in flutter?  The language tests look like legitimate breakages, for those you can approve and file an issue (@eernstg do you mind handling updating those)?

@jakemac53 @kevmoo There's a possibility that this change, when fixed and landed (see the CL linked by @scheglov above) will break ported code.  

cc @davidmorgan as well for the heads up for when this rolls.  Maybe we should consider doing a presubmit with the CL from @scheglov first to see if it will break anything?

eernstg:
@scheglov wrote:
> When I run my analyzer CL, this breaks many `language/` tests.
> Do you plan updating them?

Yes, certainly. The breakages that I immediately expect are the following (where there is an error with a function declaration due to `void` as the return type or future value type, and the corresponding error does not occur with a function literal):

```
async_invalid_return_05_test.dart async_invalid_return_08_test.dart async_invalid_return_11_test.dart  
async_invalid_return_14_test.dart async_invalid_return_17_test.dart async_invalid_return_20_test.dart 
async_invalid_return_52_test.dart async_invalid_return_53_test.dart  async_invalid_return_54_test.dart 
async_invalid_return_55_test.dart async_invalid_return_56_test.dart async_invalid_return_57_test.dart 
async_invalid_return_58_test.dart async_invalid_return_59_test.dart
```

Any test where a function literal has a context type whose return type has `void` as the return type or future value type is subject to new constraints, so there could be more.

Comparing this to the [CL results](https://dart-ci.firebaseapp.com/cl/154620/1), I can see that _5{6,7,8,9}_ are not reported as errors, even though the future value type is `void` for each of those functions.

@leafpetersen wrote:
> @scheglov This was intended to be a null safety only change, but in looking at
> this PR again, I notice it wasn't specified as such. @eernstg we may want to clarify that.

Right, the PR changes [inference.md](https://github.com/dart-lang/language/blob/master/resources/type-system/inference.md), and that document had sentences like 'If the greatest closure of `K` is `S` and `S` is a subtype of `Null`, then `T` is `Object?`' before this PR, so I assumed that any change to that document would be null-safety only.

Looking again, it looks like most of inference.md does _not_ refer to null-safety specific constructs. Do we need to adjust inference.md such that it specifically marks certain paragraphs as valid with null-safety and others as valid without null-safety, and the rest is valid for both?

eernstg:
Test updates [here](https://dart-review.googlesource.com/c/sdk/+/154661).

eernstg:
Implementation issues https://github.com/dart-lang/sdk/issues/42720, https://github.com/dart-lang/sdk/issues/42721.

davidmorgan:
Thanks. If null safe only it's low risk for google3. If it applies everywhere, yes, we should presubmit :)

eernstg:
Yep, the change was intended to be applied with null-safety only, and we just need to make sure the documents say so.

leafpetersen:
@eernstg I do still think it's worth calling out this change (in the same way that the change in local function inference is called out).

@davidmorgan are you worried about the possibility that this will break opted in third_party packages during the roll?

davidmorgan:
Doesn't seem likely--should roll in on Monday.

eernstg:
> it's worth calling out this change

Sounds good, cf. https://github.com/dart-lang/language/pull/1102.

PR #1093, opened 2020-07-14
Add new number operator rules to language specification.

lrhn:


eernstg:
Looks good, but we shouldn't put this into the language specification right now: We're on the way to a language specification for the language version 2.8, and I plan to deal with the null-safety elements explicitly (possibly by reverting certain syntax changes that are null-safety specific).

So what's the best way to keep this PR on track? A separate branch which is used to hold various non 2.8 features until the stable 2.8 spec is out? I really think we should be able to get the 2.8 spec out at some point this summer, so maybe it can simply wait.

lrhn:
Let's just keep it waiting for now.

eernstg:
OK!

lrhn:
@eernstg Is it time yet?

eernstg:
I believe we should release the stable spec (at this point it might be possible to mark it as 2.10, but at least 2.8), and for that we need to settle extension methods (#1156). The number operator rules would then be integrated into the language specification along with all other things related to null safety.

specification/dartLangSpec.tex
```diff
@@ -15860,18 +15939,43 @@ \subsection{Multiplicative Expressions}
...
-is \code{int} if the static type of $e_2$ is \code{int},
-and \code{double} if the static type of $e_2$ is \code{double}.
+
+\LMHash{}%
+Let $e$ be a multiplicative expression of the form \code{$e_1$ $op$ $e_2$}
+where $op$ is one of \code{*} or \code{\%},
+let $T$ be the static type of $e_1$,
+and let $C$ be the context type of $e$.
+
+If $T\,\,<:\,\,\code{num}$ and not $T\,\,<:\,\,\code{Never}$, then:
```
eernstg, line 15954:
```suggestion
If \SubtypeNE{T}{\code{num}} and not \SubtypeNE{T}{\code{Never}}, then
the context type of $e_2$ is determined as follows:
```

specification/dartLangSpec.tex
```diff
@@ -15860,18 +15939,43 @@ \subsection{Multiplicative Expressions}
...
+\begin{itemize}
+  \item{} If $\code{int}\,\,<:\,\,C$ and not $\code{num}\,\,<:\,\,C$,
```
eernstg, line 15956:
```suggestion
  \item{} If \SubtypeNE{\code{int}}{C} and not \SubtypeNE{\code{num}}{C},
```

specification/dartLangSpec.tex
```diff
@@ -15860,18 +15939,43 @@ \subsection{Multiplicative Expressions}
...
+      and $T\,\,<:\,\,\code{int}$
```
eernstg, line 15957:
```suggestion
      and \SubtypeNE{T}{\code{int}}
```

specification/dartLangSpec.tex
```diff
@@ -15860,18 +15939,43 @@ \subsection{Multiplicative Expressions}
...
+      then the context type of $e_2$ is \code{int}.
+  \item{} If $\code{double}\,\,<:\,\,C$ and not $\code{num}\,\,<:\,\,C$,
```
eernstg, line 15959:
```suggestion
  \item{} If \SubtypeNE{\code{double}}{C} and not \SubtypeNE{\code{num}}{C},
```

specification/dartLangSpec.tex
```diff
@@ -15860,18 +15939,43 @@ \subsection{Multiplicative Expressions}
...
+      and not $T\,\,<:\,\,\code{double}$
```
eernstg, line 15960:
```suggestion
      and not \SubtypeNE{T}{\code{double}}
```

specification/dartLangSpec.tex
```diff
@@ -15860,18 +15939,43 @@ \subsection{Multiplicative Expressions}
...
+     then the context type of $e_2$ is \code{double}.
+  \item{} Otherwise the context type of $e_2$ is \code{num}.
+\end{itemize}
+Let further $S$ be the static type of $e_2$.
+If $T\,\,<:\,\,\code{num}$ and not $T\,\,<:\,\,\code{Never}$
```
eernstg, line 15965:
```suggestion
If \SubtypeNE{T}{\code{num}} and not \SubtypeNE{T}{\code{Never}}
```

specification/dartLangSpec.tex
```diff
@@ -15860,18 +15939,43 @@ \subsection{Multiplicative Expressions}
...
+and $S$ is assignable to \code{num},
+then the static type of $e$ is determined as follows:
+\begin{itemize}
+  \item{} If $T\,\,<:\,\,\code{double}$
```
eernstg, line 15969:
```suggestion
  \item{} If \SubtypeNE{T}{\code{double}}
```

specification/dartLangSpec.tex
```diff
@@ -15860,18 +15939,43 @@ \subsection{Multiplicative Expressions}
...
+    then the static type of $e$ is $T$.
+  \item{} Otherwise, if $S\,\,<:\,\,\code{double}$
```
eernstg, line 15971:
```suggestion
  \item{} Otherwise, if \SubtypeNE{S}{\code{double}}
```

specification/dartLangSpec.tex
```diff
@@ -15860,18 +15939,43 @@ \subsection{Multiplicative Expressions}
...
+    and not $S\,\,<:\,\,\code{Never}$,
```
eernstg, line 15972:
```suggestion
    and not \SubtypeNE{S}{\code{Never}},
```

specification/dartLangSpec.tex
```diff
@@ -15860,18 +15939,43 @@ \subsection{Multiplicative Expressions}
...
+    then the static type of $e$ is \code{double}.
+  \item{} Otherwise, if $T\,\,<:\,\,\code{int}$,
```
eernstg, line 15974:
```suggestion
  \item{} Otherwise, if \SubtypeNE{T}{\code{int}},
```

specification/dartLangSpec.tex
```diff
@@ -15860,18 +15939,43 @@ \subsection{Multiplicative Expressions}
...
+    $S\,\,<:\,\,\code{int}$ and not $S\,\,<:\,\,\code{Never}$,
```
eernstg, line 15975:
```suggestion
    \SubtypeNE{S}{\code{int}} and not \SubtypeNE{S}{\code{Never}},
```

eernstg (approved):
LGTM, with a bunch of suggested changes at the LaTeX level, aimed at following current conventions in dartLangSpec.tex.

Many changes use `\SubtypeNE`. The `NE` part means "no environment". A subtype judgment usually requires an environment (specifying the bounds of type variables), but this is often not mentioned explicitly in the language specification, and `\SubtypeNE` will omit the symbol for the environment such that the text can just leave it implicit as well.

We have a large number of situations where a specific rule is applicable when a given type is in a type range (that is, `T` satisfies `T <: S1` and `T <\: S2` for two specific types `S1` and `S2`, where `S2` is often `Never`). It might make things somewhat more concise to introduce a separate concept for this, but I did not recommend doing that because it is unlikely to be used very frequently except for exactly the text changed by this PR.

eernstg:
Most suggestions in the review are for consistency with the use of LaTeX in the rest of the language specification. Proceed?

eernstg (approved):
Still LGTM. I think there's no reason to keep this update waiting: We have already started putting language about post-null-safety features into the language specification, so let's just get this kind of update landed wherever we can.

PR #1096, opened 2020-07-15
Remove rule about library name clashes in exports

eernstg:
Following up on #1083, we need to remove the rule about library name clashes in exports as well. This PR does that.


eernstg:
I think I'll assume that this change is accepted along with the import name clash change: The discussion was focused on the conflict as such (and the uselessness of detecting it without transitivity), and not on whether it was reported for an import or for an export directive.

So: Landing.

PR #1105, opened 2020-07-21
Typo fix: Change `?` to `_`

eernstg:
In #1102, Leaf said 'I have an open PR which is changing all uses of `?` to `_`, so let's keep this as `_`.' (click the only 'Show resolved' link on #1102 to see it).

This PR is a tiny fix that changes `?` back to `_` and deletes a commentary paragraph about `?`. It is needed because I forgot to upload a commit which did the same thing just before landing #1102.

### DIFF
--- a/resources/type-system/inference.md
+++ b/resources/type-system/inference.md
@@ -284,13 +284,8 @@ The function **futureValueTypeSchema** is defined as follows:
 - **futureValueTypeSchema**(`FutureOr<S>`) = `S`, for all `S`.
 - **futureValueTypeSchema**(`void`) = `void`.
 - **futureValueTypeSchema**(`dynamic`) = `dynamic`.
-- **futureValueTypeSchema**(`?`) = `?`.
+- **futureValueTypeSchema**(`_`) = `_`.
 - Otherwise, for all `S`, **futureValueTypeSchema**(`S`) = `Object?`.
-
-_In this definition, when `?` occurs as a type on its own, it is the type
-schema that imposes no constraints, cf. section 'Type Schemas'. Note that it
-has nothing to do with nullability, and in particular it does not stand for
-`T?` for any `T`._
 
 _Note that it is a compile-time error unless the return type of an asynchronous
 non-generator function is a supertype of `Future<Never>`, which means that


resources/type-system/inference.md
```diff
@@ -284,14 +284,9 @@ The function **futureValueTypeSchema** is defined as follows:
 - **futureValueTypeSchema**(`FutureOr<S>`) = `S`, for all `S`.
 - **futureValueTypeSchema**(`void`) = `void`.
 - **futureValueTypeSchema**(`dynamic`) = `dynamic`.
-- **futureValueTypeSchema**(`?`) = `?`.
+- **futureValueTypeSchema**(`_`) = `_`.
```
lrhn, line 287: What does `_` mean?

Is it a placeholder for "any (other) type"? Because then it is covered by the next item.
Is it the "empty context type"? If so, state that, there is no other use of `_` in this document.

eernstg: It is the type schema that does not impose any constraints, and you could say that it is then a placeholder for any type. But it's only a placeholder for other types in the sense that inference can replace it by any type&mdash;it is still a separate term, different from all types, and part of the concrete syntax of type schemas.

We can't ignore it: If we leave out line 287 then line 288 would apply, and that would erase `_` from the future value type schema; so `[1]` might then be inferred as `<dynamic>[1]` currently because it gets the context type `List<_>`, but if we make it `Object?` then that `[1]` would be inferred as `<Object?>[1]`. 

The reason why I changed it to `?` was exactly that this document doesn't otherwise use `_`; but as Leaf mentions in #1102, line 287, he has a not-yet-landed PR where `?` is consistently changed to `_` in this document. This is the reason why he suggested that I should just keep it as `_`.

lrhn: OK to keep it, but do document what `_` means because it's not  obvious.

eernstg: I'm trusting Leaf's not-yet-landed PR to document this ('I have an open PR which is changing all uses of `?` to `_`, so let's keep this as `_`.'). Can you live with that?

lrhn: ACK. Let's make sure to check that `_` is properly defined there then :)

eernstg: Thanks, will do that!  ;-)

PR #1106, opened 2020-07-21
Adjust `<type> implements` to include type variables

eernstg:
This PR changes the language specification such that `$S$ implements $T$` is generalized to allow `$S$` to be a type variable whose nearest non-type-variable bound is an interface type that has `$T$` as a superinterface. This means that we are relying on the interface associated with `$S$` rather than `$S$` itself. In particular, `X extends C` is treated the same as `C`, which is also what is needed for meaningful and useful behavior of the rules.

For example, the current definition makes this an error, and that is neither intended nor helpful:

```dart
void f<X extends Iterable<int>>(X x) {
  var v = [...x]; // Error!
}
```

The error occurs because the static type of `a` is required to be a type that "implements `Iterable<T>`" for some `T`. We already have special cases for the situation where `a` has a bottom-ish type (`Null`, `Never`, `X extends Null` and so on), but the case where we have a type variable like `X extends Iterable<int>` doesn't match any of the cases, and ends up being an error. I've checked all usages of `<type> implements` in the language specification, and they all justify this change.

This change is also consistent with the previously added text [here](https://github.com/dart-lang/language/blob/b196daf6bdeb6e591fd90a3a683d736365cb8526/specification/dartLangSpec.tex#L4176), which provides the same effect for ordinary method invocations and property extractions. (That part can't use the new meaning of `<type> implements ...` for simplification, because it is only concerned with the the interface associated with a type, not with the relationship to a different type which is found in the superinterface hierarchy).

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -4427,7 +4427,10 @@ or $J$ is a superinterface of a direct superinterface of $I$.
 When we say that a type $S$
 \IndexCustom{implements}{type!implements a type}
 another type $T$,
-this means that $T$ is a superinterface of $S$.
+this means that $T$ is a superinterface of $S$,
+or $S$ is $S_0$ bounded for some type $S_0$
+(\ref{bindingActualsToFormals}),
+and $T$ is a superinterface of $S_0$.
 Assume that $G$ is a raw type
 (\ref{instantiationToBound})
 whose declaration declares $s$ type parameters.
@@ -4435,7 +4438,7 @@ When we say that a type $S$
 \IndexCustom{implements}{type!implements a raw type}
 $G$,
 this means that there exist types \List{U}{1}{s}
-such that \code{$G$<\List{U}{1}{s}>} is a superinterface of $S$.
+such that $S$ implements \code{$G$<\List{U}{1}{s}>}.
 
 \commentary{%
 Note that this is not the same as being a subtype.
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -4427,15 +4427,18 @@ \subsection{Superinterfaces}
 When we say that a type $S$
 \IndexCustom{implements}{type!implements a type}
 another type $T$,
-this means that $T$ is a superinterface of $S$.
+this means that $T$ is a superinterface of $S$,
```
lrhn, line 4430: Where do we define "is a superinterface of"?

Is `Iterable<num>` a superinterface of `List<int>`?

eernstg: > is there some existential check of "There exists a `T <: num` such that `list implements Iterable<T>`"?

I'm not sure what this is about. Considering `Iterable<num> foo<T extends List<int>>(T list) => [...list];` again, we're allowed to use `list` in the spread element because of the change that this CL makes. Then we have an expression `[...list]` of type `List<int>`. That expression is returned, the function is "normal" (non-generator and so on), and it has return type `Iterable<num>`. Assignability holds because `List<int> <: Iterable<num>`. So we only use 'implements' for the check on the spread element, and returning only involves assignability, that is: subtyping.

lrhn: The expression `[...list]` has type `List<num>` because of the context.
So, I guess I'm just checking that `<num>[...listOfInt]` is allowed. If it currently is, it will sure stay valid with this change.
And yes, reading it again, it indeed says "must implement Iterable\<*T*> for some *T*", so that's fine.

eernstg: Are you OK with this at this point?

PR #1109, opened 2020-07-22
Specify the set of super-interfaces used for mixed mode upper bounds

leafpetersen:
In response to https://github.com/dart-lang/sdk/issues/42706, this specifies the set of super-interfaces used in computing upper bounds in mixed mode.

No modification to the super-interfaces is performed when computing from an opted in library, but the `LEGACY_ERASURE` is used when computing from an opted out library.

### DIFF
--- a/resources/type-system/upper-lower-bounds.md
+++ b/resources/type-system/upper-lower-bounds.md
@@ -227,7 +227,7 @@ We define the upper bound of two types T1 and T2 to be **UP**(`T1`,`T2`) as foll
     super-interfaces used consists of the `LEGACY_ERASURE` of the
     super-interfaces of the two types.
   - For an upper bound computation in an opted in library, no modification of
-    the set of super-interfaces if performed.
+    the set of super-interfaces is performed.
 
 ## Lower bounds
 
--- a/resources/type-system/upper-lower-bounds.md
+++ b/resources/type-system/upper-lower-bounds.md
@@ -4,6 +4,9 @@ leafp@google.com
 
 ## CHANGELOG
 
+2020.07.21
+  - **CHANGE** Specify treatment of mixed hierarchies.
+
 2020.03.30
   - **CHANGE** Update DOWN algorithm with extensions for FutureOr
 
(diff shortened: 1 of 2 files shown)


leafpetersen:
cc @eernstg @lrhn @munificent 

cc @johnniwinther @scheglov 

Note that for out of order migration, this seems to be consistent with what is already implemented.  For in order migration, this may be a change, but I'm not sure I understand the behavior I'm seeing as described [here](https://github.com/dart-lang/sdk/issues/42706#issuecomment-662226333).  

Tests forthcoming.

leafpetersen:
In progress tests [here](https://dart-review.googlesource.com/c/sdk/+/155380).  In order only so far.

resources/type-system/upper-lower-bounds.md
```diff
@@ -217,8 +220,14 @@ We define the upper bound of two types T1 and T2 to be **UP**(`T1`,`T2`) as foll
...
-- **UP**(`C0<T0, ..., Tn>`, `C1<S0, ..., Sk>`) = least upper bound of two interfaces
-  as in Dart 1.
+- **UP**(`C0<T0, ..., Tn>`, `C1<S0, ..., Sk>`) = least upper bound of two
+  interfaces as in Dart 1, with modifications for handling mixed null safe and
+  legacy code as follows:
+  - For an upper bound computation in a legacy library, the set of
+    super-interfaces used consists of the `LEGACY_ERASURE` of the
+    super-interfaces of the two types.
+  - For an upper bound computation in an opted in library, no modification of
+    the set of super-interfaces if performed.
```
eernstg, line 230:
```suggestion
    the set of super-interfaces is performed.
```

Of course, the superinterfaces in the algorithm in the language specification are gathered by traversing the graph, but here we'd assume that we use the ones which are associated with the direct superinterfaces (the consolidated ones, all legacy for a legacy class, and conflict resolved for others). So `C` implements `A<int*>` in the following example, and we don't use the set `{A<int>, A<int?>}` in LUBs:

```dart
// Opted in 'lib_in.dart'.
class A<X> {}
class B1 implements A<int> {}
class B2 implements A<int?> {}

// Opted out 'lib_out.dart'.
import 'lib_in.dart';
class C implements B1, B2 {} // OK, `C` implements `A<int*>`.

// Opted in 'main.dart'.
import 'lib_out.dart';

void main() {
  bool b = ...;
  A<int> a = b ? C() : C(); // OK, initializer type is A<int*>.
}
```

It would surely be inconvenient if we were to make this LUB `Object`, based on having the set `{A<int>, A<int?>}` rather than `{A<int*>}` at depth 1.

Next, when we are working with the consolidated 'implements' data we would presumably have lost some information: What is the depth? I'd assume that we should make it the maximal depth among all the types that we have collapsed into a single element in these sets of parameterized types.

Next again, is it a problem that we will not apply conflict resolution during the LUB computation?

```dart
...

void main() {
  ...
  A<int> a2 = b ? C() : A<int>(); // Error, initializer type is `Object`.
}
```

In this case we'd have the set `{A<int*>, A<int>}`. We could also apply conflict resolution to each of these sets before we ask whether they have size 1, that is, we could replace `{t1, ... tk}` by `{s1, ... sm}` which is obtained by replacing `ti` and `tj` by `NNBD_TOP_MERGE(ti, tj)` whenever that computation of `NNBD_TOP_MERGE` is defined, with iteration until no such step is possible.

leafpetersen: As discussed off line, I'm proposing to not do anything smarter for the out of order case right now.  For the non-mixed mode case, we could consider doing something smarter about top types via normalization and top merge, but I'd rather avoid blocking on that now.

PR #1110, opened 2020-07-22
Clarify rules about initializing formals

### LINKED ISSUE #1108: Initializing formal subtype error is specified with a variable defined in unrelated text.
https://github.com/dart-lang/language/blob/2c70f53b9de33145f452f2ec6cc96ad78eee5e92/specification/dartLangSpec.tex#L2988..L2990

> An \Index{initializing formal} has the form \code{\THIS{}.\id}, where \id{} is the name of an instance variable of the immediately enclosing class.
...
> If an explicit type is attached to the initializing formal, that is its static type.
> Otherwise, the type of an initializing formal named \id{} is $T_{id}$, where $T_{id}$ is the type of the instance variable named \id{} in the immediately enclosing class.
> It is a compile-time error if the static type of \id{} is not a subtype of $T_{id}$.

_T\_{id}_ is only specified in the "Otherwise" case, but the compile-time error can only be found in the "If an explicit type is attached" case.

eernstg:
Issue https://github.com/dart-lang/language/issues/1108 reports that the rules about initializing formals use `$T_{id}$` in a location where it is undefined. This PR resolves that, and several other imprecision issues at that location in the specification.

Note that the syntax of `<fieldFormalParameter>` was changed to allow an old-style function type to be nullable, but that there is no corresponding implementation effort because the analyzer as well as the CFE already supports this syntax.

Resolves #1108.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -2984,9 +2984,8 @@ A \Index{formal constructor parameter} is a formal parameter
 (\ref{formalParameters}),
 which can be an initializing formal.
 An \Index{initializing formal} contains a term of the form \code{\THIS.\id},
-where \id{} is the name of an instance variable of
-the immediately enclosing class.
-It is a compile-time error if \id{} is not
+where \id{} is the name of the initializing formal.
+It is a compile-time error if \id{} is not also the name of
 an instance variable of the immediately enclosing class.
 It is a compile-time error if an initializing formal is declared in
 a function declaration which is not a non-redirecting generative constructor.
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -2978,16 +2978,27 @@ and either a redirect clause or an initializer list and an optional body.
 \end{grammar}
 
 \LMHash{}%
-A \Index{constructor parameter list} is a parenthesized, comma-separated list of formal constructor parameters.
-A \Index{formal constructor parameter} is either a formal parameter (\ref{formalParameters}) or an initializing formal.
-An \Index{initializing formal} has the form \code{\THIS{}.\id}, where \id{} is the name of an instance variable of the immediately enclosing class.
-It is a compile-time error if \id{} is not an instance variable of the immediately enclosing class.
-It is a compile-time error if an initializing formal is used by a function other than a non-redirecting generative constructor.
+A \Index{constructor parameter list} is
+a parenthesized, comma-separated list of formal constructor parameters.
+A \Index{formal constructor parameter} is a formal parameter
+(\ref{formalParameters}),
+which can be an initializing formal.
+An \Index{initializing formal} contains a term of the form \code{\THIS.\id},
+where \id{} is the name of an instance variable of
+the immediately enclosing class.
+It is a compile-time error if \id{} is not
+an instance variable of the immediately enclosing class.
+It is a compile-time error if an initializing formal is declared in
+a function declaration which is not a non-redirecting generative constructor.
 
 \LMHash{}%
-If an explicit type is attached to the initializing formal, that is its static type.
-Otherwise, the type of an initializing formal named \id{} is $T_{id}$, where $T_{id}$ is the type of the instance variable named \id{} in the immediately enclosing class.
-It is a compile-time error if the static type of \id{} is not a subtype of $T_{id}$.
+Assume that $p$ is a declaration of an initializing formal parameter named \id.
+Let $T_{id}$ be the type of the instance variable named \id{} in
+the immediately enclosing class.
+If $p$ has a type annotation $T$ then the declared type of $p$ is $T$.
+Otherwise, the declared type of $p$ is $T_{id}$.
+It is a compile-time error if the declared type of $p$
+is not a subtype of $T_{id}$.
 
 \LMHash{}%
 Initializing formals constitute an exception to the rule that
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -30,7 +30,8 @@
 % - Change several warnings to compile-time errors, matching the actual
 %   behavior of tools.
 % - Eliminate error for library name conflicts in imports and exports.
-% - Clarify the specification of initializing formals.
+% - Clarify the specification of initializing formals. Fix error: Add
+%   missing `?' at the end for function typed initializing formal.
 %
 % 2.7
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -2978,16 +2978,26 @@ \subsubsection{Generative Constructors}
...
+a function declaration which is not a non-redirecting generative constructor.
+
+\LMHash{}%
+Assume that $p$ is a declaration of an initializing formal parameter named \id.
+Let $T_{id}$ be the type of the instance variable named \id{} in
+the immediately enclosing class.
+If $p$ has a type annotation $T$ then the declared type of $p$ is $T$.
+Otherwise, the declared type of $p$ is $T_{id}$.
+It is a compile-time error if the declared type of $p$
+is not a subtype of $T_{id}$.
```
lrhn, line 3000: (I it worth mentioning that a constructor parameter cannot be `covariant`, so "subtype of" is sufficient).

eernstg: Interesting corner! ;)

Actually, I'd say that we need to ensure that it would not be an error to use an expression whose type is the type of the formal in an initialization of the instance variable; next, we don't want `dynamic`, so the formal type must be a subtype of the variable type. If the instance variable is declared to be covariant then an overriding setter can have a more special parameter type, but that shouldn't matter, because that overriding setter would not be invoked in order to initialize the variable in this class anyway. So we can rely on the declared type of the variable in this class even when the variable is covariant.

specification/dartLangSpec.tex
```diff
@@ -2978,16 +2978,26 @@ \subsubsection{Generative Constructors}
...
-If an explicit type is attached to the initializing formal, that is its static type.
-Otherwise, the type of an initializing formal named \id{} is $T_{id}$, where $T_{id}$ is the type of the instance variable named \id{} in the immediately enclosing class.
-It is a compile-time error if the static type of \id{} is not a subtype of $T_{id}$.
+A \Index{constructor parameter list} is
+a parenthesized, comma-separated list of formal constructor parameters.
+A \Index{formal constructor parameter} is a formal parameter
+(\ref{formalParameters}),
+which can be an initializing formal.
+An \Index{initializing formal} contains a term of the form \code{\THIS.\id},
+where \id{} is the name of the initializing formal.
```
lrhn, line 2987: That sentence sounds descriptive, not prescriptive, but it appears to be defining "initializing formal".

> An \\Index{initializing formal} is a \\syntax{\<fieldFormalParameter\>}. It contains a term of the form \code{\THIS.\id}, where \id{} is the name of the initializing formal.

specification/dartLangSpec.tex
```diff
@@ -2978,16 +2978,26 @@ \subsubsection{Generative Constructors}
...
+It is a compile-time error if \id{} is not also the name of
+an instance variable of the immediately enclosing class.
```
lrhn, line 2989: Maybe move this below the next paragraph, because an initializing formal which is not in a non-redirecting generative constructor, may not have any enclosing class.

specification/dartLangSpec.tex
lrhn: Also: is the grammar for parameters correct. I can't see a production for an initializing function parameter, `(int this.foo(int x))`, but our implementations accept it.

srawlins:
Thanks so much!

specification/dartLangSpec.tex
eernstg: (Aha, so I have to reload the page to see comment number 2 and up.)

I was thinking about this sentence, too. We have the grammar rule that introduces initializing formals in a completely different place, `<fieldFormalParameter>` on line 1566.  That rule _does_ allow for the old style function typed parameter (like `int this.foo(int x)`). (It also allows for `var this.foo(int x)` that I've recommended that we turn into an error ;-).

But the point is that (presumably) the text around line 2984 used to refer to a grammar rule that actually had an alternative for "normal" parameters and another alternative for initializing formal parameters. And then the grammar was updated, some time around 1812 (OK, 2012 maybe), but the text right here wasn't updated.

This means that it's now a non-syntax error to have an initializing formal in any other function than a generative non-redirecting constructor, and that error has been added as it should.

But it's confusing that we're not describing the properties of initializing formals at the same location where the grammar is specified.

lrhn: Ah, yes:
> ```
>  fieldFormalParameter> ::= \gnewline{}
>       <finalConstVarOrType>? \THIS{} `.' <identifier> <formalParameterPart>?
>```
which allows parameters afterwards. It's missing the `?`? to allow for a nullable function type, but otherwise it's fine.
(So, it should be:

> ```
> <fieldFormalParameter> ::= \gnewline{}
>       <finalConstVarOrType>? \THIS{} `.' <identifier> (<formalParameterPart> `?'?)?
>```

specification/dartLangSpec.tex
eernstg: Moved.

specification/dartLangSpec.tex
eernstg: Uploaded a commit which is considerably clearer on these things. Also, it moves the error for having an init. formal to the section about parameters, with commentary referring to it in the constructor section. After all, this is a rule that everybody _else_ than those constructors need to follow.

eernstg: We got rid of `var` a while ago via a separate compile-time error, in the text below. In order to rock the boat a bit less, I've proposed a new grammar rule that keeps the `<finalConstVarOrType>` and just adds the optional question mark at the end of a parameter list.

eernstg: @lrhn, PTAL.

specification/dartLangSpec.tex
```diff
@@ -1573,6 +1575,17 @@ \subsubsection{Required Formals}
 It is a compile-time error if \VAR{} occurs as
 the first token of a \synt{fieldFormalParameter}.
 
+\LMHash{}%
+It is a compile-time error if a parameter derived from
+\synt{fieldFormalParameter} occurs in a function declaration
+which is not a non-redirecting generative constructor.
```
lrhn, line 1581: "in" is a little vague. Doesn't definitely rule out something like 
```dart
Foo(int foo(int this.x));
```
maybe:
> It's a compile-time error if a parameter derived from
> \sybt{fieldFormalParameter} occurs as a parameter of anything 
> except a non-redirecting generative constructor.

eernstg: Done.

PR #1114, opened 2020-07-23
Create draft proposal for `x!+=1`

lrhn:


working/1113 - null-asserting compound assignment/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+
+```latex
+<assignmentOperator> ::= `='
+  \alt `!'? <compoundAssignmentOperator>
+  
+<unaryExpression> ::= …
+  \alt <incrementOperator> <assignableExpression> `!'?
+
+<postfixOperator> ::= `!'? <incrementOperator>  
+```
```
eernstg, line 88: I think that works.

working/1113 - null-asserting compound assignment/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+
+This allows one optional `!` exactly where it affects a composite operation, and nowhere else.
+
+This grammar is not ambiguous. The `++x!` looks like it could be a problem,
+but because prefix-`++` only applies to an `<assignableExpression>`, a
+ny existing valid expression starting with `++x!` would need to be followed
```
eernstg, line 94: Typo: `a ny` --> `any`.

working/1113 - null-asserting compound assignment/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+when followed by another selector like`.foo`.
+
+### Semantics
+
+For each LHS, the semantics of an expression of the form 
+``lhs `!' op= expr`` is is specified in almost the same way
+as `lhs op= expr` except that:
+
+* After *reading* `lhs`, the operation throws if the value is `null` in the same way as `lhs!` would.
+* The static analysis uses the **NON_NULL** type of the static type of `lhs` for the operator lookup.
```
eernstg, line 107: We may want to say that if the static type of `lhs` is `T` and `T` is `T0` bounded, then we use `NON_NULL(T0)` for the operator lookup.

This will allow `++x!` where the type of `x` is `X extends int?`, where the static type `X` is the same as `NON_NULL(X)`.

working/1113 - null-asserting compound assignment/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+
+For each LHS, an expression of the form ``lhs `!' incrementOp`` is specified
+in almost the same ways as `lhs incrementOp`, except that:
+
+* After *reading* `lhs`, the operation throws if the value is `null` in the same way as `lhs!`.
+* The static analysis uses the **NON_NULL** type of the static type of `lhs` for the `+` operator lookup.
```
eernstg, line 113: Same issue as line 107.

working/1113 - null-asserting compound assignment/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+```dart
+id
+e1.id
+e1?.id
+e1[e2]
+e1?[e2]
+super.id
+super[e2]
+ClassName.id  
+```
```
eernstg, line 30: These forms are of course assumed to be derivable in the grammar, so `e1` is a `(<constructorInvocation>|<primary>) <selector>*`, and `ClassName` could be a subset of that (just `<identifier>|<qualified>`), with the constraint that it is a type literal (so it dominates `e1.id` when applicable, and should perhaps occur earlier in the list).

working/1113 - null-asserting compound assignment/proposal.md
lrhn: I though that **NON_NULL**(<code>*X* extends int?</code>)  would be <code>*X*&int</code> (aka <code>X & **NON_NULL**(T0)</code>. If not, I might just have to say that the static computation also uses `lhs!` where the original reads `lhs`.

It won't allow `++x!` because you can't assign `int` back to `X extends int?`.

working/1113 - null-asserting compound assignment/proposal.md
lrhn: ACK

working/1113 - null-asserting compound assignment/proposal.md
eernstg: > I though that **NON_NULL**(`X extends int?`) would be `X&int`

Ah, that's true. So with **NonNull** (maybe it was renamed?) we actually already ensure that `X extends T` gets a similar treatment as `T` itself (as opposed to several other cases where we had to fix this recently).

working/1113 - null-asserting compound assignment/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+This proposal's `!` modifier handles the case where we *know* that the LHS is non-null.
+We could also introduce a version which checks whether the LHS is `null`,
+and does nothing if it is, similar to the distinction between `?.` and `!.`.
+
+We already have `x ??= y` which does something if `x` is null.
+This one would do something if `x` is non-`null` instead.
+Let's go with one `?` as strawman marker syntax:
+
+```dart
+e.x ?+= 2
```
munificent, line 150: Eh, I think this is a -1 for me. It's starting to get a little too close to operator soup.

PR #1116, opened 2020-07-24
Flow analysis: Fix promote()'s update of tested set.

stereotype441:
When promoting via a test on type T, the type that should be added to
the test set is T, even if the actual promoted type is different.
This brings the spec in line with the implementation.


stereotype441:
See discussion here: https://github.com/dart-lang/language/pull/1091/files/cf4905c9a78366400f9debe3105bee938c862e52#r457448403

PR #1130, opened 2020-08-04
Add error for uninitialized final instance variable

eernstg:
This PR adds a compile-time error for an uninitialized final instance
variable in the case where there are no generative constructors at all
in the enclosing class.

Cf. https://github.com/dart-lang/language/issues/1131.

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -6,6 +6,10 @@ Status: Draft
 
 ## CHANGELOG
 
+2020.08.06
+  - Specify error for uninitialized final instance variable in class
+    with no generative constructors.
+
 2020.07.09
   - Specify combined member signature and spread element typing
     with null-safety.
@@ -501,13 +505,10 @@ non-nullable type and no initializer expression, and the class has a generative 
 where the variable is not initialized via an initializing formal or an initializer list entry,
 unless the variable is marked with the `late` modifier.
 
-It is an error if a mixin declaration or a class declaration with no generative constructors
-declares an instance variable with a potentially non-nullable type and no initializer expression
-unless the variable is marked with the `late` modifier.
-
-It is an error to derive a mixin from a class declaration which contains
-an instance variable with a potentially non-nullable type and no initializer expression
-unless the variable is marked with the `late` modifier.
+It is an error if a mixin declaration or a class declaration with no generative
+constructors declares an instance variable without an initializing expression
+which is final or whose type is potentially non-nullable, unless the variable
+is marked with the `late` modifier.
 
 It is an error if a potentially non-nullable local variable which has no
 initializer expression and is not marked `late` is used before it is definitely
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -497,18 +497,20 @@ fields on `Object`.
 It is an error to call an expression whose type is potentially nullable and not
 `dynamic`.
 
-It is an error if a top level variable or static variable with a non-nullable type
-has no initializer expression unless the variable is marked with the `late` modifier.
+It is an error if a top level variable or static variable with a non-nullable
+type has no initializer expression unless the variable is marked with a
+`late` or `external` modifier.
 
-It is an error if a class declaration declares an instance variable with a potentially
-non-nullable type and no initializer expression, and the class has a generative constructor
-where the variable is not initialized via an initializing formal or an initializer list entry,
-unless the variable is marked with the `late` modifier.
+It is an error if a class declaration declares an instance variable with a
+potentially non-nullable type and no initializer expression, and the class has
+a generative constructor where the variable is not initialized via an
+initializing formal or an initializer list entry, unless the variable is marked
+with a `late`, `abstract`, or `external` modifier.
 
 It is an error if a mixin declaration or a class declaration with no generative
 constructors declares an instance variable without an initializing expression
 which is final or whose type is potentially non-nullable, unless the variable
-is marked with the `late` modifier.
+is marked with a `late`, `abstract`, or `external` modifier.
 
 It is an error if a potentially non-nullable local variable which has no
 initializer expression and is not marked `late` is used before it is definitely


srawlins:
Drive-by tangent: I thought it was so weird that we even _parse_ uninitialized _static_ final fields. https://github.com/dart-lang/sdk/issues/41718. Since there is no other place to initialize such a field than at its declaration. Not closely related to this PR though, as this is for instance variables.

specification/dartLangSpec.tex
```diff
@@ -2963,17 +2963,36 @@ \subsection{Constructors}
...
 }
 
 % In what scope do constructors go? The simple names of named constructors go in the static scope of the class. Unnamed ones go nowhere, but we use the class name to refer to them; the class name could also in the static scope of the class as well to prevent weird errors, or we could ban it explicitly and avoiding duplication. Similarly, the instance scope could contain the constructor names and class name, or we could have special rules to prevent collisions between instance members and constructors or the class.
 
 % The enclosing scope of a generative constructor is the instance scope of the class in which it is declared (but what about redirecting?)
 
 \LMHash{}%
-If{}f no constructor is specified for a class $C$, it implicitly has a default constructor \code{C() : \SUPER{}() \{\}}, unless $C$ is class \code{Object}.
+If{}f no constructor is specified for a class $C$,
+it implicitly has a default constructor \code{C():\ \SUPER();}
```
munificent, line 2977: Maybe say "default generative constructor"? It's redundant but maybe good to be explicit.

specification/dartLangSpec.tex
```diff
@@ -2963,17 +2963,36 @@ \subsection{Constructors}
...
+unless $C$ is the built-in class \code{Object}.
+
+\LMHash{}%
+If a class $C$ has no generative constructors
```
munificent, line 2981: And then here clarify, "no generative constructors (either explicitly declared or the implicit default one above)"

leafpetersen:
I'm a bit confused about this.  Is this a pre-null safety fix?  Is it breaking?  Does it require implementation changes?  Or is this a null safety change?  If so, this should be landed in the feature specification if possible, and we need to reconcile it with the null safety feature specification which covers at least the non-nullable part of this.

```
It is an error if a class declaration declares an instance variable with a potentially
non-nullable type and no initializer expression, and the class has a generative constructor
where the variable is not initialized via an initializing formal or an initializer list entry,
unless the variable is marked with the `late` modifier.

It is an error if a mixin declaration or a class declaration with no generative constructors
declares an instance variable with a potentially non-nullable type and no initializer expression
unless the variable is marked with the `late` modifier.

It is an error to derive a mixin from a class declaration which contains
an instance variable with a potentially non-nullable type and no initializer expression
unless the variable is marked with the `late` modifier.
```

specification/dartLangSpec.tex
eernstg: Done.

specification/dartLangSpec.tex
eernstg: Done.

eernstg:
@leapetersen wrote:
> Is this a pre-null safety fix?

That was the intention, to fix it everywhere, because the error is already reported. However, I just re-ran the example programs again and discovered that the error is _not_ reported by the analyzer in a legacy library, so this means that it is indeed breaking (in principle, although the relevant kind of class may be rare).

So we may wish to change this to a null-safety-only update in order to avoid the potential breakage, which means that it should be an update to the null safety spec:

This one shouldn't be affected, because this is about the case where there are no generative constructors:

```
It is an error if a class declaration declares an instance variable with a potentially
non-nullable type and no initializer expression, and the class has a generative constructor
where the variable is not initialized via an initializing formal or an initializer list entry,
unless the variable is marked with the `late` modifier.
```

As mentioned in https://github.com/dart-lang/language/issues/1131#issuecomment-669137475, the next one would be updated to handle the anomaly that we allow a final instance variable to remain uninitialized in just this single case:

```
It is an error if a mixin declaration or a class declaration with no generative constructors
declares an instance variable with no initializing expression which is final or whose type
is potentially non-nullable, unless the variable is marked with the `late` modifier.
```

For the third rule, I think there is no need for updates, because (1) if said class does not have any generative constructors then the previous rule applies; (2) if said class has a generative constructor (which must be a default constructor in order to avoid other errors) then the legacy rules apply (and they make it an error to have an uninitialized final instance variable):

```
It is an error to derive a mixin from a class declaration which contains
an instance variable with a potentially non-nullable type and no initializer expression
unless the variable is marked with the `late` modifier.
```

I suspect that the third rule is redundant. Assume that said error occurs, call the class `C`; we then know that `C` contains an instance variable `v` with a potentially non-nullable type `T` and no initializer, and `v` is not `late`. Now (1) if `C` has no generative constructors then rule 2 applies, and (2) if `C` has any generative constructors then it can only have one, which is the default constructor (or another error would occur), and in that case the first rule applies. So unless there is some special situation that I missed we should delete the third rule.

eernstg:
@munificent, having decided that the language specification should remain unchanged and this change should be made in the null safety specification, this PR is now completely different. Could you PTAL?

@lrhn, could you take a look at the last paragraph in https://github.com/dart-lang/language/pull/1130#issuecomment-669160076 and speak up if you believe that it is incorrect to delete the sentence mentioned there?

eernstg:
Test in https://dart-review.googlesource.com/c/sdk/+/157490.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -501,13 +505,10 @@ non-nullable type and no initializer expression, and the class has a generative
...
 unless the variable is marked with the `late` modifier.
 
-It is an error if a mixin declaration or a class declaration with no generative constructors
-declares an instance variable with a potentially non-nullable type and no initializer expression
-unless the variable is marked with the `late` modifier.
-
-It is an error to derive a mixin from a class declaration which contains
-an instance variable with a potentially non-nullable type and no initializer expression
-unless the variable is marked with the `late` modifier.
+It is an error if a mixin declaration or a class declaration with no generative
```
leafpetersen, line 508: I think we need to make an exception for `abstract` and `external` fields as well, right?  Or we can specify that as part of the abstract fields proposal/spec, but either way we should make sure it's covered (and tested).

eernstg: Yes, and that fits perfectly into the null-safety specification, because that feature is bundled as part of null safety.

It seems funny that we mention that there is no error when the modifier `abstract` or `external` is specified in a non-local variable declaration, because it is a syntax error to _have_ it. But the wording is still necessary because we would otherwise say that _every_ program with a final abstract or external variable is an error because there is no initialization of it anywhere.

Done.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -493,21 +497,20 @@ fields on `Object`.
...
+initializing formal or an initializer list entry, unless the variable is marked
+with a `late`, `abstract`, or `external` modifier.
 
-It is an error to derive a mixin from a class declaration which contains
-an instance variable with a potentially non-nullable type and no initializer expression
-unless the variable is marked with the `late` modifier.
+It is an error if a mixin declaration or a class declaration with no generative
+constructors declares an instance variable without an initializing expression
+which is final or whose type is potentially non-nullable, unless the variable
+is marked with a `late`, `abstract`, or `external` modifier.
```
lrhn, line 513: I guess this situation for a *class* is only an issue if the class can be used as a mixin (so only if it's superclass is `Object`, and only as long as we allow extracting mixins from class declarations.
A class with no generative constructor cannot be instantiated, so it can never have an uninitialized instance variable.

lrhn:
I agree that the third rule is redundant for the reasons mentioned: If you can extract a mixin at all, it would be covered by the second rule.

accepted/future-releases/nnbd/feature-specification.md
eernstg: > only an issue if the class can be used as a mixin

Agreed, and I was thinking about this distinction as well. I think there's little value, however, in allowing an uninitialized final instance variable in this very special case where it is guaranteed to never exist. Such a class should probably be abstract, and the final instance variable could then be expressed as an abstract getter (or an abstract final instance variable).

If you wish to pursue that we allow it then I'd handle that in a separate PR.

PR #1133, opened 2020-08-05
Add null-safety specification of super-bounded types

eernstg:
This PR adds a section to the null-safety specification specifying how to determine whether a given type is correctly super-bounded.

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -6,9 +6,12 @@ Status: Draft
 
 ## CHANGELOG
 
+2020.08.05
+  - Specify updates to super-bounded type rules for null safety.
+
 2020.07.09
   - Specify combined member signature and spread element typing
-    with null-safety.
+    with null safety.
 
 2020.06.02
   - Fix the diff to the spec for potentially constant instance checks
@@ -699,10 +702,26 @@ the element, key, and value type of `...e` and `...?e` is `Never`.
 *When the static type _S_ of `e` is strictly non-nullable, such as when _S_
 is `Never`, `...?e` is a warning, but it may still occur.*
 
-### Instantiate to bounds
+### Instantiation to bound
 
-The computation of instantiation to bounds is changed to substitute `Never` for
+The computation of instantiation to bound is changed to substitute `Never` for
 type variables appearing in contravariant positions instead of `Null`.
+
+### Super-bounded types
+
+Null safety requires two changes to the section 'Super-Bounded Types' in the
+language specification.
+
+The definition of a top type is changed: _T_ is a top type if and only if
+`Object?` is a subtype of _T_.
+
+The definition of a super-bounded type is changed such that occurrences of
+`Null` are replaced by types involving `Never`, and `Object` is replaced by
+`Object?`. This causes one sentence to change, with the following result:
+
+Let _T'_ be the result of replacing every occurrence in _T_ of a top type in
+a covariant position by `Never`, and every occurrence in _T_ of a type
+_S_ where _S <: Never_ in a contravariant position by `Object?`.
 
 ### Least and greatest closure
 


accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -699,11 +702,27 @@ the element, key, and value type of `...e` and `...?e` is `Never`.
...
+The computation of instantiation to bound is changed to substitute `Never` for
 type variables appearing in contravariant positions instead of `Null`.
 
+### Super-bounded types
+
+Null safety requires two changes to the section 'Super-Bounded Types' in the
+language specification.
+
+The definition of a top type is changed: _T_ is a top type if and only if
+`Object?` is a subtype of _T_.
```
scheglov, line 716: Is this the same as the `TOP` predicate declared [here](https://github.com/dart-lang/language/blob/master/resources/type-system/upper-lower-bounds.md#helper-predicates)?

eernstg: This document already uses 'top type' several times, and does not refer to **TOP**, so we should in any case resolve that question.

I think **TOP** can be simplified to use the definition of 'top type' given here. If we have a top type `T`, then **TOP**(`T`) holds:

- If we used Reflexivity then we know that `T == Object?`
    OK: `TOP(Object?)` because `OBJECT(Object)`.
- Right top: `T == dynamic/void/Object?`
    OK: `TOP(dynamic)`, `TOP(void)`, `TOP(Object?)`.
- Right legacy, when `T == S*`: `Object? <: S?`
    OK: `TOP(S*)` by lemma 2, 1.
- Left nullable: `Object <: T && Null <: T`
    OK: `TOP(T)` by lemma 2: If `TOP(T)` we're done, if `OBJECT(T)` then
       `T == FutureOr^k<Object>`, contradicting `Null <: T`.
- Right FutureOr, when `T == FutureOr<S>`: `Object? <: S`
    OK: by induction `TOP(S)`, so `TOP(FutureOr<S>)`.
- Right Nullable, when `T == S?`: `Object? <: S`
    OK: by induction `TOP(S)`, so `TOP(S?)`.

Lemma 1: `Object? <: S?` => `Object <: S`
Lemma 2: `Object <: T` => `OBJECT(T) || TOP(T)`

It's rather easy to see that if `TOP(T)` then `T` is a top type.

So we should actually be able simplify `TOP`.

eernstg: Pull helper predicates to the null safety feature spec (TOP, BOTTOM, ...) and refer to them from upper-lower-bounds.md

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -699,11 +702,32 @@ the element, key, and value type of `...e` and `...?e` is `Never`.
...
+
+The definition of a super-bounded type is changed such that occurrences of
+`Null` are replaced by types involving `Never`, and `Object` is replaced by
+`Object?`. Moreover, top types in invariant positions and in positions that
+have no variance (*unused type parameters in a type alias*) are given the
+same treatment as top types in covariant positions. This causes one
+sentence to change, with the following result:
+
+Let _T'_ be the result of replacing every occurrence in _T_ of a type _S_
+in a contravariant position where _S <: Never_ by `Object?`, and every
```
scheglov, line 727: `S <: Never` includes not only `Never` itself, but also type parameters, e.g. `void Function<T extends Never>(T)`?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -699,11 +702,32 @@ the element, key, and value type of `...e` and `...?e` is `Never`.
...
+occurrence in _T_ of a top type in a position which is not contravariant by
```
scheglov, line 728: What is an example of a position that is not definitely covariant or contravariant?
I thought that a type is either a return type, or a parameter type.

The bound of a type parameter?
So, the position of the bound is the variance of the type parameter?

`T` is not used: `void Function<T extends S, S extends Object?>()`.
`T` is invariant (both covariant and contravariant): `T Function<T extends S, S extends Object?>(T)`.

accepted/future-releases/nnbd/feature-specification.md
eernstg: Yes, the point is that all such types denote the same type `Never`, only with different syntax.

accepted/future-releases/nnbd/feature-specification.md
eernstg: The type parameters of a type alias obtain their variance based on how they occur in the body (that is, how they occur in `T`, with `typedef F<X> = T`), so `typedef F2<X> = X Function()` has a covariant type parameter, `typedef F3<X> = X Function(X)` has an invariant type parameter and so does `typedef F4<X> = void Function<Y extends X>()`. An unused type parameter doesn't have any variance at all, `typedef F5<X extends C<X>> = void Function()`.

So `dynamic` occurs in a covariant position in `F2<List<dynamic>>` and in an invariant position in `F3<dynamic>` and `F3<void Function(dynamic)>.

The check for being correctly super-bounded used to be specified to leave the corresponding type arguments unchanged when the position is invariant or variance-less, and this caused some raw types to be an error:

```dart
class C<X extends C<X>> {}
typedef F5<X extends C<X>> = void Function();
F5? f; // Used to be an error.
```

The error arose because instantiation to bound would yield `F5<C<dynamic>>`, and that's not a correct super-bounded type: `F5<C<dynamic>>` is the result of replacing extreme types by their opposites according to the variance, and that isn't a regular-bounded type. The point was that we can't choose a variance for an unused type parameter, so we should abstain from taking any steps where a variance is required. Hence, `dynamic` wasn't replaced by `Never` in this case.

However, with the relaxed rules proposed here, variance-less and invariant positions are treated the same as covariant positions. This means that more top types will be replaced by `Never`, and that will change some types from being an error to being OK. In particular, the raw `F5` above is not an error any more.

It turns out that the tests that I wrote back in 2019 were used with the common front end and not with the analyzer, and also that the `//# 01: compile-time error` markers that I had written were ignored by the test runner software which is used for the common front end, so those errors were apparently never implemented. This proposal removes some errors because it would be breaking to implement them.

(And also because it's obviously safe to relax to constraints on unused type parameters of a type alias, and weird-but-not-unsound to do it for invariant parameters, and in general for extreme types in invariant positions.)

> The bound of a type parameter? ..
> `T` is not used: `void Function<T extends S, S extends Object?>()`.

That is a different situation, the "unused parameter" I was referring to is a type parameter declared by a type alias which is not used in the body of the type alias.

eernstg:
Tests under development [here](https://dart-review.googlesource.com/c/sdk/+/158397).

leafpetersen (approved):

eernstg:
Tests under review [here](https://dart-review.googlesource.com/158397).

PR #1134, opened 2020-08-05
Rework flow analysis of `==` and `!=`.

stereotype441:
Previously, we specified flow states `null()` and `nonNull()` (similar
to `true()` and `false()`), and the special behaviors of `== null` and
`!= null` checks fell out of a general analysis of `E1 == E2`.  But it
didn't work in practice, because it required doing joins like
`join(null(E1), null(E2))`, which isn't sound in general (because it
carries the flow control implication that either E1 or E2 is evaluated
but not both).  Because of these problems, the implementation never
matched this specification, and instead used ad-hoc pattern matching
to recognize `== null` and `!= null`.

In the new formulation, we remove the flow states `null()` and
`nonNull()` and retain the ad-hoc pattern matching for `== null` and
`!= null`, consistent with the current implementation.  However, we
also add cases to recognize when an equality test is guaranteed to
evaluate to `true` or `false`; this addresses
https://github.com/dart-lang/sdk/issues/41985 and ensures that we
recognize dead code associated with unnecessary null checks.

### DIFF
--- a/resources/type-system/flow-analysis.md
+++ b/resources/type-system/flow-analysis.md
@@ -434,13 +434,13 @@ Definitions:
   `Null <: T`; otherwise false.
 
 - `promote(E, T, M)` where `E` is an expression, `T` is a type which it may be
-  promoted to, and `M1 = FlowModel(r, VI)` is the flow model in which to
-  promote, is defined to be `M3`, where:
-  - If `stripParens(E)` is not a promotion target, then `M3` = `M1`
+  promoted to, and `M = FlowModel(r, VI)` is the flow model in which to promote,
+  is defined to be `M3`, where:
+  - If `stripParens(E)` is not a promotion target, then `M3` = `M`
   - If `stripParens(E)` is a promotion target `x`, then
     - Let `VM = VariableModel(declared, promoted, tested, assigned, unassigned,
       captured)` be the variable model for `x` in `VI`
-    - If `x` is not promotable via type test to `T` given `VM`, then `M3` = `M1`
+    - If `x` is not promotable via type test to `T` given `VM`, then `M3` = `M`
     - Else
       - Let `S` be the current type of `x` in `VM`
       - If `T <: S` then let `T1` = `T`
@@ -578,16 +578,16 @@ then they are all assigned the same value as `after(N)`.
 
 TODO: This isn't really right, `E1` isn't really an expression here.
 
-- **Non local-variable conditional assignment**: If `N` is an expression of the
-  form `E1 ??= E2` where `E1` is not a local variable, and the type of `E1` is
-  `T1`, then:
+- **Conditional assignment to a non local-variable**: If `N` is an expression of
+  the form `E1 ??= E2` where `E1` is not a local variable, and the type of `E1`
+  is `T1`, then:
   - Let `before(E1) = before(N)`.
-  - If `T1` is potentially nullable, then:
-    - Let `before(E2) = split(after(E1))`.
-    - Let `after(N) = merge(after(E2), split(after(E1)))`.
-  - Otherwise:
+  - If `T1` is strictly non-nullable, then:
     - Let `before(E2) = unreachable(after(E1))`.
     - Let `after(N) = after(E1)`.
+  - Otherwise:
+    - Let `before(E2) = split(after(E1))`.
+    - Let `after(N) = merge(after(E2), split(after(E1)))`.
 
   TODO(paulberry): this doesn't seem to match what's currently implemented.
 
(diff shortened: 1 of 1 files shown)


stereotype441:
I will follow up with an implementation change that brings the implementation in line with these spec changes.  A few small changes to the Flutter SDK will be necessary, to ignore dead code warnings that result from the fact that dead code analysis now detects unnecessary null checks.

stereotype441:
Implementation change: https://dart-review.googlesource.com/c/sdk/+/155926

stereotype441:
Necessary Flutter changes are here: https://github.com/flutter/flutter/pull/63007

resources/type-system/flow-analysis.md
```diff
@@ -596,14 +599,16 @@ TODO: This isn't really right, `E1` isn't really an expression here.
...
   - Let `before(E1) = before(N)`.
-  - Let `before(E2) = split(null(E1))`.
-  - Let `null(N) = unsplit(null(E2))`.
-  - Let `notNull(N) = merge(split(notNull(E1)), notNull(E2))`.
+  - If `T1` is potentially nullable, then:
+    - Let `before(E2) = split(after(E1))`.
+    - Let `after(N) = merge(after(E2), split(after(E1)))`.
+  - Otherwise:
+    - Let `before(E2) = unreachable(after(E1))`.
+    - Let `after(N) = after(E1)`.
```
eernstg, line 611: Same issue as line 592: I think we need to reorder the cases and let them target "`T1` is strictly non-nullable" respectively "otherwise".

resources/type-system/flow-analysis.md
```diff
@@ -623,27 +628,21 @@ TODO: This isn't really right, `E1` isn't really an expression here.
...
-    - Let `null(N) = before(N)`.
-    - Let `notNull(N) = unreachable(before(N))`.
-  - Otherwise if `T` is non-nullable then:
-    - Let `null(N) = before(N)`.
-    - Let `notNull(N) = unreachable(before(N))`.
+    - Let `after(N) = unreachable(after(E2))`.
   - Otherwise:
-    - Let `null(N) = promote(x, Null, before(N))`
-    - Let `notNull(N) = promoteToNonNull(x, before(N))`
+    - Let `after(N) = promoteToNonNull(E1, after(E2))`
```
eernstg, line 641: That wouldn't be true for members of the interface of `Object`, when invoked in a way that conforms to the signature. For instance `E1` _can_ be a variable with the value null after evaluating `E1.toString()` or `E1.toString`, but we _do_ know that it is non-null after evaluating `E1.toString(named: 24)` (but in that case `E1` is guaranteed to have type `dynamic` so I guess promotion to non-null won't do anything, and it's in any case sound to ignore that we have that bit of extra knowledge).

So: Don't promote for members of the interface of `Object`.

resources/type-system/flow-analysis.md
```diff
@@ -429,11 +422,22 @@ Definitions:
...
+  - If `E1` is a parenthesized expression of the form `(E2)`, then `E3` =
+    `stripParens(E2)`.
+  - Otherwise, `E3` = `E1`.
+
+- `equivalentToNull(T)`, where `T` is a type, indicates whether `T` is
+  equivalent to the `Null` type.  It is defined to be true if `T <: Null` and
+  `Null <: T`; otherwise false.
+
 - `promote(E, T, M)` where `E` is an expression, `T` is a type which it may be
   promoted to, and `M1 = FlowModel(r, VI)` is the flow model in which to
```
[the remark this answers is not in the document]

resources/type-system/flow-analysis.md
```diff
@@ -582,11 +578,18 @@ wrong.  This needs reconsideration.
 
 TODO: This isn't really right, `E1` isn't really an expression here.
 
-- **Non local-variable conditional assignment**: If `N` is an expression of the form
-  `E1 ??= E2` where `E1` is not a local variable, then:
-  - Let `before(E1) = before(N)`
-  - Let `before(E2) = split(null(E1))`.
-  - Let `after(N) = merge(after(E2), split(notNull(E1)))`
+- **Non local-variable conditional assignment**: If `N` is an expression of the
```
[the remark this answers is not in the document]
stereotype441: Fixed

resources/type-system/flow-analysis.md
```diff
@@ -582,11 +578,18 @@ wrong.  This needs reconsideration.
...
+  `T1`, then:
+  - Let `before(E1) = before(N)`.
+  - If `T1` is potentially nullable, then:
+    - Let `before(E2) = split(after(E1))`.
+    - Let `after(N) = merge(after(E2), split(after(E1)))`.
+  - Otherwise:
+    - Let `before(E2) = unreachable(after(E1))`.
+    - Let `after(N) = after(E1)`.
+
+  TODO(paulberry): this doesn't seem to match what's currently implemented.
```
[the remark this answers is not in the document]
stereotype441: Good point--I had forgotten about the distinction between "non-nullable" and "strictly non-nullable".  Fixed.  At some point later I'll re-read through the whole spec to make sure there aren't any other cases of "non-nullable" (either explicitly, or implicitly as in this case) that need to be changed to "strictly non-nullable".

resources/type-system/flow-analysis.md
stereotype441: Fixed

resources/type-system/flow-analysis.md
stereotype441: Fixed, thanks!

Issue #1129, opened 2020-08-04
Dart numbers, when compiled to JavaScript, are unsigned 32 bit


eernstg:
Cf. https://github.com/dart-lang/sdk/issues/42892#issuecomment-667414031, the appendix about the treatment of numbers in Dart when compiled to JavaScript needs to be updated such that it mentions that bit operations consider the operands to be unsigned 32 bit values rather than signed ones:

> We chose to make all bit operations unsigned because it makes more code work
> than accepting JavaScript's signed behaviour. In particular the following kinds of
> code often have bugs with signed values: implementations of bit-arrays as an array
> of 32-bit integers; code that packs bool or int fields into an integer; code that does
> rotates by combining shifts.
>
> If a signed result is needed it can be obtained via `.toSigned(32)`, e.g. `(-1 << 10).toSigned(32)`.

The current wording implies that Dart uses the JavaScript semantics.

eernstg:
Spec adjustment [here](https://github.com/dart-lang/language/pull/1145).

eernstg:
Tests [here](https://dart-review.googlesource.com/c/sdk/+/158014).

PR #1145, opened 2020-08-11
Specify that JavaScript compiled numbers use unsigned bitops

eernstg:
Cf. https://github.com/dart-lang/language/issues/1129. The decision to change the specification rather than `dart2js` was supported at the language team meeting Aug 5.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -19062,8 +19062,8 @@ which means that this seemingly expensive step can be confined to some extent.
 The \code{int} type represents integers.
 The specification is written with 64-bit two's complement integers as the
 intended implementation, but when Dart is compiled to JavaScript,
-the implementation of \code{int} will instead use the JavaScript
-number type.
+the implementation of \code{int} will instead use the JavaScript number type,
+with bit operations applied to the low 32 bits as an unsigned value.
 
 This introduces a number of differencs:
 \begin{itemize}
@@ -19072,12 +19072,13 @@ Valid values of JavaScript \code{int} are any
 IEEE-754 64-bit floating point number with no fractional part.
 This includes positive and negative \Index{infinity},
 which can be reached by overflowing
-(integer division by zero is still not allowed).
+(integer division by zero is still a dynamic error).
 Otherwise valid integer literals (including any leading minus sign)
 that represent invalid JavaScript \code{int} values
-cannot be compiled to JavaScript.
-Operations on integers may lose precision since 64-bit floating point numbers
-are limited to 53 significant bits.
+are compile-time errors when compiling to JavaScript.
+Operations on integers may lose precision,
+because they are represented as 64-bit floating point numbers
+that are limited to 53 significant bits.
 \item[$\bullet$]
 JavaScript \code{int} instances also implement \code{double},
 and integer-valued \code{double} instances also implement \code{int}.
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -19062,8 +19062,8 @@ \section*{Appendix: Integer Implementations}
 The \code{int} type represents integers.
 The specification is written with 64-bit two's complement integers as the
 intended implementation, but when Dart is compiled to JavaScript,
-the implementation of \code{int} will instead use the JavaScript
-number type.
+the implementation of \code{int} will instead use the JavaScript number type,
+with bit operations applied to the low 32 bits as an unsigned value.
```
lrhn, line 19066: > with bit-wise operations only work on the first 32 bits of an integer value. 
> The operands are first truncated to 32-bit two's complement integers, then 32-bit bitwise operations are applied, and finally the resulting 32 bits are interpreted as an unsigned 32-bit integer.

I don't want to say that the operands are unsigned, because `>>` is sign-extending (so it must have a sign, so it can't be "unsigned"). The above description is precise. JavaScript converts bit-operation operands to 32-bit integers using their spec-function ToInt32, then it performs the bit-operation on that, and finally Dart performs a `>>>0` on the result, which calls `ToUint32` with the result.

specification/dartLangSpec.tex
```diff
@@ -19072,23 +19072,27 @@ \section*{Appendix: Integer Implementations}
...
-(integer division by zero is still not allowed).
+(integer division by zero is still a dynamic error).
 Otherwise valid integer literals (including any leading minus sign)
 that represent invalid JavaScript \code{int} values
-cannot be compiled to JavaScript.
-Operations on integers may lose precision since 64-bit floating point numbers
-are limited to 53 significant bits.
+are compile-time errors when compiling to JavaScript.
+Operations on integers may lose precision,
+because they are represented as 64-bit floating point numbers
```
lrhn, line 19080: > ... because the result is represented as a 64-bit floating point number,
> which are limited to 53 significant bits.

The "they" is unclear, it can refer both to integers and operations, and it makes sense for neither since it is the result which may lose precision.

specification/dartLangSpec.tex
```diff
@@ -19072,23 +19072,27 @@ \section*{Appendix: Integer Implementations}
...
 act as if they are actually instances of a common subclass implementing both
-\code{int} and \code{double}. Fractional numbers only implement \code{double}.
+\code{int} and \code{double}.
+Fractional numbers only implement \code{double}.
 \item[$\bullet$]
-Bitwise operations on integers (and, or, xor, negate and shifts)
-all truncate the operands to 32-bit values.
+Bitwise operations on integers (and, or, xor, negate, and shifts)
+all truncate the operands to 32-bit values
+and operate on them as unsigned values
```
lrhn, line 19094: This seems to be where the details go, so:

> Bitwise operations in integers (and, or, xor, negation and shifts)
> all truncate the operands to 32-bit two's complement integers,
> perform 32-bit operations on those,
> and then the resulting 32 bits are then interpreted as a 32-bit unsigned integer.

I don't want to say that the *operands* are unsigned, because `>>` is sign-extending (so it must have a sign, so it can't be "unsigned"). The above description is precisely following the JavaScript spec, which converts bit-operation operands to 32-bit integers using their spec-function ToInt32, then it performs the bit-operation on those bits, and then interpret then as signed or unsigned 32-bit integers. Finally Dart performs an extra `>>>0` on the result, which calls `ToUint32`, to make it unsigned.

specification/dartLangSpec.tex
eernstg: The point is that "when compiling to JS, Dart uses the JavaScript number type" seems to imply that the semantics of all operations are taken directly from similarly named operations in JS. So I wanted to add "but with respect to bit operations, the semantics is different".

I changed the wording to make this more explicit.

specification/dartLangSpec.tex
eernstg: Clarified the wording: Both the operands and the result are represented as mentioned, which implies that precision may be lost (for instance, when the result is an odd integer >=2^53).

specification/dartLangSpec.tex
eernstg: Done.

eernstg:
> I think we should give this _one more though_ before codifying the current behavior.

This PR is concerned with the adjustment that makes the language specification (appendix) describe the current behavior faithfully.

Wouldn't it be reasonable to handle the potential breaking change in a different PR? Otherwise it seems that we wouldn't be able to land this before we've had a lengthy breaking change process.

eernstg:
@lrhn, PTAL: I did not make any changes to this PR concerned with the breaking change. It seems reasonable to me to handle that issue separately.

eernstg:
@lrhn, are you happy about handling the breaking change separately (that is, with a language repo discussion and a new PR)?

In that case I'll land this one, and since it has no implementation implications this should be a lightweight step to take.

Issue #1120, opened 2020-07-28
Static and runtime errors for `main` seem under-specified


leafpetersen:
There seem to be no static errors specified for the signature provided for `main`, and the runtime semantics don't seem to make it clear that as a result the invocation needs to do a dynamic type check.  In practice, the VM at least seems to do so. 
This program:

```dart
void main(int x) {
  x.isEven;
}
```

produces the following runtime error when run:

```
Unhandled exception:
type 'List<String>' is not a subtype of type 'int' of 'x'
#0      _startIsolate.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:299:32)
#1      _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:168:12)
```

This seems like it needs clarification in the spec, and possibly tests.

natebosch:
Adding any static error here will be breaking. If we add one it needs to be gated by the language version.

The name `main` is not (currently) reserved for the entrypoint method. You can define `main` in any library and invoke it with any arguments.

 I'm not opposed to it, but we should be aware that we are enforcing a (quite reasonable) _convention_ by adding static checking to the signature of functions named `main`.

leafpetersen:
> The name `main` is not (currently) reserved for the entrypoint method. You can define `main` in any library and invoke it with any arguments.

That's not quite true (at least per spec).  The restrictions are small, but as I understand the spec (quoted below):
  - Any symbol named "main" is treated as an entry point
  - It is a compile time error if that symbol is a non-function, and has more than two positional parameters
  - That symbol will be invoked according to a specific procedure.

It's just that the spec above is not restrictive enough to be sound.

```
A script $S$ is executed as follows:

\LMHash{}%
First, $S$ is compiled as a library as specified above.
Then, the top-level function defined by \code{main}
in the exported namespace of $S$ is invoked (\ref{functionInvocation})
as follows:
If \code{main} can be called with with two positional arguments,
it is invoked with the following two actual arguments:
\begin{enumerate}
\item An object whose run-time type implements \code{List<String>}.
\item An object specified when the current isolate $i$ was created,
for example through the invocation of \code{Isolate.spawnUri} that spawned $i$,
or the null object (\ref{null}) if no such object was supplied.
\end{enumerate}
If \code{main} cannot be called with two positional arguments,
but it can be called with one positional argument,
it is invoked with an object whose run-time type implements \code{List<String>}
as the only argument.
[...]

PR #1146, opened 2020-08-12
Propose rules about `main` for null-safety

eernstg:
This is a proposal for new rules about `main`, adding a new section in the null safety spec (which will replace the section 'Scripts' in the language specification in the future).

Potentially noteworthy aspects (compared to [this](https://github.com/dart-lang/language/issues/1120#issuecomment-670097180)):

This proposal maintains that the initial function has the name `main`. I think it's helpful for developers to know for sure that the initial function has this name, so that they don't suddenly have to use some other name, just because they have switched to use a different compiler.

This proposal still defines the term _script_; however, it is now a very straightforward concept. If we think the word 'script' sounds too "PHP-ish" then we can rename it (here, and later in the language specification in general), but it seems reasonable to be able to talk about the ability to be the provider of the function that defines what it means to run a program (the "initial function").

I have preserved the specification of how to invoke `main` (with 2, 1, or 0 arguments), but broadened it to allow implementations to pass additional arguments. This means that an embedder can allow or require `main` to have additional optional parameters (positional or named), and it can pass whatever actual arguments it wants to those additional parameters.

But we do maintain that `T main()` for any `T`, `T main(U u)` where `List<String> <: U`, and `T main(U u, V v)` _must_ work (except for the catchall clause that "implementations can have arbitrary extra constraints on the signature of `main`"). Also, all invocations must enforce the declared types, and we do statically prevent weird signatures like `void main(int i)` and `void main(arg1, arg2, arg3)`.

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -910,11 +910,12 @@ The section 'Scripts' in the language specification is replaced by the
 following:
 
 Let _L_ be a library that exports a declaration _D_ named `main`.  It is a
-compile-time error unless _D_ is a function.  It is a compile-time error if
-_D_ declares more than two required positional parameters, or if there are
-any required named parameters.  It is a compile-time error if _D_ declares
-at least one positional parameter, and the first positional parameter has a
-type which is not a supertype of `List<String>`.
+compile-time error unless _D_ is a function declaration.  It is a
+compile-time error if _D_ declares more than two required positional
+parameters, or if there are any required named parameters.  It is a
+compile-time error if _D_ declares at least one positional parameter, and
+the first positional parameter has a type which is not a supertype of
+`List<String>`.
 
 Implementations are free to impose any additional restrictions on the
 signature of `main`.
@@ -933,7 +934,7 @@ it is invoked with the following two actual arguments:
 - An object specified when the current isolate _i_ was created,
   for example through the invocation of `Isolate.spawnUri` that spawned _i_,
   or the null object if no such object was supplied.
-  A dynamic error occurs unless the run-time type of this object is a
+  A dynamic error occurs if the run-time type of this object is not a
   subtype of the declared type of the corresponding parameter of `main`.
 
 If `main` cannot be called with two positional arguments, but it can be
@@ -944,12 +945,14 @@ If `main` cannot be called with one or two positional arguments, it is
 invoked with no arguments.
 
 In each of the above three cases, an implementation is free to provide
-additional arguments allowed by the signature of `main`, but must ensure
-that a dynamic error occurs if an actual argument does not have a run-time
-type which is a subtype of the declared type of the parameter.
+additional arguments allowed by the signature of `main` (*the above rules
+ensure that the corresponding parameters are optional*).  But the
+implementation must ensure that a dynamic error occurs if an actual
+argument does not have a run-time type which is a subtype of the declared
+type of the parameter.
 
-A Dart program will typically be executed by executing a script.
-The procedure whereby this script is chosen is implementation specific.
+A Dart program will typically be executed by executing a script.  The
+procedure whereby this script is chosen is implementation specific.
 
 ## Runtime semantics
 
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -6,6 +6,9 @@ Status: Draft
 
 ## CHANGELOG
 
+2020.08.12
+  - Specify constraints on the `main` function.
+
 2020.08.06
   - Specify error for uninitialized final instance variable in class
     with no generative constructors.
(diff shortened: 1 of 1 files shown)


accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -901,6 +904,53 @@ These are extended as
 per
 [separate proposal](https://github.com/dart-lang/language/blob/master/resources/type-system/flow-analysis.md).
 
+### The main function
+
+The section 'Scripts' in the language specification is replaced by the
+following:
+
+Let _L_ be a library that exports a declaration _D_ named `main`.  It is a
```
natebosch, line 912: To make sure I understand - are we considering going forward with a breaking change in the implementations and applying these restrictions to _all_ declarations named `main`?

eernstg: My intention was to make these changes applicable to code with null safety only, by putting them into the null safety spec.

In the past, we have considered changes that may break opted-in code as non-breaking, but we will of course need to consider them breaking when the amount of pre-existing opted-in code is large enough, and in particular when null safety has been released as a stable feature.

At this point I think it is still fair to say that it is not a breaking change.

natebosch: > At this point I think it is still fair to say that it is not a breaking change.

I think the edge where it would impede a migration to null safety (if there is a package with a non-conforming `main` in the public API) is so unlikely that's probably safe to consider it that way.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -901,6 +904,53 @@ These are extended as
...
+compile-time error unless _D_ is a function.  It is a compile-time error if
```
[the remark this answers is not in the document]
eernstg: Done.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -901,6 +904,53 @@ These are extended as
...
+in the exported namespace of _L_ is invoked as follows:
+
+If `main` can be called with with two positional arguments,
+it is invoked with the following two actual arguments:
+
+- An object whose run-time type implements `List<String>`.
+- An object specified when the current isolate _i_ was created,
+  for example through the invocation of `Isolate.spawnUri` that spawned _i_,
+  or the null object if no such object was supplied.
+  A dynamic error occurs unless the run-time type of this object is a
```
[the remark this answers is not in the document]
eernstg: Done.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -901,6 +904,53 @@ These are extended as
...
+
+If `main` cannot be called with two positional arguments, but it can be
+called with one positional argument, it is invoked with an object whose
+run-time type implements `List<String>` as the only argument.
+
+If `main` cannot be called with one or two positional arguments, it is
+invoked with no arguments.
+
+In each of the above three cases, an implementation is free to provide
+additional arguments allowed by the signature of `main`, but must ensure
```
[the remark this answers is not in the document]
eernstg: Added commentary to say that those parameters must be optional.

eernstg:
@lrhn wrote:
> How does this actually differ from what we have now?

It specifies that it is a compile-time error to have a first positional argument whose type isn't a supertype of `List<String>` (so we get static errors for a case like `void main(int args)`). It allows the positional parameters to be optional (allowing `void main([List<String> args])` and similar signatures which used to be a compile-time error), and it allows additional optional positional parameters (making the spec text allow signatures that the implementations actually allow today). It also allows optional named parameters (again, allowed today by implementations).

It allows implementations to add their own constraints on the signature of `main` (which is desired by the `dart2js` team), but hints that such constraints should be reasonable. It's a matter of good taste to ensure this.

Finally, it specifies that dynamic errors must occur such that the invocation of `main` is type safe in the sense that actual arguments must have the declared parameter types. I claimed that this would be implied because it is just another function call (and all of them must be sound), but it could be argued that the invocation of the initial function is somehow a "special" function call which is not automatically covered by the existing rules. The text given here makes it explicit that we do have the guarantee for that kind of soundness.

eernstg:
@leafpetersen, @rakudrama, @natebosch, do you have further comments on this PR?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -901,6 +904,56 @@ These are extended as
 per
 [separate proposal](https://github.com/dart-lang/language/blob/master/resources/type-system/flow-analysis.md).
 
+### The main function
+
+The section 'Scripts' in the language specification is replaced by the
+following:
+
+Let _L_ be a library that exports a declaration _D_ named `main`.  It is a
```
johnniwinther, line 912: This includes setters, right?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -901,6 +904,56 @@ These are extended as
...
+compile-time error unless _D_ is a function declaration.  It is a
```
johnniwinther, line 913: This excludes getters, right? co19/Language/Libraries_and_Scripts/Scripts/top_level_main_t05 seems to assume it doesn't.

accepted/future-releases/nnbd/feature-specification.md
eernstg: No, that would not apply to setters, because the spec insists that their name ends in `=`. It may seem silly that we can have a setter like `set main(_) {}` when we can't have `Function get main => ...;` but it is a consequence of the existing terminology, and I don't think it causes any harm.

johnniwinther: Since no user thinks in terms of this, wouldn't it be more coherent just to ban 'main=' as well? (Dart tells me that 'main' must be a function declaration but I'm still allowed to create a setter).

accepted/future-releases/nnbd/feature-specification.md
eernstg: It is intended to exclude getters. The old version of this text in section `Scripts` said explicitly in commentary that we couldn't have a 'top-level getter or field named `main`', and we did not discuss nor intend to change that when this spec change was landed.

However, section 'Functions' implies that 'function' includes getters, setters, constructors, and "functions" (no, we don't have a separate term for the ones that have a `<formalParameterList>` and aren't constructors, I had a CL a long time ago to fix that, but it hasn't been landed). So the text should say `unless _D_ is a non-getter function declaration`. It can't be a top-level constructor and a setter can't have the name `main`, so that should be enough.

eernstg: https://github.com/dart-lang/language/pull/1146

accepted/future-releases/nnbd/feature-specification.md
eernstg: I added this proposal in a comment on #1248. It's in principle a breaking change, but I agree that it is confusing that `set main(_) {}` is allowed.

johnniwinther: It's only breaking in null safe code, right? We'll still allow whatever in legacy.

eernstg: True, this is a null-safety only update.

accepted/future-releases/nnbd/feature-specification.md
rakudrama: When calling `main` from external code, e.g. JavaScript, it is necessary to _convert_ arguments. We can't just say that `main` is called dynamically from 'outside'. This conversion happens in the context of the program but before entering `main`, e.g. needs to access the type `List<String>` for that program. This interstitial place a special place that we want to constrain heavily in a platform dependent way. For dart2js we want limit the conversion to just `Array`-to-`List<String>`, for which we need to know the static type of the entry point.

leafpetersen: @rakudrama does line 920 not address your concerns?

rakudrama: Yes, it does. I was just adding colour for Johnni.

PR #1156, opened 2020-08-17
Specify extension methods

eernstg:
This PR integrates the extension methods feature spec into the language specification.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -611,9 +611,11 @@ Let $e_{\metavar{let}}$ be a \LET{} expression of the form
 \code{\LET\,\,$v_1$\,=\,$e_1$, $\cdots$, $v_k$\,=\,$e_k$\ \IN\ $e$}.
 It is required that $v_j$ is a fresh variable, $j \in 1 .. k$
 (\commentary{which is tacitly assumed whenever a \LET{} expression is used}).
-$e_{\metavar{let}}$ contains $k$ nested scopes, \List{S}{1}{k};
-the current scope of $e_1$ is the current scope of $e_{\metavar{let}}$,
-the current scope of $e_{j+1}$ is $S_j$, $j \in 1 .. k - 1$,
+$e_{\metavar{let}}$ contains $k$ nested scopes, \List{S}{1}{k}.
+The enclosing scope for $S_1$ is the current scope for $e_{\metavar{let}}$,
+and the enclosing scope for $S_j$ is $S_{j-1}$, $j \in 2 .. k$.
+The current scope of $e_1$ is the current scope of $e_{\metavar{let}}$,
+the current scope of $e_j$ is $S_{j-1}$, $j \in 2 .. k$,
 and the current scope of $e$ is $S_k$.
 For $j \in 1 .. k$, $v_j$ introduces a final, local variable into $S_j$,
 with the static type of $e_j$ as its declared type.
@@ -5366,6 +5368,14 @@ except that static extensions can also declare operators.%
 }
 
 \LMHash{}%
+A static extension application does not have a type.
+
+\commentary{%
+This is consistent with the fact that any use of a static extension application
+where a type is needed is a compile-time error.%
+}
+
+\LMHash{}%
 Let $i$ be a simple member invocation
 (\ref{memberInvocations})
 whose receiver $r$ is a static extension application of the form
@@ -5520,7 +5530,7 @@ whenever the declared name is in scope,
 it is sufficient to consider the fresh name.
 %
 When the fresh name of $E$ is in the library scope,
-it is availably in \emph{any} scope,
+it is available in \emph{any} scope,
 because the name is fresh and hence it cannot be shadowed
 by any declaration in any intermediate scope.
 %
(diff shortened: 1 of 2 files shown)


eernstg:
@lrhn, could you take a look at this one? I just resolved conflicts, so it's based on a fresh master commit now.

specification/dartLangSpec.tex
```diff
@@ -5028,6 +5078,1074 @@ \subsection{Mixin Application}
...
+extension denoted by $E$ which is accessible
+(\ref{extensionAccessibility})
+and applicable
+(\ref{extensionApplicability})
+to $i$.
+
+\LMHash{}%
+In the case where no compile-time error occurs,
+$i$ is treated as $i'$, which is obtained from $i$ by
+replacing $r$ by \code{$E$($r$)}.
```
lrhn, line 5541: by replacing $r$ by -> by replacing the leading $r$ by

(Just to cover the case where `r` occurs more than once in the invocation.)

specification/dartLangSpec.tex
```diff
@@ -5028,6 +5078,1074 @@ \subsection{Mixin Application}
...
+the context type may affect the member invocation.%
+}
+
+
+\subsubsection{Specificity of an Extension}
+\LMLabel{extensionSpecificity}
+
+\LMHash{}%
+When \List{E}{1}{k}, $k > 1$, are extensions
+which are is accessible and applicable to a member invocation $e$
```
lrhn, line 5672: remove "is"

specification/dartLangSpec.tex
```diff
@@ -5028,6 +5078,1074 @@ \subsection{Mixin Application}
...
+As specified elsewhere
+(\ref{lexicalLookup}),
+this implies that extension instance members
+will shadow class instance members
+when called from another instance member inside the same extension
+using an unqualified function invocation
+(that is, invoking it as \code{m()} and not \code{\THIS.m()},
+\ref{unqualifiedInvocation}).
+%
+This is an exception,
```
lrhn, line 5819: To what?

Maybe just say:

> This is the only situation where implicit invocation of an extension member with basename \id{} can succeed even if the interface of the receiver has a member with basename \id{}.

specification/dartLangSpec.tex
```diff
@@ -5028,6 +5078,1074 @@ \subsection{Mixin Application}
...
+and $T_j = [S_1/Y_1, \ldots, S_m/Y_m]T'_j, j \in 1 .. n+k$,
+where $T'_j$ is the type of the corresponding parameter in
+the declaration of \id.
+Capture of type variables in \List{S}{1}{m} must be avoided,
+so $X_j$ must be renamed if \List{S}{1}{m} contains any occurrences of $X_j$,
+for all $j \in 1 .. s$.
+
+\commentary{%
+In other words, the closurization is
+a function literal whose signature is the same as that of \id,
```
lrhn, line 5962: ... is the value of a function literal ...

specification/dartLangSpec.tex
```diff
@@ -11404,6 +12542,264 @@ \subsection{Top level Getter Invocation}
...
+        \code{$r$?.\id{} = $e$} &
+        \code{$r'$\,==\,\NULL\ ?\ \NULL\ :\ $r'$.\id{} = $e$}
+        \\
+        \code{$r$?.\id(\metavar{args})} &
+        \code{$r'$\,==\,\NULL\ ?\ \NULL\ :\ $r'$.\id(\metavar{args})}
+        \\
+        \code{$r$?.\id<\metavar{types}>(\metavar{args})} &
+        \code{$r'$\,==\,\NULL\ ?\ \NULL\ :\ $r$.\id<\metavar{types}>(\metavar{args})}
+        \\
+        %% TODO(eernst): Come nnbd, uncomment this.
```
lrhn, line 12691: NNBD has come. (Or are we aiming for a 2.8 spec first?)

specification/dartLangSpec.tex
```diff
@@ -11404,6 +12542,264 @@ \subsection{Top level Getter Invocation}
...
+as the receiver of a member invocation,
+and otherwise $r'$ is $v_r$.
+
+\LMHash{}%
+When $r$ is not an extension application,
+$r'$ is a fresh variable bound to the value of $r$,
+with the same static type as $r$.
+
+\commentary{%
+This corresponds to an extra outermost \LET{} in a number of rules
```
lrhn, line 12795: So this is where you use \LET?

I can see how it's very convenient. The table does not have room for the prose version.
(Now, if we had declaration expressions, like `(var fresh = e).foo = fresh.foo + 2`, then we wouldn't need the "\LET" here. 😁 Should we make `\LET` into a macro so it's easier to change the syntax in the future?).

specification/dartLangSpec.tex
```diff
@@ -13966,8 +15437,10 @@ \subsection{Lexical Lookup}
   (\ref{classes}).
 \item
   When $\ell$ has access to \THIS{}
```
lrhn, line 15439: Where was `\ell` specified? I haven't expanded *everything*, but I can't find a reference close above (like, in the first 4k lines)

Also what does "has access to \THIS{}" mean? That is accesses `this` or that it *could* access `this` if it wanted to. From the context, it sounds like the former. Consider rewriting to make it clear.

lrhn (approved):

eernstg:
@lrhn, that's great, thanks! Haven't started working on it, but that'll be real soon. ;)

eernstg (commented):
Thanks for helpful comments! I think the issues have been addressed, but you could take a look if you wish.

specification/dartLangSpec.tex
```diff
@@ -13969,8 +15441,12 @@ \subsection{Lexical Lookup}
...
-  it is a compile-time error if the interface of the enclosing class
-  does not have a member named $n$.
+  it is a compile-time error if $\ell$ occurs in
+  a class \DefineSymbol{C} and
+  the interface of $C$ does not have a member named $n$;
+  and it is a compile-time error if $\ell$ occurs in
+  an extension \DefineSymbol{E},
+  and the interface of the \ON{} type of $E$ does not have a member named $n$.
```
leafpetersen, line 15449: @eernstg  I don't think this is correct, unless I'm missing something.  Consider:

```dart
extension F on int {
  void foo() {}
}

extension G on int {
  void bar() {
    foo();
  }
}
```

I believe this implies that the call to `foo()` inside of `bar` should be an error, right?

eernstg: Good catch! We should indeed take implicit extension member invocations into account here.

Cf. https://github.com/dart-lang/language/pull/1501.

I'll close this thread because the discussion can be handled in the new PR.

PR #1162, opened 2020-08-19
Added null-shorting rule to treat selector `!` like other selectors

eernstg:
Cf. https://github.com/dart-lang/sdk/issues/43093.

Update: That issue is now closed, and more recent discussions can be found in #1163.

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -6,6 +6,9 @@ Status: Draft
 
 ## CHANGELOG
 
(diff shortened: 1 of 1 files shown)


eernstg:
@lrhn, didn't we conclude that `!` participates in null-shorting, both when it occurs in a chain of selectors and when it occurs at the end (the language team meeting minutes are somewhat ambiguous)?

@leafpetersen, do you agree that this behavior can be specified as proposed in this PR?

leafpetersen (approved):
LGTM, thanks.

Issue #1193, opened 2020-09-01
Update needed in rule for UP on F-bounded type variables


eernstg:
The following program demonstrates that we can have a variable of type `Object` that stores the value null, running with sound null safety:

```dart
bool b = true;

class C<X extends C<X, X>?, Y extends C<Y, Y>?> {
  X x;
  C(this.x);
  Object m(X x, Y y) {
    // UP(X extends C<X, X>?, Y extends C<Y, Y>?) ==
    // C<Object, Object>?.
    var z = b ? x : y;
    if (z == null) throw 0;
    return z.x;
  }
}

class D<X> extends C<D<X>?, D<X>?> {
  D(): super(null);
}

main() {
  var d = D<Null>();
  Object o = d.m(d, null);
  print("o == $o"); // 'o == null'.
}
```

I believe that this behavior is correct according to the specification of the least upper bound, cf. [this section](https://github.com/dart-lang/language/blob/master/resources/type-system/upper-lower-bounds.md#upper-bounds), but the use of `Object` should be updated to `Object?` in the 4 rules about type variables and promoted type variables, in order to maintain soundness for this example.

Apart from the soundness issue, it is also unfortunate that the type obtained by substitutions like `[Object/X1]` is not well-bounded (`C<Object, Object>` is not a correctly super-bounded type). That's fixed by using `Object?` for the cases where the type variable occurs in a covariant position.

However, when it occurs in a contravariant position we may get an unsoundly special type:

```dart
bool b = true;

void f(Object o) {}

class C<X extends void Function(X)?> {
  X x;
  C(this.x);
  void m() {
    // UP(X extends void Function(X)?, void Function(Object)) ==
    // void Function(Object)?.
    var z = b ? x : f;
    if (z == null) return;
    z(42); // `z` unsoundly has type `void Function(Object)`.
  }
}

main() {
  var c = C<void Function(Function?)?>((Function? f) => f?.call());
  c.m(); // Dynamic error because `42` isn't a `Function?`.
}
```

This issue wouldn't be eliminated by changing `Object` to `Object?`.

We might need to consider using instantiation to bound when that yields a supertype of the bound of the type variable (that works in almost all situations, and we can fall back to a more general type like `Object?` when it doesn't work). That is, if we're approximating the bound for `X` then we'd replace contravariant occurrences of `X` by `Never` and other occurrences of `X` by `Object?`, and then we'd check that the resulting type is a supertype of the specified bound, assuming the specified bound.

[...]

PR #1195, opened 2020-09-02
Adjust UP for null safety, cf. language issue #1193

eernstg:
As mentioned in #1193, the rules for **UP** involving type variables need to be adjusted for soundness. This PR makes an adjustment to achieve this (as a special case of instantiation to bound plus a check).

It also adjusts the rules for **UP**(`T Function<...>(...)`, `T2`) and **UP**(`T1`, `T Function<...>(...)`), where the current rule rule may incorrectly discard nullability.

### DIFF
--- a/resources/type-system/upper-lower-bounds.md
+++ b/resources/type-system/upper-lower-bounds.md
@@ -167,37 +167,29 @@ We define the upper bound of two types T1 and T2 to be **UP**(`T1`,`T2`) as foll
   - `T2` if `X1 <: T2`
   - otherwise `X1` if `T2 <: X1`
   - otherwise **UP**(`B1a`, `T2`)
-    where `B1a` is obtained from `B1` by replacing contravariant occurrences
-    of `X1` by `Never`, and other occurrences of `X1` by `Object?`
-    when the resulting type is a supertype of `B1` assuming `B1`,
-    and `B1a` is otherwise `Object?`.
+    where `B1a` is the greatest closure of `B1` with respect to `X1`,
+    as defined in [inference.md].
 
 - **UP**(`X1 & B1`, `T2`) =
   - `T2` if `X1 <: T2`
   - otherwise `X1` if `T2 <: X1`
   - otherwise **UP**(`B1a`, `T2`)
-    where `B1a` is obtained from `B1` by replacing contravariant occurrences
-    of `X1` by `Never`, and other occurrences of `X1` by `Object?`
-    when the resulting type is a supertype of `B1` assuming `B1`,
-    and `B1a` is otherwise `Object?`.
+    where `B1a` is the greatest closure of `B1` with respect to `X1`,
+    as defined in [inference.md].
 
 - **UP**(`T1`, `X2 extends B2`) =
   - `X2` if `T1 <: X2`
   - otherwise `T1` if `X2 <: T1`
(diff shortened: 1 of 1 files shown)


resources/type-system/upper-lower-bounds.md
```diff
@@ -14,7 +18,7 @@ This documents the currently implemented upper and lower bound computation,
 modified to account for explicit nullability and the accompanying type system
 changes (including the legacy types).  In the interest of backwards
 compatibility, it does not try to fix the various issues with the existing
-algorithm.  
+algorithm.
```
eernstg, line 21: Just removing trailing space.

resources/type-system/upper-lower-bounds.md
```diff
@@ -213,8 +233,8 @@ We define the upper bound of two types T1 and T2 to be **UP**(`T1`,`T2`) as foll
           `Named1`
 
 - **UP**(`T Function<...>(...)`, `S Function<...>(...)`) = `Function` otherwise
-- **UP**(`T Function<...>(...)`, `T2`) = `Object`
-- **UP**(`T1`, `T Function<...>(...)`) = `Object`
+- **UP**(`T Function<...>(...)`, `T2`) = **UP**(`Object`, `T2`)
```
eernstg, line 236: Could use `Object?`, but this is slightly better when `T2` is non-nullable.

resources/type-system/upper-lower-bounds.md
```diff
@@ -420,7 +440,7 @@ that are not identical are `T0` and `S0` respectively, and **MORETOP**(`T0`,
 `S0`).
 
 A similar treatment would need to be done for the bottom types as well, since
-there are two equivalences there. 
+there are two equivalences there.
```
eernstg, line 443: Just deleting trailing space.

resources/type-system/upper-lower-bounds.md
```diff
@@ -162,22 +166,38 @@ We define the upper bound of two types T1 and T2 to be **UP**(`T1`,`T2`) as foll
 - **UP**(`X1 extends B1`, `T2`) =
   - `T2` if `X1 <: T2`
   - otherwise `X1` if `T2 <: X1`
-  - otherwise **UP**(`B1[Object/X1]`, `T2`)
+  - otherwise **UP**(`B1a`, `T2`)
+    where `B1a` is obtained from `B1` by replacing contravariant occurrences
```
leafpetersen, line 170: I'm not quite parsing this text well.  I think what we want here is something like:

```
otherwise **UP(`B1a`, `T2`) where `B1a` is the greatest closure of `B1` with respect to `X1` as defined in inference.md
```

WDYT?

eernstg: Yes, I think that will work, based on the assumption that we first expand all type aliases (so invariant occurrences of a type variable is simply an occurrence in the bound of a type parameter of a function type).

With declaration-site variance added in we can't expand away all the invariant type parameters, but we would then revisit the definition of 'greatest closure of a type'.

eernstg: Done.

eernstg:
@leafpetersen, do you have further comments on this?

leafpetersen (approved):
LGTM, thanks!

Issue #1035, opened 2020-06-17
Inference cannot solve for `T?` == `dynamic`


lrhn:
Example:
```dart
class C<T extends Object> {
  List<T?> foo;
  C(this.foo);
}
main() {
  C(<dynamic>[]);
}
```
This fails in `nullsafety.dartpad.dev` with the analyzer errors:
```
error
'dynamic' doesn't extend 'Object' - line 7
error
Couldn't infer type parameter 'T'. Tried to infer 'dynamic' for 'T' which doesn't work: Type parameter 'T' declared to extend 'Object'. The type 'dynamic' was inferred from: Parameter 'foo' declared as 'List<T?>' but argument is 'List<dynamic>'. Consider passing explicit type argument(s) to the generic. - line 7
```

Should we infer `Object` for `T` here? We do so for `<Object?>[]`, and in practice `dynamic` and `Object?` are equivalent types wrt. subtyping, so choosing `T` as `Object` would allow a `List<dynamic>` as constructor argument.

(I hit the issue with an extension not applying to a dynamic list, even though it applies to a `List<Object?>`).

leafpetersen:
This is a bug in the spec for constraint generation.  We're missing the left top rule.  Would you mind adding a language test for this if you get a change?

lrhn:
Will do.

leafpetersen:
Spec [landed](https://github.com/dart-lang/language/pull/957), tests are out, implementation issues filed.

leafpetersen:
Fix to the fix landed above: https://github.com/dart-lang/language/pull/1200 (ordering was incorrect).

leafpetersen:
Implementation issues for reference:
  - [Analyzer](https://github.com/dart-lang/sdk/issues/43162)
  - [CFE](https://github.com/dart-lang/sdk/issues/43163)

leafpetersen:
Re-opening, since this is having difficulty landing.

leafpetersen:
Ok, this seems to have landed and stuck.

leafpetersen:
The following example seems to produce incorrect results still:

```dart
void main() {
  Future<Object?> bar1 = foo1(null); // Infers Object
  Future<dynamic> bar2 = foo1(null); // Infers Object?
  Future<Object?> bar3 = foo2(null); // Infers Object
  Future<dynamic> bar4 = foo2(null); // Infers Object
}


Future<T?> foo1<T extends Object?>(T x) {
  print('T is $T');
[...]

PR #1200, opened 2020-09-02
Update left top rule order in constraint solving

leafpetersen:
Update to fix for https://github.com/dart-lang/language/issues/1035 .

### DIFF
--- a/resources/type-system/inference.md
+++ b/resources/type-system/inference.md
@@ -779,8 +779,7 @@ clauses need be tried.
 - If `Q` is a legacy type `Q0*` then the match holds under constraint set `C`:
   - If `P` is `dynamic` or `void` and `P` is a subtype match for `Q0` under
     constraint set `C`.
-  - Or if `P` is not `dynamic` or `void` and `P` is a subtype match for `Q0?`
-    under constraint set `C`.
+  - Or if `P` is a subtype match for `Q0?` under constraint set `C`.
 - If `Q` is `FutureOr<Q0>` the match holds under constraint set `C`:
   - If `P` is `FutureOr<P0>` and `P0` is a subtype match for `Q0` under
     constraint set `C`.
--- a/resources/type-system/inference.md
+++ b/resources/type-system/inference.md
@@ -787,15 +787,15 @@ clauses need be tried.
   - Or if `P` is a subtype match for `Future<Q0>` under **empty** constraint set
     `C`
 - If `Q` is `Q0?` the match holds under constraint set `C`:
-  - If `P` is `dynamic` or `void` and `Object` is a subtype match for `Q0` under
-    constraint set `C`.
   - If `P` is `P0?` and `P0` is a subtype match for `Q0` under
     constraint set `C`.
+  - Or if `P` is `dynamic` or `void` and `Object` is a subtype match for `Q0`
+    under constraint set `C`.
   - Or if `P` is a subtype match for `Q0` under **non-empty** constraint set
-    `C`
-  - Or if `P` is a subtype match for `Null` under constraint set `C`
+    `C`.
+  - Or if `P` is a subtype match for `Null` under constraint set `C`.
   - Or if `P` is a subtype match for `Q0` under **empty** constraint set
-    `C`
+    `C`.
 - If `P` is `FutureOr<P0>` the match holds under constraint set `C1 + C2`:
   - If `Future<P0>` is a subtype match for `Q` under constraint set `C1`
   - And if `P0` is a subtype match for `Q` under constraint set `C2`
--- a/resources/type-system/inference.md
+++ b/resources/type-system/inference.md
@@ -777,9 +777,9 @@ clauses need be tried.
 - If `P` is a legacy type `P0*` then the match holds under constraint set `C`:
   - Only if `P0` is a subtype match for `Q` under constraint set `C`.
 - If `Q` is a legacy type `Q0*` then the match holds under constraint set `C`:
-  - If `P` is `dynamic` or `void` and `P` is a subtype match for `Q` under
+  - If `P` is `dynamic` or `void` and `P` is a subtype match for `Q0` under
     constraint set `C`.
-  - Or if `P` is not `dynamic` or `void` and `P` is a subtype match for `Q?`
+  - Or if `P` is not `dynamic` or `void` and `P` is a subtype match for `Q0?`
     under constraint set `C`.
 - If `Q` is `FutureOr<Q0>` the match holds under constraint set `C`:
   - If `P` is `FutureOr<P0>` and `P0` is a subtype match for `Q0` under
--- a/resources/type-system/inference.md
+++ b/resources/type-system/inference.md
@@ -777,7 +777,10 @@ clauses need be tried.
 - If `P` is a legacy type `P0*` then the match holds under constraint set `C`:
   - Only if `P0` is a subtype match for `Q` under constraint set `C`.
 - If `Q` is a legacy type `Q0*` then the match holds under constraint set `C`:
-  - Only if `P` is a subtype match for `Q?` under constraint set `C`.
+  - If `P` is `dynamic` or `void` and `P` is a subtype match for `Q` under
+    constraint set `C`.
+  - Or if `P` is not `dynamic` or `void` and `P` is a subtype match for `Q?`
+    under constraint set `C`.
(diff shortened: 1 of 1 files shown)


leafpetersen:
cc @scheglov @stefantsov

lrhn (approved):

leafpetersen:
@stefantsov Good catch.   I worked through the rules, here's another take.  I've moved the left top rule into the right nullability rule.  The reasoning is that:
  - The right FutureOr rule might apply (this is your counter-example), in which case we want to match using that rule.  `dynamic` is a subtype match for `FutureOr<Q0>` iff `dynamic` is a subtype match for `Q0`.
  - The right nullability rule might apply, and this is where we actually need the left top rule, so I've added a clause there.  If `P` is `dynamic`, then the rule can match in one of two ways: either by matching `Object` against `Q0`, or by matching `dynamic` against `Q0`.  we prefer the former (heuristically), and fall back on the latter.
  
PTAL, thanks.

cc @scheglov

scheglov:
I implemented the latest changes in the analyzer, but it affects inference for legacy libraries.
```dart
T f<T>(T a) => a;

test(dynamic a) {
  f(a);
}
```

We infer `Object*` for `T`.
```
[trySubtypeMatch][P: dynamic][Q: T*]
[trySubtypeMatch][P: dynamic][Q: T?]
[trySubtypeMatch][P: Object][Q: T]
[Object*]
```
I uploaded [work in progress CL](https://dart-review.googlesource.com/c/sdk/+/162621) to run it on bots, and see how many failures it has.

stefantsov:
> I implemented the latest changes in the analyzer, but it affects inference for legacy libraries.
> 

FWIW, there are two separate implementations of the algorithm in the CFE — one for the legacy mode and one for the Null Safe mode — so the CFE isn't affected in the same way.

stefantsov:
I filed an issue to update co19 tests according to the spec proposal: https://github.com/dart-lang/co19/issues/936.

leafpetersen:
@scheglov That will almost certainly be true breaking.  Update with an additional clause to preserve the legacy behavior.  PTAL.

resources/type-system/inference.md
```diff
@@ -774,7 +777,10 @@ clauses need be tried.
 - If `P` is a legacy type `P0*` then the match holds under constraint set `C`:
   - Only if `P0` is a subtype match for `Q` under constraint set `C`.
 - If `Q` is a legacy type `Q0*` then the match holds under constraint set `C`:
-  - Only if `P` is a subtype match for `Q?` under constraint set `C`.
+  - If `P` is `dynamic` or `void` and `P` is a subtype match for `Q` under
+    constraint set `C`.
```
stefantsov, line 781: It seems that this rule makes the algorithm recurse infinitely.

leafpetersen: Sorry, there was a typo in the original rule that I carried over to the new rule.  Fixed in both rules now.

scheglov:
I landed the implementation into the analyzer, but we had to revert it because of breaking google3.

The case how it breaks is following.
Before:
```
[trySubtypeMatch][P: Actor* Function(Map<String*, Object*>*)*][Q: R* Function(Map<String*, dynamic>*)*]
[trySubtypeMatch][P: Actor* Function(Map<String*, Object*>*)][Q: R* Function(Map<String*, dynamic>*)*]
[trySubtypeMatch][P: Actor* Function(Map<String*, Object*>*)][Q: R* Function(Map<String*, dynamic>*)?]
[trySubtypeMatch][P: Actor* Function(Map<String*, Object*>*)][Q: R* Function(Map<String*, dynamic>*)]
[trySubtypeMatch][P: Actor*][Q: R*]
[trySubtypeMatch][P: Actor][Q: R*]
[trySubtypeMatch][P: Actor][Q: R?]
[trySubtypeMatch][P: Actor][Q: R]
[trySubtypeMatch][P: Map<String*, dynamic>*][Q: Map<String*, Object*>*]
[trySubtypeMatch][P: Map<String*, dynamic>][Q: Map<String*, Object*>*]
[trySubtypeMatch][P: Map<String*, dynamic>][Q: Map<String*, Object*>?]
[trySubtypeMatch][P: Map<String*, dynamic>][Q: Map<String*, Object*>]
[trySubtypeMatch][P: String*][Q: String*]
[trySubtypeMatch][P: dynamic][Q: Object*]
[trySubtypeMatch][P: dynamic][Q: Object?]
[trySubtypeMatch][P: dynamic][Q: Object]
[trySubtypeMatch][P: dynamic][Q: Null]
[trySubtypeMatch][P: Map<String*, dynamic>][Q: Null]
[inferrer][t1: Actor* Function(Map<String*, Object*>*)*][t2: R* Function(Map<String*, dynamic>*)*][success: true]
```

After:
```
[trySubtypeMatch][P: Actor* Function(Map<String*, Object*>*)*][Q: R* Function(Map<String*, dynamic>*)*]
[trySubtypeMatch][P: Actor* Function(Map<String*, Object*>*)][Q: R* Function(Map<String*, dynamic>*)*]
[trySubtypeMatch][P: Actor* Function(Map<String*, Object*>*)][Q: R* Function(Map<String*, dynamic>*)?]
[trySubtypeMatch][P: Actor* Function(Map<String*, Object*>*)][Q: R* Function(Map<String*, dynamic>*)]
[trySubtypeMatch][P: Actor*][Q: R*]
[trySubtypeMatch][P: Actor][Q: R*]
[trySubtypeMatch][P: Actor][Q: R?]
[trySubtypeMatch][P: Actor][Q: R]
[trySubtypeMatch][P: Map<String*, dynamic>*][Q: Map<String*, Object*>*]
[trySubtypeMatch][P: Map<String*, dynamic>][Q: Map<String*, Object*>*]
[trySubtypeMatch][P: Map<String*, dynamic>][Q: Map<String*, Object*>?]
[trySubtypeMatch][P: Map<String*, dynamic>][Q: Map<String*, Object*>]
[trySubtypeMatch][P: String*][Q: String*]
[trySubtypeMatch][P: dynamic][Q: Object*]
[trySubtypeMatch][P: dynamic][Q: Object]
[trySubtypeMatch][P: Map<String*, dynamic>][Q: Null]
[trySubtypeMatch][P: Actor* Function(Map<String*, Object*>*)][Q: Null]
[inferrer][t1: Actor* Function(Map<String*, Object*>*)*][t2: R* Function(Map<String*, dynamic>*)*][success: false]
```

So, we apply "If `P` is `dynamic` or `void` and `P` is a subtype match for `Q0` under constraint set `C`." and `dynamic` is not a subtype match for `Object`, so we fail.

leafpetersen:
@scheglov I don't understand how the first trace is succeeding, can you elaborate?  Specifically, the diff between the two traces is that the original trace tries:

[trySubtypeMatch][P: dynamic][Q: Object*]
[trySubtypeMatch][P: dynamic][Q: Object?]
[trySubtypeMatch][P: dynamic][Q: Object]
[trySubtypeMatch][P: dynamic][Q: Null]

which looks like a failing trace to me: it tries to match `dynamic <# Object` and fails, then tries `dynamic <: Null` and fails.

The other trace looks like the following:

[trySubtypeMatch][P: String*][Q: String*]
[trySubtypeMatch][P: dynamic][Q: Object*]
[trySubtypeMatch][P: dynamic][Q: Object]

It's different, but gets the same answer: it just doesn't try to the `Null` branch that the other does.  This matches my intuition behind why I thought my fix was correct: asking `dynamic <: Q0?` can only succeed by matching `dynamic <: Q0`.

What am I missing?

scheglov:
When we match `dynamic <# Object*` we now replace it with `dynamic <# Object` which fails.

But before this change we replace it with `dynamic <# Object?` which first tries `Q?` rules (and fails, while doing this fails `dynamic <# Object`, also then fails `dynamic <# Null`), but then we fall through from `Q?` and apply "If `Q` is `dynamic`, `Object?`, or `void` then the match holds under no no constraints", with succeeds.

leafpetersen:
@scheglov I see, thanks.  I want to think this through again, but I'll follow up again shortly.

leafpetersen:
OK, I've removed the restriction that I'd added on the right legacy rule, it was just there as an (incorrect) optimization.  PTAL.

scheglov:
Thank you. The internal presubmit was green, so I landed [it](https://dart-review.googlesource.com/c/sdk/+/164240).

PR #1214, opened 2020-09-10
Adjust promotion via type test to handle recursion

eernstg:
The rules on promotion via type test have a funny anomaly: With a null test like `x != null`, nullability can be eliminated in nested constructs like `X extends (Y extends int?)`, but with a type test like `x is int`, the (otherwise very similar) approach stops without applying recursion.

This PR changes the approach to enable the recursive step.

A simple example where it makes a difference is

```dart
void f<X extends Y, Y extends int?>(X x) {
  if (x != null) x.isEven; // OK.
  if (x is int) x.isEven; // Error.
}
```

It seems reasonable to include the recursive step because it allows more complex situations to be handled in a way that is consistent and sound, even though it may be a situation that doesn't arise very frequently.

### DIFF
--- a/resources/type-system/flow-analysis.md
+++ b/resources/type-system/flow-analysis.md
@@ -365,6 +365,16 @@ We say that the **current type** of a variable `x` in variable model `VM` is `S`
   - `VM = VariableModel(declared, promoted, tested, assigned, unassigned, captured)`
   - `promoted = S::l` or (`promoted = []` and `declared = S`)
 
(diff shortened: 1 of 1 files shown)


eernstg:
This change is likely to be non-breaking in practice, so we can choose to delay the implementation a bit.

stereotype441 (approved):
lgtm

leafpetersen:
Just to avoid confusion: we do not currently plan to put this change out for implementation.  Once higher priority implementation issues have been resolved, we may reconsider.

PR #1219, opened 2020-09-11
Specify flow analysis of for-each loops.

stereotype441:
Note: this CL adjusts the implementation to match: https://dart-review.googlesource.com/c/sdk/+/162480

### DIFF
--- a/resources/type-system/flow-analysis.md
+++ b/resources/type-system/flow-analysis.md
@@ -691,6 +691,15 @@ TODO: Add missing expressions, handle cascades and left-hand sides accurately
   - Let `before(E) = join(after(S), continue(N))`
   - Let `after(N) = join(false(E), break(S))`
 
+- **for each statement**: If `N` is a for statement of the form `for (T X in E)
+  S`, `for (var X in E) S`, or `for (X in E) S`, then:
+  - Let `before(E) = before(N)`
+  - Let `before(S) = conservativeJoin(after(E), assignedIn(N), capturedIn(N))`
+  - Let `after(N) = join(before(S), break(S))`
+
+  TODO(paulberry): this glosses over how we handle the implicit assignment to X.
+  See https://github.com/dart-lang/sdk/issues/42653.
+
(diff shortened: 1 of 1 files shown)


leafpetersen (approved):

resources/type-system/flow-analysis.md
```diff
@@ -691,6 +691,15 @@ TODO: Add missing expressions, handle cascades and left-hand sides accurately
...
   - Let `after(N) = join(false(E), break(S))`
 
+- **for each statement**: If `N` is a for statement of the form `for (T X in E)
+  S`, `for (var X in E) S`, or `for (X in E) S`, then:
+  - Let `before(E) = before(N)`
+  - Let `before(S) = conservativeJoin(after(E), assignedIn(N), capturedIn(N))`
+  - Let `after(N) = join(before(S), break(S))`
+
+  TODO(paulberry): this glosses over how we handle the implicit assignment to X.
+  See https://github.com/dart-lang/sdk/issues/42653.
```
eernstg, line 701: It may or may not be better --- in https://github.com/dart-lang/language/pull/1220 I suggested that this could be handled directly by a reference to the desugaring (in which case the implicit assignment becomes an explicit assignment).

stereotype441: Good point.  I suppose we could do a similar thing for `for (D; C; U) S` loops as well, since I believe they're equivalent to `{ D; while (C) { { S } U; } }`.  Personally, I have a minor preference for specifying it separately rather than relying on the desugaring, though, for two reasons:  (1) The implementations do flow analysis and type inference prior to desugaring, and it's nice for the spec and the implementations to match closely.  (2) for-loops appearing inside lists, sets, and maps can't be desugared in the same way.

eernstg: >  it's nice for the spec and the implementations to match closely.

Indeed. I did use `{ D; while (C) { { S } U; } }` as a starting point for thinking about the flow analysis of C-style for statements earlier today, so the approach with sugar and without sugar can probably coexist in a useful manner.

Issue #1203, opened 2020-09-03
Types of interest differ for different loops, is that useful?


eernstg:
Consider the following program snippet:

```dart
main() {
  bool b = true;
  Object i = 42;
  ... // Insert loop here!
  i = 24; // Promote iff `int` is a type of interest.
  i.isEven; // Error unless `i` was promoted.
}
```

The missing statement is one of the following loops:

```dart
while (b) { // Does not make `int` a type of interest for `i`.
  if (i is int) print("");
}

do { // Does make `int` a type of interest.
  if (i is int) print("");
} while (b);

for (int j in []) { // Does make `int` a type of interest.
  if (i is int) print("");
}
```

Promotion occurs when we insert the do-while loop or the for loop into the example, but not if we insert the while loop.

There is no soundness issue because types of interest never force an unsound promotion, they just prevent some sound ones where the type is not 'of interest'. So the control flow differences can't justify the different treatment (also, the do-while loop is guaranteed to reach the type test, but the for loop and the while loop are not, so that shouldn't single out the while loop anyway).

Perhaps we should adjust the treatment of the while loop such that types of interest emerging in its body are included? Alternatively we could change the treatment of for loops to exclude them. The [flow-analysis spec](https://github.com/dart-lang/language/blob/master/resources/type-system/flow-analysis.md) seems to omit the types of interest from the body of a while loop (and doesn't include for loops), so we may need to make an adjustment.

@stereotype441, WDYT?

munificent:
> Perhaps we should adjust the treatment of the while loop such that types of interest emerging in its body are included?

+1 from me, for what it's worth. :)

lrhn:
Agree. There is a show of interest in the code, whether it's executed or not.

Even if we only care about executed tests, I'd have expected that `for` and `while` work the same (body not guaranteed to be run) and `do`/`while` could then be different (body guaranteed to run). The current behavior feels like a bug, no matter what consistent behavior we'd want.

stereotype441:
Agreed, it's really surprising that different types of loops would interact with types of interest differently like this.  I'm not sure exactly what's going on.  I will look in more detail when I return from vacation on Wednesday.

stereotype441:
[...]

PR #1220, opened 2020-09-12
Flow analysis: specify for loops and fix types of interest for for/while loops

stereotype441:
Fixes https://github.com/dart-lang/language/issues/1203.  Corresponding implementation change is here: https://dart-review.googlesource.com/c/sdk/+/162483

### DIFF
--- a/resources/type-system/flow-analysis.md
+++ b/resources/type-system/flow-analysis.md
@@ -710,10 +710,10 @@ TODO: Add missing expressions, handle cascades and left-hand sides accurately
 - **for statement**: If `N` is a for statement of the form `for (D; C; U) S`,
   then:
   - Let `before(D) = before(N)`.
-  - Let `before(E1) = conservativeJoin(after(D), assignedIn(N), capturedIn(N))`.
+  - Let `before(C) = conservativeJoin(after(D), assignedIn(N), capturedIn(N))`.
   - Let `before(S) = split(true(C))`.
   - Let `before(U) = merge(after(S), continue(S))`.
-  - Let `after(N) = inheritTested(join(false(E), unsplit(break(S))), after(U))`.
+  - Let `after(N) = inheritTested(join(false(C), unsplit(break(S))), after(U))`.
 
 - **do while statement**: If `N` is a do while statement of the form `do S while
   (E)` then:
--- a/resources/type-system/flow-analysis.md
+++ b/resources/type-system/flow-analysis.md
@@ -356,6 +356,28 @@ We also make use of the following auxiliary functions:
       `VariableModel(d0, [], s0, a0, false, c0)`
     - Otherwise `VI1` maps `v` to `VM0`
 
+- `inheritTestedV(VM1, VM2)`, where `VM1` and `VM2` are variable models,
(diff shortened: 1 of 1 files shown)


resources/type-system/flow-analysis.md
```diff
@@ -683,7 +705,15 @@ TODO: Add missing expressions, handle cascades and left-hand sides accurately
   (E) S` then:
   - Let `before(E) = conservativeJoin(before(N), assignedIn(N), capturedIn(N))`.
   - Let `before(S) = split(true(E))`.
-  - Let `after(N) = join(false(E), unsplit(break(S))`
+  - Let `after(N) = inheritTested(join(false(E), unsplit(break(S))), after(S))`.
+
+- **for statement**: If `N` is a for statement of the form `for (D; C; U) S`,
+  then:
+  - Let `before(D) = before(N)`.
+  - Let `before(E1) = conservativeJoin(after(D), assignedIn(N), capturedIn(N))`.
```
eernstg, line 713: Wouldn't that be 'Let `before(C) = ...`'?

resources/type-system/flow-analysis.md
```diff
@@ -683,7 +705,15 @@ TODO: Add missing expressions, handle cascades and left-hand sides accurately
   (E) S` then:
   - Let `before(E) = conservativeJoin(before(N), assignedIn(N), capturedIn(N))`.
   - Let `before(S) = split(true(E))`.
-  - Let `after(N) = join(false(E), unsplit(break(S))`
+  - Let `after(N) = inheritTested(join(false(E), unsplit(break(S))), after(S))`.
+
+- **for statement**: If `N` is a for statement of the form `for (D; C; U) S`,
```
eernstg, line 710: Woohoo! ;-)

resources/type-system/flow-analysis.md
```diff
@@ -683,7 +705,15 @@ TODO: Add missing expressions, handle cascades and left-hand sides accurately
...
+  then:
+  - Let `before(D) = before(N)`.
+  - Let `before(E1) = conservativeJoin(after(D), assignedIn(N), capturedIn(N))`.
+  - Let `before(S) = split(true(C))`.
+  - Let `before(U) = merge(after(S), continue(S))`.
+  - Let `after(N) = inheritTested(join(false(E), unsplit(break(S))), after(U))`.
```
eernstg, line 716: Wouldn't `E` be `C`?

Also, can we now add a new bullet point along the lines of this?:

```
- **for-in statement**: If `N` is a for-in statement of the form
  `for (D in I) S` then the flow analysis is performed according
  to the desugaring of `N`. *The desugaring is specified in the
  language specification, and it yields a `while` loop.*
```

eernstg (approved):
LGTM, with the suggestion that we might add a bullet point about for-in statements, and pointing out what seems to be a couple of typos.

eernstg:
Removed the comment about for-in statements, it's being handled in #1219.

resources/type-system/flow-analysis.md

resources/type-system/flow-analysis.md
stereotype441: Fixed!

PR #1230, opened 2020-09-21
Add declaration-site variance feature specification

eernstg:
This PR adds a feature specification for declaration-site variance, which is a sound and explicit mechanism for specifying that a type parameter of a generic class can be covariant (marked by `out`), invariant (`inout`), or contravariant (`in`), cf. #524.


accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,415 @@
...
+_X_ to occur in a non-contravariant position in a member signature in the
+body of _D_, except that it is not an error if it occurs in a contravariant
+position in the type of a covariant formal parameter. *For instance, _X_
+can not be the return type of a method or getter, and it can not be the
+bound of a type parameter of a generic method.*
+
+*If _X_ has the variance modifier `inout` then there are no variance
+related restrictions on the positions where it can occur.*
+
+*For superinterfaces we need slightly stronger rules than the ones that
```
eernstg, line 218: I'll relax this tomorrow, such that it matches the implementation.

accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,373 @@
...
+### Variance Rules
+
+The rules for determining the variance of a position are updated as follows:
+
+We say that a type parameter of a generic class is _covariant_ if it has no
+variance modifier or it has the modifier `out`; we say that it is
+_contravariant_ if it has the modifier `in`; and we say that it is
+_invariant_ if it has the modifier `inout`.
+
+The covariant occurrences of a type (schema) `T` in another type (schema)
```
leafpetersen, line 105: This definition wigs me out a bit: I know what it's saying, but if I really get pedantic it's weird.  It's basically computing a set which will have cardinality zero or one.  That is, the base case says that `S` is a covariant occurrence of `T` (that is, they are same the type variable).  Presumably we're meant to interpret this as a singleton set.  But the only thing ever put in such as set is... `T`.  So all of the unions are either of empty sets, or of the same singleton set.  So all this is computing is "yes this variable occurs" or "no it doesn't".  So I think it would be better to phrase these as predicates: "We say that a type (schema) `T` occurs covariantly in another type (schema) `S` if: ...."

accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,373 @@
...
+`S` are:
+
+  - if `S` and `T` are the same type,
+    - `S` is a covariant occurrence of `T`.
+  - if `S` is `Future<U>`
```
leafpetersen, line 110: Add `S is U?` ?

accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,373 @@
...
+    - the covariant occurrences of `T` in `U`
+  - if `S` is `FutureOr<U>`
+    - the covariant occurrencs of `T` in `U`
+  - if `S` is an interface type `C<T0, ..., Tk>`
```
leafpetersen, line 114: I think this is ok, but it feels slightly off to me that we don't consider an invariant occurrence to be both a covariant occurrence and an invariant occurrence.  I think it works out with the way the definitions are used here?  But it's worth double checking

accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,373 @@
...
+    - all occurrences of `T` in `Bi` for `i` in `0, ..., k`
+  - if `S` is `U Function<X0 extends B0, ...., Xk extends Bk>(T0 x0, ...., Tn xn, {Tn+1 xn+1, ..., Tm xm})`
+      the union of:
+    - the invariant occurrences of `T` in `U`
+    - the invariant occurrences of `T` in `Ti` for `i` in `0, ..., m`
+    - all occurrences of `T` in `Bi` for `i` in `0, ..., k`
+
+It is a compile-time error if a variance modifier is specified for a type
+parameter declared by a static extension, a generic function type, a
+generic function or method, or a type alias.
```
leafpetersen, line 176: Presumably allowed for enums (post enhanced enums)?

accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,373 @@
...
+
+*Variance is not relevant to static extensions, because there is no notion
+of subsumption. Each usage will be a single call site, and the value of
+every type argument associated with an extension method invocation is
+statically known at the call site. Similar reasons apply for functions and
+function types. Finally, the variance of a type parameter declared by a
+type alias is determined by the usage of that type parameter in the body of
```
leafpetersen, line 183: Do these meaningfully have variance?

accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,373 @@
...
+the type alias.*
+
+We say that a type parameter _X_ of a type alias _F_ _is covariant_ if it
+only occurs covariantly in the body of _F_; that it _is contravariant_ if
+it occurs contravariantly in the body of _F_ and does not occur covariantly
+or invariantly; that it _is invariant_ if it occurs invariantly in the body
+of _F_ (*with no constraints on other occurrences*), or if it occurs both
+covariantly and contravariantly.
+
+*In particular, an unused type parameter is considered covariant.*
```
leafpetersen, line 193: What does this mean?  Usually I think of variance as specifying the subtyping relation.  But given `typedef F<X> = int`, I expect that `F<num> <: F<double>` and `F<double> <: F<num>`, so in what sense is `X` covariant?

accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,373 @@
...
+class B<out U, inout V, in W> implements
+    A<U Function(W), V Function(V), W Function(V)> {}
+
+// B<int, String, num> <: B<num, String, int>, and hence
+// A<int Function(num), String Function(String), num Function(String)> <:
+// A<num Function(int), String Function(String), int Function(String)>.
+```
+
+*In a superinterface, a type parameter without a variance modifier can be
+used in an actual type argument for a parameter with a variance modifier,
```
leafpetersen, line 243: Worth making it explicit that this is subject to the checks above, which de facto means that the vice versa only allows the "explicitly covariant or invariant to implicitly covariant" cases?  That is, `class C<in X> implements List<X>` is a no-no, right?

And don't we have the "twisted super-hierarchy" problem to deal with here too, so we have to outlaw "class C<X> implements Contra<X>, Invariant<X>"?

accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,373 @@
...
+covariantly and contravariantly.
+
+*In particular, an unused type parameter is considered covariant.*
+
+Let _D_ be the declaration of a class or mixin, and let _X_ be a type
+parameter declared by _D_.
+
+If _X_ has the variance modifier `out` then it is a compile-time error for
+_X_ to occur in a non-covariant position in a member signature in the body
+of _D_, except that it is not an error if it occurs in a covariant position
```
leafpetersen, line 200: Only covariant?  Or also invariant?

accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,373 @@
...
+in the type annotation of a covariant formal parameter (*this is a
+contravariant position in the member signature as a whole*).
+
+*In particular, _X_ can not be the type of a method parameter (unless
+covariant), and it can not be the bound of a type parameter of a generic
+method.*
+
+If _X_ has the variance modifier `in` then it is a compile-time error for
+_X_ to occur in a non-contravariant position in a member signature in the
+body of _D_, except that it is not an error if it occurs in a contravariant
```
leafpetersen, line 210: or invariant?

accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,373 @@
...
+
+*In other words, source code in _L2_ is seen as having variance modifiers
+available, but it is simply not using them.*
+
+
+### Soundly variant libraries seen from a legacy library
+
+When a legacy library _L_ imports a library _L2_ with sound variance, the
+declarations imported from _L2_ are _legacy erased_. This means that all
+variance modifiers in type parameter declarations are ignored.
```
leafpetersen, line 365: Does this work out?   Doesn't this mean I can recreate your "twisted hierarchy" soundness bug by viewing a contravariant class from a legacy library?

accepted/future-releases/variance/feature-specification.md
leafpetersen: > And don't we have the "twisted super-hierarchy" problem to deal with here too, so we have to outlaw "class C implements Contra, Invariant"?

On further thought, this is I think disallowed by our existing super-interface check, but probably worth pointing that out here.

accepted/future-releases/variance/feature-specification.md
eernstg: Did that.

accepted/future-releases/variance/feature-specification.md
eernstg: I think it's working (it's the same approach as in the language specification, and I think that's in a good shape).

accepted/future-releases/variance/feature-specification.md
eernstg: I made it more explicit that the notion of the variance of a type parameter of a type alias is unchanged.

accepted/future-releases/variance/feature-specification.md
eernstg: Thinking about this again, I noticed that we don't have any equivalence of this kind. The unused type variable is handled in the few situations where that is needed, because a rule applies to, e.g., any non-contravariant type variable, and that includes both unused, invariant, and covariant ones.

accepted/future-releases/variance/feature-specification.md
eernstg: As soon as we have a dynamic type check on the actual argument it is of course sound with respect to the execution of the function body, so we can just allow the declared type to be anything the developer wants, no limits.

However, we have maintained a certain level of discipline so far: A covariant parameter should have a declared type in the dynamic type of the receiver which is a subtype or a supertype of the statically known declared type. With invariant positions I'd expect that we allow unrelated types.

I did think about this, but I haven't changed anything yet.

accepted/future-releases/variance/feature-specification.md
eernstg: Same as line 200.

accepted/future-releases/variance/feature-specification.md
eernstg: I added a separate section about the dynamic checks.

accepted/future-releases/variance/feature-specification.md
eernstg: I changed this to say that the legacy library simply gets the declaration-site based subtype relationships whenever it refers to types that involve declarations using declaration-site variance. Surely that's not only simpler, it is also much more useful.

I haven't written anything about the management of breakage associated with corelib changes.

eernstg (commented):
Review response.

kallentu:
@eernstg What's the status of this PR? Could we merge this into the language repo?

eernstg:
@kallentu wrote:
> What's the status of this PR? Could we merge this into the language repo?

I haven't worked on this PR recently, but it could be useful to make sure it is up to date and then use that during the subsequent finalization of the feature.

@leafpetersen, WDYT? Do you want to start afresh on a specification of declaration-site variance, or do we proceed to do it based on this PR?

accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,482 @@
...
+```
+<typeParameter> ::= // Modified rule.
+    <metadata> <typeParameterVariance>? <typeIdentifier>
+    ('extends' <typeNotVoid>)?
+
+<typeParameterVariance> ::= // New rule.
+    'out' | 'inout' | 'in'
+```
+
+`out` and `inout` are added to the set of built-in identifiers (* and `in`
```
lrhn, line 67: Is that needed, or could they just be contextual?
Are we afraid that `<out out>` or `<out>` will be hard to parse?

accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,482 @@
...
+
+It is a compile-time error if a variance modifier is specified for a type
+parameter declared by a static extension, a generic function type, a
+generic function or method, or a type alias.
+
+*Variance is not relevant to static extensions, because there is no notion
+of subsumption. Each usage will be a single call site, and the value of
+every type argument associated with an extension method invocation is
+statically known at the call site. Similar reasons apply for functions and
+function types. Finally, the variance of a type parameter declared by a
```
lrhn, line 175: I'd expect extension types to support variance too. They are interface types.

I'm a little worried that you don't need variance for extensions and function types.
That probably means I don't understand the variance feature, which I don't, and have no intuition about it either, which is worrisome.

Function types already have variance implicit in their type variable occurrences, they don't need extra variance markers.

For extensions, I guess all the heavy lifting is part of the "is this interface type a subtype of the `on` type" check. If that's satisfied, then whatever it does to `this` will be safe. Also, it doesn't introduce a type.

Extension types may have something similar for their representation type and `implements` clauses for that, but they'll need variance for their own methods. And because their representation value may use variance.

```dart
extension type ConvertingBuffer<in S, T out>(({Queue<T> buffer, Transformer<S, T> transform}) _) {
   void add(S value) { _.buffer.add(_.transform(value)); }
   T removeNext() => _.buffer.removeFirst();
}

void main() {
  var c = ConvertingBuffer<String, num>((buffer: Queue(), transform: Transformer<num, String>.from((v) => "$v"));
  c.add(1);
  c.add(3.14);
  var d = c as ConvertingBuffer<Object, int>; // VALID IF VARIANCE WORKS.
  d.add(42);
}
```

I want tests. Lots of tests.

accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,482 @@
...
+      and `S` occurs in a contravariant position in `Tj`.
+  - `T` is `FutureOr<U>` and `S` occurs in a covariant position in `U`.
+  - `T` is `U Function<X1 extends B1, ...., Xk extends Bk>(T1 x1, ...., Tn xn, [Tn+1 xn+1, ..., Tm xm])` and:
+    - `S` occurs in a covariant position in `U`, or
+    - `S` occurs in a contravariant position in `Tj` for some `j` in `1..m`.
+  - `T` is `U Function<X1 extends B1, ...., Xk extends Bk>(T1 x1, ...., Tn xn, {Tn+1 xn+1, ..., Tm xm})` and:
+    - `S` occurs in a covariant position in `U`, or
+    - `S` occurs in a contravariant position in `Tj` for some `j` in `1..m`.
+  - `T` is `U?` and `S` occurs in a covariant position in `U`.
+  - `T` is `X &amp; U` and `S` occurs in a covariant position in `U`.
```
lrhn, line 132: (Don't use `&amp;` inside a code quote, just `&`.)

lrhn, line 132: Or `T` is `X`? (Which _is_ a type variable in this case!)

accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,482 @@
...
+
+We say that a type `S` occurs in a covariant position in a type `T` if at
+least one of the following conditions is true:
+  - `T` is `S`.
+  - `T` is an interface type `C<T1, ..., Tk>`, `j` is in `1..k`, and:
+    - the `j`th type parameter of `C` is covariant
+      and `S` occurs in a covariant position in `Tj`, or
+    - the `j`th type parameter of `C` is contravariant
+      and `S` occurs in a contravariant position in `Tj`.
+  - `T` is `FutureOr<U>` and `S` occurs in a covariant position in `U`.
```
lrhn, line 124: Or `T` is `Future<U>`.

(Or is the "type `S`" _always_ going to be a type parameter, not a general type?)

accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,482 @@
...
+  - `T` is `U Function<X1 extends B1, ...., Xk extends Bk>(T1 x1, ...., Tn xn, [Tn+1 xn+1, ..., Tm xm])` and:
+    - `S` occurs in a covariant position in `U`, or
+    - `S` occurs in a contravariant position in `Tj` for some `j` in `1..m`.
+  - `T` is `U Function<X1 extends B1, ...., Xk extends Bk>(T1 x1, ...., Tn xn, {Tn+1 xn+1, ..., Tm xm})` and:
+    - `S` occurs in a covariant position in `U`, or
+    - `S` occurs in a contravariant position in `Tj` for some `j` in `1..m`.
+  - `T` is `U?` and `S` occurs in a covariant position in `U`.
```
lrhn, line 131: Or `T` is `Null`, if we see `U?` as `U | Null`?

accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,482 @@
...
+    - `S` occurs in `Bj` (*at any position*) for some `j` in `1..k`.
+  - `T` is `U?` and `S` occurs in an invariant position in `U`.
+  - `T` is `X &amp; U` and `S` occurs in an invariant position in `U`.
+
+It is a compile-time error if a variance modifier is specified for a type
+parameter declared by a static extension, a generic function type, a
+generic function or method, or a type alias.
+
+*Variance is not relevant to static extensions, because there is no notion
+of subsumption. Each usage will be a single call site, and the value of
```
lrhn, line 172: Maybe lead with this. A clear description of what this feature is, and what it _is not_ (and why),
to set the context for the reader.
All the time until I got to here, I was questioning why there was no mention of generic functions.

accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,482 @@
...
+every type argument associated with an extension method invocation is
+statically known at the call site. Similar reasons apply for functions and
+function types. Finally, the variance of a type parameter declared by a
+type alias is determined by the usage of that type parameter in the body of
+the type alias.*
+
+*Variance _can_ be used with a generic enum declaration.*
```
lrhn, line 179: First line of section say "type parameter of a generic class, mixin, or enum", so not sure why _that_ particular case needs repeating.

accepted/future-releases/variance/feature-specification.md
```diff
@@ -0,0 +1,482 @@
...
+body of _D_, except that it is not an error if it occurs in a contravariant
+position in the type of a formal parameter which is covariant. *For
+instance, _X_ can never be the return type of a method or getter, and it can
+never be the bound of a type parameter of a generic method.*
+
+*If _X_ has the variance modifier `inout` then there are no variance
+related restrictions on the positions where it can occur in member
+signatures.*
+
+Let _D_ be a class or mixin declaration, let _S_ be a direct superinterface
```
lrhn, line 215: or enum.

lrhn:
I think this specification can probably be used as the basis.

My biggest worry is how to migrate platform libraries.

If we can't make `Future`, `Stream` and `Iterable` be covariant, we have a problem.
(So we have a problem in `reduce`. Can we deprecate and remove it, and/or replace it with an extension method? Or introduce statically typed instance members, aka extension members declared on the type iteself?)

We need some migration path that allows adding restrictions to platform libraries.
(I'd love if it's not just another "platform library exception" that makes old libraries not see the new types, that means we can't get any benefits from the change.)

working/0524-variance/feature-specification.md
```diff
@@ -0,0 +1,373 @@
...
+covariantly and contravariantly.
+
+*In particular, an unused type parameter is considered covariant.*
+
+Let _D_ be the declaration of a class or mixin, and let _X_ be a type
+parameter declared by _D_.
+
+If _X_ has the variance modifier `out` then it is a compile-time error for
+_X_ to occur in a non-covariant position in a member signature in the body
+of _D_, except that it is not an error if it occurs in a covariant position
```
eernstg: This has been clarified to indicate that the non-error applies when the parameter is covariant by declaration. In this situation every actual argument will be subjected to a run-time type check, ensuring that no soundness violation can proceed. There is no corresponding notion of being "invariant by declaration".

working/0524-variance/feature-specification.md
```diff
@@ -0,0 +1,482 @@
...
+```
+<typeParameter> ::= // Modified rule.
+    <metadata> <typeParameterVariance>? <typeIdentifier>
+    ('extends' <typeNotVoid>)?
+
+<typeParameterVariance> ::= // New rule.
+    'out' | 'inout' | 'in'
+```
+
+`out` and `inout` are added to the set of built-in identifiers (* and `in`
```
eernstg: It is not required that `out` and `inout` are built-in identifiers. I updated the text to say that they are not built-in identifiers.

working/0524-variance/feature-specification.md
```diff
@@ -0,0 +1,482 @@
...
+      and `S` occurs in a contravariant position in `Tj`.
+  - `T` is `FutureOr<U>` and `S` occurs in a covariant position in `U`.
+  - `T` is `U Function<X1 extends B1, ...., Xk extends Bk>(T1 x1, ...., Tn xn, [Tn+1 xn+1, ..., Tm xm])` and:
+    - `S` occurs in a covariant position in `U`, or
+    - `S` occurs in a contravariant position in `Tj` for some `j` in `1..m`.
+  - `T` is `U Function<X1 extends B1, ...., Xk extends Bk>(T1 x1, ...., Tn xn, {Tn+1 xn+1, ..., Tm xm})` and:
+    - `S` occurs in a covariant position in `U`, or
+    - `S` occurs in a contravariant position in `Tj` for some `j` in `1..m`.
+  - `T` is `U?` and `S` occurs in a covariant position in `U`.
+  - `T` is `X &amp; U` and `S` occurs in a covariant position in `U`.
```
eernstg: Done.

eernstg: We don't need a case to handle `T` is `X` because type variables do not contain subterms where a type can occur. If `S` is a type variable then we can use the rule about `T` is `S`.

working/0524-variance/feature-specification.md
```diff
@@ -0,0 +1,482 @@
...
+
+We say that a type `S` occurs in a covariant position in a type `T` if at
+least one of the following conditions is true:
+  - `T` is `S`.
+  - `T` is an interface type `C<T1, ..., Tk>`, `j` is in `1..k`, and:
+    - the `j`th type parameter of `C` is covariant
+      and `S` occurs in a covariant position in `Tj`, or
+    - the `j`th type parameter of `C` is contravariant
+      and `S` occurs in a contravariant position in `Tj`.
+  - `T` is `FutureOr<U>` and `S` occurs in a covariant position in `U`.
```
eernstg: `Future<U>` is an interface type, which is covered by item 2 in this definition.

The type `S` is indeed going to be a type parameter whenever we are using these notions of "`S` occurs in a something-variant position" to determine whether a given signature is a compile-time error, or whether it gives rise to some specific run-time type checks.

However, I don't think it creates any issues that `S` is a general type, and the tiny generalization may be useful, e.g., when we're considering situations involving substitutions.

working/0524-variance/feature-specification.md
```diff
@@ -0,0 +1,482 @@
...
+  - `T` is `U Function<X1 extends B1, ...., Xk extends Bk>(T1 x1, ...., Tn xn, [Tn+1 xn+1, ..., Tm xm])` and:
+    - `S` occurs in a covariant position in `U`, or
+    - `S` occurs in a contravariant position in `Tj` for some `j` in `1..m`.
+  - `T` is `U Function<X1 extends B1, ...., Xk extends Bk>(T1 x1, ...., Tn xn, {Tn+1 xn+1, ..., Tm xm})` and:
+    - `S` occurs in a covariant position in `U`, or
+    - `S` occurs in a contravariant position in `Tj` for some `j` in `1..m`.
+  - `T` is `U?` and `S` occurs in a covariant position in `U`.
```
eernstg: We already know by rule 1 that `Null` occurs covariantly in `Null`, and `Null` doesn't contain subterms which may be in a ...variant position of any kind. I don't think we need to add anything here.

working/0524-variance/feature-specification.md
```diff
@@ -0,0 +1,482 @@
...
+
+*Variance is not relevant to static extensions, because there is no notion
+of subsumption. Each usage will be a single call site, and the value of
+every type argument associated with an extension method invocation is
+statically known at the call site. Similar reasons apply for functions and
+function types. Finally, the variance of a type parameter declared by a
+type alias is determined by the usage of that type parameter in the body of
+the type alias.*
+
+*Variance _can_ be used with a generic enum declaration.*
```
eernstg: Right, removed it.

working/0524-variance/feature-specification.md
```diff
@@ -0,0 +1,482 @@
...
+body of _D_, except that it is not an error if it occurs in a contravariant
+position in the type of a formal parameter which is covariant. *For
+instance, _X_ can never be the return type of a method or getter, and it can
+never be the bound of a type parameter of a generic method.*
+
+*If _X_ has the variance modifier `inout` then there are no variance
+related restrictions on the positions where it can occur in member
+signatures.*
+
+Let _D_ be a class or mixin declaration, let _S_ be a direct superinterface
```
eernstg: Yep, done.

working/0524-variance/feature-specification.md
```diff
@@ -0,0 +1,482 @@
...
+    - `S` occurs in `Bj` (*at any position*) for some `j` in `1..k`.
+  - `T` is `U?` and `S` occurs in an invariant position in `U`.
+  - `T` is `X &amp; U` and `S` occurs in an invariant position in `U`.
+
+It is a compile-time error if a variance modifier is specified for a type
+parameter declared by a static extension, a generic function type, a
+generic function or method, or a type alias.
+
+*Variance is not relevant to static extensions, because there is no notion
+of subsumption. Each usage will be a single call site, and the value of
```
eernstg: Certainly, done.

working/0524-variance/feature-specification.md
```diff
@@ -0,0 +1,482 @@
...
+every type argument associated with an extension method invocation is
+statically known at the call site. Similar reasons apply for functions and
+function types. Finally, the variance of a type parameter declared by a
```
eernstg: Included extension types. I agree that it should work. Also, we need lots of tests. ;-)

The fundamental soundness requirement is that the extension type and the representation type must co-vary, that is, if `E1` and `E2` are generic instantiations of the extension type `E` such that `E1 <: E2` then it must also be true that `R1 <: R2`, where `R1` and `R2` are the instantiated representation types corresponding to `E1` and `E2`, respectively.

eernstg:
We're moving on this one! :tada:

I'll land this document now. Leaf and I talked about a couple of questions in need of further discussion, and they're on the language team agenda for the next meeting. Further discussions and revisions will occur as needed.

working/0524-variance/feature-specification.md
```diff
@@ -0,0 +1,510 @@
...
+
+The interface compositionality rule in [subtyping.md] is updated
+as follows:
+
+[subtyping.md]: https://github.com/dart-lang/language/blob/master/resources/type-system/subtyping.md
+
+- **Interface Compositionality**: `T0` is an interface type or an extension
+  type `C0<S0, ..., Sk>` and `T1` is `C0<U0, ..., Uk>`. For `i` in `0..k`,
+  let `vi` be the declared variance of the `i`th type parameter of
+  `C0`. Then, for each `i` in `0..k`, one of the following holds:
```
leif-zuttermeister, line 95: Should this sentence be "Then, `T0 <: T1` if and only if for each ..."? I feel like it's missing what these conditions define.

working/0524-variance/feature-specification.md
```diff
@@ -0,0 +1,510 @@
...
+  - `T` is `U Function<X1 extends B1, ...., Xk extends Bk>(T1 x1, ...., Tn xn, {Tn+1 xn+1, ..., Tm xm})` and:
+    - `S` occurs in a invariant position in `U`, or
+    - `S` occurs in a invariant position in `Tj` for some `j` in `1..m`, or
+    - `S` occurs in `Bj` *(at any position)* for some `j` in `1..k`.
+  - `T` is `U?` and `S` occurs in an invariant position in `U`.
+  - `T` is `X & U` and `S` occurs in an invariant position in `U`.
+
+It is a compile-time error if a variance modifier is specified for a type
+parameter declared in the type parameter list of a declaration of an
+extension, a generic function or method, an extension type, or a type
```
leif-zuttermeister, line 184: At the top, it says
> The explicitly declared variance can be used with [...], that is, class, mixin class,
enum, **extension type**, and mixin declarations.

Here, it says that It's a compile time error with extension types.
Should they be removed from one of the lists?

working/0524-variance/feature-specification.md
eernstg: It could have been written more explicitly along the lines of your comment, but several other rules in the same definition rely on the context in the same manner. For example:

> **Record Type/Record**: `T0` is a record type and `T1` is `Record`

which could have been "`T0 <: T1` if `T0` is a record type and `T1` is `Record`".

Another matter is that we'd have to be careful about "if and only if": We definitely can't use it before the matching criteria (that is, we can't say "`T0 <: T1` if and only if `T0` is a record type and `T1` is `Record`"). The "if and only if" relationship may hold if we have already confirmed the matching properties (e.g. "`T0` is a record type"), because otherwise we might need to use a later rule rather than just concluding that the subtype relationship doesn't hold.

I think it's OK to keep the rule the way it is because it matches the style used in the context where the rule occurs.

working/0524-variance/feature-specification.md
eernstg: Good catch! Extension types should be removed from this list. Thanks!

eernstg (commented):
Thanks for the input, @leif-zuttermeister, extension types should indeed be removed from the list that you mentioned. I'll do that in the next PR on this topic.

PR #1231, opened 2020-09-21
Small fixes

leafpetersen:
Change inference with a Null initializer to not use a subtype rule, and eliminate some out of date wording.


leafpetersen:
Both implementations are consistent on inferring `dynamic` from a `Null` initializer and not from a `Never` initializer, so I see no real benefit to mandating a change here.

stereotype441 (approved):
lgtm

PR #1242, opened 2020-09-29
Do not use promoteToNonNull in flow analysis of method invocations.

stereotype441:
It's not necessary, because a method invocation is only valid if
either the target is already non-nullable, in which case it would have
no effect, or the method is available on a nullable type
(e.g. toString()), in which case we shouldn't be promoting the type of
the target anyhow.


leafpetersen (approved):

eernstg (commented):
LGTM.

PR #1248, opened 2020-10-05
Clarify that `main` cannot be a getter

eernstg:
@johnniwinther just noted that the wording about errors on `main` introduced by https://github.com/dart-lang/language/pull/1146 needs to be clarified: It is not the intention that `main` can be a getter (and the old text says so explicitly in commentary), but the word `function` in the spec refers to "functions" (with a `<formalParameterPart>`) as well as setters, getters, and constructors.

In the context we already know that `main` is a top-level declaration, so it can't be a constructor. It can't be a setter because a setter name ends in `=`. So we just need to specify that it can't be a getter. This PR does that.

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -6,6 +6,9 @@ Status: Draft
 
 ## CHANGELOG
 
+2020.10.05
+  - Clarify that `main` cannot be a getter.
+
 2020.09.10
   - Specify updates to super-bounded type rules for null safety.
 
@@ -1112,7 +1115,7 @@ The section 'Scripts' in the language specification is replaced by the
 following:
 
 Let _L_ be a library that exports a declaration _D_ named `main`.  It is a
-compile-time error unless _D_ is a function declaration.  It is a
+compile-time error unless _D_ is a non-getter function declaration.  It is a
 compile-time error if _D_ declares more than two required positional
 parameters, or if there are any required named parameters.  It is a
 compile-time error if _D_ declares at least one positional parameter, and


accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1112,7 +1115,7 @@ The section 'Scripts' in the language specification is replaced by the
 following:
 
 Let _L_ be a library that exports a declaration _D_ named `main`.  It is a
-compile-time error unless _D_ is a function declaration.  It is a
+compile-time error unless _D_ is a non-getter function declaration.  It is a
```
eernstg, line 1118: Though not strictly necessary, we could also say 'non-getter and non-setter function declaration', in order to outlaw the confusing permission to have `set main(_) {}`.

lrhn: This change isn't *technically* needed. A "function declaration" is a concept separate from a "getter declaration". I do believe the spec occasionally refers to getters as functions, though, so it's probably better to double down on the redundancy overhead.

lrhn (approved):

accepted/future-releases/nnbd/feature-specification.md
eernstg: Thanks!

We actually have the following in section `Function Declarations`: `A function declaration consists of an identifier indicating the function’s name, possibly prefaced by a return type. The function name is followed by a signature and body. For getters, the signature is empty.` So a function declaration can be a getter declaration, just like a function can be a getter.

But what do you think about outlawing the _setter_? It is confusing that it is allowed, but its name is of course `main=`, so there is no harm in allowing it.

eernstg: Let's take the setter separately, to get this one settled.

PR #1249, opened 2020-10-05
Add nnbd warning for cases like `C?.staticMethod()`

eernstg:
When `T` is a type literal denoting a class, mixin, or extension `C` (`T` could be `C` or `p.C`), `T.m()` will invoke the static method `m` declared by `C`. In that case `T?.m()` was previously equivalent to `T.m()` and caused no diagnostic messages. This PR changes it to be a warning.

This was discussed by mail in May-June ('ClassName?.foo diagnostics'), but apparently hasn't yet been specified.

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -6,6 +6,10 @@ Status: Draft
 
 ## CHANGELOG
 
+2020.10.05
+  - Specify that a null-aware static member access (e.g., `C?.staticMethod()`)
+    is a warning.
+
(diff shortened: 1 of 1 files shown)


leafpetersen (approved):
Thanks, can you file implementation issues two, but specifically not as blockers for beta.  Let's target beta 2.

eernstg:
Thanks! Implementation issues in https://github.com/dart-lang/sdk/issues/43701, https://github.com/dart-lang/sdk/issues/43702, https://github.com/dart-lang/sdk/issues/43703.

Issue #1250, opened 2020-10-05
Least upper bound involving FutureOr has a soundness problem


eernstg:
The function [**UP**](https://github.com/dart-lang/language/blob/master/resources/type-system/upper-lower-bounds.md#upper-bounds) does not have any cases covering `FutureOr`.

I believe this implies that such examples as `UP(FutureOr<int?>, num)` does not match any rules.

In particular, the last rule about the 'least upper bound of two interfaces as in Dart 1' would treat `FutureOr<int?>` as an interface type, but such types are assumed to be subtypes of `Object`, and that is not true for `FutureOr<int?>`. So we can't just pretend that `FutureOr` has `Object` as a superinterface and proceed with the Dart 1 algorithm, because it is unsound to let `UP(FutureOr<int?>, num) == Object`.

The implementations do indeed fail to handle this today, e.g., with the following example:

```dart
import 'dart:async';

bool condition = true;

void main() {
  FutureOr<int?> x = null;
  num n = 1;
  var z = condition ? x : n;
  // Unsound: `z` has the type `Object` and the value null.
  Object o = z;
  print(o); // 'null'.
}
```

This crashes the front end and is accepted with no issues by the analyzer.

If we adjust 'Note that both types must be class types at this point' to 'Note that both types must be class types or `FutureOr<T>` for some `T` at this point' in the two subtype rules then they will handle some of these cases, e.g., `UP(FutureOr<int>, int) == FutureOr<int>`.

A set of sound rules could be expressed as follows (but we may not wish to use them directly because they use nested applications of **UP**, and they rely on least upper bound being commutative):

- **UP**(`FutureOr<T> `, `T2`) = **UP**(**UP**(`Future<T>`, `T2`), **UP**(`T`, `T2`)).
- **UP**(`T1`, `FutureOr<T> `) = **UP**(**UP**(`T1`, `Future<T>`), **UP**(`T1`, `T`)).
- **UP**(`Future<T>`, `T`) = `FutureOr<T>`.
- **UP**(`T`, `Future<T>`) = `FutureOr<T>`.

@leafpetersen, WDYT?

leafpetersen:
@eernstg Interesting, good catch! Pre null-safety, essentially the rules that were implemented were as follows:

- **UP**(`FutureOr<T1> `, `FutureOr<T2>`) = `FutureOr<T3>` where `T3 = **UP**`(T1, T2)`
- **UP**(`T1`, `FutureOr<T> `) = `Object`
[...]

PR #1254, opened 2020-10-07
Add rules about FutureOr to UP

eernstg:
This PR proposes a set of rules about **UP** with arguments including `FutureOr` based on the discussion in https://github.com/dart-lang/language/issues/1250. It does not introduce `FutureOr` as the result of **UP** when `FutureOr` is not among the arguments.

### DIFF
--- a/resources/type-system/upper-lower-bounds.md
+++ b/resources/type-system/upper-lower-bounds.md
@@ -173,6 +173,11 @@ We define the upper bound of two types T1 and T2 to be **UP**(`T1`,`T2`) as foll
 - **UP**(`T Function<...>(...)`, `S Function<...>(...)`) = `Function` otherwise
 - **UP**(`T Function<...>(...)`, `T2`) = **UP**(`Object`, `T2`)
 - **UP**(`T1`, `T Function<...>(...)`) = **UP**(`T1`, `Object`)
+- **UP**(`FutureOr<T1>`, `FutureOr<T2>`) = `FutureOr<T3>` where `T3` = **UP**(`T1`, `T2`)
+- **UP**(`Future<T1>`, `FutureOr<T2>`) = `FutureOr<T3>` where `T3` = **UP**(`T1`, `T2`)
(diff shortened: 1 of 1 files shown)


resources/type-system/upper-lower-bounds.md
```diff
@@ -173,6 +173,11 @@ We define the upper bound of two types T1 and T2 to be **UP**(`T1`,`T2`) as foll
 - **UP**(`T Function<...>(...)`, `S Function<...>(...)`) = `Function` otherwise
 - **UP**(`T Function<...>(...)`, `T2`) = **UP**(`Object`, `T2`)
 - **UP**(`T1`, `T Function<...>(...)`) = **UP**(`T1`, `Object`)
+- **UP**(`FutureOr<T1>`, `FutureOr<T2>`) = `FutureOr<T3>` where `T3` = **UP**(`T1`, `T2`)
+- **UP**(`Future<T1>`, `FutureOr<T2>`) = `FutureOr<T3>` where `T3` = **UP**(`T1`, `T2`)
+- **UP**(`FutureOr<T1>`, `Future<T2>`) = `FutureOr<T3>` where `T3` = **UP**(`T1`, `T2`)
+- **UP**(`T1`, `FutureOr<T2>`) = `FutureOr<T3>` where `T3` = **UP**(`T1`, `T2`)
+- **UP**(`FutureOr<T1>`, `T2`) = `FutureOr<T3>` where `T3` = **UP**(`T1`, `T2`)
```
lrhn, line 180: I think this works for nullable types too, but what if we have: `FutureOr<int>` and `Future<int>?`, will we be able to get to `FutureOr<int>?` ?

leafpetersen: Yes, this is handled correctly.  The nullability rule kicks in above first, we recursively find the LUB of the FutureOr and Future, and then make it nullable.  I'm not positive that all combinations of unions will be handled correctly though.

leafpetersen (approved):
LGTM, thanks.

Issue #1081, opened 2020-07-09
Null-shorting to the left by desugaring?


eernstg:
Cf. https://github.com/dart-lang/sdk/issues/42379.

Null-shorting allows expressions like `e?.a().b().c()` because execution of both `a()`, `b()`, and `c()` is omitted in the case where `e` evaluates to null.

However, `(e?.a()).b().c()` is an error, because `e?.a()` may yield null, and null-shorting does not propagate through parentheses (so we must use `(e?.a())?.b().c()`).

In general, null-shorting proceeds to the right and includes only constructs derived from `<selector>` (that is, member invocations like `.getter`, `.method(...)`), so we can't have `e1?.a().b().c() + e2` and expect null-shorting to skip the invocation of `+` if `e1` evaluates to null. Instead, `e1?.a() + e2` is an error because the receiver may be null.

It turns out that we have a couple of exceptions:

```dart
class C {
  C e() => this;
  C operator -() => this;
  C operator ~() => this;  
  C operator +(C other) => this;
  int f = 1;
}

void main() {
  C? c = new C();
  -c?.e();
  ~c?.e();
}
```

The above program is accepted by the analyzer (SDK 2.9.0-edge.2ed302bfbbbad1d10f6c839c504ae3871fc96dc2), but rejected by the common front end.

Acceptance of the program may be justified by [the following sentence](https://github.com/dart-lang/language/blob/219a9b15d8e34dfaf2a03c23d7bdbebf70c1f045/specification/dartLangSpec.tex#L13414) in the language specification:

> Any other expression of the form `op e` is equivalent to the
> method invocation `e.op()`.

This rule actually allows null-shorting to proceed from a subexpression of `e` and all the way up to, and including, `.op()`. This affects unary `-` and `~`.

[...]

PR #1256, opened 2020-10-09
Specify that null shorting does not include operators

eernstg:
Cf. #1081, this PR updates the null-safety spec of null shorting to say that it does _not_ use the equivalence between operator invocations and method invocations that the language specification specifies.

For instance `-e` is specified as equivalent to `e.unaryMinus()` (for a suitable declaration of method `unaryMinus`), and similarly for `e1 + e2` respectively `e1.plus(e2)`.

The tools already treat operators in this manner, so there are no implementation issues.

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -1362,6 +1362,24 @@ continuation.
   - A list literal `[e1, ..., en]` translates to `TERM[ [EXP(e1), ..., EXP(en)] ]`
   - A parenthesized expression `(e)` translates to `TERM[(EXP(e))]`
 
+The language specification specifies that an invocation of any of several
+operators is considered equivalent to a member access (this applies to
+relational expressions, bitwise expressions, shift expressions, additive
+expressions, multiplicative expressions, and unary expressions).
+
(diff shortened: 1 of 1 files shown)


accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1362,6 +1366,25 @@ continuation.
...
+Similarly, `-e` is equivalent to `e.unaryMinus()`.*
+
+This equivalence is not applicable in the above rules, so operators not
+mentioned specifically in a rule are handled in the case for 'other'
+expressions, not in the case for `e.m(args)`.
+
+*This means that the null-shorting transformation stops at operators. For
+instance, `e?.f + b` is a compile-time error because `e?.f` can be null, it is
+not an expression where both `.f` and `+ b` will be skipped if `e` is null.
+Similarly, both `-a?.f` and `~a?.f` is an error, it does not null-short like
```
leafpetersen, line 1385: "both ... are errors, and do not..." or "each of ... is an error, and does not...".

leafpetersen (approved):

accepted/future-releases/nnbd/feature-specification.md
eernstg: Done.

PR #1259, opened 2020-10-14
Clarify conflict merging

eernstg:
Thanks for all the hard work on 1161! Here's a follow-up, clarifying that conflict merging never considers import prefixes as conflicted.


lrhn (approved):
Super!

eernstg:
Thanks!

PR #1260, opened 2020-10-15
Grammar fixes

eernstg:
This one is basically a bunch of typos, shouldn't be heavy.

In preparation for releasing a stable 2.8 specification, this PR makes a number of corrections to the language specification grammar rules.

It does not remove elements (like `\LATE{}` and `?` on types) associated with null safety, that will be done later.

Note that `<metadata>` was corrected in several rules (added or removed). CL https://dart-review.googlesource.com/c/sdk/+/167801 tests that tools accept metadata in all locations specified by the grammar, after this change.

To avoid some clutter, a number of terms were simplified as in `\LATE{}?` --> `\LATE?`.

In a few locations `>>>` and `>>` were swapped. This is needed in the spec parser (Dart.g), because it only parses as '>' '>' '>' and the longest match must come first. So technically this is not necessary, but it seems helpful to avoid that as a source of confusion, especially given that the spec explicitly says that the rules are ordered.

Added `late` and `required` as built-in identifiers. They will be removed in a CL that eliminates null-safety elements, but for now it's better to be consistent (also: reverting that CL will then make it correct for null safety).

`?` was removed from top-level external declarations of functions/getters/setters. This has been around forever, but there is no way it can be correct to use `external? <functionSignature> ';'`, it should be `external <functionSignature> ';'`. So that's a bug fix. Implementations don't allow the `external` to be omitted, so there is no implementation effort.

Added `<EOF>` at the end of `<libraryDeclaration>`. We had it in `<partDeclaration>` but not `<libraryDeclaration>` so we should add it to libraries or remove it from parts. It may seem reasonable to say that a library and a part cannot be followed by arbitrary garbage, and having `<EOF>` will say that.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -995,11 +995,11 @@ Variables are storage locations in memory.
 \begin{grammar}
 <variableDeclaration> ::= <declaredIdentifier> (`,' <identifier>)*
 
-<declaredIdentifier> ::= <metadata> \COVARIANT{}? <finalConstVarOrType> <identifier>
+<declaredIdentifier> ::= <metadata> \COVARIANT? <finalConstVarOrType> <identifier>
 
-<finalConstVarOrType> ::= \LATE{}? \FINAL{} <type>?
+<finalConstVarOrType> ::= \LATE? \FINAL{} <type>?
   \alt \CONST{} <type>?
-  \alt \LATE{}? <varOrType>
+  \alt \LATE? <varOrType>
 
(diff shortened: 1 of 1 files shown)


lrhn (approved):
LGTM. Check that it doesn't change the result (except where removing something unnecessary)

eernstg:
Thanks! https://dart-review.googlesource.com/c/sdk/+/167805 was used to do just that, using a version of Dart.g which has been updated to match the spec grammar.

Issue #447, opened 2019-07-11
Should asserts in const expressions always be evaluated?


leafpetersen:
A question was raised as to whether asserts in const expressions should always be evaluated, or only when asserts are otherwise enabled.  Note that the notion of "asserts are otherwise enabled" is not necessarily well-defined at the point that consts are evaluated if we allow turning assertions on and off dynamically.  In discussion, it seems useful to allow normal assertions to be:
  - stripped out/not stripped out during compilation
  - if not stripped out, turned on/turned off at runtime

Since consts are evaluated at compiled time, the only way we could turn const asserts on/off would be via the compile time control.  That implies that we would only not evaluate const asserts if the compile time flag was off.

If we allow the compile time flag to turn off const assertions, then in table form the behavior looks like this:

|                             | runtime flag on | runtime flag off |
| ---------------- | --------------- | --------------- |
| compile flag on  | const + normal | const only         |
| compile flag off  | none                  | none                  |

If we always evaluate const asserts, then the behavior looks like this:

|                             | runtime flag on | runtime flag off |
| ---------------- | --------------- | --------------- |
| compile flag on  | const + normal | const only         |
| compile flag off  | const only        | const only         |

I don't see a lot of value in not evaluating const asserts.  I also suspect that the analyzer always evaluates them: @stereotype441 is that correct?

So my inclination would be to go with the second behavior, where const asserts are always evaluated.  The implementations are going with this behavior for now, since it is simpler to implement.  If we decide to require the ability to turn on/off const asserts they can add this later as a non-breaking change.

cc @lrhn @munificent @eernstg @mraleph @a-siva

munificent:
> I don't see a lot of value in not evaluating const asserts. 

+100. The only real value I see in *not* evaluating an assertion is performance. Const asserts have zero performance cost, so there's no reason to not always evaluate them.

lrhn:
Agree, it brings real value to evaluate assertions, and the main reason not to is to avoid production overhead.

However, it means that:
```dart
class C {
  const C(int x) : assert(x != 0);
}
```
behaves differently between compile-time and run-time.
A `const C(1)` is a compile-time error and `new C(1)` is perhaps no error at all. 
We have to make sure that we don't introduce run-time errors that shouldn't be there.

eernstg:
Agreed, same reasons as @munificent and @lrhn! ;-)

[...]

PR #1271, opened 2020-10-27
Specify that constant assertions are always evaluated

lrhn:
See https://github.com/dart-lang/language/issues/447 for context.

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -6,6 +6,9 @@ Status: Draft
 
 ## CHANGELOG
 
+2020.10.26
+  - Specify that constant evaluation executes all assertions.
+  
 2020.10.14
   - Include selector `!` among the null-shorting constructs.
 
@@ -1007,6 +1010,33 @@ void main() {
   print(const ConstantClass<int>());
 }
 ```
+
+### Constant assertions
+
+Currently assertions are enabled only when the compiler is passed an `enable-asserts` flag.
+With null safety, when an assert is evaluated as part of a constant invocation of a `const`
+constructor, any `assert`s in the initializer list are evaluated. 
+The `enable-asserts` flag then only applies to *run-time assertions*.
+
+More precisely, a assert _s_ of the form <code>assert(_e_<sub>1</sub>, _e_<sub>2</sub>)</code>
+is executed as follows:
+  * If the execution of _s_ happens during evaluation of a const invocation of the 
+    surrounding `const` constructor, then _e_<sub>1</sub> is evaluated to a constant value _c_.
+    It is a compile-time error if this evaluation would have thrown, or if _c_ is not the
+    `bool` value `true`, and otherwise _s_ completes normally.
+  * Otherwise (when it's not part of a constant evaluation), if run-time assertions are enabled
+    then then _e_<sub>1</sub> is evaluated to a value _v_.
+    If _v_ is not an instance of the type `bool`, then a dynamic error occurs.
+    If _v_ is the value `false` then _e_<sub>2</sub> is evaluated to a value _m_,
+    and an object, _o_, implementing the class `AssertionError` and containing _m_ is created.
+    Then execution of _s_ throws _o_ and a stack trace corresponding 
+    to the current evaluation context.
+    If _v_ is the value `true` then execution of *s* completes normally.
+  * Otherwise (when execution is not not part of a constant evaluation 
+    and run-time assertions are disabled), execution of _s_ completes normally without evaluating
+    either of _e_<sub>1</sub> or _e_<sub>2</sub>.
+
+(An assert of the form <code>assert(_e_)</code> is equivalent to <code>assert(_e_, null)</code>.)
 
 ### Null promotion
 


accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1008,6 +1011,33 @@ void main() {
 }
 ```
 
+### Constant assertions
+
+Currently assertions are enabled only when the compiler is passed an `enable-asserts` flag.
+With null safety, when an assert is evaluated as part of a constant invocation of a `const`
+constructor, any `assert`s in the initializer list are evaluated.
```
eernstg, line 1018: Could say `are enabled`, to keep the number of concepts about assertions lower, and allowing implementations to compile non-enabled assertions away entirely, or keep them and skip them, etc.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1008,6 +1011,33 @@ void main() {
 }
 ```
 
+### Constant assertions
+
+Currently assertions are enabled only when the compiler is passed an `enable-asserts` flag.
+With null safety, when an assert is evaluated as part of a constant invocation of a `const`
```
eernstg, line 1017: We could call them `an assertion` as a rule, and only use `assert` when referring to actual code. The language specification has `assert statement` which is a statement that is an assertion, but otherwise it refers to assertions as `assertion`s. Might as well do that here, too.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1008,6 +1011,33 @@ void main() {
...
+constructor, any `assert`s in the initializer list are evaluated. 
+The `enable-asserts` flag then only applies to *run-time assertions*.
+
+More precisely, a assert _s_ of the form <code>assert(_e_<sub>1</sub>, _e_<sub>2</sub>)</code>
+is executed as follows:
+  * If the execution of _s_ happens during evaluation of a const invocation of the 
+    surrounding `const` constructor, then _e_<sub>1</sub> is evaluated to a constant value _c_.
```
eernstg, line 1024: How about `then it is a compile-time error if _e_<sub>1</sub> is not a constant expression with the value `true``?

Then we don't have to mention 'completes normally' as if it's part of the constant evaluation semantics.

eernstg (approved):
LGTM. I put the question about going ahead and doing this on the language team meeting agenda for tomorrow.

accepted/future-releases/nnbd/feature-specification.md
lrhn: I think that sentence is garbled. That said, an assertion "being enabled" is a concept more than simply saying what its semantics are in different cases. Implementations can still completely remove run-time assertions from the compiled output when assertions are disabled, because they have the "null-semantics": completes normally with no side effects.

accepted/future-releases/nnbd/feature-specification.md
lrhn: Good idea. I think I did that deliberately when I introduced assertions into initializer lists (where they are not statements).

accepted/future-releases/nnbd/feature-specification.md
lrhn: Good point. We already do the "compile-time error if expression would throw" for constant evaluation.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1008,6 +1011,33 @@ void main() {
 }
 ```
 
+### Constant assertions
+
+Currently assertions are enabled only when the compiler is passed an `enable-asserts` flag.
```
eernstg, line 1016: The text in section 'Assert' of the language specification already says almost everything which is said here.

Wouldn't it suffice to have a very short section here? We would then rely on the existing specification text for all the details:

```suggestion
Assertions are enabled during the evaluation of a constant object expression.

*This means that properties checked by assertions will always be enforced with constant expressions. The implementation specific mechanism to enable assertions is used to enable the remaining assertions, namely the ones that are executed at run time.*
```

eernstg (approved):
I think we could rely more on the existing specification text, and just have a very short section in this document.

accepted/future-releases/nnbd/feature-specification.md
lrhn: It's tricky, because we don't actually separate compile-time and run-time in the specification, we just specify the meaning of constants such that they *can* be pre-computed, but we don't actually say that you have to do so.

We still need a way to specify that something is happening as part of constant evaluation because it matters in some situations. For example `const x = 1 ~/ 0;` is a compile-time error instead of throwing a `DivisionByNullError`.

eernstg: Right, we specify that evaluation of a constant object expression occurs at run time and then it is (recursively, bottom-up) canonicalized. It makes no difference whether that constant object expression is in a constant context, whether it is used in the initialization of a `const` variable, etc., it is a rule for `<constObjectExpression>` anywhere. Of course, if it is in a constant context then the `const` keyword may be absent, it is still statically known to be there.

This rule is compatible with an evaluation at compile time. However, it doesn't _require_ an implementation to perform the evaluation at compile time and reuse that result every time it is needed at run time. For instance, the expression could be evaluated at compile time and the result could be discarded, and then evaluation and canonicalization could be done at run time. Also, because the constant sublanguage is small and Turing-incomplete, execution is guaranteed to be possible at compile time. So when we want to determine whether an error should be emitted, we can evaluate the constant object expression.

So if we do say 'Assertions are enabled during the evaluation of a constant object expression' then it is allowed for an implementation to raise the compile-time error early ("at compile time"). That's probably exactly what we want.

This also means that _if_ a tool gives rise to an execution where a constant object expression is actually evaluated at run time then there will be a run-time error, even though we set out to say that these errors should occur at compile-time. But tools don't have to do this, so in practice we can maintain that these errors occur early.

If an execution uses hot reload then compilation will occur at run time anyway.

Don't you think that would be OK for tool teams to work with?

PR #1298, opened 2020-11-09
Change terminology to use unsound/sound rather than weak/strong

eernstg:
We may wish to update the nnbd spec to use the sound/unsound terminology rather than weak/strong. This CL performs such a rename. The `expect.dart` library and language tests have recently been changed similarly.


leafpetersen (approved):
LGTM, thanks!

PR #1309, opened 2020-11-12
Add possible compile-time error for external function

eernstg:
Cf. https://github.com/dart-lang/sdk/issues/42730:

This PR adds an implementation specific compile-time error at `external` function declarations. It is intended to allow for the behavior of `dart2js` in the situation where an external function does not have a `@JS(...)` annotation. It does not mention that specific motivation, because the same text should allow for an error in situations that are similar, but not identical.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -2018,25 +2018,51 @@ will necessarily have the same run-time type.
 \LMHash{}%
 An \IndexCustom{external function}{function!external}
 is a function whose body is provided separately from its declaration.
-An external function may be a top-level function (\ref{librariesAndScripts}), a method (\ref{instanceMethods}, \ref{staticMethods}), a getter (\ref{getters}), a setter (\ref{setters}) or a non-redirecting constructor (\ref{generativeConstructors}, \ref{factories}).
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -2018,25 +2018,51 @@ \subsection{External Functions}
...
+An external function may be
+a top-level function (\ref{librariesAndScripts}),
+a method (\ref{instanceMethods}, \ref{staticMethods}),
+a getter (\ref{getters}),
+a setter (\ref{setters}),
+or a non-redirecting constructor
+(\ref{generativeConstructors}, \ref{factories}).
+External functions are introduced via the built-in identifier \EXTERNAL{}
+(\ref{identifierReference})
+followed by the function signature.
```
lrhn, line 2030: Maybe mention already that a variable declaration (top-level, static or instance) can be marked external, and that it's equivalent to declaring an external getter and (unless `final`) setter.

lrhn (approved):

specification/dartLangSpec.tex
eernstg: I think I'll do that separately. It would be somewhat bigger, because it involves grammar changes as well, and changes to section 'Variables'.

PR #1315, opened 2020-11-16
Little script to extract the spec's grammar as plaintext.

munificent:
When working on Markdown feature specs, I end up copying pieces of the grammar out of the spec. In Chrome, it ends up putting `h` before and after every grammar rule and does some other weird stuff. I got tired of fixing that up manually every time, so this little script extracts all of the grammar rules from the spec, cleans them up, and prints them out.

Kind of hacky, but it seems to work. :)

cc @leafpetersen @lrhn

### DIFF
--- a/tools/plaintext_grammar.dart
+++ b/tools/plaintext_grammar.dart
@@ -1,0 +1,68 @@
+import 'dart:io';
+
+/// Reads the "dartLangSpec.tex" and prints out all of the grammar rules it
+/// contains as plaintext.
+
+/// Matches Tex keyword references like:
+///
+///     \LATE
+///     \FINAL{}
+final keywordRegExp = RegExp(r'\\([A-Z]+)(\{\})?');
+
+/// Matches non-terminals like:
+///
+///     <declaredIdentifier>
+final ruleRegExp = RegExp(r'<(\w+)>');
+
+void main(List<String> arguments) {
+  if (arguments.isEmpty) {
+    print("Usage: dart plaintext_grammar.dart <path to dartLangSpec.tex>");
+    exit(1);
+  }
+
+  var specFile = File(arguments[0]);
+
+  var inGrammar = false;
+  for (var line in specFile.readAsLinesSync()) {
+    line = line.trimRight();
+
+    if (line == r'\begin{grammar}') {
+      inGrammar = true;
+    } else if (line == r'\end{grammar}') {
+      print('');
+      inGrammar = false;
+    } else if (inGrammar) {
+      line = line
+          .replaceAll('`', "'")
+          .replaceAll(r'\alt', '|')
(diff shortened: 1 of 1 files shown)


leafpetersen (approved):

tools/plaintext_grammar.dart
```diff
@@ -0,0 +1,68 @@
...
+          .replaceAll(r'\\u', r'\u')
+          .replaceAll(r'\\v', r'\v')
+          .replaceAll(r'\\x', r'\x')
+          .replaceAll(r'\sqsqsq', r"\'\'\'")
+          .replaceAll(r'\sqsq', r"\'\'")
+          .replaceAll(r'\sq', r"\'")
+          .replaceAll(r'\\', r'\')
+          .replaceAll('<NEWLINE>', "'\\n'")
+          .replaceAllMapped(
+              keywordRegExp, (match) => "'${match[1]!.toLowerCase()}'")
```
eernstg, line 63: Right, except that `\FUNCTION{}` needs to turn into `'Function'`.

tools/plaintext_grammar.dart
```diff
@@ -0,0 +1,68 @@
...
+          .replaceAll(r'\\r', r'\r')
+          .replaceAll(r'\\t', r'\t')
+          .replaceAll(r'\\u', r'\u')
+          .replaceAll(r'\\v', r'\v')
+          .replaceAll(r'\\x', r'\x')
+          .replaceAll(r'\sqsqsq', r"\'\'\'")
+          .replaceAll(r'\sqsq', r"\'\'")
+          .replaceAll(r'\sq', r"\'")
+          .replaceAll(r'\\', r'\')
+          .replaceAll('<NEWLINE>', "'\\n'")
```
eernstg, line 61: I think this line should be removed: It should work to have a lexical rule `NEWLINE ::= '\n' | '\r' | '\r\n'` and using it in a few places.

If we try to inline that lexical rule then the rule itself becomes `'\n' ::= ...`, which is presumably not a grammar rule in the target notation, and we'd need to put `'\n' | '\r' | '\r\n'` in each position where `<NEWLINE>` is used. So it's easier to just keep the `NEWLINE ::= ...` rule as it is.

eernstg (approved):
LGTM, noting a couple of glitches in comments.

tools/plaintext_grammar.dart
munificent: Good catch! Done.

tools/plaintext_grammar.dart
munificent: Oops, right. I didn't mean to inline the rule. I didn't realize when I first saw `<NEWLINE>` that it was referring to another rule. Fixed.

PR #1349, opened 2020-12-04
modify grammer

Naoki-Hidaka:


devongalat:
Why exactly do you need to make 'help' an infinitive here? There is nothing to be fixed because the sentence is grammatically correct. Your change makes it grammatically **incorrect**.

munificent:
Hi, thanks for contributing, but I think the sentence is correct already.

PR #1355, opened 2020-12-07
Add missing word

mnordine:
@munificent This paragraph starts out confusing to me, since you use `Point(...)` in 2 places, but I think you're trying to differentiate them by using the word `outer` for the matcher, but maybe you missed `inner` for the binder?


munificent:
Yeah, I didn't word this well. Both places are referring to the same `Point(...)` pattern. By "outer", I was referring to the `Point(...)` pattern itself as opposed to the `var x` and `var y` patterns inside it. There is no "inner" `Point(...)` pattern since there's just the one.

Issue #848, opened 2020-02-20
Non-function type aliases: Add restrictions


eernstg:
Working on non-function type aliases, we have some interactions with type inference. I believe that we need to reduce the expressive power of this mechanism in order to maintain tractability, that is, the language design needs to be refined/restricted.

In particular, we aimed to allow a type alias to be used for instance creation (that is, it should be possible to use it as a constructor), because this would be required in order to allow a type alias to work as a "rename" on a class, or as a shorthand for passing a specific pattern of type arguments to a class (for instance, `typedef C<X> = p.C<X, int>` could allow us to add a type parameter to an existing class `C`, and then have existing code import the new class with a prefix and pass the extra type argument using a typedef -- and then the existing code wouldn't have to be updated).

But the ability to use a type alias for instance creation creates some new situations that we might not want to address, and possibly inherently intractable situations. The core problem is that instance creation is subject to type inference. There are other situations where a type alias is similarly subject to inference, in particular when a generic mixin is specified using a raw type alias. The constructor case is probably the most problematic, so I'll use that for illustration.

The following kind of type alias introduces a new kind of inference:

```dart
// Questionable kind of type alias: Right hand side is a type variable.
typedef F<X, Y extends Iterable<X>> = X;

// Illustration of issue.
class Foo { Foo.named(); }
class Bar { Bar.named(); }

void main() {
  // Infer `F<Foo, Iterable<Foo>>` because `Foo.named` denotes a constructor,
  // and because `Y` is unconstrained by the right hand side, but has a bound?
  Foo foo = F.named();
}
```

**We could say** that it is simply an error for a type alias to have a right hand side which is a type variable.

In general, a non-function type alias could have another non-function type alias as its right hand side, so we would have a function from a list of types to another list of types, composed from any number of such functions. Each of them could omit some of its arguments or use some of its arguments more than once, or in some other parameterized type (type alias or class).

This gives rise to a function that may not have an inverse (if a type variable is unused in one of these steps, the value is unconstrained; if it is used twice then it creates an equality constraint; constraints on the declared bounds may exist as well, at each step).

```dart
// Setup with type aliases affecting inference. 
typedef F1<X> = F2<X, List<X>>;
typedef F2<X, Y> = Map<X, Y>;

void main() {
  Map<num, Iterable<int>> map = F1(); // Infer type argument `int`?
}
```

[...]

PR #1361, opened 2020-12-10
Is an error to create an instance of a type alias that expands to a type variable

eernstg:
Cf. #848; we agreed that at should be an error to have instance creations like `new F<C>()` when `F` is a type alias of the form `typedef F<X> = X;`. This PR adds that rule to the non-function type alias feature specification.


scheglov:
Do we allow type aliases that expand to a type variable in redirecting factory constructors?

```dart
class A implements B {}

typedef X<T> = T;

class B {
  factory B() = X<A>;
}
```

eernstg:
No, that should be an error as well. Thanks! &mdash; I just added this error to the PR.

leafpetersen (approved):

PR #1367, opened 2020-12-12
Update patterns-feature-specification.md

timsneath:
Spotted this tiny nit as I was reading.


munificent:
Maybe I leave mistakes in like this deliberately to see who reads the proposal. :)

PR #1374, opened 2020-12-17
Clarify .call on function type bounded types

eernstg:
The language specification is currently too vague on the ability to look up `.call` on a receiver whose type is function type bounded. This PR specifies that the `.call` is available, and gives its type.

Example: When `F extends void Function(int)`, with `F f`, `f.call` is allowed and has type `void Function(int)`.

Cf. https://github.com/dart-lang/sdk/issues/44506: The action requested there relies on the interpretation which is made explicit with this PR. We already specify explicitly that any object whose static type is function type bounded [can be called](https://github.com/dart-lang/language/blob/988d82bd0f126192fa4af8616654f75e76f675eb/specification/dartLangSpec.tex#L10612), and it would be inconsistent to allow `f(42)` and disallow `f.call(42)`, just because the static type of `f` is function-type bounded rather than being that function type.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -4326,6 +4326,9 @@ any type $T$ which is $T_0$ bounded
 (\ref{bindingActualsToFormals}),
 where $T_0$ is a class with interface $I$,
 is also considered to have interface $I$.
+Similarly, when $T$ is $F$ bounded where $F$ is a function type,
+$T$ is considered to have a method named \CALL{} with signature $m$,
+such that the function type of $m$ is $F$.
 
 \LMHash{}%
 The \Index{combined interface} $I$ of a list of interfaces \List{I}{1}{k}
@@ -10551,10 +10554,10 @@ corresponds to $f$.
 %
 A reference to this section is also used in other sections
 when actual arguments are to be bound to the corresponding formal parameters,
-and $f$ is about to be invoked, to specify the dynamic semantics.
+and $f$ is about to be invoked, to specify the dynamic semantics.%
 }
 
-\commentary{
+\commentary{%
 We do not call $f$ a `function object' here, because we do not wish to imply
 that every function invocation must involve a separate evaluation
 of an expression that yields a function object,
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -4326,6 +4326,9 @@ \section{Interfaces}
 (\ref{bindingActualsToFormals}),
 where $T_0$ is a class with interface $I$,
 is also considered to have interface $I$.
+Similarly, when $T$ is $F$ bounded where $F$ is a function type,
```
lrhn, line 4329: Is "$F$ bounded" a term we generally use for a type variable having an upper bound of `$F$`.
I feel it's very confusable with having F-bounded polymorphism. 😉 

I can see we use "any type $T$ which is $T_0$ bounded`" above. Does it just mean "which has $T_0$ as supertype"? Or must $T$ be a type variable and $T_0$ its bound?
In the former case, it's not true, we need the most specific function super-type for the type of `call`, not any function super-type.

lrhn (approved):

specification/dartLangSpec.tex
eernstg: Yes, it's a bit of a conflict to use `$F$` to name the type, but function types are often named like that. We do use `$T$ bounded` to mean types like `X extends T` and `X extends Y`, `Y extends T`, with special cases like 'function-type bounded' as a common term for all `$F$` bounded types where `$F$` is a function type. I changed the name to `$T_0$` (that's consistent with the previous sentence, so at least it's consistent with _something_ ;-).

We can have cases where `$T$` is `$T_0$` bounded and `$T$` isn't a type variable: For instance, `$T_0$` is `$T_0$` bounded for all `$T_0$`. But whenever we _use_ "`$T_0$ bounded", `$T_0$` is a less general type: It can be an interface type or a function type, and we might need additional cases in the future. But the point is that we just want to skip over the `extends` steps for a type variable until we get to a type that has some affordances, and then we'll give the same affordances to that type variable.

The notion of being `$T_0$` bounded allows us to skip from a type variable to its bound until we encounter a type which is not a type variable, but it does not allow any other subtype steps. So there should not be a problem with the most specific function supertype vs. other function types that are supertypes of the most specific one.

PR #1375, opened 2020-12-17
Review based corrections to #1374

eernstg:
I'm sorry, I forgot to press enter so the review related changes weren't uploaded before #1374 was landed. Here they are.


eernstg:
@lrhn, this one is quick: I just forgot to press `ENTER` before landing #1374, and this PR is the missing updates.

lrhn (approved):

PR #1378, opened 2020-12-19
Update 3-BUG.md

chintanmalaviya:
input issue

import 'dart:io';

void main() {
  stdout.writeln('Type something');
  String input = stdin.readLineSync();
  stdout.writeln('You typed: $input');
}


natebosch:
Were you trying to file an issue instead of opening a PR?

You don't mention what the problem you encountered was, so this isn't actionable for us. FWIW on the VM your example does exactly what one would expect. You do need to hit `Enter` after typing something in case that was the confusion.

Issue #1350, opened 2020-12-04
Should const evaluation normalize type arguments?


eernstg:
Consider the following program:

```dart
import 'dart:async';

const xs = <Object?>[];
const ys = <FutureOr<Object?>?>[];
const z = identical(xs, ys)? 1 ~/ 0 : 0;

void main() {}
```

This program compiles and runs with no errors with `dart` (from commit 74be6670c9b5fff68caec4112f13216031c4b0fc and in version 2.12.0-96.0.dev), both with `--sound-null-safety` and with `--no-sound-null-safety`.

But with the analyzer, a compile-time error is reported:

```
ERROR|COMPILE_TIME_ERROR|CONST_EVAL_THROWS_IDBZE|/usr/local/google/home/eernst/lang/dart/scratch/202012/n008.dart|5|30|6|Evaluation of this constant expression throws an IntegerDivisionByZeroException.
```

This shows that constant evaluation in the analyzer performs [normalization](https://github.com/dart-lang/language/blob/master/resources/type-system/normalization.md) on the type arguments passed to expressions that are subject to constant expression evaluation, but no such normalization takes place in the common front end.

We should decide on which of these behaviors is correct, and ask "the other" implementation to make adjustments.

leafpetersen:
I have proposed to do normalization of type arguments to constant constructors for the purposes of canonicalization [here](https://github.com/dart-lang/language/pull/1389).

Issue #1346, opened 2020-12-03
Does the analyzer need `--sound-null-safety` / `--no-sound-null-safety`?


eernstg:
It seems likely that we need to adjust the definition of canonicalization of the values of constant expressions, or we need to allow the analyzer to take the execution modes with sound null checking (`--sound-null-safety`) vs. unsound null checking (`--no-sound-null-safety`) into account.

### Issue

The static analysis of Dart programs with null safety is mostly identical for the two execution modes. However, there is a difference in the treatment of constant expressions: With unsound null checking, the type arguments passed to collection literals and constructor invocations [are treated](https://github.com/dart-lang/language/blob/master/accepted/future-releases/nnbd/feature-specification.md#const-evaluation-and-canonicalization) as legacy types, and with sound null checking they are treated as the given types without any transformations. This makes a difference for canonicalization:

```dart
const xs = <Never>[];
const ys = <Null>[];
const z = identical(xs, ys)? 1 ~/ 0 : 0;

void main() {}
```

This program is accepted with no errors by `dartanalyzer` (commit a9d212b94aa73472824f4218a567a8e004575112 as well as 2.12.0-96.0.dev). It is also accepted by `dart` (same versions) with sound null checking, but it encounters a compile-time error with unsound null checking. This is consistent with the fact that the type `Never` is considered to denote the same type as `Null` when encountered in a legacy library (it does not mean `Never*`).

## Possible resolution

If it is considered impractical to let the analyzer have such modes then we'd need to make sure that canonicalization never differs with sound and unsound null checking.

The only situation where we have the choice is with an entry point where every directly or indirectly imported library enables null safety (that is, in a purely null safe program): With a mixed-version program we can only use unsound null checking.

So when do types `S` and `T` differ, but when "treated as a legacy type" they do not differ?

I believe that it may be sufficient to say that "`T` treated as a legacy type" denotes the type `S` where:

- If `T` is atomic then `S` is `T*`.
- If `T` is `FutureOr<T1>` then `S` is `FutureOr<S1>*` where `S1` is the result of treating `T1` as a legacy type.
- If `T` is a parameterized type `T0<T1, .. Tk>` then `S` is `T0<S1, .. Sk>*` where `Sj` is the result of treating `Tj` as a legacy type, `1 <= j <= k`.
- (Similar structural rules for type variables, intersections, function types, etc.)

So we're adding `*` everywhere, but we're _not_ treating `Never` as another name for `Null`. 

[...]

PR #1389, opened 2020-12-31
Specify type literal and const object canonicalization

leafpetersen:
Specify canonicalization for type literals and constant objects with respect to legacy types.  Remove the warning when
an overriding method uses a different default value.  Specify that constant object canonicalization uses normalization on type parameters.

https://github.com/dart-lang/sdk/issues/44246
https://github.com/dart-lang/language/issues/1350
https://github.com/dart-lang/language/issues/1346

### DIFF
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -6,6 +6,12 @@ Status: Draft
 
 ## CHANGELOG
 
+2020.12.30
+  - Remove the warning for overrides with different default values.
+  - Specify canonicalization of type literals.
+  - Specify the unsound mode legacy rewrite of type arguments to const
+    constructors and literals.
+
 2020.11.09
   - Change terminology to use sound or unsound null checking rather than
     weak or strong mode.
@@ -654,6 +660,10 @@ It is a warning to use a null aware operator (`?.`, `?[]`, `?..`, `??`, `??=`, o
 It is a warning to use the null check operator (`!`) on an expression of type
 `T` if `T` is **strictly non-nullable** .
 
+It is no longer a warning to override a method which has a default value for a
+parameter with a method with a different default value for the corresponding
+parameter.
+
 ### Local variables and definite (un)assignment.
 
 As part of the null safety release, errors for local variables are specified to
@@ -1236,18 +1246,92 @@ the `LEGACY_ERASURE` of the types.
 
 ### Const evaluation and canonicalization
 
-With unsound null checking, all generic const constructors and generic
-const literals are treated as if all type arguments passed to them were legacy
-types (both at the top level, and recursively over the structure of the
-types), regardless of whether the constructed class is defined in a legacy
-library or not, and regardless of whether the constructor invocation or
-literal occurs in a legacy library or not.  This ensures that const objects
-continue to canonicalize consistently across legacy and opted-in libraries.
+Const evaluation is modified so that legacy and opted-in instances canonicalize
+more consistently as defined below, using the operation **LEGACY_TYPE**(`T`) on
+types `T` defined as follows.
+
+- **LEGACY_TYPE**(`T`) = `T` if `T` is `dynamic`, `void`, `Null`
+- **LEGACY_TYPE**(`T`) = `T*` if `T` is `Never` or `Object`
+- **LEGACY_TYPE**(`FutureOr<T>`) = `FutureOr<S>*`
+- **LEGACY_TYPE**(`T?`) =
+  - let `S` be **LEGACY_TYPE**(`T`)
+  - if `S` is `R*` then `R?`
+  - else `S?`
+- **LEGACY_TYPE**(`T*`) = **LEGACY_TYPE**(`T`)
+- **LEGACY_TYPE**(`X extends T`) = `X*`
+- **LEGACY_TYPE**(`X & T`) =
+  - This case should not occur, since intersection types are not permitted as
+    generic arguments.
+- **LEGACY_TYPE**(`C<T0, ..., Tn>`) = `C<R0, ..., Rn>*`
+  - where `Ri` is **LEGACY_TYPE**(`Ti`)
+- **LEGACY_TYPE**(`R Function<X extends B>(S)`) = `F*`
+  - where `F = R1 Function<X extends B1>(S1)`
+  - and `R1` = **LEGACY_TYPE**(`R`)
+  - and `B1` = **LEGACY_TYPE**(`B`)
+  - and `S1` = **LEGACY_TYPE**(`S`)
+
+Note that if `T` is a normal form type, then **LEGACY_TYPE**(`T`) is also a
+normal form type.
+
+#### Type literals
+
+Two constant type literals `T1` and `T2` canonicalize to the same object if they
+are equal using the definition of runtime type equality specified above.  The
+object `R` which serves as the canonical form for `T1` and `T2` is defined as
+follows. Let `S1` and `S2` be the types to which `T1` and `T2` correspond.  Note
+that by assumption, **NORM**(`S1`) and **NORM**(`S2`) are syntactically
+identical up to the identity of bound variables and the placement of legacy type
+modifiers.  We choose `R` arbitrarily as the type literal corresponding to one
+of **NORM**(`S1`) or **NORM**(`S2`) with all legacy type modifiers erased.
+
+The effect of this definition is to ensure that type literals which look
+indentical in the source syntax but which may differ by the presence of legacy
+type modifiers are canonicalized consistenly in the sense that any two type
+literals which would compare equal via the definition of runtime type equality
+given above will be canonicalized to the same type object.  The choice of what
+type object to canonicalize to is arbitrary in that placement of legacy
+modifiers in type literals is not otherwise observable in the language.  Any
+consistent choice of placement of legacy modifiers is hence an equally valid
+implementation.
+
+Note that the choice of canonicalization for type literals does not depend
+directly on whether sound or unsound null safety semantics are in use.  However,
+soundly null safe programs will never contain legacy types, and hence the
+erasure of legacy type modifiers is never necessary when running in sound mode.
+
+#### Constant instances
+
+In both sound and unsound null checking, canonicalization of generic object
+instances is performed relative to the normal forms of their generic type
+arguments.  That is, an instance of `C<T0>` canonicalizes to the same object as
+`C<T1>` if `T0` and `T1` have the same normal form (up to the identity of bound
+variables), and the objects are otherwise identical.
 
 With sound null checking, all generic const constructors and generic const
-literals are evaluated using the actual type arguments provided, whether legacy
-or non-legacy.  This ensures that with sound null checking, the final
-consistent semantics are obeyed.
+literals are evaluated using the actual type arguments provided.  This ensures
+that with sound null checking, the final consistent semantics are obeyed.
+
+With unsound null checking, all generic const constructors and generic const
+literals are additionally treated as if all type arguments passed to them were
+legacy types regardless of whether the constructed class is defined in a legacy
+library or not, and regardless of whether the constructor invocation or literal
+occurs in a legacy library or not.  That is, a constructor invocation or object
+literal with generic type parameters `Ti` is treated as if the parameters were
+**LEGACY_TYPE**('Ti'). Implementations may choose to eagerly normalize the type
+arguments before applying the legacy rewrite, as desired.
+
+This ensures that const objects which appear identical in the syntax continue to
+canonicalize consistently across legacy and opted-in libraries.  Unlike the case
+of type literals, the choice to rewrite the type arguments is not arbitrary,
+since the outcome of constant instance checks and casts may depend on the legacy
+status of the type arguments.
+
+The Dart static analysis tool does not distinguish between sound and unsound
+checking mode, and hence it is expected that there will be some small level of
+infidelity in the constant evaluation semantics in the analyzer.  In practice,
+programs that observe object identity at compile time are almost non-existent,
+and so it may prove most pragmatic in the analyzer to always use the provided
+type arguments.
 
 ### Null check operator
 
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -1267,7 +1267,7 @@ canonicalize to is arbitrary in that placement of legacy modifiers in type
 literals is not otherwise observable in the language.
 
 Note that the choice of canonicalization for type literals does not depend
-directly on whether sound or unsound null safety semantics are in use.
+directly on whether sound or unsound null checking is in use.
 
 #### Constant instances
 
@@ -1297,10 +1297,11 @@ non-legacy type is used in the canonical representative.  For example:
 ```dart
 // null safe code.
 class C<T> {
+  final T x;
   void test(Object? o) {
-    assert(o is T);
+    o as T;
   }
-  const C(Object? o) : assert(o is T);
+  const C(Object? o) : x = o as T;
 }
 
 // If the canonical instance uses `int`, this is a compile time error
@@ -1333,16 +1334,15 @@ ensures that with sound null checking, the final consistent semantics are
 obeyed, since it is not observable which instance is chosen as the canonical
 representative in sound mode.
 
-With unsound null checking, all generic const constructors and generic const
-literals are additionally treated as if all type arguments passed to them were
-legacy types regardless of whether the constructed class was defined in a legacy
-library or not, and regardless of whether the constructor invocation or literal
-occured in a legacy library or not.  Specifically, a constructor invocation or
-object literal with generic type parameters `Ti` is treated as if the parameters
-were **LEGACY_TYPE**('Ti') as defined below. Implementations may choose to
-eagerly normalize the type arguments before applying the legacy rewrite, as
-desired.  This ensures that const objects which appear identical in the syntax
-continue to canonicalize consistently across legacy and opted-in libraries.
+With unsound null checking, all generic constant object expressions are
+additionally treated as if all type arguments passed to them were legacy types
+regardless of whether the constructed class was defined in a legacy library or
+not, and regardless of whether the constructor invocation or literal occured in
+a legacy library or not.  Specifically, a constant object expression with
+generic type parameters `Ti` is treated as if the parameters were
+**CONST_CANONICAL_TYPE**('Ti') as defined below.  This ensures that const
+objects which appear identical in the syntax continue to canonicalize
+consistently across legacy and opted-in libraries.
 
 The Dart static analysis tool does not distinguish between sound and unsound
 checking mode, and hence it is expected that there will be some small level of
@@ -1357,35 +1357,35 @@ to evaluate those in opted out libraries using unsound mode semantics.  Hence in
 the example above, the definition of `c1` would be a compile time error, but the
 definition of `c2` would not.
 
-The exact definition of the **LEGACY_TYPE**(`T`) erasure operation on types `T`
-used above is defined as follows.
+The **CONST_CANONICAL_TYPE**(`T`) erasure operation on types `T` used above is
+defined as follows.
 
-- **LEGACY_TYPE**(`T`) = `T` if `T` is `dynamic`, `void`, `Null`
-- **LEGACY_TYPE**(`T`) = `T*` if `T` is `Never` or `Object`
-- **LEGACY_TYPE**(`FutureOr<T>`) = `FutureOr<S>*`
-  - where `S` is **LEGACY_TYPE**(`T`)
-- **LEGACY_TYPE**(`T?`) =
-  - let `S` be **LEGACY_TYPE**(`T`)
+- **CONST_CANONICAL_TYPE**(`T`) = `T` if `T` is `dynamic`, `void`, `Null`
+- **CONST_CANONICAL_TYPE**(`T`) = `T*` if `T` is `Never` or `Object`
+- **CONST_CANONICAL_TYPE**(`FutureOr<T>`) = `FutureOr<S>*`
+  - where `S` is **CONST_CANONICAL_TYPE**(`T`)
+- **CONST_CANONICAL_TYPE**(`T?`) =
+  - let `S` be **CONST_CANONICAL_TYPE**(`T`)
   - if `S` is `R*` then `R?`
   - else `S?`
-- **LEGACY_TYPE**(`T*`) = **LEGACY_TYPE**(`T`)
-- **LEGACY_TYPE**(`X extends T`) = `X*`
-- **LEGACY_TYPE**(`X & T`) =
+- **CONST_CANONICAL_TYPE**(`T*`) = **CONST_CANONICAL_TYPE**(`T`)
+- **CONST_CANONICAL_TYPE**(`X extends T`) = `X*`
+- **CONST_CANONICAL_TYPE**(`X & T`) =
   - This case should not occur, since intersection types are not permitted as
     generic arguments.
-- **LEGACY_TYPE**(`C<T0, ..., Tn>`) = `C<R0, ..., Rn>*`
-  - where `Ri` is **LEGACY_TYPE**(`Ti`)
+- **CONST_CANONICAL_TYPE**(`C<T0, ..., Tn>`) = `C<R0, ..., Rn>*`
+  - where `Ri` is **CONST_CANONICAL_TYPE**(`Ti`)
   - Note this includes the case of an interface type with no generic parameters
     (e.g `int`).
-- **LEGACY_TYPE**(`R Function<X extends B>(S)`) = `F*`
+- **CONST_CANONICAL_TYPE**(`R Function<X extends B>(S)`) = `F*`
   - where `F = R1 Function<X extends B1>(S1)`
-  - and `R1` = **LEGACY_TYPE**(`R`)
-  - and `B1` = **LEGACY_TYPE**(`B`)
-  - and `S1` = **LEGACY_TYPE**(`S`)
+  - and `R1` = **CONST_CANONICAL_TYPE**(`R`)
+  - and `B1` = **CONST_CANONICAL_TYPE**(`B`)
+  - and `S1` = **CONST_CANONICAL_TYPE**(`S`)
   - Note, this generalizes to arbitrary number of type and term parameters.
 
-Note that if `T` is a normal form type, then **LEGACY_TYPE**(`T`) is also a
-normal form type.
+Note that if `T` is a normal form type, then **CONST_CANONICAL_TYPE**(`T`) is
+also a normal form type.
 
 
 ### Null check operator
--- a/accepted/future-releases/nnbd/feature-specification.md
+++ b/accepted/future-releases/nnbd/feature-specification.md
@@ -11,6 +11,8 @@ Status: Draft
   - Specify canonicalization of type literals.
   - Specify the unsound mode legacy rewrite of type arguments to const
     constructors and literals.
+  - Specify that constant object canonicalization is done with respect to normal
+    forms of generic type arguments.
 
 2020.11.09
   - Change terminology to use sound or unsound null checking rather than
@@ -1285,8 +1287,8 @@ modifiers.  We choose `R` arbitrarily as the type literal corresponding to one
 of **NORM**(`S1`) or **NORM**(`S2`) with all legacy type modifiers erased.
 
 The effect of this definition is to ensure that type literals which look
-indentical in the source syntax but which may differ by the presence of legacy
-type modifiers are canonicalized consistenly in the sense that any two type
+identical in the source syntax but which may differ by the presence of legacy
+type modifiers are canonicalized consistently in the sense that any two type
 literals which would compare equal via the definition of runtime type equality
 given above will be canonicalized to the same type object.  The choice of what
 type object to canonicalize to is arbitrary in that placement of legacy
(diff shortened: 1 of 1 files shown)


leafpetersen:
@johnniwinther @scheglov Can you please take a look and see what you think with respect to feasibility (comments on desirability also welcome)?  I'm aiming for something relatively low impact on the implementations.

cc @lrhn @eernstg @stereotype441 @alexmarkov @matanlurey 

This is an initial proposal for a fix to a variety of issues around consistent canonicalization of type literals and constant objects when running mixed mode programs.   Please take a look and let me know what you think.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1236,18 +1248,92 @@ the `LEGACY_ERASURE` of the types.
...
-library or not, and regardless of whether the constructor invocation or
-literal occurs in a legacy library or not.  This ensures that const objects
-continue to canonicalize consistently across legacy and opted-in libraries.
+Const evaluation is modified so that legacy and opted-in instances canonicalize
+more consistently as defined below, using the operation **LEGACY_TYPE**(`T`) on
+types `T` defined as follows.
+
+- **LEGACY_TYPE**(`T`) = `T` if `T` is `dynamic`, `void`, `Null`
+- **LEGACY_TYPE**(`T`) = `T*` if `T` is `Never` or `Object`
+- **LEGACY_TYPE**(`FutureOr<T>`) = `FutureOr<S>*`
```
scheglov, line 1257: It seems that the meaning of `S` is missing for this case.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1236,18 +1248,92 @@ the `LEGACY_ERASURE` of the types.
...
+- **LEGACY_TYPE**(`R Function<X extends B>(S)`) = `F*`
+  - where `F = R1 Function<X extends B1>(S1)`
+  - and `R1` = **LEGACY_TYPE**(`R`)
+  - and `B1` = **LEGACY_TYPE**(`B`)
+  - and `S1` = **LEGACY_TYPE**(`S`)
+
+Note that if `T` is a normal form type, then **LEGACY_TYPE**(`T`) is also a
+normal form type.
+
+#### Type literals
```
lrhn, line 1278: This section doesn't use **LEGACY_TYPE**, so maybe move it above the definition of **LEGACY_TYPE**, so that definition comes just before the section using it.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1236,18 +1248,92 @@ the `LEGACY_ERASURE` of the types.
...
+soundly null safe programs will never contain legacy types, and hence the
+erasure of legacy type modifiers is never necessary when running in sound mode.
+
+#### Constant instances
+
+In both sound and unsound null checking, canonicalization of generic object
+instances is performed relative to the normal forms of their generic type
+arguments.  That is, an instance of `C<T0>` canonicalizes to the same object as
+`C<T1>` if `T0` and `T1` have the same normal form (up to the identity of bound
+variables), and the objects are otherwise identical.
```
lrhn, line 1310: No "and ignoring legacy type modifiers"? Is that correct?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1236,18 +1248,92 @@ the `LEGACY_ERASURE` of the types.
...
+that with sound null checking, the final consistent semantics are obeyed.
+
+With unsound null checking, all generic const constructors and generic const
+literals are additionally treated as if all type arguments passed to them were
+legacy types regardless of whether the constructed class is defined in a legacy
+library or not, and regardless of whether the constructor invocation or literal
+occurs in a legacy library or not.  That is, a constructor invocation or object
+literal with generic type parameters `Ti` is treated as if the parameters were
+**LEGACY_TYPE**('Ti'). Implementations may choose to eagerly normalize the type
+arguments before applying the legacy rewrite, as desired.
```
lrhn, line 1323: (I assume that type aliases are always expanded eagerly).

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1236,18 +1248,92 @@ the `LEGACY_ERASURE` of the types.
...
+of type literals, the choice to rewrite the type arguments is not arbitrary,
+since the outcome of constant instance checks and casts may depend on the legacy
+status of the type arguments.
+
+The Dart static analysis tool does not distinguish between sound and unsound
+checking mode, and hence it is expected that there will be some small level of
+infidelity in the constant evaluation semantics in the analyzer.  In practice,
+programs that observe object identity at compile time are almost non-existent,
+and so it may prove most pragmatic in the analyzer to always use the provided
+type arguments.
```
lrhn, line 1336: I can see the problem.
```dart
// null safe code.
class C<T> {
  const C(Object o) : assert(o is T Function(T));
}
int? f(int? x) => x;
const breaks = C<int>(f); // Throws in sound mode, not in unsound mode.
```
I can't see a good solution. I don't like the analyzer doing having a decidedly wrong semantics (one which matches neither unsound mode nor sound mode), but since we are erring on the side of detecting problems in sound mode programs, it's probably the best available alternative in the long run. And statically checking identity of constants in potentially constant code is likely to be rare.


Maybe the analyzer could constant-evaluate null safe code twice, one in sound mode and one in unsound mode. It'll still look weird to have "potential, but not certain, compile-time error".

lrhn (approved):

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1236,18 +1248,92 @@ the `LEGACY_ERASURE` of the types.
...
+literals are additionally treated as if all type arguments passed to them were
+legacy types regardless of whether the constructed class is defined in a legacy
+library or not, and regardless of whether the constructor invocation or literal
+occurs in a legacy library or not.  That is, a constructor invocation or object
+literal with generic type parameters `Ti` is treated as if the parameters were
+**LEGACY_TYPE**('Ti'). Implementations may choose to eagerly normalize the type
+arguments before applying the legacy rewrite, as desired.
+
+This ensures that const objects which appear identical in the syntax continue to
+canonicalize consistently across legacy and opted-in libraries.  Unlike the case
```
eernstg, line 1326: It seems likely that we wish to ensure that a library _L1_ can be opted in, and a different library _L2_ can still be opted out, and _L1_ and _L2_ can contain constant expressions that have the same spelling and are canonicalized to the same object. In other words, they keep working if they are at first opted out, and then _L1_ is switched over to enable null safety.

Why, then, is it not important to avoid the breaking change (in _L1_ as well as _L2_) introduced by normalization in mixed version programs?

If this is because normalization is almost never used anyway then perhaps it is also not important to introduce it as part of canonicalization?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1236,18 +1248,92 @@ the `LEGACY_ERASURE` of the types.
...
-types), regardless of whether the constructed class is defined in a legacy
-library or not, and regardless of whether the constructor invocation or
-literal occurs in a legacy library or not.  This ensures that const objects
-continue to canonicalize consistently across legacy and opted-in libraries.
+Const evaluation is modified so that legacy and opted-in instances canonicalize
+more consistently as defined below, using the operation **LEGACY_TYPE**(`T`) on
+types `T` defined as follows.
+
+- **LEGACY_TYPE**(`T`) = `T` if `T` is `dynamic`, `void`, `Null`
+- **LEGACY_TYPE**(`T`) = `T*` if `T` is `Never` or `Object`
```
eernstg, line 1256: I suppose this case doesn't apply when `Never` occurs in a legacy library, because in that case `Never` simply means `Null` (in a constant expression, and anywhere else)?

eernstg (commented):
Added several comments. I suspect that we'll be more happy about an approach where normalization isn't applied. This is mainly because that's the approach we have used elsewhere, and it does create some issues, as described in the comments.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1236,18 +1248,92 @@ the `LEGACY_ERASURE` of the types.
...
-const literals are treated as if all type arguments passed to them were legacy
-types (both at the top level, and recursively over the structure of the
-types), regardless of whether the constructed class is defined in a legacy
-library or not, and regardless of whether the constructor invocation or
-literal occurs in a legacy library or not.  This ensures that const objects
-continue to canonicalize consistently across legacy and opted-in libraries.
+Const evaluation is modified so that legacy and opted-in instances canonicalize
+more consistently as defined below, using the operation **LEGACY_TYPE**(`T`) on
+types `T` defined as follows.
+
```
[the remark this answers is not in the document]
leafpetersen: > So, I read it as only being used for the _run-time values_, not the static types.
> Is this correct?

Correct.  I'll clarify that.



> It's used only on type arguments? (Type objects like the `int` in `<Type>[int]` are covered separately and doesn't use **LEGACY_TYPE**).

Correct.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: Fixed.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1236,18 +1248,92 @@ the `LEGACY_ERASURE` of the types.
...
+- **LEGACY_TYPE**(`T`) = `T` if `T` is `dynamic`, `void`, `Null`
+- **LEGACY_TYPE**(`T`) = `T*` if `T` is `Never` or `Object`
+- **LEGACY_TYPE**(`FutureOr<T>`) = `FutureOr<S>*`
+- **LEGACY_TYPE**(`T?`) =
+  - let `S` be **LEGACY_TYPE**(`T`)
+  - if `S` is `R*` then `R?`
+  - else `S?`
+- **LEGACY_TYPE**(`T*`) = **LEGACY_TYPE**(`T`)
```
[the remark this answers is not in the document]
leafpetersen: I don't think so. I'm taking the interpretation that `int` is a class type.  The invariant is that the result of LEGACY_TYPE(T) is always either dynamic, void, Null, `S?` for some `S` or `S*` for some `S`.  In your example, `LEGACY_TYPE(int*) == LEGACY_TYPE(int) == int*`.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1236,18 +1248,92 @@ the `LEGACY_ERASURE` of the types.
...
+- **LEGACY_TYPE**(`X extends T`) = `X*`
```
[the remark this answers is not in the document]
leafpetersen: True.  Fixed.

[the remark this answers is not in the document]
leafpetersen: Actually - not true. `X` could be a bound variable in a generic function type.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1236,18 +1248,92 @@ the `LEGACY_ERASURE` of the types.
...
+- **LEGACY_TYPE**(`X & T`) =
+  - This case should not occur, since intersection types are not permitted as
+    generic arguments.
+- **LEGACY_TYPE**(`C<T0, ..., Tn>`) = `C<R0, ..., Rn>*`
+  - where `Ri` is **LEGACY_TYPE**(`Ti`)
+- **LEGACY_TYPE**(`R Function<X extends B>(S)`) = `F*`
+  - where `F = R1 Function<X extends B1>(S1)`
+  - and `R1` = **LEGACY_TYPE**(`R`)
+  - and `B1` = **LEGACY_TYPE**(`B`)
+  - and `S1` = **LEGACY_TYPE**(`S`)
```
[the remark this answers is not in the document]
leafpetersen: Done.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: Done.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1236,18 +1248,92 @@ the `LEGACY_ERASURE` of the types.
...
+
+Note that if `T` is a normal form type, then **LEGACY_TYPE**(`T`) is also a
+normal form type.
+
+#### Type literals
+
+Two constant type literals `T1` and `T2` canonicalize to the same object if they
+are equal using the definition of runtime type equality specified above.  The
```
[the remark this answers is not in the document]
leafpetersen: > If we do decide to keep normalization to a minimum then we may be able to simply say that two types `T1` and `T2` are transformed into `**LEGACY_TYPE**(Tj)` for `j = 1` resp. `j = 2`, canonicalized to the same object iff the resulting types are identical.

I think you can avoid normalization this way, but since this is not specified as being specific to sound or unsound mode, we cannot (or at least should not) choose the legacy version of the type literal as the representative of the class.

I'm not sure what you mean by removing the arbitrary choice - the choice of instance to represent the equivalence class still seems arbitrary, regardless of whether you normalize.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1236,18 +1248,92 @@ the `LEGACY_ERASURE` of the types.
...
+object `R` which serves as the canonical form for `T1` and `T2` is defined as
+follows. Let `S1` and `S2` be the types to which `T1` and `T2` correspond.  Note
+that by assumption, **NORM**(`S1`) and **NORM**(`S2`) are syntactically
+identical up to the identity of bound variables and the placement of legacy type
+modifiers.  We choose `R` arbitrarily as the type literal corresponding to one
```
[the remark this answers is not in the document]
leafpetersen: Yes, I was being imprecise.  It's arbitrary which you choose as the representative, but you must choose the same representative for the entire equivalence class.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1236,18 +1248,92 @@ the `LEGACY_ERASURE` of the types.
...
+given above will be canonicalized to the same type object.  The choice of what
+type object to canonicalize to is arbitrary in that placement of legacy
+modifiers in type literals is not otherwise observable in the language.  Any
+consistent choice of placement of legacy modifiers is hence an equally valid
+implementation.
+
+Note that the choice of canonicalization for type literals does not depend
+directly on whether sound or unsound null safety semantics are in use.  However,
+soundly null safe programs will never contain legacy types, and hence the
+erasure of legacy type modifiers is never necessary when running in sound mode.
```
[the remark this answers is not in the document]
leafpetersen: line 1287.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1236,18 +1248,92 @@ the `LEGACY_ERASURE` of the types.
...
+
+#### Constant instances
+
+In both sound and unsound null checking, canonicalization of generic object
+instances is performed relative to the normal forms of their generic type
```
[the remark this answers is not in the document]
leafpetersen: Done.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1236,18 +1248,92 @@ the `LEGACY_ERASURE` of the types.
...
+literals are evaluated using the actual type arguments provided.  This ensures
+that with sound null checking, the final consistent semantics are obeyed.
+
+With unsound null checking, all generic const constructors and generic const
+literals are additionally treated as if all type arguments passed to them were
+legacy types regardless of whether the constructed class is defined in a legacy
+library or not, and regardless of whether the constructor invocation or literal
+occurs in a legacy library or not.  That is, a constructor invocation or object
+literal with generic type parameters `Ti` is treated as if the parameters were
+**LEGACY_TYPE**('Ti'). Implementations may choose to eagerly normalize the type
```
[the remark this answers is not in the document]
leafpetersen: `int?` remains `int?`.  So `<int?>[]` and `<int?>[]` canonicalize to the same object, which is different from `<int>[]`.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1236,18 +1246,92 @@ the `LEGACY_ERASURE` of the types.
...
+- **LEGACY_TYPE**(`X & T`) =
+  - This case should not occur, since intersection types are not permitted as
+    generic arguments.
+- **LEGACY_TYPE**(`C<T0, ..., Tn>`) = `C<R0, ..., Rn>*`
+  - where `Ri` is **LEGACY_TYPE**(`Ti`)
+- **LEGACY_TYPE**(`R Function<X extends B>(S)`) = `F*`
+  - where `F = R1 Function<X extends B1>(S1)`
+  - and `R1` = **LEGACY_TYPE**(`R`)
+  - and `B1` = **LEGACY_TYPE**(`B`)
+  - and `S1` = **LEGACY_TYPE**(`S`)
```
crelier, line 1273: Shouldn't `required` markers be removed from the signature? They are removed by LEGACY_ERASURE, which is informally specified in the [spec](https://github.com/dart-lang/language/blob/master/accepted/future-releases/nnbd/feature-specification.md#classes-defined-in-legacy-libraries). For clarity, shouldn't LEGACY_ERASURE be specified in the same way as the new LEGACY_TYPE?

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -6,6 +6,14 @@ Status: Draft
 
 ## CHANGELOG
 
+2020.12.30
+  - Remove the warning for overrides with different default values.
+  - Specify canonicalization of type literals.
+  - Specify the unsound mode legacy rewrite of type arguments to const
+    constructors and literals.
+  - Specify that constant object canonicalization is done with respect to normal
+    forms of generic type arguments.
```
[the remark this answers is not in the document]
leafpetersen: I think I would probably propose to just do this as a breaking change.  It won't affect anyone.  We have already made this breaking change for non const runtime type objects, by the way.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: This kind of analyzer infidelity in mixed mode has been baked in for a year plus, I'm not too worried about it.  In practice, I think users will be happy if the const evaluation is performed according to the library semantics: that is, in an opted in library, the above would throw, but if the `const breaks` definition were in an opted out library, it would not throw (since the type would be `int*`).  I actually think that from a behavior standpoint, this is superior to the CFE semantics.  Unfortunately, it is inconsistent with the separate desire to have two instances of `<int>[]` compare as identical, even if one appears in an opted in library and one does not.  So I'm proposing to give the CFE the consistent identity semantics, but permit the analyzer to be looser with identity (which is basically never observed at compile time in practice) but relatively strict with evaluation semantics.  Though actually, after discussions with @scheglov and @bwilkerson today, I believe that the analyzer can implement accurate identity semantics.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: @crelier I don't think it should.  Informally, my goal is that two constant type literals which are "spelled the same" should compare identical.  More formally, my goal is that two constant type literals should compare identical iff they are equal according to the runtime type equality definition specified just above this.  That definition does ignore `*` modifiers, but does not ignore `required` markers.   That was the idea anyway, do you see issues with it?  It does bother me a bit that I have two definitions  (LEGACY_TYPE and LEGACY_ERASURE) that are so close but not the same.  

And yes, I should have specified LEGACY_ERASURE precisely too, if I get a chance I'll follow up.

leafpetersen:
I took another whack at this based on discussion with @scheglov, @bwilkerson and @stereotype441 today, as well as incorporating feedback above.  Please take another look, and let's plan on discussing this in the language team meeting tomorrow.

accepted/future-releases/nnbd/feature-specification.md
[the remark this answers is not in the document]
eernstg: Ah yes, we've had that behavior since `dart-sdk-2.8.0-dev.17.0`, even a 100% legacy program will use normalization during canonicalization. I hadn't thought about that. So it was a breaking change that didn't actually break anything. Amazing, there are many lines of Dart code out there! ;-)

In that case I certainly think it's reasonable to keep using normalization.

We might even be in a situation where normalization cannot be observed in very many other cases, and that might allow us to use normalization more aggressively. We might then be able to think "normalization occurs whenever possible" rather than "normalization occurs only in a few specific situations".

We may then be able to apply some specialized optimizations. Perhaps each normalized run-time type representation will make subtype checks faster, perhaps other optimizations will be allowed only when _every_ type is normalized.

Nice!

accepted/future-releases/nnbd/feature-specification.md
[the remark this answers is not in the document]
eernstg: Ah, `n` can be minus one in the `LEGACY_TYPE(C<T0..Tn>)` rule. OK!

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1236,18 +1248,92 @@ the `LEGACY_ERASURE` of the types.
...
+directly on whether sound or unsound null safety semantics are in use.  However,
+soundly null safe programs will never contain legacy types, and hence the
+erasure of legacy type modifiers is never necessary when running in sound mode.
+
+#### Constant instances
+
+In both sound and unsound null checking, canonicalization of generic object
+instances is performed relative to the normal forms of their generic type
+arguments.  That is, an instance of `C<T0>` canonicalizes to the same object as
+`C<T1>` if `T0` and `T1` have the same normal form (up to the identity of bound
```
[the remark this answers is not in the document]
eernstg: (We have normalization even for legacy programs already, no breaking change)

accepted/future-releases/nnbd/feature-specification.md
eernstg: Why do we not require eager normalization? It will occur at compile-time, so there's no run-time cost, and it seems to be a useless source of ambiguity.

If the outcome is provably the same as an approach where normalization is not performed eagerly, each implementation is of course free to use such an approach. This would allow the reader of the code to reason about canonicalization based on a simpler mental model than the one which is needed if normalization is optional.

accepted/future-releases/nnbd/feature-specification.md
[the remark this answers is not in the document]
eernstg: (Could resolve this thread, but I'll keep it open for now because it seems important to be aware of this fact.)

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1236,18 +1248,92 @@ the `LEGACY_ERASURE` of the types.
...
+variables), and the objects are otherwise identical.
 
 With sound null checking, all generic const constructors and generic const
-literals are evaluated using the actual type arguments provided, whether legacy
-or non-legacy.  This ensures that with sound null checking, the final
-consistent semantics are obeyed.
+literals are evaluated using the actual type arguments provided.  This ensures
+that with sound null checking, the final consistent semantics are obeyed.
```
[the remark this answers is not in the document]
eernstg: In order to avoid that "2nd breaking change", we will need to specify that normalization is applied during canonicalization for a program with pure null safety. Surely we don't want to change `<int>[]` to `<int*>[]` when it occurs in a pure program, so we'll stop using `**LEGACY_TYPE**` and start using some other transformation when we switch to sound null checking. That other transformation might simply be normalization, because `**LEGACY_TYPE**(T) == T` whenever `T` is a type in a pure program (it doesn't contain any `*`, and it doesn't contain any type alias applications that denote a type that contains a `*`).

accepted/future-releases/nnbd/feature-specification.md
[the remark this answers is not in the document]
eernstg: > `int?` remains `int?`. So `<int?>[]` and `<int?>[]` canonicalize to the same object, which is different from `<int>[]`.

I'm referring to the wording: Noting that `int?` remains `int?`, it does not make sense to say that we are treating `int?` as if it were a legacy type (line 1318).

accepted/future-releases/nnbd/feature-specification.md
eernstg: (Resolving: Normalization is not a new behavior, we've had it since 2.8.0-dev.17.0, hence there is no breakage).

accepted/future-releases/nnbd/feature-specification.md
crelier: @leafpetersen Thanks for the clarification! I was just checking, because the VM currently applies LEGACY_ERASURE on constants when  reading them in, and therefore removes the markers. My understanding is that the VM will now apply LEGACY_TYPE instead, so the markers will now stay.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1248,13 +1248,122 @@ the `LEGACY_ERASURE` of the types.
...
+  const C(Object? o) : assert(o is T);
+}
+
+// If the canonical instance uses `int`, this is a compile time error
+// If the canonical instance uses `int*`, this is not a compile time error
+const c1 = C<int>(null);
+
+// If the canonical instance uses `int`, this throws
+// If the canonical instance uses `int*`, this does not throw
+void test1() => c1.test(null);
```
crelier, line 1312: I would declare a valid `c2` constant in this second example, since `c1` does not compile, e.g. `const c2 = C<int>(0);`

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1248,13 +1248,122 @@ the `LEGACY_ERASURE` of the types.
...
+representative in sound mode.
+
+With unsound null checking, all generic const constructors and generic const
+literals are additionally treated as if all type arguments passed to them were
+legacy types regardless of whether the constructed class was defined in a legacy
+library or not, and regardless of whether the constructor invocation or literal
+occured in a legacy library or not.  Specifically, a constructor invocation or
+object literal with generic type parameters `Ti` is treated as if the parameters
+were **LEGACY_TYPE**('Ti') as defined below. Implementations may choose to
+eagerly normalize the type arguments before applying the legacy rewrite, as
```
eernstg, line 1343: I believe normalization here is intended to be an optimization opportunity, and it should not be observable because it only affects the creation of a (canonical) type object, and there is no way (other than `.toString()`) to detect whether normalization occurred (because normalization is used for `Type.==`, and `Type.==` is used to determine canonicalization).

It would then be helpful to make this sentence commentary, and mention that it is not observable: It would not make any difference if this sentence were omitted, so it shouldn't be normative text.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1248,13 +1248,122 @@ the `LEGACY_ERASURE` of the types.
...
+choosing a canonical representative for each equivalence class.  However, the
+lack of a canonical representative is observable at compile time in rare cases,
+such as the example shown above.  We propose that the analyzer should choose to
+evaluate those constants in opted in libraries using sound mode semantics, and
+to evaluate those in opted out libraries using unsound mode semantics.  Hence in
+the example above, the definition of `c1` would be a compile time error, but the
+definition of `c2` would not.
+
+The exact definition of the **LEGACY_TYPE**(`T`) erasure operation on types `T`
+used above is defined as follows.
```
eernstg, line 1361: Maybe delete 'exact definition of the'.

eernstg (commented):
Added a few comments.

I still think that `**LEGACY_TYPE**` is a highly surprising name for a function that may return `int?`, couldn't we rename it, maybe to something like `**CANONICAL_TYPE**`? ;-)

The treatment of `Never` in a legacy library is another source of confusion. See the comments.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: Probably so.

accepted/future-releases/nnbd/feature-specification.md
leafpetersen: No, we ignore legacy type modifiers.

accepted/future-releases/nnbd/feature-specification.md
```diff
@@ -1236,18 +1248,92 @@ the `LEGACY_ERASURE` of the types.
...
 
-With unsound null checking, all generic const constructors and generic
-const literals are treated as if all type arguments passed to them were legacy
-types (both at the top level, and recursively over the structure of the
-types), regardless of whether the constructed class is defined in a legacy
-library or not, and regardless of whether the constructor invocation or
-literal occurs in a legacy library or not.  This ensures that const objects
-continue to canonicalize consistently across legacy and opted-in libraries.
+Const evaluation is modified so that legacy and opted-in instances canonicalize
+more consistently as defined below, using the operation **LEGACY_TYPE**(`T`) on
```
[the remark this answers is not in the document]
leafpetersen: Change the name to CONST_CANONICAL_TYPE.

accepted/future-releases/nnbd/feature-specification.md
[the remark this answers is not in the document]
leafpetersen: I am saying, we treat `int?` as a legacy type, and thereupon immediately defining what it means to treat `int?` as a legacy type.  Not sure how to be more explicit.

leafpetersen:
Landed the core of this.  If there are more tweaks desired, I'll follow up.

leafpetersen:
Proposed tests for the type literal canonicalization portion [here](https://dart-review.googlesource.com/c/sdk/+/178929).

PR #1400, opened 2021-01-13
Eliminate the static scope of a class

### LINKED ISSUE #550: Specification: Eliminate the distinction between the static and instance scope of a class?
The language specification says (and has basically always said) that a class has the scope structure library scope > type-parameter scope > static scope > instance scope, where static/instance members live in the static/instance scope, respectively:

> A class has several scopes:
>
> - A type-parameter scope which is empty if the class is not generic.
>   The enclosing scope of the type-parameter scope of a class is the enclosing scope
>   of the class declaration.
> - A static scope. The enclosing scope of the static scope of a class is
>   the type parameter scope of the class.
> - An instance scope. The enclosing scope of a class' instance scope is the 
>   class' static scope.
>
> The enclosing scope of an instance member declaration is
> the instance scope of the class in which it is declared.
> 
> The enclosing scope of a static member declaration is
> the static scope of the class in which it is declared.

Surprisingly, constructors do not live in any scopes (I can't find an answer to that question in the specification, except that the question is raised in a couple of old comments). So that's one little thing to fix.

However, and more seriously: This implies that an instance member is simply not in scope inside a static declaration. In particular, it implies that the following program has no errors:

```dart
int x;

class A {
  String x;
  static foo() => x.isEven; // OK!
}
```

We have special rules about the treatment of names of the form `x` and `x=`, but in this case that makes no difference: As seen from the static scope of `A` there is only one declaration of `x` (and none named `x=`), and that's the top-level variable.

However, the common front end and the analyzer are (of course ;-) not happy about this program. First the response from the analyzer, commit 0c06fbd88fed6943959192932ca480d73db19377:

```
ERROR|COMPILE_TIME_ERROR|INSTANCE_MEMBER_ACCESS_FROM_STATIC|/usr/local/google/home/eernst/lang/dart/scratch/201908/n035.dart|5|19|1|Instance members can't be accessed from a static method.
ERROR|STATIC_TYPE_WARNING|UNDEFINED_GETTER|/usr/local/google/home/eernst/lang/dart/scratch/201908/n035.dart|5|21|6|The getter 'isEven' isn't defined for the class 'String'.
```

Similarly for `dart` from the same commit:

```
n035.dart:5:19: Error: Getter not found: 'x'.
  static foo() => x.isEven; // OK!
                  ^
```

I have no idea how the CFE arrives at the conclusion that there are too few getters around, but the point is that the currently specified scope structure is not what the implementations are using.

It seems much more reasonable to say that the tools are merging the static scope and the instance scope into one scope, and then there is a special rule that instance members cannot be accessed using the scope rules (but we of course can do things like `e.x` when `e` has type `A`).

So we have some freedom to change the specified scope structure, and still consider this to be a non-breaking change.

So here's a proposal:

- We eliminate the notion of 'the static scope' and 'the instance scope' of a class.
- We introduce the notion of 'the body scope' of a class, and let all static members, instance members, and constructors live in the body scope.
- Special rules are already in place in #548, ensuring that it is an error to access an instance member using implicit instance member access (such as an identifier reference or assignment, `myInstanceVariable`/`myInstanceVariable = "Hest"`; or an unqualified invocation `myInstanceMethod(42)`) from any location which is not in the body of an instance member or generative constructor (soon: or an initializing expression of a `late` instance variable).

@munificent, @lrhn, @leafpetersen, WDYT?

eernstg:
The static scope of a class is not implemented, and #550 gathered support for eliminating it. This means that all of the constructors, static members, and instance members have the new 'class scope' of the class as their current scope, which is already matched by the current behavior of the analyzer and the common front end.

Cf. https://github.com/dart-lang/sdk/issues/38588 is an example where it actually matters.

Resolves #550.
Resolves https://github.com/dart-lang/sdk/issues/38588.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -2126,22 +2126,25 @@ The \IndexCustom{members}{members} of a class
 are its static and instance members.
 
 \LMHash{}%
-A class has several scopes:
+A class declaration has two scopes:
 \begin{itemize}
-\item A \IndexCustom{type-parameter scope}{scope!type parameter},
+\item
+  A \IndexCustom{type-parameter scope}{scope!type parameter},
   which is empty if the class is not generic (\ref{generics}).
-The enclosing scope of the type-parameter scope of a class is the enclosing scope of the class declaration.
-\item A \IndexCustom{static scope}{scope!static}.
-The enclosing scope of the static scope of a class is the type parameter scope (\ref{generics}) of the class.
-\item An \IndexCustom{instance scope}{scope!instance}.
-The enclosing scope of a class' instance scope is the class' static scope.
+  The enclosing scope of the type-parameter scope of a class declaration is
+  the current scope of the class declaration
+  (\commentary{which is the library scope of the current library}).
+\item
+  A \IndexCustom{class scope}{scope!class}.
+  The enclosing scope of the class scope of a class declaration is
+  the type parameter scope of the class declaration.
 \end{itemize}
 
 \LMHash{}%
-The enclosing scope of an instance member declaration is the instance scope of the class in which it is declared.
-
-\LMHash{}%
-The enclosing scope of a static member declaration is the static scope of the class in which it is declared.
+The enclosing scope of an instance member declaration,
+a static member declaration,
+or a constructor declaration is
+the class scope of the class in which it is declared.
 
 \LMHash{}%
 The current instance
@@ -3076,17 +3079,29 @@ We may also omit such a substitution when the given context is
 the instance scope of $C$, where $X_1, \ldots, X_m$ are in scope.
 }
 
-\commentary{
+\commentary{%
 A constructor declaration may conflict with static member declarations
 (\ref{classMemberConflicts}).
 }
 
-% In what scope do constructors go? The simple names of named constructors go in the static scope of the class. Unnamed ones go nowhere, but we use the class name to refer to them; the class name could also in the static scope of the class as well to prevent weird errors, or we could ban it explicitly and avoiding duplication. Similarly, the instance scope could contain the constructor names and class name, or we could have special rules to prevent collisions between instance members and constructors or the class.
-
-% The enclosing scope of a generative constructor is the instance scope of the class in which it is declared (but what about redirecting?)
+\commentary{%
+A constructor declaration does not introduce a name into a scope.
+The resolution of an instance creation expression
+(\ref{new})
+to a particular constructor
+relies on the library scope to determine the class
+(possibly via an import prefix),
+and for a constructor denoted by \code{$C$.\id} or \code{$prefix$.$C$.\id},
+it is directly specified that the class $C$ must declare
+a constructor with the name \code{$C$.\id}.
+In other words, the constructor is never determined via a lexical lookup,
+only the class.%
+}
 
 \LMHash{}%
-If{}f no constructor is specified for a class $C$, it implicitly has a default constructor \code{C() : \SUPER{}() \{\}}, unless $C$ is class \code{Object}.
+If{}f no constructor is specified for a class $C$,
+it implicitly has a default constructor \code{C():\ \SUPER()\ \{\}},
+unless $C$ is the built-in class \code{Object}.
 
 
 \subsubsection{Generative Constructors}
@@ -3523,8 +3538,6 @@ is a constructor prefaced by the built-in identifier
   \CONST? \FACTORY{} <constructorName> <formalParameterList>
 \end{grammar}
 
-%The enclosing scope of a factory constructor is the static scope \ref{} of the class in which it is declared.
-
 \LMHash{}%
 The return type of a factory whose signature is of the form \FACTORY{} $M$ or the form \FACTORY{} \code{$M$.\id} is $M$ if $M$ is not a generic type;
 otherwise the return type is \code{$M$<$T_1, \ldots,\ T_n$>} where $T_1, \ldots, T_n$ are the type parameters of the enclosing class.
@@ -3870,13 +3883,9 @@ for class \code{Object}.
 <mixins> ::= \WITH{} <typeNotVoidList>
 \end{grammar}
 
-%The superclass clause of a class C is processed within the enclosing scope of the static scope of C.
-%\commentary{
-%This means that in a generic class, the type parameters of the generic are available in the superclass clause.
-%}
-
 \LMHash{}%
-The scope of the \EXTENDS{} and \WITH{} clauses of a class $C$ is the type-parameter scope of $C$.
+The scope of the \EXTENDS{} and \WITH{} clauses of a class $C$ is
+the type-parameter scope of $C$.
 
 \LMHash{}%
 It is a compile-time error if the type in the \EXTENDS{} clause of a class $C$ is
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -3076,7 +3076,7 @@ $[T_1/X_1, \ldots, T_m/X_m]F$,
 where $X_j, j \in 1 .. m$ are the formal type parameters of $C$
 and $T_j, j \in 1 .. m$ are specified in the given context.
 We may also omit such a substitution when the given context is
-the instance scope of $C$, where $X_1, \ldots, X_m$ are in scope.
+the class scope of $C$, where $X_1, \ldots, X_m$ are in scope.
 }
 
 \commentary{%
@@ -14044,7 +14044,7 @@ it is a compile-time error if $D$ is an instance member,
 and the interface of the enclosing class does not have a member named $n$.
 When $\ell$ does not have access to \THIS,
 it is a compile-time error if $D$ is an instance member.
-% An instance variable initializer is in the instance scope, so it is
+% An instance variable initializer is in the class scope, so it is
 % actually possible for $D$ to be an instance member here.
 \EndCase
 
@@ -19294,13 +19294,18 @@ Comments may nest.
 \LMHash{}%
 \IndexCustom{Documentation comments}{documentation comments}
 are comments that begin with the tokens \code{///} or \code{/**}.
-Documentation comments are intended to be processed by a tool that produces human readable documentation.
+Documentation comments are intended to be processed by
+a tool that produces human readable documentation.
 
 \LMHash{}%
-The scope of a documentation comment immediately preceding the declaration of a class $C$ is the instance scope of $C$.
+The scope of a documentation comment immediately preceding
+the declaration of a class $C$ is
+the class scope of $C$.
 
 \LMHash{}%
-The scope of a documentation comment immediately preceding the declaration of a function $f$ is the scope in force at the very beginning of the body of $f$.
+The scope of a documentation comment immediately preceding
+the declaration of a function $f$ is
+the current scope at the very beginning of the body of $f$.
 
 
 \subsection{Operator Precedence}
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -3082,17 +3082,18 @@ A constructor declaration may conflict with static member declarations
 
 \commentary{%
 A constructor declaration does not introduce a name into a scope.
-The resolution of an instance creation expression
-(\ref{new})
-to a particular constructor
-relies on the library scope to determine the class
-(possibly via an import prefix),
-and for a constructor denoted by \code{$C$.\id} or \code{$prefix$.$C$.\id},
-it is directly specified that the class $C$ must declare
-a constructor with the name \code{$C$.\id}.
-In other words, the constructor is never determined via a lexical lookup,
-only the class
-(\ref{functionExpressionInvocation}).%
+If a function expression invocation
+(\ref{functionExpressionInvocation})
+or an instance creation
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -2126,22 +2126,25 @@ \section{Classes}
 are its static and instance members.
 
 \LMHash{}%
-A class has several scopes:
+A class declaration has two scopes:
```
lrhn, line 2129: Maybe change "has" to "introduces"?

specification/dartLangSpec.tex
```diff
@@ -2126,22 +2126,25 @@ \section{Classes}
...
+\item
+  A \IndexCustom{type-parameter scope}{scope!type parameter},
   which is empty if the class is not generic (\ref{generics}).
-The enclosing scope of the type-parameter scope of a class is the enclosing scope of the class declaration.
-\item A \IndexCustom{static scope}{scope!static}.
-The enclosing scope of the static scope of a class is the type parameter scope (\ref{generics}) of the class.
-\item An \IndexCustom{instance scope}{scope!instance}.
-The enclosing scope of a class' instance scope is the class' static scope.
+  The enclosing scope of the type-parameter scope of a class declaration is
+  the current scope of the class declaration
```
lrhn, line 2135: The "current scope" confuses me here&mdash;"current" suggests variability over time. 
Maybe: "the scope containing the class declaration"?

specification/dartLangSpec.tex
```diff
@@ -2126,22 +2126,25 @@ \section{Classes}
...
+  The enclosing scope of the class scope of a class declaration is
+  the type parameter scope of the class declaration.
 \end{itemize}
 
 \LMHash{}%
-The enclosing scope of an instance member declaration is the instance scope of the class in which it is declared.
-
-\LMHash{}%
-The enclosing scope of a static member declaration is the static scope of the class in which it is declared.
+The enclosing scope of an instance member declaration,
```
lrhn, line 2144: We use "enclosing scope" about a scope, meaning the an *outer scope*, above, and here we use "enclosing scope" of a declaration to be the scope containing the declaration. That's confusing.

If we had done the same above, we could say
"The enclosing scope of the type-parameter scope of a class declaration is the enclosing scope of the class declaration" (which I first tried as an alternative to "current scope", and then found quite unreadable).

Could we use "enclosing scope" only for scopes (scopes are nested) and "containing scope" for declarations (each declaration belongs to exactly one scope)?

specification/dartLangSpec.tex
```diff
@@ -3073,20 +3076,32 @@ \subsection{Constructors}
...
+\commentary{%
 A constructor declaration may conflict with static member declarations
 (\ref{classMemberConflicts}).
 }
 
-% In what scope do constructors go? The simple names of named constructors go in the static scope of the class. Unnamed ones go nowhere, but we use the class name to refer to them; the class name could also in the static scope of the class as well to prevent weird errors, or we could ban it explicitly and avoiding duplication. Similarly, the instance scope could contain the constructor names and class name, or we could have special rules to prevent collisions between instance members and constructors or the class.
-
-% The enclosing scope of a generative constructor is the instance scope of the class in which it is declared (but what about redirecting?)
+\commentary{%
+A constructor declaration does not introduce a name into a scope.
```
lrhn, line 3088: (Not sure that was a good idea, but it is what it is).

specification/dartLangSpec.tex
```diff
@@ -3073,20 +3076,32 @@ \subsection{Constructors}
...
+a constructor with the name \code{$C$.\id}.
+In other words, the constructor is never determined via a lexical lookup,
+only the class.%
+}
 
 \LMHash{}%
-If{}f no constructor is specified for a class $C$, it implicitly has a default constructor \code{C() : \SUPER{}() \{\}}, unless $C$ is class \code{Object}.
+If{}f no constructor is specified for a class $C$,
+it implicitly has a default constructor \code{C():\ \SUPER()\ \{\}},
+unless $C$ is the built-in class \code{Object}.
```
lrhn, line 3104: > ... is one of the built-in classes \code{Object} or \code{Null}.

(If we are post-null-safety).

specification/dartLangSpec.tex
```diff
@@ -3073,20 +3076,32 @@ \subsection{Constructors}
...
+The resolution of an instance creation expression
+(\ref{new})
+to a particular constructor
+relies on the library scope to determine the class
+(possibly via an import prefix),
+and for a constructor denoted by \code{$C$.\id} or \code{$prefix$.$C$.\id},
+it is directly specified that the class $C$ must declare
+a constructor with the name \code{$C$.\id}.
+In other words, the constructor is never determined via a lexical lookup,
+only the class.%
```
lrhn, line 3098: This feels weird (and wrong) because we don't know whether `Foo.id()` (no `new`) *is* an instance creation expression without first checking whether `Foo.id` is a constructor or a static member. 

Does "instance creation expression" only refer to `new C.id(...)` and `const C.id(...)`, not the expressions without the `new`/`const`? If so, the text here doesn't handle the most common case, so it doesn't add much v alue.
If not, then it's wrong that the expression `C.id(..)` must denote a constructor, it could be a static function.

That is, we can't really say that an instance creation expression without a `new` *must* denote a constructor, because it isn't an instance creation expression if it doesn't denote a constructor. The `must` requirement is tautologically true.
(We don't "do lookup", we just "look up" whether `Foo` has a `Foo.id` constructor.)

Maybe be more elaborate:

> A constructor declaration does not introduce a name into a scope.
> The resolution of an instance creation expression with a \NEW{} (\ref{new}) or \CONST{} (\ref{const})
> to a particular constructor
> relies on the library scope to determine the class
> (possibly via an import prefix),
> and for a constructor denoted by \code{$C$.\id} or \code{$prefix$.$C$.\id},
> it is directly specified that the class $C$ must declare
> a constructor with the name \code{$C$.\id}.
> An invocation of the form \code{$C$.\id(...)} or \code{$prefix}$.$C$.\id(...)} 
> first resolves $C$ or \code{$prefix$.$C$}, and if that resolves to a class,
> it directly checks whether that class declares a constructor with the name \code{$C$.\id}
> (or whether it perhaps has a static member instead).
> In other words, the constructor is never determined via a \emph{lexical} lookup, only the class.%

specification/dartLangSpec.tex
```diff
@@ -19285,13 +19294,18 @@ \subsubsection{Comments}
...
-The scope of a documentation comment immediately preceding the declaration of a class $C$ is the instance scope of $C$.
+The scope of a documentation comment immediately preceding
+the declaration of a class $C$ is
+the class scope of $C$.
 
 \LMHash{}%
-The scope of a documentation comment immediately preceding the declaration of a function $f$ is the scope in force at the very beginning of the body of $f$.
+The scope of a documentation comment immediately preceding
+the declaration of a function $f$ is
+the current scope at the very beginning of the body of $f$.
```
lrhn, line 19308: Actually not true since local variables are in scope in the entire body. It's the scope just *outside* the body. Is that the parameter scope?

specification/dartLangSpec.tex
eernstg: Done

specification/dartLangSpec.tex
eernstg: True, it sounds like a temporal association, but it is the standard phrase that we've always used to denote the innermost scope that contains a given syntactic entity.

In this case we don't have to use that phrase, because the current scope for a class declaration is always the library scope. So I just changed it to directly refer to the library scope, normatively.

This breaks if we introduce nested classes, of course, but we'll probably be able to fix it again if needed.

specification/dartLangSpec.tex
eernstg: That's a mistake, it should be 'The current scope'.

> Could we use "enclosing scope" only for scopes 

Yes, that's how it should be. Changed 13 locations (a few of them were obsolete comments that I just deleted) to make it so.

We may wish to change 'current scope' to some other phrase, but this at least makes the text more consistent such that we can change 'current scope' more reliably, if we choose to do that.

I did not change a couple of locations where it talks about 'enclosing scopes' (of a term). That's taken to mean 'the current scope of the term and its enclosing scopes'. We could change that, too, but I haven't done that.

specification/dartLangSpec.tex
eernstg: Yes, I think this is a rule which is consistent with the actual treatment of constructors that we've had all the time. I don't think it's particularly problematic, but it might need to be considered an exception by 'dart:mirrors'.

specification/dartLangSpec.tex
eernstg: > we don't know whether `Foo.id()` (no `new`) is an instance 
> creation expression without first checking whether `Foo.id` is
> a constructor or a static member.

The omitted `new`/`const` is handled exactly like that: We add a `new` or a `const` keyword to `Foo.id()` when `Foo.id` denotes a constructor, not when it is a static member, cf. https://github.com/dart-lang/language/blob/6df1612f667f7fdd28dd3b66a89e06b75ea98d2a/specification/dartLangSpec.tex#L10947.

In that step, we just look up the class `$C$`, and then we require explicitly that the declaration of `$C$` contains a declaration of a constructor named `$C$` respectively `$C$.\id`.

So we don't need to worry about how we handle an instance creation of the form `Foo.id()` because it's gone after desugaring, and the desugaring itself looks up the class, not the constructor.

I believe that the elaborated form gives information which is already given in other sections, so I added a reference to \ref{functionExpressionInvocation} at the end of this commentary.

specification/dartLangSpec.tex
eernstg: Thanks! I added a TODO comment to do that.

specification/dartLangSpec.tex
eernstg: Argh, this is very old text, present when the spec was added to the SDK repo. ;-)

I guess the intention is that parameters (obviously) must be in scope, but local variables in the function are not. There is no restriction in a DartDoc comment that it can't talk about instance members, so there's no need to worry about the special scope for initializing formals in a constructor, so the formal parameter scope is fine. Using that now!

eernstg (commented):
One issue remains: I haven't changed all occurrences of `current scope` to anything else ("the innermost enclosing scope", perhaps). We have 16 of them, now that a few have been added as a result of corrections of 'enclosing scope' of non-scopes.

specification/dartLangSpec.tex
lrhn: Not sure it affects `dart:mirrors`. Mirrors are about declarations, not scopes. 
(I see "scopes" as lexical constructs used to look up single identifiers. Since constructors have names that are not single identifiers, it makes sense that they are not in any scope. We also do non-scope lookup on static accesses like `Function.apply`, we look for static members in the declaration of the *class*, not names in a scope introduced by the class. The exception is import prefix "scopes", which are named scopes accessed by a qualified identifier instead of a single identifier).

specification/dartLangSpec.tex
lrhn: Desugaring cannot look up something, like $C$, before scopes exist, so it still feels to me like this text skips an important part - the part where we recognize constructors as constructors before adding the `new`.

It might just be the phrasing

Maybe:

> If an expression denotes a constructor, either as \code{$C$.\id} or \code{$prefix$.$C$.\id},
> or as $C$ or \code{$prefix$.$C$} in an invocation context,
> resolution relies on the library scope to determine the class (possibly via an import prefix),
> and then the class declaration is directly checked for whether it has a constructor named \code{$C$.\id}
> or $C$. It is not possible for an identifier to directly refer to a constructor, since the constructor is not
> in any scope used for resolving identifiers.

lrhn (approved):

specification/dartLangSpec.tex
eernstg: I agree on those things.

specification/dartLangSpec.tex
eernstg: I like that! Now using the new 'Maybe:' text, with a few adjustments.

> Desugaring cannot look up something, like `$C$`, before scopes exist,

It's not important to me that we call this step desugaring, so if we don't want to use that term to denote a transformation that has access to name resolution (or types, or any other non-context-free properties) then we'd just use a different name.

However, we have lots of transformations where source code is changed, and the changes are influenced by information from static analysis. For example, the transformation that defines the semantics of a for-in statement needs the static type of `$e$`, the transformation that transforms `id` to `this.id` needs the interface of the enclosing class, etc.

I know that Johnni is very worried about every piece of syntactic sugar in the specification (and I'm worried as well, whenever it may turn out to be imprecise), but I do believe that we can maintain a reasonably well-defined notion of source code transformations that are able to use information from static analysis. The crucial point is probably that we can't allow those transformations to change the static analysis of non-transformed expressions. With that, I think it's OK to use this sparingly and cautiously.

MrBirb:
@eernstg Any thoughts on when the spec will be finalized?

eernstg:
The spec will continue to evolve because the language evolves. But I'd very much like to release a spec which is considered stable and which is a specification of Dart 2.11 (the last version without null safety), and henceforth the bleeding edge spec will include null safety. I hope the updates and reviews for the new stable spec will be finished in Q1.

PR #1417, opened 2021-01-25
Add feature specification and implementation plan for small features for Q1 of 2021

lrhn:


accepted/future-releases/small-features-21Q1/feature-specification.md
```diff
@@ -0,0 +1,109 @@
+# Dart "Small Features" 21Q1
+
```
leafpetersen, line 2: Add a CHANGELOG at the top?  I found it very useful with the null safety feature spec.

accepted/future-releases/small-features-21Q1/feature-specification.md
```diff
@@ -0,0 +1,109 @@
...
+The language disallows generic function types as type arguments.
+
+```dart
+List<T Function<T>(T)> idFunctions; // INVALID
+var callback = [foo<T>(T value) => value]; // Inferred as above, then invalid.
+```
+
+We remove that restriction, so a type argument *can* be a generic function type.
+
+This requires no new syntax, and in some cases only the removal of a single check. There might be some platforms where the implementation currently assumes that generic function types cannot occur as the value of type variables (an proof-of-concept attempt hit an assert in the VM). Such assumptions will need to be flushed out with tests and fixed.
```
leafpetersen, line 75: @crelier Do you have any concerns about this?

accepted/future-releases/small-features-21Q1/implementation-plan.md
```diff
@@ -0,0 +1,153 @@
...
+
+For generic metadata, the analyzer needs to represent the *source*, and therefore must be able to make the distinction between an omitted-and-inferred type argument and an explicit type argument on a metadata constructor invocation. This is important for other source-manipulating tools like the formatter. The AST for metadata invocations need to have a place to store the type arguments, which isn't there now.
+
+If the analyzer assumes, and relies on the assumption, that type arguments cannot be generic function types, then it might need to fix that.
+
+The analyzer also needs to support `const Symbol(">>>")`, but otherwise seems to support the `>>>` feature (`operator>>>`  and `#>>>` works).
+
+Otherwise the analyzer is unlikely to need significant change for any of the features.
+
+## Phase 2 (Tool Implementation)
```
leafpetersen, line 58: We should have sections for dartdoc and IDE tooling.  

@bwilkerson @devoncarew @jcollins-g  We talked about making space for you to plan out IDE tooling work (quick fixes, etc), do you want to add some content here in a follow up, or just have it noted here (with a tracking issue filed?).

accepted/future-releases/small-features-21Q1/implementation-plan.md
```diff
@@ -0,0 +1,153 @@
...
+
+The language team updates the experimental flags `generic-metadata` and `triple-shift` to
+always be enabled and no longer be available to users, and releases this update
+in the next stable Dart release.
+
+### Use
+
+The Dart library team introduces `int.operator>>>` as soon as possible.
+
+### Documentation
```
leafpetersen, line 132: @kwalrath I think there's probably some minor documentation work to be done here (e.g. I seem to remember there's a table of operators with their precedence floating around somewhere).

leafpetersen (approved):
LGTM

accepted/future-releases/small-features-21Q1/feature-specification.md
```diff
@@ -0,0 +1,109 @@
...
+
+Because we already infer `List<T Function<T>(T)>` in the code above, this change will not affect type inference, it will just make the inferred type not be an error afterwards.
+
+We do not expect the removal of this restriction to affect the feasibility of type inference. After all, it's already possible to have a generic function type occurring covariantly in a type argument, like `List<T Function<T>(T) Function()>`.
+
+## Allow `>>>` as overridable operator
```
srawlins, line 81: This section does not specify that `>>>=` is also legal syntax. Should it? Looks like the VM currently supports it with `--enable-experiment=triple-shift`.

accepted/future-releases/small-features-21Q1/implementation-plan.md
srawlins: On line 92 of this section, it says:

> There are no new quick-fixes needed.

accepted/future-releases/small-features-21Q1/feature-specification.md
crelier: No, I do not have concerns. As mentioned on line 75, tests should point us to the places where we assumed (and hopefully verified with asserts) that we did not have to bother with handling generic types.

accepted/future-releases/small-features-21Q1/feature-specification.md
lrhn: ACK

accepted/future-releases/small-features-21Q1/feature-specification.md
lrhn: It follow from being an `<operator>`, because we allow composite assignment for all `<operator>`s, but I should probably write it explicitly.

accepted/future-releases/small-features-21Q1/implementation-plan.md
lrhn: Ack. That table should probably be updated.

accepted/future-releases/small-features-21Q1/feature-specification.md
srawlins: Excellent, thanks!

accepted/future-releases/small-features-21Q1/feature-specification.md
```diff
@@ -0,0 +1,109 @@
...
+
+When the operator has been enabled, we'll quickly (in the same dev-release series) introduce:
+
+```dart
+int operator >>>(int shift);
+```
+
+on the `int` class, which will work similarly to `>>`, but will zero-extend instead of sign-extend. 
+
+At that point, the `>>>` operator on `int` must be **valid in constant and potentially constant expressions**, so `1 << 3` is a compile-time constant expression with the value `8`, and `const C(int x) : y = 1 << x;` is valid (although potentially throwing if `x > 63`) as a constant constructor.
```
scheglov, line 107: Should the example be updated to describe `>>>`?

lrhn: It probably should, yes.

Issue #496, opened 2019-08-06
Allow generic function types as type arguments and bounds


eernstg:
Feature specification: https://github.com/dart-lang/language/blob/master/accepted/future-releases/small-features-21Q1/feature-specification.md

---

It is currently an error to use a generic function type as a type argument or as the bound of a type variable. This issue is a proposal to lift that restriction. Here is an example:

```dart
void f<X>() {}

main() {
  var xs = [f]; // Error: ''.. a function with type parameters cannot be used as a type argument'.
}
```

The restriction mainly serves to keep type inference tractable, but we do not prevent function types _containing_ such types from being used as type arguments, so we presumably already handle the complexity arising from having such types when dealing with the parts of types that we already allow:

```dart
void Function<X>() g(void Function<X>() f) => f;

main() {
  var xs = [g]; // OK.
}
```

The restriction is at times inconvenient, e.g., because it forces data structures like lists to have less informative types than they could otherwise have, forcing expressions accessing elements of that list to use casts or dynamic invocations.

If the restriction is lifted then some new values of type `Type` can be obtained (because a generic function type can now be the value of a type variable), e.g., a reification of `void Function<X>()`&mdash;only reflection could deliver such reified types previously. But it was already possible to obtain a reification of, say, `void Function<X>() Function()`, so it might be a small change.

In any case, it involves both the analyzer and the common front end (because some types are no longer rejected), and it also involves backends (in particular, if there is something genuinely new in supporting these new reified types).

We need to ensure that it is indeed true for all implementations that they do not rely in any essential way on having this restriction. @mraleph, @stereotype441, if this restriction is lifted, is it your impression that this would break any invariants that our tools currently rely on?

@munificent, @leafpetersen, @lrhn, what do you think?

lrhn:
I'd love to remove the restriction. I have already been hit by it once, when I tried to make a list of generic functions.

We are exacerbating the issue by actually actually *inferring* the type, and then reporting it as an error. That should never happen. We must not report errors in "code" that the user didn't write. We should refuse to infer a type instead of inferring a known invalid type.

I would be surprised if lifting the restriction introduces any problem for `List<T Function<T>(T)>` that wouldn't already be hit by `List<T Function<T>(T) Function()>`. The type occurs just as covariantly in the latter case.

(I can be surprised, I hope I won't be).

lrhn:
See https://github.com/dart-lang/sdk/issues/41871 for a concrete problem.

[...]

PR #1445, opened 2021-02-09
Small-features-21Q1 fix: Allow generic functions as bounds

eernstg:
The small-features-21Q1 spec includes the feature described in #496, except that it does not mention generic function types used as bounds. I think we should include bounds: There is a need to specify a generic function type as a bound if a particular type variable is intended to have a value which is a generic function type, e.g., in order to know how many type arguments to pass in invocations.

```dart
class C<F extends void Function<X>(X)> {
  F f;
  C(this.f);
  void m() { f<int>(42); }
}

class D<F extends void Function<X>(X, [List<X>?]) extends C<X> {...}
```

Also, this PR fixes a typo `tripple-shift` --> `triple-shift`.

### DIFF
--- a/accepted/future-releases/small-features-21Q1/feature-specification.md
+++ b/accepted/future-releases/small-features-21Q1/feature-specification.md
@@ -15,12 +15,12 @@ We have a [list](https://github.com/dart-lang/language/issues/1077) of smaller l
 For the first quarter of 2021, we schedule three of these:
 
 1. Allow type arguments on annotations ([#1297](https://github.com/dart-lang/language/issues/1297)).
-2. Allow generic function types as type arguments ([#496](https://github.com/dart-lang/language/issues/496)).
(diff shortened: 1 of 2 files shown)


eernstg:
@lrhn, do you wish to omit the bounds from #496?

lrhn:
No, I want to allow all types everywhere, as arguments and as bounds.

It gets problematic otherwise: `foo<F extends Function, G extends F>(){}` with `foo<T Function<T>(T), Never Function<T>(Object?)>()` would either be a run-time type error because the bound is a generic function type, or it would always be a compile-time error to use a type variable as bound unless it cannot possibly be bound to a generic function type (so not an unbounded type variable, and not one bounded by `Function`, `Object`, `FutureOr<Object>`, or similar supertypes of generic function types).

lrhn (approved):
LGTM, thanks.

Issue #1434, opened 2021-02-05
Semantics of JavaScript compiled `>>` as intended?


eernstg:
We are probably not going to change the semantics of `>>` when compiling to JavaScript, but the implemented behavior was actually surprising to me. Consider the following program:

```dart
// Assuming compilation to JS.

void main() async {
  // Signed shift right on positive number that fits in 32 bits: No surprise.
  print(4294967295 >> 10); // 4194303

  // 64 bit literals: Truncated in JS, zeros at the end to avoid error.
  // This shows that we're masking off all bits >31.
  print(0x0000000080000000 >> 0); // bit 31 is one: 2147483648.
  print(0x0000000100000000 >> 0); // bit 32 is one: 0.
  print(0xffffffff00000000 >> 0); // all bits >31 are one: 0.

  // Signed shift right _does_ mask off all bits >31, and it's _not_ bit 31
  // which plays the role as a sign bit, it must be the sign of the IEEE 754
  // representation, that is, a *33rd* bit.
  print(0xffffffff00000000 >> 31); // 0.
  print(0xffffffff80000000 >> 31); // 1.

  // Signed shift right _is_ signed, in the sense that we preserve the sign
  // bit of a representation whose IEEE 754 interpretation is negative. But
  // then we mask off the signed result at 32 bits, and consider it unsigned.
  print(-1 >> 1000); // 4294967295.
}
```

This shows that the `operator >>` on `int` in JavaScript-compiled programs behaves as a signed shift on a bit string of length *33*. The 32 bits are taken from the mantissa of the IEEE 754 representation, and the 33rd bit is the sign bit.

If this is working as intended then we should specify it.

@rakudrama, @lrhn, @munificent, @leafpetersen, @stereotype441, @jakemac53, @natebosch, WDYT?

lrhn:
The `>>` surprises me too.

I *though* the specified behavior of bitwise operations on the web are to use the underlying JS operations, possibly after some argument validation - we don't do (mod 32) on the shift counts. The JS operation coerces to Int32 first, then does 32-bit operations. Afterwards we convert the result to Uint32 (by doing `>>> 0`).

[...]

PR #1448, opened 2021-02-10
Adding a sentence about the semantics of >> in JavaScript compiled code

eernstg:
Cf. https://github.com/dart-lang/language/issues/1434.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -19501,9 +19501,14 @@ This introduces a number of differences:
 \item[$\bullet$]
   Bitwise operations on integers (and, or, xor, negation, and shifts)
   all truncate the operands to 32-bit two's complement integers,
-  perform 32-bit operations on those,
-  and the resulting 32 bits are interpreted as a 32-bit unsigned integer
-  (for example, \code{-1 <{}< 1} evaluates to 4294967294).
+  perform 32-bit operations on those, and the resulting 32 bits are
+  interpreted as a 32-bit unsigned integer. For example,
+  \code{-1\,\ltlt\,1} evaluates to 4294967294. The right shift
+  operator, \lit{\gtgt}, performs a signed shift when the sign bit of
+  the JavaScript number is 1, and an unsigned shift when it is 0. For
+  example, with one in bits 1 and 31 and zeros elsewhere the shift is
+  unsigned in \code{0x80000002\ \gtgt\ 1\ ==\ 0x40000001}, and signed in
+  \code{-0x7ffffffe\ \gtgt\ 1\ ==\ 0xc0000001}.
 \item[$\bullet$]
   The \code{identical} method cannot distinguish the values $0.0$ and
   $-0.0$, and it cannot recognize any \Index{NaN} value as identical
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -423,7 +423,7 @@ An example would be:
   \alt <oneOrMoreThings>+
   \alt <anOptionalThing>?
   \alt (<some> <grouped> <things>)
-  \alt \gtilde{}<notAThing>
+  \alt \gtilde<notAThing>
   \alt `aTerminal'
   \alt <A\_LEXICAL\_THING>
 \end{grammar}
@@ -2383,7 +2383,7 @@ The following names are allowed for user-defined operators:
 \lit{-},
 \lit{+},
 \lit{/},
-\lit{\gtilde{}/},
+\lit{\gtilde/},
 \lit{*},
 \lit{\%},
 \lit{|},
@@ -2394,7 +2394,7 @@ The following names are allowed for user-defined operators:
 \lit{\gtgt},
 \lit{[]=},
 \lit{[]},
-\lit{\gtilde{}}.
+\lit{\gtilde}.
 
 \LMHash{}%
 It is a compile-time error if the arity of the user-declared operator
@@ -2407,7 +2407,7 @@ It is a compile-time error if the arity of a user-declared operator with one of 
 \lit{==},
 \lit{-},
 \lit{+},
-\lit{\gtilde{}/},
+\lit{\gtilde/},
 \lit{/},
 \lit{*},
 \lit{\%},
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -19501,9 +19501,14 @@ \section*{Appendix: Integer Implementations}
...
-  and the resulting 32 bits are interpreted as a 32-bit unsigned integer
-  (for example, \code{-1 <{}< 1} evaluates to 4294967294).
+  perform 32-bit operations on those, and the resulting 32 bits are
+  interpreted as a 32-bit unsigned integer. For example,
+  \code{-1\,\ltlt\,1} evaluates to 4294967294. The right shift
+  operator, \lit{\gtgt}, performs a signed shift when the sign bit of
+  the JavaScript number is 1, and an unsigned shift when it is 0. For
+  example, with one in bits 1 and 31 and zeros elsewhere the shift is
+  unsigned in \code{0x80000002\ \gtgt\ 1\ ==\ 0x40000001}, and signed in
+  \code{-0x7ffffffe\ \gtgt\ 1\ ==\ 0xc0000001}.
```
lrhn, line 19511: That's not what the `>>` operator actually does, is it?
It bases its choice on whether to do signed or unsigned shift (sign-extending or zero-extending) on the sign of the *original* value, not the 32-bit two's complement value it was truncated to.

eernstg: Right, it _is_ using the sign bit of the IEEE 754 representation, and that's the reason why we have to go around corners to get a number whose sign bit is 1 and whose mantissa has 0x80000002 as its lower 32 bits (down to the "binary point", cf. decimal point, as determined by the exponent).

Can you think of a really nicely readable way to obtain such a number? `-0x7ffffffe` will do, but the reader needs to mentally flip all bits and add one in order to obtain the lowest 32 bits.

eernstg: Ah, one more thing: It never sign-extends, it just copies a 1 from bit 31 to bit 30 for each step of the shift operation when the shift is signed, and then it produces a result which has zero in every bit >31. Sign-extension based on bit 31 as the sign bit would be done by `toSigned(32)`.

lrhn: It does sign extend the 32-bit number as a 32-bit number. 
It's really as if `n >> s` was:
```dart
  if (s < 0) throw RangError(...);
  If (n < 0) {
    if (s >= 32) {
      return (-1).toUnsigned(32);
    } 
    return shiftRight(n.toSigned(32), x).toUnsigned(32);
  } 
  if (s >= 32) {
    return 0;
  } 
  return shiftRight(n.toUnsigned(32), x);
```
where shiftRight can be either right-shift operation because the top 32 bits never reaches the result.
In practice, it's probably this JavaScript:
```javascript
 // check s >= 0 before calling into JavaScript.
 if (n < 0) return (s >= 32 ? -1 : n >> s) >>> 0;
 return s >= 32 ? 0 : n >>> s;
}
```

So, phrasing that (if we want to enshrine it in the spec) would something like:
> Bitwise operations on integers (`&`, `|`, `^`, `~`, `<<` and `>>`)
> all truncate the operands to 32-bit two's complement integers,	  
> perform 32-bit operations on those,
> and the resulting 32 bits are interpreted as a 32-bit unsigned integer.
> For example, `-1 << 1` evaluates to 4294967294 (also known as `(-2).toUnsigned(32)`. 
> The right shift operator, `>>`, performs a sign-extending shift on 
> the 32-bit two's complement integer when the original number is negative,
> and zero-extending shift when the original number is non-negative.
> For example: `0x80000002 >> 1 == 0x40000001`, but 
> `-0x7FFFFFFD >> 1 == 0xC0000001`.

eernstg: > It does sign extend the 32-bit number as a 32-bit number.

I'm not sure what that would mean. Sign extension would normally be a conversion from a _k_ bit two's complement representation to a _k+n_ bit two's complement representation for some positive _n_, preserving the denoted integer. So we'd sign extend the 8 bit signed number 0x9 (denoting -7) to 16 bit as 0xf9 (also -7). That is, we make the bit string longer by adding bits to the left, copying the sign bit of the old representation into each newly added bit (including the new sign bit).

Just checking for normalcy, this matches the description here: https://en.wikipedia.org/wiki/Sign_extension.

The distinction here is between a signed right shift and an unsigned right shift. At each step, with _k_ bits, they will both write bit 1 unto position 0 (thus eliminating the least significant bit), then bit 2 unto position 1, etc., up to bit _k-1_ unto position _k-2_. With the signed shift we're done (and the last thing we did was to "copy the sign bit into the most significant position"); with the unsigned shift we have one more action: store 0 into position _k-1_ (so we make the most significant bit 0, no matter what it was previously).

I don't think it's going to help anyone if we try to use the phrase sign extension to describe signed and unsigned shift operations. We could call them logical and arithmetic shift operations, but I suspect that this would be less informative for many readers, especially because `>>>` tends to be described as the 'unsigned right shift' operator in JavaScript.

I used the new text, but adjusted the part about sign extension to talk about signed and unsigned shift operations. I believe those two terms are quite well-established.

eernstg (commented):
Thanks for the review! Here's a response. I haven't used the phrase 'sign extension' (or any variant thereof) because that would generally refer to a different operation, but I do think the new text is a bit more readable.

specification/dartLangSpec.tex
lrhn: ACK. I think that "sign extending shift" is also used about the "signed shift"/"arithmetic shift" (vs "unsigned shift"/"logical shift"). I'm sure I have seen it used, but not as commonly in authoritative sources as in people asking questions.

For example, the [Java language spec](https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#:~:text=sign-extension) says that:
> The value of `n >> s` is `n` right-shifted `s` bit positions with sign-extension.

The phrasing "sign extending shift" makes sense if you think about it in certain ways. If you right shift a 32 bit number by 17, you get 15 bits. if you then sign extend those 15 bits back to 15+17 bits, it matches the meaning of sign extension. The https://en.wikipedia.org/wiki/Arithmetic_shift article mentions that "(this is a kind of sign extension)" ... but not enough to be non-parenthetically mentioned.

I agree that "signed shift" and "arithmetic shift" are more common, and we should use those instead. (Then we should probably specify what it means, just to be sure).

lrhn (approved):

specification/dartLangSpec.tex
eernstg: Thanks!

> Then we should probably specify what it means

Did that.

PR #1450, opened 2021-02-10
add initial static metaprogramming doc

jakemac53:
This adds an initial document laying out the use cases we would like to solve with static metaprogramming, some rough ideas in terms of the general way we want approach the problem, and the known challenges that we will face.

### DIFF
--- a/working/static metaprogramming/intro.md
+++ b/working/static metaprogramming/intro.md
@@ -220,7 +220,9 @@ remaining code is almost pure boilerplate that is identical across all
 `StatefulWidget` and `State` subclasses.
 
 There are other patterns in Flutter like this that are commonly applied to
-widget classes but require Flutter users to write a lot of repetitive code.
+widget classes but require Flutter users to write a lot of repetitive code,
+such as the `dispose()` method which needs to do any required cleanup of widget
+state (removing listeners, calling custom `dispose` methods on fields, etc).
 Metaprogramming could enable the Flutter framework to express these patterns
 more tersely and with less chance for user error.
 
@@ -301,15 +303,17 @@ The approach we propose is **macros**, and the above requirements list can be
 stated more concretely as:
 
 1. We design some sort of syntax to apply a macro by name to a given class,
-   method declaration, etc. We haven’t spent time here yet.
-2. Metaprogramming happens by defining a macro, which can be called at compile
-   time.
-3. The introspection system is TBD. It could be an imperative API similar to the
-   existing mirrors API or analyzer API (but run at compile-time), or perhaps
-   something declarative like to pattern matching over syntax examples.
-4. We could use an imperative API to construct new pieces of Dart code in a
-   macro, but some sort of quotation syntax would make this nicer and make the
-   API less brittle.
+   method declaration, etc.
+2. We propose to design a way of defining macros which can be then be applied
+   as above.
+3. We propose to design an introspection system to allow macros to introspect
+   on the program during their execution. This might be an imperative API
+   similar to the existing mirrors API or analyzer API, or something
+   declarative like to pattern matching over syntax examples.
+4. We propose to consider designing a quotation syntax so that macros that
+   generate Dart code can generate code in a readable fashion. This also means
+   code generators only depend on language *syntax* for code generation and not
+   an imperative api, which will better stand the test of time.
 
 Most of this is as-yet-undesigned, but the key piece is that **we think macros
 should be written in normal imperative Dart code which is executed at compile
--- a/working/static metaprogramming/intro.md
+++ b/working/static metaprogramming/intro.md
@@ -262,6 +262,9 @@ ensure that we don't compromise the existing advantages of Dart.
   workflow—any change in the program should not have to rerun all
   metaprogramming in the entire program. It should at worst require a rerun of
   the metaprogramming in libraries that depend on the modified library.
+* **Must support expression evaluation**. Expression evaluation is a vital tool
+  for the development experience and we need to ensure it is not negatively
+  affected.
 
 For the most part the design constraints here align with the modular compilation
 design constraints. In short, metaprogramming should not introduce global action
@@ -354,7 +357,9 @@ accumulate imperative features or end up replaced by those that do.
 
 In the Dart ecosystem, in the absence of metaprogramming in the language, users
 often resort to code generators. Some of those are quite sophisticated and are
-written in imperative Dart code.
+written in imperative Dart code. These typically take the form of kernel
+transformers or separate build processes like build_runner which run before the
+compilers and are not directly integrated.
 
 Rust has a nice [declarative, rule-based macro system][rust_declarative_macros].
 But, because that rule-based system is limited, Rust also has a second
@@ -389,6 +394,9 @@ both are applied to the same class, does the memoized field get serialized?
 Do macros run before type checking, after, or interleaved? What is the static
 type context in which a macro runs?
 
+How do macros interact with const canonicalization and evaluation, do they run
+before, after, or interleaved? Can macros read const values *and* generate them?
+
 There are additional considerations if we allow recursive macros, and cycles
 between macro defining libraries.
 
@@ -398,12 +406,14 @@ Macros can easily lead to incomprehensible programs if they become overly
 magical, or are a complete black box to the user. In order to demystify macros
 you should:
 
-* Be able to easily navigate to their implementation code.
-* Be able to visualize in some way the code they generated in your program, at
+* Be able to easily navigate to the macro implementation.
+* Be able to visualize in some way the code generated into your program, at
   development time.
 * Be able to debug generated code (step through it, etc).
+  * Possibly provide an opt out for this, if feasible/desirable.
 * Be able to trace errors that flow through generated code, as well as navigate
   back to the line in the macro implementation that produced the code.
+* Be able to auto-complete macro generated apis.
 
 ### Performance
 
--- a/working/static metaprogramming/intro.md
+++ b/working/static metaprogramming/intro.md
@@ -244,6 +244,8 @@ From these sample use cases and others, we can generalize some requirements:
   class code generator for instance would ideally be a combination of several
   more targeted code generators for each of the desired features (`copyWith()`,
   `==`, `hashCode`, etc).
+* **Generated code must be debuggable and user visible**. See the Usability
+  section below.
 
 ## Design constraints
 
@@ -394,11 +396,14 @@ between macro defining libraries.
 
 Macros can easily lead to incomprehensible programs if they become overly
 magical, or are a complete black box to the user. In order to demystify macros
-you would ideally be able to easily navigate to their implementation code, as
-well as visualize in some way the code they generated in your program.
+you should:
 
-When errors occur in generated code, how does the user trace backwards to
-understand where that code came from and how it was generated?
+* Be able to easily navigate to their implementation code.
+* Be able to visualize in some way the code they generated in your program, at
+  development time.
+* Be able to debug generated code (step through it, etc).
+* Be able to trace errors that flow through generated code, as well as navigate
+  back to the line in the macro implementation that produced the code.
 
 ### Performance
 
--- a/working/static metaprogramming/intro.md
+++ b/working/static metaprogramming/intro.md
@@ -293,11 +293,11 @@ Dart needs to provide a way for users to:
 2. **Define some metaprogramming** that specifies how the copyWith() method is
    defined in terms of the class’s fields. That metaprogramming needs to be
    authored in some kind of metaprogramming language.
-3. It uses some sort of **introspection mechanism** to walk over the class’s
-   field declarations.
-4. Based on that it uses some other **construction mechanism** to produce a new
-   method including its signature and body. The resulting is then added to the
-   original class declaration.
+3. Use an **introspection mechanism** to walk over the class’s field
+   declarations.
+4. Use a **construction mechanism** to produce a new method including its
+   signature and body. The resulting is then added to the original class
+   declaration.
 
 The approach we propose is **macros**, and the above requirements list can be
 stated more concretely as:
--- a/working/static metaprogramming/intro.md
+++ b/working/static metaprogramming/intro.md
@@ -1,0 +1,416 @@
+# Static Metaprogramming in Dart
+
+## Problem Statement
+
+Programmers hate repetition—having to write the same or very similar code over
+and over again. It’s boring, tedious, error-prone, and harder to maintain in
+parallel correctly. A key property of what makes a programming language
+productive is its facilities for eliminating repetition.
+
+The humble loop is a simple example. Instead of:
+
+```dart
+list.add(1);
+list.add(2);
+list.add(3);
+```
+
+Which is asking for a copy/paste error, we’d rather write:
+
+```dart
+for (var i = 1; i <= 5; i++) {
+  list.add(i);
+}
+```
+
+A more powerful example is functions. If you need to build a Flutter widget
+containing three similar columns, would you rather maintain this:
+
+```dart
+class MyApp extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return Row(
+      children: [
+        Column(
+          mainAxisSize: MainAxisSize.min,
+          mainAxisAlignment: MainAxisAlignment.center,
+          children: [Text('CALL')],
+        ),
+        Column(
+          mainAxisSize: MainAxisSize.min,
+          mainAxisAlignment: MainAxisAlignment.center,
+          children: [Text('ROUTE')],
+        ),
+        Column(
+          mainAxisSize: MainAxisSize.min,
+          mainAxisAlignment: MainAxisAlignment.center,
+          children: [Text('SHARE')],
+        ),
+      ],
+    );
+  }
+}
+```
+
+Or this:
+
+```dart
+class MyApp extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return Row(
+      children: [
+        _buildColumn('CALL'),
+        _buildColumn('ROUTE'),
+        _buildColumn('SHARE'),
+      ],
+    );
+  }
+
+  Column _buildColumn(String label) {
+    return Column(
+      mainAxisSize: MainAxisSize.min,
+      mainAxisAlignment: MainAxisAlignment.center,
+      children: [Text(label)],
+    );
+  }
+}
+```
+
+Loops and functions are powerful, but fundamentally limited. They only execute
+within imperative code—initializers and function bodies—and they can only
+operate on data—values that are first-class in Dart. If you want to, say,
+iterate over all the fields in a class (maybe to serialize them), you have to
+use the “dart:mirrors” library which exists to expose data about classes as
+first-class Dart objects. That works, but sacrifices performance, causes code
+size problems, and isn’t supported at all in AOT builds.
+
+In the above example, we can hoist the repetitive code to create a column into a
+function because Flutter widgets are built using imperative code inside
+`build()` methods. But say you wanted to define each column as a separate widget
+class, like this:
+
+```dart
+class CallBuildColumn extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return Column(
+      mainAxisSize: MainAxisSize.min,
+      mainAxisAlignment: MainAxisAlignment.center,
+      children: [Text('CALL')],
+    );
+  }
+}
+
+class RouteBuildColumn extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return Column(
+      mainAxisSize: MainAxisSize.min,
+      mainAxisAlignment: MainAxisAlignment.center,
+      children: [Text('ROUTE')],
+    );
+  }
+}
+
+class ShareBuildColumn extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return Column(
+      mainAxisSize: MainAxisSize.min,
+      mainAxisAlignment: MainAxisAlignment.center,
+      children: [Text('SHARE')],
+    );
+  }
+}
+```
+
+There is no function you can write in Dart and call three times to output these
+three classes, short of a heavy code generator. Dart doesn’t let you abstract
+over classes or other top-level declarations.
+
+**Metaprogramming** refers to code that can do this—code that operates on other
+code as if it were data. It can take code in as parameters, reflect over it,
+inspect it, create it, modify it, and return it. ***Static* metaprogramming**
+means doing that work at compile-time. This avoids the safety, performance, and
+code size problems of runtime metaprogramming. Examples are the C preprocessor,
+C++ templates, Rust macros, Swift [function builders][function_builders], and
+[compile-time execution in Zig][zig_compile_time_execution].
+
+This document proposes an approach to adding static metaprogramming in Dart that
+we’d like to investigate.
+
+## Sample use cases
+
+Here is a non-exhaustive list of some of the problems users have run into where
+it feels like there should be a reusable general solution but instead they have
+to write the same repetitive boilerplate code every time:
+
+### JSON serialization and deserialization
+
+Given some class declaration, it’s trivial to write a method that serializes it
+to JSON:
+
+```dart
+class Dog {
+  final String breed;
+  final String name;
+
+  Object toJSON() => {'breed': breed, 'name': name};
+}
+```
+
+But that code requires iterating over the fields of the class somehow. Short of
+using mirrors or code generation, there’s no way to write a reusable system that
+can add JSON serialization/deserialization to any given class.
+
+### Data classes
+
+The [most requested open language issue][data_classes_issue] is to add data
+classes. A data class is essentially a regular Dart class that comes with an
+automatically provided constructor and implementations of `==`, `hashCode`, and
+`copyWith()` (called `copy()` [in Kotlin][kotlin_copy]) methods based on the
+fields the user declares in the class.
+
+The reason this is a *language* feature request is because there’s no way for a
+Dart library or framework to add data classes as a reusable mechanism. Again,
+this is because there isn’t any easily available abstraction that lets a Dart
+user express “given this set of fields, add these methods to the class”. The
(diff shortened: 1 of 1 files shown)


working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,416 @@
...
+between macro defining libraries.
+
+### Usability
+
+Macros can easily lead to incomprehensible programs if they become overly
+magical, or are a complete black box to the user. In order to demystify macros
+you would ideally be able to easily navigate to their implementation code, as
+well as visualize in some way the code they generated in your program.
+
+When errors occur in generated code, how does the user trace backwards to
```
kevmoo, line 400: Or an error flows through generated code.

kevmoo: Nit, but it'd be nice to explicitly mention `debugging` here. I searched for it and couldn't find it.

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,416 @@
...
+The widget name is repeated six times, `fixedStuff` appears twice, and the
+remaining code is almost pure boilerplate that is identical across all
+`StatefulWidget` and `State` subclasses.
+
+There are other patterns in Flutter like this that are commonly applied to
+widget classes but require Flutter users to write a lot of repetitive code.
+Metaprogramming could enable the Flutter framework to express these patterns
+more tersely and with less chance for user error.
+
+## Requirements
```
kevmoo, line 227: What about configuration? What inputs will a code generator have access to? Only the input (Dart) file?

Can they read other Dart files? Other (configuration) files? Yaml?

jakemac53: This is all TBD

kevmoo, line 227: Related to the debugging bits below, I'd say a pretty hard requirement is that generated code is user-visible in development tools *during* development.

Also static errors in generated code should be visible without having to run the code through a compiler.

Basically I want it to feel like build-to-source, even though we're not building to source.

If this feels like Ruby (circa 2008) we'll create a mess.

jakemac53: Ya we will get more concrete about this stuff, but we are on the same page.

I can add a requirement bullet about this though.

working/static metaprogramming/intro.md
jakemac53: I just updated this section and linked from requirements

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,421 @@
...
+
+The widget name is repeated six times, `fixedStuff` appears twice, and the
+remaining code is almost pure boilerplate that is identical across all
+`StatefulWidget` and `State` subclasses.
+
+There are other patterns in Flutter like this that are commonly applied to
+widget classes but require Flutter users to write a lot of repetitive code.
+Metaprogramming could enable the Flutter framework to express these patterns
+more tersely and with less chance for user error.
+
```
jacob314, line 226: Would be nice to call out disposable objects specifically here.

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,421 @@
...
+In addition to user-level requirements, there are some practical constraints to
+ensure that we don't compromise the existing advantages of Dart.
+
+* **Must support modular compilation**. Any individual library should be able to
+  be fully compiled (including running all macros) using only the transitive
+  dependencies of the library. To this end we should not expose any API or
+  functionality that allows global access to the program, outside of the current
+  library and/or its dependencies. Such functionality would defeat modular
+  compilation.
+* **Must support incremental compilation**. This is key to the hot reload
```
jacob314, line 261: also for expression evaluation.

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,421 @@
...
+  metaprogramming in the entire program. It should at worst require a rerun of
+  the metaprogramming in libraries that depend on the modified library.
+
+For the most part the design constraints here align with the modular compilation
+design constraints. In short, metaprogramming should not introduce global action
+at a distance.
+
+## Non-goals
+
+* **Modifying existing code**. We don’t see a need to allow modifiying existing
```
jacob314, line 272: Can you get more specific on what modifying existing code does and does not mean? Does it mean no deleting existing code but inserting arbitrary lines in-between existing code is fair game? If it ok to add conditionals that may cause blocks of existing code to not execute (for example turn off asserts)?

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,421 @@
...
+macros *correctly*, but has always been sufficiently powerful.
+
+Experience from many UI templating languages (which are surprisingly similar to
+macro metaprogramming in many ways) is that most declarative ones gradually
+accumulate imperative features or end up replaced by those that do.
+
+In the Dart ecosystem, in the absence of metaprogramming in the language, users
+often resort to code generators. Some of those are quite sophisticated and are
+written in imperative Dart code.
+
```
jacob314, line 358: might be worth calling out Dart kernel transformers here which perhaps have a superset of the functionality you need.

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,421 @@
...
+distinguish between these?
+
+### Ordering and staging
+
+In what order are macros run? If they can both generate new APIs and introspect
+on the program, then ordering can be user-visible. Consider a macro that adds a
+field for memoization and another that adds a method to serialize all fields. If
+both are applied to the same class, does the memoized field get serialized?
+
+Do macros run before type checking, after, or interleaved? What is the static
```
jacob314, line 389: one more: after or before const canonicalization.

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,421 @@
...
+There are additional considerations if we allow recursive macros, and cycles
+between macro defining libraries.
+
+### Usability
+
+Macros can easily lead to incomprehensible programs if they become overly
+magical, or are a complete black box to the user. In order to demystify macros
+you should:
+
+* Be able to easily navigate to their implementation code.
```
jacob314, line 401: Is this the code defining the macro?

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,421 @@
...
+* Be able to visualize in some way the code they generated in your program, at
```
jacob314, line 402: I would be interested in seeing some mocks of what this looks like.

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,421 @@
...
+  development time.
+* Be able to debug generated code (step through it, etc).
```
jacob314, line 404: probably also need to be able to opt to not-debug into it unless you really want to.

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,421 @@
...
+* Be able to trace errors that flow through generated code, as well as navigate
+  back to the line in the macro implementation that produced the code.
```
jacob314, line 406: It makes sense that the line in the macro implementation would be useful for a macro author debugging. I'm skeptical it would often be useful for users. For users, I'd want more a sourcemap back to the relevant data driving the macro. For example, what features of my source code caused this macro to trigger at all? Was there some annotation that I can change to turn off the functionality?

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,421 @@
...
+
+There are additional considerations if we allow recursive macros, and cycles
+between macro defining libraries.
+
+### Usability
+
+Macros can easily lead to incomprehensible programs if they become overly
+magical, or are a complete black box to the user. In order to demystify macros
+you should:
+
```
jacob314, line 400: What about auto-complete? Will results from macros need to show up in autocomplete?

working/static metaprogramming/intro.md
kevmoo: Good point! I assumed the granularity of meta programming was at the member level. Add members, add classes, etc.

Would be good to clarify.

jakemac53: By "Modify" we basically mean "Overwrite" - is that more clear?

@munificent are you fine with that wording change?

See the sentence here "Dart code that already has clearly defined meaning today should continue to mean the same thing after metaprogramming." that is the guiding principle. Allowing you to overwrite code clearly violates this principle (or at least would give you the power to violate it).

I don't think we want to restrict this purely to new top level and class declarations quite yet, but that may very well be where we end up (also likely allowing implementing abstract/external declarations).

working/static metaprogramming/intro.md
jakemac53: Yes, clarified

working/static metaprogramming/intro.md
jakemac53: Added an explicit bullet for this, but yes

working/static metaprogramming/intro.md
jakemac53: I am sure there will be a lot of collaboration here once we get to filling in the details for how we will achieve each of these requirements :)

working/static metaprogramming/intro.md
jakemac53: We don't have that for build_runner generated code today - I added that as a possibility but I don't see it as a requirement (nobody has asked for it in build_runner for instance).

working/static metaprogramming/intro.md
jakemac53: I don't think that would be feasible here. Pointing at the macro implementation should give you some hints though (you can see why it made the choice it did).

working/static metaprogramming/intro.md
jakemac53: done

working/static metaprogramming/intro.md
jakemac53: done

working/static metaprogramming/intro.md
jakemac53: I added a separate bullet for this

working/static metaprogramming/intro.md
jakemac53: done

working/static metaprogramming/intro.md
jacob314: Thanks that is bit clearer. Does wrapping a a block of code with an if statement count as overwriting or not?

working/static metaprogramming/intro.md
jacob314: Is the goal to just know the macro was "Jake's awesome serializer" or that it was Line 75 of jake_serializer.dart?

working/static metaprogramming/intro.md
jacob314: It probably depends on how invasive the uses of static metaprogramming are. If they are all use cases that code generators could also solve I agree that always having the code always be debuggable is fine. If the use cases are more invasive then accidentally stepping into the generated code could be quite annoying. For example, stepping into some synthetic code for every get of a `disposable` field would be a bit annoying just as stepping into the synthetic getter for a `late` field would be annoying.

working/static metaprogramming/intro.md
jacob314: Pointing to the macro implementation could be relevant for the macro author and very advanced users but I doubt it would be actionable for average user. For example, an average dart2js user could care less what line of the dart2js compiler emitted a line of code but they care a lot about the sourcemap telling them what line of their original dart code the generated JS code is related to.

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,431 @@
...
+  Column _buildColumn(String label) {
+    return Column(
+      mainAxisSize: MainAxisSize.min,
+      mainAxisAlignment: MainAxisAlignment.center,
+      children: [Text(label)],
+    );
+  }
+}
+```
```
Hixie, line 79: ironically that's considered bad style in flutter... better to pull out the _buildColumn into its own widget class. :-)

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,431 @@
...
+three classes, short of a heavy code generator. Dart doesn’t let you abstract
+over classes or other top-level declarations.
+
+**Metaprogramming** refers to code that can do this—code that operates on other
+code as if it were data. It can take code in as parameters, reflect over it,
+inspect it, create it, modify it, and return it. ***Static* metaprogramming**
+means doing that work at compile-time. This avoids the safety, performance, and
+code size problems of runtime metaprogramming. Examples are the C preprocessor,
+C++ templates, Rust macros, Swift [function builders][function_builders], and
+[compile-time execution in Zig][zig_compile_time_execution].
```
Hixie, line 139: and generics in Dart, to a very small limited extent.

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,431 @@
...
+compile time is a big ask. Here’s why we think it’s a good choice:
+
+### We already do it
+
+A subset of the Dart language is already allowed at compile time, which is
+specifically carved out to support limited const expressions. This means users
+are familiar with the concept of “running” Dart code at compile time, and we
+have invested some implementation time in supporting it. You could think of this
+proposal as merely (massively) expanding the subset of Dart allowed in a const
+expression.
```
Hixie, line 330: There are separate proposals to do just that, are we considering them part of this, a layer below (i.e. prerequisite to) this, or orthogonal to this?

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,431 @@
...
+
+[function_builders]: https://github.com/apple/swift-evolution/blob/9992cf3c11c2d5e0ea20bee98657d93902d5b174/proposals/XXXX-function-builders.md
+[zig_compile_time_execution]: https://andrewkelley.me/post/zig-programming-language-blurs-line-compile-time-run-time.html#:~:text=Compile%2DTime%20Parameters,-Compile%2Dtime%20parameters&text=In%20Zig%2C%20types%20are%20first,functions%2C%20and%20returned%20from%20functions.&text=At%20the%20callsite%2C%20the%20value,is%20known%20at%20compile%2Dtime.
+[data_classes_issue]: https://github.com/dart-lang/language/issues/314
+[kotlin_copy]: https://kotlinlang.org/docs/reference/data-classes.html#copying
+[observable]: https://docs.google.com/document/d/1L2NgM-Kl1PKt8iWQIrFugH0JUQ7Ky86MrM-WM1CPf0I/edit#heading=h.sb9uvy8myyqv
+[differentiable]: https://en.wikipedia.org/wiki/Differentiable_programming
+[rust_declarative_macros]: https://doc.rust-lang.org/book/ch19-06-macros.html#declarative-macros-with-macro_rules-for-general-metaprogramming
+[rust_procedural_macros]: https://doc.rust-lang.org/book/ch19-06-macros.html#procedural-macros-for-generating-code-from-attributes
```
jacob314, line 431: might want to reference flutter_hooks and react native hooks + possible js features.
https://github.com/flutter/flutter/issues/51752
https://github.com/rrousselGit/flutter_hooks
Would be interesting to understand if the features this doc is discussing are powerful for the use cases @rrousselGit is thinking about.

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,431 @@
...
+  * Possibly provide an opt out for this, if feasible/desirable.
+* Be able to trace errors that flow through generated code, as well as navigate
+  back to the line in the macro implementation that produced the code.
+* Be able to auto-complete macro generated apis.
+
+### Performance
+
+A Turing-complete programming language that runs in your typechecker opens the
+door to user-code that locks the IDE. How do we ensure that users maintain a
+fast edit refresh cycle when arbitrary Dart code may be running during compilation?
```
Hixie, line 422: The halting problem has one solution: timeouts. I recommend displaying a message after a few seconds that says something along the lines of "Macro blabla is taking a long time to execute. Press ^C to abort, O to obtain an Observatory URL with a debugger, S to print a stack trace, and W to get information on what code was being compiled when this macro was found."

working/static metaprogramming/intro.md
esDotDev: A problem this proposal would fix! Best of both worlds, devs can write the syntax that is familiar, but a pre-compiler can inline things, or make wrapper classes where it makes sense.

working/static metaprogramming/intro.md
jakemac53: The proposal to expand const is effectively a prerequisite to this proposal. That will essentially lay the groundwork for how we actually run user code at compile time, what rules are in place for that, etc.

working/static metaprogramming/intro.md
jakemac53: Yes, something like this will needed. We will otherwise run the risk of people filing issues about slow builds when its just the fault of some very expensive macro :). I am sure that will happen anyways, but we want to mitigate it for sure.

We may want to do it specifically by limiting recursion depth or something like that instead of a raw timeout (to have more consistent results across machines). But it is the same principle.

working/static metaprogramming/intro.md
esDotDev: Would this not support the use case of inlining then? Imagine I have some function, used 3 times, and it's better practice in flutter to have that inlined.

```
build(){
  return Column(
     _btn("1", onPressed: ...),
     _btn("2", onPressed: ...),
  )
}

@inline
Widget _btn(String label, onPressed) => Padding(padding: EdgeInsets.all(10), child: FlatButton(label, label, onPressed))
```
Could be turned into: 
```
build(){
  return Column(
    Padding(padding: EdgeInsets.all(10), child: FlatButton("1", label, onPressed)),
    Padding(padding: EdgeInsets.all(10), child: FlatButton("1", label, onPressed)),
  )
}
```
But that would require replacing the code inside the Widget tree.

It would basically get us Remi's functional widgets, without the annoying code generation phase:
https://pub.dev/packages/functional_widget

working/static metaprogramming/intro.md
PetarKirov: D popularized CTFE quite some time before Zig, by the way: https://dl.acm.org/doi/pdf/10.1145/3386323

working/static metaprogramming/intro.md
jakemac53: > Would this not support the use case of inlining then? Imagine I have some function, used multiple times, and it's better performance in flutter to have that inlined.

Technically, that wouldn't violate the principles as long as the inlining maintained the same scoping rules for the inlined code as it had in its own function. This isn't really a use case we are targeting though I don't think, and the compiler already should be doing this where it makes sense.

> It would basically get us Remi's functional widgets, without the annoying code generation phase:

functional_widget is one of the use cases we have looked at and we are targeting something that could support that - but you don't need inlining for it. You really just need to be able to code generate a class IIRC.

working/static metaprogramming/intro.md
jakemac53: Ideally you would know exactly where generated the code, whether that is feasible will depend on the more specific implementation we go with.

working/static metaprogramming/intro.md
rrousselGit: functional_widget is a weird one. It probably should not be done by code-generation, but function composition instead.

Code generation is far from ideal for such pattern. Because it's a very common mistake for users to use the function instead of the class, and it also causes naming issues.

esDotDev: Ya it occured to me you could do this a different way, by making `_btn` be the class. But it seems like that could get quite complicated from a scope standpoint because the method will have access to all the scope of the containing stateful widget, including context, and any member vars, so the generated class would need to sniff out all those deps.  Where as if we just inline, everything would just kinda work.

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,431 @@
...
+  be fully compiled (including running all macros) using only the transitive
+  dependencies of the library. To this end we should not expose any API or
+  functionality that allows global access to the program, outside of the current
+  library and/or its dependencies. Such functionality would defeat modular
+  compilation.
+* **Must support incremental compilation**. This is key to the hot reload
+  workflow—any change in the program should not have to rerun all
+  metaprogramming in the entire program. It should at worst require a rerun of
+  the metaprogramming in libraries that depend on the modified library.
+* **Must support expression evaluation**. Expression evaluation is a vital tool
```
lrhn, line 265: What is "expression evaluation"? Is it evaluating user-input expressions during debugging?

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,431 @@
...
+at a distance.
+
+## Non-goals
+
+* **Modifying existing code**. We don’t see a need to allow modifiying existing
+  code through metaprogramming. This leads to unexpected behavior, and
+  negatively affects usability. Dart code that already has clearly defined
+  meaning today should continue to mean the same thing after metaprogramming.
+  The goal is not to be able to hijack existing Dart syntax to do wildly
+  different things, but to augment code with new capabilities.
```
lrhn, line 280: One common use-case is to augment existing methods, maybe with an enter/exit protocol, without changing the code that actually runs in the middle. 
So *adding* something to a function, but not *changing* what is there (which really means that the new code can only go before/after/around the original code).

(One possible approach to that, which doesn't change existing code, is to allow a (possibly generated) mixin to be applied *after* the class declarations instead of only *before*. Maybe something like `class Foo with M1 { method(){}  } with M2;` which applies the mixin `M2` on top of `Object&M1{ method(){} }` and makes that the `Foo` class.)

working/static metaprogramming/intro.md
jakemac53: Yes

working/static metaprogramming/intro.md
jakemac53: Another possible approach is to generate code into a normal mixin, and require that users call `super.<method>` in order to get that behavior (and that they do it in the right place). However I think that is error prone, and is boiler plate of its own.

I do worry a bit about allowing arbitrary new code to be injected at the top/bottom of a method, as that could potentially change the behavior of it, but I do also recognize the usefulness of such an ability. Possibly some way of "wrapping" a function which is sort of what you are describing would make sense.

working/static metaprogramming/intro.md
leafpetersen: I think one thing that is being raised here is whether or not conditional compilation is covered by this proposal.  I think as stands, the answer might be "somewhat".  For example, I think this proposal is very compatible with a pattern where the user writes a static `MyClass.debugToString` method, and a macro generates a toString method which either calls that static method or not depending on compilation mode.  But I think we do not intend to support inclusion/exclusion of arbitrary fragments of text (a la #ifdef include) in the limit.

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,431 @@
...
+
+## Approach
+
+To understand what a static metaprogramming solution might look like, we’ll walk
+through one of the use cases. Say a Dart user wants to write a library that
+automatically defines `copyWith()` methods for classes. For them to do that,
+Dart needs to provide a way for users to:
+
+1. **Invoke the metaprogramming** to control which classes should get the
```
leafpetersen, line 289: I like this summary of the problem.

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,431 @@
...
+   `copyWith()` method.
+2. **Define some metaprogramming** that specifies how the copyWith() method is
+   defined in terms of the class’s fields. That metaprogramming needs to be
+   authored in some kind of metaprogramming language.
+3. It uses some sort of **introspection mechanism** to walk over the class’s
```
leafpetersen, line 294: Bullet 3 loses the tense.  Each of these bullets is a continuation of "provide a way for users to:", and "provide a way for users to It uses some sort of" doesn't work.  Same for bullet 4.

working/static metaprogramming/intro.md
```diff
@@ -0,0 +1,431 @@
...
+   field declarations.
+4. Based on that it uses some other **construction mechanism** to produce a new
+   method including its signature and body. The resulting is then added to the
+   original class declaration.
+
+The approach we propose is **macros**, and the above requirements list can be
+stated more concretely as:
+
+1. We design some sort of syntax to apply a macro by name to a given class,
```
leafpetersen, line 303: This list is too apologetic.  It's ok to be vague, I'd suggest being less concrete.  Just state the problems to be solved.  For example something along the following lines:

1) We propose to design a syntax to apply a macro to a given class, method declaration, etc.
2) We propose to design a way of defining macros which can be then be applied as above.
3) We propose to design an introspection system to allow macros to introspect on the program during their execution.  This might be an imperative API.... etc
4) We propose to consider designing a quotation syntax so that macros that generate Dart code can generate code in a readable fashion and take advantage of existing Dart tooling... etc.

leafpetersen:
LGTM.  I think there are a few comments to consider incorporating, but I think this is pretty close to the state where we should just land it, file a meta-issue, choose a label for the related issues (e.g. static-metaprogramming) and file discussions issues for any open points of discussion.

cc @mit-mit

working/static metaprogramming/intro.md
esDotDev: There's a common conditional-compilation, use case we have where we have some native .io calls we want to make from our UI, and we want them to just be no-ops on Web (or anywhere else .io is not supported). 

It would be nice if this could support that. Currently we have to maintain 3 files, 1 interface, and 2 implementations. It would be nice if we could just maintain the 1file, and flag some methods, or an entire class, as no ops if .io is missing.

As a bit of an aside, it would be nice if this could ease conditional compilation in general. Right now it's quite a rough spot imo, very few junior to intermediate developers even understand how to set these up.

Unity's approach of
```
#if UNITY_STANDALONE
  // Do desktop stuff
#endif 
```
Is very handy, especially as C# allows you to inline the entire package path, keeping it all out of the compilers way if it's not supported. 
Like:
```
#if UNITY_EDITOR
  // Do stuff with UnityEditor.* namespace, which does not exist in a published build
#endif
```

This opens the door for the code we really want to write as developers: 
```
Future<double> getTitleBarHeight() async {
  #if CAN_USE_IO
      await io.methodChannel.invokeMethod("getTitlebarHeight") as double
  #else
     return 40;
  #endif
}
```
Which lets us just have a single class, share state/logic, and just swap out the offending lines on each platform. This is most maintainable, easiest to grok, and most productive, just from developer POV.
Cheers!

PR #1453, opened 2021-02-15
Fix typo in PR 1448

eernstg:
I just noticed that 1448 was changed to use `-0x7FFFFFFD` rather than `-0x7ffffffe` at some point. I didn't have a strong opinion about upper/lower case, but the change from `e` to `D` is wrong, it should be `E`. This PR fixes that. It also adds a missing changelog entry, and the promised definition of signed/unsigned shift.


lrhn (approved):

eernstg:
Thanks! Sorry about forgetting to press 'ENTER' to upload the last changes, next time I'll triple-check that it's done! ;-)

lrhn:
It happens. 
Far. Far. Too. Often. 😁

PR #1485, opened 2021-03-01
Clean up \commentary, \rationale, specify \DefineSymbol

eernstg:
This is a LaTeX clean-up and enhancement PR (no changes to the contents, except for a small change at two locations, which have been marked by a comment from me).

It corrects all `\commentary{}` and `\rationale{}` such that they do not introduce spaces in front or at the end (this is needed in order to avoid spurious extra line breaks when the text fills up the last line of the paragraph). It also reformats the `\commentary{}` and `\rationale{}` to fit in 80 columns.

Finally, it introduces `\DefineSymbol` from the beginning and until & including section 'Enums'. This is a macro that adds the given symbol in the margin, such that it is possible to quickly look up the location where a symbol is defined. Symbols are only shown like this if they are used in more than a few lines of text. So if $C$ is used and we want to see how it's defined, scan backwards in the text: It could be defined just a couple of lines earlier, otherwise it should occur in the margin.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -361,8 +361,6 @@ For undated references, the latest edition of the referenced document (including
 
 \LMHash{}%
 Terms and definitions used in this specification are given in the body of the specification proper.
-Such terms are highlighted in italics when they are introduced, e.g., `we use the term \NoIndex{verbosity} to refer to the property of excess verbiage',
-and add a marker in the margin.
 % End Ecma Boilerplate
 
 
@@ -377,27 +375,28 @@ At this time, non-normative text includes:
 \begin{itemize}
 \item[Rationale]
   Discussion of the motivation for language design decisions appears in italics.
-\rationale{
-Distinguishing normative from non-normative helps clarify what part of the text is binding and what part is merely expository.
+\rationale{%
+Distinguishing normative from non-normative helps clarify
+what part of the text is binding and what part is merely expository.%
 }
 \item[Commentary]
   Comments such as ``\commentary{The careful reader will have noticed that the name Dart has four characters}'' serve to illustrate or clarify the specification, but are redundant with the normative text.
-\commentary{
-The difference between commentary and rationale can be subtle.
+\commentary{%
+The difference between commentary and rationale can be subtle.%
 }
-\rationale{
-Commentary is more general than rationale, and may include illustrative examples or clarifications.
+\rationale{%
+Commentary is more general than rationale,
+and may include illustrative examples or clarifications.%
 }
-\item[Open questions] (\Q{in this font}).
-Open questions are points that are unsettled in the mind of the author(s) of the specification; expect them (the questions, not the authors; precision is important in a specification) to be eliminated in the final specification.
-\Q{Should the text at the end of the previous bullet be rationale or commentary?}
 \end{itemize}
 
 \LMHash{}%
-Reserved words and built-in identifiers (\ref{identifierReference}) appear in {\bf bold}.
+Reserved words and built-in identifiers
+(\ref{identifierReference})
+appear in {\bf bold}.
 
-\commentary{
-Examples would be \SWITCH{} or \CLASS{}.
+\commentary{%
+Examples would be \SWITCH{} or \CLASS{}.%
 }
 
 \LMHash{}%
@@ -413,8 +412,8 @@ Negation is represented by prefixing an element of a production with a tilde.
 Negation is similar to the not combinator of PEGs, but it consumes input if it matches.
 In the context of a lexical production it consumes a single character if there is one; otherwise, a single token if there is one.
 
-\commentary{
-An example would be:
+\commentary{%
+An example would be:%
 }
 
 \begin{grammar}\color{commentaryColor}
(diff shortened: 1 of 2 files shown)


specification/dartLangSpec.tex
```diff
@@ -2337,52 +2519,54 @@ \subsection{Instance Methods}
...
+(\ref{interfaces}).
+It is a compile-time error if $D$ overrides a declaration
 % Note that $m'$ is accessible, due to the definition of 'overrides'.
-which overrides a member signature $m'$
+with member signature $m'$
 from a direct superinterface of $C$
 (\ref{interfaceInheritanceAndOverriding}),
-unless this is a correct member override
+unless $m$ is a correct member override of $m'$
 (\ref{correctMemberOverrides}).
```
eernstg, line 2538: Adjusted the language here. It was a bit too sloppy.

specification/dartLangSpec.tex
```diff
@@ -21009,7 +21570,7 @@ \section*{Appendix: Integer Implementations}
   operands to 32-bit two's complement integers, perform 32-bit
   operations on those, and the resulting 32 bits are interpreted as a
   32-bit unsigned integer. For example, \code{-1\,\ltlt\,1} evaluates
-  to 4294967294 (also known as \code{(-2).toUnsigned(32)}. The right
+  to 4294967294 (also known as \code{(-2).toUnsigned(32)}). The right
```
eernstg, line 21573: Typo fix: Added missing ')'.

specification/dartLangSpec.tex
```diff
@@ -986,46 +1047,69 @@ \section{Errors and Warnings}
...
+Typically, once a compile-time error is thrown and $A$ is suspended,
+$A$ will then be terminated.
 However, this depends on the overall environment.
 A Dart engine runs in the context of an \Index{embedder},
-a program that interfaces between the engine and the surrounding computing environment.
-The embedder will often be a web browser, but need not be; it may be a C++ program on the server for example.
-When an isolate fails with a compile-time error as described above, control returns to the embedder, along with an exception describing the problem.
+a program that interfaces between the engine and
+the surrounding computing environment.
+The embedder will often be a web browser, but need not be;
```
lrhn, line 1097: Phew, good thing it doesn't need to be one, since it *never is*.
(Might be worth fixing this sentence 😁)

specification/dartLangSpec.tex
```diff
@@ -8320,54 +8719,63 @@ \subsection{Strings}
...
 print("A simple sum: 2 + 2 = " +
             2 + 2);
 \end{dartCode}
 
-\rationale{
-which this prints 'A simple sum: 2 + 2 = 22' rather than 'A simple sum: 2 + 2 = 4'.
-However, the use of the concatenation operation is still discouraged for efficiency reasons.
-Instead, the recommended Dart idiom is to use string interpolation.
+\rationale{%
+which prints 'A simple sum: 2 + 2 = 22' rather than
```
lrhn, line 8760: "which prints" -> "which would print". (Now that I read it).

specification/dartLangSpec.tex
```diff
@@ -8320,54 +8719,63 @@ \subsection{Strings}
...
+'A simple sum: 2 + 2 = 4'.
+However, the use of the concatenation operation is still discouraged
```
lrhn, line 8762: That paragraph should be rewritten or removed.
It should be used judiciously, but it is not *discouraged* when used correctly.

specification/dartLangSpec.tex
```diff
@@ -10552,8 +10974,10 @@ \subsection{Throw}
 the stack trace $t$ is stored on $v$ so that it will be returned
 by the $v$'s \code{stackTrace} getter
```
lrhn, line 10993: (Maybe say: `by the \code{stackTrace} getter inherited from \code{Error}.`, because it might be overridden in a subclass, so it's not the `stackTrace` getter of *v*).

lrhn (approved):

specification/dartLangSpec.tex
eernstg: Adjusted this sentence to drop the web browser part, and changed the word 'embedder' to 'runtime'.

You could say that 'runtime' is more ambiguous than 'runtime system', but I think we'll be OK because it is defined explicitly at this location, and it can be looked up in the index. WDYT?

specification/dartLangSpec.tex
eernstg: Right, done.

specification/dartLangSpec.tex
eernstg: Adjusted the text to say that string interpolation 'recommended', which leaves concatenation as "unspecified, but not exactly recommended".

specification/dartLangSpec.tex
eernstg: Good catch, done!

specification/dartLangSpec.tex
lrhn: LGTM

eernstg: Thanks!

PR #1486, opened 2021-03-02
Fix upper range in example

mnordine:


jakemac53 (approved):

jakemac53:
Ah  yes I shortened the above example but didn't update the loop :D

PR #1489, opened 2021-03-04
Specify identifier references denoting extension members

eernstg:
This corrects an omission in the recent update where extension methods where specified: How to find the static type and execute an identifier reference, when it denotes an extension member.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -16352,10 +16352,9 @@ The evaluation of $e$ proceeds as follows:
 \item
   If $D$ is an instance method declaration in an extension $E$
   with type parameters \List{X}{1}{s},
-  then $e$ evaluates to the extension method closurization
-  \code{$E$<\List{t}{1}{s}>(\THIS).\id}
-  (\ref{extensionMethodClosurization}),
-  where \List{t}{1}{s} are the current bindings of said type parameters.
+  then $e$ evaluates to the result of the extension method closurization
+  \code{$E$<\List{X}{1}{s}>(\THIS).\id}
+  (\ref{extensionMethodClosurization}).
 \item
   If $D$ is a local variable $v$
   (\commentary{which can be a formal parameter})
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -41,6 +41,7 @@
 %   'class' scope.
 % - Add specification of `>>` in JavaScript compiled code, in appendix.
 % - Integrate the specification of extension methods into this document.
+% - Specify identifier references denoting extension members.
 %
 % 2.7
 % - Rename non-terminals `<...Definition>` to `<...Declaration>` (e.g., it is
@@ -16245,9 +16246,18 @@ Let $D$ be the declaration yielded by the lexical lookup of \id.
   such that $v$ is known to have type $S$,
   in which case the static type of $e$ is $T$.
 \item
+  % Extension getter invocation by scope.
+  If $D$ is a declaration of an instance getter in an extension $E$
+  then the static type of $e$ is the return type of $D$.
+\item
+  % Extension method tear-off by scope.
+  If $D$ is a declaration of an instance method
+  in an extension $E$ then the static type of $e$ is
+  the function type of $D$.
+\item
   \commentary{%
     A lexical lookup will never yield a declaration $D$
-    which is an instance member.%
+    which is an instance member of a class.%
   }
 \end{itemize}
 \vspace{-1ex}
@@ -16334,6 +16344,18 @@ The evaluation of $e$ proceeds as follows:
   then $e$ evaluates to the function object obtained by closurization
   (\ref{functionClosurization})
   of $D$.
+\item
+  If $D$ is an instance getter declaration in an extension $E$,
+  then $e$ evaluates to the result of invoking said getter
+  with the current binding of \THIS{}, and
+  the current bindings of the type parameters declared by $D$.
+\item
+  If $D$ is an instance method declaration in an extension $E$
+  with type parameters \List{X}{1}{s},
+  then $e$ evaluates to the extension method closurization
+  \code{$E$<\List{t}{1}{s}>(\THIS).\id}
+  (\ref{extensionMethodClosurization}),
+  where \List{t}{1}{s} are the current bindings of said type parameters.
 \item
   If $D$ is a local variable $v$
   (\commentary{which can be a formal parameter})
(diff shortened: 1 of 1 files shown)


eernstg:
@lrhn, WDYT?

specification/dartLangSpec.tex
```diff
@@ -16244,10 +16245,19 @@ \subsection{Identifier Reference}
...
   in which case the static type of $e$ is $T$.
+\item
+  % Extension getter invocation by scope.
+  If $D$ is a declaration of an instance getter in an extension $E$
+  then the static type of $e$ is the return type of $D$.
+\item
+  % Extension method tear-off by scope.
+  If $D$ is a declaration of an instance method
+  in an extension $E$ then the static type of $e$ is
+  the function type of $D$.
```
lrhn, line 16256: Do we need to mention type parameters of *E* or are they assumed to be part of *E* (because we say "extension *E*", not "extension declaration *E*")?

specification/dartLangSpec.tex
```diff
@@ -16244,10 +16245,19 @@ \subsection{Identifier Reference}
...
 \item
   \commentary{%
     A lexical lookup will never yield a declaration $D$
-    which is an instance member.%
+    which is an instance member of a class.%
```
lrhn, line 16260: Why not? (Github is preventing me from seeing most of the context here, but this is "Unqualified Invocation", right?)

Do we assume that it's *already* rewritten to `this.id(...)`? (If so, say so).


Also, while we are here, above (where Github won't let me see) it says:
```latex
\item
  When $D$ is a type declaration, that is,
  a declaration of a class, mixin, type alias, or type parameter,
  the following applies:
```
It needs to also mention `enum` (unless that's implicitly declaring a class, which it probably, technically, is. but it's confusing when a "declaration of a class" and a "class declaration" is not the same thing).

specification/dartLangSpec.tex
```diff
@@ -16334,6 +16344,17 @@ \subsection{Identifier Reference}
...
+  If $D$ is an instance getter declaration in an extension $E$,
+  then $e$ evaluates to the result of invoking said getter
+  with the current binding of \THIS{}, and
+  the current bindings of the type parameters declared by $D$.
+\item
+  If $D$ is an instance method declaration in an extension $E$
+  with type parameters \List{X}{1}{s},
+  then $e$ evaluates to the result of the extension method closurization
+  \code{$E$<\List{X}{1}{s}>(\THIS).\id}
+  (\ref{extensionMethodClosurization}).
```
lrhn, line 16357: I'd say:

> ... to the tear-off of *D* closing over the current values of `this` and type parameters of *E*.

Do we need to handle instantiated tear-off here, or is that (implicitly) handled in type inference?

specification/dartLangSpec.tex
```diff
@@ -16334,6 +16344,17 @@ \subsection{Identifier Reference}
...
 \item
   If $D$ is a local variable $v$
   (\commentary{which can be a formal parameter})
```
lrhn, line 16360: Same comments about mentioning `enum` above and 
> ```latex
> \item
>   \commentary{%
>     A lexical lookup will never yield a declaration $D$
>     which is an instance member.%
>   }
> ```
looking wrong.

specification/dartLangSpec.tex
eernstg: I don't think so: Given that a lexical lookup has yielded a declaration of a member of an extension declaration $E$, the current location is inside some member of $E$ as well, and this means that the type arguments that occur free in the function type of $D$ are the available denotations of those types (we don't have any actual type arguments that we can use in a substitution, and it's OK to use the type parameters because they are in scope).

I corrected 'extension $E$' to 'extension declaration $E$', because it is actually a declaration.

It is still possible that we will need to perform a generic function instantiation on $e$ (both for a getter invocation and for a method tear-off), but we don't mention this: It would have to be mentioned for _every_ form of expression because they could all occur in a context where a non-generic function type is the context, and they could all have a type which is a generic function type.

specification/dartLangSpec.tex
eernstg: The enclosing section is still 'Identifier Reference', not 'Unqualified Invocation'.

The reason why lexical lookup won't ever yield a declaration that is an instance member of a class is that it yields a member signature in that case.

I did it like that because this ensures that we have a simple rule for when to add `\THIS{}` in front of an expression (identifier reference or unqualified invocation), and that's exactly when the result from the lexical lookup is a member signature.

(That member signature is actually never used, it's just taken as a signal that we should prepend `\THIS.` to the expression and proceed with that; so I've changed that in another PR, reviewed by Leaf, such that we just yield a "special value" indicating that we should prepend `\THIS.`).

> Do we assume that it's already rewritten to `this.id(...)`?

No, lexical lookup has selected a declaration $D$, and that's never a member of a class interface, because (both for an inherited member and for a locally declared member) we always obtain a member signature in that case. That's the case approx. 10 lines further down from here.

specification/dartLangSpec.tex
eernstg: About 'closing over'. We have explicitly specified the syntax `E<T1, .. Tk>(e).id` as an extension method closurization, don't you think we should stick to that syntax rather than switching to the less-explicitly defined 'closing over'?

As mentioned at line 16256, we have a general rule about generic function instantiation that we rely on in many locations.

specification/dartLangSpec.tex
eernstg: Added 'enumerated type' to the list of kinds of type declarations that we may encounter.

specification/dartLangSpec.tex
eernstg: I don't understand why I'd need to mention enumerated types in the item about local variables?

The reason why the lexical lookup will always yield a member signature rather than a declaration in the case where the selected entity is an instance member is unchanged from line 16260.

eernstg (commented):
Thanks for comments, review response added!

specification/dartLangSpec.tex
lrhn: So we did a hack in "lexical lookup" to not actually find "a declaration", even when there literally is an instance declaration with that name in the surrounding class. Might be worth reminding the reader about :)

specification/dartLangSpec.tex
lrhn: Referring to line 16216:
> ```latex
>  If $D$ declares a class, mixin, type alias or type parameter,
>  the static type of $e$ is \code{Type}.
> ```
which should also mention `enum` declarations.

lrhn (approved):

PR #1501, opened 2021-03-08
Adjust lexical lookup to include implicit extension method invocation

eernstg:
The current specification of lexical lookup specifies a compile-time error in certain cases where it should specify that a name `$n$` is instead transformed into `\THIS.$n$` (which is an identifier reference or a part of an unqualified function invocation). This PR fixes that.

Note that lexical lookup yields a member signature as a "signal to caller" that the expression should have `\THIS{}` prepended. The static analysis will then proceed from scratch on the transformed expression, and this will allow us to detect that it is a regular getter/method invocation on `\THIS{}` or an implicit extension getter/method invocation.

Update: The information about _which_ member signature was yielded by a lookup was ignored (we just proceed with `this.id` resp. `this.id<...>(...)`), so I changed the lookup to yield a special value `PREPEND_THIS`.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -15998,7 +15998,7 @@ Now proceed as described in the first applicable case from the following list:
   Otherwise, the lexical lookup yields $D$.
 \item
   Consider the case where $D$ is an instance member declaration in
-  a class \DefineSymbol{C}.
+  a class $C$.
   The lexical lookup then yields the member signature named $n$
   in the interface of $C$.
   \commentary{%
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -15861,20 +15861,32 @@ the following errors apply:
 \item
   Consider the case where $\ell$ has access to \THIS,
   and $\ell$ occurs in a class \DefineSymbol{C}.
-  A compile-time error occurs
-  if the interface of $C$ does not have a member named $n$,
-  and there is no accessible extension $E$
-  such that $C$ matches the \ON{} type of $E$
-  and $E$ has an instance member named $n$.
+  A compile-time error occurs in each of the following two situations:
+  \begin{itemize}
+  \item
+    The interface of $C$ does not have a member named $n$,
+    but it does have a member with basename \id.
+  \item
+    The interface of $C$ does not have a member with basename \id,
+    and there is no accessible extension $E$,
+    such that $C$ matches the \ON{} type of $E$,
+    and $E$ has an instance member named $n$.
+  \end{itemize}
 \item
   \BlindDefineSymbol{E, T}%
   Consider the case where $\ell$ has access to \THIS,
   and $\ell$ occurs in an extension $E$ with \ON{} type $T$.
-  A compile-time error occurs
-  if the interface of $T$ does not have a member named $n$,
-  and there is no accessible extension $E_1$
-  such that $T$ matches the \ON{} type of $E_1$
-  and $E_1$ has an instance member named $n$.
+  A compile-time error occurs in the following two situations:
+  \begin{itemize}
+  \item
+    The interface of $T$ does not have a member named $n$,
+    but it does have a member with basename \id.
+  \item
+    The interface of $T$ does not have a member with basename \id,
+    and there is no accessible extension $E_1$,
+    such that $T$ matches the \ON{} type of $E_1$,
+    and $E_1$ has an instance member named $n$.
+  \end{itemize}
 \end{itemize}
 \vspace{-2ex}
 \EndCase
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -15845,6 +15845,39 @@ even though it may have the name \id{} when $n$ is \code{\id=} or vice versa.
 That situation may cause an error, as specified below.%
 }
 
+% Even when we have found a declaration, it may have the wrong name.
+\LMHash{}%
+\Case{$D$ exists}
+In this case, at least one declaration with basename \id{}
+is in scope at the location $\ell$.
+
+\LMHash{}%
+It is a compile-time error if the name of $D$ is not $n$,
+unless $D$ is an instance member or a local variable
+(\commentary{which may be a formal parameter}).
+
+%% TODO(eernst): Come NNBD, `this` is accessible in a `late` instance variable
+%% initializer, so they must also be included in the first case.
+\LMHash{}%
+% Has chosen instance member, wants getter/setter, has only the opposite.
+When $\ell$ has access to \THIS{}
+(\ref{classes})
+and is located in a class or mixin \DefineSymbol{A},
+it is a compile-time error if $D$ is an instance member,
+and the interface of $A$ does not have a member named $n$.
+% Has chosen extension instance member, wants getter/setter,
+% has only the opposite.
+When $\ell$ has access to \THIS{}
+and occurs in an extension \DefineSymbol{E},
+it is a compile-time error if $D$ is an instance member,
+and $E$ does not declare a member named $n$.
+% No access to this.
+When $\ell$ does not have access to \THIS,
+it is a compile-time error if $D$ is an instance member.
+% An instance variable initializer is in the body scope, so it is
+% actually possible for $D$ to be an instance member here.
+\EndCase
+
 \LMHash{}%
 \Case{$D$ does not exist}
 When no declaration with basename \id{} is in scope at the location $\ell$,
@@ -15862,18 +15895,23 @@ the following errors apply:
   (\ref{classes}).
 \item
   Consider the case where $\ell$ has access to \THIS,
-  and $\ell$ occurs in a class \DefineSymbol{C}.
+  and $\ell$ occurs in a class or a mixin \DefineSymbol{A}.
   A compile-time error occurs in each of the following two situations:
   \begin{itemize}
   \item
-    The interface of $C$ does not have a member named $n$,
+    The interface of $A$ does not have a member named $n$,
     but it does have a member with basename \id.
   \item
-    The interface of $C$ does not have a member with basename \id,
-    and there is no accessible extension $E$,
-    such that $C$ matches the \ON{} type of $E$,
-    and $E$ has an instance member named $n$.
+    The interface of $A$ does not have a member with basename \id,
+    and there is no accessible extension $E$
+    such that $A$ matches the \ON{} type of $E$,
+    and $E$ has an instance member with basename \id.
   \end{itemize}
+  \commentary{%
+    In other words, we get an error if
+    we find a getter, but we are looking for a setter, or vice versa;
+    and we get an error if we cannot find anything at all.%
+  }
 \item
   \BlindDefineSymbol{E, T}%
   Consider the case where $\ell$ has access to \THIS,
@@ -15894,45 +15932,11 @@ the following errors apply:
 \EndCase
 
 \commentary{%
-So it is an error if there is no member with the right basename,
-and there is also no applicable extension member with the right basename.
-It is also an error if there is one such member,
-but it is a non-setter and we are looking for a setter,
-or vice versa.%
+We could describe these rules as having a built-in priority system:
+The lexically enclosing scopes have the highest priority;
+members of the interface of \THIS{} have the next priority;
+and implicitly invoked extension methods have the lowest priority.%
 }
-
-% Even when we have found a declaration, it may have the wrong name.
-\LMHash{}%
-\Case{$D$ exists}
-In this case, at least one declaration with basename \id{}
-is in scope at the location $\ell$.
-
-\LMHash{}%
-It is a compile-time error if the name of $D$ is not $n$,
-unless $D$ is an instance member or a local variable
-(\commentary{which may be a formal parameter}).
-
-%% TODO(eernst): Come NNBD, `this` is accessible in a `late` instance variable
-%% initializer, so they must also be included in the first case.
-\LMHash{}%
-% Has chosen instance member, wants getter/setter, has only the opposite.
-When $\ell$ has access to \THIS{}
-(\ref{classes})
-and is located in a class \DefineSymbol{C},
-it is a compile-time error if $D$ is an instance member,
-and the interface of $C$ does not have a member named $n$.
-% Has chosen extension instance member, wants getter/setter,
-% has only the opposite.
-When $\ell$ has access to \THIS{}
-and occurs in an extension \DefineSymbol{E},
-it is a compile-time error if $D$ is an instance member,
-and $E$ does not declare a member named $n$.
-% No access to this.
-When $\ell$ does not have access to \THIS,
-it is a compile-time error if $D$ is an instance member.
-% An instance variable initializer is in the body scope, so it is
-% actually possible for $D$ to be an instance member here.
-\EndCase
 
 \rationale{%
 We are always looking up \emph{both} \id{} and \code{\id=},
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -12384,8 +12384,8 @@ When the lexical lookup of \id{} yields an import prefix,
 a compile-time error occurs.
 \EndCase
 
-\Case{Lexical lookup yields \PrependThisValue}
-When the lexical lookup of \id{} yields \PrependThisValue,
+\Case{Lexical lookup yields nothing}
+When the lexical lookup of \id{} yields nothing,
 $i$ is treated as
 (\ref{notation})
 the ordinary method invocation
@@ -14541,7 +14541,7 @@ Perform a lexical lookup of \code{\id=} from the location of \id.
   Otherwise, a compile-time error occurs,
   unless the static type of $e$ is assignable to the parameter type of $D$.
 \item
-  When the lexical lookup yields \PrependThisValue,
+  When the lexical lookup yields nothing,
   $a$ is treated as
   (\ref{notation})
   \code{\THIS.\id{} = $e$}.
@@ -14602,7 +14602,7 @@ Perform a lexical lookup of \code{\id=} from the location of \id.
 \item
   \commentary{%
     The case where the lexical lookup of \code{\id=}
-    yields \PrependThisValue{} cannot occur,
+    yields nothing cannot occur,
     because that case is treated as
     \code{\THIS.\id{} = $e$},
     whose evaluation is specified elsewhere.%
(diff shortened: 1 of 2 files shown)


specification/dartLangSpec.tex
```diff
@@ -15945,23 +15951,46 @@ \subsection{Lexical Lookup}
...
-    and also if there is one such member,
-    but it is a non-setter and we are looking for a setter,
-    or vice versa.%
-  }
+  Consider the case where $\ell$ has access to \THIS,
+  and $\ell$ occurs in a class \DefineSymbol{C}.
+  A compile-time error occurs in each of the following two situations:
+  \begin{itemize}
+  \item
+    The interface of $C$ does not have a member named $n$,
```
leafpetersen, line 15959: I don't know how to think about these checks here.  This feels very non-modular.  Why is it useful to catch some (but not all) of these kinds of errors here?  Why do we not just say that lexical lookup failed, and let the caller of lexical lookup figure out what to do?

specification/dartLangSpec.tex
```diff
@@ -15861,7 +15863,11 @@ \subsection{Lexical Lookup}
 
 \LMHash{}%
 A lexical lookup yields a result which is
-a declaration, an import prefix, or a member signature.
+a declaration, an import prefix,
+or the special value
+\Index{\PrependThisValue},
```
leafpetersen, line 15868: Why signal this with a special value instead of just saying that "lexical lookup fails"?  That seems much more straightforward.  Either lexical lookup succeeds and yields a declaration, or it fails and we try to prepend `this`?

specification/dartLangSpec.tex
```diff
@@ -15945,23 +15951,46 @@ \subsection{Lexical Lookup}
...
+    and $E_1$ has an instance member named $n$.
+  \end{itemize}
 \end{itemize}
 \vspace{-2ex}
 \EndCase
 
+\commentary{%
+So it is an error if there is no member with the right basename,
+and there is also no applicable extension member with the right basename.
+It is also an error if there is one such member,
```
leafpetersen, line 15989: "one or more such members"?

specification/dartLangSpec.tex
eernstg: That's tempting, but it's not obvious to me that it would be a simplification.

The idea is that we handle the compile-time errors in one location (the case list starting at line 15941), such that there will not be any compile-time errors caused by a lookup failure at the locations where lexical lookup is used. (We never specify propagation of errors, so when the "client" says `perform a lexical lookup for the name $n$` we proceed with the outcome and do not mention "but we drop any further processing of this expression if an error occurred").

So the good thing about this organization is that we mention all the errors of lexical lookups in one piece of text.

I think we do need to mention at least some errors here, and I think we'd have to repeat the same errors "at the client location" if we just report a failure of the lookup, covering all the real errors plus the situations where we will prepend `this` and try again.

We have 50 occurrences of `lexical lookup`, 30 of them outside the `Lexical Lookup` section, so it is likely to cause a substantial duplication of each of these phrases specifying when the failure is actually an error.

We could try to do it, it is possible that we can rule out some of the errors for some occurrences of `lexical lookup` outside the `Lexical Lookup` section.

specification/dartLangSpec.tex
eernstg: We would have to repeat the same rule about 30 times: It is an error if we find a getter but we're looking for a setter, and so on.

For instance, an identifier reference `id` would have that rule (at several locations in section `Identifier Reference`), and so would an unqualified  function invocation `id<...>(...)`, and an assignment `id = e`, and a composite assignment `id += e`, and so on.

It is true that we do not catch all errors: When there is no declaration with the requested basename in scope, and there is access to `this`, and the interface of the enclosing class does not have a member with the requested basename, we may still have multiple applicable extensions.

Do you think we have other missing errors? I think this one isn't so bad, because any reports about the error would deal with exactly the same problem even if we had had an explicit `this` in the first place. So it isn't going to be wrong if we simply proceed with `this.id...` and get the error later on.

specification/dartLangSpec.tex
eernstg: I think we can't actually have more than one: We can have two members with the given basename, but then we must have both a getter and a setter. ;-)

eernstg (commented):
Review response added.

eernstg:
@leafpetersen, WDYT?

specification/dartLangSpec.tex
leafpetersen: I'm not sure I understand.  Currently, you claim that "lexical lookup cannot fail".  But of course it can, you just mean that either lexical lookup causes an error, or it produces something, and that something can be either a declaration or a special value saying "lexical lookup failed to find something".    I'm saying, why not just get rid of the special value and say that one of three things happen:
  - lexical lookup causes an error (as it may now)
  - lexical lookup finds a declaration (as it may now)
  - lexical lookup fails to find a declaration (what you're calling PrependThisValue)

This seems exactly isomorphic, but avoids indirecting through an additional definition (PrependThisValue)

specification/dartLangSpec.tex
leafpetersen: Ok, I think I might get what you're saying.  Paraphrasing, I think what you are saying is that there are really two different kinds of errors we might be concerned about:
  - We didn't find anything with the right name
     - This error cannot be handled here, because there are additional lookup steps that we need to follow if lexical lookup fails
  - We found something with the right name, but it was a setter and we wanted a getter or vice versa
     -   We can't treat this as just another case of "lexical lookup failed to find what we wanted" because we want to stop subsequent non-lexical lookup from happening.

Every caller of "lexical lookup" already has to deal with the first situation in which lexical lookup doesn't find what we want (which was why I suggested not checking things here) but what you're saying is that even when lexical lookup succeeds, we may still have an error, and it's better to catch that in one place, here.  

Is that a fair summary?

specification/dartLangSpec.tex
leafpetersen: So there's an implicit assumption that this code is non-erroneous?  e.g. doesn't have duplicate definitions?

specification/dartLangSpec.tex
eernstg: I very nearly agree. ;-)

The approach I've taken will find the errors that we can characterize as lookup failures ("we don't have anything suitable with that name"), and treat them as compile-time errors.

The 'caller' can just say `perform a lexical lookup for the name $n$`, and the outcome is a declaration, an import prefix, or \PrependThis{}. The caller doesn't have to worry about "we don't have anything...", because we never describe propagation of compile-time errors, and the caller _will_ be able to trust that something suitable was found.

When the outcome is \PrependThis{} we know that the correct next step is to process `this.id` rather than `id`. At this point we do know that there is an instance member or an applicable extension method with the requested name; the only error that we can still encounter is that there is more than one applicable extension method.

One nice property with this approach is that it keeps _everything_ about getter/setter bundling during lookups in one location, namely here: If we do find a declaration with the right basename in an enclosing scope, but the nearest one is a getter and we requested a setter (or vice versa), we get an error. If there's nothing in any enclosing scope with that basename, but we have the same getter/setter failure with instance members, we get an error. If there is no instance member with the requested basename, but we get the same getter/setter failure with extension methods, we get an error.

So the rest of the spec doesn't need to know that getters and setters are bundled, or that basenames matter.

Returning to the two kinds of errors you mentioned:

> * We didn't find anything with the right name

Agreed that this is always an error, can be handled here.

> * We found something with the right name, but it was a setter and we wanted a getter or vice versa

This is also always an error, can be handled here.

But in both cases we have to use the right perspective: "We didn't find anything" means (1) nothing had the right basename in enclosing scopes, and (2) nothing had the right basename in the interface of `this` (if applicable), and (3) nothing had the right basename in applicable extension declarations.

"We found something with the right name, but .." means (1) the nearest enclosing scope with a declaration with the right basename had the getter/setter failure, or (2) enclosing scopes have no such basename, but the interface of `this` (if applicable) had the getter/setter failure, or (3) enclosing scopes and the interface of `this` didn't have that basename, but an applicable extension declaration had the getter/setter failure.

When none of all those errors occurred, we can say that the lexical lookup 'completes normally', and in this case we have to remember that we must prepend `this` in some cases, and we must avoid prepending `this` in other cases. In the former case we have a declaration or an import prefix; in the latter we just know that we must prepend `this` and restart the analysis.

We could of course say that we don't detect errors here, we just allow the outcome of lexical lookups to be "failed!".

But I don't think that would be helpful. For instance, if there is a top-level setter `foo=` and we are looking for a `foo` getter, we certainly shouldn't receive the outcome "failed!" from the lexical lookup, prepend `this`, and succeed in the search for an instance or extension method. We must maintain that there is an error, and it's the setter/getter failure at the top-level.

eernstg (commented):
Oops, forgot to click 'finish your review'.

eernstg:
Swapped the `$D$ exists` and `$D$ does not exist` cases (I think it's more readable like that), and fixed a bug: We deal with a `class or mixin` not just a `class`.

specification/dartLangSpec.tex
```diff
@@ -15862,7 +15866,11 @@ \subsection{Lexical Lookup}
 
 \LMHash{}%
 A lexical lookup yields a result which is
-a declaration, an import prefix, or a member signature.
+a declaration, an import prefix,
+or the special value
+\Index{\PrependThisValue},
+which indicates that the lookup should be transformed into
+a member access on \THIS{}.
 A lexical lookup can not fail
```
leafpetersen, line 15874: This is a really confusing comment.  It most certainly can fail, right?  What does it mean to say that a compile-time error can occur but it cannot fail?

specification/dartLangSpec.tex
```diff
@@ -15980,9 +15955,9 @@ \subsection{Lexical Lookup}
 % Has chosen instance member, wants getter/setter, has only the opposite.
 When $\ell$ has access to \THIS{}
 (\ref{classes})
-and is located in a class \DefineSymbol{C},
+and is located in a class or mixin \DefineSymbol{A},
 it is a compile-time error if $D$ is an instance member,
```
leafpetersen, line 15959: What is D here?  As best I can tell, at this point in the text, we haven't said anything about finding D or what it is.  Did this end up out of order in a refactor?

specification/dartLangSpec.tex
```diff
@@ -15930,39 +15938,6 @@ \subsection{Lexical Lookup}
...
-    but it is a non-setter and we are looking for a setter,
-    or vice versa.%
-  }
-\end{itemize}
-\vspace{-2ex}
-\EndCase
-
 % Even when we have found a declaration, it may have the wrong name.
 \LMHash{}%
 \Case{$D$ exists}
```
leafpetersen, line 15943: As below, what is D?

leafpetersen: Never mind - this seems to be something broken with github's preview.  It's showing line 15876 followed immediately by line 15938 with no option to expand the lines between, and I think that's where the missing text is.  I think I'm going to have to pull this down to my machine to review tomorrow.

specification/dartLangSpec.tex
leafpetersen: > We could say that lexical lookup failed to find a declaration, but that doesn't match the situation precisely: lots of situations where we fail to find a declaration are an error, but PrependThisValue implies that a number of error checks have been performed and a suitable declaration _does_ exist.

I don't really agree with this, at least if I'm understanding what's here.  I think saying that lexical lookup failed does match the situation precisely.  This algorithm either finds a single declaration in the lexical scope; or it emits an error because there is no possible suitable thing in either the lexical or the instance scope; or it says "there's nothing in the lexical scope, but there's something that might (or might not) work in the instance scope, go do some additional work".  The last is what you are calling "PrependThisValue".  It still seems much cleaner to me to just call these "Found a declaration", "Caused an error", and "Didn't find a declaration".

leafpetersen: > So the 'caller' doesn't have to take into account that the lexical lookup failed to find anything. When it actually fails to find anything, a compile-time error has occurred.

Nothing I'm suggesting changes this.

leafpetersen: > because 'fails to find a declaration' would presumably also cover at least some of the many error case

I'm not proposing that.  Why would it presumably also cover that?

leafpetersen: > because 'fails to find a declaration' would presumably also cover at least some of the many error cases, and PrependThisValue is only yielded when we have actually found a member with the requested name and established that prepending `this` is appropriate.

If you prefer, you could say "fails to return a declaration"?  But I actually think "fails to find a declaration" is just fine.  It found something that might (or might not) be the right thing, but it's not returning anything because it doesn't know whether there is an unambiguous answer.

eernstg:
I didn't get around to this today, but I'll comment on the open threads tomorrow. I think I might be able to reduce the amount of checks performed before "returning" `\PrependThis{}`.

specification/dartLangSpec.tex
```diff
@@ -12483,9 +12485,10 @@ \subsubsection{Unqualified Invocation}
 
 \commentary{%
 This occurs when the lexical lookup has determined that
-$i$ must invoke an instance member,
+$i$ must invoke an instance member of a class or an extension,
 and the location of $i$ can access \THIS{},
-and the interface of the enclosing class has a member named \id.
+and the interface of the enclosing class has a member named \id,
```
leafpetersen, line 12490: Why do we bother to check and assert this here?  As best I can tell, section \ref{ordinaryInvocation} which we delegate to does this check anyway.  So it seems like we should just not talk about this at all.  That is, we should say "When lexical lookup fails to find a declaration, i is treated as this.i".  And then if i is not in the interface for this, then ordinaryInvocation will handle it just fine.  What am I missing?

specification/dartLangSpec.tex
leafpetersen: > As I mentioned, we're using lexical lookup in about 30 locations, and I don't think it's helpful to repeat that many times which errors we should raise when nothing was found (or we found a getter and wanted a setter, etc.).

This is part of what I'm not getting entirely.  I agree that you probably need to catch the getter/setter mismatch here (are we sure we're specifying this as implemented, BTW?).  But the case where it's not found I think is already handled by the logic for `this.id`, so why catch it here?

eernstg: > This algorithm either finds a single declaration in the lexical scope; or it emits an
> error because there is no possible suitable thing in either the lexical or the instance
> scope; or it says "there's nothing in the lexical scope, but there's something that
> might (or might not) work in the instance scope, go do some additional work"

>  It still seems much cleaner to me to just call these "Found a declaration",
> "Caused an error", and "Didn't find a declaration".

I don't think we should start specifying propagation of errors, we otherwise never do that. So if the lexical lookup encountered an error then we just silently ignore it at the 'call site', exactly like we'd ignore a type error in `b` when we specify `if (b) S`: we require that `b` has a type assignable to `bool`, but we never mention what to do if `b` has no type because it isn't type correct, so it's like a `throw` that we don't catch at the call site. So I wouldn't consider "Caused an error" to be the result of a lexical lookup in a way that is similar to "Found a declaration": The call site needs to talk about the latter, but would not mention the former.

"Found a declaration" matches the situation where the lexical lookup yields a declaration, so that's fine. We're currently handling import prefixes separately, but that's just because it seems somewhat odd to consider an import directive as a declaration of an import prefix, especially because we can have the same prefix on multiple imports.

"Didn't find a declaration" matches the situation where there was no declaration with the requested basename at all (enclosing lexical scopes + interface of `this` + extension methods), but that gives rise to an error (so, again, we ignore it at the call site).

The remaining situations are those where we _have_ found a declaration with the requested name in the interface of `this`, or `this` has no such basename and we found one in an extension. Is the best possible name for this situation really "Didn't find a declaration"?

specification/dartLangSpec.tex
eernstg: I rewrote this to emphasize the handling of lexical lookups, avoiding the words "cannot fail".

specification/dartLangSpec.tex
eernstg: `$D$` is defined before the commentary. It could be helpful to have the generated PDF available: It shows symbols in the right margin at the point where they are introduced (in the parts of the spec where I've done that, including this section).

specification/dartLangSpec.tex
eernstg: Yes, I've been looking for better expansion support in github reviews, too.

specification/dartLangSpec.tex
eernstg: I'm just explaining exactly what we know to be true here. I could delete the whole commentary.

However, we can't treat `i` as `this.i` if we found a setter, but there was no getter. In that case actually did not find any declarations named `i` (OK, there could have been a getter in an enclosing scope, but if we find a setter then we insist on using the getter from the same scope), so "When lexical lookup fails to find a declaration" must be interpreted as "When lexical lookup did not encounter a compile-time error, and it did not find a declaration named `i`".

I think it's simpler to determine precisely when we need to prepend `this`, and then communicate that fact unambiguously to the 'caller'.

eernstg:
At this point I've concentrated on revising the text in the spec rather than explaining my view of things in the comment threads. ;-)

In particular, the text no longer says a `lexical lookup cannot fail`. Instead, it explains that property by indicating how to deal with a lexical lookup (that is, how to "call" it) in other parts of the specification.

I added commentary in a couple of places where we have "<a declaration> does not have name <some name>", to spell out that this is concerned with setter/getter mismatches.

I managed to remove all the `this` related error detection: The errors will be raised on the expression where `this.` has been prepended, and they will have the same nature (so we are not misrepresenting what went wrong when we just change `\id` to `this.\id`, and proceed with that).

The part that I hadn't noticed is that we _do_ handle getter/setter mismatches in one more location: When invoking an extension method `m` implicitly, we first check that the interface of `this` does not have a member with _basename_ `m`.

So I'm now convinced that we can simply bail out here, yield `\PrependThis`, and leave all the error checking to `this.\id`.

eernstg:
In the latest patchset I'm consistently relying on error handling applied to `this.n`, which gets rid of quite some lines of text.

specification/dartLangSpec.tex
leafpetersen: > Is the best possible name for this situation really "Didn't find a declaration"?

Yes.  Because the algorithm you are specifying is **lexical** lookup.  Not **lexical and instance** lookup.  I get that you want to do some error checking in this algorithm, and that we just silently propagate error cases.  But if we are doing lexical lookup, and we ignore the error cases, then either it found something in the lexical scope or it didn't.  If it didn't, we lookup in the instance scope.  To use your terminology, you are specifying this as returning an element of `datatype LexicalDeclaration of declaration | PrependThis`.  I'm saying that that datatype is precisely isomorphic to `Declaration option` so why go to the trouble to introduce a new datatype instead of just using the `option` datatype?  The new datatype is just one more thing for the reader to load into their head.

leafpetersen (approved):
LGTM.  I'm still not a fan of the PrependThisValue formulation, but I don't really see it as worth keeping on about.  thanks!

specification/dartLangSpec.tex
eernstg: Resolving this thread; I think the contents are covered by later threads.

eernstg (commented):
\PrependThis{} is gone! Landing.

PR #1513, opened 2021-03-11
Adjust feature spec for small features 21Q1

eernstg:
The grammar already allows the type arguments on metadata. Changed the feature spec accordingly.

### DIFF
--- a/accepted/future-releases/small-features-21Q1/feature-specification.md
+++ b/accepted/future-releases/small-features-21Q1/feature-specification.md
@@ -44,21 +44,15 @@ However, the *grammar* does not allow type arguments, meaning that it's not poss
 
 There is no technical reason for this restriction. It was just simpler, and probably didn't seem necessary at the time metadata was introduced. It does now.
 
-The only change is in the grammar, from
+There is no change in the grammar, it already allows the type arguments:
 
 ```
-<metadatum> ::= \gnewline{}
+<metadatum> ::=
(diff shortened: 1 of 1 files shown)


lrhn (approved):
Did we change the grammar at some point, or why are we not accepting the type arguments already?

LGTM and thanks.

eernstg:
I'm afraid I did not discover that the change from

```ebnf
<metadata> ::= ('@' <qualified> ('.' <identifier>)? (<arguments>)?)*
```

to

```ebnf
<metadata> ::= ('@' <metadatum>)*
<metadatum> ::= <identifier> | <qualifiedName> | <constructorDesignation> <arguments>
```

introduced syntactic support for type arguments. One of the issues resolved by that change was that `<qualified> ('.' <identifier>)?` allows for built-in identifiers in too many locations (like `abstract . abstract . abstract`), and `<constructorDesignation>` had already been scrutinized in order to handle constructor designations correctly. But I forgot that we should have avoided the type arguments in just this case.

Luckily we don't have to do that any more. :smile:

jensjoha:
This now says

> However, the *grammar* does not allow type arguments
>
> [...]
> 
> There is no change in the grammar [...]

which seems... strange.

eernstg:
Sorry about that, cf. https://github.com/dart-lang/language/pull/1516.

lrhn:
So, the story is now that "we have already fixed the spec". There was much rejoicing.

eernstg:
:grin: and thanks!

PR #1538, opened 2021-03-22
Take out the null-safety related elements of the specification

eernstg:
In preparation for a stable 2.11 language specification release, this PR takes out those few elements in the language specification that are part of the null safety feature bundle.

In particular, it removes `?` from types, `?..` from cascades, `late` from variables, and `required` from formal parameters.

I went over every occurrence of `null` in the specification, but did not see any other elements that are part of null safety (in general, we've avoided adding those things, because we were planning to release a stable version of the pre-null-safety language specification).

@lrhn, can you think of other elements that we need to take out in order to make it a clean 2.11 spec?

It would be really nice to complete the work on adding `\DefineSymbol` at every introductory occurrence of a mathematical symbol, but it might also be OK to finish that work later on, in preparation for a release of a spec with null safety. (We currently have a reasonably consistent use of `\DefineSymbol` in the spec up to page 75). I'd suggest that we simply keep it as-is and publish that, and then complete the work later on.

We _should_ land https://github.com/dart-lang/language/pull/1501, https://github.com/dart-lang/language/pull/1468 (which is about type aliases, but it prevents them from being 'non-function' because it uses `<functionType>` rather than `<type>` in the grammar rule, and we _should_ include the corrections in that PR as soon as possible). But they wouldn't conflict with this PR, so we can just land them after this one, or before, whatever fits better.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -5990,14 +5990,14 @@ to $e$ if the following conditions are all satisfied:
   }
 \item
   The type $S$ does not have a member with the basename $m$,
-  and $S$ is neither \DYNAMIC{} nor \code{Never}.
+  and $S$ is not \DYNAMIC.
 
   \commentary{%
-    \DYNAMIC{} and \code{Never} are considered to have all members.
+    \DYNAMIC{} is considered to have all members.
     Also, it is an error to access a member on a receiver of type \VOID{}
     (\ref{typeVoid}),
     so extensions are never applicable to receivers of
-    any of the types \DYNAMIC, \code{Never}, or \VOID.%
+    any of the types \DYNAMIC{} or \VOID.%
    }
 
   For the purpose of determining extension applicability,
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -13,8 +13,8 @@
 \usepackage{makeidx}
 \makeindex
 \title{Dart Programming Language Specification\\
-{5th edition draft}\\
-{\large Version 2.8.0-dev}}
+{5th edition}\\
+{\large Version 2.11.0}}
 \author{}
 
 % For information about Location Markers (and in particular the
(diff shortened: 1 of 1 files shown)


lrhn:
Other null-safety-only features:
* `e1?[e2]`
* `e1?.s2.s3` being short-circuited.
* "no implicit downcasts" (hard to *remove*).
* The type `Never`.
* Subtyping rules for `null` (`null is Object` being false).
* New special-case rules for integer arithmetic.
* `e is T` and `e as T` being potentially constant expressions.

At least we don't have type promotion in the spec 😉 .

eernstg:
Thanks!

> `e1?[e2]`

That one has already been removed in this PR.

> `e1?.s2.s3` being short-circuited.
> "no implicit downcasts" (hard to remove).
> Subtyping rules for null (null is Object being false).
> New special-case rules for integer arithmetic.

Those things were never in dartLangSpec.tex.

> The type Never.

Ah, yes: It occurs once in normative text in 'Applicability of an Extension', and twice in commentary immediately following that occurrence. Fixed. (It's easy to unfix later, I'll just revert this CL, so I haven't added any reminder comments there.)

> `e is T` and `e as T` being potentially constant expressions.

I thought that was older, and I can see the following in the git log, commit 7dc3b79f97b994c889e6bfe9ad757f25ab75d749 (Feb 2019):

```
% New in 2.1.
\item An expression of the form \code{$e$ is $T$} is potentially constant if $e$ is a potentially constant expression and $T$ is a constant type expression, and it is further constant if $e$ is constant.
```

So the new part is that `T` can be a type variable, and that's again something which was never in dartLangSpec.tex, only in other documents.

lrhn (approved):

leafpetersen:
Note that this was the only reference for the null safety syntax.  Do we have a plan for when/how to add this back in?

eernstg:
Yes, I'll revert commit 9bf9692ea7ffae3b0282fa3f4d2f52ed7fc6a343 as soon as the 2.11 stable language specification has been released.

PR #1594, opened 2021-04-21
Tiny fix: A top-level declaration pair (same basename) must be a getter and a setter

eernstg:
Section 'Class Member Conflicts' specifies several conflicts (we can't have a method and a setter with the same basename, etc), but similar rules are missing for top-level declarations.

Only one of these rules is relevant (there are no top-level constructors so we can't have a conflict at the top level involving a constructor, etc.): It is an error to have a setter with basename `n` together with any other declaration with basename `n`, except when that other declaration is a getter.

This PR adds that rule.

Update: The rule in 'Class Member Conflicts' about setter/method conflicts was a bit vague: It referred to "member" with the intent that it should cover both static and instance members. It is now split into two sentences where one says 'the interface of $C$ contains an instance member' respectively '$C$ declares a static member'. The static/instance and instance/static cases are already covered by the previous sentence.

There is no associated implementation work, this behavior is already implemented.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -26,9 +26,11 @@
 %
 % Significant changes to the specification.
 %
-% 2.13 (there was no 2.12)
+% 2.12 - 2.13 (there was no 2.11)
 % - Revert the CL where certain null safety features were removed (to enable
 %   publishing a stable version of the specification).
+% - Add rule that a top-level pair of declarations with the same basename
+%   is a compile-time error except when it is a getter/setter pair.
 %
 % 2.8 - 2.10
 % - Change several warnings to compile-time errors, matching the actual
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -18222,8 +18222,6 @@ the Dart implementation.
 Libraries are units of privacy.
 A private declaration declared within a library $L$
 can only be accessed by code within $L$.
-Any attempt to access a private member declaration from outside $L$
-will cause a method, getter or setter lookup failure.
 
 \commentary{%
 Since top level privates are not imported,
@@ -18245,6 +18243,13 @@ and its namespace is the library namespace of $L$
 It is a compile-time error if the local namespace of library $L$
 has two declarations with the same basename,
 except when they are a getter and a setter.
+
+\commentary{%
+Two distinct names $n_1$ and $n_2$ can only have the same basename
+when they are of the form \id{} and \code{\id=},
+so this kind of conflict \emph{always} involves a setter.
+But the other declaration could be a function, a class, etc.%
+}
 
 
 \subsection{Imports}
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -18783,7 +18783,7 @@ In the second step we compute the exported namespace of $L$.
 Let \NamespaceName{\metavar{merged}} be the result of applying
 conflict merging
 (\ref{conflictMergingOfNamespaces})
-to \List{\metavar{NS}}{\metavar{one},1}{\metavar{one},m}.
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -18209,16 +18209,19 @@ \section{Libraries and Scripts}
...
 the Dart implementation.
 
 \LMHash{}%
 Libraries are units of privacy.
-A private declaration declared within a library $L$ can only be accessed by code within $L$.
-Any attempt to access a private member declaration from outside $L$ will cause a method, getter or setter lookup failure.
+A private declaration declared within a library $L$
+can only be accessed by code within $L$.
+Any attempt to access a private member declaration from outside $L$
+will cause a method, getter or setter lookup failure.
```
lrhn, line 18224: This sentence is confusing. If I do `c._x`, it's a matter of intent whether I'm "attempt"ing to look up an inaccessible private member of `C` (the type of `C`) ... or the (potentially accessible) private member of `C`'s superclass. Both of them will find the superclass member if it's accessible, so not necessarily a "lookup failure".
We can't determine intent from syntax.

Probably just remove it. Maybe:

> A name starting with an underscore cannot refer to a declaration from another library.

specification/dartLangSpec.tex
```diff
@@ -18227,14 +18230,20 @@ \section{Libraries and Scripts}
 
 \LMHash{}%
 The \Index{public namespace} of library $L$ is the namespace that maps
-the simple name of each public top-level member declaration $m$ of $L$ to $m$.
+the name of each public top-level member declaration $m$ of $L$ to $m$.
```
lrhn, line 18233: "maps" suggests being a function (a name only maps to one declaration), but ....

specification/dartLangSpec.tex
```diff
@@ -18227,14 +18230,20 @@ \section{Libraries and Scripts}
...
 to the corresponding declaration.
 The \Index{library scope} of library $L$ is the outermost scope in $L$,
 and its namespace is the library namespace of $L$
 (\ref{theImportedNamespace}).
 
+\LMHash{}%
+% A setter can not be paired with a function, class, etc.
+It is a compile-time error if the local namespace of library $L$
+has two declarations with the same basename,
+except when they are a getter and a setter.
```
lrhn, line 18245: ... then a namespace *cannot* hold two members with the same name. This then only catches a setter and a non-getter, but not a getter and a method, or two getters.

Maybe make it a compile-time error for the *library* to have two declarations with the same basename unless one is a getter declaration and the other is a setter declaration.

That catches the conflict "before" creating the namespace, and the namespace can then be treated as injective.

(Or is this handled elsewhere by the general rule that no scope can contain two declarations with the same name, in which case it is fine to only catch the setter + non-getter situation)

lrhn (approved):

specification/dartLangSpec.tex
eernstg: Right, it's better to simply remove that sentence. Done.

specification/dartLangSpec.tex
eernstg: We have defined a (compile-time) namespace to be a partial function, so let's stick to that.

specification/dartLangSpec.tex
eernstg: There is no conflict here: I'm taking into account that the namespace is a partial function, so it does not contain two bindings with the same key.

So "same basename" implies "a pair of names of the form `id` and `id=`.

We can't have a clash between a getter and a method etc., because then the namespace wouldn't be a partial function, and that's already a compile-time error (page 12). So we are already catching the "same name" conflicts before creating the namespace. This extra error is concerned with `id`/`id=` conflicts only.

But this is not easy to see from the text, so we need a clarification. I've added commentary to achieve that.

eernstg (commented):
Thanks! Some changes were made in response to the review.

PR #1599, opened 2021-04-23
Specification clean-up

eernstg:
This CL does not change the contents of the specification. It changes the following formatting properties:
- Lines with normal text (not `\code{}` and not math stuff) >80 chars are split to fit in 80 columns.
- `\code{}` and math elements not at the beginning of a paragraph that are large enough to need their own line are now consistently without indentation (`\noindent`).
- A number of unnecessary `{}` after LaTeX commands have been removed.
- One reference (`\ref{...}`) was undefined. A relevant reference occurred 3 lines earlier, so this undefined one was deleted.
- The wording in 2-3 comments was changed slightly to fit in 80 columns (only comments!).

We have preferred this kind of formatting for a long time, but generally changed long lines as part of other changes. It's a cleaner approach to fix all the long lines and inconsistent indentations in a single PR which is known to do just that, hence this PR.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -1771,7 +1771,7 @@ It is a compile-time error to preface a function declaration
 with the built-in identifier \STATIC.
 
 \LMHash{}%
-When we say that a function $f_1$ \Index{forwards} to another function $f_2$, 
+When we say that a function $f_1$ \Index{forwards} to another function $f_2$,
 we mean that invoking $f_1$ causes $f_2$ to be executed
 with the same arguments and/or receiver as $f_1$,
 and returns the result of executing $f_2$ to the caller of $f_1$,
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -4720,7 +4789,7 @@ \section{Interfaces}
 such that we can say that an interface 'declares' or 'has' a member,
```
lrhn, line 4789: Why are these red in the review tool?

lrhn (approved):

specification/dartLangSpec.tex
eernstg: I think it considers the use of a single quote at both ends of a phrase as a syntax error, in the sense that it should have been a backquote and a single quote. Fixed approx. 50 locations where this was done inconsistently. There will still be a few red spots, because github highlighting doesn't understand all situations as intended.

eernstg (commented):
Thanks! Corrected a bunch of inconsistent quoted phrases (using `<backquote>phrase<single_quote>` rather than `<single_quote>phrase<single_quote>`, but keeping code in comments in the `<backquote>usual(md.notation)<back_quote>`).

PR #1615, opened 2021-05-05
Correct `part of` syntax

eernstg:
Tiny bug fix, cf. https://github.com/dart-lang/sdk/issues/45916:

The specification currently allows for a `part of` directive where the enclosing library is specified using `<identifier> ('.' <identifier>)*`, but not using `<uri>`. The latter is intended to work, and it has been working in all tools for a while, and this PR just brings the specification up to date.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -19558,7 +19558,7 @@ may be found.
 \begin{grammar}
 <partDirective> ::= <metadata> \PART{} <uri> `;'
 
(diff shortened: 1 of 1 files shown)


eernstg:
@lrhn, this one is tiny. ;)

eernstg:
@lrhn, do you have any specific worries about this change?

lrhn (approved):

eernstg:
Thanks!

PR #1616, opened 2021-05-06
Specify that generic method instantiation doesn't apply to functionType.call

eernstg:
As decided at the language meeting yesterday, `void Function() f = o.call;` should no longer be specified to apply generic method instantiation to the call method of a function object, even in the situation where all other conditions are satisfied. The treatment of any other method, including methods named `call` in the interface of a user-written class, remains unchanged.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -14545,7 +14545,8 @@ would have been a compile-time error.%
 Let $i$ be a property extraction expression of the form
 \code{$e$?.\id}, \code{$e$.\id}, or \code{\SUPER.\id}
 (\ref{propertyExtraction}, \ref{superGetterAccessAndMethodClosurization}),
-which is statically resolved to denote an instance method named \id,
+which is statically resolved to denote an instance method named \id{}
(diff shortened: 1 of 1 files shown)


lrhn (approved):

specification/dartLangSpec.tex
```diff
@@ -14545,7 +14545,8 @@ \subsubsection{Generic Method Instantiation}
 Let $i$ be a property extraction expression of the form
 \code{$e$?.\id}, \code{$e$.\id}, or \code{\SUPER.\id}
 (\ref{propertyExtraction}, \ref{superGetterAccessAndMethodClosurization}),
-which is statically resolved to denote an instance method named \id,
+which is statically resolved to denote an instance method named \id{}
+that is not the \CALL{} method of a function type,
```
lrhn, line 14549: So this simply says that if it's the instance `call` method of a function type, then we don't do instantiation (which will make it a type error), right?

eernstg: Exactly.

PR #1617, opened 2021-05-06
Add first version of a `view` proposal

eernstg:
This PR adds a proposal for a `view` language construct which uses the concept of a 'view' to provide support for zero-cost abstraction.

The basic idea is the same as that of an extension type, but the word `view` matches the actual intention behind this language construct much better than `extension type`. So it's basically a re-wording of the [existing proposal](https://github.com/dart-lang/language/blob/master/working/1426-extension-types/feature-specification.md).

Apart from the terminology there is one substantial difference between the [extension types](https://github.com/dart-lang/language/blob/master/working/1426-extension-types/feature-specification.md) proposal and this proposal: With the extension types proposal, a plain `extension` declaration (available since Dart 2.6) will declare a type as well as a set of extension methods. This means that an existing extension declaration `E` can be re-interpreted as an extension type `E`, and the members of `E` can be invoked on a receiver of type `E` as well as via the existing extension method invocation mechanism (where the type of the receiver is some type that matches the on-type of `E`). With this proposal, no new affordances are provided for existing extension declarations. In short, "an extension is not a view".

This means that the extension types proposal offers a slightly greater amount of expressive power and flexibility. However, that particular affordance has been eliminated in this proposal because it seems unlikely that a declaration `E` which works well as a set of extension methods would also work well as the type of a receiver. In other words, that particular affordance will probably just be a constant source of confusion, and rarely if at all a useful feature.

### DIFF
--- a/working/1426-extension-types/feature-specification-views.md
+++ b/working/1426-extension-types/feature-specification-views.md
@@ -200,16 +200,17 @@ for late binding of a view member, and hence there is no notion of
 overriding. In return for this lack of expressive power, we get improved
 performance.
 
-Here is another example. It illustrates the fact that a view type `V` with
-on-type `T` introduces a type `V` which is a supertype of `T`. This makes
-it possible to assign an expression of type `T` to a variable of type
-`V`. This corresponds to "entering" the view type (accepting the specific
-discipline associated with `V`). Conversely, a cast from `V` to `T` is a
-downcast, and hence it must be written explicitly. This cast corresponds to
-"exiting" the view type (allowing for violations of the discipline
-associated with `V`), and the fact that the cast must be written explicitly
-helps developers maintaining the discipline as intended, rather than
-dropping out of the view type by accident, silently.
+Here is another example. It illustrates the fact that a plain view with
+on-type `T` introduces a view type `V` which is a supertype of `T`. (There
+are two other kinds, `open` and `closed` views, with different subtyping
+relationships.) This makes it possible to assign an expression of type `T`
+to a variable of type `V`. This corresponds to "entering" the view type
+(accepting the specific discipline associated with `V`). Conversely, a cast
+from `V` to `T` is a downcast, and hence it must be written explicitly.
+This cast corresponds to "exiting" the view type (allowing for violations
+of the discipline associated with `V`), and the fact that the cast must be
+written explicitly helps developers maintaining the discipline as intended,
+rather than dropping out of the view type by accident, silently.
 
 ```dart
 view ListSize<X> on List<X> {
@@ -245,6 +246,7 @@ rules for elements used in view declarations:
       'on' <type>
       <viewShowHidePart>
       <interfaces>?
+      ('box' 'as' <typeName>)?
   '{'
     (<metadata> <viewMemberDefinition>)*
   '}'
@@ -334,11 +336,14 @@ denotes an invocation of the view method `m` with the same bindings.*
 
 The static analysis of `invokeViewMethod` is that it takes exactly three
 positional arguments and must be the receiver in a member access. The first
-argument must be a name that denotes a view declaration, the next argument
-must be a type argument list, together yielding a view type _V_. The third
-argument must be an expression whose static type is _V_ or the
-corresponding instantiated on-type (defined below). The member access must
-be a member of `V`.
+argument must be a type name that denotes a view declaration, the next
+argument must be a type argument list, together yielding a view type
+_V_. The third argument must be an expression whose static type is _V_ or
+the corresponding instantiated on-type (defined below). The member access
+must be a member of `V` or an associated member of a superview of `V`.
+
+*Superviews and associated members are specified in the section 'Composing
+view types'.*
 
 If the member access is a method invocation (including an invocation of an
 operator that takes at least one argument), it is allowed to pass an actual
@@ -413,7 +418,7 @@ class or mixin, or if a view type is used to derive a mixin.
 `V2`, and `V3` are view types, and `mixin M on V1 implements V2 {}`
 has two errors.*
 
-If `e` is an expression whose static type is the view type
+If `e` is an expression whose static type `V` is the view type
 <code>View<S<sub>1</sub>, .. S<sub>k</sub>></code>
 and the basename of `m` is the basename of a member declared by `V`,
 then a member access like `e.m(args)` is treated as
@@ -484,7 +489,7 @@ on-type is the same in either case.
 
 We say that `D` is _open_ respectively _closed_ if its declaration
 starts with the keyword `open` respectively `closed`, and similarly we
-say that a view type <code>View<T<sub>1</sub>, .. T<sub>k</sub>></code>
+say that a view type <code>View<S<sub>1</sub>, .. S<sub>k</sub>></code>
 where `View` denotes `D` is _open_ respectively _closed_.
 If `D` starts with the keyword `view` we say that `D` is _plain_
 and that corresponding view types are _plain_.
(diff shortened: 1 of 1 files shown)


eernstg:
(Other things came up, so the new version of this document has not yet been uploaded. I'll upload it after the language meeting).

eernstg:
Uploaded a version of the document where many typos and glitches have been fixed. It should be in rather good shape now.

eernstg:
@leafpetersen, I believe we agreed to land this document now? We can then handle further updates/discussions in separate PRs.

working/1426-extension-types/feature-specification-views.md
```diff
@@ -0,0 +1,955 @@
...
+
+```dart
+view TinyJson on Object {
+  Iterable<num> get leaves sync* {
+    var self = this;
+    if (self is num) {
+      yield self;
+    } else if (self is List<dynamic>) {
+      for (var element in self) {
+        yield* element.leaves;
```
leafpetersen, line 94: I think this code is quadratic, since every time you yield an element, you re-traverse element.leaves.  Might be better to use a different example, or not use sync*?

leafpetersen (approved):

working/1426-extension-types/feature-specification-views.md
eernstg: The number of invocations of `leaves` is actually identical to the number of nodes in the given tree, so if there is anything which is too expensive it would be the use of `sync*` (it is known that `sync*` ought to be optimized better than it is today).

I ran the following:

```dart
class TinyJson {
  final Object self;

  TinyJson(this.self);

  Iterable<num> get leaves sync* {
    var self = this.self;
    if (self is num) {
      yield self;
    } else if (self is List<dynamic>) {
      for (var element in self) {
        yield* TinyJson(element).leaves;
      }
    } else {
      throw "Unexpected object encountered in TinyJson value";
    }
  }
}

var b = false;

void use(dynamic d) {
  if (b) d.fooBar(); // Pretend that `d` is not ignored.
  try {} finally {} // Prevent inlining.
}

void testFlatList() {
  print('---------- FlatList ----------');
  for (int length = 1024; length < 1000000000; length *= 2) {
    var tiny = TinyJson(List<dynamic>.filled(length, 1));
    var sw = Stopwatch()..start();
    use(tiny.leaves);
    print('Length: $length, time: ${sw.elapsed}');
  }
}

void testTwoLevelList() {
  print('---------- TwoLevelList ----------');
  for (int length = 1024; length < 1000000000; length *= 2) {
    var tiny = TinyJson(List<dynamic>.filled(length, <dynamic>[1, 2]));
    var sw = Stopwatch()..start();
    use(tiny.leaves);
    print('Length: $length, time: ${sw.elapsed}');
  }
}

void testManyLevelList() {
  print('---------- TwoLevelList ----------');
  var max = 1000;
  List<dynamic> tinies = List.filled(max, [<dynamic>[0]]);
  for (int depth = 1; depth < max; depth++) {
    var tiny = tinies[depth] = <dynamic>[tinies[depth - 1]];
    var sw = Stopwatch()..start();
    use(TinyJson(tiny).leaves);
    print('Depth: $depth, time: ${sw.elapsed}, tiny: $tiny');
  }
}

void main() {
  testFlatList();
  testTwoLevelList();
  testManyLevelList();
}
```

I got the following output, which doesn't seem to imply anything excessively slow:

```
---------- FlatList ----------
Length: 1024, time: 0:00:00.001751
Length: 2048, time: 0:00:00.000004
Length: 4096, time: 0:00:00.000001
Length: 8192, time: 0:00:00.000001
Length: 16384, time: 0:00:00.000001
Length: 32768, time: 0:00:00.000001
Length: 65536, time: 0:00:00.000003
Length: 131072, time: 0:00:00.000005
Length: 262144, time: 0:00:00.000002
Length: 524288, time: 0:00:00.000002
Length: 1048576, time: 0:00:00.000002
Length: 2097152, time: 0:00:00.000002
Length: 4194304, time: 0:00:00.000008
Length: 8388608, time: 0:00:00.000008
Length: 16777216, time: 0:00:00.000006
Length: 33554432, time: 0:00:00.000007
Length: 67108864, time: 0:00:00.000008
Length: 134217728, time: 0:00:00.000008
Length: 268435456, time: 0:00:00.000007
Length: 536870912, time: 0:00:00.000008
---------- TwoLevelList ----------
Length: 1024, time: 0:00:00.000006
Length: 2048, time: 0:00:00.000000
Length: 4096, time: 0:00:00.000000
Length: 8192, time: 0:00:00.000000
Length: 16384, time: 0:00:00.000000
Length: 32768, time: 0:00:00.000008
Length: 65536, time: 0:00:00.000003
Length: 131072, time: 0:00:00.000003
Length: 262144, time: 0:00:00.000003
Length: 524288, time: 0:00:00.000002
Length: 1048576, time: 0:00:00.000001
Length: 2097152, time: 0:00:00.000001
Length: 4194304, time: 0:00:00.000002
Length: 8388608, time: 0:00:00.000004
Length: 16777216, time: 0:00:00.000007
Length: 33554432, time: 0:00:00.000007
Length: 67108864, time: 0:00:00.000007
Length: 134217728, time: 0:00:00.000007
Length: 268435456, time: 0:00:00.000007
Length: 536870912, time: 0:00:00.000008
---------- TwoLevelList ----------
Depth: 1, time: 0:00:00.000008
Depth: 2, time: 0:00:00.000000
Depth: 3, time: 0:00:00.000000
Depth: 4, time: 0:00:00.000000
Depth: 5, time: 0:00:00.000000
Depth: 6, time: 0:00:00.000000
Depth: 7, time: 0:00:00.000000
Depth: 8, time: 0:00:00.000000
Depth: 9, time: 0:00:00.000000
Depth: 10, time: 0:00:00.000000
...
Depth: 990, time: 0:00:00.000000
Depth: 991, time: 0:00:00.000000
Depth: 992, time: 0:00:00.000000
Depth: 993, time: 0:00:00.000000
Depth: 994, time: 0:00:00.000000
Depth: 995, time: 0:00:00.000000
Depth: 996, time: 0:00:00.000000
Depth: 997, time: 0:00:00.000000
Depth: 998, time: 0:00:00.000000
Depth: 999, time: 0:00:00.000000
```

leafpetersen: Interesting, I'm surprised.  I think `yield* TinyJson(element).leaves;` is equivalent to `for(var e in TinyJson(element).leaves) yield e`, and with the right input, I think `TinyJson(element).leaves` should be a stack of iterables of depth `n`, and so the `ith` `.next` call should have to traverse `i` stacked iterables to find the next element, hence quadratic.

@rakudrama am I getting myself confused here (I admit to finding sync* confusing) or is this one of the patterns that dart2js and the VM are now able to recognize and optimize?

eernstg: (I didn't see this comment before I landed, I'll make changes as needed in a new PR).

leafpetersen: FWIW, if I build lists of increasing length using:
```
dynamic mk(int i) {
  if (i == 0) {
    return [];
  }
  return [i, mk(i-1)];
}
```

then the VM seems to exhibit linear time growth, but DDC exhibits quadratic growth:

```
128 elements takes 3000 ms  (23.4375 ms/elements)
256 elements takes 11000 ms  (42.96875 ms/elements)
512 elements takes 51000 ms  (99.609375 ms/elements)
1024 elements takes 212000 ms  (207.03125 ms/elements)
2048 elements takes 889000 ms  (434.08203125 ms/elements)
4096 elements takes 4496000 ms  (1097.65625 ms/elements)
```

So looks like maybe this is quadratic, but the VM (and maybe dart2js?) are able to optimize it away?

PR #1626, opened 2021-05-10
Specify the private override error caused by a mixin application

eernstg:
Proposed rules for a mixin related name clash error, cf. https://github.com/dart-lang/sdk/issues/45959.

This PR adjusts the language specification section 'Mixin Application' such that it specifies the error which arises when an override relationship arises among two declarations with a private name `n` in a library `L` because of a mixin application that occurs outside `L`.

```dart
// Library 'lib.dart'.
class A {
  void _foo() {}  
}

class B {
  void _foo() {}
}

mixin M {
  void _foo() {}
}

mixin MA implements A {
  void _foo() {}
}

// Library 'other.dart'.
import 'lib.dart';

class AB extends A with B {} // Error.
class AM extends A with M {} // Error.
class AMA extends A with MA {} // Error.
```

Note that it is an error in all cases where the mixin application outside `lib.dart` creates an override relationship for a member with a private name (for instance, such that `A._foo` is overridden by `MA._foo`), even in the cases where the given members are already associated with each other by declarations (e.g., `MA._foo` is connected to `A._foo` by `implements A`, but it is still an error to have `A with MA` in `other.dart`).

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -5452,42 +5452,93 @@ and the instance members declared in $M$ as \Index{mixin member declarations}.
 A mixin may be applied to a superclass, yielding a new class.
 
 \LMHash{}%
-Let $S$ be a class,
-$M$ be a mixin with \NoIndex{required superinterface}s $T_1$, \ldots, $T_n$,
-\NoIndex{combined superinterface} $M_S$,
-\NoIndex{implemented interfaces} $I_1$, \ldots, $I_k$ and
-\metavar{members} as \NoIndex{mixin member declarations},
-and let $N$ be a name.
+\BlindDefineSymbol{S, S'}%
+Let $S$ be a class, and let $S'$ be a parameterized type of the form
+\code{$S$<$\cdots$>}.
+\commentary{%
+This includes the case where $S$ is non-generic and $S'$ is $S$.%
+}
 
 \LMHash{}%
-It is a compile-time error to apply $M$ to $S$ if $S$ does not implement,
-directly or indirectly, all of $T_1$, \ldots, $T_n$.
+\BlindDefineSymbol{M, \metavar{members}}%
+Let $M$ be a mixin with member declarations \metavar{members}.
+\BlindDefineSymbol{M', T_j}%
+Let $M'$ be a parameterized type of the form
+\code{$M$<$\cdots$>}
+\commentary{(again including the case where $M'$ is $M$)}.
+Let \List{T}{1}{n} be the required superinterfaces,
+\BlindDefineSymbol{M_S, I_j}
+$M_S$ the combined superinterface,
+\List{I}{1}{k} the implemented interfaces,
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -5452,42 +5452,93 @@ \subsection{Mixin Application}
...
+Let \DefineSymbol{L_C} be the library containing the mixin application.
+\commentary{%
+That is, the library containing the clause \code{$S$ \WITH{} $M$}
+or the clause \code{$S_0$ \WITH{} $M_1$, \ldots,\ $M_k$, $M$} giving rise
+to the mixin application.%
+}
+Let \DefineSymbol{L_M} be the library containing the declaration of $M$.
+
+\LMHash{}%
+Assume that $S$ has a concrete member $m_S$ which is accessible to $L_M$,
```
lrhn, line 5504: Did we agree on it needing to be *concrete*?

specification/dartLangSpec.tex
```diff
@@ -5452,42 +5452,93 @@ \subsection{Mixin Application}
...
+that $m_S$ has the name $n$ which is private
+(\commentary{so $m_S$ is declared in $L_M$}),
+and that $M$ declares a concrete member $m_M$ which is also named $n$
+\commentary{(note that $m_M$ is also declared in $L_M$)}.
+In this case a compile-time error occurs.
```
lrhn, line 5509: Should it be an error if the name is accessible in L_C (so L_C is L_M)?

specification/dartLangSpec.tex
eernstg: No, this PR awaits the discussion in the language team: Do we want to support these mixin applications in all the cases where it does not give rise to an override, or do we want to raise an error on all clashes (abstract or concrete, that doesn't matter)?

My preferred approach is the former, because I don't want to eliminate expressive power gratuitously.

However, we certainly need to raise an error when there is an override because of an implicitly induced implementation (say, a stub that checks the type of an actual argument and calls `super.sameMember`).

This means that the text in this PR needs to be updated, but it should be updated in two different ways if we choose the former respectively the latter approach.

eernstg: Cf. the decision at the language meeting yesterday, the PR has now been updated to make it an error when the private collision as created by a mixin application, no matter what kind of clash we have (concrete/concrete, concrete/abstract, concrete/stub, abstract/abstract, it's always an error).

specification/dartLangSpec.tex
eernstg: That's a mistake, added a few words to say that `L_C != L_M`.

eernstg:
PTAL, the spec change has been updated to reflect the decision at the language meeting yesterday: All clashes are conflicts, even abstract/abstract.

eernstg:
The commentary about the private collisions was extended to motivate the error for some abstract-on-concrete cases.

Issue #1631, opened 2021-05-14
Should `super` be callable?


stereotype441:
Here's a brain teaser for my fellow parsing nerds (I suspect @eernstg and @lrhn will particularly want to weigh in).

Is the following code valid?  If so, what do we expect its output to be?
```dart
f(x, [y]) => 'f(x=$x, y=$y)';

class B {
  call<T, U>(x) => '$this.call<$T, $U>($x)';

  operator<(x) => '($this < $x)';
}

class C extends B {
  toString() => 'C()';
  
  void test() {
    print(f(super<int, int>(0))); // (1)
  }
}

extension on Type {
  operator>(x) => '($this > $x)';
}

main() {
  C().test();
}
```
Stop reading now if you want to guess before I spoil it for you.

AFAICT, the answer depends who you ask:
- The spec, if I'm reading it correctly, doesn't have a grammar production for `super` followed by &lt;argumentPart&gt;, so `f(super<int, int>(0))` should unambiguously parse as a call to `f` with two arguments: `super < int` and `int > (0)`.  Thanks to the definitions of operators `<` and `>` I've provided, this is valid.  So I believe according to the spec, the program should print `f(x=(C() < int), y=(int > 0))`.
- The analyzer accepts the code without error.  But it doesn't seem to be interpreting `f(super<int, int>(0))` as a call to `f` with two arguments.  If you comment out the definition of `B.call`, the analyzer gives an error; if you comment out the definition of `B.operator<`, it doesn't.  This suggests that the analyzer is interpreting it as a call to `f` with one argument, and that argument is an invocation of `super.call` with type arguments `<int, int>` and argument `0`.
- The CFE gives an error at (1): "Can't use 'super' as an expression."  If you change (1) to `print(f(super(0)));`, you get the exact same message.  If instead you change (1) to `print(f(super<int, int>0));`, then the output is `f(x=(C() < int), y=(int > 0))`.  If you change (1) to `print(f(super.call<int, int>(0)));`, then the output is `f(x=C().call<int, int>(0), y=null)`.  This suggests that the CFE is parsing `f(super<int, int>(0))` the same way as the analyzer (not really surprising, since the parser implementation is shared), but it has an extra rule that "super" isn't callable unless the `.call` is explicit.

It seems like we ought to fix this disparity.  Anyone have an opinion about which interpretation we should pick as the "correct" one?  We aren't in danger of breaking anyone because the CFE rejects the code today.

[...]

PR #1636, opened 2021-05-17
Add support for implicit `.call` superinvocations

eernstg:
Cf. #1631. This PR adds the grammar rule to allow constructs like `super()` and `super<int>()`, and specifies them in the section `Superinvocations`.

The PR also changes a couple of occurrences of `super invocation` to `superinvocation`, for consistency with this term in itself, and for consistency with terms like `superinitializer`.

Finally, this PR adds `\DefineSymbol` (and related commands) such that the section `Superinvocations` is up to date with symbol introductions.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -149,8 +149,8 @@
 %   declaration must be taken into account during static checks.
 % - Disallow any expression statement starting with `{`, not just
 %   those that are map literals.
-% - Define a notion of lookup that is needed for super invocations, adjust
-%   specification of super invocations accordingly.
+% - Define a notion of lookup that is needed for superinvocations, adjust
+%   specification of superinvocations accordingly.
 % - Specify that it is a dynamic error to initialize a non-static variable
 %   with an object that does not have the declared type (e.g., a failed cast).
 % - Specify for constructor initializers that target variable must exist and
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -13800,15 +13801,17 @@ \subsubsection{Cascades}
 }
 
 
-\subsubsection{Super Invocation}
-\LMLabel{superInvocation}
+\subsubsection{Superinvocations}
+\LMLabel{superInvocations}
 
-% Conditional super invocation is meaningless: \THIS{} is not null.
+% Conditional superinvocations are meaningless: \THIS{} is not null.
```
lrhn, line 13807: Weeeeeell, technically ... `this` can be `null` for members of `Null`. (But it's perfectly fine to ignore that, it only affects, well, me.)

lrhn (approved):
This cleans up the section on `super.method(...)` invocations, touches super-tearoffs, but doesn't touch `super operator ...` at all. Should it? (Or did it not say "super invocation" anywhere?)

specification/dartLangSpec.tex
eernstg: ;-)

eernstg:
We don't have any terminology covering operator invocations on super, we only have references to the pseudo-syntax where the operators are considered as methods, e.g.,

> A relational expression of the form \SUPER{} $op$ $e_2$ is equivalent to
> the method invocation \code{\SUPER.$op$($e_2$)}.

That would be the reason why we have no occurrences of `super invocation` beyond the ones that I found. But it's a simple text search (searching `super invocat` and `super^Jinvocat`), so the search should be rather robust.

PR #1676, opened 2021-06-10
Fix commentary about class member conflicts and add missing error

eernstg:
Fix a couple of typos in the language specification about class member conflicts, and add a missing compile-time error about cyclic redirecting generative constructors.


lrhn (approved):
LGTM!

Issue #1682, opened 2021-06-15
Double word typos


jmewes:
The word "to" is unnecessarily repeated in the following sentence:

> If the static type of `e` is `double` then `e`  evaluates to to an instance of the double class

https://dart.dev/guides/language/specifications/DartLangSpec-v2.10.pdf, p. 172

```
\LMHash{}%
If the static type of $e$ is \code{int} then $e$ evaluates to
an instance of the \code{int} class representing the numeric value $-i$.
If $i$ is zero and the \code{int} class can represent a negative zero value,
then the resulting instance instead represents that negative zero value.
It is a compile-time error if the integer $-i$ cannot be represented
exactly by an instance of \code{int}.
```

https://github.com/dart-lang/language/blob/master/specification/dartLangSpec.tex#L15795-L15796

Same here:

![Auswahl_311](https://user-images.githubusercontent.com/5235584/121998419-209a8a80-cdac-11eb-9d3c-8a980a6806c4.png)

munificent:
Thanks for letting us know! I fixed the latter with https://github.com/dart-lang/language/commit/15673cb972ee0fb29466c6db2845688b432cfaf5.

I'll leave it to @eernstg to tweak the spec.

eernstg:
Yeah, let's just remove the double word typos: https://github.com/dart-lang/language/pull/1685.

jmewes:
Thanks!

PR #1685, opened 2021-06-17
Fix several double word typos, cf. dart-lang/language#1682

eernstg:
This PR just fixes a bunch of typos where the same word is repeated twice, and the author was unlikely to intend that that word should be repeated.


lrhn (approved):

eernstg:
Thanks!

PR #1687, opened 2021-06-17
Fix placement of metadata in required named parameters

bwilkerson:
Without this I believe the grammar specifies that the `required` keyword appears before metadata.

If this change doesn't get the LaTex right and it would be easier to just make the change yourself than to explain to me what to do I won't be offended if you just close the PR.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -1986,7 +1986,8 @@ Optional parameters may be specified and provided with default values.
 <defaultFormalParameter> ::= <normalFormalParameter> (`=' <expression>)?
 
 <defaultNamedParameter> ::= \gnewline{}
-  \REQUIRED? <normalFormalParameter> ((`=' | `:') <expression>)?
+  <metadata> \REQUIRED? <normalFormalParameterNoMetadata>
+  ((`=' | `:') <expression>)?
 \end{grammar}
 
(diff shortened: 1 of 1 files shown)


eernstg (approved):
Indeed, thanks!

specification/dartLangSpec.tex
```diff
@@ -1986,7 +1986,8 @@ \subsubsection{Optional Formals}
 <defaultFormalParameter> ::= <normalFormalParameter> (`=' <expression>)?
 
 <defaultNamedParameter> ::= \gnewline{}
-  \REQUIRED? <normalFormalParameter> ((`=' | `:') <expression>)?
+  <metadata> \REQUIRED? <normalFormalParameterNoMetadata>
+  ((`=' | `:') <expression>)?
```
eernstg, line 1990: If a \gnewline{} is added to line 1990 then we avoid that the rule runs into the margin:

```suggestion
  \gnewline{} ((`=' | `:') <expression>)?
```

eernstg (approved):
Tried out the LaTeX, recommended adding a special grammar newline.

bwilkerson:
Sorry, I saw the approval so I merged. I'm happy to put together another PR is you'd like me to.

eernstg:
That would be fine! I could approve and land it tomorrow morning CEST.

PR #1695, opened 2021-06-21
Restore errors about derived mixins

eernstg:
Commit 89347169c715d39e5d3e6ae588cd2bdcbb875492 removed two errors from the language specification: It is an error to derive a mixin from a class whose superclass is not `Object`, or a class that explicitly declares a generative constructor. This PR restores those two errors. We do not have the required language design needed in order to lift those restrictions, so they should be restored.

There is no associated implementation effort, the errors have been reported by the CFE and the analyzer all the time.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -5248,7 +5248,7 @@ A mixin describes the difference between a class and its superclass.
 A mixin is either derived from an existing class declaration
 or introduced by a mixin declaration.
 It is a compile-time error to derive a mixin from
-a class that explicitly declares a generative constructor,
+a class that declares a generative constructor,
 or from a class that has a superclass other than \code{Object}.
 
 \LMHash{}%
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -5247,6 +5247,9 @@ \section{Mixins}
 A mixin describes the difference between a class and its superclass.
 A mixin is either derived from an existing class declaration
 or introduced by a mixin declaration.
+It is a compile-time error to derive a mixin from
+a class that explicitly declares a generative constructor,
```
lrhn, line 5251: Is "explicitly" needed? Can you *declare* a generative constructor implicitly, or do we reserve the word "declare" for when you actually write something.

lrhn (approved):

specification/dartLangSpec.tex
eernstg: I thought about the same thing: We don't otherwise say 'explicitly' (along with that, we don't consider any implicitly induced declarations to be part of the program), so the consistent choice is to remove the word `explicitly` here. Done.

PR #1727, opened 2021-07-07
Fix a typo in Constructor tear-off spec

sgrekhov:
Seems there was a typo in an example in a Summary section


lrhn (approved):
Good catch. Thanks.

PR #1729, opened 2021-07-08
Clarify override error for inherited member with non-covariant parameter

eernstg:
This PR updates the language specification to make it explicit that it is a compile-time error for a member signature _m_ to override another member signature _m2_ if the latter has a parameter which is covariant-by-declaration, and the former does not have that.

This error is independent of the function types of the method signatures, so let's consider an example where the function types are identical (so it's trivially a correct member override in that respect):

```dart
class A { num foo(num n) => 1.1; }
abstract class B { num foo(covariant num x); }
class C extends A with B {} // Error
```

Here, `C` is an error because its interface has a member signature `num foo(covariant num)`, but it has an implementation with signature `num foo(num)` (inherited from `A`), and the `covariant` property is missing. (Note that we're talking about member signatures, and `covariant` is always explicit in a member signature, as opposed to syntactic declarations where it may be present implicitly, because it is present in a superinterface.)

The point is that the inherited method implementation `A.foo` doesn't "magically" get a parameter which is covariant-by-declaration just because it's inherited by `C`.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -5222,21 +5222,20 @@ if{}f the following criteria are all satisfied:
   (\ref{superBoundedTypes}).
 
   \commentary{%
-    This requirement can actually be violated:
-    If $m$ is the signature of a concrete member declaration
-    inherited from a superclass $S$ into a class $C$,
-    a parameter $p$ may not be covariant-by-declaration
-    even though the parameter corresponding to $p$ in a superinterface of $C$
-    is covariant-by-declaration.
-    In this case the inherited member fails to be a correct override
-    of the member in that superinterface
-    (unless $p$ has a top type, where the covariance cannot cause unsoundness).
-
-    Otherwise, when $m$ is the member signature of a declaration in a class $C$
+    This requirement is satisfied in the typical case, but not in all cases:
+    When $m$ is the member signature of a declaration in a class $C$
     and $m'$ is a member signature from a superinterface of $C$,
     this requirement will never fail.
     This is because $p$ is automatically and implicitly covariant-by-declaration
     whenever $p'$ is covariant-by-declaration.
+
+    However, if $m$ is the signature of a concrete member declaration
+    inherited from a superclass $S$ into a class $C$,
+    a parameter $p$ may not be covariant-by-declaration
+    even though $p'$ is covariant-by-declaration.
+    In this case the inherited member fails to be a correct override
+    of the member in that superinterface
+    (unless $p$ has a top type, where the covariance cannot cause unsoundness).
   }
 \item
   %% TODO(eernst): Come nnbd, this warning is removed.
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -5216,26 +5216,16 @@ if{}f the following criteria are all satisfied:
     This is a built-in property of the function type subtype rules.
   }
 
-  If $p'$ is a formal parameter in $m'$ which is covariant-by-declaration,
-  the corresponding parameter $p$ in $m$ must also be covariant-by-declaration
-  unless the type of $p$ is a top type
-  (\ref{superBoundedTypes}).
-
   \commentary{%
-    This requirement is satisfied in the typical case, but not in all cases:
-    When $m$ is the member signature of a declaration in a class $C$
-    and $m'$ is a member signature from a superinterface of $C$,
-    this requirement will never fail.
-    This is because $p$ is automatically and implicitly covariant-by-declaration
-    whenever $p'$ is covariant-by-declaration.
-
-    However, if $m$ is the signature of a concrete member declaration
+    Note that if $m$ is the signature of a concrete member declaration
     inherited from a superclass $S$ into a class $C$,
     a parameter $p$ may not be covariant-by-declaration
     even though $p'$ is covariant-by-declaration.
-    In this case the inherited member fails to be a correct override
-    of the member in that superinterface
-    (unless $p$ has a top type, where the covariance cannot cause unsoundness).
+    In this case, for soundness, there may be a need to perform
+    a dynamic type check on the actual argument passed to $p$,
+    which is not performed by the inherited implementation.
+    An implementation may choose to generate a forwarding method
+    to ensure that the required type check is performed.
   }
 \item
   %% TODO(eernst): Come nnbd, this warning is removed.
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -34,6 +34,8 @@
 % - Change grammar to enable non-function type aliases. Correct rule for
 %   invoking `F.staticMethod()` where `F` is a type alias.
 % - Add missing error for cyclic redirecting generative constructor.
+% - Clarify the treatment of `covariant` parameters in the interface of a class
+%   that inherits an implementation where those parameters are not covariant.
 %
 % 2.8 - 2.10
 % - Change several warnings to compile-time errors, matching the actual
@@ -5221,9 +5223,10 @@ if{}f the following criteria are all satisfied:
     so that modifier need not be present syntactically in $D$.
   }
 
-  It is a compile-time error if $m'$ has a parameter
-  with the modifier \COVARIANT,
-  and the corresponding parameter in $m$ does not have that modifier.
+  If $p'$ is a formal parameter in $m'$ that has the modifier \COVARIANT,
+  the corresponding parameter $p$ in $m$ must also have the modifier \COVARIANT,
+  unless the type of $p$ is a top type
+  (\ref{superBoundedTypes}).
 
   \commentary{%
     This can only occur in the case where $m$ is the signature of
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -5216,16 +5216,26 @@ if{}f the following criteria are all satisfied:
     This is a built-in property of the function type subtype rules.
   }
 
+  If $p'$ is a formal parameter in $m'$ which is covariant-by-declaration,
+  the corresponding parameter $p$ in $m$ must also be covariant-by-declaration
+  unless the type of $p$ is a top type
+  (\ref{superBoundedTypes}).
+
   \commentary{%
-    Note that if $m$ is the signature of a concrete member declaration
+    This requirement is satisfied in the typical case, but not in all cases:
+    When $m$ is the member signature of a declaration in a class $C$
+    and $m'$ is a member signature from a superinterface of $C$,
+    this requirement will never fail.
+    This is because $p$ is automatically and implicitly covariant-by-declaration
+    whenever $p'$ is covariant-by-declaration.
+
+    However, if $m$ is the signature of a concrete member declaration
     inherited from a superclass $S$ into a class $C$,
     a parameter $p$ may not be covariant-by-declaration
     even though $p'$ is covariant-by-declaration.
-    In this case, for soundness, there may be a need to perform
-    a dynamic type check on the actual argument passed to $p$,
-    which is not performed by the inherited implementation.
-    An implementation may choose to generate a forwarding method
-    to ensure that the required type check is performed.
+    In this case the inherited member fails to be a correct override
+    of the member in that superinterface
+    (unless $p$ has a top type, where the covariance cannot cause unsoundness).
   }
 \item
   %% TODO(eernst): Come nnbd, this warning is removed.
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -5250,11 +5250,12 @@ if{}f the following criteria are all satisfied:
 
 \commentary{%
 Note that a parameter which is covariant-by-declaration
-must have a type which is subject to one more check
+must have a type which satisfies one more requirement,
+relative to the corresponding parameters in all superinterfaces,
+both direct and indirect
 (\ref{instanceMethods}).
-We cannot cover that property here,
-because it is concerned with declarations in all superinterfaces,
-whereas the notion of a correct member override is only concerned with
+We cannot make that requirement a part of the notion of correct overrides,
+because correct overrides are only concerned with
 the relation to a single superinterface.%
 }
 
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -5204,28 +5204,42 @@ if{}f the following criteria are all satisfied:
   $F$ must then be a subtype of $F'$.
 
   \commentary{%
-  The subtype requirement ensures that argument list shapes
-  that are admissible for an invocation of a method with signature $m'$
-  are also admissible for an invocation of a method with signature $m$.
-  For instance, $m'$ may accept 2 or 3 positional arguments,
-  and $m$ may accept 1, 2, 3, or 4 positional arguments, but not vice versa.
-  This is a built-in property of the function type subtype rules.
-  %
-  Note that a member signature differs from
-  an underlying syntactic declaration $D$ in a class $C$.
-  In particular,
-  a parameter in a member signature has the modifier \COVARIANT{}
-  if and only if the parameter is covariant-by-declaration
-  (\ref{covariantParameters}),
-  and that may be the case due to declarations in a supertype of $C$,
-  so that modifier need not be present in $D$.
-  %
-  There is an additional potential compile-time error associated with
-  a parameter which is covariant-by-declaration
-  (\ref{instanceMethods}).
-  But we cannot cover that here as a property of member overrides,
-  because it is concerned with declarations in all superinterfaces,
-  indirect as well as direct.%
+    The subtype requirement ensures that argument list shapes
+    that are admissible for an invocation of a method with signature $m'$
+    are also admissible for an invocation of a method with signature $m$.
+    For instance, $m'$ may accept 2 or 3 positional arguments,
+    and $m$ may accept 1, 2, 3, or 4 positional arguments, but not vice versa.
+    This is a built-in property of the function type subtype rules.
+
+    Note that a member signature differs from
+    an underlying syntactic declaration $D$ in a class $C$.
+    In particular,
+    a parameter in a member signature has the modifier \COVARIANT{}
+    if and only if the parameter is covariant-by-declaration
+    (\ref{covariantParameters}),
+    and that may be the case due to declarations in a supertype of $C$,
+    so that modifier need not be present syntactically in $D$.
+  }
+
+  It is a compile-time error if $m'$ has a parameter
+  with the modifier \COVARIANT,
+  and the corresponding parameter in $m$ does not have that modifier.
+
+  \commentary{%
+    This can only occur in the case where $m$ is the signature of
+    a concrete member declaration inherited from a superclass.
+    Otherwise, if $m$ needs to be a correct override of $m'$ then
+    $m'$ is already a member of a superinterface of the enclosing class
+    of the declaration whose signature is $m$.
+    Every \COVARIANT{} parameter of $m'$ is then also \COVARIANT{} in $m$,
+    with or without an explicit occurrence of the modifier \COVARIANT.
+
+    There is one additional potential compile-time error associated with
+    a parameter which is covariant-by-declaration
+    (\ref{instanceMethods}).
+    We cannot cover that error here as a property of member overrides,
+    because it is concerned with declarations in all superinterfaces,
+    indirect as well as direct.%
   }
 \item
   If $m$ and $m'$ are both methods,
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -5195,13 +5195,15 @@ if{}f the following criteria are all satisfied:
 \begin{itemize}
 \item
   $m$ and $m'$ are both methods, both getters, or both setters.
-  % We cannot have a setter and a method, say, because they cannot have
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -5204,28 +5204,42 @@ \subsubsection{Correct Member Overrides}
   $F$ must then be a subtype of $F'$.
```
lrhn, line 5204: Should it have type `Object?` instead of `Object`, or is the spec just not null safe yet?

specification/dartLangSpec.tex
```diff
@@ -5204,28 +5204,42 @@ \subsubsection{Correct Member Overrides}
...
+    In particular,
+    a parameter in a member signature has the modifier \COVARIANT{}
+    if and only if the parameter is covariant-by-declaration
+    (\ref{covariantParameters}),
+    and that may be the case due to declarations in a supertype of $C$,
+    so that modifier need not be present syntactically in $D$.
+  }
+
+  It is a compile-time error if $m'$ has a parameter
+  with the modifier \COVARIANT,
```
lrhn, line 5225: We are in a predicate definition ("is a valid override if .... "), so it's the wrong place to introduce compile-time errors. Just say that m is not a correct override of m'.
It's then up to other places in the spec to make it an error if something is not a valid override of something else.

*Also*, I think that if m has a top type as parameter type, it is a valid override of a covariant function. After all, it can accept all the same arguments.

specification/dartLangSpec.tex
eernstg: I haven't started putting null-safety elements into the spec yet, we only have those snippets of the grammar where the null safety elements have been there for many months.

specification/dartLangSpec.tex
eernstg: > Just say ..
Good point! Done.

> .. top type ..
Good point, too!

This is likely to be a new behavior for the analyzer (as well as the CFE), so this means that there will be an implementation effort in the analyzer that wasn't previously expected for this spec change. However, there's no rush to do it because it is a non-breaking change. A developer who _would_ have used this feature would always be able to write a forwarding method (which is what they need to do in the non-top type case, anyway).

eernstg:
Made some adjustments, PTAL.

specification/dartLangSpec.tex
```diff
@@ -5204,28 +5206,43 @@ \subsubsection{Correct Member Overrides}
...
+    This is a built-in property of the function type subtype rules.
+
+    Note that a member signature differs from
+    an underlying syntactic declaration $D$ in a class $C$.
+    In particular,
+    a parameter in a member signature has the modifier \COVARIANT{}
+    if and only if the parameter is covariant-by-declaration
+    (\ref{covariantParameters}),
+    and that may be the case due to declarations in a supertype of $C$,
+    so that modifier need not be present syntactically in $D$.
```
lrhn, line 5223: Slightly confusing to read (to me). Consider changing the last line to:

  rather than a modifier that is syntactically present in $D$.

specification/dartLangSpec.tex
```diff
@@ -5204,28 +5206,43 @@ \subsubsection{Correct Member Overrides}
...
+  }
+
+  If $p'$ is a formal parameter in $m'$ that has the modifier \COVARIANT,
+  the corresponding parameter $p$ in $m$ must also have the modifier \COVARIANT,
+  unless the type of $p$ is a top type
+  (\ref{superBoundedTypes}).
+
+  \commentary{%
+    This can only occur in the case where $m$ is the signature of
```
lrhn, line 5232: This is still a predicate, not a situation. The predicate does not say where it will be used.
(Also, try to avoid the unqualified "this".)

Maybe phrase this as:

  This requirement is relevant in the case where $m$ is the signature of a concrete
  member declaration inherited from a superclass where the parameter was not covariant,
  and $m'$ is the signature of an interface method, added in a subclass, 
  that $m$ is required to implement.

specification/dartLangSpec.tex
```diff
@@ -5204,28 +5206,43 @@ \subsubsection{Correct Member Overrides}
...
+    a concrete member declaration inherited from a superclass.
+    Otherwise, if $m$ needs to be a correct override of $m'$ then
+    $m'$ is already a member of a superinterface of the enclosing class
+    of the declaration whose signature is $m$.
+    Every \COVARIANT{} parameter of $m'$ is then also \COVARIANT{} in $m$,
+    with or without an explicit occurrence of the modifier \COVARIANT.
+
+    There is one additional potential compile-time error associated with
```
lrhn, line 5240: Again with the compile-time errors in a predicate :)

I'd just drop this. It's not relevant to the *definition* of when something is a valid override of something else.

lrhn (approved):

specification/dartLangSpec.tex
eernstg: I rewrote most of the commentary in this section, I think it's more readable now.

specification/dartLangSpec.tex

specification/dartLangSpec.tex
eernstg: I really think it's important to mention that there is a check which is performed relative to all superinterfaces, and also that it can't be a property of 'correct member override' because we want to make that property specific to the relation to a single superinterface.

I moved this paragraph out of the bullet list, such that it is a comment on the section as a whole rather than a comment on one of the cases in the predicate. I think it's more readable now.

eernstg:
Rewrote several parts of 'Correct Member Overrides', PTAL.

specification/dartLangSpec.tex
```diff
@@ -5193,52 +5195,70 @@ \subsubsection{Correct Member Overrides}
...
-  % the same name. However, we must _allow_ $m$ and $m'$ to be two setters.
+\item
+  If $m$ and $m'$ are both getters:
+  The return type of $m$ must be a subtype of the return type of $m'$.
 \item
   If $m$ and $m'$ are both methods or both setters:
   Let $F$ be the function type of $m$
   except that the parameter type is the built-in class \code{Object}
-  for each parameter of $m$ which has the modifier \COVARIANT.
+  for each parameter of $m$ which is covariant-by-declaration
```
lrhn, line 5205: Do we only do this for covariant-by-declaration and not covariant-by-generics?

(Or handle covariant-by-generics in genereal: What if `m` is the signature of `void add(int x) {}` and `m'` is `add` from `List<int>`? It seems that `m` is still not a valid implementation of `m'` when cast to `List<Object>`.)

What is the real difference between the two?
* Covariant by declaration: Inherited, allows sub-type as parameter type in override.
* Covariant by generics: Inherited, does not allow sub-type as parameter type in override.

That suggests that we do need to distinguish the two in the "valid override" relation, but only in the case where the overriding method's parameter type is a subclass of the overridden one, in all other contexts, they should be treated equivalently?

Not sure. I'll think some more and come back.

specification/dartLangSpec.tex
```diff
@@ -5193,52 +5195,70 @@ \subsubsection{Correct Member Overrides}
...
+    The subtype requirement ensures that argument list shapes
+    that are admissible for an invocation of a method with signature $m'$
+    are also admissible for an invocation of a method with signature $m$.
+    For instance, $m'$ may accept 2 or 3 positional arguments,
+    and $m$ may accept 1, 2, 3, or 4 positional arguments, but not vice versa.
+    This is a built-in property of the function type subtype rules.
+  }
+
+  If $p'$ is a formal parameter in $m'$ which is covariant-by-declaration,
+  the corresponding parameter $p$ in $m$ must also be covariant-by-declaration
```
lrhn, line 5220: Again, should covariant by generics be included for `m'`?
So if `m'` is covariant-by-generics, then `m` must accept any argument.

For `m`, you don't need to say that it must be covariant, because if it was covariant, you already changed its parameter type to `Object` above (line 5204), so it is guaranteed to satisfy "$p$ is a top type". You can just have:

> the type of the corresponding parameter $p$ in $m$ must be a top type (which it is if $p$ is is covariant-by-declaration or covariant-by-generics).

specification/dartLangSpec.tex
```diff
@@ -5193,52 +5195,70 @@ \subsubsection{Correct Member Overrides}
...
+  unless the type of $p$ is a top type
+  (\ref{superBoundedTypes}).
+
+  \commentary{%
+    This requirement can actually be violated:
```
lrhn, line 5225: This phrase is confusing. Does it say that you can write valid programs which violates the rule? Or just that the rule is required because it's possible to write programs where it matters.
I'm guessing the latter, but that's not how I read it.

So, maybe:

>  This requirement matters in the situation where ....

specification/dartLangSpec.tex
eernstg: Yes, with covariant-by-class we require the tools to create forwarding methods as needed, cf. https://github.com/dart-lang/sdk/issues/31580#issuecomment-350420199, we only report an error for parameters that are covariant-by-declaration.

eernstg: > only in the case where the overriding method's parameter type is a subclass of the overridden one,

This would be a global property in any case: If the inherited concrete member has a parameter type which is a supertype of the overridden one, it could still be a non-supertype of an indirectly overridden one.

```dart
class A1 {}
class A2 {}
class B implements A1, A2 {}

class C {
  void foo(A1 b) { print(b.runtimeType); }
}

abstract class D {
  void foo(A2 b);
}

abstract class E implements D {
  void foo(covariant B c);
}

class F extends C implements E {}

void main() {
  D d = F();
  d.foo(A2());
}
```

specification/dartLangSpec.tex
eernstg: We decided to make it a non-error when the parameter is covariant-by-class. It isn't specified when the implementations need to implicitly induce a forwarding method to ensure soundness, we just specify that actual soundness violations must be detected (so an implementation can generate type checks everywhere, or it can optimize for removing as many as possibly&mdash;and using a forwarding method here and there is one of the tools available to reduce the number of generated type checks).

> if `m'` is covariant-by-generics, then `m` must accept any argument.

No, but there must be a dynamic check on the actual parameter type. An implicitly induced forwarder can do that.

> you already changed its parameter type

No, not quite: The function type `$F$` was computed in this manner, but the member signatures `$m$` and `$m'$` are derived from the program with no modifications.

specification/dartLangSpec.tex
lrhn: Argh.
So what are the rules that we actually want to specify here? I've lost track.

The *situation* we want to avoid is that a normal implementation method ends up implementing a covariant interface signature. That's not *necessarily* a problem (maybe there are no super-classes to upcast to which would make it unsound), but we see it as *potentially* unsound and want to refuse it. (We currently don't. AFAICS, both VM and dart2js introduce forwarders, at least in the simple cases I tested. Why don't we just specify that and call it a day?)

If that happens, the rules we try to specify here trigger and make it a compile-time error because a non covariant-by-declaration implementation tries to implement a covariant-by-declaration interface, and is presumably not able to handle the potential unsound upcast.
So, the rule should say that:
* if *p'* is covariant-by-declaration and *p* can't accept all arguments (which it can if its type is a top-type, or if it's covariant-by-generics or covariant-by-declaration, because then we know that it's prepared to be used in unsound situations), then *m* is not a valid override of *m'*.

So, something like:

> A member signature *m* is a valid override of a member signature *m'* iff all the following conditions hold:
> 
> * *m* and *m'* are either both getter signatures, both setter signatures or both method signatures.
> * If *m* and *m'* are getter or method signatures, then the return type of *m* is a subtype of the return type of *m'*.
> * If *m* and *m'* are setter or method signatures, then
>   * For each parameter *p* of *m*, if there is no corresponding parameter of *m'*, then *p* is optional.
>   * For each parameter *p'* of *m'* there is a corresponding parameter *p* of *m*, such that:
>     * If *p* is required, so is *p'*,
>     * Either 
>       * the type of *p* is a supertype of the type of *p'*, or
>       * the type of *p* is a subtype of the type of *p'* and *p* is covariant-by-declaration.
>     * If *p'* is covariant-by-declaration, then either *p* is covariant-by-declaration, *p* is covariant-by-generics, or the type of *p* is a top-type.

I wouldn't change the covariant parameter types of `m` to `Object` early. It seems lossy, and you never known when you'd need the information. For example, it won't be able to check the "subtype and covariant-by-declaration" rule here.

specification/dartLangSpec.tex
eernstg: I rewrote that commentary again, I think it's considerably more straightforward now! ;-)

It it intended to say that it is possible to write a program where the correct override relation does not hold, because this criterion is not satisfied. If that hadn't been the case then we wouldn't need the criterion at all.

When the correct override relation doesn't hold, it is specified elsewhere that it is an error:

We [specify in section 'Instance Methods'](https://github.com/dart-lang/language/blob/a3f50843c5937135f8542d04aa5a0e239158df7d/specification/dartLangSpec.tex#L2609) that it is an error if a class `C` declares a member, and an accessible member of the same name exists in a superinterface of `C`, and it is not a correct override. So that's the location where we connect 'correct override' and 'compile-time error' for the standard override relation.

[Section 'Classes'](https://github.com/dart-lang/language/blob/a3f50843c5937135f8542d04aa5a0e239158df7d/specification/dartLangSpec.tex#L2522) specifies that an inherited member in a concrete class must be a correct override of any members with the same name in superinterfaces. So that's the location where we connect 'correct override' and 'error' for inherited methods.

specification/dartLangSpec.tex
lrhn: When I say "accept any argument" I mean doing something semantically correct for it, rather than, well, *actually* accepting something of the wrong type. So bad phrasing, what I meant was "Must behave soundly for all arguments", which means properly throwing if an argument allowed by the covariant upcast is not valid for the actual implementation.

specification/dartLangSpec.tex
lrhn: I made some tests: https://dartpad.dev/?id=91a6cfc411dc4cdf9c392ea49a62b9aa&

It does seem that we currently add forwarders for both covariant-by-generics and covariant-by-declaration.
Should we just accept and specify that? It seems to work, and anything touching `covariant` is prone to run-time errors, so I wouldn't expect a compile-time error.

eernstg: > So what are the rules that we actually want to specify here

We're specifying the relationship between the properties of two corresponding formal parameters in a correct override. One property is the declared type, the other one is covariance-by-declaration.

When the parameter is covariant-by-declaration, the types are checked by the rule in section 'Classes' (so we disable that check here by building an `$F$` with a parameter type `Object`/`Object?` for those parameters). When it isn't, we rely on the function type subtype rules.

> avoid [..] normal implementation [..] implementing a covariant interface signature. That's not necessarily a problem

I agree that we may be able to carve out a little bit of extra flexibility: No dynamic check is required when the overriding parameter has a type which is a supertype of all the types of the (directly or indirectly) overridden parameters (covariant or not, the important point is simply that we must take into account all possible static types of the receiver where that member could be called).

> (We currently don't. AFAICS, both VM and dart2js introduce forwarders,
> at least in the simple cases I tested. Why don't we just specify that and call it a day?)

I argued many times that we should do that, but the decision was to not do it, cf. https://github.com/dart-lang/sdk/issues/31580#issuecomment-350420199. ;-)

But we do get the error in some cases, including Paul's original example:

```
class A {}
class B extends A {}
class C {
  void f(B x, B y) {}
}
abstract class I1 {
  void f(covariant A x, B y);
}
class D extends C implements I1 {}
```

The analyzer flags `D` as an error, indicating that it is not a correct override.

> So, something like:

We could rewrite these rules to re-state the function type subtype requirements, but that introduces a substantial amount of redundancy that we must then maintain, and it may not be obvious to a reader that this is all the same stuff that we're saying about subtypes, repeated.

> `the type of p is a subtype of the type of p' and p is covariant-by-declaration.`

This is redundant with the all-superinterfaces rule, so that's even more redundancy to keep in sync (and I do not think we should change 'correct override' to be a relationship between a member signature and _all_ the member signatures that it overrides, because we don't want to imply that all the checks must be made globally&mdash;the global check should preferably only be performed in the presumably rare cases where we actually have a cov-by-decl parameter).

> `either p is covariant-by-declaration, p is covariant-by-generics`

It is again an interesting generalization that we could allow the inheritance when `p` is cov-by-class, because that method must then already perform a dynamic check on the actual argument type.

specification/dartLangSpec.tex
eernstg: I think we agree here, so I'll close this thread because we have enough threads already. ;)

specification/dartLangSpec.tex
lrhn: But we do introduce forwarders in some cases, like when only the covariance marker differs, but the type without the covariance is actually fine. Do we really want to make that an error instead? (It seems reasonable to users for all the same reasons it does for covariance-by-generic).

So, if the function signature of *m* (with `Object` for covariant parameters, either by-declaration or both) is a subtype of the function signature of *m'*, there is no problem, even if *p'* is covariant-by-generics.
That seems to be what we have implemented.

If it's not a subtype, then it must be because a parameter *p* of *m*, has a type that is not a supertype of the type of the corresponding parameter *p'* of *m'*. Since we did the `Object` change, that means *p* is not covariant. So, we catch that case automatically. 

That only leaves the case where the type of *p* is a supertype of the type of *p'*, *p* is not covariant and *p'* is ... which is not currently an error, and which you are trying to make one. Not sure it's worth it.

eernstg: I'd be quite happy about asking for forwarding methods whenever needed, and then we wouldn't have the cov-by-decl error at all.

eernstg: But it is not exactly what we have implemented: Paul's original example is rejected at compile-time.

lrhn: In that example (if it's the one above), the interface requires `A` and the implementation only accepts the subtype `B`. So, that's not a valid implementation of the interface if you ignore covariance, and you ignore covariance because `C`'s implementation method is not coviriant.
If the implementation had accepted an `A` as argument, then it would be sound ignoring covariance, and then the forwarder would just be for allowing upcasts, not changing the visible implementation from accepting `B` to accepting `A`.

eernstg: Actually, the CFE could cut down on the forwarding stubs: The only situation where a cov-by-decl parameter `p` with type `T` must have a dynamic check is when there is a supertype of the receiver where that parameter has a non-subtype of `T`.

For instance, the CFE would not have to generate the forwarder in the following case:

```dart
class A {}

class B extends A {}

class C {
  void f(A x) {}
}

abstract class I1 {
  void f(covariant A x);
}

class D extends C implements I1 {}

void main() {}
```

But it does generate that forwarder:

```
library /*isNonNullableByDefault*/;
import self as self;
import "dart:core" as core;

class A extends core::Object {
  synthetic constructor •() → self::A
    : super core::Object::•()
    ;
}
class B extends self::A {
  synthetic constructor •() → self::B
    : super self::A::•()
    ;
}
class C extends core::Object {
  synthetic constructor •() → self::C
    : super core::Object::•()
    ;
  method f(self::A x) → void {}
}
abstract class I1 extends core::Object {
  synthetic constructor •() → self::I1
    : super core::Object::•()
    ;
  abstract method f(covariant self::A x) → void;
}
class D extends self::C implements self::I1 {
  synthetic constructor •() → self::D
    : super self::C::•()
    ;
  forwarding-stub method f(covariant self::A x) → void
    return super.{self::C::f}(x);
}
static method main() → void {}
```

eernstg: > If the implementation had accepted an A as argument, then it would be sound ignoring covariance

This is a non-local property: We're comparing the type of the formal parameter of the implementation to all the types of the corresponding formal parameters of overridden declarations, directly or indirectly.

So if we want to make a distinction between "good" and "bad" occurrences of inheritance where covariance-by-decl is added to a parameter of an implementation, it can't be expressed in terms of a correct override relation to a direct superinterface, it must be in terms of the complete set of superinterfaces.

We did not make that distinction when we discussed this topic.

I think it would be considerably nicer to allow addition of covariance-by-decl implicitly, and ask tool teams to generate the required forwarders, rather than starting to carve out all these exceptions: The exceptions will just be some cases where an implementation doesn't have to generate a forwarder anyway.

specification/dartLangSpec.tex
eernstg: I think this thread is done, too.

specification/dartLangSpec.tex
eernstg: The newest commit removes the criterion about covariant parameters and adds a sentence about forwarding methods to the commentary.

It is then up to the implementations to detect exactly where there is a need for a forwarding method, and various carved out pieces of flexibility are now a matter of optimization: Developers are allowed to rely on covariance-by-declaration being added to an inherited method, just like covariance-by-class was always added as needed.

eernstg: @lrhn, I landed this before the vacation, but we did have some ongoing discussions. Do you wish to take another turn in another PR?

lrhn: Let's discuss what we want before writing more spec.

I remember loosing track a few times, back when I could remember what this even was, so let's go slowly and make sure we are talking about the same thing. :)

eernstg: Sounds good! I'll close this thread because the further discussions will occur elsewhere.

PR #1744, opened 2021-07-15
Adjust the views proposal to use `super`

eernstg:
The current views feature specification uses `this` to denote the current receiver of a view instance method invocation using the on-type as the static type. This proposal gives `this` the enclosing view type as the static type, and it uses `super` to access the receiver with the on-type as the static type.

This PR is intended to be an option: We could land it, or we could drop it, depending on the underlying language design decision. The PR just illustrates that it is a rather small change in terms of specification language. It may be less convenient from an implementation point of view (e.g., `return super;` should be possible, and that seems to be a rather deep syntactic change), but it might be possible to adjust the lexer such that it produces an `<identifier>` when it encounters `super` inside a `view` declaration, and then treat that identifier as a final variable whose type is the on-type.

### DIFF
--- a/working/1426-extension-types/feature-specification-views.md
+++ b/working/1426-extension-types/feature-specification-views.md
@@ -32,12 +32,12 @@ the set of members, with subsetting as a special case.
 
 The functionality is entirely static. Invocation of a view member is
 resolved at compile-time, based on the static type of the receiver.  Inside
-the view declaration, the scoping and the type and meaning of `this` is the
-same as for extension methods (a feature which was added to Dart in version
-2.6). This is important because it implies that the language Dart has a
-single and consistent semantics for all statically resolved member
-invocations, rather than having one set of rules for extension methods, and
-a different set of rules for view members.
+the view declaration, the reserved word `super` is treated as an identifier
+denoting a final variable whose type is the on-type of the enclosing view
+declaration. At run time, the value of `super` is the current on-type instance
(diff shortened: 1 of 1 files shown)


working/1426-extension-types/feature-specification-views.md
```diff
@@ -32,12 +32,12 @@ the set of members, with subsetting as a special case.
...
-2.6). This is important because it implies that the language Dart has a
-single and consistent semantics for all statically resolved member
-invocations, rather than having one set of rules for extension methods, and
-a different set of rules for view members.
+the view declaration, the reserved word `super` is treated as an identifier
+denoting a final variable whose type is the on-type of the enclosing view
+declaration. At run time, the value of `super` is the current on-type instance
+of the view. Similarly, the type of `this` is the view type declared by the
+enclosing declaration, and the run-time value of `this` is also the current
+on-type instance.
```
lrhn, line 40: This requires grammar changes.

Probably just making `super` a `primary`, where currently the primary contains:
```ebnf
  \alt \SUPER{} <unconditionalAssignableSelector>
  \alt \SUPER{} <argumentPart>
```
Then make it a compile-time error if `super` is not followed by an unconditinoalAssignableSelector or an argumentPart unless the `super` occurs inside a member declaration of a view.

eernstg:
Many hours of discussion later, we've settled on an approach where this idea is no longer relevant. Closing.

PR #1746, opened 2021-07-16
Mention cascades in constructor-tearoffs feature spec

eernstg:
The 'constructor-tearoffs' feature specification does not mention cascades. We could make them a compile-time error (because things like `x..instanceMethod<int>` has no effect), but we allow things like `myLocalVariable;` as a statement as well. So I chose to include support for instantiated tearoffs in cascades, but added commentary saying that it isn't likely to be useful.

### DIFF
--- a/accepted/future-releases/constructor-tearoffs/feature-specification.md
+++ b/accepted/future-releases/constructor-tearoffs/feature-specification.md
@@ -317,7 +317,24 @@ It applies to instance methods as well as local, static and top-level function d
 
 `instanceMethod<int>` (with implicit `this`), `object.instanceMethod<int>` (including `this`) and `super.instanceMethod<int>`.
 
-Cascades can contain explicitly instantiated tearoffs, e.g., `receiver..foo()..instanceMethod<int>..bar`. *Note that this is allowed for consistency, but it will compute a value and discard it, so it is hardly useful.*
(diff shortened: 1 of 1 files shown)


accepted/future-releases/constructor-tearoffs/feature-specification.md
```diff
@@ -317,6 +317,8 @@ It applies to instance methods as well as local, static and top-level function d
 
 `instanceMethod<int>` (with implicit `this`), `object.instanceMethod<int>` (including `this`) and `super.instanceMethod<int>`.
 
+Cascades can contain explicitly instantiated tearoffs, e.g., `receiver..foo()..instanceMethod<int>..bar`. *Note that this is allowed for consistency, but it will compute a value and discard it, so it is hardly useful.*
```
lrhn, line 320: Not necessarily true, since you can call extension members on the function, maybe something like `apply` which calls `Function.apply`:
```dart
foo
  ..bar<int>.apply(argumentList)
  ..bar<String>.apply(otherArgumentList);
```
Probably not going to be widely used, but still potentially useful.

lrhn (approved):
LGTM, thanks.

accepted/future-releases/constructor-tearoffs/feature-specification.md
eernstg: Of course! Added a small example to this effect.

PR #1760, opened 2021-07-27
Replace "white box" with "clear box" testing

natebosch:
The terms "white box" and "black box" are less clear in isolation than
"clear box" and "opaque box".


munificent:
Thanks!

Issue #1763, opened 2021-07-28
False positive with the `late` keyword when using `const factory` constructor


rrousselGit:
Consider the following code:

```dart
class Base {
  Base();

  const factory Base.empty() = _Empty;

  late final property = ...
}

class _Empty implements Base {
  const _Empty();

  @override
  get property => ...
}
```

This code currently does not compile with "Can't have a late final field in a class with a const constructor.
Try removing the 'late' modifier, or don't declare 'const' constructors."

This error is caused by a `const factory`. But since this is a factory constructor, this should not have any impact.

eernstg:
True. Considering the issue in https://github.com/dart-lang/language/pull/1764.

eernstg:
Done in https://github.com/dart-lang/language/pull/1764.

PR #1764, opened 2021-07-28
Adjust nnbd spec to allow a const factory in a class with a late final variable

eernstg:
Cf. issue https://github.com/dart-lang/language/issues/1763. It looks like we can allow a const factory constructor (which is then necessarily redirecting) in a class that has a late final instance variable.

There will be an implementation effort (it is currently, correctly according to the spec, reported as an error).


lrhn (approved):
Add an entry to the changelog at the top as well.

Otherwise LGTM.

PR #1768, opened 2021-07-28
[macros] Flesh out some of the ordering behavior a little more.

munificent:
Trying to give readers a little more context on the challenges and pin
down the semantics a bit more precisely. For now, I'm just going with
the simplest approach for separate compilation, but I put a TODO in
there pointing to modules since we'll likely end up going there.


jakemac53 (approved):
🥳

PR #1770, opened 2021-07-29
Add rule about covariant-by-decl parameter when inherited

eernstg:
In the case where a parameter `p` is covariant-by-declaration in a superinterface, and a concrete member is inherited from a superclass which has a parameter `p1` that corresponds to `p`, the type of `p1` is required to be a subtype or supertype of each type of a parameter that corresponds to `p1` in all superinterfaces.

We already have this rule for a declaration which occurs in the class itself, but (by mistake) we did not have the rule when a method is inherited. This PR adds the rule for that case.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -26,6 +26,10 @@
 %
 % Significant changes to the specification.
 %
+% 2.14
+% - Add constraint on type of parameter which is covariant-by-declaration in
+%   the case where the method is inherited (that case was omitted by mistake).
+%
 % 2.12 - 2.13 (there was no 2.11)
 % - Revert the CL where certain null safety features were removed (to enable
 %   publishing a stable version of the specification).
@@ -2531,10 +2535,21 @@ unless $C$ has a non-trivial \code{noSuchMethod}
 (\ref{theMethodNoSuchMethod}).
 It is a compile-time error if $C$ has
 a concrete accessible member with the same name as $m$,
-with a method signature $m'$ which is not a correct override of $m$
+with a method signature \DefineSymbol{m'}
+which is not a correct override of $m$
 (\ref{correctMemberOverrides}),
 unless that concrete member is a \code{noSuchMethod} forwarder
 (\ref{theMethodNoSuchMethod}).
+
+\LMHash{}%
+Consider the case where $m'$ is a correct override of $m$.
+For each parameter $p$ of $m'$ where \COVARIANT{} is present,
+it is a compile-time error if there exists
+a direct or indirect superinterface of $C$ which has
+an accessible method signature $m''$ with the same name as $m$,
+such that $m''$ has a parameter $p''$ that corresponds to $p$
+(\ref{covariantParameters}),
+unless the type of $p$ is a subtype or a supertype of the type of $p''$.
 
 \commentary{%
 In particular, it is an error for a class to be concrete even if it inherits
@@ -2635,7 +2650,7 @@ a direct or indirect superinterface of $C$ which has
 an accessible method signature $m''$ with the same name as $m$,
 such that $m''$ has a parameter $p''$ that corresponds to $p$
 (\ref{covariantParameters}),
-unless the type of $p$ is assignable to the type of $p''$.
+unless the type of $p$ is a subtype or a supertype of the type of $p''$.
 
 \commentary{%
 This means that
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -2528,28 +2528,18 @@ that name denotes the interface of the class.
 A concrete class must fully implement its interface:
 \BlindDefineSymbol{C, I, m}%
 Let $C$ be a concrete class with interface $I$.
-Assume that $I$ has an accessible member signature $m$.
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -2531,11 +2535,22 @@ \section{Classes}
 (\ref{theMethodNoSuchMethod}).
 It is a compile-time error if $C$ has
 a concrete accessible member with the same name as $m$,
-with a method signature $m'$ which is not a correct override of $m$
+with a method signature \DefineSymbol{m'}
+which is not a correct override of $m$
```
lrhn, line 2539: So if `m'` is `int Function(covariant int)` and `m` is `int Function(num)`, then `m'` is a correct override of `m`?
But `m'` as `int Function(int)` is not a correct override of `m` as `int Function(covariant num)`?

(I think this is probably correct for what we currently intended to specify, but I do want both to be invalid. Eventually :).

lrhn (approved):

specification/dartLangSpec.tex
eernstg: First a few remarks on the newest upload:

I added a few words at 'must fully implement', to make it explicit that this rule is about inherited methods, not about methods that are declared in `$C$`. That should be a benign change (because the declarations in `$C$` must satisfy the 'correct override' constraint as well, but that's in a different section).

Next, I moved the new paragraph down below the commentary, in order to avoid some confusion about the topic of that commentary.

I added a few words to the new paragraph (introducing `m'` afresh, because it's now further away). I think that's clearer.

Now, about the actual rule. I think it says what you want, and also that it was possible to get this effect without too many twisted turns. So that's nice!

The question about whether any given parameter in the inherited signature `m'` is cov-by-decl is answered by checking `m'` itself. That is, a parameter inherited from `B` is _not_ cov-by-decl at this point if it isn't known to be cov-by-decl in `B`.

This ensures that we get an error for "magically becoming cov-by-decl when inherited" (using the rule in Correct Member Overrides), e.g.

```dart
class A { m(num n) {}}
abstract class I { m(covariant num n); }
class B extends A implements I {} // Error.
```

> So if `m'` is `int Function(covariant int)` and `m` is `int Function(num)`, 
> then `m'` is a correct override of `m`?

I think we couldn't actually have those member signatures:

`m'` is the inherited signature (the spec writes that as `int m(covariant int)` because we don't support `covariant` in function types) and `m` is the one from the interface (`int m(num)`).

But if the parameter is cov-by-decl in any superclass then it would also be cov-by-decl in the interface of the current class.

But if the superclass has signature `int m(int)` and the class interface has `int m(covariant num)` then we get a compile-time error: `int m(int)` is not a correct override. We don't "magically" add `covariant` to the inherited signature, and it's a violation of the cov-by-decl rule in 'Correct Member Overrides' when `covariant` is missing.

Consider the opposite situation:

```dart
class A { m(covariant int i) {}}
abstract class I { m(num n); }
class B extends A implements I {} // OK.
```

This is actually accepted: The inherited signature is `dynamic m(covariant int)` and that is a correct override of `dynamic m(num)`.

It seems rather reasonable to me, but do you want this to be an error?

> But `m'` as `int Function(int)` is not a correct override of `m` as `int Function(covariant num)`?

So this would correspond to the following code:

```dart
class A { int m(int i) {}}
abstract class I { int m(covariant num n); }
class B extends A implements I {}
```

Yes, that would be an error, because `int m(int)` is not a correct override of `int m(covariant num)`: `covariant` is missing (and more, but that's enough to make it an error).

eernstg:
Several issue raised, PTAL.

specification/dartLangSpec.tex
```diff
@@ -2528,29 +2528,19 @@ \section{Classes}
 A concrete class must fully implement its interface:
 \BlindDefineSymbol{C, I, m}%
 Let $C$ be a concrete class with interface $I$.
-Assume that $I$ has an accessible member signature $m$.
+Assume that $I$ has an accessible member signature $m$,
+and $C$ does not declare a member with the same name as $m$.
```
lrhn, line 2532: Reasonable constraint since a member declared in the class will force the class interface member signature to match.

specification/dartLangSpec.tex
lrhn: I think this is fine, and we'll have to further discuss whether
```dart
class A { m(covariant int i) {}} 
abstract class I { m(num n); } 
class B extends A implements I {} // OK.
```
should be an error too (which I think it should because `B().m(2.2)` is known to throw a type error, meaning that the class doesn't implement it *own* interface, without even needing to up-cast it, so it's not a *covariance* based error, it's just a plain error).

lrhn (approved):

specification/dartLangSpec.tex
eernstg: Right, this is basically just removing some redundancy: The rule in 'Instance Methods' takes care of declarations in `$C$`, this rule takes care of declarations inherited by `$C$` (when that's a concrete class). Without line 2532, this rule covers both (there is no contradiction, but the redundancy is confusing).

eernstg:
Very good, thanks!

PR #1778, opened 2021-08-04
add table of contents to macro doc

jakemac53:
This doc is starting to get longer so I think a TOC is in order :).


munificent (approved):
P R O F E S S I O N A L

PR #1779, opened 2021-08-04
initial pass on the scoping section

jakemac53:
I chose for now to make it an error to introduce an ambiguity. Open to discuss other options though :)

### DIFF
--- a/working/macros/feature-specification.md
+++ b/working/macros/feature-specification.md
@@ -308,8 +308,8 @@ class, allowing you to build up your code fragments however you like.
 
 Macros will likely want to introduce references to identifiers that are not in
 the scope of the library in which they are running, but are in the scope of the
-macro itself, or possibly even in some other library that is known to the
-macro, but not imported by the library where the macro is applied.
+macro itself, or possibly even references which are not in scope of either the
+macro itself or the library where it is applied.
 
 Even if an identifier is expected to be in scope of the library in which the
 macro is running (lets say its exported by the macro library), that identifier
@@ -331,9 +331,21 @@ if it happens.
 
 ### Generated declarations
 
-Macros may add new declarations which collide with existing symbols in the
-library. We want to ensure that the intent of any user written code is always
-clear in this case. Consider the following example:
+A key use of macros is to generate new declarations, and hand-authored code may
+refer to them — it may call macro-generated functions, read macro-generated
+fields, construct macro-generated classes, etc. This means that before macros
+are applied, code may contain identifiers that cannot be resolved. This is not
+an error. Any identifier that can't be resolved before the macro is applied is
+allowed to be resolved to a macro-produced identifier after macros are applied.
+
+Macros are not permitted to introduce declarations that directly conflict with
+existing declarations in the same library. These rules are the same as if the
+code were hand written.
+
+Macros may also end up adding new declarations which shadow existing symbols in
+the library, but don't directly conflict. In this case we want to ensure that
+the intent of any user written code is always clear. Consider the following
+example:
 
 ```dart
 int get x => 1;
@@ -347,14 +359,46 @@ class Bar {
 }
 ```
 
-Every choice here has a lot of downsides, so it is instead an error. More
-specifically, if any identifier *could be* resolved prior to macros running,
-then a macro cannot alter the program in a way that changes *how* that
-identifier would resolve.
+There are several potential choices to we could make here:
+
+1.  We could say that any identifier that can be resolved before macro
+    application keeps its original resolution (so `x` would still resolve to the
+    original, top level `x`).
+2.  We could re-resolve all identifiers after the macros are applied, which can
+    possibly change what they resolve to (in this case `x` would resolve to the
+    generated instance getter `x`).
+3.  We could make it some kind of error for a macro to introduce an identifier
+    that shadows another.
+4.  We could make it error to use an identifier shadowed by one produced by a
+    macro.
+
+The first two choices could be very confusing to users, some will expect one
+behavior while others expect the other. The third choice would work but might be
+overly restrictive. The final option still avoids the ambiguity, and is a bit
+more permissive than the third.
+
+It similarly also is not allowed for one macro to produce an declaration that
+the identifier resolves to and then another macro to produce another declaration
+that then shadows that one. In other words, any hand-authored identifier may be
+resolved at any point during macro application, but it may only be resolved
+once.
+
+At the same time, a macro application in a library must be able to shadow
+identifiers in other libraries, in order to allow for modular compilation. It
+shouldn't be the case that a library can cause an error in one of the libraries
+that it imports.
+
+These constraints produce this rule:
+
+*   It is a compile-time error if any hand-authored identifier in a library
+    containing a macro application would bind to a different declaration when
+    resolved before and after macro application. In other words, it is a
+    compile-time error if a macro introduces an identifier that shadows a
+    hand-authored identifier that is used in the same library.
 
 This follows from the general principle that macros should not alter the
-meaning of existing code. Adding the getter `x` here shadows the top level `x`,
-changing the meaning of the original code.
+meaning of existing code. Adding the getter `x` in the example above shadows the
+top level `x`, changing the meaning of the original code.
 
 Note, that if the getter were written as `int get y => this.x;`, then a macro
 *would* be allowed to introduce the new getter `x`, because `this.x` could not
@@ -363,10 +407,6 @@ previously be resolved.
 There is an exception to this rule, which is that macros are always allowed to
 override declarations from their super types, and that is not considered to be
 a violation of this rule.
-
-Macros similarly cannot introduce declarations that directly conflict with
-existing declarations in the same library. These rules are the same as if the
-code were hand written.
 
 ## Limitations
 
--- a/working/macros/feature-specification.md
+++ b/working/macros/feature-specification.md
@@ -304,7 +304,7 @@ class, allowing you to build up your code fragments however you like.
 
 ## Scoping
 
-### Generated identifiers
+### Resolved identifiers
 
 Macros will likely want to introduce references to identifiers that are not in
 the scope of the library in which they are running, but are in the scope of the
--- a/working/macros/feature-specification.md
+++ b/working/macros/feature-specification.md
@@ -316,12 +316,12 @@ macro is running (lets say its exported by the macro library), that identifier
 could be shadowed by another identifier in the library.
 
 To enable a macro to safely emit a reference to a known identifier, there is
-a `ResolvedIdentifier` subtype of `Code`. This class takes both a simple name
-for the identifier (no prefix allowed), as well as a library uri, where that
-identifier should be looked up.
+a `Identifier` subtype of `Code`. This class takes both a simple name for the
+identifier (no prefix allowed), as well as a library URI, where that identifier
+should be looked up.
 
 The generated code should be equivalent to adding a new import to the library,
-with the specified uri and a unique prefix. In the code the identifier will be
+with the specified URI and a unique prefix. In the code the identifier will be
 emitted with that unique prefix followed by its simple name.
 
 Note that technically this allows macros to add references to libraries that
@@ -332,7 +332,7 @@ if it happens.
 ### Generated declarations
 
 A key use of macros is to generate new declarations, and hand-authored code may
-refer to them — it may call macro-generated functions, read macro-generated
+refer to them—it may call macro-generated functions, read macro-generated
 fields, construct macro-generated classes, etc. This means that before macros
 are applied, code may contain identifiers that cannot be resolved. This is not
 an error. Any identifier that can't be resolved before the macro is applied is
@@ -340,12 +340,11 @@ allowed to be resolved to a macro-produced identifier after macros are applied.
 
 Macros are not permitted to introduce declarations that directly conflict with
 existing declarations in the same library. These rules are the same as if the
-code were hand written.
+code were handwritten.
 
-Macros may also end up adding new declarations which shadow existing symbols in
-the library, but don't directly conflict. In this case we want to ensure that
-the intent of any user written code is always clear. Consider the following
-example:
+Macros may also add declarations which shadow existing symbols in the library,
+but don't directly conflict. In this case we want to ensure that the intent of
+any user written code is always clear. Consider the following example:
 
 ```dart
 int get x => 1;
@@ -369,16 +368,16 @@ There are several potential choices to we could make here:
     generated instance getter `x`).
 3.  We could make it some kind of error for a macro to introduce an identifier
     that shadows another.
-4.  We could make it error to use an identifier shadowed by one produced by a
-    macro.
+4.  We could make it a compile-time error to *use* an identifier shadowed by one
+    produced by a macro.
 
 The first two choices could be very confusing to users, some will expect one
 behavior while others expect the other. The third choice would work but might be
 overly restrictive. The final option still avoids the ambiguity, and is a bit
 more permissive than the third.
 
-It similarly also is not allowed for one macro to produce an declaration that
-the identifier resolves to and then another macro to produce another declaration
+It similarly also is not allowed for one macro to produce a declaration that the
+identifier resolves to and then another macro to produce another declaration
 that then shadows that one. In other words, any hand-authored identifier may be
 resolved at any point during macro application, but it may only be resolved
 once.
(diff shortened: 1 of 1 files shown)


working/macros/feature-specification.md
```diff
@@ -304,20 +304,69 @@ class, allowing you to build up your code fragments however you like.
...
 ## Scoping
 
-**TODO**: Fill in this section with more detail and a real proposal.
+### Resolved identifiers
 
 Macros will likely want to introduce references to identifiers that are not in
 the scope of the library in which they are running, but are in the scope of the
-macro itself.
+macro itself, or possibly even in some other library that is known to the
+macro, but not imported by the library where the macro is applied.
```
munificent, line 312: "known to the macro" is vague. Instead of "in the scope of the macro itself or possibly...", I would just say: "in scope where the macro is defined and not where it is applied".

working/macros/feature-specification.md
```diff
@@ -304,20 +304,69 @@ class, allowing you to build up your code fragments however you like.
...
+identifier should be looked up.
+
+The generated code should be equivalent to adding a new import to the library,
+with the specified uri and a unique prefix. In the code the identifier will be
+emitted with that unique prefix followed by its simple name.
+
+Note that technically this allows macros to add references to libraries that
+the macro itself does not depend on, and the users application also may not
+depend on. This is discouraged, but not prevented, and should result in an error
+if it happens.
```
munificent, line 330: I'm not crazy about this. It adds a *lot* of power to let a macro freely compose a library URI and synthesize an identifier in its scope, and I don't see much justification for that. It means you could do potentially really weird stuff like have a macro application create an import cycle that wasn't there previously.

My initial pitch back when I wrote the Code API strawman is that there is a Scope object that you can use to create identifiers that resolve to that scope. Then there would be some way for a macro definition to get a Scope for the current scope where the macro is defined, and maybe one for where the macro is applied. I think that's probably all we need?

working/macros/feature-specification.md
```diff
@@ -304,20 +304,69 @@ class, allowing you to build up your code fragments however you like.
...
+
+### Generated declarations
+
+Macros may add new declarations which collide with existing symbols in the
```
munificent, line 334: "collide with" -> "shadow".

working/macros/feature-specification.md
```diff
@@ -304,20 +304,69 @@ class, allowing you to build up your code fragments however you like.
...
+@generateX 
+class Bar {
+  // Generated: int get x => 2;
+
+  // Should this return the top level `x`, or the generated instance getter?
+  int get y => x; 
+}
+```
+
+Every choice here has a lot of downsides, so it is instead an error. More
```
munificent, line 350: It might not be clear to the reader what "choices" are being made. Maybe spell it out:

> We could say that any identifier that can be resolved before macro application keeps its original resolution. Or we could re-resolve all identifiers after the macros are applied, which can possibly change what they resolve to. We could make it some kind of error for a macro to introduce an identifier that shadows another. Or we could make it error to *use* an identifier shadowed by one produced by a macro.
>
> Every choice here has downsides...

working/macros/feature-specification.md
```diff
@@ -304,20 +304,69 @@ class, allowing you to build up your code fragments however you like.
...
+specifically, if any identifier *could be* resolved prior to macros running,
+then a macro cannot alter the program in a way that changes *how* that
+identifier would resolve.
```
munificent, line 353: "An error" isn't super precise. I would say:

> It is a compile-time error if any identifier in the program would bind to a different declaration when resolved before and after a macro application. In other words, it is a compile-time error if a macro introduces an identifier that shadows an identifier that is used where the generated identifier is in scope.

Also, I think we should explicitly point out that code may contain references to non-existent identifiers. Maybe something like:

> A key use of macros is to generate new declarations, and hand-authored code may refer to them—it may call macro-generated functions, read macro-generated fields, construct macro-generated classes, etc. This means that before macros are applied, code may contain identifiers that *cannot* be resolved. This is *not* an error. Any identifier that can't be resolved before the macro is applied is allowed to be resolved to a macro-produced identifier after macros are applied.
>
> However, it is *not* allowed for one macro to produce an declaration that the identifier resolves to and then another macro to produce another declaration that then shadows that one. In other words, any hand-authored identifier may be resolved at any point during macro application, but it may only be resolved *once*.

Although... oof... that all sounds pretty hairy.

working/macros/feature-specification.md
```diff
@@ -304,20 +304,69 @@ class, allowing you to build up your code fragments however you like.
...
+meaning of existing code. Adding the getter `x` here shadows the top level `x`,
+changing the meaning of the original code.
+
+Note, that if the getter were written as `int get y => this.x;`, then a macro
+*would* be allowed to introduce the new getter `x`, because `this.x` could not
+previously be resolved.
+
+There is an exception to this rule, which is that macros are always allowed to
+override declarations from their super types, and that is not considered to be
+a violation of this rule.
```
munificent, line 365: I would be surprised if we needed this exception. Why do you think we might?

working/macros/feature-specification.md
```diff
@@ -304,20 +304,69 @@ class, allowing you to build up your code fragments however you like.
...
+
+Macros similarly cannot introduce declarations that directly conflict with
+existing declarations in the same library. These rules are the same as if the
+code were hand written.
```
munificent, line 369: I'd consider hoisting this before all the previous stuff since colliding in the *same* scope is simpler and more fundamental than *shadowing*, which is what the rest is about.

munificent (approved):
This looks good!

working/macros/feature-specification.md
jakemac53: I restructured this section, calling out 4 explicit choices and then explaining the logic for the final rule. PTAL.

Note that I changed the rule a bit such that this rule doesn't apply to any identifier in the program (due to concerns about that breaking modular compilation).

working/macros/feature-specification.md
jakemac53: Reworked this a bunch

working/macros/feature-specification.md
munificent: Ooh, that's a really good point. I think that's going to require more thought and design. Maybe file a tracking issue so we can work through it outside of this PR?

working/macros/feature-specification.md
munificent: That should be allowed even without the exception. Before the macro is applied, `x` will be resolved to `this.x`. After the macro is applied, it's still `this.x`. The actual `x` member invoked *polymorphically at runtime* changes, but the static resolution of what `x` refers to is the same. The only reason you'd need this exception is if you wanted to allow:

```dart
// a.dart
class A {
  int get x => 1;
}

// b.dart
import 'a.dart';

int x = 'top level';

@generateX
class B extends A {
  // Generated: int get x => 2;
  
  int get y => x; 
}
```

But I *don't* think we want to allow this, do you?

working/macros/feature-specification.md
```diff
@@ -304,20 +304,109 @@ class, allowing you to build up your code fragments however you like.
...
 the scope of the library in which they are running, but are in the scope of the
-macro itself.
+macro itself, or possibly even references which are not in scope of either the
+macro itself or the library where it is applied.
 
-We will want some way of providing an affordance to emit a reference to
-something from the macro scope, but in the code generated for the original
-library.
+Even if an identifier is expected to be in scope of the library in which the
+macro is running (lets say its exported by the macro library), that identifier
```
munificent, line 315: "in which the macro is running" is vague since "running" is a dynamic and not textual concept. How about: 

> Even if an identifier is in scope where the macro is applied...

working/macros/feature-specification.md
```diff
@@ -304,20 +304,109 @@ class, allowing you to build up your code fragments however you like.
...
+could be shadowed by another identifier in the library.
 
-We do have the start of something like this already available in the `builder`
-api - these have APIs to get a reference to a `Type` object. We will want to add
-the ability to do the same for any arbitrary identifier, and then the ability
-to emit references to these inside of [Code][] objects.
+To enable a macro to safely emit a reference to a known identifier, there is
+a `ResolvedIdentifier` subtype of `Code`. This class takes both a simple name
```
munificent, line 319: Are there *un*resolved identifiers? If not, how about just `Identifier`?

working/macros/feature-specification.md
```diff
@@ -304,20 +304,109 @@ class, allowing you to build up your code fragments however you like.
...
+for the identifier (no prefix allowed), as well as a library uri, where that
```
munificent, line 320: Nit: "uri" -> "URI".

working/macros/feature-specification.md
```diff
@@ -304,20 +304,109 @@ class, allowing you to build up your code fragments however you like.
...
+
+Note that technically this allows macros to add references to libraries that
+the macro itself does not depend on, and the users application also may not
+depend on. This is discouraged, but not prevented, and should result in an error
+if it happens.
+
+### Generated declarations
+
+A key use of macros is to generate new declarations, and hand-authored code may
+refer to them — it may call macro-generated functions, read macro-generated
```
munificent, line 335: Super nit: Chicago Manual of Style does not put spaces around em dashes. (This is what writing and editing the book has done to me.)

working/macros/feature-specification.md
```diff
@@ -304,20 +304,109 @@ class, allowing you to build up your code fragments however you like.
...
+fields, construct macro-generated classes, etc. This means that before macros
+are applied, code may contain identifiers that cannot be resolved. This is not
+an error. Any identifier that can't be resolved before the macro is applied is
+allowed to be resolved to a macro-produced identifier after macros are applied.
+
+Macros are not permitted to introduce declarations that directly conflict with
+existing declarations in the same library. These rules are the same as if the
+code were hand written.
```
munificent, line 343: Also, "handwritten" is one word. What have I become?

working/macros/feature-specification.md
```diff
@@ -304,20 +304,109 @@ class, allowing you to build up your code fragments however you like.
...
+
+Macros may also end up adding new declarations which shadow existing symbols in
```
munificent, line 345: Nit: "end up adding new" -> "add". Brevity is the soul of wit.

working/macros/feature-specification.md
```diff
@@ -304,20 +304,109 @@ class, allowing you to build up your code fragments however you like.
...
+
+1.  We could say that any identifier that can be resolved before macro
+    application keeps its original resolution (so `x` would still resolve to the
+    original, top level `x`).
+2.  We could re-resolve all identifiers after the macros are applied, which can
+    possibly change what they resolve to (in this case `x` would resolve to the
+    generated instance getter `x`).
+3.  We could make it some kind of error for a macro to introduce an identifier
+    that shadows another.
+4.  We could make it error to use an identifier shadowed by one produced by a
```
munificent, line 372: "it error" -> "a compile-time error"

Maybe emphasize "use" (`*use*`) to make it clearer how this differs from the previous point.

working/macros/feature-specification.md
```diff
@@ -304,20 +304,109 @@ class, allowing you to build up your code fragments however you like.
...
+    macro.
+
+The first two choices could be very confusing to users, some will expect one
+behavior while others expect the other. The third choice would work but might be
+overly restrictive. The final option still avoids the ambiguity, and is a bit
+more permissive than the third.
+
+It similarly also is not allowed for one macro to produce an declaration that
```
munificent, line 380: "an declaration" -> "a declaration"

working/macros/feature-specification.md
```diff
@@ -304,20 +304,109 @@ class, allowing you to build up your code fragments however you like.
...
+the identifier resolves to and then another macro to produce another declaration
+that then shadows that one. In other words, any hand-authored identifier may be
+resolved at any point during macro application, but it may only be resolved
+once.
+
+At the same time, a macro application in a library must be able to shadow
+identifiers in other libraries, in order to allow for modular compilation. It
```
munificent, line 387: I'm not sure what this means, I don't know if "modular compilation" (an implementation technique) is the right thing to focus on. Are you trying to say that a macro application can produce a declaration that shadows an imported one?

working/macros/feature-specification.md
```diff
@@ -304,20 +304,109 @@ class, allowing you to build up your code fragments however you like.
...
+shouldn't be the case that a library can cause an error in one of the libraries
+that it imports.
+
+These constraints produce this rule:
+
+*   It is a compile-time error if any hand-authored identifier in a library
+    containing a macro application would bind to a different declaration when
+    resolved before and after macro application. In other words, it is a
+    compile-time error if a macro introduces an identifier that shadows a
+    hand-authored identifier that is used in the same library.
```
munificent, line 397: What about:

```dart
@generateXTopLevelVariable
library foo;

@generateXGetter
class C {
  int get y => x;
}
```

Is this an error? If `@generateXGetter` were at some point in the future be able to introspect over the body of `y`, what would it see that `x` has been resolved to, if anything?

I think the answer is probably that this is allowed—multiple macro applications can produce identifiers that *do* shadow each other. The user's expectation is only how identifiers in the actual text file are resolved and how identifiers after all macro applications are done are resolved. In both cases, the behavior is well-defined. It might be worth calling this out explicitly.

In the example, introspection should see it as an unresolved identifier.

munificent (approved):
A few more things, but I really like how this is looking.

working/macros/feature-specification.md
jakemac53: For this PR, do you think I should land it as is or replace the details with an TODO?

working/macros/feature-specification.md
jakemac53: No, I don't want to allow that :).

I wasn't sure how the resolution worked with overrides but `this.x` makes sense and would work fine so I will remove the exception.

working/macros/feature-specification.md
jakemac53: There are, although we don't need a special class for them (you could just use `Expression` I think).

jakemac53: I am fine with just calling this Identifier

working/macros/feature-specification.md
jakemac53: haha, ok. That seems super weird to me but 🤷‍♂️ .

working/macros/feature-specification.md
jakemac53: What I am trying to say is that if I import a library, macros in that imported library *are* allowed to change the resolution of identifiers in my library. Otherwise my library can cause a compile time error in a library I import, which is bad. And in general that sounds like a very expensive thing to check for.

Case in point:

```dart
import 'a.dart'; // has a macro generated top level getter `x`

class A {
  int get x => 2;
}

class B {
  // Which `x`? Its actually the one that comes from `a.dart`, because top level getters
  // take precedence over super getters.
  int get y => x; 
}

void main() {
  print(B().y); // 1
}
```

I don't know if there are more situations than just this that exist, this is just kind of a weird quirk, but it does exist.

working/macros/feature-specification.md
jakemac53: Ya I have the same reasoning for calling out that its only hand-authored things that can't ever have ambiguity. The user will probably only ever see macro code in its pre-expanded or fully expanded states, and not any potentially confusing in-between states.

I don't think we have to worry about introspection because you can't introspect at all on these things until all declarations are created anyways.

working/macros/feature-specification.md
munificent: How about leaving what you have but adding a TODO saying that it needs further thought?

working/macros/feature-specification.md
munificent: Ah, I get it. Yes, I was just taking for granted that all macro applications in imported libraries are done before the import is processed at all. Once you import a library, you can't tell what parts of it came from macros and what didn't.

I think there's probably a better way to express this. Maybe early in the entire section talk about this is how scoping works *within* a single library and explicitly state that macro applications are handled separately for each library. Or maybe that's implicit given the ordering section? Maybe you just need to say "Here is how scoping works within a library."?

working/macros/feature-specification.md
munificent: In the example here, I think that depends on whether we consider a library-level macro like `@generateXTopLevelVariable` to be a parent of the macros on top-level declarations. I think I was assuming here it would be, so that a library-level macro could see all of the declarations in the library.

munificent (approved):

working/macros/feature-specification.md
jakemac53: I added a part above that clarifies the rules only apply within a single library.

working/macros/feature-specification.md
jakemac53: > I think I was assuming here it would be, so that a library-level macro could see all of the declarations in the library.

I think that the same rules still work out - since the introspective capabilities could still be limited to guarantee that. So for instance a Library macro would not be able to do subtype checks on types it sees, because other library type macros might still be messing with types.

jakemac53: We can follow up and discuss this further

PR #1782, opened 2021-08-06
Update section Symbols, resolving several issues

eernstg:
Section Symbols in the language specification had several issues, arising over time and also because of the recent change to allow `Symbol(s)` where `s` is an arbitrary string.

- The section used to specify that certain symbols are _not_ the same object (not `identical`), which is something we otherwise do not specify (because we do not want to preclude compile-time or run-time canonicalization of symbols, where possible, as an optimization).
- The section did not mention canonicalization or equality of symbols obtained from `Symbol(_)`.

This PR adjusts the text such that these issues are resolved, and brings the language specification in sync with the newly written tests about symbol equality/identity.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -29,6 +29,7 @@
 % 2.14
 % - Add constraint on type of parameter which is covariant-by-declaration in
 %   the case where the method is inherited (that case was omitted by mistake).
+% - Clarify symbol equality and identity.
 %
 % 2.12 - 2.13 (there was no 2.11)
 % - Revert the CL where certain null safety features were removed (to enable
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -9168,33 +9169,56 @@ \subsection{Symbols}
...
+\item $t$ is a period separated sequence of identifiers
+  \code{$id_1$.$id_2$.$\cdots$.$id_n$}.
+\item $t$ is derived from \synt{operator}.
+\end{itemize}
+
+\LMHash{}%
+Any two occurrences of \code{\#t} evaluate to the same object,
+and so does a symbol obtained as the value of a constant expression
+of the form \code{Symbol($s$)} or \code{\CONST{} Symbol($s$)},
+where $s$ is a constant string whose contents is $t$.
```
lrhn, line 9187: Would:
 ... a constant expression evaluating to a string ...
be more precise, or is this an imprecision we're fine with otherwise?

Also, the contents of a string is a sequence of code units, it's not a term, so maybe "whose contents is the same characters as $t$". (Very nit-picky).

specification/dartLangSpec.tex
```diff
@@ -9168,33 +9169,56 @@ \subsection{Symbols}
...
 (\ref{equality}).
 
 \LMHash{}%
 A symbol literal \code{\#\_\id} where \id{} is an identifier
 evaluates to an instance of \code{Symbol}
-representing the private identifier \code{\_\id} of the containing library.
+representing the private identifier \code{\_\id} of the enclosing library.
 All occurrences of \code{\#\_\id} \emph{in the same library} evaluate to
 the same object,
 and no other symbol literals evaluate to the same object,
```
lrhn, line 9218: -> "and no other symbol literal or \code{Symbol} constructor invocation evaluates to that object,"

specification/dartLangSpec.tex
```diff
@@ -9168,33 +9169,56 @@ \subsection{Symbols}
...
+\LMHash{}%
+In these cases, we say that the given symbol is a
+\IndexCustom{non-private symbol based on}{symbol!non-private, based on}
+the string $s$.
+If $o$ is an instance of \code{Symbol} obtained by
+a non-constant evaluation of an expression
+of the form \code{Symbol($s'$)} or the form \code{new Symbol($s'$)}
+then we also say that the given symbol is a
+\NoIndex{non-private symbol based on}
+the string $s'$.
```
lrhn, line 9199: What is `$s'$`? Should the last line be "the string that $s'$ evaluate to." (since $s'$ is used as an expression above and seemingly a value here).

specification/dartLangSpec.tex
eernstg: 'a constant expression evaluating to a string' is surely better. I used the nit, too. ;)

specification/dartLangSpec.tex
eernstg: Right, I changed `$s'$` to `$e$` everywhere (it's just an expression, and static or dynamic checks will enforce that it is a `String`, or we wouldn't have obtained `$o$`). I also changed the end of the sentence to explicitly say that the relevant string is the value of `$e$` at the invocation of the constructor.

specification/dartLangSpec.tex
eernstg: Indeed, done!

eernstg (commented):
Thanks, PTAL.

specification/dartLangSpec.tex
```diff
@@ -9168,33 +9169,58 @@ \subsection{Symbols}
...
+a string whose contents is the same characters as $t$.
+\commentary{That is, symbol instances are canonicalized.}
+
+\LMHash{}%
+In these cases, we say that the given symbol is a
+\IndexCustom{non-private symbol based on}{symbol!non-private, based on}
+the string $s$.
+If $o$ is an instance of \code{Symbol} obtained by
+a non-constant evaluation of an expression
+of the form \code{Symbol($e$)} or the form \code{new Symbol($e$)}
```
lrhn, line 9197: I think Github lost my comment here ...
I see no case covering `const Symbol("_x")`. This case only does non-const, and the one above only does non-private identifiers.

Consider:
```latex
An invocation of the \code{Symbol} constructor, whether \code{Symbol($e$)}, \code{new Symbol($e$)} or \code{const Symbol($e$)}, where $e$ is an expression (constant if necessary) which evaluates to a string $s$, creates a non-private symbol based on $s$. All constant symbol instance creations which creates a non-private symbol based on the same string $s$, evaluate to the same such object.
\commentary{That is, constant symbols are canonicalized.}
```

Maybe move that before dealing with literals, so the sequence is "symbol constructor", "non-private literals", "private literals" instead of "non-private literals", "constructors", "private literals". (Just a hunch that it might read better).

lrhn (approved):

specification/dartLangSpec.tex
eernstg: Good catch! Revised the section based on this.

specification/dartLangSpec.tex
```diff
@@ -9168,33 +9169,52 @@ \subsection{Symbols}
...
-or that $t$ is derived from \synt{operator}.
-Any two occurrences of \code{\#t} evaluate to the same object
-(\commentary{that is, symbol instances are canonicalized}),
-and no other symbol literals evaluate to that object,
-nor to a \code{Symbol} instance that is equal to that object
-according to the \lit{==} operator
+For the value $o$ of a symbol literal representing a source code term
+like an identifier or an operator, we say that $o$ is a
+\IndexCustom{non-private symbol based on}{symbol!non-private, based on}
+the string whose contents is the characters of that term, without whitespace.
```
lrhn, line 9175: Unless it's a private identifier?

specification/dartLangSpec.tex
```diff
@@ -9168,33 +9169,52 @@ \subsection{Symbols}
...
+\LMHash{}%
+If $o_1$ and $o_2$ are non-private symbols
+(\commentary{not necessarily constant})
+based on strings $s_1$ and $s_2$
+then $o_1$ and $o_2$ are equal according to operator \lit{==}
+if and only if \code{$s_1$ == $s_2$}
 (\ref{equality}).
 
 \LMHash{}%
 A symbol literal \code{\#\_\id} where \id{} is an identifier
```
lrhn, line 9208: Actually, it should be "where \_\id{} is an identifier" because it covers `#_9` as well.

specification/dartLangSpec.tex
eernstg: The intention is that 'representing' (or 'represents', etc.) is a reference to the 4 occurrences of that word in the previous ~20 lines of text. However, I can see that it is ambiguous, so we should adjust it.

Cf. https://github.com/dart-lang/language/issues/1783.

specification/dartLangSpec.tex
eernstg: Cf. https://github.com/dart-lang/language/issues/1783 as well.

PR #1805, opened 2021-08-13
Add `is` and `as` disambiguation tokens to constructor tearoffs proposal.

stereotype441:
See https://github.com/dart-lang/sdk/issues/46888 for the rationale.

### DIFF
--- a/accepted/future-releases/constructor-tearoffs/feature-specification.md
+++ b/accepted/future-releases/constructor-tearoffs/feature-specification.md
@@ -1,6 +1,6 @@
 # Dart Constructor Tear-offs
 
-Author: lrn@google.com<br>Version: 2.12
+Author: lrn@google.com<br>Version: 2.13
 
 Dart allows you to tear off (aka. closurize) methods instead of just calling them. It does not allow you to tear off *constructors*, even though they are just as callable as methods (and for factory methods, the distinction is mainly philosophical).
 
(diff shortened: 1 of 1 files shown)


accepted/future-releases/constructor-tearoffs/feature-specification.md
```diff
@@ -360,14 +360,16 @@ The look-ahead tokens which force the prior tokens to be type arguments are:
 
 > `(`  `)`  `]`  `}`  `:`  `;`  `,`  `.`  `?`  `==`  `!=` `..` `?.` `??` `?..` 
 >
-> `&` `|` `^` `+` `*`  `%`  `/`  `~/`
+> `&` `|` `^` `+` `*`  `%`  `/`  `~/` `is` `as`
```
eernstg, line 363: This whole list is somewhat dangerous: We're basically deciding that we will _never_ have an expression that starts with `:`, `.`, `&`, `|`, `^`, `+` (note that Kotlin uses unary plus a lot), `*`, `%`, and now `as` (which is already possible), or at least that those expressions will be 2nd class citizens because they must be enclosed in parentheses when they occur as the right operand in a greater-than expression.

Maybe we should think about this list one more time: It could be small (and sufficient, in some sense) rather than maximal (in terms of which tokens we _currently_ know can't be the beginning of an expression).

How much should the rest of the language pay for the ability to write `f<int, String>`?

eernstg (commented):
I think the change is fine on its own, but we should probably think about the premise for having a long list of disambiguation tokens in the first place. It's now! -- if we want to make that list shorter.

accepted/future-releases/constructor-tearoffs/feature-specification.md
stereotype441: I think that's a fair question and deserves discussion.  I've opened https://github.com/dart-lang/language/issues/1806 so we can discuss it further.

In the meantime, though, I'd like to move ahead with this PR, since I've already made the corresponding code change (based on what appeared to be a consensus on https://github.com/dart-lang/sdk/issues/46888 that it was a good idea), and I'd like to keep the spec in line with the implementation as much as possible.

Is that ok with you?  I definitely don't want to stifle discussion on the topic.

eernstg (approved):

accepted/future-releases/constructor-tearoffs/feature-specification.md
eernstg: Sure! It is also useful to get to know how it works, no matter how #1806 ends.

PR #1812, opened 2021-08-20
Allow generic function instantiation of an existing function object

eernstg:
Update section "Generic Function Instantiation" to allow this operation on an existing function object, in addition to the currently supported cases (where the expression denotes a declaration).

This update does _not_ yet mention explicit instantiation, but I think the work needed to allow the current generalization and the work needed to support explicit instantiation are rather well separated. So it's not a waste of time to those two things in two steps (and we can then do 'constructor-tearoffs' as one single update, which is probably also nice).

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -26,6 +26,10 @@
 %
 % Significant changes to the specification.
 %
+% 2.15
+% - Allow generic instantiation of expressions with a generic function type
+%   (until now, it was an error unless the expression denoted a declaration).
+%
 % 2.14
 % - Add constraint on type of parameter which is covariant-by-declaration in
 %   the case where the method is inherited (that case was omitted by mistake).
@@ -12925,24 +12929,17 @@ are canonicalized.%
 
 \LMHash{}%
 Generic function instantiation is a mechanism that yields
-a non-generic function object,
-based on a reference to a generic function.
-
-\commentary{%
-It is a mechanism which is very similar to function closurization
-(\ref{functionClosurization}),
-but it only occurs in situations where
-a compile-time error would otherwise occur.%
-}
+a non-generic function object based on a given generic function.
 
 \rationale{%
 The essence of generic function instantiation
 is to allow for ``curried'' invocations,
 in the sense that a generic function can receive its actual
-type arguments separately during closurization
+type arguments separately
 (it must then receive \emph{all} type arguments, not just some of them),
 and that yields a non-generic function object.
 The type arguments are passed implicitly, based on type inference;
+%% TODO(eernst): Come constructor-tearoffs, revise this.
 a future version of Dart may allow for passing them explicitly.%
 }
 \commentary{Here is an example:}
@@ -12970,16 +12967,19 @@ to the corresponding generic function.%
 }
 
 \LMHash{}%
-Let $f$ of the form
+Let \DefineSymbol{f} of the form
 \syntax{<identifier> (`.'~<identifier>\,(`.'~<identifier>)?)?}~be
-an expression that denotes
-a declaration of a local function, a static method, or a top-level function,
-and let $G$ be the static type of $f$.
-Consider the situation where $G$ is a function type of the form
-\RawFunctionType{T_0}{X}{B}{s}{\metavar{parameters}}
-with $s > 0$
-(\commentary{that is, $G$ is a generic function type}),
-and the context type is a non-generic function type $F$.
+an expression whose static type \DefineSymbol{G} is
+\BlindDefineSymbol{T_0,X_j,B_j,s,p}%
+a generic function type of the form
+
+\noindent
+\RawFunctionType{T_0}{X}{B}{s}{\metavar{p}}
+
+\noindent
+where \code{($p$)} is derived from \synt{formalParameterList}.
+\commentary{Note that $s > 0$ because $G$ is generic.}
+Assume that the context type is a non-generic function type $F$.
 In this situation a compile-time error occurs
 (\ref{variables},
 \ref{functions},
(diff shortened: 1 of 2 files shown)


eernstg:
@lrhn, I removed the review request because I'm revising this PR. I'll restore it when the text is in good shape again.

eernstg:
The rules are now much more compositional (identical/== for the whole is determined by identical/== on the parts).

specification/dartLangSpec.tex
```diff
@@ -13013,129 +13009,74 @@ \subsubsection{Generic Function Instantiation}
...
-List<int> fooOfInt(int i) => [i];
-\\
-String bar(List<int> f(int)) => "\${f(42)}";
-\\
-\VOID{} main() \{
-  print(bar(foo));
-\}
-\end{dartCode}
+\LMHash{}%
+Let $f_1$ and $f_2$ be two constant expressions that are subject to
```
lrhn, line 13035: Have we added that construct to the constants section as well:

* `c<typeArgs>` is a compile-time constant when `c` is a constant expression with a static type which is a generic function type, and `<typeArgs>` is a type argument clause where all the types are compile-time constant types.

(Can they be potentially constant expressions if *c* is potentially constant, and `typeArgs` are really constant? I'd say no, it's object creation at compile-time, and we only allow that for basic types, not something compositional with non-trivial equality).

specification/dartLangSpec.tex
```diff
@@ -13013,129 +13009,74 @@ \subsubsection{Generic Function Instantiation}
...
-but denoting the same function,
-respectively the situation where
-an execution of the program containing $f$ evaluates it twice.
-Let $o_1$ and $o_2$ denote the function objects obtained by
-evaluation of those two expressions,
-respectively the two evaluations of that expression.
+Let $g_1$ and $g_2$ be two expressions
+(\commentary{which may or may not be constant})
+that are subject to generic function instantiation.
+Assume that $g_1$ and $g_2$ without a context type evaluate to $o_1$
```
lrhn, line 13056: This assume *implicit* instantiation. Maybe "without a context type and without explicit type arguments".

Consider moving away from talking about context types here, and instead assume that type inference inserts the actual type arguments, then work with those from there on. Just like we (I hope) do for inferred type arguments to classes and in invocations.

specification/dartLangSpec.tex
```diff
@@ -13013,129 +13009,74 @@ \subsubsection{Generic Function Instantiation}
...
+respectively $o_2$ such that \code{$o_1$ == $o_2$)} is true.
+Assume that the given context types give rise to a successful
+generic function type instantiation with
+the same actual type arguments for $g_1$ and $g_2$,
```
lrhn, line 13060: Where do we define "same ... type arguments"?
Refer to that here (if we have it).

eernstg: We do not yet have anything on 'same type' (other than exactly that phrase) in the language specification.

We have equality for reified `Type`s in the null-safety spec, and that's going to be integrated here soon. So I'd like to handle that as part of the integration of null-safety.

For now, I added a few comments to known places where this change will be relevant.

eernstg (commented):
Review response.

eernstg:
Added one change which was omitted by accident at first: The PR now also removes the rule which makes it an error to perform a generic method instantiation on the `call` method of a function object. We only introduced that error because we thought we couldn't do that, but this whole PR is about doing it anyway.

specification/dartLangSpec.tex
```diff
@@ -12970,16 +12970,14 @@ \subsubsection{Generic Function Instantiation}
 }
 
 \LMHash{}%
```
lrhn, line 12972: If *f* is an expression with a callable object type with a generic `call`method, will "type inference" magically have inserted the `.call` at this point? (I guess it will, since the context type is a function type and the expression's static type is otherwise a callable object. Is it worth mentioning?)

specification/dartLangSpec.tex
```diff
@@ -13013,129 +13009,74 @@ \subsubsection{Generic Function Instantiation}
...
-\commentary{Here is an example:}
+Execution of $f$ proceeds as follows.
+Evaluate $f$ to an object $o$.
+Let \RawFunctionType{S_0}{Y}{B'}{s}{$q$} be the dynamic type of $o$
+(\commentary{by soundness, this type is a subtype of $G$}).
+$f$ then evaluates to a function object $o'$ with dynamic type
+$[t_1/Y_1, \ldots, t_s/Y_s](\FunctionTypeSimple{S_0}{$q$})$,
+where $t_j$ is the actual value of $T_j$, for $j \in 1 .. k$.
+An invocation of $o'$ with actual arguments \metavar{args} has
+the same effect as an invocation of $o$
```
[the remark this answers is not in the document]
lrhn: It's probably fine. If the reader is just a little cooperative, it should be clear what it means.

specification/dartLangSpec.tex
lrhn: Ack. Fine then.

lrhn (approved):

specification/dartLangSpec.tex
eernstg: I don't think we have support for that. ['Function closurization'](https://github.com/dart-lang/language/blob/a8b6b66346d975c9d475a5fa8b43b9bf71d3cce4/specification/dartLangSpec.tex#L12853) in the language specification never mentions `.call` methods, so generic function instantiation would need to be applicable to a "callable object type" rather than a generic function type, and we don't have that either.

eernstg (commented):
Review response.

eernstg:
A few experiments with the current implementations show that we _do_ actually allow for the tearoff.

One might think "Sure, why not?", but we specified that it isn't possible, so I created https://github.com/dart-lang/language/issues/1827 to raise the question and perhaps decide that we should just support both the closurization and the generic function instantiation on an `e` that has been desugared to `e.call` (including `e<T>` which has been extended to `e.call<T>`).

PR #1816, opened 2021-08-25
Macros: add section on language and api versioning

jakemac53:
cc @leafpetersen

### DIFF
--- a/working/macros/feature-specification.md
+++ b/working/macros/feature-specification.md
@@ -422,6 +422,43 @@ Note, that if the getter were written as `int get y => this.x;`, then a macro
 *would* be allowed to introduce the new getter `x`, because `this.x` could not
 previously be resolved.
 
+## Language and API Evolution
+
+### Language Versioning
+
+Macros generate code directly into existing libraries, and we want to maintain
+the behavior that a library only has one language version. Thus, the language
+version of macro generated code is always that of the library it is generating
+code _into_.
+
+This means that macros need the ability to ask for the language version of a
+given library. This will be allowed through the library introspection class,
+which is available from the introspection apis on all declarations via a
+`library` getter.
+
+TODO: Fully define the library introspection api for each phase. 
+
+### API Versioning
+
+TODO: Finalize the approach here.
+
+It is possible that future language changes would require a breaking change to
+an existing imperative macro API. For instance you could consider what would
+happen if we added multiple return values from functions. That would
+necessitate a change to many apis so that they would support multiple return
+types instead of a single one.
+
+#### Proposal: Ship Macro APIs as a Pub Package
+
+Likely, this package would only export directly an existing `dart:` uri, but
+it would be able to be versioned like a public package, including tight sdk
+constraints (likely on minor version ranges). This would work similar to the
+`dart:_internal` library.
+
+This approach would involve more work on our end (to release this package with
+each dart release). But it would help keep users on the rails, and give us a
+lot of flexibility with the API going forward.
+
 ## Limitations
 
 - Macros cannot be applied from within the same library cycle as they are


working/macros/feature-specification.md
```diff
@@ -422,6 +422,43 @@ Note, that if the getter were written as `int get y => this.x;`, then a macro
...
+### Language Versioning
+
+Macros generate code directly into existing libraries, and we want to maintain
+the behavior that a library only has one language version. Thus, the language
+version of macro generated code is always that of the library it is generating
+code _into_.
+
+This means that macros need the ability to ask for the language version of a
+given library. This will be allowed through the library introspection class,
+which is available from the introspection apis on all declarations via a
```
munificent, line 436: "apis" -> "APIs". :)

working/macros/feature-specification.md
```diff
@@ -422,6 +422,43 @@ Note, that if the getter were written as `int get y => this.x;`, then a macro
...
+#### Proposal: Ship Macro APIs as a Pub Package
+
+Likely, this package would only export directly an existing `dart:` uri, but
+it would be able to be versioned like a public package, including tight sdk
+constraints (likely on minor version ranges). This would work similar to the
+`dart:_internal` library.
+
+This approach would involve more work on our end (to release this package with
+each dart release). But it would help keep users on the rails, and give us a
+lot of flexibility with the API going forward.
```
munificent, line 460: I don't know if this will buy us much. If every macro ends up having a tight constraint on the package, version lock will become more common and that will be so painful that we'll be highly disincentivized from even changing the pub package.

The analyzer team has been able to evolve their API through a large number of language changes while being mostly backwards compatible, and that's while exposing a complete AST API. The current macro API doesn't give you a complete set of classes for expressions and statements, so the surface area is much smaller. Given that, I think we can probably get away with treating the macro API as just another "dart:" library where we simply have to evolve it very carefully and avoid breaking changes.

If we do need to make a breaking change, we might be able to use language versioning to evolve the macro API too. We did something roughly similar with null safety, where things like `List()` existed when called from legacy libraries but not null safe libraries.

jakemac53: I think that any pain incurred by this shipping as a package is well worth the effort on our end. It buys us a lot in terms of future flexibility, and would even enable us to expose an AST api if so desired. I don't expect a lot of breaking changes in this package, but the ability to make some breaking changes is very valuable and we will thank ourselves later, imo.

jakemac53: Note that the analyzer today actually has fairly consistent breaking changes (not related to AST apis usually). It does cause some pain to get the world updated, but it works itself out. I would expect far fewer breaking changes in this package anyways.

The macro authors also don't need tight constraints, just normal constraints restricting to <= the current major version.

munificent: > I think that any pain incurred by this shipping as a package is well worth the effort on our end. It buys us a lot in terms of future flexibility, and would even enable us to expose an AST api if so desired.

I guess I don't see how it helps. The package API must be implemented in terms of *some* core `dart:` API. The assumption here is that there will be many changes to the package-level code that don't require any corresponding change to the `dart:`-level code (or at least don't require a breaking change to the latter), but I'd want to see a clearer picture about what that core API looks like before I have confidence that it's the case.

(If we do go down the road of having macros compile to something like generated files + partial classes, etc. then I definitely *can* see moving almost all of the macro system implementation into a package with some minimal hooks in the language for invoking it.)

jakemac53: I probably didn't describe it well enough but the `dart:` api would be a private one - and exported by the package. Only that package would be allowed to import it (we have this pattern in at least one other place I think?).

PR #1818, opened 2021-08-25
Macros: add section about adding new macro applications

jakemac53:
This handles the use cases around adding macro applications to both existing and new declarations.

### DIFF
--- a/working/macros/feature-specification.md
+++ b/working/macros/feature-specification.md
@@ -341,7 +341,11 @@ following the normal ordering rules.
 
 Macro applications can be added to existing declarations through the `*Builder`
 APIs. Macros added in this way are always prepended to the list of existing
-macros on the declaration.
+macros on the declaration (which makes them run _last_).
+
+Note that macros can already _immediately_ invoke another macro on a given
+declaration manually, by simply instantiating the macro and then invoking
+it.
 
 TODO: Update the builder apis to allow this.
 


working/macros/feature-specification.md
```diff
@@ -319,6 +319,63 @@ be constructed, but where you need to build up something in smaller pieces you
...
+
+#### Adding Macro Applications to New Declarations
+
+When creating [Code][] instances, a macro may generate code which includes
+macro applications. These macro applications must be from either the current
+phase or a later phase, but cannot be from previous phases.
+
+If a macro application is added which implements an earlier phase, that phase
+is not ran. This should result in a warning if the macro does not also
+implement some phase that will be ran.
```
munificent, line 334: I think it's probably safest to make this an explicit error. I think a macro that defines multiple phases will probably want to be able to reliably assume that its earlier phases have run and that it won't spontaneously pick up in the middle of its phased execution.

working/macros/feature-specification.md
```diff
@@ -319,6 +319,63 @@ be constructed, but where you need to build up something in smaller pieces you
...
+```
+
+Normally, macros always run "inside-out". But in this case `@macroC` runs after
+both `@macroB` and `@macroC` which were applied to the class.
+
+We still allow this because it doesn't cause any ambiguity in ordering, even
+though it violates the normal rules.
+
+We could instead only allow adding macros from _later_ phases, but that would
+restrict the functionality in unnecessary ways.
```
munificent, line 377: Do we have known use cases for macros generating code that contains macro applications? Are there cases where the macro can't simply directly invoke the corresponding macro code to achieve the same effect?

Unless we know this is needed, I'd be very inclined to leave it out because, if nothing else, it's pretty confusing for users to reason about.

munificent (commented):
👍

working/macros/feature-specification.md
jakemac53: It am mostly thinking of a macro applying _itself_ to other declarations - granted it could split itself into multiple macros just to avoid the error, but that seems less than ideal.

Consider a macro that runs on a state class, generates a stateful widget class, and also applies itself to some lifecycle methods in the current class. It would likely implement _all three_ phases of macros, but would apply itself to some declarations in the declaration phase. That wouldn't be allowed now and it would have to split itself into multiple macros. Granted macros are easily composable and so it wouldn't be that arduous to do, but it also feels unnecessary.

working/macros/feature-specification.md
jakemac53: Yes we have examples - namely this comes up when you want to hook into lifecycle methods whether the user implements them already or not. If the method exists you want to apply yourself to it, and if it doesn't you want to generate a version of it with yourself applied.

working/macros/feature-specification.md
munificent: Ah, that's an interesting example, thanks.

PR #1826, opened 2021-08-30
Add a FAQ for the constructor tear-offs feature

lrhn:
Attempted to do a less-formal FAQ-style introduction to the constructor tear-off feature.
I've seen a lot of questions which suggest that the specification is not accessible enough to casual readers.

### DIFF
--- a/accepted/future-releases/constructor-tearoffs/faq.md
+++ b/accepted/future-releases/constructor-tearoffs/faq.md
@@ -1,5 +1,7 @@
 # Dart Constructor Tearoffs Feature FAQ
 
+Author: lrn@google.com<br>Version: 1.0
+
 This is a short summary of the _Constructor Tearoffs_ feature. This document is not intended as a specification, look at the feature specification for that. Instead it hopes to be a brief *introduction* to the feature set that we intend to release, and to answer some of the questions that it is hard to find short answers for in the specification.
 
 ## What are the new features?
@@ -9,7 +11,7 @@ In short:
 * Named constructor tear-off (`C.name` is a valid expression).
 * Named unnamed constructor (`C.new` is a constructor name, refers to the same constructor as "unnamed" `C` constructor).
 * Function value instantiation (you can instantiate function *values*, not just tear-offs).
-* Explicit instantiation (`List<int>` and `Future.then<int>` are valid type- and function-expressions.)
+* Explicit instantiation (`List<int>` and `Future.then<int>` are valid type- and function-expressions).
 
 ## Named constructor tear-off
 
@@ -125,11 +127,11 @@ Everywhere except the tear-offs, you can remove the `.new`  and it means the sam
 
 **Q:** Will `dart:mirrors` be able to see the `.new` on a constructor declaration?
 
-**A:** Most likely not. There are no plans to change the `dart:mirrors`, and it's just different syntax for the same declaration. The constructor name is still going to be just the class name, and that's what `dart:mirrors` expose.
+**A:** Most likely not. There are no plans to change the `dart:mirrors` library, and it's just different syntax for the same declaration. The constructor name is still going to be just the class name, and that's what `dart:mirrors` expose.
 
 ## Function value instantiation
 
-Until now you've could instantiate *tear-offs* of function declarations or instance methods. You could write:
+Until now you could only instantiate *tear-offs* of function declarations or instance methods. You could write:
 
 ```dart
 T id<T>(T value) => value;
@@ -143,9 +145,11 @@ T Function<T>(T) id = <T>(T value) => value;
 int Function(int) intId = id; // INVALID
 ```
 
-There were reasons for this, mainly worries about implementations not being able to be efficient. The implementors have told us that it's not a problem, so we remove that restriction and allow you to instantiate any function-typed expression. The `INVALID` above becomes valid and well-typed.
+There were reasons for this, mainly worries about implementations not being able to be efficient. The implementors have told us that it's not a problem, so we now remove that restriction and allow you to instantiate any function-typed expression. The `INVALID` above becomes valid and well-typed.
 
-This also applies to *callable objects* (objects which has an interface type with a `call` method), which we treat like function values in most places.
+_This is not actually part of the constructor tear-off feature, it's considered a bug-fix because some (but not all) compilers already supported the feature, and it will be available in older language versions too. See language issue [#1812](https://github.com/dart-lang/language/pull/1812) for details. This feature-fix will still very likely ship at the same time as constructor tear-offs, and it's affecting the explicit instantiation feature, so we include it here._
+
+This also applies to *callable objects* (objects that have an interface type with a `call` method), which we treat like function values in most places.
 
 **Q:** Where does that even matter?
 
@@ -159,7 +163,7 @@ Hypothetical example:
 
 **Q:** Couldn't I just instantiate the `call` method using method instantiation anyway?
 
-**A:** Yes and no. You could for callable objects, and we'd even add the `.call` implicitly for you (we do that for any callable object in a function-typed context, before checking whether the types actually work). It just didn't work for *real* function values. Dart2js never implemented tearing off the `call` method because it would be equivalent to instantiating the function value itself, which wasn't a supported feature (until now), so your code would just crash. We recognized this and initially planned to disallow doing a instantiated tear-off of a function's `call` method. Instead it turned out we can just support consistently.
+**A:** Yes and no. You could for callable objects, and we'd even add the `.call` implicitly for you (we do that for any callable object in a function-typed context, before checking whether the types actually work). It just didn't work for *real* function values. Dart2js never implemented instantiated tear-off of the `call` method of function values because that would be equivalent to instantiating the function value itself, which wasn't a supported feature (until now), so your code would just crash. We recognized this and initially planned to disallow doing a instantiated tear-off of a function's `call` method. Instead it turned out we can just support it consistently.
 
 **Q:** Are function value instantiations canonicalized? Or equal?
 
@@ -192,7 +196,7 @@ Whenever such a type argument list is followed by an argument list, it exactly m
 
 **Q:** Doesn't that make the grammar, like, totally ambiguous?
 
-**A:** Yes! *Thank you* for noticing! And that is a problem. With Dart 2.0 we introduced generic function invocations, and had to decide how to parse `f(a<b,c>(d))`. The argument(s) to `f` could be either two comparison operator expressions or a single generic function invocation. We decided on always choosing the latter when `b` and `c` can be parsed as types and the `>` is followed by a `(`. (We have to decide while we parse the program, long before we can even begin to figure out what `b` and `c` actually refer to, so the choice is entirely grammar based). We now have even more similar ambiguous cases. For example `f(a<b,c>-d)` is ambiguous because `-` can both be a prefix operator after a greater-than operator, or an infix operator after an explicit type-instantiation. Our choice is to be very restrictive in when we parse `expr <` as starting a type argument. We only do so the following *can* be parsed as a type argument list, and the only if the *next token* after the final `>` of the type arguments is one of:
+**A:** Yes! *Thank you* for noticing! And that is a problem. With Dart 2.0 we introduced generic function invocations, and had to decide how to parse `f(a<b,c>(d))`. The argument(s) to `f` could be either two comparison operator expressions or a single generic function invocation. We decided on always choosing the latter when `b` and `c` can be parsed as types and the `>` is followed by a `(`. (We have to decide while we parse the program, long before we can even begin to figure out what `b` and `c` actually refer to, so the choice is entirely grammar based.) We now have even more similar ambiguous cases. For example `f(a<b,c>-d)` is ambiguous because `-` can both be a prefix operator after a greater-than operator, or an infix operator after an explicit type-instantiation. Our choice is to be very restrictive in when we parse `expr <` as starting a type argument. We only do so when the following tokens *can* be parsed as a type argument list, and the only if the *next token* after the final `>` of the type arguments is one of:
 
 > `)`, `}`, `]`, `;`, `:`, `,`,`(`, `.`, `==`, or `!=`
 
@@ -200,7 +204,7 @@ If the next token is *any other token*, then the `<` is parsed as a less-than op
 
 **Q:** Can I call a static method from `List` on `List<int>`.
 
-**A:** No. You can write `List.copyRange` but not `List<int>.copyRange`. This is a grammar based restriction. Even if you have a type alias like `typedef Stupid<X> = int;`, where `Stupid<void>` just means `int`, you can't do `Stupid<void>.parseInt`. The only thing you can access trough an instantiated type literal is constructors. (So `Stupid<void>.fromEnvironment` is valid, just please don't do it.)
+**A:** No. You can write `List.copyRange` but not `List<int>.copyRange`. This is a grammar based restriction. Even if you have a type alias like `typedef Stupid<X> = int;`, where `Stupid<void>` just means `int`, you can't do `Stupid<void>.parseInt`. The only thing you can access through an instantiated type literal is constructors. (So `Stupid<void>.fromEnvironment` is valid, just please don't do it.)
 
 **Q:** What if I want to call an extension method defined on `Type` on a `List<int>`?
 
@@ -246,3 +250,7 @@ object..someFuture.then<int>.doWith((f) => f((_) => 42));
 **Q:** Are explicitly instantiated functions and types canonicalized? Are they equal?
 
 **A:** Explicitly instantiated tear-offs work exactly like implicitly instantiated tear-offs, they just don't need to infer the types from the context first. For instantiated type literals, they will be constant and canonicalized if the type arguments are constant (contains no type variables), and otherwise equal if the same type is instantiated with "the same" type arguments.
+
+## Versions
+
+1.0: Initial version
--- a/accepted/future-releases/constructor-tearoffs/faq.md
+++ b/accepted/future-releases/constructor-tearoffs/faq.md
@@ -1,0 +1,248 @@
+# Dart Constructor Tearoffs Feature FAQ
+
+This is a short summary of the _Constructor Tearoffs_ feature. This document is not intended as a specification, look at the feature specification for that. Instead it hopes to be a brief *introduction* to the feature set that we intend to release, and to answer some of the questions that it is hard to find short answers for in the specification.
+
+## What are the new features?
+
+In short:
+
+* Named constructor tear-off (`C.name` is a valid expression).
+* Named unnamed constructor (`C.new` is a constructor name, refers to the same constructor as "unnamed" `C` constructor).
+* Function value instantiation (you can instantiate function *values*, not just tear-offs).
+* Explicit instantiation (`List<int>` and `Future.then<int>` are valid type- and function-expressions.)
+
+## Named constructor tear-off
+
+If *C* refers to a class, and *C* has a constructor *C*.*name*, then <code>*C*.*name*</code> can now be an expression which evaluates to a function with the same function signature as the constructor, and which, when called, does the same thing as the constructor.
+
+Example:
+
+```dart
+// Has type: DateTime Function(int, [int, int, int, int, int, int, int])
+var makeUtcDate = DateTime.utc; 
+```
+
+**Q:** What if the class is generic?
+
+**A:** Then the function is also generic, with the same type arguments as the class.
+
+Example:
+
+```dart
+// Has type: List<T> Function<T>(int, T)
+var makeList = List.filled;
+// Has type: Map<K, V> Function<K, V>(Iterable<MapEntry<K, V>>)
+var makeMap = Map.fromEntries;
+```
+
+**Q:** Can I tear it off at a specific type argument?
+
+**A:** Yes. Just like the current function tear-offs, the context type can be used to specify a type argument. It's really like the constructor tear-off is tearing off a generic function.
+
+Example:
+
+```dart
+List<String> Function(int, String) makeList = List.filled; // Works!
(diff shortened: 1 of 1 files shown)


accepted/future-releases/constructor-tearoffs/faq.md
```diff
@@ -0,0 +1,248 @@
...
+
+**A:** No. The *name* of the constructor is still `C`, the `C.new` is just another syntax for declaring a constructor named `C`, and you still can't declare two constructors with the same name.
+
+**Q:** Will `dart:mirrors` be able to see the `.new` on a constructor declaration?
+
+**A:** Most likely not. There are no plans to change the `dart:mirrors`, and it's just different syntax for the same declaration. The constructor name is still going to be just the class name, and that's what `dart:mirrors` expose.
+
+## Function value instantiation
+
+Until now you've could instantiate *tear-offs* of function declarations or instance methods. You could write:
```
bwilkerson, line 132: "you've could" --> "you could"?

accepted/future-releases/constructor-tearoffs/faq.md
```diff
@@ -0,0 +1,248 @@
...
+
+Hypothetical example:
+
+```dart
+// Not a clue, mate. You tell me.
+```
+
+**Q:** Couldn't I just instantiate the `call` method using method instantiation anyway?
+
+**A:** Yes and no. You could for callable objects, and we'd even add the `.call` implicitly for you (we do that for any callable object in a function-typed context, before checking whether the types actually work). It just didn't work for *real* function values. Dart2js never implemented tearing off the `call` method because it would be equivalent to instantiating the function value itself, which wasn't a supported feature (until now), so your code would just crash. We recognized this and initially planned to disallow doing a instantiated tear-off of a function's `call` method. Instead it turned out we can just support consistently.
```
bwilkerson, line 162: "just support consistently" --> "just support it consistently"?

accepted/future-releases/constructor-tearoffs/faq.md
```diff
@@ -0,0 +1,248 @@
...
+
+**Q:** Doesn't that make the grammar, like, totally ambiguous?
+
+**A:** Yes! *Thank you* for noticing! And that is a problem. With Dart 2.0 we introduced generic function invocations, and had to decide how to parse `f(a<b,c>(d))`. The argument(s) to `f` could be either two comparison operator expressions or a single generic function invocation. We decided on always choosing the latter when `b` and `c` can be parsed as types and the `>` is followed by a `(`. (We have to decide while we parse the program, long before we can even begin to figure out what `b` and `c` actually refer to, so the choice is entirely grammar based). We now have even more similar ambiguous cases. For example `f(a<b,c>-d)` is ambiguous because `-` can both be a prefix operator after a greater-than operator, or an infix operator after an explicit type-instantiation. Our choice is to be very restrictive in when we parse `expr <` as starting a type argument. We only do so the following *can* be parsed as a type argument list, and the only if the *next token* after the final `>` of the type arguments is one of:
```
bwilkerson, line 195: "We only do so the" --> "We only do so if the"?

accepted/future-releases/constructor-tearoffs/faq.md
```diff
@@ -0,0 +1,248 @@
...
+This is a short summary of the _Constructor Tearoffs_ feature. This document is not intended as a specification, look at the feature specification for that. Instead it hopes to be a brief *introduction* to the feature set that we intend to release, and to answer some of the questions that it is hard to find short answers for in the specification.
+
+## What are the new features?
+
+In short:
+
+* Named constructor tear-off (`C.name` is a valid expression).
+* Named unnamed constructor (`C.new` is a constructor name, refers to the same constructor as "unnamed" `C` constructor).
+* Function value instantiation (you can instantiate function *values*, not just tear-offs).
+* Explicit instantiation (`List<int>` and `Future.then<int>` are valid type- and function-expressions.)
```
pq, line 12: => 

function-expressions).

accepted/future-releases/constructor-tearoffs/faq.md
```diff
@@ -0,0 +1,248 @@
...
+
+**A:** For consistency *and* because it's expected to be useful for other things too, like generic constructors where we'll also need a way to add type arguments to both the class and the constructor. We'd rather introduce a full feature once than a partial feature now and then having to add another part to it later.
+
+**Q:** Can I declare both `C` and `C.new` in the same class?
+
+**A:** No. The *name* of the constructor is still `C`, the `C.new` is just another syntax for declaring a constructor named `C`, and you still can't declare two constructors with the same name.
+
+**Q:** Will `dart:mirrors` be able to see the `.new` on a constructor declaration?
+
+**A:** Most likely not. There are no plans to change the `dart:mirrors`, and it's just different syntax for the same declaration. The constructor name is still going to be just the class name, and that's what `dart:mirrors` expose.
```
pq, line 128: to change the `dart:mirrors`

=>

to change `dart:mirrors`

?

accepted/future-releases/constructor-tearoffs/faq.md
```diff
@@ -0,0 +1,248 @@
...
+
+but you couldn't instantiate function *values* with a type argument:
+
+```dart
+T Function<T>(T) id = <T>(T value) => value;
+int Function(int) intId = id; // INVALID
+```
+
+There were reasons for this, mainly worries about implementations not being able to be efficient. The implementors have told us that it's not a problem, so we remove that restriction and allow you to instantiate any function-typed expression. The `INVALID` above becomes valid and well-typed.
+
```
pq, line 147: remove

=>

removed

accepted/future-releases/constructor-tearoffs/faq.md
```diff
@@ -0,0 +1,248 @@
...
+This also applies to *callable objects* (objects which has an interface type with a `call` method), which we treat like function values in most places.
```
pq, line 148: objects which has an interface type

=>

objects that have an interface type

accepted/future-releases/constructor-tearoffs/faq.md
pq, line 195: (We have to decide while we parse the program, long before we can even begin to figure out what `b` and `c` actually refer to, so the choice is entirely grammar based). 

=>

(We have to decide while we parse the program, long before we can even begin to figure out what `b` and `c` actually refer to, so the choice is entirely grammar based.)

accepted/future-releases/constructor-tearoffs/faq.md
```diff
@@ -0,0 +1,248 @@
...
+
+> `)`, `}`, `]`, `;`, `:`, `,`,`(`, `.`, `==`, or `!=`
+
+If the next token is *any other token*, then the `<` is parsed as a less-than operator.
+
+**Q:** Can I call a static method from `List` on `List<int>`.
+
+**A:** No. You can write `List.copyRange` but not `List<int>.copyRange`. This is a grammar based restriction. Even if you have a type alias like `typedef Stupid<X> = int;`, where `Stupid<void>` just means `int`, you can't do `Stupid<void>.parseInt`. The only thing you can access trough an instantiated type literal is constructors. (So `Stupid<void>.fromEnvironment` is valid, just please don't do it.)
```
pq, line 203: trough

=>

through

pq (commented):
Amazingly useful.  Thank you! 💎 

A grammar few nits...

accepted/future-releases/constructor-tearoffs/faq.md
lrhn: Actually intended as present tense, matching the later "allow"

accepted/future-releases/constructor-tearoffs/faq.md
lrhn: Took me a while to notice the difference between those two sentences :)

accepted/future-releases/constructor-tearoffs/faq.md
```diff
@@ -0,0 +1,256 @@
...
+  var tearoff = C.new; // New!
+  var explicitlyTypedTearoff = C<int>.new; // New!
+}
+```
+
+Everywhere except the tear-offs, you can remove the `.new`  and it means the same thing. The tearoff is the only place which *requires* the `.new`.
+
+**Q:** Should I use `.new` or not. What does the style guide say?
+
+**A:** The style guide says nothing yet. The general rule is to not write something which isn't necessary, so don't write `new` unless you need to. For now, only use it for tear-offs and possibly DartDoc (it will be more Dart-like to write `[C.new]` than the current DartDoc-only `[C.C]` to refer to the unnamed constructor). _Don't ever write `new Foo.new()`._
```
stereotype441, line 118: Personally I'm not a fan of this.  The reason (as I understand it) that we're not making any recommendations in the style guide yet is that we don't want to unduly influence people in one direction or another during this early time period where Dart coders are experimenting and figuring out what kind of style works best.  By making a specific recommendation here, we're subverting that because (hopefully) a lot of people are going to read this FAQ and follow its recommendations, and then when we later look to see what kind of style the community adopted, we'll just see that they adopted our recommendation.

I would rather that we said something like:

> **A:** The style guide says nothing yet.  As with most language features, we are deliberately avoiding making any style recommendations early on so that we can see what kind of a style the community as a whole prefers, and the consenquences of people's stylistic choices.  Once the feature has been in use for a while we'll settle on some specific stylistic recommendations.

accepted/future-releases/constructor-tearoffs/faq.md
```diff
@@ -0,0 +1,256 @@
...
+
+**A:** Yes! *Thank you* for noticing! And that is a problem. With Dart 2.0 we introduced generic function invocations, and had to decide how to parse `f(a<b,c>(d))`. The argument(s) to `f` could be either two comparison operator expressions or a single generic function invocation. We decided on always choosing the latter when `b` and `c` can be parsed as types and the `>` is followed by a `(`. (We have to decide while we parse the program, long before we can even begin to figure out what `b` and `c` actually refer to, so the choice is entirely grammar based.) We now have even more similar ambiguous cases. For example `f(a<b,c>-d)` is ambiguous because `-` can both be a prefix operator after a greater-than operator, or an infix operator after an explicit type-instantiation. Our choice is to be very restrictive in when we parse `expr <` as starting a type argument. We only do so when the following tokens *can* be parsed as a type argument list, and the only if the *next token* after the final `>` of the type arguments is one of:
+
+> `)`, `}`, `]`, `;`, `:`, `,`,`(`, `.`, `==`, or `!=`
+
+If the next token is *any other token*, then the `<` is parsed as a less-than operator.
```
stereotype441, line 203: In case people are worried they're going to have to memorize this rule, it might be worth adding a final sentence saying something like:

> In practice, we believe the parser will just do the right thing without any effort on your part.  In the unlikely event that it doesn't, you can force the interpretation you want by adding parentheses.

accepted/future-releases/constructor-tearoffs/faq.md
```diff
@@ -0,0 +1,256 @@
...
+
+**Q:** If I can instantiate a generic function, and `List.filled` is a generic function tear-off, why is `List.filled<int>` invalid?
+
+**A:** Because we say so. We could make it mean `(List.filled)<int>`, but we prefer to reserve the syntax for if/when we add proper generic constructors. And you *can* write the parentheses, just please don't. Prefer `List<int>.filled` if that's what you mean&mdash;and if it's not what you mean, you probably can't yet do what you mean.
+
+**Q:** So where can I write an explicit type argument instantiation.
+
+**A:** In short, after an expression *e*, where the type arguments are not then followed by an argument list (then it's just a generic invocation, not an instantiation), and where *e*:
```
stereotype441, line 219: This answer is bascially just a restatement of the cases we enumerate in the spec.  I wonder if it would be more helpful to people as an FAQ if we simplified it and restated it with examples, e.g.:

> **A:** After any expression.  The meaning depends on what kind of expression it follows.  For example:

* If the expression is a class, mixin, or type alias, it just supplies type parameters to the type, e.g. `Type t = List<int>;`  Or, if the type is followed by a constructor name, then it's a constructor tear-off, e.g. `List<int> Function(int, int) f = List<int>.filled;`
* If the expression is a generic function, it instantiates it, e.g. `Iterable<int> Function(int Function(double)) = {1.0}.map<int>;`
* If the expression is a callable object with a generic `call` method, it instantiates the `call` method, e.g. if `foo.call` is a generic method, then `var f = foo<int>y` is equivalent to `var f = foo.call<int>;`

(Yes, I realize this glosses over a lot of details, but that's what a FAQ should do)

accepted/future-releases/constructor-tearoffs/faq.md
```diff
@@ -0,0 +1,256 @@
...
+
+In all other cases, it's an error.
+
+**Q:** Can I do an explicit instantiation on a `dynamic` value?
+
+**A:** No. We do allow you to *invoke* a `dynamic` value *v* as <code>*v*\<typeArgs>(args)</code>, but we do *not* allow <code>*v*\<typeArgs></code> as a dynamic instantiation. It's not one of the cases above because they all require or imply having a non-`dynamic` static type. (Same for expressions with static type `Function` or `Never`). In short, we need to know the static type of the type arguments before we will try to instantiate them.
+
+**Q:** If `<typeArgs>` is a selector, can't I write `foo<int><int>`?
+
+**A:** Nice try, but no. Since the token after the first `<int>` is `<`, and not one of the tokens listed above, the first `<` is parsed as a less-than operator, and then parsing will fail glamorously when it reaches the `>`.
```
stereotype441, line 235: Possible follow-up sentence you might want to consider:

> You could of course try doing `(foo<int>)<int>` but there's no way that would work in practice, because the type of `foo<int>` would be a non-generic function.

stereotype441 (approved):
All my comments below are nits.  Feel free to ignore them if you want.

accepted/future-releases/constructor-tearoffs/faq.md
lrhn: I agree. I could see that people were trying to introduce lints anyway, so I'd give them this, but it is probably better to not go there. Will remove the recommendation.

accepted/future-releases/constructor-tearoffs/faq.md
lrhn: Done.

accepted/future-releases/constructor-tearoffs/faq.md
lrhn: Very nice! Taken and rephrased slightly.

accepted/future-releases/constructor-tearoffs/faq.md
lrhn: Done.

accepted/future-releases/constructor-tearoffs/faq.md
```diff
@@ -0,0 +1,262 @@
...
+Example:
+
+```dart
+// Has type: DateTime Function(int, [int, int, int, int, int, int, int])
+var makeUtcDate = DateTime.utc; 
+```
+
+**Q:** What if the class is generic?
+
+**A:** Then the function is also generic, with the same type arguments as the class.
```
eernstg, line 29: I usually use `type arguments` to mean `actual type arguments` and `type parameters` to mean `formal type parameters`. In this case (where the reader can't be assumed to know the habits of the Dart specification documents) it would perhaps be best to use 'formal type parameters', such that anyone who has had a compiler course would get a rather firm hint about the intended meaning.

accepted/future-releases/constructor-tearoffs/faq.md
```diff
@@ -0,0 +1,262 @@
...
+
+**Q:** Can I write the type on the class, `List<int>.filled`, just like when calling?
+
+**A:** *Yes*!
+
+Example:
+
+```dart
+// Has type: List<String> Function(int, String) makeList
+var makeList = List<String>.filled; // Works!
```
eernstg, line 58: I think it could be helpful to show the regular call first.

accepted/future-releases/constructor-tearoffs/faq.md
```diff
@@ -0,0 +1,262 @@
...
+This also applies to *callable objects* (objects that have an interface type with a `call` method), which we treat like function values in most places.
+
+**Q:** Where does that even matter?
+
+**A:** If you have a generic function *value*, it's usually something you've received as a parameter at some point (if you knew which value it was, you'd just refer directly to a function declaration). It's probably going to be fairly rare to then need to instantiate that function to a specific type, instead of keeping it generic until it's called. It can happen. It makes explicit instantiation easier to explain too!
+
+Hypothetical example:
+
+```dart
+// Not a clue, mate. You tell me.
```
eernstg, line 161: Funny, but not so helpful. ;-)

accepted/future-releases/constructor-tearoffs/faq.md
```diff
@@ -0,0 +1,262 @@
...
+```
+
+**Q:** Couldn't I just instantiate the `call` method using method instantiation anyway?
+
+**A:** Yes and no. You could for callable objects, and we'd even add the `.call` implicitly for you (we do that for any callable object in a function-typed context, before checking whether the types actually work). It just didn't work for *real* function values. Dart2js never implemented instantiated tear-off of the `call` method of function values because that would be equivalent to instantiating the function value itself, which wasn't a supported feature (until now), so your code would just crash. We recognized this and initially planned to disallow doing a instantiated tear-off of a function's `call` method. Instead it turned out we can just support it consistently.
```
eernstg, line 166: Maybe we can skip the history part, and just say how it works when the callable objects generalization has been achieved? (And then, briefly: "This didn't work previously"; most developers don't care exactly how or why it didn't work ;-)

accepted/future-releases/constructor-tearoffs/faq.md
```diff
@@ -0,0 +1,262 @@
...
+
+Examples:
+
+```dart
+Type intListType = List<int>; // Explicit type literal instantiation.
+
+T id<T>(T value) => value; // Our standard generic function example.
+var idValue = id; // A function *value*.
+
+var intId = id<int>; // Explicit instantiation, saves on writing the function type.
```
eernstg, line 184: I think this should be split on two lines, to fit well in a not-so-large window on screen, and even in a printout. Perhaps there are similarly long lines elsewhere.

accepted/future-releases/constructor-tearoffs/faq.md
```diff
@@ -0,0 +1,262 @@
...
+var intId2 = idValue<int>; // Still works!
+var intId3 = (id)<int>; // Still works!
+
+var makeList = (List.filled)<String>; // List<String> Function(int, String)
+```
+
+The last example shows the generality of the feature, because the `(List.filled)` tear-off is a generic function value, you can instantiate it. *Don't even write that*, always use `List<String>.filled` instead!
+
+In short, we allow you to use `<typeArguments>` as a *selector*, like `.name` and `[expr]`, which you can chain after an expression. Previously we only allowed type arguments to occur before an argument list (`<typeArguments>(arguments)`) or after a class name in a constructor invocation (`ClassName<typeArguments>.name(arguments)`). Now we allow it after any expression, and with the same precedence as other selectors like argument lists, `.name` and `[expr]`.
```
eernstg, line 193: `after any expression` may seem less bureaucratic than the precise information, but it could also be a source of confusion (say, if someone tries `e as T <int>`). Maybe 'Now we allow it at all locations where a `<selector>` can occur, ...'

accepted/future-releases/constructor-tearoffs/faq.md
```diff
@@ -0,0 +1,262 @@
...
+**Q:** If I can instantiate a generic function, and `List.filled` is a generic function tear-off, why is `List.filled<int>` invalid?
+
+**A:** Because we say so. We could make it mean `(List.filled)<int>`, but we prefer to reserve the syntax for if/when we add proper generic constructors. And you *can* write the parentheses, just please don't. Prefer `List<int>.filled` if that's what you mean&mdash;and if it's not what you mean, you probably can't yet do what you mean.
+
+**Q:** So where can I write an explicit type argument instantiation.
+
+**A:** In short, after any expression which can actually use type arguments. The meaning depends on what generic thing is being instantiated.
+
+- If the expression denotes a generic class, mixin, or type alias, it just supplies type parameters to the type, e.g., `Type t = List<int>;` If the type is then followed by a constructor name, then it's a constructor tear-off, e.g., `var f = List<int>.filled;`. Otherwise it's a `Type` literal.
+- If the expression is a generic function, it instantiates it, e.g., `var f = [1.0].map<int>;`.
```
eernstg, line 230: It might be helpful to mention that `f` is a function that accepts an argument of type `int Function(double)` (if I got that right ;-).

accepted/future-releases/constructor-tearoffs/faq.md
```diff
@@ -0,0 +1,262 @@
...
+- If the expression is a callable object with a generic `call` method, it instantiates the `call` method, e.g., if `foo.call` is a generic method, then `var f = foo<int>;` is equivalent to `var f = foo.call<int>;`.
+
+In all other cases, it's an error. You can only instantiate something which is generic.
+
+**Q:** Can I do an explicit instantiation on a `dynamic` value?
+
+**A:** No. We do allow you to *invoke* a `dynamic` value *v* as <code>*v*\<typeArgs>(args)</code>, but we do *not* allow <code>*v*\<typeArgs></code> as a dynamic instantiation. It's not one of the cases above because they all require or imply having a non-`dynamic` static type. (Same for expressions with static type `Function` or `Never`). In short, we need to know the static type of the type arguments before we will try to instantiate them.
```
eernstg, line 237: Maybe change 'having a non-`dynamic` static type ...' to 'having a static type that cannot be `dynamic` (and also cannot be `Function`, or `Never`)'.

eernstg (commented):
Finally reached this one, sorry about the delay! LGTM, too! Added just a few comments about locations where I think the text could be more helpful.

eernstg:
Ah, this one was merged after the latest refresh of the browser tab. Anyway, I think my suggestions could be useful for a separate PR.

PR #1831, opened 2021-08-31
Macros: Add a concept of "macro libraries"

jakemac53:
This introduces a new concept of a `macro` library, which gives us maximum predictability in terms of tree shaking of macros themselves from the final program. These can still be used just like normal libraries from a user perspective, and the extra boilerplate is just one keyword in the macro library itself.

### DIFF
--- a/working/macros/feature-specification.md
+++ b/working/macros/feature-specification.md
@@ -582,7 +582,7 @@ modifier on the library directive: `macro library my_macro;`.
   applications when the original library is imported.
 - The library containing the Macro interfaces and APIs is itself a macro
   library. This makes it only useful to import from other macro libraries, and
-  ensures its apis never leak unexpectedly into runtime apps.
+  ensures its APIs never leak unexpectedly into runtime apps.
 
 In this way, macros (and their imports) have no effect on the final program
 beyond the code that they generate. It also draws a clear distinction between
--- a/working/macros/feature-specification.md
+++ b/working/macros/feature-specification.md
@@ -559,6 +559,67 @@ though it violates the normal rules.
 We could instead only allow adding macros from _later_ phases, but that would
 restrict the functionality in unnecessary ways.
 
+## Macro Libraries and Tree Shaking
+
+Macros, and any dependencies of macros, should be guaranteed to be tree shaken
+in their entirety from the final program. At runtime, the macro definition
+libraries should not be available, nor any stubs of them. The imports should
+be entirely removed.
+
+In order to support this, we introduce a new type of library - a "macro"
+library. You can mark your library as a macro library through a `macro`
+modifier on the library directive: `macro library my_macro;`.
+
+- Macro libraries can import and freely use code from other macro libraries
+  or normal libraries as they choose.
+- If a normal library imports a macro library, it can only use symbols from
+  that library within macro application code.
+- Once macros are expanded, all macro applications and imports of macro
+  libraries are removed from the program.
+- Regular libraries _are_ allowed to export macro libraries (for convenience),
+  and that _does not_ make them a macro library. The _export_ is removed from
+  the final program, and its symbols are still only visible within macro
+  applications when the original library is imported.
+- The library containing the Macro interfaces and APIs is itself a macro
+  library. This makes it only useful to import from other macro libraries, and
+  ensures its apis never leak unexpectedly into runtime apps.
+
+In this way, macros (and their imports) have no effect on the final program
+beyond the code that they generate. It also draws a clear distinction between
+macro libraries and normal libraries, without most users having to understand
+the actual distinction between them (they appear similar from the outside).
+
+### Testing Macros
+
+While it may be desirable to write functional unit tests of macros, this
+approach does not generally allow for that, since macro code is not available
+at runtime.
+
+Instead users are expected to write integration style tests, where they use
+their macro directly in their test and then test the behavior of the resulting
+generated code.
+
+TODO: Can we come up with a better testing pattern? Possibly something where the
+tests are actually ran at compile time, as a macro?
+
+TODO: Can we expose some sort of golden tests functionality to match the
+expected output?  
+
+### Modular compilation
+
+During modular compilation, macro libraries should be compiled as their own
+module. These modules should never be provided to compiler backends, and are
+only used during kernel compilation.
+
+When creating a merged dill file, you should simply omit the macro library
+modules entirely from the concatenated dill. The same goes for dill manifest
+files, they should be omitted from the manifest.
+
+### Macros and `dart:mirrors`
+
+Since macros are removed entirely after compilation, we do not allow reflecting
+on them with `dart:mirrors`. 
+
 ## Scoping
 
 ### Resolved identifiers


working/macros/feature-specification.md
```diff
@@ -559,6 +559,67 @@ though it violates the normal rules.
...
+While it may be desirable to write functional unit tests of macros, this
+approach does not generally allow for that, since macro code is not available
+at runtime.
+
+Instead users are expected to write integration style tests, where they use
+their macro directly in their test and then test the behavior of the resulting
+generated code.
+
+TODO: Can we come up with a better testing pattern? Possibly something where the
+tests are actually ran at compile time, as a macro?
```
bwilkerson, line 603: I'm not sure what you mean by compile-time. Do you mean the time at which the macro is being loaded into the VM, which presumably happens when loading code that uses the macro?

Would this include the analyzer's execution of macros?

working/macros/feature-specification.md
```diff
@@ -559,6 +559,67 @@ though it violates the normal rules.
...
+
+TODO: Can we expose some sort of golden tests functionality to match the
+expected output?
```
bwilkerson, line 606: Does "expected output" mean the content of the file that would be generated by the analyzer, VM, or any other tool that requires a textual representation of the generated code?

Does the expected output include the imports of macro libraries and macro applications?

bwilkerson (commented):
@scheglov

working/macros/feature-specification.md
jakemac53: Essentially I was thinking you could maybe run a "test" where the entire test is actually just a macro (so the macro apis are available). This would mean the analyzer _would_ also run these just like any other macros. And so your "tests" would actually run as a part of the analyzer, which might be kind of cool actually.

But I would not be asking the analyzer to implement a testing framework or anything like that, just execute macros normally.

working/macros/feature-specification.md
jakemac53: In general I made these TODOs because I don't have the answers to these questions. Just future avenues to explore.

I don't have any specific ideas on how this could/should work right now, I just know people sometimes like to use golden files for testing codegen output today.

working/macros/feature-specification.md
```diff
@@ -559,6 +559,67 @@ though it violates the normal rules.
...
 restrict the functionality in unnecessary ways.
 
+## Macro Libraries and Tree Shaking
+
+Macros, and any dependencies of macros, should be guaranteed to be tree shaken
+in their entirety from the final program. At runtime, the macro definition
+libraries should not be available, nor any stubs of them. The imports should
+be entirely removed.
+
+In order to support this, we introduce a new type of library - a "macro"
```
munificent, line 569: Do we have evidence that this feature is actually needed? Do we know that we need to make users explicitly decide which libraries should be macro libraries and which shouldn't (and also lose the ability to test their macros!) just to ensure that the compiler does what it should be able to do automatically?

This feels like a lot of user effort for almost no value.

If we *do* want to go in this direction, then I think we shouldn't have macro *libraries*, we should have macro *imports*. Whether some code is compiled into the final program is a property of how it's used, not how it's declared. We could say that you can mark any import as (maybe) `static`:

```dart
static import 'package:macro_stuff/macro_stuff.dart';
```

The behavior then is:

- It is a compile-time error to refer to any identifier imported from a `static` import outside of a macro application expression.

This would let you then test your macros by having the test code import the macro normally.

jakemac53: That is another approach that I considered for sure. The problem with it is that people are going to create libraries that contain both macros and other normal code. But they won't use the macro at all. Then, there will exist in the program some pattern that defeats the tree shaking of the macro, and we end up portions of the macro libraries and macros themselves shipped with real apps.

If we believe fully in tree shaking, then we don't have to do anything at all (including removing the import, it will all be tree shaken anyways!). 

The mechanism I proposed here should be 100% foolproof, and won't have the same pitfalls. It also doesn't require most users to even know these types of libraries exist - only the authors of the libraries have to know.

lrhn: I'm worried about intermingling macro code and run-time code too, so much that I don't like importing the *code* of the macros at all.

Consider instead:
* We have `dart:macros` which defines an abstract API and some base annotation classes.
* An annotation can extend those and specify a macro implementation *by URL* along with data to pass to that code.
* The macro is a script which gets an actual implementation of the macro API as second argument to `main`. We can define any API we want there, including allowing the script to stay-alive and handle more macro expansion requests, or just shut down when it's done.
* The macros are then defined in macro-libraries, which are perfectly normal libraries with a `main` method taking a `MacroContext` as second argument. You can run them with mock macro implementations for testing.

Then my `package:data_classes/data_classes.dart` would expose a public annotation `DataClass(...data...)` which extends `MacroAnnotation("package:data_classes/generator.dart", ...data...)`. The macro processor will see that, spawn that script (along with an implementation of the macros API, maybe fully fledged or maybe a shell which communicates with the real processor using ports), and will give a `MacroContext` to it containing `data` and anything it needs. It generates code and inspects existing code by interacting with the context.

But no library is a special *kind* of library. It's all which APIs it has access to, and who will invoke it.
And it keeps the macro implementation at arms length from the code which uses it.

jakemac53: It feels to me like that approach starts tying our hands to a specific implementation strategy, and it also opens up a fair bit of extra api without much benefit over this proposal? It does open up a strategy for testing, but does it solve much else?

The `MacroAnnotation` class and other apis exposed through `dart:macros` also leak into the program still in that case.

kevmoo: Yes...let's PLEASE be careful here.

TimWhiting: Mix of this proposal plus @munificent's comments.

What about having a macro folder next to lib. The end user imports the macro as follows:
```dart
import macro 'package:my_macro/my_macro.dart';
```
Users can only reference identifiers in the macro library in macro contexts. Macro libraries are recommended to keep all macro specific code in the macro folder, but can import from other libraries / their own lib. The lib folder is used for any abstract classes / etc that the macro author wants to provide for end users that actually needs to be in their code. All macro imports and code in the macro folder get tree-shaken. Nothing in the lib folder can import from the macro folder of the same package, thus ensuring that tree-shaking should run into no problems. As a benefit the macro author does not need to add a library annotation for each library containing a macro, because it is path based.

lrhn: Can you *use* a macro defined in the same package in the `lib` directory? (That seems like a primary use-case: defining your own project-specific macros.)


We could put `MacroAnnotation` in `dart:core`, and only make the `dart:macro` API available in some modes (compiling macros, and compiling tests for macros). If you don't explicitly opt in to it somehow, it won't be available in production code.

TimWhiting: I don't think use should be restricted, only referencing any identifier from the macro folder in a non-macro application context.

jakemac53: I haven't been convinced yet that these proposals are better than the one I have here, for a few reasons:

- This proposal guarantees that it does what it sets out to do - remove all macro related code from compiled apps. That is really the only goal of this proposal, and its not  a property I think we should compromise on.
- I think that a clear separation of macro libraries and non-macro libraries is a _good thing_. It overall simplifies the model a lot, without much downside. You can still make them appear as one library to users by exporting a macro library from a normal library.
- I think it is also a good thing to have macros themselves be imported as actual code - it makes it much easier to go to definition and inspect the macro itself, look at its documentation, or try to see where it is going wrong (or how you might be using it wrong). You can also immediately see the macro interfaces it implements (and thus know on what types of declarations it is allowed).
- It requires less overall boilerplate, and less people have to understand new concepts (only macro authors).

It does have the downside of making unit testing being more difficult, but that is only because it is _so good_ at doing its job. I don't think that enabling unit tests is actually intractable either - we should be able to provide a framework for unit testing macros by writing your tests themselves as macros. It will probably be a bit annoying to do, but should be possible.

munificent: > The problem with it is that people are going to create libraries that contain both macros and other normal code.

Users tend not to do things that end up hurting them. They won't do this if it causes them or their users pain. And if it doesn't cause them pain... then I see no real reason to prohibit it. I trust that they will organize their code in ways that make sense for them.

In particular, macro code will likely import something like `dart:macro` in order to access the macro API. We can make that library inaccessible in production (i.e. non-test) compilation, which will force users to separate out their macro code into other libraries.

I think the worry you have here is analogous to worrying that people will lump code using `dart:html` and other code using `dart:io` into the same library. But, in practice, they don't because that causes problems that they experience themselves.

bwilkerson: > I think it is also a good thing to have macros themselves be imported as actual code - it makes it much easier to go to definition and inspect the macro itself, look at its documentation, or ...

I'm not arguing either for or against your proposal, but as long as there's a way for the analyzer to find the macro definition (whether that's via explicit import or some other mechanism), and it's guaranteed that there will be, then it will be able to gather enough information to allow navigation, display documentation, etc.

> You can also immediately see the macro interfaces it implements (and thus know on what types of declarations it is allowed).

And we can also provide more tooling support if that's useful. At a minimum I would expect that the analyzer would produce a diagnostic if a macro were associated with a declaration that wasn't appropriate for it. We could also produce a list of valid targets in the hover text if that would be useful.

scheglov: I don't see how importing a library with macro builders and not using them is different from importing some other library and not using some of its parts. In both cases the result is that your compiled application is bigger than it could be. In both cases the solution is not avoid these (which?) patterns that don't allow tree shaking do its work.

jakemac53: The difference is we _can_ make a clear distinction here about code that we know is only useful compile time (excluding the test use case, though). That isn't true of other code, so it is different.

Also, I believe that the set of dependencies used by macro code and runtime code will be _very_ different. Consider the set of `dev_dependencies` that are typically added in an app that uses build_runner today. This quite often is many many more packages that are actually used at runtime.  

It is true we don't _have_ to take this approach. But I do think it would be a mistake not to enforce that these build time dependencies get removed at runtime.

jakemac53: If we don't want to make a separate type of library to enforce this, then I would probably want to go with something like the proposal @lrhn suggested.

One advantage of that is it could simplify some of the questions around arguments to macros. We could treat it more as just serializable data, and that has its appeals for sure.

jakemac53: Macros have to be regular dependencies - unless we decided packages should ship with macros already ran - but I don't think we want to do that.

But we do still want all the macro dependencies to be eliminated from the final application.

munificent:
I'm still leaning quite strongly against this. It feels like a (fairly complex) solution in search of a problem. I think we should leave solving this problem out until we know it's an actual problem.

jakemac53:
Closing this and will instead file an issue regarding code bloat/tree shaking.

jakemac53:
I am re-opening this issue given the [reflected imports](https://github.com/dart-lang/language/blob/master/working/reflected-imports/feature-specification.md) proposal, as I think it solves an actual requirement for that proposal.

Specifically, the proposal says this today:

```
This implies that reflected imports can only appear in libraries that are run in the macro execution environment. You can think of the macro environment as a distinct Dart "platform" and only it supports reflected imports. When targeting any other execution environment, a Dart compiler reports a compile-time error if it encounters a reflected import.
```

We have no way of actually enforcing that requirement without something like this proposal, which can separate the "compile-time" imports from the "run-time" imports.

jakemac53:
I also believe that the reflected imports will anyways cause a problem for unit testing macros, so this proposal also not allowing that becomes a non-issue.

jakemac53:
Closing this

Issue #1830, opened 2021-08-31
Specification of constructor tearoffs on a mixin application class may need clarification


eernstg:
[Edit: Using the latest commit, the program below is accepted and working. So this issue is now only about the specification. Adjusted the text below accordingly.]

Consider the following program:

```dart
class A { A.named(); A(); }
mixin M {}

class B = A with M;

void main() {
  var f = B.named;
  var g = B.new;
}
```

The mixin application class `B` makes constructors accessible (we can do `B.named()` and `B()`) ~~but it does not support tearoffs of the same constructors. This seems inconsistent.~~ In the latest commit, the tearoffs are supported by both the analyzer and the CFE.

If we agree that these tearoffs should be allowed then we may need to clarify the topic in the constructor-tearoffs specification.

leafpetersen:
Both those are valid constructors to call on the class `B`, right?  So given that, I can't see any justification for not allowing them to be torn off.

eernstg:
Right, that was my reasoning as well.

eernstg:
We could proceed directly and create an SDK issue for this immediately, and then separately make sure that the specification is adjusted as needed.

leafpetersen:
SGTM, thanks!

eernstg:
Haha, it works in commit 65b5125223386a4d9cdf275ff82a4853b793c9b9, so there is no need to change anything in the implementations! (And I did not create an SDK issue).

The only remaining issue is whether we'd need to adjust the specification. @lrhn, do you think this case is covered already?

lrhn:
I think the case is covered. The mixin application specification requires the mixin application class to have those constructors. Nothing outside of that should care how those constructors came to be. Tearing them off should just work like any other constructor.

eernstg:
SG, cf. https://github.com/dart-lang/language/pull/1832.

eernstg:
Spec change done in https://github.com/dart-lang/language/pull/1823.

PR #1832, opened 2021-09-01
Add section about tearoffs of implicitly induced constructors

eernstg:
Cf. https://github.com/dart-lang/language/issues/1830, this CL adds a short section making it explicit that implicitly induced constructors can be torn off, just like the explicitly declared ones.


eernstg:
Closed (abandoned), done in #1823.

PR #1833, opened 2021-09-01
Adjust the rules about parameters that are covariant due to inheritance: No longer an error

eernstg:
New tests concerned with the treatment of parameters that are covariant-by-declaration due to inheritance are introduced by https://dart-review.googlesource.com/c/sdk/+/208504. The language team decided that this kind of inheritance should be allowed (until now, it was specified to be an error), partly for consistency/simplicity and partly because the CFE already generates the required forwarding stubs.

This PR makes the corresponding changes to the language specification. It introduces a new section by splitting the section 'Classes' in two, because the latter half of this section is really not about classes per se, it is about a topic that we should show explicitly as a section title: 'Fully Implementing an Interface'.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -34,6 +34,8 @@
 %   error for `int.toString()`.
 % - Add support for function closurization for callable objects, in the sense
 %   that `o` is desugared as `o.call` when the context type is a function type.
+% - Clarify the treatment of `covariant` parameters in the interface of a class
+%   that inherits an implementation where those parameters are not covariant.
 %
 % 2.14
 % - Add constraint on type of parameter which is covariant-by-declaration in
@@ -48,8 +50,6 @@
 % - Change grammar to enable non-function type aliases. Correct rule for
 %   invoking `F.staticMethod()` where `F` is a type alias.
 % - Add missing error for cyclic redirecting generative constructor.
-% - Clarify the treatment of `covariant` parameters in the interface of a class
-%   that inherits an implementation where those parameters are not covariant.
 %
 % 2.8 - 2.10
 % - Change several warnings to compile-time errors, matching the actual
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -2521,49 +2521,76 @@ \section{Classes}
...
+For this check, any missing \COVARIANT{} modifiers are implicitly added
+to the inherited signature.
+When the modifier \COVARIANT{} is added to one or more parameters
+(which will only happen when the concrete member is inherited),
+an implementation may choose to implicitly induce a forwarding method
+with the same signature as $m'$,
+in order to perform the required dynamic type check
+and then invoke the inherited method.
+
+When there is a non-trivial \code{noSuchMethod} it is allowed
```
lrhn, line 2578: &mapsto; When a class has a non-trivial \code{noSuchMethod}, the class may leave ...

specification/dartLangSpec.tex
```diff
@@ -2521,49 +2521,76 @@ \section{Classes}
...
 to leave some members unimplemented,
 and it is allowed to have a \code{noSuchMethod} forwarder which does not
```
lrhn, line 2580: Which "it"? (Avoid unqualified "it"s, it's ambiguous! - pun intended!)

Try to avoid "it's is allowed" as a phrase meaning "it's not prohibited". I can't read it, and usually want at least a "by something" (which can then be rephrased to "something allows ..."). Others might have similar problems.

(I know these lines weren't really changed by this CL).

specification/dartLangSpec.tex
```diff
@@ -2521,49 +2521,76 @@ \section{Classes}
...
-Assume that $C$ has a concrete accessible member with the same name as $m$,
-with a method signature $m'$ which is a correct override of $m$.
-For each parameter $p$ of $m'$ where \COVARIANT{} is present,
-it is a compile-time error if there exists
-a direct or indirect superinterface of $C$ which has
-an accessible method signature $m''$ with the same name as $m$,
-such that $m''$ has a parameter $p''$ that corresponds to $p$
-(\ref{covariantParameters}),
-unless the type of $p$ is a subtype or a supertype of the type of $p''$.
+One more constraint must be satisfied
```
lrhn, line 2586: Nice foreshadowing, but be more direct. :)

> Parameters that are covariant-by-declaration must also satisfy the following constraint:

specification/dartLangSpec.tex
```diff
@@ -2521,49 +2521,76 @@ \section{Classes}
...
 
 \commentary{%
-In particular, it is an error for a class to be concrete even if it inherits
+So it is an error for a class to be concrete even if it declares or inherits
 a member implementation for every member signature in its interface,
 unless each of them has parameters and types such that they satisfy
 the corresponding member signature.
-But when there is a non-trivial \code{noSuchMethod} it is allowed
+For this check, any missing \COVARIANT{} modifiers are implicitly added
+to the inherited signature.
```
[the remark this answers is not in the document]
eernstg: No, because the "inherited signature" is the signature of the concrete member declaration which is inherited by `$C$`, and there is only one of those.

specification/dartLangSpec.tex
```diff
@@ -2521,49 +2521,76 @@ \section{Classes}
...
+When the modifier \COVARIANT{} is added to one or more parameters
+(which will only happen when the concrete member is inherited),
```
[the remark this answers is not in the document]
eernstg: The modifier is added to parameters of the member signature, but that signature will never fail to have them already, unless the member is inherited.

specification/dartLangSpec.tex
eernstg: Right, done!

specification/dartLangSpec.tex
eernstg: I think it's better now!

specification/dartLangSpec.tex

eernstg (commented):
Much improved, PTAL!

specification/dartLangSpec.tex
```diff
@@ -2531,62 +2531,20 @@ \section{Classes}
...
-as the constraint which is specified for a declaration in $C$
-(\ref{instanceMethods}).%
-}
+It is a compile-time error if a class named $C$ declares
+a member with basename (\ref{classMemberConflicts}) $C$.
+If a generic class named $G$ declares a type variable named $X$,
+it is a compile-time error
+if $X$ is equal to $G$,
+or if $G$ has a member whose basename is $X$,
+or if $G$ has a constructor named \code{$G$.$X$}.
```
lrhn, line 2540: (We haven't defined the basename of a constructor name *G.X* to be *G* yet?)

eernstg: No, basename still simply strips a trailing `=`.

specification/dartLangSpec.tex
```diff
@@ -2521,49 +2521,76 @@ \section{Classes}
...
 unless $C$ has a non-trivial \code{noSuchMethod}
 (\ref{theMethodNoSuchMethod}).
-It is a compile-time error if $C$ has
-a concrete accessible member with the same name as $m$,
-with a method signature $m'$ which is not a correct override of $m$
+
+\LMHash{}%
+Each concrete member must have a suitable signature.
+Assume that $C$ has a concrete accessible member with the same name as $m$,
+and let $m''$ be its member signature.
```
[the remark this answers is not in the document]
eernstg: (I did a little bit of renaming, I think it's clearer now.)

eernstg:
Friendly ping?

lrhn (approved):
My bad, forgot to publish comments :(

specification/dartLangSpec.tex
```diff
@@ -2530,91 +2530,169 @@ \section{Classes}
...
-unless each of them has parameters and types such that they satisfy
-the corresponding member signature.
-But when there is a non-trivial \code{noSuchMethod} it is allowed
-to leave some members unimplemented,
-and it is allowed to have a \code{noSuchMethod} forwarder which does not
-satisfy the class interface
-(in which case it will be overridden by another \code{noSuchMethod} forwarder).%
+Consider a concrete class \code{C},
+and assume that \code{C} declares or inherits
+a member implementation with the right name
```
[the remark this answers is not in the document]
eernstg: Yep, done!

specification/dartLangSpec.tex
```diff
@@ -2530,91 +2530,169 @@ \section{Classes}
...
+for every member signature in its interface.
+It is still an error if one or more of them has parameters or types
```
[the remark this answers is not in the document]
eernstg: Done!

specification/dartLangSpec.tex
```diff
@@ -2530,91 +2530,169 @@ \section{Classes}
...
-it is a compile-time error if there exists
-a direct or indirect superinterface of $C$ which has
-an accessible method signature $m''$ with the same name as $m$,
-such that $m''$ has a parameter $p''$ that corresponds to $p$
-(\ref{covariantParameters}),
-unless the type of $p$ is a subtype or a supertype of the type of $p''$.
+The use of an implicitly induced forwarding method
+when the modifier \COVARIANT{} is added to
+one or more parameters in $m'$
+is allowed,
```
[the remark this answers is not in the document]
eernstg: "is allowed" is intended to mean that a conforming implementation can do it, but it could also use some other implementation technique. I've changed it to use "implementation specific", that's better.

> it can then get run-time failures on seemingly
> type-valid invocations

We must have the dynamic checks in any case, because those upcasts on the receiver can make a seemingly type-valid invocation incur a type error at run time.

> the forwarder is then one implementation approach
> to desugaring that behavior, but it's visible at runtime
> whether you have a forwarder or not on the runtime type
> of the tearoff

Exactly.
When we say that it is allowed for an implementation to use the forwarding method, we avoid requiring that the inherited method (somehow) performs the dynamic type checks on its actual arguments. This could yield better performance, because we might otherwise perform those dynamic type checks in a lot of method implementations

specification/dartLangSpec.tex
```diff
@@ -2530,91 +2530,169 @@ \section{Classes}
...
+the member signature in the interface of $C$ is $m$.
+
 \commentary{%
-Here are simple examples, that illustrate the difference between
-``has a member'' and ``declares a member''.
-For example, \code{B} \IndexCustom{declares}{declares member}
- one member named \code{f},
-but \IndexCustom{has}{has member} two such members.
-The rules of inheritance determine what members a class has.%
+The forwarding method should not change the interface of \code{C},
```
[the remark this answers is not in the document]
eernstg: 'does not', because it should not! ;-)

eernstg (commented):
Review response.

One thing remains: I haven't renamed member signatures to `$s$` and variants thereof, because we have about 5 times as many occurrences of `$m$` and such used to denote member signatures, and it clashes with the use of `$s$` to denote the number of type parameters/arguments in various situations.

PR #1837, opened 2021-09-02
Add section about errors to tearoffs spec

eernstg:
There may be a need to specify some additional compile-time errors in the constructor-tearoffs spec. This PR adds a section for that, and includes one error which was needed according to https://github.com/dart-lang/sdk/issues/47077.

### DIFF
--- a/accepted/future-releases/constructor-tearoffs/feature-specification.md
+++ b/accepted/future-releases/constructor-tearoffs/feature-specification.md
@@ -402,6 +402,18 @@ A compile-time error occurs if a constructor tear-off denotes a generative const
 
 *A generative constructor declared in an abstract class may well be executed as part of the execution of a constructor tear-off, via a superinitializer. It cannot, however, be torn off on its own, because the execution of a constructor tear-off is an instance creation, and abstract classes do not have instances. There is no similar constraint on a factory constructor.*
 
+### Compile-time errors
+
+This section lists the compile-time errors associated with this feature. The _italic_ entries are commentary, which indicates that they are stated elsewhere already.
+
+It is compile-time error if an expression of the form `e.new <parameterPart>?` occurs, and `e` is not a type literal.
+
+_It is a compile-time error if an expression of the form `e <typeArguments>` occurs, and the static type of `e` is `Function` or `dynamic`._
+
+_It is a compile-time error occurs if a constructor tear-off denotes a generative constructor declared in an abstract class._
+
+_The basename of a constructor declared with `C.new` is `C`. It is a compile-time error if a class declares more than one constructor with the same basename._
+
 ### Grammar changes
 
 The grammar changes necessary for these changes are provided separately (as [changes to the spec grammar](https://dart-review.googlesource.com/c/sdk/+/197161)).


accepted/future-releases/constructor-tearoffs/feature-specification.md
```diff
@@ -402,6 +402,18 @@ A compile-time error occurs if a constructor tear-off denotes a generative const
 
 *A generative constructor declared in an abstract class may well be executed as part of the execution of a constructor tear-off, via a superinitializer. It cannot, however, be torn off on its own, because the execution of a constructor tear-off is an instance creation, and abstract classes do not have instances. There is no similar constraint on a factory constructor.*
 
+### Compile-time errors
+
+This section lists the compile-time errors associated with this feature. The _italic_ entries are commentary, which indicates that they are stated elsewhere already.
+
```
lrhn, line 408: I'd say "summarizes" instead of "lists", since it only lists *one* entry, then the rest are then stated elsewhere. (A "list of one thing" is just a thing).

Also, it doesn't really summarize all the errors.

accepted/future-releases/constructor-tearoffs/feature-specification.md
```diff
@@ -402,6 +402,18 @@ A compile-time error occurs if a constructor tear-off denotes a generative const
...
+It is compile-time error if an expression of the form `e.new <parameterPart>?` occurs, and `e` is not a type literal.
```
lrhn, line 409: Can you can drop the `<parameterPart>?`? Since it's optional, it's an error without it anyway, so why mention it at all?
If `e.new` occurs *as an expression* and *e* is not a type literal, then it doesn't matter what it's followed by.

accepted/future-releases/constructor-tearoffs/feature-specification.md
```diff
@@ -402,6 +402,18 @@ A compile-time error occurs if a constructor tear-off denotes a generative const
...
+
+_It is a compile-time error if an expression of the form `e <typeArguments>` occurs, and the static type of `e` is `Function` or `dynamic`._
```
lrhn, line 411: It's an error *unless*
- `e` is a type literal denoting a generic class, mixin or type alias declaration, or
- the static type of `e` is a generic function type.

Since neither `Function` nor `dynamic` satisfies that, they're included (but probably still worth calling out specifically).

accepted/future-releases/constructor-tearoffs/feature-specification.md
eernstg: I don't think that would work (to drop `<parameterPart>?`). We generally specify the rules for certain syntactic forms, and there is no reason to assume that a short sequence _s_ of tokens has any properties that are valid for any sequence _s1_ of tokens which is obtained from _s_ by adding extra tokens at the end. In particular, just because we've said that `e.new` is an error when `e` satisfies certain criteria, we haven't said anything at all about the error/non-error status of something which is derived from `e.new <parameterPart>`.

The reason why we can talk about the latter at all is that we deal with the properties of such constructs (like an instance creation expression) as a whole. The rules about an `e.new` which is followed by something else are simply irrelevant for `e.new <parameterPart>`.

We might require that we specify this more explicitly, such that we talk about `e.new`-which-is-not-followed-by-`<parameterPart>` vs. `e.new <parameterPart>`, but we don't do that consistently, and I think it would give rise to a very verbose style.

So I did not change anything here at this point.

accepted/future-releases/constructor-tearoffs/feature-specification.md
eernstg: Those two requirements follow from the normal static analysis. I'm aiming at specifying the special cases which are otherwise given special treatment: We _do_ allow `f<int>()` if the static type of `f` is `Function` or `dynamic`, so why wouldn't we support that with a tearoff, that is, `f<int>`, as well?

Do

eernstg (commented):
Review response.

eernstg:
@mateusfccp, the bots cannot confirm author consent from you on this PR, can you give that?

mateusfccp:
Sorry, @eernstg.

mateusfccp:
@googlebot I consent.

eernstg:
Thx, @mateusfccp!

eernstg:
Thank you!

PR #1838, opened 2021-09-03
Specify that C.m, C.m() cannot invoke an extension method

eernstg:
It has been an error for years to use `int.toString()` to call the `Object` instance method `toString` on an instance of `Type`. This syntax ("`<typeLiteral> '.' <memberAccess>`") is reserved for static member access.

It is inconsistent if we allow the same syntax to invoke an extension method (because an extension method will always be dominated by an instance method on the static type of the receiver, and hence extension methods are presumed to _always_ be invoked in a situation where we could also invoke an instance method).

The current specification wording does not make this obvious, and you could argue that it allows for `int.myExtensionMethod()` to invoke an extension method on `Type`.

This PR makes it explicit and unambiguous that it is not possible to implicitly call an extension method on a type literal.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -30,6 +30,8 @@
 % - Allow generic instantiation of expressions with a generic function type
 %   (until now, it was an error unless the expression denoted a declaration).
 % - Allow generic instantiation of the `call` method of a function object.
+% - Disallow `TypeLiteral.extensionMethod()`, in line with the error for
+%   `int.toString()`.
 %
 % 2.14
 % - Add constraint on type of parameter which is covariant-by-declaration in
@@ -6045,9 +6047,10 @@ An implicit extension member invocation occurs
 for a member invocation $i$
 (\ref{memberInvocations})
 with receiver $r$ and corresponding member name $m$ if{}f
-(1) the interface of the static type of $r$ does not have a member
+(1) $r$ is not a type literal,
+(2) the interface of the static type of $r$ does not have a member
 whose basename is the basename of $m$,
-and (2) there exists a unique most specific
+and (3) there exists a unique most specific
 (\ref{extensionSpecificity})
 extension denoted by $E$ which is accessible
 (\ref{extensionAccessibility})
@@ -13654,16 +13657,30 @@ and the static type of $i$ is as specified there.
 
 \LMHash{}%
 It is a compile-time error to invoke any of the methods of class \code{Object}
-on a constant type literal that is immediately followed by the token `.'\,.
+on a type literal that is immediately followed by the token `.' (a period).
 
 \rationale{%
-The reason for the latter is that this syntax is reserved for
-invocation of static methods.
+The reason for this rule is that this syntax is reserved for
+invocation of static members.
+Invocation of a static member of a class, mixin, or extension
+uses said entity as a \emph{namespace},
+not as an actual class, mixin, or extension.
+In particular, the syntactic receiver is not evaluated
+to an object---that would not even be possible for an extension.%
+}
+
+\commentary{%
 For instance, \code{int.toString()} is similar to
-\code{C.someStaticMethod()}, and
-it would be confusing if just a couple of expressions of this form
-were instance method invocations.
-If needed, \code{(int).toString()} may be used instead.%
+\code{C.someStaticMethod()}, and it would be confusing
+if the former would evaluate the receiver and call an instance method,
+and the latter (which is much more common and useful)
+would use the type literal as a namespace.
+If needed, \code{(int).toString()} may be used instead.
+
+As a natural consequence of this rule,
+a type literal cannot be the receiver in
+an implicit invocation of an extension method
+(\ref{implicitExtensionInvocations}).%
 }
 
 \LMHash{}%
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -13676,7 +13676,7 @@ to an object---that would not even be possible for an extension.%
 An member access on a type literal 
 (e.g., \code{C.id()}, \code{C.id}, or \code{C?.id()}),
 always treats the declaration denoted by the literal as
-a namespace for accessing static members.
+a namespace for accessing static members or constructors.
 For instance, \code{int.toString()} is an error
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -30,6 +30,8 @@
 % - Allow generic instantiation of expressions with a generic function type
 %   (until now, it was an error unless the expression denoted a declaration).
 % - Allow generic instantiation of the `call` method of a function object.
+% - Disallow `TypeLiteral.extensionMethod()`, in line with the error for
```
lrhn, line 33: Maybe say that we make it *explicit* that it was already disallowed.
I think it was. The alternative was that the language was ambiguous and `C.m` could mean *both* a static member and an extension member, and we didn't provide any disambiguation.

specification/dartLangSpec.tex
```diff
@@ -13654,16 +13657,30 @@ \subsubsection{Ordinary Invocation}
 
 \LMHash{}%
 It is a compile-time error to invoke any of the methods of class \code{Object}
-on a constant type literal that is immediately followed by the token `.'\,.
+on a type literal that is immediately followed by the token `.' (a period).
```
lrhn, line 13660: Should we mention `?.` as well.
`C?.toString()` is "treated like" `C.toString()`, but it is not "immediately followed by the token `.'".
I could argue that it's not a compile time error, and then it's treated like `C.toString()` would be as a non-compile-time error. (I can argue differently too, but that's still bad).

Or should we just remove this, and just rely on the rule that if `C` is a type literal then `C.m` is a compile time error if `C` does not declare a static member named `m`.
(This text comes from back when static lookups were lookups on objects with the static members as instance members. It never worked.)

Basically (which is what I think this CL intends, just summarizing):

If `T` is a type literal or denotes an extension, then the conditional property extraction `T?.m` is equivalent to `T.m` (and the conditional method invocation `T?.m()` is equivalent to `T.m()`).
If `T` is a type literal denoting a class or mixin `C`, or denoting an extension `E`, then `T.m` and `T.m()` are compile-time errors if `C`, respectively `E`, does not declare a static member named `m`.

(Come constructor tear-offs, make it "a static member or constructor with base name `m`", and only allow static accesses on non-instantiated type literals).

Then make the ordinary method invocation/property extraction `e.m` split into either when `e` is a type literal, or when it's not, and have separate behaviors for those.


Then it's a compile-time error to write `C.toString()` not (directly) because `Object` declares `toString`, but because `C` does not declare a static with that name (which it can't because it cannot declare a static with the same name as an inherited instance member, so only indirectly because it's declared by `Object`).

The result should be the same, but without the "one token lookahead" which doesn't really scale.

specification/dartLangSpec.tex
```diff
@@ -13654,16 +13657,30 @@ \subsubsection{Ordinary Invocation}
...
-The reason for the latter is that this syntax is reserved for
-invocation of static methods.
+The reason for this rule is that this syntax is reserved for
+invocation of static members.
+Invocation of a static member of a class, mixin, or extension
+uses said entity as a \emph{namespace},
+not as an actual class, mixin, or extension.
+In particular, the syntactic receiver is not evaluated
+to an object---that would not even be possible for an extension.%
+}
```
lrhn, line 13670: Not sure what I'm expected to learn from this.
Especially since "this syntax" is just syntax of the form `id.id` which can be used for almost anything. It's not the syntax which is important, but perhaps the "construction" of a member access on a type literal.

specification/dartLangSpec.tex
```diff
@@ -13654,16 +13657,30 @@ \subsubsection{Ordinary Invocation}
...
 For instance, \code{int.toString()} is similar to
-\code{C.someStaticMethod()}, and
-it would be confusing if just a couple of expressions of this form
-were instance method invocations.
-If needed, \code{(int).toString()} may be used instead.%
+\code{C.someStaticMethod()}, and it would be confusing
+if the former would evaluate the receiver and call an instance method,
+and the latter (which is much more common and useful)
+would use the type literal as a namespace.
+If needed, \code{(int).toString()} may be used instead.
```
lrhn, line 13678: This also confuses me a little, rather than illuminating me.
In general, don't talk about hypotheticals that we could have done instead, and then explain why we didn't. That's introducing more concepts for the reader to understand, without actually helping them understanding *what is*.

Maybe just make it short:

> An member invocation or property access on a type literal, `L.id` or `L?.id`, 
> always treas the declaration denoted by the literal as a namespace
> for accessing static members.
> For instance, `int.toString()` will be an error because `int`
> does not declare a static `toString` member.
> It will not evaluate `int` to a `Type` object and then call its `toString` instance method,
> to do that, you should write `(int).toString()` instead.
> (Cascades are different, though, since they always evaluate their receiver to an object first.).

eernstg: Yes, that's better. Done.

eernstg (commented):
Review response.

One issue remains: I did not change the wording around "the token '.'" to talk about static methods, it is still talking about errors for instance methods. PTAL.

specification/dartLangSpec.tex
lrhn: If we just remove the entire paragraph, then I don't think there is any problem at all.
Or make it descriptive instead of nominative.

The semantics of `C.m` and `C.m()` where `C` is a type literal are specified such that we don't even *try* to look up an instance member, only a static member (or constructor for invocations), and we also won't go to the extension method phase which comes after failing to find an instance member.


We'll definitely need to revisit a lot of things :)

specification/dartLangSpec.tex
```diff
@@ -13653,17 +13656,39 @@ \subsubsection{Ordinary Invocation}
...
+not as an actual class, mixin, or extension.
+In particular, the syntactic receiver is not evaluated
+to an object---that would not even be possible for an extension.%
+}
+
+\commentary{%
+An member access on a type literal 
+(e.g., \code{C.id()}, \code{C.id}, or \code{C?.id()}),
+always treats the declaration denoted by the literal as
+a namespace for accessing static members.
```
lrhn, line 13679: Well, technically, "or constructors".

lrhn (approved):
LGTM

specification/dartLangSpec.tex

PR #1839, opened 2021-09-03
Deploy specification with Firebase

mit-mit:


.github/workflows/ci.yml
```diff
@@ -34,13 +41,10 @@ jobs:
...
-      - name: Deploy
-        if: |
-          github.event_name == 'push'
-          && github.ref == 'refs/heads/master'
-          && github.repository == 'dart-lang/language'
-        run: |
-          cd specification
-          firebase deploy --project=dart-specification
-        env:
-          FIREBASE_TOKEN: ${{ secrets.FIREBASE_TOKEN }}
```
athomas, line 46: Please delete this token after merging this PR.

athomas (approved):

.github/workflows/ci.yml
```diff
@@ -34,13 +41,10 @@ jobs:
...
+      - uses: FirebaseExtended/action-hosting-deploy@0e16177d01e10c4560245404f959e75a9c633533
+        with:
+          repoToken: '${{ secrets.GITHUB_TOKEN }}'
+          firebaseServiceAccount: '${{ secrets.FIREBASE_SERVICE_ACCOUNT_PACKAGE_QUALITY }}'
```
athomas, line 47: Looks like this is an empty string when the PR is filed from a fork causing the workflow to fail. That's perhaps an unintended effect?

mit-mit: No, I just got the key name wrong!

PR #1842, opened 2021-09-03
Specify callable object closurization

eernstg:
This PR adds a new case to the set of possible 'property extraction' expressions: Implicit property extraction. It occurs where an expression `e` has a static type as a callable object, and the context type is a function type. In this case we desugar `e` into `e.call`. Note that this will also serve as the first step in a generic function instantiation, because it is still true that the context type is a function type also in the case where `e.call` isn't assignable to that context type (but we can fix it by means of an implicit instantiation).

Smaller corrections:

The subsubsection about generic function instantiation was turned into a subsection: It used to be the case that a generic function instantiation would necessarily include a function closurization as a sub-action. This is no longer true, we can perform the generic instantiation on an existing function object. This means that the generic instantiation material should no longer be a subsection of 'Function Closurization', it's a stand-alone feature.

Subtitle uses version 2.15-dev, not 2.13-dev.

Note that this change is _not_ expected to give rise to any implementation efforts. The following works today already:

```dart
class A { call() {}}
void f(void Function() g) => print(g);
void main() => f(A());
```

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -14,7 +14,7 @@
 \makeindex
 \title{Dart Programming Language Specification\\
 {6th edition draft}\\
-{\large Version 2.13-dev}}
+{\large Version 2.15-dev}}
 \author{}
 
 % For information about Location Markers (and in particular the
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -12783,7 +12783,7 @@ $i$ is treated as
 the ordinary invocation
 
 \noindent
-\code{$e_f$.call<$A_1, \ldots,\ A_r$>($a_1, \ldots,\ a_n,\ x_{n+1}$: $a_{n+1}, \ldots,\ x_{n+k}$: $a_{n+k}$)}.
+\code{$e_f$.\CALL<$A_1, \ldots,\ A_r$>($a_1, \ldots,\ a_n,\ x_{n+1}$: $a_{n+1}, \ldots,\ x_{n+k}$: $a_{n+k}$)}.
 
 \LMHash{}%
 Otherwise, the static analysis of $i$ is performed as specified
@@ -14088,7 +14088,8 @@ or it is an expression of the form \code{\SUPER.\id}
 \Case{Implicit}
 Let $e$ be an expression whose static type is
 an interface type that has a method named \CALL.
-In the case where the context type for $e$ is a function type,
+In the case where the context type for $e$ is a function type
+or the type \FUNCTION,
 $e$ is treated as \code{$e$.\CALL}.
 
 \commentary{%
@@ -14096,8 +14097,8 @@ This means that a ``callable object'' may be treated as a function
 that supports a mechanism similar to function closurization
 (\ref{functionClosurization})
 by desugaring it to a method closurization on \CALL.
-It is required that it is statically known that it is a callable object,
-and that the context type requires a function.%
+This only occurs when it is statically known that it is a callable object,
+and when the context type requires a function.%
 }
 \EndCase
 
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -2294,7 +2294,7 @@ Let $u$ be the run-time type of a function object $o$ obtained by
 function closurization
 (\ref{functionClosurization})
 or instance method closurization
-(\ref{ordinaryMethodClosurization})
+(\ref{instanceMethodClosurization})
 applied to $F$,
 and let $t$ be the actual type corresponding to $T$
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -14081,6 +14084,23 @@ \subsection{Property Extraction}
...
+\LMHash{}%
+\Case{Implicit}
+Let $e$ be an expression whose static type is
+an interface type that has a method named \CALL.
+In the case where the context type for $e$ is a function type,
+$e$ is treated as \code{$e$.\CALL}.
+
+\commentary{%
+This means that a ``callable object'' may be treated as a function
+that supports a mechanism similar to function closurization
```
lrhn, line 14096: What is a "function closurization" here? (Words, they mean things! But what?!?)

Callable objects are usually treated like function objects, which are inherently closed, so closurization isn't necessary. So to me "function" means "already closed", but I guess it can also be used about static functions.

specification/dartLangSpec.tex
```diff
@@ -14081,6 +14084,23 @@ \subsection{Property Extraction}
...
+(\ref{functionClosurization})
+by desugaring it to a method closurization on \CALL.
+It is required that it is statically known that it is a callable object,
```
lrhn, line 14099: &mapsto; It must be statically known that the value is ...

To many "it"s.

specification/dartLangSpec.tex
```diff
@@ -14081,6 +14084,23 @@ \subsection{Property Extraction}
 (\ref{superGetterAccessAndMethodClosurization}).
 \EndCase
 
+\LMHash{}%
+\Case{Implicit}
+Let $e$ be an expression whose static type is
+an interface type that has a method named \CALL.
+In the case where the context type for $e$ is a function type,
```
lrhn, line 14091: &mapsto; a function type or the type \FUNCTION,

lrhn (approved):

lrhn (approved):

specification/dartLangSpec.tex
```diff
@@ -14081,6 +14084,23 @@ \subsection{Property Extraction}
...
+$e$ is treated as \code{$e$.\CALL}.
```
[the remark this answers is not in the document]
eernstg: I want to have a consistent terminology for desugaring (and 'desugar' seems perhaps a bit too informal, even though it would probably offer very good readability).

However, we're not there yet: Old parts of the specification use 'is equivalent to', but it is a considerable unnecessary complication if we start implying that desugaring is a relation (so we may need to go back and forth, searching for a "solution" of sorts in a large space of possible rewritings), I want this to be a simple function: Take the given source code, desugar it, and that's the foundation for further work.

The phrase I've used (in many places, at least 63 places right now)  is 'A is treated as B'. At least, this is directional.

(Vague definitions mean exactly what's the most convenient thing right now ;-)

specification/dartLangSpec.tex
eernstg: Check the index: Function closurization is a concept and a section.

It's the process that takes something which is not specified to have any particular properties (a 'function') except for having a declaration, and yields a function object, whose behavior is specified to be the same as running the code in the body of said declaration with suitable bindings of formal parameters and type parameters the function declaration.

The reason why one could say that this is a mechanism which is similar to a function closurization is that we have some expression (in this case we do start with an expression whose value is actual object), and then we take some steps in order to obtain a function object that "corresponds" to the given expression.

However, it's commentary so it's always a discussion whether it's too imprecise to be helpful.

Technically, after desugaring, it is simply an instance method closurization (somehow named 'ordinary method closurization' in the spec, maybe I should fix that immediately...).

specification/dartLangSpec.tex
eernstg: Done!

specification/dartLangSpec.tex
eernstg: I left out `\FUNCTION` because we don't have that for the invocation, https://github.com/dart-lang/language/blob/123bad8162497c4fd54ef50fac186f8ed88773a1/specification/dartLangSpec.tex#L12781.

But I can see that the implementations (a couple of them, anyway) already support type `Function` as well, so we could also include it in both cases.

specification/dartLangSpec.tex
eernstg: Fixed that, too.

specification/dartLangSpec.tex
eernstg: Ah, by the way, this is not the receiver type, it is the context type, so there is no contradiction in using the type `Function` in the context as a request for a `.call`. Done!

PR #1861, opened 2021-09-17
add section on macro instantiation and parameters

jakemac53:
Followup to our discussion on making macro parameters more meta-y


TimWhiting:
This resolves to some extent https://github.com/jakemac53/macro_prototype/issues/26 via the `Code` interface if I am right? The only limitation would be that the AST will be unresolved and in the form of `Code` objects. 

Is there a reason that `Symbol` is omitted? With `Code` parameters I guess you could just take an unresolved identifier in the form of an expression which is a lot nicer than a `String` / `Symbol` based api for telling a macro what you want it to name some declaration in my opinion, but I just wondered why `Symbol` wasn't included since it seems like another 'primitive' object in dart.

jakemac53:
> This resolves to some extent [jakemac53/macro_prototype#26](https://github.com/jakemac53/macro_prototype/issues/26) via the `Code` interface if I am right? The only limitation would be that the AST will be unresolved and in the form of `Code` objects.

Correct it does in effect provide the interface for doing that - since you get `Code` objects you can use them directly in your output as well.

> Is there a reason that `Symbol` is omitted?

Nope, I just took a look through the `dart:core` types and listed the ones that seemed reasonable to support. I am actually not entirely sure about Symbol... passing a Symbol from one app to the other might not actually make sense (especially in release mode, the symbols lose all references to the textual value of the symbol I think?). cc @lrhn any thoughts on that?

working/macros/feature-specification.md
```diff
@@ -246,6 +246,41 @@ introspection APIs.
 These macros can fully introspect on any type reachable from the declarations
 they annotate, including introspecting on members of classes, etc.
 
+## Macro Instantiation
+
+Macros should be able to be able to be instantiated using only the libraries
+that are available to the macro definition library. This allows us to compile an
+app (or spawn an isolate) that can run macro applications for that macro using
+only its own dependencies.
```
lrhn, line 254: Who performs this instantiation? Where do they get the arguments from?

working/macros/feature-specification.md
```diff
@@ -246,6 +246,41 @@ introspection APIs.
...
+In order to facilitate this, parameters to Macro constructors are limited to
+a specific set of allowable types:
+
+- [Code][] or its subtypes
+- String
+- int, double, num
+- bool
+- null (really this just allows nullable types)
+- Uri
+- DateTime
```
lrhn, line 267: Why not just allow any value which can be safely deserialized by the isolate which is going to run the macro?
(And make it a compile-time error to specify any argument which can't)

working/macros/feature-specification.md
```diff
@@ -246,6 +246,41 @@ introspection APIs.
...
+
+### Macro Constructor Arguments
+
+When any type other than [Code][] is used, arguments are only allowed to be
+object literals. Or in the case of `Uri` and `DateTime` they must be direct
```
lrhn, line 272: So, the source code for this macro occurs as an annotation in the library using the macro. This *source code* is then executed in a separate isolate. Therefore the source code must be valid in that library. Is that correctly understood?
(And the annotation is not available at run-time in the original library).

In that case, I'd definitely allow any *literal* (including symbol literals, but not including "type literals").

I'm vary about `Uri` and `DateTime` because they do not have const constructors, and therefore shouldn't be in an annotation to begin with. I guess we'll need special rules for macro annotations to allow them to do anything, becaus they'll be instantiated in a non-`const` way eventually. Or you can just put in the string and parse it at the other end.

working/macros/feature-specification.md
```diff
@@ -246,6 +246,41 @@ introspection APIs.
...
+- [Code][] or its subtypes
+- String
+- int, double, num
+- bool
+- null (really this just allows nullable types)
+- Uri
+- DateTime
+
+### Macro Constructor Arguments
+
```
munificent, line 270: This is a little unclear to me about whether arguments are values or meta-values. How about something kind of like:

---

Arguments to macros are passed as *unevaluated code*. In other words, if you apply a macro like:

```dart
@myMacro(1 + 2)
class SomeClass {}
```

The argument passed is not the number 3, but a Code object representing the expression `1 + 2`. This is useful if a macro wants to take an arbitrary piece of syntax as an argument that it then injects into the generated code.

In many cases, though, what the macro author wants is an argument *value*. For example:

```dart
@coordinates(dimensions: 3)
class Vec3 {}
```

Here, the `@coordinates` macro generates, `x`, `y`, etc. fields based on the given dimension passed to the macro. To support this, a macro constructor can accept parameters of primitive types and it will receive the arguments as values, not Code objects representing the value. However, the argument expression must be a *literal* of that type, and not any arbitrary expression that happens to evaluate to that type.

For example:

```dart
@coordinates(dimensions: 1 + 2) // Error.
class Vec3 {}
```

?

munificent (approved):
Some suggestions on how to word it, but overall I'm excited to see you filling in this part of the proposal. :)

working/macros/feature-specification.md
lrhn: If the arguments are recorded as expressions, then can you use variables references?
Say:
```dart
const string = "something or other";
...
@MyMacro(string, 42)
...
```
How about interpolations:
```dart
@MyMacro("Header:$string:Trailer")
...
```

If you can't access const variables, then you also can't use enums, which is likely going to affect the APIs that you'd want to build around the macro annotation. 

(How about just serializing the macro annotation constant, deserializing it in the isolate running the macro, then failing compilation if you can't do both).

jakemac53: I took another shot at rewording this also. Borrowed some stuff from @munificent, but tried not to mislead people by oversimplifying things early on still.... I think its important to make the pass by value vs Code object distinction early on.

working/macros/feature-specification.md
jakemac53: Some of the other wording below starts to address this - I don't want to get overly specific as this could be implementation specific? For now its just "the compiler".

PR #1872, opened 2021-09-27
Clarify that `e<T>` can desugar to `e.call<T>` when `call` is an extension method

eernstg:
This PR adds an extra bullet to the list where it is specified what `e<typeArgs>` can mean, such that it can desugar to `e.call<typeArgs>` in the case where said `call` is an extension method.

It also changes the wording of the previous bullet such that it uses `tear-off` rather than `tear off` when that phrase is used as a noun phrase, for consistency with the rest of this document.

We might want to perform another change, but I haven't done that (yet): It is required several times in this itemized list that a `call` method is generic, and otherwise we do not desugar it by adding `.call`. However, the current language specification does not require any properties of the `call` method, which means that we'll commit to the `.call` desugaring and then (perhaps) discover that it is not a generic method.

I think this view on the situation is more informative, because it fits naturally with an error message that specifically says "`call` is not a generic method", just like it would say "`call` is a generic method, but you can't pass three type arguments because it only accepts two".

The current wording (where we skip out of that bullet as soon as we note that `call` isn't generic) tends to put this situation into a much more general bucket where `e<typeArgs>` goes wrong because we have tried all cases.

Of course, the error messages should always refer to the actual source code rather than a desugared form, but I still think that it would make sense to remove the requirements about `call` being a generic method from this bullet list.


accepted/future-releases/constructor-tearoffs/feature-specification.md
```diff
@@ -355,8 +355,9 @@ For an expression of the form <code>*e*\<*typeArgs*></code>, which is not follow
...
-* If *e* has a static type which is a generic callable object type (a non-function type with a generic method named `call`), then <code>*e*\<*typeArgs*></code> is equivalent to the instantiated method-tear off <code>*e*\.call<*typeArgs*></code>.
-* Otherwise, if *e* has a static type which is a generic function type, then <code>*e*\<*typeArgs*></code> is equivalent to the instantiated method-tear off <code>*e*\.call<*typeArgs*></code>.
+* If *e* has a static type which is a generic callable object type (an interface type with a generic method named `call`), then <code>*e*\<*typeArgs*></code> is equivalent to the instantiated method tear-off <code>*e*.call\<*typeArgs*></code>.
+* If *e* has an accessible and applicable generic extension method named `call` then <code>*e*\<*typeArgs*></code> is equivalent to the instantiated extension method tearoff <code>*e*.call\<*typeArgs*></code>.
```
lrhn, line 359: I guess that's OK now that we can instantiate anything. If you can tear off `e.call`, you can also instantiate it. (IIRC, we don't do implicit extension-`.call` tear-offs, but the type arguments make this explicit. So `Function f = e;` is invalid but `Function f = e<int>;` is valid.).

accepted/future-releases/constructor-tearoffs/feature-specification.md
```diff
@@ -735,3 +736,4 @@ In this case, most of the parameters are *unnecessary*, and a tear-off expressio
 * 2.14: Remove many disambiguation tokens. Allow instantiating function *objects* and *callable objects*. Mention forwarding constructors from mixin applications.
 * 2.15: Add section about constants and specify new rules about potentially constant and constant expressions of the form <code>e\<T<sub>1</sub>..T<sub>k</sub>></code>.
 * 2.16: Add one more kind of potential constant, type parameters.
+* 2.17: Specify that <code>*e*\<*typeArgs*></code> can desugar to <code>*e*.call\<*typeArgs*></code> when `call` is an extension method.
```
lrhn, line 739: Also update the current version at the top. It's been stuck on 2.14.

lrhn (approved):

accepted/future-releases/constructor-tearoffs/feature-specification.md
eernstg: Right, we did make it an error to tear off an extension method named `call` implicitly. E.g., `Function f = e;` where `e` has an interface type with a method named `call` is an error, it does not desugar to `Function f = e.call;`, presumably because there is no syntactic hint at all on `e` about the special desugaring.

It does not appear to be overwhelmingly consistent, but we do have a rule which says that a `.call` desugaring step is always marked by explicit syntax.

accepted/future-releases/constructor-tearoffs/feature-specification.md
eernstg: Ah, yes! Done.

eernstg:
And what do you think about removing 'generic' from that bullet list? The considerations that I mentioned are actually relevant to type literals and other non-`call`-method cases as well.

lrhn:
I think that's probably a good idea. In general, don't determine whether it's a type instantiation, a function instantiation, or a plain early error, depending on the static type of the receiver, *then* worry about whether that thing is actually a *generic* thing:

* If `e` denotes a type declaration, then `e<typeArgs>` is a _type instantiation_ of that type. If the type arguments do not apply to the type denoted by `e`, then that's an error, including if the type is not generic at all.
* If `e` has a static type which is a *function type*, then `e<typeArgs>` is a _function instantiation_ of that function. It's a compile-time error if the type arguments do not match type parameters of the type of `e`, including that type not being generic at all
* if `e` has a static type which is a callable object type (interface type declaring a `call` method), then `e<typeArgs>` is equivalent to the function instantiation `e.call<typeArgs>`, including whether that instantiation is an error.
* If `e` has a static type which has an accessible and applicable `call` extension method declared through the extension `E` (IIRC "applicable" means it's uniquely applicable, we're past conflict resolution), then `e<typeArgs>` is equivalent to the function instantiation `E(e).call<typeArgs>`. (Just to be absolutely sure that adding `.call<typeArgs>` doesn't change resolution in some weird way, which I think it doesn't. At least today.)
* Otherwise `e<typeArgs>` is a compile-time error because we don't know what the type instantiation would do. (This includes static types of `dynamic`, `Function` and `Never`, where we'd otherwise allow `e<typeArgs>(args)`.)

Am I missing something?

eernstg:
Sounds good! I think that's what I have with the most recent upload.

eernstg:
Forgot one thing:

> uniquely applicable

I believe the current wording implies that we commit to having the ambiguity failure: If there is _at least_ one accessible and applicable extension method named `call` then we desugar to `e.call<typeArgs>`, and that will incur the compile-time error (we have no instance method, but there are several extension methods, and none of them is most specific).

I think that's consistent with the treatment of the other cases.

PR #1944, opened 2021-10-28
Add WIP macro apis to the proposal

jakemac53:
I made a fair number of changes here from the original prototype apis. A lot of this was influenced by subsequent prototypes for actually loading and running macros in a separate isolate - it became clear we will want some asynchronous apis to enable us to split up information into smaller chunks for sending it efficiently to a separate macro expansion isolate.

The major differences are as follows:

- all types in all apis now appear as `TypeAnnotation` objects, and then the builder apis expose operations on those based on the capabilities in that phase.
- the ClassDeclaration api doesn't directly have information about its members - you have to also query separately for those through asynchronous apis on the builder objects.
  - this does feel a bit weird, so we could change it potentially

There are also still some holes I am sure in terms of missing information that we may want to expose for certain objects.

I also ported over the data class macro example to this API.

### DIFF
--- a/working/macros/api/builders.dart
+++ b/working/macros/api/builders.dart
@@ -1,0 +1,115 @@
+import 'code.dart';
+import 'introspection.dart';
+import 'macros.dart'; // For dart docs :(
+
+abstract class Builder {
+  /// Used to construct a [TypeAnnotation] to a runtime type available to the
+  /// the macro implementation code.
+  ///
+  /// This can be used to emit a reference to it in generated code, or do
+  /// subtype checks (depending on support in the current phase).
+  TypeAnnotation typeAnnotationOf<T>();
+}
+
+/// The api used by [TypeMacro]s to contribute new type declarations to the
+/// current library, and get [TypeAnnotation]s from runtime [Type] objects.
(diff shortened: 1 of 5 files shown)


munificent (approved):

jakemac53:
Superceded by https://github.com/dart-lang/language/pull/1964

jakemac53:
Re-opening this pull request, I think we need to re-evaluate the revised api and probably move back more to an approach like this one.

Primarily the reason is that this API gives us a cleaner separation of the phases - in particular it allows us to run only certain phases of a macro and not others. This allows for optimization by toolchains - for instance Phase 2 only has to be re-ran if the _current file_ is updated, since it cannot reflect on anything in the program.  Phase 2 only can reflect on types (from dependencies at least) and only has to be re-ran if types have changed.

The newer api also causes problems with macro composition. It becomes difficult to define how a macro might apply another macro, especially in any phase after phase 1. You end up only being able to directly invoke utility methods from other macros, which they must explicitly expose, instead of being able to invoke them directly.

jakemac53:
actually going to close this one and start over from scratch as that will be easier at this point

PR #1964, opened 2021-11-05
Alternate macro api proposal, less visible phases

jakemac53:
In this proposal, there are no explicit classes you implement for each phase, there is only a single class to implement based on the type of declaration you will annotate.

Each of these classes has a single visit method which only gets a declaration argument (no builder). They have external `buildTypes`, `buildDeclarations`, and `buildDefinition` methods. These methods take a callback as an argument, and the callback receives the corresponding builder.

Some advantages of this api are:

- Can have some shared state between phases, managed by local variables in the `visit` method.
- Less complicated class  heirarchy - far fewer classes to understand.
- The phases are less directly exposed, you don't really have to understand them so much any more.

Some disadvantages of this api are:

- Can result in runtime errors if violating phases - you can't call the `buildTypes` method from within the callback passed to `buildDeclarations` for instance. That would violate ordering, and we would have to throw at runtime. We could lint for this though.
- People may try to share state that isn't shareable across phases (for instance store the TypeBuilder and use it in the declaration phase).
- If only implementing one phase there is more boilerplate.

### DIFF
--- a/working/macros/api/builders.dart
+++ b/working/macros/api/builders.dart
@@ -1,0 +1,115 @@
+import 'code.dart';
+import 'introspection.dart';
+import 'macros.dart'; // For dart docs :(
+
+abstract class Builder {
+  /// Used to construct a [TypeAnnotation] to a runtime type available to the
+  /// the macro implementation code.
+  ///
+  /// This can be used to emit a reference to it in generated code, or do
+  /// subtype checks (depending on support in the current phase).
+  TypeAnnotation typeAnnotationOf<T>();
+}
+
+/// The api used by [TypeMacro]s to contribute new type declarations to the
+/// current library, and get [TypeAnnotation]s from runtime [Type] objects.
+abstract class TypeBuilder implements Builder {
+  /// Adds a new type declaration to the surrounding library.
+  void addTypeToLibary(DeclarationCode typeDeclaration);
+}
+
+/// The api used by [DeclarationMacro]s to contribute new (non-type)
+/// declarations to the current library.
+///
+/// Can also be used to do subtype checks on types.
+abstract class DeclarationBuilder implements Builder {
+  /// Adds a new regular declaration to the surrounding library.
+  ///
+  /// Note that type declarations are not supported.
+  void addToLibrary(DeclarationCode declaration);
+
+  /// Returns true if [leftType] is a subtype of [rightType].
+  bool isSubtypeOf(TypeAnnotation leftType, TypeAnnotation rightType);
+
+  /// Retruns true if [leftType] is an identical type to [rightType].
+  bool isExactly(TypeAnnotation leftType, TypeAnnotation rightType);
+}
+
+/// The api used by [DeclarationMacro]s to contribute new members to a class.
+abstract class ClassMemberDeclarationBuilder implements DeclarationBuilder {
+  /// Adds a new declaration to the surrounding class.
+  void addToClass(DeclarationCode declaration);
+}
+
+/// The api used to introspect on a [ClassDeclaration].
+abstract class ClassIntrospector {
+  /// The fields available for [clazz].
+  ///
+  /// This may be incomplete if in the declaration phase and additional macros
+  /// are going to run on this class.
+  Stream<FieldDeclaration> fieldsOf(ClassDeclaration clazz);
+
+  /// The methods available so far for the current class.
+  ///
+  /// This may be incomplete if additional declaration macros are running on
+  /// this class.
+  Stream<MethodDeclaration> methodsOf(ClassDeclaration clazz);
+
+  /// The constructors available so far for the current class.
+  ///
+  /// This may be incomplete if additional declaration macros are running on
+  /// this class.
+  Stream<ConstructorDeclaration> constructorsOf(ClassDeclaration clazz);
+
+  /// The class that is directly extended via an `extends` clause.
+  Future<ClassDeclaration?> superclassOf(ClassDeclaration clazz);
+
+  /// All of the classes that are mixed in with `with` clauses.
+  Stream<ClassDeclaration> mixinsOf(ClassDeclaration clazz);
+}
+
+/// The api used by [DeclarationMacro]s to reflect on the currently available
+/// members, superclass, and mixins for a given [ClassDeclaration]
+abstract class ClassDeclarationBuilder
+    implements ClassMemberDeclarationBuilder, ClassIntrospector {}
+
+/// The api used by [DefinitionMacro] to get a [TypeDeclaration] for any given
+/// [TypeAnnotation].
+abstract class TypeIntrospector {
+  Future<TypeDeclaration> typeDeclarationOf(TypeAnnotation annotation);
+}
+
+/// The base class for builders in the definition phase. These can convert
+/// any [TypeAnnotation] into its corresponding [TypeDeclaration], and also
+/// reflect more deeply on those.
+abstract class DefinitionBuilder
+    implements Builder, ClassIntrospector, TypeIntrospector {}
+
+/// The apis used by [DefinitionMacro]s to define the body of a constructor
+/// or wrap the body of an existing constructor with additional statements.
+abstract class ConstructorDefinitionBuilder implements DefinitionBuilder {
+  /// Augments an existing constructor body with [body].
+  ///
+  /// TODO: Link the library augmentations proposal to describe the semantics.
+  void augment({FunctionBodyCode? body, List<Code>? initializers});
+}
+
+/// The apis used by [DefinitionMacro]s to augment functions or methods.
+abstract class FunctionDefinitionBuilder implements DefinitionBuilder {
+  /// Augments the function.
+  ///
+  /// TODO: Link the library augmentations proposal to describe the semantics.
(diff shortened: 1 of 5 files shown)


working/macros/api/builders.dart
```diff
@@ -0,0 +1,115 @@
...
+  /// This can be used to emit a reference to it in generated code, or do
+  /// subtype checks (depending on support in the current phase).
+  TypeAnnotation typeAnnotationOf<T>();
+}
+
+/// The api used by [TypeMacro]s to contribute new type declarations to the
+/// current library, and get [TypeAnnotation]s from runtime [Type] objects.
+abstract class TypeBuilder implements Builder {
+  /// Adds a new type declaration to the surrounding library.
+  void addTypeToLibary(DeclarationCode typeDeclaration);
```
munificent, line 18: Bikeshed: Maybe `declareType()`?

working/macros/api/builders.dart
```diff
@@ -0,0 +1,115 @@
...
+}
+
+/// The api used by [DeclarationMacro]s to contribute new (non-type)
+/// declarations to the current library.
+///
+/// Can also be used to do subtype checks on types.
+abstract class DeclarationBuilder implements Builder {
+  /// Adds a new regular declaration to the surrounding library.
+  ///
+  /// Note that type declarations are not supported.
```
munificent, line 28: What does "type declaration" mean here?

working/macros/api/builders.dart
```diff
@@ -0,0 +1,115 @@
...
+
+/// The api used by [Macro]s to reflect on the currently available
+/// members, superclass, and mixins for a given [ClassDeclaration]
+abstract class ClassDeclarationBuilder
+    implements ClassMemberDeclarationBuilder, ClassIntrospector {}
+
+/// The api used by [Macro] to get a [TypeDeclaration] for any given
+/// [TypeAnnotation].
+abstract class TypeIntrospector {
+  Future<TypeDeclaration> typeDeclarationOf(TypeAnnotation annotation);
```
munificent, line 79: Maybe `resolve()`?

Also, we'll have to figure out how this works with non-nominal types like function types and `FutureOr<T>`. Maybe just leave a TODO.

working/macros/api/code.dart
```diff
@@ -0,0 +1,129 @@
+/// A scope in which to resolve a chunk of code.
+///
+/// TODO: Handle more deeply nested scopes (such as a scope specific to a class
+/// or method body).
+class Scope {
+  /// Identifiers should be resolved as if they existed in this library.
+  final Uri libraryUri;
+
+  Scope(this.libraryUri);
```
munificent, line 9: I don't think users should be able to create a Scope for an arbitrary URI. Instead, they should only be able to get access to a couple of existing Scope instances through the macro API: The top-level Scope where the macro is declared and maybe one where the macro is applied.

working/macros/api/code.dart
```diff
@@ -0,0 +1,129 @@
...
+}
+
+/// The base class representing an arbitrary chunk of Dart code, which may or
+/// may not be syntacically or semantically valid yet.
```
munificent, line 13: Nit: "syntactically".

working/macros/api/macros.dart
```diff
@@ -0,0 +1,82 @@
+import 'dart:async';
+
+import 'builders.dart';
+import 'introspection.dart';
+
+/// The marker interface for all types of macros.
+abstract class Macro {
+  const Macro();
+
+  external void buildTypes(FutureOr<void> Function(TypeBuilder) callback);
```
munificent, line 10: `external`? Why not just abstract?

working/macros/example/data_class.dart
```diff
@@ -0,0 +1,241 @@
...
+      ].joinAsCode(' ^ ');
+
+      builder.addToClass(DeclarationCode.fromParts([
+        '''
+@override
+int get hashCode =>''',
+        ...hashCodeExprs,
+        ';',
+      ]));
+    });
```
munificent, line 160: These look like they are both introspecting on members and declaring members in the same phase. Is that a problem? What if they need to see the declarations of fields produced by other macros? Would something like this work?

```dart
class _HashCode extends ClassMacro {
  const _HashCode();

  @override
  void visitClass(ClassDeclaration clazz) {
    // The visitClass() method body runs in phase 2...
    clazz.addMethod('@override int get hashCode', (builder)  async {
      // ...and then the macro engine defers running the callback (which has access to the builder and
      // instrospection) until phase 3:
      var hashCodeExprs = [
        await for (var field in builder.allFields(builder))
          ExpressionCode.fromString('${field.name}.hashCode')
      ].joinAsCode(' ^ ');

      return DeclarationCode.fromParts([...hashCodeExprs, ';']);
    });
  }
}
```

munificent (approved):
I'm still digesting this, but overall it looks really good to me. :)

working/macros/api/builders.dart
jakemac53: Any declaration that defines a new type (class, enum, type alias, etc)

working/macros/api/code.dart
jakemac53: Ya, I think ultimately I do agree with you. I wanted to expose this so that macros could add references to things they don't directly import - but on second that thought would create weird situations with dependencies, especially in google3 but even in pub as well. It could in theory add a reference to something that isn't depended on at all, forcing users to explicitly add their own dependency on something they don't use.

working/macros/api/macros.dart
jakemac53: A few reasons:

- It makes it so that the example macros are valid without actually implementing these.
- I actually do expect the implementations to be external (filled in with patch files). They probably won't be identical across analyzer/cfe/etc.

jakemac53: I could instead hang these types off of another interface, and pass a 2nd argument to the `visit` methods which is of that type.

working/macros/example/data_class.dart
jakemac53: > These look like they are both introspecting on members and declaring members in the same phase. Is that a problem?

That is a good point, for the ones where the signature doesn't also depend on the fields, I should split this up.

For things like `CopyWith` and `AutoConstructor` they have to use the fields to declare their signatures though. 

> ```dart
> clazz.addMethod('@override int get hashCode', (builder)  async {
> ```

I don't want to hang things off of the `ClassDeclaration` - you can get one of those for anything in the program.

I did consider an api like this, but you may also want to provide a body for something you didn't define, (or more commonly wrap something you didn't define). So we would have to provide two different apis for doing that. So that is why I went with the separate api here.

working/macros/api/builders.dart
jakemac53: Done, I also renamed some other things to follow that naming a bit more, `addToLibrary => declareInLibrary` and `addToClass => declareInClass`.

jakemac53:
Ok I updated this based on feedback. I ended up adding the new `ClassDefinitionBuilder` which lets you get builders for its own fields/methods/constructors, to support defining a declaration later on. It is a bit clunky, but works.

One thing in particular that is weird right now in general is how fields are treated compared to explicit getters/setters and regular methods. Right now I treat explicit getters/setters as methods, but then the naming gets weird, because you have two "methods" with the same name. In `dart:mirrors` we just say the setter has a trailing `=` on the name, and we could do that here, but it feels pretty hacky.

working/macros/api/macros.dart
munificent: Ah, I get it. So this is a concrete method that macro implementations can call. The question then is where to put that method. I think there's basically two options:

1. Put it on a class that we pass an instance of to the macro when it runs. Sort of how the builder API works.
2. Put it on the macro class itself so that the user's macro class can call it on itself.

If we're going to do the latter, then I think the natural way to model it is to make `Macro` be an abstract base class (or mixin) that concrete macro classes extend. Then `buildTypes()` can be a normal concrete instance method. If Dart had protected methods, it would probably be protected (and non-virtual).

I'm not sure which is the best approach, but if we do want this to be a method on the macro itself, then I wouldn't make it an `external` method on a `Macro` interface that gets magically implemented. That's pretty weird for users.

It's probably a little safer to take option (1) and hang it off a parameter that we pass to the macro.

jakemac53: > It's probably a little safer to take option (1) and hang it off a parameter that we pass to the macro.

Ya, I agree. It would be very bizarre for somebody to override these methods haha.

I can work on moving these to other interfaces.

jakemac53: Ok I did this - there are now a _lot_ of builder interfaces though. And the naming gets interesting (I ended up calling them all `builder` and shadowing local variables, to avoid verbosity in the names, but its gross).

jakemac53:
Landing this as is for now - we may want to revisit adding an optional argument to the `addDeclaration*` and `addType` methods, that allows you to pass the closure to fill them in more directly.

I didn't do that for now because its a bit harder than originally anticipated because at least for `addDeclaration*` we can't put a very good type on the builder argument - because we don't know the type of declaration being added.

We could split these out into `addField*`, `addMethod*` etc though? But its a bit larger discussion.

PR #1986, opened 2021-11-23
Revert back to original macro api style

jakemac53:
Opening for discussion.

This PR moves us back towards the original proposal, where there is a different type to implement for each phase, and each declaration type you can run on.

**Advantages**:

- It is easier for an implementation to know up front exactly which phases a given macro will run in.
- It allows an implementation to do more efficient invalidation safely - since macros can't share state across phases with local variables you are free to run only a specific phase of a given macro.
- Macros are more composable - you have per-phase functions you can directly delegate work to, instead of only a more general function.
- We don't have to worry about  users trying to share objects they shouldn't through local variables (`TypeBuilders` etc).
- Less nesting, avoids the "callback" style which can be a bit messy.

**Disadvantages**:

- You have to delegate each phase explicitly for any macro you are composing with (see changes to the example).
- There is a disconnect between where you declare something and where you define it.

I also added the missing `VariableDefinition` and `VariableDefinitionBuilder` classes for introspecting/augmenting top level variables.

### DIFF
--- a/working/macros/api/introspection.dart
+++ b/working/macros/api/introspection.dart
@@ -106,22 +106,25 @@ abstract class ConstructorDeclaration implements MethodDeclaration {
   bool get isFactory;
 }
 
-/// Field introspection information ..
-abstract class FieldDeclaration implements Declaration {
+/// Cariable introspection information.
+abstract class VariableDeclaration implements Declaration {
   /// Whether this function has an `abstract` modifier.
   bool get isAbstract;
 
   /// Whether this function has an `external` modifier.
   bool get isExternal;
 
-  /// Type type of this field.
+  /// The type of this field.
   TypeAnnotation get type;
-
-  /// The class that defines this method.
-  TypeAnnotation get definingClass;
 
   /// A [Code] object representing the initializer for this field, if present.
   Code? get initializer;
+}
+
+/// Field introspection information ..
+abstract class FieldDeclaration implements VariableDeclaration {
+  /// The class that defines this method.
+  TypeAnnotation get definingClass;
 }
 
 /// Parameter introspection information.
--- a/working/macros/api/introspection.dart
+++ b/working/macros/api/introspection.dart
@@ -106,7 +106,7 @@ abstract class ConstructorDeclaration implements MethodDeclaration {
   bool get isFactory;
 }
 
-/// Cariable introspection information.
+/// Variable introspection information.
 abstract class VariableDeclaration implements Declaration {
   /// Whether this function has an `abstract` modifier.
   bool get isAbstract;
--- a/working/macros/example/data_class.dart
+++ b/working/macros/example/data_class.dart
@@ -160,17 +160,16 @@ external int get hashCode;'''));
   @override
   Future<void> buildDefinitionForClass(
       ClassDeclaration clazz, ClassDefinitionBuilder builder) async {
-    await builder.buildMethod('hashCode', (builder) async {
-      var hashCodeExprs = [
-        await for (var field in clazz.allFields(builder))
-          ExpressionCode.fromString('${field.name}.hashCode')
-      ].joinAsCode(' ^ ');
-      builder.augment(FunctionBodyCode.fromParts([
-        ' => ',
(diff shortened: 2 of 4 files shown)


working/macros/api/introspection.dart
```diff
@@ -106,24 +106,27 @@ abstract class ConstructorDeclaration implements MethodDeclaration {
   bool get isFactory;
 }
 
-/// Field introspection information ..
-abstract class FieldDeclaration implements Declaration {
+/// Cariable introspection information.
```
munificent, line 109: "Variable"

working/macros/example/data_class.dart
```diff
@@ -5,225 +5,235 @@ import '../api/code.dart';
...
-        builder.augment(FunctionBodyCode.fromParts([
-          ' => ',
-          ...hashCodeExprs,
-          ';',
-        ]));
-      });
+  @override
+  Future<void> buildDefinitionForClass(
+      ClassDeclaration clazz, ClassDefinitionBuilder builder) async {
+    await builder.buildMethod('hashCode', (builder) async {
```
munificent, line 163: Since the API is explicitly phased, we don't need the callbacks anymore, do we? At this point, it seems like all it does is give you a FunctionDefinitionBuilder for the right method. But a simpler API for that is something like:

```dart
@override
Future<void> buildDefinitionForClass(
    ClassDeclaration clazz, ClassDefinitionBuilder builder) async {
  var functionBuilder = builder.buildMethod('hashCode');
  var hashCodeExprs = [
    await for (var field in clazz.allFields(builder))
      ExpressionCode.fromString('${field.name}.hashCode')
  ].joinAsCode(' ^ ');
  functionBuilder.augment(FunctionBodyCode.fromParts([
    ' => ',
    ...hashCodeExprs,
    ';',
  ]));
}
```

working/macros/example/data_class.dart
```diff
@@ -5,225 +5,235 @@ import '../api/code.dart';
...
 
   @override
-  void visitClass(ClassDeclaration clazz, ClassContext context) {
-    context.buildDeclarations((builder) {
-      builder.declareInClass(DeclarationCode.fromString('''
+  void buildDeclarationsForClass(
+      ClassDeclaration clazz, ClassDeclarationBuilder builder) {
+    builder.declareInClass(DeclarationCode.fromString('''
 @override
 external int get hashCode;'''));
```
munificent, line 157: Say a user wrote:

```dart
class _HashCode implements ClassDeclarationsMacro {
  const _HashCode();

  @override
  void buildDeclarationsForClass(
      ClassDeclaration clazz, ClassDeclarationBuilder builder) {
    var hashCodeExprs = [
      await for (var field in clazz.allFields(builder))
        ExpressionCode.fromString('${field.name}.hashCode')
    ].joinAsCode(' ^ ');

    builder.declareInClass(DeclarationCode.fromParts([
      '@override int get hashCode =>',
      ...hashCodeExprs,
      ';']));
  }
}
```

So they both declare and define the method all in the declaration phase. The API will let them do this, and it will work correctly in cases where there are no other macros adding fields. But if that isn't the case this simpler code will silently do the wrong thing. Do I have that right?

If we going to foist the complexity of explicit phased interfaces onto users, it would be really nice to at least protect them from misusing the phases. I'm not necessarily saying we should go back to an explicitly phased introspection API because that was a *lot* of classes. But this feels like a footgun to me.

munificent (approved):
I wouldn't say I'm *thrilled* by having to have `phases × application_target` different macro interfaces, but it does seem like trying to do the indirect context stuff didn't actually buy us much simplicity either. I think this is probably a step in the right direction.

working/macros/example/data_class.dart
jakemac53: Yes true, ill switch to that.

working/macros/example/data_class.dart
jakemac53: Ya I think we should continue discussion on this but I am on the fence right now. It is both more convenient and more efficient to declare the body as well, if you have enough information to do so.

PR #1990, opened 2021-11-24
Address final comments around FinalizationRegistry

mraleph:
* Rename `[Native]FinalizationRegistry` to `[Native]Finalizer`
* Rename `WeakRef` to `WeakReference`
* Remove `WeakMap` typedef
* Make `NativeFinalizer` not inherit from `Finalizer`.
* Move proposal to accepted folder.

### DIFF
--- a/accepted/1847 - FinalizationRegistry/proposal.md
+++ b/accepted/1847 - FinalizationRegistry/proposal.md
@@ -1,4 +1,4 @@
-# Proposal for adding `FinalizationRegistry` and `WeakRef` to core libraries.
+# Proposal for adding `Finalizer` and `WeakReference` to core libraries.
 
 Author: vegorov@google.com<br>Version: 0.1
 
@@ -6,9 +6,8 @@ Author: vegorov@google.com<br>Version: 0.1
 
 This proposal includes two interconnected pieces of functionality:
 
-- an ability to create _weak references_ (`WeakRef`);
-- an ability to associate finalization actions with objects
-  (`FinalizationRegistry`).
+- an ability to create _weak references_ (`WeakReference`);
+- an ability to associate finalization actions with objects (`Finalizer`).
 
 ### Weak references
 
@@ -132,14 +131,14 @@ would release the native resource.
 
 We split the proposed finalization API into two parts:
 
-- `FinalizationRegistry` which:
+- `Finalizer` which:
   - does not provide strong guarantees around promptness of finalization,
   - does not impose any restrictions on objects you could associate a
     finalization action with,
   - does not impose any restrictions on finalization actions and invokes them
     asynchronously,
   - is implementable on the Web.
-- `NativeFinalizationRegistry` which:
+- `NativeFinalizer` which:
   - provides stronger guarantees around promptness of finalization,
   - guarantees that finalization actions are invoked when isolate is shutting
   down,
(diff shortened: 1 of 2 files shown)


mraleph:
PTAL @lrhn

accepted/1847 - FinalizationRegistry/proposal.md
```diff
@@ -247,15 +238,27 @@ abstract class Finalizable {
...
+/// becomes inaccessible to the program, then any number of those finalizers
+/// may trigger their associated callback. It will not necessarily be all or
+/// none of them.
+///
+/// Finalization callbacks will happen as early as possible, without waiting
+/// for control to return to the event loop.
+abstract class NativeFinalizer<T> {
+  /// Creates a finalizer with the given finalization callback.
   ///
   /// Note: [callback] is expected to be a native function which can be
```
lrhn, line 263: `[callback]` -> `The [callback]`

(because I don't like starting sentences with a lower-case letter, not even if it's code).

accepted/1847 - FinalizationRegistry/proposal.md
```diff
@@ -247,15 +238,27 @@ abstract class Finalizable {
 typedef NativeFinalizer = Void Function(Pointer<Void>);
 typedef NativeFinalizerPtr = Pointer<NativeFunction<NativeFinalizer>>
 
-/// [FinalizationRegistry] which will execute its finalizers as early as
-/// possible without waiting for control to return to the event loop.
+/// [Finalizer] *may* call the given callback when objects it is attached to
```
lrhn, line 241: Not a good first line. Describe the instances of this class, using noun phrases.
This line is describing a completely different class.
In some cases (doc overview sections), this is the *only* line you see describing the class.

> /// A native finalizer which can be attached to Dart objects.

Then describe when and how to use it:

> ///
> /// When [attach]ed to a Dart object, this finalizer's native callback is called
> /// after the Dart object is garbage collected or becomes inaccessible for other
> /// reasons.
> ///
> /// Callbacks will happen as early as possible, when the object becomes 
> /// inaccessible to the program,
> /// and may happen at any moment during execution of the program.
> /// At the latest, when an isolate group shuts down, 
> /// this callback is guaranteed to be called for each object in that isolate group
> /// that the finalizer is still attached to.
> ///
> /// If a callback is no longer desired, the [detach] method can remove 
> /// any number of attachments to objects by providing the detachment key 
> /// used when the finalizer was attached to those objects.

Then add perspective:

> /// 
> /// Compared to the [Finalizer] from `dart:core`,
> /// which makes no promises to ever call an attached `callback`,
> /// this native finalizer promises that all attached finalizers are 
> /// definitely called at least once before the program ends,
> /// and the callbacks are called as soon as possible after an object 
> /// is recognized as inaccessible.

(The reader should be able to stop at any time and have gotten the most important parts of the documentation.)

accepted/1847 - FinalizationRegistry/proposal.md
```diff
@@ -247,15 +238,27 @@ abstract class Finalizable {
...
-/// this finalization registry is shutting down.
-abstract class NativeFinalizationRegistry<F extends Finalizable>
-    extends FinalizationRegistry<Pointer> {
-  /// Creates a finalization registry with the given finalization
-  /// callback.
+/// No other promises are made that the callback will ever be called,
+/// only that *if* it is called with a finalization token as argument,
+/// at least one object registered in the registry with that finalization token
+/// is no longer accessible to the program.
+///
```
lrhn, line 251: Is the "no other promises" true for the native finalizer?

accepted/1847 - FinalizationRegistry/proposal.md
```diff
@@ -247,15 +238,27 @@ abstract class Finalizable {
...
+/// If multiple finalizers has been attached to an object or the same
+/// finalizer has been attached multiple times to an object and that object
+/// becomes inaccessible to the program, then any number of those finalizers
+/// may trigger their associated callback. It will not necessarily be all or
+/// none of them.
```
lrhn, line 256: Is the "not all or none" true, won't it be "all"?

accepted/1847 - FinalizationRegistry/proposal.md
```diff
@@ -265,37 +268,60 @@ abstract class NativeFinalizationRegistry<F extends Finalizable>
   ///
   /// [callback] might be invoked on an arbitrary thread and not necessary
```
lrhn, line 269: > The \[callback\]

accepted/1847 - FinalizationRegistry/proposal.md
```diff
@@ -265,37 +268,60 @@ abstract class NativeFinalizationRegistry<F extends Finalizable>
...
-  /// Same as [super.register] but allows to specify an [externalSize] to
-  /// guide GC heuristics.
-  void register(covariant F value,
-                Pointer finalizationToken,
-                {Object? unregisterToken, int externalSize});
+  /// Attaches this finalizer to the given [value].
+  ///
+  /// When [value] is no longer accessible to the program,
+  /// the registry *may* call its callback function with [finalizationToken]
+  /// as argument.
```
lrhn, line 277: May or will?

lrhn (approved):

accepted/1847 - FinalizationRegistry/proposal.md
```diff
@@ -132,75 +131,69 @@ would release the native resource.
...
-/// callback. It will not necessarily be all or none of them.
+/// If multiple finalizers has been attached to an object or the same
+/// finalizer has been attached multiple times to an object and that object
+/// becomes inaccessible to the program, then any number of those finalizers
+/// may trigger their associated callback. It will not necessarily be all or
+/// none of them.
 ///
 /// Finalization callbacks will happen as *events*, not during execution of
 /// other code and not as a microtask, but as high-level events similar to
 /// timer events.
```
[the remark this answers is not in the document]
mraleph: > Maybe add: "The callback should not throw." 

Done.

> Is the callback bound to a zone when the finalizer is created?

Added (to the constructor doc comment) that the callback is going to be bound to the current Zone. 

Note that even in `dart:async` we don't actually document binding of callbacks (e.g. neither `Future.then` nor `Stream.listen` say much about `Zone`s).

accepted/1847 - FinalizationRegistry/proposal.md
```diff
@@ -132,75 +131,69 @@ would release the native resource.
...
-abstract class FinalizationRegistry<FT> {
-  /// Creates a finalization registry with the given finalization callback.
-  external factory FinalizationRegistry(
-    void Function(FT finalizationToken) callback);
+abstract class Finalizer<T> {
+  /// Creates a finalizer with the given finalization callback.
```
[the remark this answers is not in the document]
mraleph: Done.

accepted/1847 - FinalizationRegistry/proposal.md
```diff
@@ -132,75 +131,69 @@ would release the native resource.
...
-  /// The callback may be called at most once per registration, and not
-  /// for registrations which have been unregistered since they were registered.
-  void register(Object value, FT finalizationToken, {Object? unregisterToken});
+  /// The callback may be called at most once per attachment, and not
+  /// for registrations which have been detached since they were attached.
+  void attach(Object value, T token, {Object? detachKey});
 
-  /// Unregisters any finalization callbacks registered with [unregisterToken]
-  /// as unregister-token.
+  /// Detaches the finalizer from any objects which used [detachKey] when
```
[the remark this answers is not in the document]
mraleph: > (it's as restrictive clause)

TIL. finally I know the criteria. Thanks!

accepted/1847 - FinalizationRegistry/proposal.md
```diff
@@ -212,23 +205,21 @@ abstract class FinalizationRegistry<FT> {
...
 /// supported as an [Expando] key.
-abstract class WeakRef<T extends Object> {
-  /// Create a [WeakRef] pointing to the given [target].
-  /// 
+abstract class WeakReference<T extends Object> {
+  /// Create a [WeakReference] pointing to the given [target].
+  ///
   /// The [target] must be an object supported as an [Expando] key.
-  external factory WeakRef(T target);
+  external factory WeakReference(T target);
```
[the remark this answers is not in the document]
mraleph: This reads a bit strange, cause those are the types which are *not* supported by Expando as keys. 

Instead added 

```
/// Which means [target] can not be a number, a string, a boolean, or the `null` value
```

accepted/1847 - FinalizationRegistry/proposal.md
mraleph: Applied. 

Though I find it a bit redundant to document class called `NativeFinalizer` as `A native finalizer which can be attached to Dart objects.`. It's all in the name already.

accepted/1847 - FinalizationRegistry/proposal.md
mraleph: We can't promise that GC ever happens. 

I have removed most of this text and replaced it with your variant which is actually much better.

accepted/1847 - FinalizationRegistry/proposal.md
mraleph: Removed this text.

accepted/1847 - FinalizationRegistry/proposal.md
mraleph: Fixed.

mraleph (commented):
Applied your suggestions, thanks!

mraleph:
@lrhn let me know if you would to take another look or I should just land it.

lrhn:
Ship it. 
(I'll probably look at the final CL for the implementation anyway, with fresher eyes, and catch anything missing at that point.)

PR #2003, opened 2021-12-02
clean up some out of date portions of the macro spec

jakemac53:
This was still pointing at my old repo instead of the new api, and referred to some older apis :).

Removes all references to the prototype repo.


munificent (approved):
Looks great!

PR #2004, opened 2021-12-02
add `interfacesOf` method to `ClassIntrospector`, and some cleanup

jakemac53:
I had originally thought we allowed cycles in the class hierarchy when it comes to interfaces, but it doesn't look like we do, so we can still have a well defined ordering here (which was the only reason I left this off prior).

Also removed the section about adding macro applications to existing declarations. You should be able to just directly invoke the macros inside your macro instead.


munificent (approved):
👍

PR #2008, opened 2021-12-02
add macro keyword

jakemac53:
I went with just `macro NameOfYourMacro` and not `macro class NameOfYourMacro`. Sort of like how `mixin` isn't `mixin class`.

I didn't write a full grammar for this, but could. It's just the class grammar with a different keyword and no `abstract` modifier allowed (we only want to identify concrete macro definitions here).

### DIFF
--- a/working/macros/feature-specification.md
+++ b/working/macros/feature-specification.md
@@ -293,16 +293,32 @@ introspection APIs. These macros can fully introspect on any type reachable from
 the declarations they are applied to, including introspecting on members of
 classes, etc.
 
-## Macro definitions
+## Macro declarations
 
-Macro definitions do not have any unique syntax or language features. They are
-regular Dart class declarations. They are macros by virtue of the fact that they
-implement one or more special "macro" interfaces defined by the Dart core
-libraries.
+Macros are a special type of class, which use the `macro` keyword in place of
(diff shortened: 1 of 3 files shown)


johnniwinther (approved):

working/macros/feature-specification.md
```diff
@@ -293,17 +293,33 @@ introspection APIs. These macros can fully introspect on any type reachable from
...
-## Macro definitions
+## Macro declarations
 
-Macro definitions do not have any unique syntax or language features. They are
-regular Dart class declarations. They are macros by virtue of the fact that they
-implement one or more special "macro" interfaces defined by the Dart core
-libraries.
+Macros are a special type of class, which use the `macro` keyword in place of
+the `class` keyword, and have some additional limitations that classes don't
+have.
```
munificent, line 300: Bikeshed: I think I'd lean towards `macro class` instead of just `macro`. That reinforces that it *is* a class with all of the typical affordances a user expects from a class. And it leaves the door open in the future for non-class uses of `macro`.

munificent (approved):

working/macros/feature-specification.md
jakemac53: I went with that, I think its fine. Its a fair point we may want to allow a macro function or something like that later on.

PR #2011, opened 2021-12-06
Fix typos in finalization registry

dcharkes:


dcharkes:
@mraleph The `T` in `NativeFinalizer` is not connected to the `Pointer<Void>` in

```dart
typedef NativeFinalizer = Void Function(Pointer<Void>);
typedef NativeFinalizerPtr = Pointer<NativeFunction<NativeFinalizer>>
```

We should either (1) remove `T` and change the type of `token` to `Pointer<Void>`, or (2) remove the typedefs and use `Pointer<NativeFunction<Void Function(Pointer<T>)>>` and make `T` lower bounded by `NativeType`. That would enable users to kind of type check on which `NativeFinalizer` they are trying to call `attach`.

accepted/future-releases/1847 - FinalizationRegistry/proposal.md
```diff
@@ -301,7 +301,7 @@ abstract class NativeFinalizer<T> {
   /// [externalSize] is an amount of native (non-Dart) memory owned by the
   /// given [value]. This information is used to drive garbage collection
   /// scheduling heuristics.
-  void attach(Object value, T token, {Object? detachKey, int externalSize}});
+  void attach(Finalizable value, Pointer<Void> token, {Object? detachKey, int externalSize}});
```
mraleph, line 304: Good catch!

mraleph (approved):

PR #2013, opened 2021-12-06
Api cleanup: remove Scope class and allow for unnamed types 

jakemac53:
- Drops the Scope class
- Drops the field from the `Code` class and subclasses
- Made `TypeAnnotation` a base class with two subtypes `FunctionTypeAnnotation` and `NamedTypeAnnotation`
- Also added a `code` getter to `TypeAnnotation` which converts it to a `Code` object.
  - Is this actually feasible to produce always? Should this api move to a `builder` class?

### DIFF
--- a/working/macros/api/builders.dart
+++ b/working/macros/api/builders.dart
@@ -83,9 +83,8 @@ abstract class ClassDeclarationBuilder
 /// The api used by [Macro] to get a [TypeDeclaration] for any given
 /// [TypeAnnotation].
 abstract class TypeIntrospector {
-  /// TODO: Figure out how to deal with `FutureOr<T>`, function types, and
-  /// other non-nominal types.
-  Future<TypeDeclaration> resolve(TypeAnnotation annotation);
+  /// Resolves a [NamedTypeAnnotation] to its declaration.
+  Future<TypeDeclaration> resolve(NamedTypeAnnotation annotation);
 }
 
 /// The base class for builders in the definition phase. These can convert
--- a/working/macros/api/code.dart
+++ b/working/macros/api/code.dart
@@ -1,51 +1,36 @@
-/// A scope in which to resolve a chunk of code.
-///
-/// TODO: Specify more what these mean, what fields are available (if any), etc.
-abstract class Scope {}
-
 /// The base class representing an arbitrary chunk of Dart code, which may or
 /// may not be syntactically or semantically valid yet.
 class Code {
-  /// The scope in which to resolve anything from [parts] that does not have its
-  /// own scope already defined.
-  final Scope? scope;
-
   /// All the chunks of [Code] or raw [String]s that comprise this [Code]
   /// object.
   final List<Object> parts;
 
-  Code.fromString(String code, {this.scope}) : parts = [code];
+  Code.fromString(String code) : parts = [code];
 
-  Code.fromParts(this.parts, {this.scope});
+  Code.fromParts(this.parts);
 }
 
 /// A piece of code representing a syntactically valid declaration.
 class DeclarationCode extends Code {
-  DeclarationCode.fromString(String code, {Scope? scope})
-      : super.fromString(code, scope: scope);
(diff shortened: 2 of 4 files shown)


working/macros/api/builders.dart
```diff
@@ -83,9 +83,8 @@ abstract class ClassDeclarationBuilder
 /// The api used by [Macro] to get a [TypeDeclaration] for any given
 /// [TypeAnnotation].
 abstract class TypeIntrospector {
-  /// TODO: Figure out how to deal with `FutureOr<T>`, function types, and
-  /// other non-nominal types.
-  Future<TypeDeclaration> resolve(TypeAnnotation annotation);
+  /// Resolves a [NamedTypeAnnotation] to its declaration.
+  Future<TypeDeclaration> resolve(NamedTypeAnnotation annotation);
```
munificent, line 87: You'll want to keep this accepting TypeAnnotation. We'll end up with two forms for all the structural types:

* `FunctionTypeAnnotation` has fields for the parameter and return types that are of type `TypeAnnotation`. It's an unresolved function type.

* `FunctionType` (better name?) has fields for the parameter and return types of type `TypeDeclaration`. This is a resolved function type. It's still a function type—it's still structural and doesn't refer to a named declaration—but the types within it have been resolved now.

We'll have a similar pair of types for `FutureOrTypeAnnotation` and `FutureOrType`. Maybe `NullableTypeAnnotation` and `NullableType`?

working/macros/api/code.dart
```diff
@@ -1,51 +1,36 @@
-/// A scope in which to resolve a chunk of code.
-///
-/// TODO: Specify more what these mean, what fields are available (if any), etc.
-abstract class Scope {}
```
munificent, line 4: I do think we'll probably still want this as a way to create a new identifier that explicitly resolves in a given scope. That way you can create code that refers to declarations that are themselves produced by a macro.

The old doc I wrote about the Code API sketches it out a little. But you can drop it for now if you want.

Removing the Scope parameters to all of the other APIs LGTM.

working/macros/api/introspection.dart
```diff
@@ -128,10 +139,7 @@ abstract class FieldDeclaration implements VariableDeclaration {
 }
 
 /// Parameter introspection information.
-abstract class ParameterDeclaration {
-  /// The name of the parameter.
-  String get name;
-
+abstract class ParameterDeclaration implements Declaration {
```
munificent, line 142: No more name? Is it inherited from Declaration?

What about required-ness and default values?

munificent (approved):

working/macros/api/builders.dart
jakemac53: The way I was thinking about this is for the `FunctionTypeAnnotation` you would just individually resolve the `TypeAnnotation`s within it - and not resolve the entire thing at once.

working/macros/api/code.dart
jakemac53: I think removing it for now is good until we have a better understanding of how it will be used, especially with the context of augmentation libraries.

working/macros/api/introspection.dart
jakemac53: > No more name? Is it inherited from Declaration?

Yes I made it implement `Declaration` instead which has a name, so it gets it from that.

> What about required-ness and default values?

Ya it has those (you need to expand a section below).

working/macros/api/builders.dart
jakemac53: > What if you want to ask a question like "Is this type a subtype of `int Function(Object?)`?"

The api for this today just takes two `TypeAnnotation` objects - which I know seems a bit weird but the api only exists in phase 2 and beyond, where types are resolvable. The `TypeAnnotation` objects basically need to have come from the compiler originally in order for it to be able to know how to resolve it.

We could instead have resolved versions of _all_ the declaration types, which would then have resolved types, and hang the type comparison apis off of those. But it adds a lot more api surface area.

PR #2017, opened 2021-12-07
Update enhanced enum feature specification.

lrhn:
State that `index` and `toString` are inherited from `Enum`.
State that `values` is only added if it would not conflict.

Rephrase the definition in terms of an `enum` declaration
introducing a *class*, not a `class` declaration.
Then put the corresponding class declaration into the
"implementation" section as a desugaring, not a specification.

### DIFF
--- a/accepted/future-releases/enhanced-enums/feature-specification.md
+++ b/accepted/future-releases/enhanced-enums/feature-specification.md
@@ -66,11 +66,11 @@ must redirect to a generative constructor._
 
 ## Semantics
 
-We change the publicly visible API of the  `Enum` class to be declared as:
+The `Enum` class behaves as if it was declared as:
 
 ```dart
 class Enum {
-  // No default constructor (however we enforce that).
+  // No default constructor.
   external int get index;
   external String toString();
 }
@@ -84,14 +84,15 @@ The semantics of such an enum declaration, *E*, is defined as introducing a (sem
 
 * **Name**: The name of the class *C* and its implicit interface is the name of the `enum` declaration.
 
-* **Superclass**: The superclass of *C* is (seemingly) the `Enum` class, with the mixins declared by *E* applied. _(In practice, the superclass may be another system provided class with implements `Enum`.)_
+* **Superclass**: The superclass of *C* is an implementation-specific built-in class *`EnumImpl`*, with the mixins declared by *E* applied. _(The `EnumImpl` class may be the `Enum` class itself or it may be another class which extends or implements `Enum`, but as seen from a non-platform library the interface of *`EnumImpl`* is the same as that of `Enum`, and its methods work as specified for `Enum` )_
 
-  * If *E* is declared as `enum Name with Mixin1, Mixin2 …` then the superclass of *C* is the mixin application `Enum with Mixin1, Mixin2`.
+  * If *E* is declared as `enum Name with Mixin1, Mixin2 …` then the superclass of *C* is the mixin application <Code>*EnumImpl* with Mixin1, Mixin2</code>.
 
   It’s a **compile-time error** if such a mixin application introduces any instance variables. _We need to be able to call an implementation specific superclass `const` constructor of `Enum`, and a mixin application of a mixin with a field does not make its forwarding constructor `const`. Currently that’s the only restriction, but if we add further restrictions on mixin applications having `const` forwarding constructors, those should also apply here._
 
 * **Superinterfaces**: The immediate superinterfaces of *C* are the interface of the superclass and the interfaces declared by *E*.
-  * If `E` is declares as `enum Name with Mixin1, Mixin2 implements Type1, Type2 { … }` then the immediate superinterfaces of *C* are the interfaces of `Name with Mixin1, Mixin2`, `Type1` and `Type2`.
+  
+  * If `E` is declared as `enum Name with Mixin1, Mixin2 implements Type1, Type2 { … }` then the immediate superinterfaces of *C* are the interfaces of `Name with Mixin1, Mixin2`, `Type1` and `Type2`.
 
 - **Declared members**: For each member declaration of the `enum` declaration *E*, the same member is added to the class *C*. This includes constructors (which must be `const` generative or non-`const` factory constructors.)
 
@@ -142,7 +143,7 @@ If the resulting class would have any naming conflicts, or other compile-time er
 
 If not invalid, the semantics denoted by the `enum` declaration is that class, which we’ll refer to as the *corresponding class* of the `enum` declaration. *(We don’t require the implementation to be exactly such a class declaration, there might be other helper classes involved in the implementation, and different private members, but the publicly visible interface and behavior should match.)*
 
-That is, if the corresponding class of an `enum` declaration is valid, the `enum` declaration introduces the public interface* and *type* of the corresponding class. _There are, however, restrictions on how that class and interface can be used, listed in the next section._
+That is, if the corresponding class of an `enum` declaration is valid, the `enum` declaration introduces the *public interface* and *type* of the corresponding class. _There are, however, restrictions on how that class and interface can be used, listed in the next section._
 
 ### Implementing `Enum` and enum types
 
@@ -258,7 +259,7 @@ enum Plain {
 }
 ```
 
-has corresponding class desugaring similar to:
+would have a corresponding class desugaring of:
 
 ```dart
 class Plain extends Enum {
@@ -288,7 +289,7 @@ mixin EnumIndexOrdering<T extends Enum> on Enum implements Comparable<T> {
 }
 ```
 
-has corresponding class desugaring of:
+would have a  corresponding class desugaring of:
 
 ```dart
 class Ordering extends Enum with EnumIndexOrdering<Ordering> {
@@ -304,8 +305,6 @@ class Ordering extends Enum with EnumIndexOrdering<Ordering> {
   String _$enumToString() => "Ordering.${_$name}";
 }
 ```
-
-
 
 ### Complex, one with everything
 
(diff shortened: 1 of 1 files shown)


accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -68,128 +66,175 @@ must redirect to a generative constructor._
...
-The semantics of such an enum declaration is defined as being equivalent to *rewriting into a corresponding class declaration* as follows:
-
-- Declare a class with the same name and type parameters as the `enum` declaration.
-
-- Add `extends Enum`, where `Enum` refers to the type declared in `dart:core`.
+```dart
+class Enum {
+  // No default constructor (however we enforce that).
+  external int get index;
+  external String toString();
```
eernstg, line 75: With respect to the publicly visible API, `toString` may be omitted. We don't have the notion of a "primitive `toString` method", so clients should not have to know.

Maybe we could just say that "The `Enum` class behaves as if it were declared as [lines 72-76]", thus avoiding the direct reference to "publicly visible".

accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -68,128 +66,175 @@ must redirect to a generative constructor._
...
-- Further add the mixins and interfaces of the `enum` declaration.
+We intend to (at least pretend to) let `enum` classes extend `Enum`, and let mixins and members access the default `index` and `toString()` through `super.`. _(In practice, we may use a different class implementing `Enum` as the superclass, but for checking the validity of `super.index`/`super.toString()`, we analyze against `Enum` itself, so it must have non-abstract implementations.)_
+
+This all makes it look as if `Enum` would be a valid superclass for the mixin applications and methods of the enhanced `enum` class.
+
+The semantics of such an enum declaration, *E*, is defined as introducing a (semantic) *class*, *C*, just like a similar `class` declaration.
+
+* **Name**: The name of the class *C* and its implicit interface is the name of the `enum` declaration.
+
+* **Superclass**: The superclass of *C* is (seemingly) the `Enum` class, with the mixins declared by *E* applied. _(In practice, the superclass may be another system provided class with implements `Enum`.)_
```
eernstg, line 87: Can we say something like this?: "The super class of *C* is a built-in class _EnumImpl_. _EnumImpl_ may be `Enum`, or it may be a different class, but the interface of _EnumImpl_ as seen from a non-system library is the same as that of `Enum`, and _EnumImpl_ is a subtype of `Enum`."

accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -68,128 +66,175 @@ must redirect to a generative constructor._
...
+
+  * If *E* is declared as `enum Name with Mixin1, Mixin2 …` then the superclass of *C* is the mixin application `Enum with Mixin1, Mixin2`.
+
+  It’s a **compile-time error** if such a mixin application introduces any instance variables. _We need to be able to call an implementation specific superclass `const` constructor of `Enum`, and a mixin application of a mixin with a field does not make its forwarding constructor `const`. Currently that’s the only restriction, but if we add further restrictions on mixin applications having `const` forwarding constructors, those should also apply here._
+
+* **Superinterfaces**: The immediate superinterfaces of *C* are the interface of the superclass and the interfaces declared by *E*.
+  * If `E` is declares as `enum Name with Mixin1, Mixin2 implements Type1, Type2 { … }` then the immediate superinterfaces of *C* are the interfaces of `Name with Mixin1, Mixin2`, `Type1` and `Type2`.
```
eernstg, line 94:
```suggestion
  * If `E` is declared as `enum Name with Mixin1, Mixin2 implements Type1, Type2 { … }` then the immediate superinterfaces of *C* are the interfaces of `Name with Mixin1, Mixin2`, `Type1` and `Type2`.
```

accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -68,128 +66,175 @@ must redirect to a generative constructor._
...
+
+- **Declared members**: For each member declaration of the `enum` declaration *E*, the same member is added to the class *C*. This includes constructors (which must be `const` generative or non-`const` factory constructors.)
+
+- **Default constructor**: If no generative constructors were declared, and no unnamed factory constructor was added,
+  a default generative constructor is added:
+
+  ```dart
+  const Name();
+  ```
```
eernstg, line 103: So this constructor is assumed to be transformed later on, to accept the two "extra" arguments?

accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -68,128 +66,175 @@ must redirect to a generative constructor._
...
+  
+  all of which is accessible to the `toString` and `index` member of `Enum`, and to the `EnumName.name` extension getter. The values are computed as follows.
+  
+  - `id` &mapsto; `Name()` (no arguments, equivalent to empty argument list)
+  - `id(args)` &mapsto; `Name(args)`
+  - `id<types>(args)` &mapsto; `Name<types>(args)`
+  - `id.named(args)` &mapsto; `Name._$named(args)`
+  - `id<types>.named(args)` &mapsto; `Name<types>._$named(args)`
+  
+  where `args` are considered as occurring in a `const` context, and it’s a **compile-time error** if they are then not compile-time constants.
```
eernstg, line 121: Arguments that occur in a constant context without any `const` in a syntactically enclosing term seems to be a novel thing, but I guess it's OK because it _does_ occur in the desugared declaration.

accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -68,128 +66,175 @@ must redirect to a generative constructor._
...
+- Using a non-constant expression as argument of an enum value.
+- Declaring a static member and inheriting an instance member with the same base-name.
 
-- If no `toString` member overriding `Object.toString` was declared or inherited _(from mixin applications)_,
-  A `String toString() => “Name.${_$name}”;` instance method is added.
+If not invalid, the semantics denoted by the `enum` declaration is that class, which we’ll refer to as the *corresponding class* of the `enum` declaration. *(We don’t require the implementation to be exactly such a class declaration, there might be other helper classes involved in the implementation, and different private members, but the publicly visible interface and behavior should match.)*
 
-- For each `<enumEntry>` with name `id` and index *i* in the comma-separated list of enum entries, a static constant is added as follows:
+That is, if the corresponding class of an `enum` declaration is valid, the `enum` declaration introduces the public interface* and *type* of the corresponding class. _There are, however, restrictions on how that class and interface can be used, listed in the next section._
```
eernstg, line 145:
```suggestion
That is, if the corresponding class of an `enum` declaration is valid, the `enum` declaration introduces the *public interface* and *type* of the corresponding class. _There are, however, restrictions on how that class and interface can be used, listed in the next section._
```

accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -213,7 +258,7 @@ enum Plain {
 }
 ```
 
-has corresponding class declaration:
+has corresponding class desugaring similar to:
```
eernstg, line 261: The terminology used previously was simply that there is "corresponding class", so maybe

```suggestion
has a corresponding class which is approximately the following:
```

accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -222,15 +267,46 @@ class Plain extends Enum {
...
+  few,
+  many;
+}
+
+mixin EnumIndexOrdering<T extends Enum> on Enum implements Comparable<T> {
+  int compareTo(T other) => this.index - other.index;
 }
 ```
 
+has corresponding class desugaring of:
```
eernstg, line 291:
```suggestion
has a corresponding class which is approximately the following:
```

accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -288,7 +364,7 @@ enum Complex<T extends Pattern> with EnumComparable<Complex> implements Pattern
 }
 ```
 
-has corresponding class declaration:
+has corresponding class desugaring of:
```
eernstg, line 367:
```suggestion
has a corresponding class which is approximately the following:
```

accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -347,57 +424,27 @@ enum MySingleton implements Whatever {
 }
 ```
 
-has equivalent class
+has a corresponding class desugaring of:
```
eernstg, line 427:
```suggestion
has a corresponding class which is approximately the following:
```

eernstg (approved):
LGTM, with a handful of comments.

munificent (approved):

accepted/future-releases/enhanced-enums/feature-specification.md
lrhn: ACK.
I'll retain `toString` just to suggest that it overrides `Object.toString`.

accepted/future-releases/enhanced-enums/feature-specification.md
lrhn: Something like that.

accepted/future-releases/enhanced-enums/feature-specification.md
lrhn: Done.

accepted/future-releases/enhanced-enums/feature-specification.md
lrhn: ACK.

accepted/future-releases/enhanced-enums/feature-specification.md
lrhn: Done.

accepted/future-releases/enhanced-enums/feature-specification.md
lrhn: I actually changed the phrasing generally (I hope).
There is no "corresponding class".
The `enum` declaration introduces a (semantic) *class* directly.

The corresponding class desugaring is a *suggested* implementation strategy, but is not normative.

PR #2019, opened 2021-12-07
Document new strict-casts mode

srawlins:


resources/type-system/strict-casts.md
```diff
@@ -0,0 +1,95 @@
...
+  language:
+    strict-casts: true
+```
+
+## Motivation
+
+Static analysis can provide many benefits which require meaningful static
+types. For example, when a value of one type is _not assignable_ to a variable
+of another type, the error is reported by static analysis. However, when a
+value has a static type of `dynamic`, no error is reported; such a value is
```
munificent, line 28: Grammar nit: I'd just use "." instead of ";". I've never found the semicolon to improve readability.

munificent (approved):
Drive-by review. This looks great!

resources/type-system/strict-casts.md
```diff
@@ -0,0 +1,95 @@
...
+void main() {
+  var decoded = jsonDecode(/* some JSON */);
+  if (decoded['foo']) {
+    print('Got foo.');
+  }
+}
+```
+
+The expression used in an if-statement's condition must have a static type of
+`bool`. Without the Strict casts mode, an expression with a static type of
```
leafpetersen, line 43: Strict -> strict

resources/type-system/strict-casts.md
srawlins: I love semicolons, and very probably overuse them :). Done.

resources/type-system/strict-casts.md
srawlins: done

PR #2021, opened 2021-12-08
add initial macro expansion protocol

jakemac53:
This adds a `MacroExecutor` interface which is intended for use by language frontends (cfe, analyzer) in order to load and execute macros, as well as create library augmention files.

This should help more clearly define the role of a frontend when it comes to macros (discovery and introspection), as well as provide a unified api for producing the final library augmentations.

It does not actually write files though, we still need to figure out how we want to coordinate that part, and make sure that the CFE and analyzer can consistently talk about these files.

This also allows for multiple different implementations in terms of how macros are actually loaded and executed, without the frontends having to know about them.

Followup PR here https://github.com/dart-lang/language/pull/2022 which starts an implementation using `IsolateMirror.loadUri`.

### DIFF
--- a/working/macros/api/builders.dart
+++ b/working/macros/api/builders.dart
@@ -6,14 +6,7 @@ import 'macros.dart'; // For dart docs.
 
 /// The base interface used to add declarations to the program as well
 /// as augment existing ones.
-abstract class Builder {
-  /// Used to construct a [TypeAnnotation] to a runtime type available to the
-  /// the macro implementation code.
-  ///
-  /// This can be used to emit a reference to it in generated code, or do
-  /// subtype checks (depending on support in the current phase).
-  TypeAnnotation typeAnnotationOf<T>();
-}
+abstract class Builder {}
 
 /// The api used by [Macro]s to contribute new type declarations to the
 /// current library, and get [TypeAnnotation]s from runtime [Type] objects.
@@ -22,33 +15,24 @@ abstract class TypeBuilder implements Builder {
   void declareType(DeclarationCode typeDeclaration);
 }
 
-/// The interface for checking if a type implements another type.
-abstract class TypeComparator {
-  /// Returns true if [leftType] is a subtype of [rightType].
-  bool isSubtypeOf(TypeAnnotation leftType, TypeAnnotation rightType);
-
-  /// Returns true if [leftType] is an identical type to [rightType].
-  bool isExactly(TypeAnnotation leftType, TypeAnnotation rightType);
-}
-
-/// The api used by [Macro]s to contribute new (non-type)
-/// declarations to the current library.
+/// The interface to resolve a [TypeAnnotation] to a [StaticType].
 ///
-/// Can also be used to do subtype checks on types.
-abstract class DeclarationBuilder implements Builder, TypeComparator {
-  /// Adds a new regular declaration to the surrounding library.
-  ///
-  /// Note that type declarations are not supported.
-  void declareInLibrary(DeclarationCode declaration);
-}
-
-/// The api used by [Macro]s to contribute new members to a class.
-abstract class ClassMemberDeclarationBuilder implements DeclarationBuilder {
-  /// Adds a new declaration to the surrounding class.
-  void declareInClass(DeclarationCode declaration);
+/// The [StaticType]s can be compared against other [StaticType]s to see how
+/// they relate to each other.
+///
+/// This api is only available to the declaration and definition phases of
+/// macro expansion.
+abstract class TypeResolver {
+  /// Resolves [typeAnnotation] to a [StaticType].
+  StaticType resolve(TypeAnnotation typeAnnotation);
 }
 
 /// The api used to introspect on a [ClassDeclaration].
+///
+/// Available in the declaration and definition phases, but limited in the
+/// declaration phase to immediately annotated [ClassDeclaration]s. This is
+/// done by limiting the access to the [TypeDeclarationResolver] to the
+/// definition phase.
 abstract class ClassIntrospector {
   /// The fields available for [clazz].
   ///
@@ -78,23 +62,47 @@ abstract class ClassIntrospector {
   Future<List<ClassDeclaration>> interfacesOf(ClassDeclaration clazz);
 }
 
+/// The api used by [Macro]s to contribute new (non-type)
+/// declarations to the current library.
+///
+/// Can also be used to do subtype checks on types.
+abstract class DeclarationBuilder
+    implements Builder, TypeResolver, ClassIntrospector {
+  /// Adds a new regular declaration to the surrounding library.
+  ///
+  /// Note that type declarations are not supported.
+  void declareInLibrary(DeclarationCode declaration);
+}
+
+/// The api used by [Macro]s to contribute new members to a class.
+abstract class ClassMemberDeclarationBuilder implements DeclarationBuilder {
+  /// Adds a new declaration to the surrounding class.
+  void declareInClass(DeclarationCode declaration);
+}
+
 /// The api used by [Macro]s to reflect on the currently available
 /// members, superclass, and mixins for a given [ClassDeclaration]
 abstract class ClassDeclarationBuilder
     implements ClassMemberDeclarationBuilder, ClassIntrospector {}
 
-/// The api used by [Macro] to get a [TypeDeclaration] for any given
-/// [TypeAnnotation].
-abstract class TypeIntrospector {
-  /// Resolves a [NamedTypeAnnotation] to its declaration.
-  Future<TypeDeclaration> resolve(NamedTypeAnnotation annotation);
+/// The interface used by [Macro]s to resolve any [NamedStaticType] to its
+/// declaration.
+///
+/// Only available in the definition phase of macro expansion.
+abstract class TypeDeclarationResolver {
+  /// Resolves a [NamedStaticType] to its [TypeDeclaration].
+  Future<TypeDeclaration> declarationOf(NamedStaticType annotation);
 }
 
 /// The base class for builders in the definition phase. These can convert
 /// any [TypeAnnotation] into its corresponding [TypeDeclaration], and also
 /// reflect more deeply on those.
 abstract class DefinitionBuilder
-    implements Builder, ClassIntrospector, TypeIntrospector {}
+    implements
+        Builder,
+        TypeResolver,
+        ClassIntrospector,
+        TypeDeclarationResolver {}
 
 /// The apis used by [Macro]s that run on classes, to fill in the definitions
 /// of any external declarations within that class.
--- a/working/macros/api/builders.dart
+++ b/working/macros/api/builders.dart
@@ -27,7 +27,7 @@ abstract class TypeComparator {
   /// Returns true if [leftType] is a subtype of [rightType].
   bool isSubtypeOf(TypeAnnotation leftType, TypeAnnotation rightType);
 
-  /// Retruns true if [leftType] is an identical type to [rightType].
+  /// Returns true if [leftType] is an identical type to [rightType].
   bool isExactly(TypeAnnotation leftType, TypeAnnotation rightType);
 }
 
--- a/working/macros/api/builders.dart
+++ b/working/macros/api/builders.dart
@@ -22,21 +22,24 @@ abstract class TypeBuilder implements Builder {
   void declareType(DeclarationCode typeDeclaration);
 }
 
-/// The api used by [Macro]s to contribute new (non-type)
-/// declarations to the current library.
-///
-/// Can also be used to do subtype checks on types.
-abstract class DeclarationBuilder implements Builder {
-  /// Adds a new regular declaration to the surrounding library.
-  ///
-  /// Note that type declarations are not supported.
-  Declaration declareInLibrary(DeclarationCode declaration);
-
+/// The interface for checking if a type implements another type.
+abstract class TypeComparator {
   /// Returns true if [leftType] is a subtype of [rightType].
   bool isSubtypeOf(TypeAnnotation leftType, TypeAnnotation rightType);
 
   /// Retruns true if [leftType] is an identical type to [rightType].
   bool isExactly(TypeAnnotation leftType, TypeAnnotation rightType);
+}
+
+/// The api used by [Macro]s to contribute new (non-type)
+/// declarations to the current library.
+///
+/// Can also be used to do subtype checks on types.
+abstract class DeclarationBuilder implements Builder, TypeComparator {
+  /// Adds a new regular declaration to the surrounding library.
+  ///
+  /// Note that type declarations are not supported.
+  void declareInLibrary(DeclarationCode declaration);
 }
 
 /// The api used by [Macro]s to contribute new members to a class.
--- a/working/macros/api/builders.dart
+++ b/working/macros/api/builders.dart
@@ -24,7 +24,13 @@ abstract class TypeBuilder implements Builder {
 /// macro expansion.
 abstract class TypeResolver {
   /// Resolves [typeAnnotation] to a [StaticType].
-  StaticType resolve(TypeAnnotation typeAnnotation);
+  ///
+  /// Throws an error if the type annotation cannot be resolved. This should
+  /// only happen in the case of incomplete or invalid programs, but macros
+  /// may be asked to run in this state during the development cycle. It is
+  /// helpful for users if macros provide a best effort implementation in that
+  /// case or handle the error in a useful way.
+  Future<StaticType> resolve(TypeAnnotation typeAnnotation);
 }
 
 /// The api used to introspect on a [ClassDeclaration].
--- a/working/macros/api/expansion_protocol.dart
+++ b/working/macros/api/expansion_protocol.dart
@@ -46,7 +46,8 @@ abstract class MacroExecutor {
   Future<MacroExecutionResult> executeDeclarationsPhase(
       MacroInstanceIdentifier macro,
       Declaration declaration,
-      TypeComparator typeComparator);
+      TypeComparator typeComparator,
+      ClassIntrospector classIntrospector);
 
   /// Runs the definitions phase for [macro] on a given [declaration].
   ///
@@ -69,11 +70,11 @@ abstract class MacroExecutor {
 /// All argument instances must be of type [Code] or a built-in value type that
 /// is serializable (num, bool, String, null, etc).
 class Arguments {
-  final List<Object?> positionalArgs;
+  final List<Object?> positional;
 
-  final Map<String, Object?> namedArgs;
+  final Map<String, Object?> named;
 
-  Arguments(this.positionalArgs, this.namedArgs);
+  Arguments(this.positional, this.named);
 }
 
 /// An opaque identifier for a macro class, retrieved by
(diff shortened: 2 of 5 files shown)


working/macros/api/expansion_protocol.dart
```diff
@@ -0,0 +1,109 @@
...
+  /// Invoked when an implementation discovers a new macro definition in a
+  /// library, and prepares this executor to run the macro.
+  ///
+  /// May be invoked more than once for the same macro, which will cause the
+  /// macro to be re-loaded. Previous [MacroClassIdentifier]s and
+  /// [MacroInstanceIdentifier]s given for this macro will be invalid after
+  /// that point and should be discarded.
+  ///
+  /// Throws an exception if the macro fails to load.
+  Future<MacroClassIdentifier> loadMacro(Uri library, String name);
```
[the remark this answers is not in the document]
jakemac53: It would be any valid import uri that can be resolvable with the package config. This api is intentionally abstracting away the details - so for instance I would expect one implementation to load it via `IsolateMirror.loadUri`.

For testing we could have a specific implementation of this interface that uses an in memory file system. It would also allow you to unit test in a more black boxed way with this part of the system (you could have a mock `MacroExecutor` that never actually runs anything, just returns fake declarations).

working/macros/api/expansion_protocol.dart
```diff
@@ -0,0 +1,109 @@
...
+  /// Runs the type phase for [macro] on a given [declaration].
+  ///
+  /// Throws an exception if there is an error executing the macro.
+  Future<MacroExecutionResult> executeTypesPhase(
+      MacroInstanceIdentifier macro, Declaration declaration);
+
+  /// Runs the declarations phase for [macro] on a given [declaration].
+  ///
+  /// Throws an exception if there is an error executing the macro.
+  Future<MacroExecutionResult> executeDeclarationsPhase(
```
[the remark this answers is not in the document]
jakemac53: Oh yes you are correct you need the `ClassIntrospector` here, good catch.

working/macros/api/expansion_protocol.dart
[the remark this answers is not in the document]
jakemac53: I created a very basic and partial implementation to hopefully help here https://github.com/dart-lang/language/pull/2022

working/macros/api/expansion_protocol.dart
```diff
@@ -0,0 +1,104 @@
...
+/// The interface used by Dart language implementations, in order to load
+/// and execute macros, as well as produce library augmentations from those
+/// macro applications.
+///
+/// This class more clearly defines the role of a Dart language implementation
+/// during macro discovery and expansion, and unifies how augmentation libraries
+/// are produced.
+abstract class MacroExecutor {
+  /// Invoked when an implementation discovers a new macro definition in a
+  /// library, and prepares this executor to run the macro.
```
munificent, line 14: Is `name` the name of the macro definition? If so, might want to clarify in the doc comment.

working/macros/api/expansion_protocol.dart
```diff
@@ -0,0 +1,104 @@
...
+///
+/// All modifications are expressed in terms of library augmentation
+/// declarations.
+abstract class MacroExecutionResult {
+  /// Any library imports that should be added to support the code used in
+  /// the augmentations.
+  Iterable<DeclarationCode> get imports;
+
+  /// Any augmentations that should be applied as a result of executing a macro.
+  Iterable<DeclarationCode> get agumentations;
```
munificent, line 96: "agu" -> "aug"

working/macros/api/expansion_protocol.dart
```diff
@@ -0,0 +1,104 @@
...
+abstract class MacroInstanceIdentifier {}
+
+/// A summary of the results of running a macro in a given phase.
+///
+/// All modifications are expressed in terms of library augmentation
+/// declarations.
+abstract class MacroExecutionResult {
+  /// Any library imports that should be added to support the code used in
+  /// the augmentations.
+  Iterable<DeclarationCode> get imports;
```
munificent, line 93: Is there a reason this is separated out from other kinds of added declarations? If it's important to do so, would it make more sense for this to have a different type, like `Iterable<ImportDeclaration>` or something?

working/macros/api/expansion_protocol.dart
```diff
@@ -0,0 +1,104 @@
...
+
+  /// Any augmentations that should be applied as a result of executing a macro.
+  Iterable<DeclarationCode> get agumentations;
+}
+
+/// Each of the different macro execution phases.
+enum Phase {
+  types, // Only new types are added in this phase
```
munificent, line 101: Nit: These comments should all be doc comments with "." at the end.

working/macros/api/expansion_protocol.dart
```diff
@@ -0,0 +1,104 @@
...
+/// A summary of the results of running a macro in a given phase.
+///
+/// All modifications are expressed in terms of library augmentation
+/// declarations.
+abstract class MacroExecutionResult {
+  /// Any library imports that should be added to support the code used in
+  /// the augmentations.
+  Iterable<DeclarationCode> get imports;
+
+  /// Any augmentations that should be applied as a result of executing a macro.
```
munificent, line 95: Is the result here an "augmentation" in the sense that it contains code that is syntactically valid and would be meaningful if it appeared inside an augmentation library?

If so, we'll need to extend augmentation libraries to support this. Right now, there's no syntax to declare that a type exists without defining it.

munificent (approved):
This looks great.

johnniwinther (approved):

working/macros/api/expansion_protocol.dart
jakemac53: They are separated out so that you can merge a bunch of these together to create a single library augmentation file, without actually parsing any of the `Code` objects.

I could create an `ImportDeclaration` but I don't think there is a ton of value and it might be a bit confusing for regular macro authors, and I don't today have any other specialized `Declaration` types for Code objects. It is very minimal.

working/macros/api/expansion_protocol.dart
jakemac53: I believe this is already fully supported, from this section https://github.com/dart-lang/language/blob/master/working/augmentation-libraries/feature-specification.md#static-semantics.

```
2. Else, it is a new declaration in the augmentation library:

If a declaration with that name already exists in the namespace, error. (Exception: setters do not collide with getters and final variables.)

Else, add the declaration to the namespace.
```

So to add a new type you just add the type here (no augment keywords or anything), and that gets merged into the original libraries namespace.

working/macros/api/builders.dart
```diff
@@ -22,21 +22,24 @@ abstract class TypeBuilder implements Builder {
   void declareType(DeclarationCode typeDeclaration);
 }
 
+/// The interface for checking if a type implements another type.
+abstract class TypeComparator {
+  /// Returns true if [leftType] is a subtype of [rightType].
+  bool isSubtypeOf(TypeAnnotation leftType, TypeAnnotation rightType);
```
scheglov, line 28: How will clients use this? It looks that `TypeAnnotation` is practically just a name, a string. I can see how it can be used on one side - check what is written in the code. But if we want for example to know if something is a subtype of `Widget` from `package:flutter`, we need to tell that that `Widget` name *is* from `package:flutter`.

munificent: +1. I really do think the API should make a clear distinction between a type annotation and a type.

jakemac53: Yes I definitely agree this part of the api is funky. However, adding an api that takes a `TypeAnnotation` and resolves it just moves the problem to a different API.

One other option is to allow you to resolve entire `Declaration`s, which would resolve all the `TypeAnnotation`s in the declaration. But that has several downsides as well:

- It dramatically increases the api surface area (all declarations need resolved counterparts).
- It is less granular than we might want for many situations (I might only care about the return type of a function, etc).

My general thought is that these `TypeAnnotation` objects actually must originate from the implementation that is being asked this question - and thus they will actually be subtypes of `TypeAnnotation` that contain more information than just what `TypeAnnotation` has, and do have the ability to answer these questions.

It is funky though I agree for sure, because you can't express this in the API itself.

Would you all be OK with adding a TODO on this particular part for now, to get the rest landed, and then I can come back to this?

jakemac53: Maybe the simplest solution here actually is to bring back the `Scope` object?

scheglov: OK, I see. So, `TypeAnnotation` could be anything - an unresolved annotation from AST, or an actual type requested in some way, like `typeProvider.getClassElement('package:flutter/widget.dart', 'Widget').instantiate([], noNullability)`.

One weird case if when both `left` and `right` type annotations are from AST. But I guess it still might be interesting sometimes.

jakemac53: Ya, I don't envision macro authors building these on their own (and we may even want to try and prevent them from doing so?). So they always come from either an actual AST or `instantiate`, etc. We also want to have an api for getting one from a runtime type in the macro code itself at some point.

But conceptually a `TypeAnnotation` subtype should have more underlying information that is actually exposed through its own public api. It is difficult to come up with an agnostic way of representing that though. But if we brought back the opaque `Scope` class, it might help?

scheglov: I don't know what `Scope` is. Any `TypeAnnotation` from AST has its actual scope. And as much as I can imagine, the other type is requested as I described above - by the library URI and name.

One thing that is not clear to me, is if any AST type annotation is always resolved. During the phase one (when we can declare new types), we don't have types resolved yet, right? Then, what `TypeComparator` methods will return? `false`? Exception?

jakemac53: > I don't know what `Scope` is.

It was intended as an opaque thing, that would only have a meaning to an individual implementation. It would hold the information required to know where a type annotation actually came from (which would be required to resolve it).

Ultimately I dropped it because it felt unnecessary (the implementations could just put this info on their own subtype).

It sounds to me like maybe that is still the right choice and adding the abstraction back as a user-visible but opaque thing might not really hold value (or worse, just be confusing).

> One thing that is not clear to me, is if any AST type annotation is always resolved.

They are not resolved, no. But they need to be _resolvable_, if that makes sense.

> During the phase one (when we can declare new types), we don't have types resolved yet, right? Then, what `TypeComparator` methods will return? `false`? Exception?

The phase one methods do not get a `TypeComparator`, their is no public api available to them to attempt to resolve things (and the executor function also doesn't accept one):

```dart
  Future<MacroExecutionResult> executeTypesPhase(
      MacroInstanceIdentifier macro, Declaration declaration);
```

munificent: > One other option is to allow you to resolve entire `Declaration`s, which would resolve all the `TypeAnnotation`s in the declaration.

I don't think we need that. I think there are basically three concepts:

*   A **type annotation** corresponds to the actual syntax of a type annotation as it might appear in a program. Think a return type on a function, type on a variable declaration, etc. The identifiers in it might be resolved so that we know where they come from, but the object represents the *annotation*, not the type. For example:

    ```dart
    typedef AlsoInt = int;

    main() {
      int a;
      AlsoInt b;
    }
    ```

    Here, `a` and `b` have different type annotations, even though they end up meaning the same type.

*   A **type** is the entity that the static type system works with. It's the construct for which questions like "is this a subtype?" are meaningful. (Questions like this are not meaningful for type *annotations* because things like aliases obscure it.) Types are the inputs and outputs of type inference.

*   A **type declaration** is some Dart language construct that brings a new named type into being. Class declarations, enum declarations, mixin declarations. In some cases, you can think of those declarations as *being* their corresponding types, but it's a clearer model to separate them. In particular, with generic classes, there is a real distinction between the class declaration and the types around it. Given:

    ```dart
    class Box<T> {}
    ```

    The type `Box<int>` is a real type that is meaningful, but that type is not simply a pointer to the class declaration because it also tracks the set of type arguments applied to the class.

You can move between these three concepts and they overlap in some ways, but they're distinct. You can take a type annotation and "resolve" it get the type that it describes. If that type is derived from some nominal declaration like a class, you can probably get from the type to the class. But we can't collapse those last two categories *in general*. There are types like function types where there is no declaration to go to at all (also, eventually record types). And even instantiations of generic types need to also track their type arguments.

If I remember right, the analyzer element API doesn't model these three concepts clearly. Every type implicitly tries to also be a declaration. If I remember right, this is an artifact of the analyzer being written back before Dart had real support for function types. (All it had was function-typed parameters.) So, back then, any type you could talk about also had a declaration you could point to. But that's no longer the case.

I think the API we should have for introspection should treat them separately:

*   A base `TypeAnnotation` class for anything that can appear in a type annotation with subclasses for `NamedTypeAnnotation` and `FunctionTypeAnnotation`. (Maybe `NullableTypeAnnotation` if you want to consider nullable types a wrapper around their base type.) The properties of these are also type annotations.

    You can resolve one of these to get...

*   A base, uh, `StaticType` class for an object that represents a static type. These are the things you can ask about subtypes, same type, etc. Again subclasses for `NamedType` (which has a potentially empty list of type arguments) and `FunctionType`. The properties of these are `StaticTypes`.

    The `NamedType` class then has a property or operation to get to the declaration that introduces that name. That takes you to...

*   `ClassDeclaration`, `EnumDeclaration`, etc. These are essentially syntactic classes. They let you introspect over the structure and contents of the type. They will have a list of type parameters, superclass, superinterfaces, fields, etc. References to types in all of those are `TypeAnnotation`, again because this object represents a mostly-syntactic entity.

    In principle, sure, we could also offer something like a `ResolvedClassDeclaration` where every property that was a type annotation has been resolved to its corresponding type, but I don't think that's necessary or very useful.

Type aliases are the funny one. We could also have a operation on `NamedTypeAnnotation` to take you to the declaration where that name is defined. If you do that for a type annotation that refers to a type alias, it would take you to the `typedef` declaration, not the actual declaration of the underlying base type. 

Does this make sense?

scheglov: Actually, it was my personal crusade (one of :-)) to make function types completely synthetic. So, now `FunctionTypeImpl` does not know if it was created from a method, or specified as a generic function type. For code generation purposes (e.g. Angular) however we track if a type (any, not only function type) was create from a type alias.

What you describe about separation between declarations, types, and annotations, all makes sense to me, because this is how it is done in the analyzer.

For type annotations that reference a type alias `typedef MyAlias<T> = List<T>;`, like `MyAlias<int>` the name `MyAlias` references `MyAlias` in `typedef`. But the type is `List<int>`, somewhere in `dart:core`.

jakemac53: I think this all almost basically matches the existing setup, but with the addition of a `StaticType` class.

Concretely we would change the `TypeIntrospector` api from this:

```dart
/// The api used by [Macro] to get a [TypeDeclaration] for any given
/// [NamedTypeAnnotation].
abstract class TypeIntrospector {
  /// Resolves a [NamedTypeAnnotation] to its declaration.
  Future<TypeDeclaration> resolve(NamedTypeAnnotation annotation);
}
```

To this:

```dart
/// The api used by [Macro] to get a [TypeDeclaration] for any given
/// [NamedTypeAnnotation].
abstract class TypeIntrospector {
  /// Resolves a [TypeAnnotation] to a [StaticType].
  Future<StaticType> resolve(TypeAnnotation annotation);

  /// Resolves a [NamedType] to its declaration.
  Future<TypeDeclaration> declarationOf(NamedType annotation);
}
```

And then move the `isSubtypeOf` and `isExactly` apis from the `TypeComparator` interface to this new `StaticType` interface.

I do think it makes sense to add also a `TypeAliasDeclaration` class, and give back that when asking for the declaration of a named type that is actually an alias (and that will have a `TypeAnnotation get type;` to get the actual type).

## Pros

- There is a more explicit line drawn between a type annotation and a type.

## Cons

- It requires 2 async operations in order to go from a `TypeAnnotation` to its declaration, instead of one in the current api.
  -  We could also have an api to go from a `TypeAnnotation` directly to the declaration to avoid this if desired (ie: the current api).
- It still has all the same implicit problems of the existing api where `TypeAnnotation`s are a bit weird, and must have additional implementation specific tracking information in them in order to be resolvable. So I don't think it concretely changes anything, but maybe produces a more explicit api that is easier to understand?

I am

munificent: LGTM!

jakemac53: Ok I pushed up some of these changes, ended up a bit more complex than I expected but the gist is the same. There is a bit to still work out in terms of how these StaticTypes can actually be converted to (or used by) Code objects, so we can build up a proper augmentation library from them etc.

But lets work that out separately.

working/macros/api/builders.dart
```diff
@@ -22,30 +15,24 @@ abstract class TypeBuilder implements Builder {
...
-  /// Adds a new declaration to the surrounding class.
-  void declareInClass(DeclarationCode declaration);
+/// The [StaticType]s can be compared against other [StaticType]s to see how
+/// they relate to each other.
+///
+/// This api is only available to the declaration and definition phases of
+/// macro expansion.
+abstract class TypeResolver {
+  /// Resolves [typeAnnotation] to a [StaticType].
+  StaticType resolve(TypeAnnotation typeAnnotation);
```
scheglov, line 27: Do we want to return `dynamic` or `null` for unresolvable annotations?

working/macros/api/introspection.dart
```diff
@@ -50,8 +68,8 @@ abstract class TypeDeclaration implements Declaration {
   /// The type parameters defined for this type declaration.
   Iterable<TypeParameterDeclaration> get typeParameters;
 
-  /// Create a type annotation representing this type with [typeArguments].
-  TypeAnnotation instantiate({List<TypeAnnotation> typeArguments});
+  /// Create a static type representing this type with [typeArguments].
+  StaticType instantiate({List<StaticType> typeArguments});
```
scheglov, line 72: We probably also want a `nullabilitySuffix`, to instantiate `List<int>` vs. `List<int>?`.

working/macros/api/introspection.dart
```diff
@@ -39,6 +39,24 @@ abstract class NamedTypeAnnotation implements TypeAnnotation {
   Iterable<TypeAnnotation> get typeArguments;
 }
 
+/// The interface representing a resolved type.
+///
+/// Resolved types understand exactly what type they represent, and can be
+/// compared to other static types.
+abstract class StaticType {
+  /// Returns true if this is a subtype of [other].
+  bool isSubtypeOf(StaticType other);
```
scheglov, line 48: Is it right that `StaticType` related APIs are synchronous?

If macro builders run in a separate isolate, all the type and declaration information is in the main isolate, e.g. in form of the analyzer element model. So, almost everything is RPC.

working/macros/api/builders.dart
jakemac53: We shouldn't have unresolvable ones (assuming correct programs) - so I think we should just throw.

jakemac53: I am also making this async.

working/macros/api/introspection.dart
jakemac53: Ah right good point, we have this on the `TypeAnnotation` but also need it on `StaticType`.

This should probably also be `async` I suppose :scream:

jakemac53:
Merging this for now to get the ball rolling and we can keep iterating. Once I move it into the SDK repo it should be easier also.

PR #2023, opened 2021-12-10
Update pre-null-safety spec with new null safety sections

eernstg:
This PR was created by deleting the sections from `Subtypes` to (but not including) `Reference` plus the appendix `Algorithmic Subtyping`, and inserting the section `Subtypes` from the null safety update branch plus all the following sections until (but not including) `Reference`, plus the new version of the appendix `Algorithmic Subtyping`.

The point is that this PR introduces a large chunk of the new material related to null safety which is introduced into the language specification, and it does not include all the small changes to sections that are mostly the same as before (those changes will be submitted as other PRs later on). In other words, the text before `Subtyping` is unchanged (apart from the fixes mentioned below), and references from the new material to any location before `Subtyping` may thus very well refer to obsolete language.

A couple of other updates were necessary in order to make the result LaTeX'able:

- `dart.sty` was updated to be the version used with the null safety PR (the new version includes a number of declarations that are used by the material about null safety).
- `dartLangSpec.tex` was updated to use `\Delta` rather than `\Gamma` to denote a type variable environment (that's the common notation; we previously used `\Gamma`, but that's typically a "term" variable environment, so `\Delta` would be less confusing for readers who also read research papers involving type systems).
- `\RawFunctionTypeNamed` was updated to allow for the `required` modifier; a couple of changes were necessary in sections before `Subtyping` in order to make dartLangSpec.tex LaTeX'able again, but this only involved adding the new LaTeX macro argument `{r}`, further updates will take place later on (this update just updates stuff from `Subtyping` and out).
- The symbol used to indicate that a word or phrase is being defined was changed from `\diamond` to `\triangle`, because `\triangle` is now also used over `=` in order to indicate that a certain term of the form `A = B` defines `A`.
- Compile-time errors are now also indicated in the margin (using an `\ominus` symbol, similar to the symbolic traffic sign meaning "No entrance"), this symbol is used in the new material, that is, nowhere before section `Subtyping`; for instance, it occurs in section `Type Null`.

Unfortunately, the diff algorithm in git (and, probably, almost any diff algorithm) does not quite understand that the operation performed here is "delete two big chunks of text and insert replacements that are mostly new material", but it does seem to get on track after a while.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -21133,7 +21133,8 @@ All further interpretation of URIs is implementation dependent.
 \LMLabel{types}
 
 \LMHash{}%
-Dart supports static typing based on interface types.
+Dart supports static typing based on interface types
+(\ref{interfaceTypes}).
 
 \rationale{%
   The type system is sound in the sense that
@@ -23830,15 +23831,13 @@ because the alternatives are ambiguous or complex.%
 
 \LMHash{}%
 We define the auxiliary function \NominalTypeDepthName{}
-on interface types and \code{Object?} as follows:
+on interface types as follows:
 
 \begin{itemize}
 \item
-  \DefEquals{\NominalTypeDepth{Object?}}{0}.
-\item
-  \DefEquals{\NominalTypeDepth{Object}}{1}.
-\item
-  \DefEquals{\NominalTypeDepth{Null}}{1}.
+  % We could make it 1 rather than 0, to "reserve space" for `Object?`,
+  % but this function is never used with `Object?` anyway.
+  \DefEquals{\NominalTypeDepth{Object}}{0}.
 \item
   Let $T$ be a class or a mixin,
   and let $M$ be the set of immediate superinterfaces of $T$.
@@ -23846,10 +23845,6 @@ on interface types and \code{Object?} as follows:
   where $d$ is
   $\metavar{max}\,\{\;\NominalTypeDepth{$S$}\;|\;S\;\in M\;\}$.
 \end{itemize}
-
-\commentary{%
-
-}
 
 \LMHash{}%
 \BlindDefineSymbol{I, J, M}%
@@ -24239,7 +24234,7 @@ Let $T$ be a type, and let $T'$ be the transitive alias expansion of $T$
 (\ref{typedef}).
 We say that $T$ is an \Index{interface type} if{}f
 $T'$ is of the form \code{$C$<\List{T}{1}{k}>},
-where $C$ denotes a class different from \code{Never},
+where $C$ denotes a class different from \code{Never} and \code{Null},
 or $C$ denotes a mixin.
 
 \commentary{%
@@ -24247,13 +24242,14 @@ or $C$ denotes a mixin.
   Non-generic classes are included because we can have $k = 0$.
 
   In particular, the following types are \emph{not} interface types:
-  \VOID, \DYNAMIC, \FUNCTION, \code{FutureOr<$T$>} for any $T$, \code{Never},
+  \VOID, \DYNAMIC, \FUNCTION, \code{FutureOr<$T$>} for any $T$,
+  \code{Never}, \code{Null},
   any function type, any type variable, any intersection type,
   and any type of the form \code{$T$?}.
 
-  Conversely, built-in classes
-  like \code{Object}, \code{Null}, \code{num}, \code{int},
-  \code{String}, and \code{Exception} are interface types,
+  Conversely, built-in classes like, e.g.,
+  \code{Object}, \code{num}, \code{int}, \code{String}, and \code{Exception}
+  are interface types,
   and so are
   \code{Future<$T$>}, \code{Stream<$T$>}, \code{Iterable<$T$>},
   \code{List<$T$>}, \code{Map<$S$,\,\,$T$}, and \code{Set<$T$>},
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -21864,9 +21864,6 @@ However, their subtype relations are specified without restrictions.
 \newcommand{\SrnRightTop}{2}
 \newcommand{\SrnLeftTop}{3}
 \newcommand{\SrnBottom}{4}
-%\newcommand{\SrnRightObjectOne}{} Redundant
-%\newcommand{\SrnRightObjectTwo}{} Redundant
-%\newcommand{\SrnRightObjectThree}{} Redundant
 \newcommand{\SrnRightObjectFour}{5}
 \newcommand{\SrnNullOne}{6}
 % \newcommand{\SrnNullTwo}{} Redundant
@@ -22424,12 +22421,12 @@ the set of direct superinterfaces of $C$
     %% in \ref{mixinApplication} is incorrect, and also that it will need to
     %% be rewritten completely for the integration of the new mixin construct.
     The case where the superclass is a mixin application is covered via
-    the equivalence with a declaration
-    of a regular (possibly generic) superclass
+    the equivalence with a declaration of
+    a regular (possibly generic) superclass
     (\ref{mixinApplication}),
     and this means that there may be multiple subtype steps from
     a given class declaration to the class specified in an \EXTENDS{} clause.
-\end{itemize}%
+  \end{itemize}%
 }
 
 \subsubsection{Additional Subtyping Concepts}
@@ -22481,7 +22478,8 @@ if{}f $T$ is not nullable.
   Nullable types are types which are
   definitively known to include the null object,
   regardless of the value of any type variables.
-  This is equivalent to the syntactic criterion that $T$ is any of:
+  If $T'$ is the transitive alias expansion (\ref{typedef}) of $T$
+  then this is equivalent to the syntactic criterion that $T'$ is any of:
 
   \begin{itemize}[itemsep=-0.5ex]
   \item \VOID.
@@ -22503,13 +22501,14 @@ if{}f $T$ is not non-nullable.
   Non-nullable types are types which are definitively known to
   \emph{not} include the null object,
   regardless of the value of any type variables.
-  This is equivalent to the syntactic criterion that $T$ is any of:
+  If $T'$ is the transitive alias expansion (\ref{typedef}) of $T$
+  then this is equivalent to the syntactic criterion that $T$ is any of:
 
   \begin{itemize}[itemsep=-0.5ex]
   \item \code{Never}.
   \item Any function type.
   \item The type \FUNCTION.
-  \item Any interface type except \code{Null}.
+  \item Any interface type.
   \item \code{FutureOr<$S$>}, for any non-nullable type $S$.
   \item Any type variable $X$ whose bound is non-nullable.
   \item Any type of the form \code{$X$\,\&\,$S$}, where
@@ -22871,10 +22870,11 @@ where $T_r$ is determined as follows:
 
   \noindent
   then $T_r$ is
-  \FunctionTypePositional{R_0}{ }{X}{B}{s}{R}{n}{k}
+  \FunctionTypePositional{T'\!_0}{ }{X}{B'\!}{s}{T'\!}{n}{k}
 
   \noindent
-  where $R_i$ is \NormalizedTypeOf{$T_i$} for $i \in 0 .. n+k$.
+  where $T'\!_i$ is \NormalizedTypeOf{$T_i$} for $i \in 0 .. n+k$
+  and $B'\!_i$ is \NormalizedTypeOf{$B_i$} for $i \in 1 .. s$.
 \item If $T_u$ is of the form
   \FunctionTypeNamedStd{T_0}
 
@@ -22882,10 +22882,11 @@ where $T_r$ is determined as follows:
   where $r_j$ is either \REQUIRED{} or empty
   then $T_r$ is
   \noindent
-  \FunctionTypeNamed{R_0}{ }{X}{B}{s}{R}{n}{x}{k}{r}
+  \FunctionTypeNamed{T'\!_0}{ }{X}{B'\!}{s}{T'\!}{n}{x}{k}
 
   \noindent
-  where $R_i$ is \NormalizedTypeOf{$T_i$} for $i \in 0 .. n+k$.
+  where $T'\!_i$ is \NormalizedTypeOf{$T_i$} for $i \in 0 .. n+k$
+  and $B'\!_i$ is \NormalizedTypeOf{$B_i$} for $i \in 0 .. s$.
 \end{itemize}
 
 \commentary{%
@@ -23224,8 +23225,8 @@ to parameter type declarations,
 which is defined as follows.
 Assume that $P_1$ and $P_2$ are two formal parameter type declarations
 with declared type $T_1$ respectively $T_2$,
-such that both are positional or both are named,
-with the same name \DefineSymbol{n}.
+such that both are positional,
+or both are named and have the same name \DefineSymbol{n}.
 Then \UpperBoundType{$P_1$}{$P_2$} (respectively \LowerBoundType{$P_1$}{$P_2$})
 is the formal parameter declaration $P$,
 with the following proporties:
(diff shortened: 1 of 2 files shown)


eernstg:
An update fixing a few issues with interface types is on the way, will upload in half an hour, approx. ;-)

eernstg:
Done.

chloestefantsova:
I skimmed over the changes and created the following checklist. I plan to review the text in parts and cross them out here when I'm done. The checklist itself may change in the future in case I missed anything.

- [ ] Subtype algorithm.
  - [ ] Rules of inference.
  - [ ] Subsequent commentary.
- [ ] Type nullability.
- [ ] Top merge.
- [ ] IsTopType.
- [ ] Moretop, morebottom.
- [ ] Type normalization.
- [ ] Canonical syntax of types.
- [ ] Standard bounds.
- [ ] Closures of types.
- [ ] Specific kinds of types
  - [ ] Types bounded by types.
  - [ ] Class building types.
  - [ ] Interface types.
  - [ ] Intersection types.
  - [ ] Never.
  - [ ] Null.
  - [ ] Type.
  - [ ] Function.
  - [ ] dynamic.
- [ ] Type inference.
  - [ ] Flow analysis.
  - [ ] Type promotion.
- [ ] Appendix. Algorithmic subtyping.

chloestefantsova:
I reviewed the part "Subsequent commentary" to the subtype algorithm, in the green "after" column the lines range from 20889 to 21291. I have two drive-by questions to the surrounding text. Otherwise, the part LGTM.

specification/dartLangSpec.tex
```diff
@@ -21253,39 +21261,2076 @@ \subsubsection{Informal Subtype Rule Descriptions}
...
+if{}f \SubtypeNE{T}{\code{Object}},
+and that $T$ is
+\IndexCustom{potentially nullable}{type!potentially nullable}
+if{}f $T$ is not non-nullable.
+
+\commentary{%
+Non-nullable types are types which are definitively known to
+\emph{not} include the null object,
+regardless of the value of any type variables.
+This is equivalent to the syntactic criterion that $T$ is any of:
```
chloestefantsova, line 21327: Similarly to above, I think type aliases should be mentioned explicitly.

specification/dartLangSpec.tex
```diff
@@ -21253,39 +21261,2076 @@ \subsubsection{Informal Subtype Rule Descriptions}
...
+  for any type $T$.
+\item
+  \DefEquals{\NonNullType{$X$}}{\code{$X$\,\&\,\NonNullType{$B$}}},
+  where $X$ is a type variable with bound $B$,
+  such that $\NonNullType{$B$} \not= B$.
+\item
+  \DefEquals{\NonNullType{$X$\,\&\,$T$}}{\code{$X$\,\&\,\NonNullType{$T$}}},
+  for any type $T$.
+\item
+  Otherwise, \DefEquals{\NonNullType{$T$}}{T},
```
chloestefantsova, line 21395: Should that be `$T$` at the end of the formula?

chloestefantsova:
I checked the part "Type Nullability", in the green "after" column the lines range from 21291 to 21414. I have left some comments, but otherwise LGTM.

specification/dartLangSpec.tex
```diff
@@ -21253,39 +21261,2076 @@ \subsubsection{Informal Subtype Rule Descriptions}
...
+For instance, \TopMergeTypeName{} is used during the computation of
+the interface of a class member which is declared
+in multiple superinterfaces.
+If a class $C$ inherits a method $m$ that accepts
+an argument of type \code{List<\DYNAMIC>} from one superinterface,
+and another method $m$ that accepts
+an argument of type \code{List<\VOID>}
+from another superinterface,
+the parameter will have type \code{List<Object?>} in $C$,
+and this will ensure a more strict static typing by default than, say,
```
chloestefantsova, line 21483: As a reader, I have a question: isn't `void` even more restrictive than `Object?`?

chloestefantsova:
I checked the part "Top merge", in the green "after" column the lines range from 21414 to 21494. LGTM with one comment.

chloestefantsova:
I checked the part "IsTopType", in the green "after" column the lines range from 21494 to 21539. It LGTM.

chloestefantsova:
I checked the part "Moretop, morebottom", in the green "after" column the lines range from 21539 to 21582.

specification/dartLangSpec.tex
```diff
@@ -21253,39 +21261,2076 @@ \subsubsection{Informal Subtype Rule Descriptions}
...
+  \end{itemize}
+\item If $T_u$ is of the form \code{$C$<\List{T}{1}{k}>}
+  then $T_r$ is \code{$C$<\List{R}{1}{k}>},
+  where $R_i$ is \NormalizedTypeOf{$T_i$}, for $i \in 1 .. k$.
+\item If $T_u$ is of the form
+  \FunctionTypePositionalStd{T_0}
+
+  \noindent
+  then $T_r$ is
+  \FunctionTypePositional{R_0}{ }{X}{B}{s}{R}{n}{k}
```
chloestefantsova, line 21694: The feature spec seem to require the normalization of the bounds of type variables too. Is it omitted on purpose?

chloestefantsova:
I checked the part "Type normalization", in the green "after" column the lines range from 21582 to 21737. LGTM with one question.

chloestefantsova:
I checked the part "Canonical syntax of types", in the green "after" column the lines range from 21737 to 21931. LGTM.

specification/dartLangSpec.tex
```diff
@@ -21253,39 +21261,2076 @@ \subsubsection{Informal Subtype Rule Descriptions}
...
+
+\begin{dartCode}
+\CLASS{} I \{\}
+\CLASS{} J \{\}
+\CLASS{} A \IMPLEMENTS{} I, J \{\}
+\CLASS{} B \IMPLEMENTS{} I, J \{\}
+\end{dartCode}
+
+\commentary{%
+No \emph{least} upper bound exists for \code{A} and \code{B},
```
chloestefantsova, line 21971: As I understand it, with the exception of a top type?

specification/dartLangSpec.tex
```diff
@@ -21253,39 +21261,2076 @@ \subsubsection{Informal Subtype Rule Descriptions}
...
+  if both $P_1$ and $P_2$ have this modifier
+  (respectively, if either $P_1$ or $P_2$ has this modifier).
+\end{itemize}
+
+\LMHash{}%
+The function \UpperBoundTypeName{} is defined as follows.
+
+\begin{itemize}
+\item
+  \DefEquals{\UpperBoundType{$T$}{$T$}}{T}.
```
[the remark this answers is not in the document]
chloestefantsova: Similarly below, but I'm not sure about the treatment of that argument by `\DefEquals`.

specification/dartLangSpec.tex
```diff
@@ -21253,39 +21261,2076 @@ \subsubsection{Informal Subtype Rule Descriptions}
...
+    \EXTENDS\,$B_{1m}$>($P_{11}$,\,\ldots,\,$P_{1k}$)}
+
+  \noindent
+  \code{$T_2$\,\FUNCTION<$X_1$\,\EXTENDS\,$B_{21}$,\,\ldots,\,$X_m$\,%
+    \EXTENDS\,$B_{2m}$>($P_{21}$,\,\ldots,\,$P_{2l}$)}
+
+  \noindent
+  such that each $B_{1i}$ and $B_{2i}$ are types with the same canonical syntax,
+  and both have the same number of required positional parameters.
+  Let $q$ be $\metavar{min}(k, l)$,
```
chloestefantsova, line 22254: Interesting. I think both the CFE and the Analyzer expect `k == l`. At least they both accept the code below that shows that `UP(Function(int, int, int), Function(int, int)) = Function` and not `Function(int, int)`, as this rule suggests.

```dart
foo(void Function(int, int, int) f, void Function(int, int) g) {
  var x = [f, g];
  var h = x.first;
  h(0, 1, 2, 3, 4, 5);
}
```

specification/dartLangSpec.tex
```diff
@@ -21253,39 +21261,2076 @@ \subsubsection{Informal Subtype Rule Descriptions}
...
+  \end{itemize}
+
+  \commentary{%
+    This case includes non-generic function types by allowing $m$ to be zero.%
+  }
+%%
+%% TODO(eernst), for review: Why do we not have a rule for
+%% \UpperBoundType{T1 Function(P1..Pm, [...])}{T2 Function(P1..Pk, {...}}}
+%% = T3 Function(R1..Rk), where the left operand has at least k parameters,
+%% plus the converse?
```
chloestefantsova, line 22319: It seems that this case is covered by treating the required and optional positional parameters in the positional-parameter case above uniformly. As I mentioned, the implementations seem to diverge from that. They expect the number of required positional parameters to be the same, and then just cut out the optional parameters in the upper bound.

chloestefantsova:
I've checked the part "Standard bounds", in the green "after" column the lines range from 21931 to 22759. LGTM with some comments.

chloestefantsova:
I've checked the part "Closures of types", in the green "after" column the lines range from 22759 to 22983. it LGTM.

specification/dartLangSpec.tex
```diff
@@ -21253,39 +21261,2076 @@ \subsubsection{Informal Subtype Rule Descriptions}
...
+schema is always a supertype of any type which matches the schema.%
+}
+
+
+\subsection{Types Bounded by Types}
+\LMLabel{typesBoundedByTypes}
+
+\LMHash{}%
+For a given type $T_0$, we introduce the notion of a
+\IndexCustom{$T_0$ bounded type}{type!T0 bounded}:
```
chloestefantsova, line 22987: Should that be `T_0` in the index instead of `T0`?

chloestefantsova:
I checked the part "Types bounded by types", in the green "after" column the lines range from 22983 to 23015. LGTM with one small question.

specification/dartLangSpec.tex
```diff
@@ -21253,39 +21261,2076 @@ \subsubsection{Informal Subtype Rule Descriptions}
...
+(\ref{null}).
+
+\commentary{%
+In other words, \code{Null} is a singleton type.
+Apart from top types
+(\ref{superBoundedTypes}),
+\code{Null} is a subtype of all types of the form \code{$T$?},
+and of all types $S$ such that \futureOrBase{S} is
+a top type or a type of the form \code{$T$?}.
+The only non-trivial subtypes of \code{Null} are
```
chloestefantsova, line 23199: Are type variables extending `Null` considered trivial in the remark?

specification/dartLangSpec.tex
```diff
@@ -21684,7 +23732,8 @@ \subsection{Type Void}
 must compare equal (according to the \lit{==} operator \ref{equality})
 to any instance of \code{Type} which reifies the type \code{Object}
 (\ref{dynamicTypeSystem}).
-It is not guaranteed that \code{identical(\VOID, Object)} evaluates to true.
+It is not guaranteed that \code{identical(\VOID, Object)} evaluates to
+the \TRUE{} object.
```
chloestefantsova, line 23736: I would expect `identical(void, Object)` to be `false`, actually. Maybe the text should mention `Object?` instead of the `Object`?

specification/dartLangSpec.tex
```diff
@@ -22093,108 +24137,335 @@ \subsubsection{Actual Types}
...
+\LMLabel{definiteAssignment}
+
+\LMHash{}%
+At each location where a local variable can be referred
+(\commentary{%
+e.g., as an expression, or as the left hand side of an assignment%
+}),
+the variable has a status as being
+\IndexCustom{definitely assigned}{local variable!definitely assigned} or
+\IndexCustom{definitely unassigned}{local variable!definitely unassigned}.
```
chloestefantsova, line 24190: This clause sounds as if it's strictly one of these two possibilities, but as stated below, a variable can be neither definitely assigned nor definitely unassigned.

specification/dartLangSpec.tex
```diff
@@ -22093,108 +24137,335 @@ \subsubsection{Actual Types}
...
+We might say that these promotions are ``atomic'',
+and all other promotions are derived from these,
+and from the control flow.%
+}
+
+\LMHash{}%
+Let $\ell$ be a location,
+and let $v$ be a local variable which is in scope at $\ell$.
+Assume that $\ell$ occurs after the declaration of $v$.
+The expressions that give rise to promotion or demotion
```
chloestefantsova, line 24413: "gives"

specification/dartLangSpec.tex
```diff
@@ -22093,108 +24137,335 @@ \subsubsection{Actual Types}
...
+and to \NonNullType{$T$} in the \FALSE{} continuation.
+
+%% TODO(eernst), for review: The null safety spec says that `T?` is
+%% promoted to `T`, but implementations _do_ promote `X extends int?` to
+%% `X & int`. So I've specified the latter. This is also more consistent
+%% with the approach used with `==`.
+\LMHash{}%
+A check of the form \code{$v$\,\,!=\,\,\NULL},
+\code{\NULL\,\,!=\,\,$v$},
+or \code{$v$\,\,\IS\,\,$T$}
```
chloestefantsova, line 24448: Is it true for every type `T`? What about the case of `T = S?`?

chloestefantsova:
I checked the part "Type inference", in the green "after" column the lines range from 24140 to 24613. LGTM with some minor questions.

specification/dartLangSpec.tex
```diff
@@ -22342,156 +24613,238 @@ \subsection{Operator Precedence}
...
+  $B_{0i}[Z_0/X_0, \ldots, Z_k/X_k]$:
+
+  \begin{itemize}
+  \item $p \geq n$.
+  \item $m \geq q$.
+  \item \SubtypeNE{S_i[Z_0/Y_0, \ldots, Z_k/Y_k]}{V_i[Z_0/X_0, \ldots, Z_k/X_k]}
+    for $i \in 0 .. q$.
+  \item \SubtypeNE{U_0[Z_0/X_0, \ldots, Z_k/X_k]}{U_1[Z_0/Y_0, \ldots, Z_k/Y_k]}.
+  \item $B_{0i}[Z_0/X_0, \ldots, Z_k/X_k]$ and $B_{1i}[Z_0/Y_0, \ldots, Z_k/Y_k]$
+    have the same canonical syntax, for $i \in 0 .. k$.
```
chloestefantsova, line 24796: Should we ask for a weaker requirement: the bounds should be mutual subtypes or should be equal if all top types treated as being equal to each other and the bottom types treated as being equal to each other?

specification/dartLangSpec.tex
```diff
@@ -22342,156 +24613,238 @@ \subsection{Operator Precedence}
...
+    for $i \in 0 .. n$.
+  \item \SubtypeNE{S_i[Z_0/Y_0, \ldots, Z_k/Y_k]}{V_j[Z_0/X_0, \ldots, Z_k/X_k]}
+    for $i \in n+1 .. q$, and $y_j = x_i$.
+  \item
+    for each $j$ such that $r_{0j}$ is \REQUIRED, there exists an
+    $i \in n+1 .. q$ such that $x_j = y_i$, and $r_{1i}$ is \REQUIRED.
+  \item
+    \SubtypeNE{U_0[Z_0/X_0, \ldots, Z_k/X_k]}{U_1[Z_0/Y_0, \ldots, Z_k/Y_k]}.
+  \item
+    $B_{0i}[Z_0/X_0, \ldots, Z_k/X_k]$ and $B_{1i}[Z_0/Y_0, \ldots, Z_k/Y_k]$
```
chloestefantsova, line 24836: Ditto.

chloestefantsova:
I checked the part "Appendix. Algorithmic subtyping", in the green "after" column the lines range fro 24613 to 24847. LGTM with some question. That concludes my initial pass over the PR.

specification/dartLangSpec.tex
eernstg: Yes! The requirement should be 'mutual subtypes', and it's been updated now. I overlooked the change were `===` in the null safety feature spec was changed to mean 'mutual subtypes', and that's the reason why I kept the old text because `===` was still there. That shows the power of change propagation! ;-)

specification/dartLangSpec.tex
eernstg: Ditto.

eernstg (commented):
Review response.

specification/dartLangSpec.tex
chloestefantsova: I see. Thanks for explaining it!

specification/dartLangSpec.tex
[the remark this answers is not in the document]
chloestefantsova: Thanks for the explanation!

PR #2024, opened 2021-12-10
clarify that macros can't generate other macros

jakemac53:


working/macros/feature-specification.md
```diff
@@ -260,6 +260,10 @@ Here, macros contribute new types to the program&mdash;classes, typedefs, enums,
 etc. This is the only phase where a macro can introduce a new visible name into
 the top level scope.
 
+**Note**: Macro classes _cannot_ be generated in this way, but they can rely on
+macro generated declarations for their implementation. This ensures that all
```
munificent, line 264: "rely on macro generated declarations for their implementation" isn't very clear. That may be OK, but we may want to be more concrete. Can a macro class inherit from a generated class? Implement a generated interface? Contain a field of a generated type? Apply a macro to itself in order to generate a method?

Is the only real restriction here that every name that can appear in a macro application must refer to a piece of hand-authored Dart syntax that starts with `macro class`?

munificent (approved):
+1, but I wouldn't be surprised if we end up tweaking this further.

working/macros/feature-specification.md
jakemac53: The only restriction necessary I think is that you can't literally generate a macro class (ie: one with the `macro` keyword). The point is just to ensure they are all discoverable, at least right now.

jakemac53:
> +1, but I wouldn't be surprised if we end up tweaking this further.

Ya we might change the rules but i just wanted to get this restriction written down :).

PR #2027, opened 2021-12-14
Add (first) testing plan for enhanced enums.

lrhn:


munificent:
Pretty sure we already have tests for enhanced enums at this point, so I'm going to go ahead and abandon this. :)

PR #2048, opened 2022-01-04
Spec: unnecessary leading '|'

modulovalue:


eernstg:
Hello @modulovalue, we should make this change: The leading `|` in the given lexer rules specify that `QUOTES_SQ` and `QUOTES_DQ` can be empty. This is working as intended. For example, we don't want to require that the first character after `'''` in a multi-line string _must_ be a single quote.

```dart
void main() {
  var s1 = '''It must be possible to write a multi-line string literal like this.''';
  var s2 = '''' .. we shouldn't be forced to let the _contents_ start with
a single quote (or two), like in this string literal.''';
}
```

modulovalue:
Hello Erik,

this suggestion is not related to the meaning of the grammar. I'm trying to parse and extract the grammar from the latex specification to feed it into other tools.

I assumed that this was a typo as it looked odd to me, sorry for that! 

Would you approve the following changes, which seem to me to be equivalent to the leading pipe i.e. to remove the leading pipe and to replace all references to `<QUOTES_SQ>` with `<QUOTES_SQ>?` and `<QUOTES_DQ>` with `<QUOTES_DQ>?`? 

```
...
<QUOTES\_SQ> ::= ---| ---`\sq' | `\sqsq'
<STRING\_CONTENT\_TSQ> ::= \gnewline{}
  <QUOTES\_SQ>+++?+++ (<STRING\_CONTENT\_COMMON> | `"' | `\\r' | `\\n')
<MULTI\_LINE\_STRING\_SQ\_BEGIN\_END> ::= \gnewline{}
  `\sqsqsq' <STRING\_CONTENT\_TSQ>* `\sqsqsq'
<MULTI\_LINE\_STRING\_SQ\_BEGIN\_MID> ::= \gnewline{}
  `\sqsqsq' <STRING\_CONTENT\_TSQ>* <QUOTES\_SQ>+++?+++ `${'

<MULTI\_LINE\_STRING\_SQ\_MID\_MID> ::= \gnewline{}
  `}' <STRING\_CONTENT\_TSQ>* <QUOTES\_SQ>+++?+++ `${'
...

...
<QUOTES\_DQ> ::= ---| ---`"' | `""'
<STRING\_CONTENT\_TDQ> ::= \gnewline{}
  <QUOTES\_DQ>+++?+++ (<STRING\_CONTENT\_COMMON> | `\sq' | `\\r' | `\\n')
<MULTI\_LINE\_STRING\_DQ\_BEGIN\_END> ::= \gnewline{}
  `"""' <STRING\_CONTENT\_TDQ>* `"""'
<MULTI\_LINE\_STRING\_DQ\_BEGIN\_MID> ::= \gnewline{}
  `"""' <STRING\_CONTENT\_TDQ>* <QUOTES\_DQ>+++?+++ `${'

<MULTI\_LINE\_STRING\_DQ\_MID\_MID> ::= \gnewline{}
  `}' <STRING\_CONTENT\_TDQ>* <QUOTES\_DQ>+++?+++ `${'
...
```

eernstg:
I agree that it is always possible to replace a derivation rule _R_ that derives a set of token sequences _S_ that contains the empty sequence by a rule that derives exactly every element of _S_ except the empty sequence, and then replace all occurrences of the non-terminal `N` on the left hand side of _R_ by `N?`.

For example, we could do this with the rule for `<metadata>` in the grammar, and the same reasoning allows us to do the same thing at the level of character sequences with the lexical rules `<QUOTES_SQ>` and `<QUOTES_DQ>`.

However, I don't think this is an improvement of the grammar: Before the change we have one location where it is specified that the given non-terminal can derive the empty sequence of tokens / the empty string, and after the change we have expressed the same property using one occurrence of `?` _for every usage_ of the non-terminal. This is a kind of anti-abstraction (a bit like manual inlining of code), and I don't think it's a good idea to use such an approach in general.

modulovalue:
> However, I don't think this is an improvement of the grammar

Could it be that we are looking at it from different perspectives? My suggestion is about reducing the complexity of the grammar with which the dart grammar is specified. 

My argument is basically the same as this proposal about removing the ++ and -- operators from Swift https://github.com/apple/swift-evolution/blob/master/proposals/0004-remove-pre-post-inc-decrement.md

But as your points are all valid and my confusion about the leading pipe has been resolved, I see this as resolved as well. Thank you!

eernstg:
@modulovalue, thanks for your input! I hope I did not come across as overly grumpy or inflexible. Do get in touch if you encounter other issues with the syntax of Dart, or anything Dart, of course!

> My suggestion is about reducing the complexity of the grammar with which the dart grammar is specified

I guess I just think (like ANTLR) that there's nothing special about the empty sequence of grammar expressions, so we can freely use `nt ::= | a | b` or `nt ::= a | | b` to allow `nt` to derive both `a`, `b`, and the empty string.

But I do recognize that it's a potential source of confusion. I suppose all those compiler/syntax textbooks use symbols like 𝜀 to denote the empty sequence of grammar expressions for exactly that reason, but also because they want to talk about that empty sequence in isolation (for stuff like "Assume that the non-terminal _N_ derives 𝜀, then blahblah"). In the language specification I'd probably just use words like 'the empty sequence of tokens' or even 'the empty string' to talk about the same thing.

PR #2064, opened 2022-01-11
State explicitly that an => body in a constructor is an error

eernstg:
The specification does not currently make it explicit that it is an error for a generative constructor to have a body of the form `=> e;`. This PR adds text to make that error explicit.


lrhn (approved):
LGTM!

PR #2076, opened 2022-01-26
[macros] Specify that identifiers in strings must refer to locals.

munificent:


working/macros/feature-specification.md
```diff
@@ -617,31 +622,35 @@ macros to library augmentations.
...
+*A compiler could implement this by generating an import of the library
+containing the declaration that the identifier refers to. The compiler adds a
+unique prefix to the import and then the identifier emitted as a prefixed
+identifier followed by the identifier's simple name.*
+
+Macros can also generate code from strings. When a `Code` object is created, if
+any identifiers appear inside strings or `Fragment`s used to build the code,
+they must resolve to local declarations inside that same `Code` object. Code
+generated from strings or fragments may not refer to free variables declared
+outside of the same `Code` object. To refer to non-local declarations, macro
```
jakemac53, line 648: One weird thing here is Code objects are composed of many other Code objects. I should be able to refer to local identifiers that are created in any _parent_ Code object as well, not just the current one. For instance:

```dart
DeclarationCode.fromParts([
  'void foo() {',
  DeclarationCode.fromString('var x = 1;'),
  Code.fromString('return x + 1;'), // This should be able to refer to `x` above.
  '}',
]),
```

jakemac53: You may also want to add multiple declarations in the same phase, and want to have them be able to refer to each other. Maybe really what we want to say is that you can refer to anything that is defined in the entire resulting augmentation file?

working/macros/feature-specification.md
```diff
@@ -617,31 +622,35 @@ macros to library augmentations.
...
+namespace. When the identifier object is inserted into other generated code,
+it retains its original resolution.
+
+*A compiler could implement this by generating an import of the library
+containing the declaration that the identifier refers to. The compiler adds a
+unique prefix to the import and then the identifier emitted as a prefixed
+identifier followed by the identifier's simple name.*
+
+Macros can also generate code from strings. When a `Code` object is created, if
+any identifiers appear inside strings or `Fragment`s used to build the code,
```
jakemac53, line 645: Note that there is actually no longer a `Fragment` class - there is just the base `Code` class which works like `Fragment` used to. The subtypes of `Code` are the ones that need to represent fully syntactically valid language constructs.

working/macros/feature-specification.md
munificent: > One weird thing here is Code objects are composed of many other Code objects. I should be able to refer to local identifiers that are created in any parent Code object as well, not just the current one.

Ah, good catch! Reworded to handle that case and clarify when resolution happens.

> You may also want to add multiple declarations in the same phase, and want to have them be able to refer to each other.

If they're instance members, you can (and should) use `this.`. If they're top level declarations, I think you can still go through an `Identifier`. The API you have now makes it pretty to get the `Identifier` for a declaration the macro itself created.

working/macros/feature-specification.md
```diff
@@ -633,22 +633,29 @@ the macro and not in scope where the macro is applied.
...
+method. To refer to non-local declarations, macro authors must use an
+`Identifier` object. Macros can get or create `Identifiers` using the
+introspection API.
+
+An identifier in a string used to create a `Code` object does not necessarily
+need to resolve to a local declaration *in that same `Code` object*. Macros can
+compose `Code` objects out of other `Code` objects and strings. It is only when
+the final `Code` object is returned by the macro and used to produce a
+declaration that all identifiers inside strings must be resolvable to locals
+in the generated declaration.
```
jakemac53, line 658: I guess I am still not fully sold on this. In particular when referencing instance members, it can be very convenient to just use the bare name. We could enforce that you use `this.` etc, and yes it can theoretically cause a problem to not do so, but it seems unlikely? We don't have any restrictions like this in codegen and [while it has come up](https://github.com/dart-lang/source_gen/issues/555) it is exceedingly rare?

I am concerned people will end up making async calls just to get a handle on an identifier for instance. Or end up implementing their macro across multiple phases just so they can get an identifier, when they could make a more efficient macro that probably never causes a problem if they didn't have to work around this.

munificent: > I guess I am still not fully sold on this. In particular when referencing instance members, it can be very convenient to just use the bare name.

One idea is to specify that bare identifiers are resolved in the namespace where the declaration is generated. Basically you get whatever resolution you would have gotten if you'd manually written the code there. But that's not actually how it would work: the generated augmentation library deliberately has its own top level scope separate from the main library's. We could make that work by having the generated augmentation library always import the main library, but that doesn't seem practically useful. Most macros probably don't *want* to accidentally refer to random top level declarations or stuff that was imported into the library where the macro is applied.

If it's just so that generated code has an easier way to refer to *instance* members... I don't know, writing `this.` isn't that bad, is it? That's what everyone writing JS is doing all day every day. And it's completely clear how the identifier should be resolved at that point.

I suppose we could say that identifiers are resolved as if the class appeared in an otherwise-empty top level namespace? That would let you resolve to instance and static members on the surrounding class, which is what you want, and that's about it.

cc @leafpetersen in case he has thoughts.

jakemac53: > If it's just so that generated code has an easier way to refer to _instance_ members... I don't know, writing `this.` isn't that bad, is it?

It works, but it's weird to tell macro authors they have to do this, when it provides no tangible benefit to the vast majority of them. You don't write normal Dart code that way, and we want augmentations to be as "normal" as possible.

Maybe it would help if we tried to get more concrete about the exact situations we are trying to avoid with this I could be convinced? When you consider that all top level identifiers coming from dependencies will be prefixed, that already eliminates many of the potential issues.

> I suppose we could say that identifiers are resolved as if the class appeared in an otherwise-empty top level namespace? That would let you resolve to instance and static members on the surrounding class, which is what you want, and that's about it.

You might want to refer to other _generated_ top level members as well. I think I would rather just specify that all macro augmentations are merged into a single file per library, and are all in the same scope. That is currently how I think we plan on presenting it to the user anyways.

If people generate code that resolves incorrectly or not at all, that is on them imo, but we shouldn't be overly prescriptive about exactly how they do it.

munificent: OK, I took another stab at this.

jakemac53 (approved):

PR #2086, opened 2022-01-31
Delete macro APIs, update DataClass example macro, add script to run it

jakemac53:
- The macro apis now live in the SDK, so delete them from here and update the links.
- Update the DataClass macro style a bit, and migrate to the latest API.
- Add a script to actually run the DataClass macro on a fake class, logs some timing information and the resulting augmentation library.


jakemac53:
Note that looking at the result of this macro brought up a few things that need to be fixed etc, I will start working on some of those.

jakemac53:
This should land after https://dart-review.googlesource.com/c/sdk/+/231041 which contains a few fixes

munificent (approved):

Issue #2093, opened 2022-02-04
How are macros passed names of declarations to produce?


munificent:
Some macros might want to create a declaration whose name is chosen by the caller. The approach we're currently leaning towards is that the macro takes the name as an argument, like:

```dart
@CreateClass(Foo)
library;
```

Here, the `@CreateClass` macro generates a new class whose name is `Foo`, based on the argument passed to the macro. The argument is syntactically just a identifier expression.

### When are macro arguments resolved?

That raises the question of what that `Foo` expression *means*. In most cases, the macro argument expression is treated as an actual expression that when interpolated into generated code produces the result of evaluating that expression. For example:

```dart
var foo = 'global';

@GeneratePrint(foo)
void f() {}
```

Here, this contrived macro generates a function body that looks like:

```dart
{
  var foo = 'local';
  print(<macro argument>);
}
```

In the example above, it generates this for `f()`:

```dart
{
  var foo = 'local';
  print(prefix1.foo);
}
```

Where `prefix1` is an import prefix of an import to this same library so that the expression reliably refers to the actual top-level `foo` declaration. This way it doesn't inadvertently instead evaluate to a reference to the *local* variable `foo` in the body generated by the macro.

In other words, identifiers in macro arguments expressions are interpreted as *resolved references to actual declarations* and not simply meaningless syntactic identifiers that are resolved after they get inserted into generated code.

### Macro arguments for names

That interpretation doesn't naturally make sense for the first example. In that example there *is* no `Foo` declaration that the macro argument can resolve to. It doesn't exist.

The current thinking is that that's OK. The implementation of the `@CreateClass` will be careful to access the *name* of that identifier argument and insert that as bare syntax into the class declaration it's generating. Then the macro generates a class with that name. And now *after* the macro has run, the identifier expression being passed to `@CreateClass` does exist. Because `@CreateClass` created it. Once the macro is done, now the identifier can be resolved and things like go to definition in an IDE can take you from that macro argument to the generated class.

This feels honestly pretty sketchy to me. I understand that it's consistent with other places in hand-authored code where users can refer to identifiers that won't exist until macros have run. But this feels different because the identifier that refers to a non-existent declaration is *passed to the very macro that creates it*.

We're passing an argument that isn't meaningful until after the macro receiving it has run. It looks like we're passing *a reference to the thing being declared* to the macro but we're actually passing a reference to the thing it *will* declare, which the macro then conveniently conjures into being.

### Macro arguments for non-top level names

[...]

Issue #2092, opened 2022-02-03
Allow collections of Identifiers, type literals, and literals in macro constructor parameters


srawlins:
The macro spec currently says:

> [...] macros can also accept arguments as values. However, only built-in value types (int, bool, etc.) are allowed and arguments must be _simple literal expressions_.

I think we should support "collections" of built-in value types and (adjascent or as part of https://github.com/dart-lang/language/issues/2091) type literals and identifiers. "Collections" means Lists, Sets, and Maps, though I don't have thoughts on any type restrictions on the key of a Map (maybe just String? Maybe no restriction beyond "only built-in value types"?).

For example:

```
@GenerateSomething([Foo, Bar])
```

which allows access to `Foo` and `Bar` as Identifiers, or as type literals, etc.

This raises a question about spreads. Today, the typical way to write an annotation which includes a list literal which folds in elements from a constant defined elsewhere is a spread. Can we allow:

```
const classes = [Foo, Bar];
@GenerateSomething([...classes, Baz])
```

Allowing this would be consistent with the support that constant evaluation provides, but might be extremely difficult / impossible because it _requires_ constant evaluation. Maybe we could sidestep constant evaluation by saying only a subset of constant values are supported in a collection of type literals or `Identifiers`...

cc @munificent @jakemac53

jakemac53:
Yep I was just about to file this same issue :)

jakemac53:
This is allowed now

Issue #2012, opened 2021-12-06
Macro argument scopes (specifically, Code objects), and augmentation libraries


jakemac53:
We want to be able to model the Macro output in terms of an augmentation library, and in general we want the code output by macros to not have special powers.

At the same time, we allow `Code` arguments to macro constructors, and we want the _scope_ of that code to be essentially the same scope as the macro annotation. However, we also possibly want to be able to refer to _local variables_ within those expressions.

Example:

```dart
abstract class Thing {
  int get x;
  int get y;
  
  @MemoizedVar(x + y) // `x` and `y` here are local variables. 
  int get z;
}
```

But what if `z` is a function, and it has parameters named `x` or `y`? We need to clarify whether the macro arguments scope is that of the _body_ of the declaration it annotates, or the class, or something else more custom.

Some factors to be taken into consideration:

- It might be useful for a macro arguments to be able to access the arguments to a function.
- A macro on `z` can emit code for declarations outside the scope of the body of `z` (for instance it can add fields and methods to the surrounding class and/or library, or add whole new classes).
- Macros may annotate static methods/fields on classes.

There are a few possible solutions I have been considering to this:

## Solution 1: No access to local variables or params in macro argument expressions

The scope is the lexical scope of the body of the class, static members of the class are visible (possibly without being qualified by `ClassName.`?), but no local variables.

This could be modeled by generating a static getter next to the original annotated declaration:

```dart
library augment 'original.dart';

augment class Thing {
  static  get _$someGeneratedName => x + y;
}
```

### Pros

- Easiest to define.
- Valid anywhere in the library (may need some munging to add the qualified `ClassName.` in front, or just require it).

### Cons

- Very limiting, can't access parameters or local variables.

## Solution 2: Its own lexical scope, as if the expression was inside the body of a new declaration

We could model this in the same way as Solution 1, except the getter wouldn't always be static - it would only be static if the annotated declaration was static.

### Pros

- I think this mostly behaves how users would expect, the scope is essentially the same as the annotated declaration.
- Allows access to local variables.

### Cons

- Possibly a bit magical, might be unexpected that expressions in macro annotations on static fields have a different scope.
- These expressions cannot be used in parent scopes (unless the annotated declaration was static).

## Solution 3: The scope is the same as the scope at the top of the body of the annotated declaration.

We could again model this in essentially same way, except with a local function at the top of the body of the annotated declaration.

```dart
library augment 'original.dart';

augment class Thing {
  int get z {
[...]

PR #2094, opened 2022-02-04
clarify identifier scope, allow bare Identifiers, allow Lists and Maps

jakemac53:
Attempt to close https://github.com/dart-lang/language/issues/2093, and https://github.com/dart-lang/language/issues/2092. Related to https://github.com/dart-lang/language/issues/2012.

- Adds `Identifier`, `List`, and `Map` as valid parameter types for macro constructors (and thus valid arguments for macro applications).
  - List and Map are allowed to have type arguments that are any of the supported types. This allows for `List<Identifier>`, etc.
- Specify the scope for identifiers in macro application arguments better (both bare and in code objects).
- Some other unrelated cleanup (can remove if desired).
  - Fixed up some old links
  - Removed the section on `Fragment` (you can just use `Code` for this now).


working/macros/feature-specification.md
```diff
@@ -118,6 +118,28 @@ Most of the time, like here, a macro takes the arguments you pass it and
...
+  // ....
+}
+
+class MyTypeDeserializer implements Deserializer<MyType> {
+  // ....
+}
+```
+
+Here the macro takes an `Identifier` argument, and introspects on it to know
+how to generate the desired serialization and deserialization classes.
```
munificent, line 141: Might be good to show a little more than `// ....` to hint at how the introspection is involved. Right now, it just looks like `MyType` is propagated over to the type argument in the generated class, which doesn't require introspection.

working/macros/feature-specification.md
```diff
@@ -871,23 +882,32 @@ Each argument in the metadata annotation for the macro application is converted
...
-    adjacent strings.
+*   If the parameter type is `bool`, `double`, `int`, `Null`, `num`, `String`,
+    `List`, or `Map`, (or the nullable forms of any of those), then the argument
+    expression must be a Boolean, number, null, string, list, or map literal.
+
+    * Number literals may be negated.
+    * String literals may not contain any interpolation, but may be adjacent
+      strings.
+    * List and Map literals may only contain any of the supported parameter
+      types. If the parameter type specifies a generic type argument, it must
```
munificent, line 893: "parameter types" -> "parameter types, recursively".

working/macros/feature-specification.md
```diff
@@ -896,6 +916,28 @@ It is a compile-time error if an macro class constructor invoked by a macro
...
 
+#### Identifier Scope
+
+The following rules apply to any `Identifier` passed as an argument to a macro
+application, whether as a part of a `Code` expression or directly as an
+`Identifier` instance.
+
+The scope of any `Identifier` argument is the same as the scope in which the
+identifier appears in the source code, which is the same as the scope for any
+other type of annotation on a declaration. This means:
```
munificent, line 927: "for any other type of annotation" -> "argument scope for a metadata annotation".

working/macros/feature-specification.md
```diff
@@ -896,6 +916,28 @@ It is a compile-time error if an macro class constructor invoked by a macro
...
+
+* Identifiers in macro application arguments may only refer to static and top
+  level members.
+* They cannot refer to local or instance variables, as those can never be in scope
+  where a macro application appears.
+* Identifiers referring to static class members may be unqualified if the
```
munificent, line 933: You mention "may be unqualified" here. Does this mean some identifiers *may* be qualified? Do we support:

```dart
@MyMacro(some_prefix.SomeClass)
@MyMacro(SomeClass.someStaticMember)
```

?

working/macros/feature-specification.md
```diff
@@ -896,6 +916,28 @@ It is a compile-time error if an macro class constructor invoked by a macro
...
+  annotation appears on a member of that class.
+
+All identifiers passed to macro constructors must resolve to a real declaration
+by the time macro expansion has completed. They may resolve to generated
+identifiers, including ones generated by the macro they were passed to,
+although that design may be inadvisable.
```
munificent, line 939: How about adding a TODO here pointing to https://github.com/dart-lang/language/issues/2093 to resolve this?

munificent (approved):

working/macros/feature-specification.md
jakemac53: Yes, both should be supported. I can add a section clarifying how these appear to the macro author (in both cases they would only see the unqualified name).

working/macros/feature-specification.md
jakemac53: In general I left off because I don't think the spec is the best place to be telling people what patterns to use (some other user facing doc for that would be more discoverable). So the spec just explicitly outlines that possibility, making it clear that you _can_ do it in case it affects implementations, even though I don't expect it to.

working/macros/feature-specification.md
```diff
@@ -871,23 +891,32 @@ Each argument in the metadata annotation for the macro application is converted
...
-    (or the nullable forms of any of those), then the argument expression must
-    be a Boolean, number, null, or string literal. Number literals may be
-    negated. String literals may not contain any interpolation, but may be
-    adjacent strings.
+*   If the parameter type is `bool`, `double`, `int`, `Null`, `num`, `String`,
+    `List`, or `Map`, (or the nullable forms of any of those), then the argument
+    expression must be a Boolean, number, null, string, list, or map literal.
+
+    * Number literals may be negated.
+    * String literals may not contain any interpolation, but may be adjacent
```
srawlins, line 899: How about raw Strings? (Multi-line is probably unnecessary to specify.)

working/macros/feature-specification.md
```diff
@@ -871,23 +891,32 @@ Each argument in the metadata annotation for the macro application is converted
 to a form that the corresponding constructor on the macro class expects, which
 it specifies through parameter types:
 
-*   If the parameter type is `bool`, `double`, `int`, `Null`, `num`, or `String`
-    (or the nullable forms of any of those), then the argument expression must
-    be a Boolean, number, null, or string literal. Number literals may be
-    negated. String literals may not contain any interpolation, but may be
-    adjacent strings.
+*   If the parameter type is `bool`, `double`, `int`, `Null`, `num`, `String`,
+    `List`, or `Map`, (or the nullable forms of any of those), then the argument
```
srawlins, line 895: How about Set?

working/macros/feature-specification.md
srawlins: I'll add a third, one with two dots:

`@MyMacro(some_prefix.SomeClass.someStaticMember)`

working/macros/feature-specification.md
```diff
@@ -871,23 +891,32 @@ Each argument in the metadata annotation for the macro application is converted
...
+    expression must be a Boolean, number, null, string, list, or map literal.
+
+    * Number literals may be negated.
+    * String literals may not contain any interpolation, but may be adjacent
+      strings.
+    * List and Map literals may only contain any of the supported parameter
```
srawlins, line 901: Am I allowed to pass an argument like `[false, 7]`? This would require being able to use a parameter of type `List<Object>`. I think the restrictions on type arguments for Lists, Maps, (and Sets) are currently unspecified.

srawlins (approved):

working/macros/feature-specification.md
jakemac53: Should I clarify raw strings? I don't know if there is any consideration that makes them different from any other String literal.

working/macros/feature-specification.md
jakemac53: Hmm, good question. I am not sure what the use case really would be given that you can't combine collections at all (at least for now). I am inclined to just leave it off to simplify things? But it wouldn't cause any issue to allow it afaik.

working/macros/feature-specification.md
jakemac53: Interesting point, maybe we should allow `List<Object>` (or any built in type?) for the generic types. But then the values must still conform to the rules?

working/macros/feature-specification.md
srawlins: I have no strong opinion. I just put raw strings on the same level as adjacent strings, which are commented on.

working/macros/feature-specification.md
srawlins: We use them in mockito at least so that in the IDE, users see when they've accidentally duplicated something.

working/macros/feature-specification.md
srawlins: Yeah that's not a bad idea.

jakemac53: Only enabling `List<Object>` for now, I think that is simple enough.

working/macros/feature-specification.md
jakemac53: I added it, why not 🤷‍♂️

working/macros/feature-specification.md
jakemac53: I went ahead and added it

PR #2102, opened 2022-02-08
Remove reference to being able to declare `values`

lrhn:
When removing the ability to declared a `values` member, I forgot one reference to it.
Also rephrasing that sentence for clarity.

Att: @scheglov


accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -137,7 +137,7 @@ If the resulting class would have any naming conflicts, or other compile-time er
...
 - Declaring or mixing in a member which is not a valid override of a super-interface member declaration, including, but not limited to, the `index` and `toString` members of `Enum`.
 - Declaring or inheriting an member signature with no corresponding implementation. _(For example declaring an abstract `Never get index` or `String toString([int optional])`, but not providing an implementation.)
-- Declaring a type parameter on the `enum` which does not have a valid well-bounded or super-bounded instantiate-to-bounds result *and* not declaring or inheriting a member with base-name `values` _(because the then automatically introduced `static const List<EnumName> values` requires a valid instantiate-to-bounds result which is at least super-bounded, and a value declaration may require a well-bounded instantiation)_.
+- Declaring a type parameter on the `enum` which does not have a valid well-bounded or super-bounded instantiate-to-bounds result when such a result is needed _(the automatically introduced `static const List<EnumName> values` requires a valid instantiate-to-bounds result which is at least super-bounded, and a value declaration may require a well-bounded instantiation)_.
```
eernstg, line 140: Might as well change 'well-bounded or super-bounded' to 'well-bounded', because well-bounded means 'regular-bounded or super-bounded' already. Also, instantiation to bounds is applied to a complete type parameter declaration list, so it's the generic type that has an instantiation to bounds or not, not a single type parameter. 

I think there's no need to mention a single type parameter here, it would work just fine to say 'Declaring a generic `enum` that does not have an instantiation to bounds (*because the algorithm yields a type which is not well-bounded*).'

With the new treatment of `values`, it _will_ be needed.

I think it would be fair to say that it is not an error on the enum declaration as a whole if there is a value where the instantiation to bounds causes an error because it is super-bounded: The developer could just fix that particular value declaration by giving suitable type arguments.

eernstg (approved):
LGTM, suggesting a couple of further changes.

PR #2106, opened 2022-02-09
Adjust `primitive operator ==` to `primitive equality`, and include `hashCode`

eernstg:
This PR updates the language specification such that the notion of having a 'primitive operator `==`' is replaced by the notion of having 'primitive equality', and it adds constraints on the objects and classes such that they only have primitive equality if they satisfy some constraints on operator `==` (same as before) and on `hashCode` (which is new).

The point is that we wish to maintain that equality as a whole (involving operator `==` as well as `hashCode`) can be relied upon to have a behavior which is known by language processing tools (such as compilers and analyzers). The rules of today (before this PR) allow `hashCode` to be overridden arbitrarily, and this implies that constant sets and non-constant hash sets would have different semantics in the case where the overridden `hashCode` is inconsistent with operator `==`, which seems to be a mal-feature for everybody.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -38,6 +38,8 @@
 % Dec 2022
 % - Change the definition of the type function 'flatten' to resolve soundness
 %   issue, cf. SDK issue #49396.
+% - Change 'primitive operator ==' to 'primitive equality', and include
+%   constraints on `hashCode` corresponding to the ones we have on `==`.
 %
 % 2.15
 % - Allow generic instantiation of expressions with a generic function type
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -17781,8 +17785,8 @@ \subsection{Switch}
 }
 
 \LMHash{}%
-It is a compile-time error if the operator \lit{==} of class $C$
-is not primitive
+It is a compile-time error if the class $C$
+does not have primitive equality
```
lrhn, line 17789: This is where we actually only need primitive `operator ==`, because a switch is not expected to use `hashCode`.

I can see that it's simpler to just have one notion, but I think it can be relevant to have, say, an `enum` where `hashCode` defers to `index` for speed, rather than rely on `identitiyHashCode` (which is not that efficient), but where you still want to allow switching on the enum values.

lrhn (commented):
LGTM, but I would want to consider once more whether we want to separate "primitive hashCode" from "primitive operator==", so that switches can work on classes with non-primitive `hashCode`.

specification/dartLangSpec.tex
eernstg: That is possible, but somewhat speculative: If two enum declarations do this then those objects could never be allowed to occur in the same set. I suspect that this would be a footgun that nobody would want to have in a serious software product.

lrhn: There should be no problem having them in the same set, as long as it's not a constant set literal.
An implementation of `int get hashCode => 4;` is *valid*, just not necessarily that efficient.

eernstg:
Rebased, refreshed this PR. @lrhn, PTAL.

specification/dartLangSpec.tex
```diff
@@ -3282,10 +3285,11 @@ \subsubsection{The Operator `=='}
 \end{itemize}
```
lrhn, line 3285: If I'm going to nit, `bool` instances likely override `hashCode`.

"evaluation of a list literal", should it be "of a constant list literal, ... a constant map literal ..., or a constant set literal"?

specification/dartLangSpec.tex
```diff
@@ -3233,39 +3235,40 @@ \subsubsection{The Method \code{noSuchMethod}}
...
   if it does not have an implementation of the operator \lit{==}
+  that overrides the one inherited from \code{Object},
+  and it does not have an implementation of the getter \code{hashCode}
   that overrides the one inherited from \code{Object}.
   \commentary{%
-  In particular, the following have a primitive operator \lit{==}:
+  In particular, the following have primitive equality:
   The null object (\ref{null}),
   function objects obtained by function closurization of
   a static method or a top-level function
```
lrhn, line 3274: Now we have instantiated tear-off, do we need to mention `possibly instantiated with constant types, if generic"?

lrhn (approved):

specification/dartLangSpec.tex
eernstg: Adjusted the text to specify directly that `bool`, const-able function closurizations, and collection literals have primitive equality. Then we don't need to worry about whether or not `bool` overrides `hashCode`.

eernstg (commented):
Some further changes: Moved all normative text about having primitive equality into the section '... Primitive Equality', replaced text about this topic (in section 'Booleans', and in the section about function closurization, etc.) by commentary, referring to section '... Primitive Equality'. Also added a rationale paragraph, indicating that the concept of primitive equality is relevant because of constant expression evaluation.

specification/dartLangSpec.tex
```diff
@@ -3233,59 +3235,81 @@ \subsubsection{The Method \code{noSuchMethod}}
...
-  The null object (\ref{null}),
-  function objects obtained by function closurization of
-  a static method or a top-level function
-  (\ref{functionClosurization}),
-  instances of type \code{bool}
-  (\ref{booleans}),
-  and instances obtained by evaluation of a list literal
+  has primitive equality.
+\item
+  Let $o$ be an object obtained by evaluation of a list literal
```
lrhn, line 3281: You say "constant expressions" above, but not here.
I'd say "constant list literal", etc., just to avoid possible doubt.

eernstg: Done.

specification/dartLangSpec.tex
```diff
@@ -3233,59 +3235,81 @@ \subsubsection{The Method \code{noSuchMethod}}
...
   a set literal
-  (\ref{sets}).%
-  }
+  (\ref{sets}),
+  then $o$ has primitive equality.
+\item
+  A function object obtained by function closurization of
+  a static method or a top-level function
+  (\ref{functionClosurization})
+  has primitive equality.
```
lrhn, line 3292: Probably need to say "constant" here too.
```dart
T foo<T>(T value) => value;

void bar<X>() {
  X Function(X) f = foo; // implicit instantiated tear-off.
}
```
The `foo` is not constant, but probably also doesn't have primitive equality.

lrhn (approved):

specification/dartLangSpec.tex
eernstg: Now requiring that it is a constant expression.

PR #2116, opened 2022-02-15
README.md updated

yashjani01:
Minor, but important phrases updated.


leafpetersen (changes requested):
Thanks for the suggestions!  I think in general though I prefer the original wording here.

munificent:
I agree with Leaf. I appreciate the suggestions but I like the deliberately more informal tone of the current wording.

PR #2118, opened 2022-02-16
Disallow overriding `==` and `hashCode`.

lrhn:


accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -29,7 +29,7 @@ enum Name<T extends Object?> with Mixin1, Mixin2 implements Interface1, Interfac
 where `memberDeclaration*` is almost any sequence of static and instance member declarations, or constructors, 
 with some necessary restrictions specified below.
 
-The `;` after the identifier list is optional if there is nothing else in the declaration (for backwards compatibility), and required if there is any member declaration after it. The identifier list may have a trailing comma in either case (like now).
+The `;` after the value list is optional if there is nothing else in the declaration (for backwards compatibility), and required if there is any member declaration after it. The identifier list may have a trailing comma in either case (like now).
```
mateusfccp, line 32: Shouldn't this be also changed to `value`?

```suggestion
The `;` after the value list is optional if there is nothing else in the declaration (for backwards compatibility), and required if there is any member declaration after it. The value list may have a trailing comma in either case (like now).
```

accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -152,20 +155,18 @@ It’s currently a compile-time error for a class to implement, extend or mix-in
...
+- It's a compile-time error if a `class`, `mixin` or `enum` declaration has `Enum` as a superinterface (trivially true for `enum` declarations), and that declaration contains non-abstract instance member declaration named `index`, `hashCode` or `==` (an `operator ==` declaration). _That `index` member would override the `index` getter inherited from `Enum`, and we currently do not allow that. The `hashCode` and `operator==` declarations would prevent the enum class from having "primitive equality", and we want to ensure that enums can be used in switches._
 
-Those restrictions allows abstract classes (interfaces) which implements `Enum` in order to have the `int index;` getter member available, and it allows `mixin` declarations to use `Enum` as an `on` type because `mixin` declarations cannot be instantiated directly.
+Those restrictions allow abstract classes (interfaces) which *implements* `Enum` in order to have the `int index;` getter member available, and also allow `mixin` declarations to use `Enum` as an `on` type because `mixin` declarations cannot be instantiated directly.
```
leafpetersen, line 163: implements -> implement

leafpetersen (approved):
LGTM, thanks!

leafpetersen:
cc @scheglov @johnniwinther @nshahan @rakudrama

accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -245,13 +246,13 @@ The enum values can call the declared constructors, or the default unnamed zero-
...
 - The `enum` types themselves are completely sealed. No other class can implement an `enum` type.
 
 - Because of that, `enum` types support exhaustiveness checking in `switch` cases in the language _(meaning that flow-control can see that an exhaustive switch over enum values cannot pass through without executing at least one `case`, which can then affect variable promotion)_.
 - The `EnumName.name` extension member works on `enum` values.
 
-If the *restrictions* (the type is sealed, there is only a finite, enumerable number of instances, and the class  implements `Enum`, so it must have an `int index` getter and no `values` member), are acceptable, there should no longer be any reason to *not* make your enum-like class a language-based `enum`.
+If the *restrictions* (the type is sealed, there is only a finite, enumerable number of instances, and the class  implements `Enum`, so it must have an `int index` getter and no `values` member, and it cannot override `index`, `hashCode` or `operator==`), are acceptable, there should no longer be any reason to *not* make your enum-like class a language-based `enum`.
```
nshahan, line 255: > ...so it must have an `int index` getter and no `values` member, and it cannot override `index`...

Must have `int index` and cannot override `index` sounds conflicting.

nshahan:
cc @Markzipan

accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -152,20 +155,18 @@ It’s currently a compile-time error for a class to implement, extend or mix-in
...
-- It's a compile-time error if a `class`, `mixin` or `enum` declaration has `Enum` as a superinterface, and it declares a non-abstract instance member named `index`. _That member would override the `index` getter inherited from `Enum`, and we currently do not allow that._
+- It's a compile-time error if a `class`, `mixin` or `enum` declaration has `Enum` as a superinterface (trivially true for `enum` declarations), and that declaration contains non-abstract instance member declaration named `index`, `hashCode` or `==` (an `operator ==` declaration). _That `index` member would override the `index` getter inherited from `Enum`, and we currently do not allow that. The `hashCode` and `operator==` declarations would prevent the enum class from having "primitive equality", and we want to ensure that enums can be used in switches._
```
eernstg, line 161: A couple of changes would fit here, for consistency:
```suggestion
- It's a compile-time error if a `class`, `mixin` or `enum` declaration has `Enum` as a superinterface (*trivially true for `enum` declarations*), and that declaration contains a concrete instance member declaration named `index`, `hashCode` or operator `==`. _That `index` member would override the `index` getter inherited from `Enum`, and we currently do not allow that. The `hashCode` and operator `==` declarations would prevent the enum class from having "primitive equality", and we want to ensure that enums can be used in switches._
```

However, there is another issue: This item repeats rules that have already been stated (in line 134 and 136), and this normally implies that the duplicate should be commentary. A similar issue arises for `values` in line 160.

accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -152,20 +155,18 @@ It’s currently a compile-time error for a class to implement, extend or mix-in
...
+The restrictions still ensure  `enum` values are the only object instances which implements `Enum`, while making it valid to declare `abstract class MyInterface implements Enum` and `mixin MyMixin on Enum` for interfaces and mixins intended to be used in declaring `enum` classes. It's also impossible to override the instance getters `index` , `hashCode`, or the operator `==`, and impossible to interfere with the static  `values` getter, without causing a compile-time error. _For example, if an enum declaration implements an interface containing the signature `Never get index;`, then because it's not possible to override `int get index;` from `Enum`, the resulting class will not implement its interface and that is a compile-time error._
```
eernstg, line 165: The paragraph as a whole seems to be commentary in nature.
```suggestion
_The restrictions still ensure that `enum` values are the only objects whose run-time type implements `Enum`, while making it valid to declare `abstract class MyInterface implements Enum` and `mixin MyMixin on Enum` for interfaces and mixins intended to be used in `enum` declarations. It's also impossible to override the instance getters `index` , `hashCode`, or the operator `==`, and impossible to interfere with the static  `values` getter, without causing a compile-time error. For example, if an enum declaration implements an interface containing the signature `Never get index;`, then because it's not possible to override `int get index;` from `Enum`, the resulting class will not implement its interface and that is a compile-time error._
```

accepted/future-releases/enhanced-enums/feature-specification.md
eernstg: I think we can just avoid mentioning that `index` must be there (that's implied by the following 'cannot override `index`' anyway). With a couple of other adjustments:

```suggestion
If the *restrictions* are acceptable (the type is sealed; there is only a finite, statically known set of instances; the class  implements `Enum`; it cannot override `index`, `hashCode` or operator `==`; and it cannot have a user-written member named `values`), there should no longer be any reason to *not* make your enum-like class a language-based `enum`.
```

eernstg (approved):
LGTM, with various suggested adjustments.

accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -124,19 +123,23 @@ The semantics of such an enum declaration, *E*, is defined as introducing a (sem
 
   where `args` are considered as occurring in a `const` context, and it’s a **compile-time error** if they are then not compile-time constants.
 
-The resulting constructor invocations are subject to type inference, using the empty context type. *This implies that inferred type arguments to the constructor invocation itself may depend on the types of the argument expressions of `args`.* The type of the constant variable is the static type of the resulting constant object creation expression.
+  The resulting constructor invocations are subject to type inference, using the empty context type. *This implies that inferred type arguments to the constructor invocation itself may depend on the types of the argument expressions of `args`.* The type of the constant variable is the static type of the resulting constant object creation expression.
```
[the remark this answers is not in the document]
lrhn: ... is **worst** kind of whitespace!

accepted/future-releases/enhanced-enums/feature-specification.md
lrhn: This actually *doesn't* duplicate the earlier errors.

This error only apply to *non-abstract declarations* of `index`/`hashCode`/`==` in a `class`/`mixin`/`enum` which implements `Enum`.

The errors in lines 134/136 apply to an `enum` declaration which declares or inherits `index`/`hashCode`/`==`.

So, the case of an `enum` with a mixin application of a mixin *not* implementing `Enum`, but declaring a non-abstract `index`/`hashCode`/`==`, is only covered by the earlier errors, not this one.

And that's probably an oversight, because we don't cover the case of:
```dart
class C {
  int get index => 0;
}
abstract class D extends C implements Enum {}
```
either.

It's actually not particularly *important* that a `class` implementing `Enum` has a concrete implementation of anything, because there is currently no way to ever get that implementation into an `enum` declaration.
(This error catches the class itself declaring `index`, etc., and if it's declared in a superclass, then the class has a superclass other than `Objecf` and cannot be used as a mixin.)

So, how strict should we be about classes implementing `Enum` and having implementations. Those implementations are guaranteed dead code, so we could choose to refuse it entirely.
Or we could just not care (status quo), since you'll probably figure it out if you ever try to use the class for anything except `implements` its interface.

lrhn: I actually prefer "non-abstract" to "concrete" because being `abstract` is the actual language term.

eernstg: The language specification has 58 occurrences of `concrete` and zero occurrences of 'non-abstract', so we do have prior art to support the notion that 'concrete' is well-defined. We could switch over, of course, but 'non-abstract' does seem somewhat bulky and hard to pronounce.

lrhn: The
> instance member declaration named `index`, `hashCode` or operator `==`

phrase doesn't work for me. There is no "instance member declaration named ... operator `==`".
It's either *named* `==` or it's an `operator ==` declaration.

accepted/future-releases/enhanced-enums/feature-specification.md
eernstg: The language specification uses `\IMPLEMENTS` as a reference to the actual syntax and 'implements' as a separate concept (such that `List<int>` implements `Iterable<int>` and `Object`, but not `Iterable<num>`). In this context I think we could use 'abstract classes that implement `Enum`' and remain consistent with the way this word is used in the spec.

We could include the word 'interface', but I think that's more opinion than information. ;-)

accepted/future-releases/enhanced-enums/feature-specification.md
eernstg: NB: I edited several 'suggestions' above to fit the results of the discussion at this point. Added `[Edit: ...]` there.

> The errors in lines 134/136 apply to an `enum` declaration

Indeed, I forgot about that distinction even though it came up earlier in discussions with the UniPro folks. ;-)

I changed the edit suggestion such that the overlap is reduced: The rule here will then apply to classes and mixins, not to enum declarations.

However, these error specifications should occur right next to each other, it is too hard to read the document when they are so far apart from each other.

> So, the case of an `enum` with a mixin application of a mixin not implementing `Enum`, but declaring a non-abstract `index`/`hashCode`/`==`, is only covered by the earlier errors, not this one.

Right, and we do have cases where line 161 applies and line 134/136 do not:

```dart
mixin M on Enum {
  int get index => 0; // Error.
}

abstract class B implements Enum {
  int get index => 0; // Error.
}

class C {
  int get index => 0;
}
abstract class D extends C implements Enum {} // Error.
```

I suggested that we should use the phrase '.. C has a concrete instance member `m`..', meaning that the superclass chain of `C` contains an element `B` (which could be `C` or some other class, possibly a mixin application) such that `B` declares a concrete instance member named `m`.

There is a bit of a conceptual clash here, because 'has'/'declares' is usually used with interfaces, that is, with member signatures and not actual declarations, but I think it is quite readable and not ambiguous in practice.

eernstg (approved):
Further comments in several threads.

accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -124,19 +123,23 @@ The semantics of such an enum declaration, *E*, is defined as introducing a (sem
...
 
 - **Static `values` list**: A static constant variable named `values` is added as by the declaration  `static const List<Name> values = [id1, …, idn];`
   where `id1`…`idn` are the names of the enum entries of the `enum` declaration in source/index order.
   _If `Name` is generic, the `List<Name>` instantiates `Name` to its bounds._
 
+It's a **compile-time error** if an `enum` declaration declares or inherits (from a mixin application) a non-abstract member named  `index` which overrides the `index` getter inherited from the `Enum` class.
+
+It's a **compile-time error** if an `enum` declaration declares or inherits (from a mixin application) a non-abstract member named  `hashCode` or `==` (an `operator ==` declaration) which overrides the `hashCode` getter or `==` operator inherited from the `Object` class. (The `Enum` class does not override `hashCode` or `operator==` from `Object`). _This ensures that enum values can be used as switch statement case values, which is the main advantage of using an enum over just writing a normal class._
```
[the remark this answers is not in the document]
lrhn: I'm also vary about using "has".

A declaration *contains* other declarations. It's actually even a stretch to say that a declaration *inherits* anything.
The semantic class/mixin/enum that the declaration introduces may have thing or inherit things, but if there is a compile-time error, then that class/mixin/enum does not exist, so it can't "have" anything.

So, to define things in terms of concepts that are well-defined at this point in the validation, I refer only to member declarations of the member itself, or the class/interface of its superclass/superinterfaces (which are assumed well-formed before we even start looking at this declaration). The "inherits" here is really a shorthand for "the superclass has X so the enum would inherit X, if it is otherwise valid".

accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -124,19 +123,23 @@ The semantics of such an enum declaration, *E*, is defined as introducing a (sem
...
 
-  The objects created here are *not canonicalized* like other constant object creations. _(In practice, the index value is considered part of the object, so no two objects will have the same state.)_
+   The objects created here are *not canonicalized* like other constant object creations. _(In practice, the index value is considered part of the object, so no two objects will have the same state.)_
 
 - **Static `values` list**: A static constant variable named `values` is added as by the declaration  `static const List<Name> values = [id1, …, idn];`
   where `id1`…`idn` are the names of the enum entries of the `enum` declaration in source/index order.
   _If `Name` is generic, the `List<Name>` instantiates `Name` to its bounds._
 
+It's a **compile-time error** if an `enum` declaration declares or inherits (from a mixin application) a non-abstract member named  `index` which overrides the `index` getter inherited from the `Enum` class.
```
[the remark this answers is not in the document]
lrhn: Agree. I've moved the mention of the mixin application to after the paragraph, and comment-italicized it.

accepted/future-releases/enhanced-enums/feature-specification.md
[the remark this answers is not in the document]
lrhn: Although in this case, it's probably workable, since if the enum declaration is otherwise valid, we can *then* check if it has an `index`/`hashCode`/`==` instance member which overrides the ones from `Enum`/`Object`.

I still don't like "has". It's too vague.

accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -152,20 +154,18 @@ It’s currently a compile-time error for a class to implement, extend or mix-in
...
-- It's a compile-time error if a `class`, `mixin` or `enum` declaration has `Enum` as a superinterface, and it declares a non-abstract instance member named `index`. _That member would override the `index` getter inherited from `Enum`, and we currently do not allow that._
+- It's a **compile-time error** if a `class` or `mixin` declaration has `Enum` as a superinterface, and that class or mixin declares or inherits a concrete instance member named `index`, `hashCode` or `==` *(an `operator ==` declaration)*. _That `index` member could override the `index` getter inherited from `Enum`, and we currently do not allow that. The `hashCode` and `operator==` declarations would prevent the enum class from having "primitive equality", and we want to ensure that enums can be used in switches._
```
eernstg, line 160: A mixin actually cannot inherit a concrete instance member, because the superclass in any mixin application is only required to implement any `on` types of the mixin, an `on` type doesn't have to be a superclass.

But I guess that's benign, because it just means that the 'mixin' case "can't use the word 'inherits'", it doesn't introduce anything which is wrong.

eernstg (approved):
LGTM, resolving several threads, adding just one comment.

accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -93,7 +91,7 @@ The semantics of such an enum declaration, *E*, is defined as introducing a (sem
 
   It’s a **compile-time error** if such a mixin application introduces any instance variables. _We need to be able to call an implementation specific superclass `const` constructor of `Enum`, and a mixin application of a mixin with a field does not make its forwarding constructor `const`. Currently that’s the only restriction, but if we add further restrictions on mixin applications having `const` forwarding constructors, those should also apply here._
 
-* **Superinterfaces**: The immediate superinterfaces of *C* are the interface of the superclass and the interfaces declared by *E*.
+* **Superinterfaces**: The immediate superinterfaces of *C* are the interface of the superclass and the interfaces declared bu `implements` in *E*.
```
sgrekhov, line 94:
```suggestion
* **Superinterfaces**: The immediate superinterfaces of *C* are the interface of the superclass and the interfaces declared by `implements` in *E*.
```

PR #2139, opened 2022-03-04
Clarify rethrow

eernstg:
Cf. https://github.com/dart-lang/sdk/issues/48502.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -18032,28 +18032,35 @@ re-throw an exception and its associated stack trace.
 \end{grammar}
 
 \LMHash{}%
+Consider a \RETHROW{} statement $S$.
+% $f$ as defined below is guaranteed to exist: Every statement occurs in
+% a <block> (a statement may be part of another statement $S_2$, but by
+% induction hypothesis $S_2$ occurs in a <block>), and a <block> can only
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -18031,30 +18031,37 @@ \subsection{Rethrow}
...
+and every \CATCH{} clause is treated as some \CATCH{} clause of the form
+\code{\ON\,\,$T$\,\,\CATCH\,\,(p1,\,\,p2)}.
+So we can consider the \RETHROW{} statement to be enclosed in
 a \CATCH{} clause of that form.%
 }
 
 \LMHash{}%
-The \RETHROW{} statement throws (\ref{statementCompletion})
+The \RETHROW{} statement then throws (\ref{statementCompletion})
 with $p_1$ as the exception object and $p_2$ as the stack trace.
```
lrhn, line 18063: Ob-nit: It really throws the object bound to the variable p_1, not the variable itself.
I assume we use that shorthand elsewhere, and might even say that we do so, so it's probably fine.

lrhn (approved):

specification/dartLangSpec.tex
eernstg: Fixed!

PR #2145, opened 2022-03-07
Say that an enum constructor cannot use super-parameters.

lrhn:
Just to be explicit. It would be an error for the desugared class as well, but it's better to make it an error up-front.

### DIFF
--- a/accepted/future-releases/enhanced-enums/feature-specification.md
+++ b/accepted/future-releases/enhanced-enums/feature-specification.md
@@ -1,6 +1,6 @@
 # Dart Enhanced Enum Classes
 
-Author: lrn@google.com<br>Version: 1.7<br>Tracking issue [#158](https://github.com/dart-lang/language/issues/158)
+Author: lrn@google.com<br>Version: 1.8<br>Tracking issue [#158](https://github.com/dart-lang/language/issues/158)
 
 This is a formal specification for a language feature which allows `enum` declarations to declare classes with fields, methods and `const` constructors initializing those fields. Further, `enum` declarations can implement interfaces and apply mixins.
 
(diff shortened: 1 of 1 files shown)


accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -54,12 +54,16 @@ It is a **compile-time error** if the initializer list of a non-redirecting gene
 
 _We will introduce the necessary super-invocation ourselves as an implementation detail. From the user’s perspective, they extend `Enum` which has no public constructors. We could allow `super()`, which would then be a constructor of `Enum`, but it's simpler to just disallow super invocations entirely._
 
+It is a **compile-time error** if the argument list of a non-redirecting generative constructor includes a `super` parameter.
+
+_Super parameters (a feature introduced in the same release as enhanced enums) implicitly adds arguments to the `super` constructor invocation of a non-redirecting generative constructor, and such enum constructors do have a known superclass constructor expecting such arguments._
```
munificent, line 59: "do" -> "do not"?

lrhn: Potato, not-potato. Let's call the whole thing off!

(Good catch!)

accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -54,12 +54,16 @@ It is a **compile-time error** if the initializer list of a non-redirecting gene
 
 _We will introduce the necessary super-invocation ourselves as an implementation detail. From the user’s perspective, they extend `Enum` which has no public constructors. We could allow `super()`, which would then be a constructor of `Enum`, but it's simpler to just disallow super invocations entirely._
 
+It is a **compile-time error** if the argument list of a non-redirecting generative constructor includes a `super` parameter.
+
+_Super parameters (a feature introduced in the same release as enhanced enums) implicitly adds arguments to the `super` constructor invocation of a non-redirecting generative constructor, and such enum constructors do not have a known superclass constructor expecting such arguments._
```
eernstg, line 59: If line 57 reports that an error is known to occur (based on existing rules) then it should presumably be commentary.

accepted/future-releases/enhanced-enums/feature-specification.md
```diff
@@ -97,14 +101,14 @@ The semantics of such an enum declaration, *E*, is defined as introducing a (sem
 
 - **Declared members**: For each member declaration of the `enum` declaration *E*, the same member is added to the class *C*. This includes constructors (which must be `const` generative or non-`const` factory constructors.)
 
-- **Default constructor**: If no generative constructors were declared, and no unnamed factory constructor was added,
+- **Default constructor**: If no generative constructors were declared, and also no unnamed factory constructor was declared,
```
eernstg, line 104: The language specification considers the name of a constructor like `class C { C(); }` to be `C`, and the name of `class C { C.named(); }` to be `C.named`, for all kinds of constructors including factories. This means that we don't have to struggle with "unnamed" constructors. It doesn't hurt to introduce the name of the enum at this point, because it will be needed in line 108 anyway.

eernstg (approved):
LGTM, noting that line 57 ought to be commentary if it is a consequence of existing rules.

accepted/future-releases/enhanced-enums/feature-specification.md
lrhn: We can have this as a direct rule on super-parameters, rather than here, but I'd mention it here as a rule as well.
(In super parameters, we say that it's a compile-time error to have a super parameter in any constructor which is not a non-redirecting generative constructor, and we should add "of a `class` declaration" now that we can have constructors in `enum` declarations too.)

This is not *just* an error because the desugaring would have an error. It's that too, but not just.

eernstg: Right, that makes sense, we do want to have errors that are specified directly rather than "desugar this, and then backtrack any errors to the enum".

accepted/future-releases/enhanced-enums/feature-specification.md
eernstg: I do think that it works better to say that the names of constructors are `C` and `C.named` rather than "the unnamed constructor declared by the class `C`" and "a constructor named `named` which is declared by the class `C`", even though I do recognize that "declared by the class `C`" can be obtained from the context in some cases.

However, let's get this landed with one or the other terminology, the first thing we need is to ensure consistency within any single document.

eernstg: 👍

PR #2148, opened 2022-03-09
Only allow super parameters in `class` declarations.

lrhn:
Makes it explicit that you cannot use super-parameters in an `enum` declaration constructor..


working/1855 - super parameters/proposal.md
```diff
@@ -46,9 +46,9 @@ _That is, exactly the same grammar as initializing formals, but with `super` ins
 
 ## Semantics
 
-It’s a **compile-time error** if a super-parameter in any declaration other than a non-redirecting generative constructor.
+It’s a **compile-time error** if a super-parameter in any declaration other than a non-redirecting generative constructor of a `class` declaration.
```
eernstg, line 49:
```suggestion
It’s a **compile-time error** if a super-parameter declaration occurs in any declaration other than a non-redirecting generative constructor of a class declaration.
```

(I think there's ample tradition for using 'class' with no special font, so we might as well keep the visuals simple).

working/1855 - super parameters/proposal.md
```diff
@@ -46,9 +46,9 @@ _That is, exactly the same grammar as initializing formals, but with `super` ins
...
 
-_All non-redirecting generative constructors have a super-constructor invocation at the end of their initializer list. If none is written (or there is even no initializer list), the default is an invocation of `super()`, targeting the unnamed superclass constructor. It’s a compiler-time error if the superclass does not have the specified constructor, or it’s not a generative constructor._
+_All non-redirecting generative constructors of `class` declarations have a super-constructor invocation at the end of their initializer list. If none is written (or there is even no initializer list), the default is an invocation of `super()`, targeting the unnamed superclass constructor. It’s a compiler-time error if the superclass does not have the specified constructor, or it’s not a generative constructor. The new enhanced `enum` declarations, on the other hand, do have non-redirecting generative constructors, but those cannot not have any super-invocations, and they also have no known superclass constructor to forward parameters to._
```
eernstg, line 51:
```suggestion
_All non-redirecting generative constructors of a class declaration have a super-constructor invocation at the end of their initializer list. If none is written (or there is even no initializer list), the default is an invocation of `super()`, targeting the unnamed superclass constructor. It’s a compiler-time error if the superclass does not have the specified constructor, or it’s not a generative constructor. The new enhanced `enum` declarations, on the other hand, do have non-redirecting generative constructors, but those cannot not have any super-invocations, and they also have no known superclass constructor to forward parameters to._
```

(Yeah, well, we do use `enum` in many cases with the backquotes, presumably because we haven't found a real word for that concept, and `enum` is actually programming language syntax rather than a natural language word.)

eernstg (approved):
LGTM, suggesting a couple of tiny terminology changes.

PR #2149, opened 2022-03-09
Fix </code> tag in the `enhanced-enums` spec

creativecreatorormaybenot:
This one typo previously destroyed all formatting in the document after the occurence.

cc @lrhn


lrhn (approved):
Thanks!

PR #2163, opened 2022-03-22
Introduce rules to say that missing formal parameter type means dynamic

eernstg:
The language specification definition of function declarations in section 'Function Declarations' has not been updated for a long time. Unfortunately, the definition is inconsistent with the usage of that phrase elsewhere in the spec. With this PR, we're consistently considering a function declaration as a broad concept (e.g., it includes constructors, operators, getters, and external functions).

With a consistently broad definition of function declaration in place, this PR then introduces the rule (in section 'Formal Parameters') that an omitted type annotation on a formal parameter yields the declared type `dynamic`, and also mentions that the declared type isn't omitted if it is provided by override inference.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -1564,12 +1564,16 @@ Functions abstract over executable actions.
 
 \LMHash{}%
 Functions can be introduced by function declarations
-(\ref{functionDeclarations}),
+(\ref{functionDeclarations}).
+\commentary{%
+This includes
 method declarations (\ref{instanceMethods}, \ref{staticMethods}),
 getter declarations (\ref{getters}),
 setter declarations (\ref{setters}),
-and constructor declarations (\ref{constructors});
-and they can be introduced by function literals (\ref{functionExpressions}).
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -1868,7 +1853,25 @@ \subsection{Formal Parameters}
...
+Type inference is not yet specified in this document,
+and is assumed to have taken place already
+(\ref{overview}).
+In particular, some or all formal parameters of an instance method declaration
+may have received a declared type based on override inference,
+in which case it is not considered to be omitted.
+%% TODO(eernst): Change to \ref{} when inference spec is added.
+Briefly, override inference yields a type computed from the declared types of
+the corresponding parameter of member signatures in superinterfaces,
+or it causes a compile-time error if said types can not be reconciled.%
```
munificent, line 1871: Do you also want to mention inference from initializing formals here too since you include constructors under this umbrella now?

eernstg: Yes, thanks!

PR #2175, opened 2022-03-30
Clarify the evaluation order for collection literal elements

eernstg:
The evaluation order for the elements of a collection literal (say, `[e1, ...e2, if (b) e3]` where `e2` is a set) was not mentioned. This PR adds a few words to clarify that evaluation occurs in textual order. The order of evaluation associated with the subsequences with other than a single element is already specified (e.g., the subsequence for a `Set` spread is obtained by a pseudo-code for-in statement that explicitly states how each object is obtained).

It also corrects a couple of typos where an empty sequence followed by a semicolon was shown as `[[;]]` rather than `[[]];`. One more set of typos fixed: `\ell` changed to `\ell_1`, four locations, in spec of static analysis of `<forElement>`.

The wording around "object sequences" is not wonderful, but I'd recommend that we leave it as is for now.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -10654,13 +10654,13 @@ The same compile-time errors occur for $\ell$ as
 the errors that would occur with the corresponding \FOR{} statement
 \code{\AWAIT?\,\,\FOR\,\,($P$)\,\,\{\}},
 located in the same scope as $\ell$.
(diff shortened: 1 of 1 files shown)


munificent:
This is quite old, but I suspect we do still want to land it. @eernstg, is it worth fixing the conflicts and getting this in?

Issue #2181, opened 2022-03-30
Case expressions instead of if-case statement


munificent:
This is a strawman for a more general `case` expression form to replace the narrowly targeted [`if-case` statement][if-case] in the patterns proposal.

## Background

The main use for [patterns][] in control flow is switch statements and expressions. However, those can be fairly verbose. Following [Swift][swift if], the proposal also defines an [`if-case` statement][if-case] form:

[patterns]: https://github.com/dart-lang/language/blob/master/working/0546-patterns/patterns-feature-specification.md
[swift if]: https://useyourloaf.com/blog/swift-if-case-let/
[if-case]: https://github.com/dart-lang/language/blob/master/working/0546-patterns/patterns-feature-specification.md#if-case-statement

```dart
if (case [int x, int y] = json) {
  print('Was coordinate array $x,$y');
} else {
  throw FormatException('Invalid JSON.');
}
```

In [an issue][2137], @lrhn suggested that instead of `if-case` statements, we follow [C#][csharp is] and [Java][java instanceof] and allow a pattern after `is` (or `instanceof` in the case of Java):

[2137]: https://github.com/dart-lang/language/issues/2137
[csharp is]: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/patterns#declaration-and-type-patterns
[java instanceof]: https://openjdk.java.net/jeps/394

```dart
if (this.field is int value) {   // for "field promotion"
  // use value
}

if (something is [Object error, StackTrace stack]) {
  Error.throwWithStackTrace(error, stack);
}
```

Patterns would be allowed in any `is` expression, not just as a direct expression in an if condition, allowing useful chaining like:

```dart
if (something is int a && other is int b) return a + b;
```

Unfortunately, allowing any matcher pattern after `is` would [lead to some problems][comment]. Lasse and I spent some time discussing it and we came up with another idea, described here. If we decide we like this better, I'll roll it into the main patterns proposal and remove `if-case`.

[comment]: https://github.com/dart-lang/language/issues/2137#issuecomment-1059601471

## Proposal

We define a new infix `case` operator. The left-hand side is an expression, and the right-hand side is a matcher pattern. The above examples look like:

```dart
if (json case [int x, int y]) {
  print('Was coordinate array $x,$y');
} else {
  throw FormatException('Invalid JSON.');
}

if (this.field case int value) {   // for "field promotion"
  // use value
}

if (something case [Object error, StackTrace stack]) {
[...]

PR #2191, opened 2022-04-05
[patterns] Change if-case statement syntax to a more infix-like form.

munificent:
Instead of:

  if (case [int x, int y] = json) return Point(x, y);

The proposed syntax is now:

  if (json case [int x, int y]) return Point(x, y);

This does *not* define an infix case *expression*. But it is more
forward-compatible with that syntax if we decide to do that later.

See https://github.com/dart-lang/language/issues/2181.


leafpetersen (commented):
I do find this more readable.

PR #2198, opened 2022-04-12
Adjust and clarify simple string interpolation, and clarify identifier rules.

eernstg:
Adjust lexical rules around identifiers, making it explicit that `<IDENTIFIER>` does not derive reserved words or built-in identifiers. Also introduce a few extra rules from `Dart.g` in order to define the different kinds of identifiers more explicitly.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -9168,7 +9168,7 @@ This process is known as \Index{string interpolation}.
   \alt `${' <expression> `}'
 
 <SIMPLE\_STRING\_INTERPOLATION> ::= \gnewline{}
-  `$' (<IDENTIFIER\_NO\_DOLLAR> | <BUILT\_IN\_IDENTIFIER> | \THIS{})
+  `$' (<IDENTIFIER\_NO\_DOLLAR> | <BUILT\_IN\_IDENTIFIER> | \THIS)
 \end{grammar}
 
 \commentary{%
@@ -16542,7 +16542,8 @@ which implies that \THIS{} should be added.%
 \LMHash{}%
 An \Index{identifier expression} consists of a single identifier;
 it provides access to an object via an unqualified name.
-A \synt{typeIdentifier}
+A \synt{typeIdentifier} is an identifier which can be used
+as the name of a type declaration.
 
 \commentary{%
 A \synt{qualifiedName} is not an identifier expression,
@@ -16622,7 +16623,7 @@ both for human readers and during parsing.%
 
 \LMHash{}%
 It is a compile-time error if either of the identifiers \AWAIT{} or \YIELD{}
-is used as an identifier in a function body
+is used as an \synt{identifier} in a function body
 marked with either \ASYNC, \code{\ASYNC*}, or \code{\SYNC*}.
 
 \rationale{%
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -16537,62 +16542,69 @@ \subsection{Identifier Reference}
...
 
 \LMHash{}%
-A built-in identifier is one of the identifiers produced by the production
-\synt{BUILT\_IN\_IDENTIFIER}.
+The ordering of the lexical rules above ensure that \synt{IDENTIFIER}
+and \synt{IDENTIFIER\_NO\_DOLLAR} do not derive any built-in identifiers.
+Similarly, the lexical rule for reserved words (\ref{reservedWords})
+must be considered to come before the rule for \synt{BUILT\_IN\_IDENTIFIER},
+such that \synt{IDENTIFIER} and \synt{IDENTIFIER\_NO\_DOLLAR}
+also do not derive any reserved words.
```
lrhn, line 16603: Hmm.

When we have a grammar rule like 
```ebnf
<statement> ::= ...
   | `for' ...
```
does that `for` not conflict with `<RESERVED_WORD>`? Or can the grammar unpack the `<RESERVED_WORD>` that contains the `for`?

Is that because lexical rules are different from grammatical rules?

(Or can we just do:
```ebnf
<IDENTIFIER> ::= 
   NEGATIVE_LOOKAHEAD(<RESERVED_WORD>|<BUILT_IN_IDENTIFIER>)
   <IDENTIFIER\_START> <IDENTIFIER\_PART>*
```
instead, and not rely on ordering?

specification/dartLangSpec.tex
```diff
@@ -16611,7 +16623,7 @@ \subsection{Identifier Reference}
 \LMHash{}%
 It is a compile-time error if either of the identifiers \AWAIT{} or \YIELD{}
 is used as an identifier in a function body
```
lrhn, line 16625: Here we say "as an identifier" too. Should we elaborate that ("as a production of \<identifier\>")?

lrhn (approved):

specification/dartLangSpec.tex
eernstg: It is completely standard that a language has certain reserved words, and identifiers are defined using a simple regular expression. In practice this is achieved by giving the lexical rule for identifiers after the rules for reserved words. I would expect most language specification documents to rely on this approach implicitly. It may seem somewhat hacky to rely on the ordering, but the real difference would simply be that we're admitting it.

ANTLR requires lexical productions for each word (so we have `FOR ::= 'for'` and such), but the grammar in the language specification uses the keywords directly. I believe this is a reasonable compromise: We rely on the spec grammar to identify multiple occurrences of words. However, ANTLR already treats sequences of punctuation characters in that manner (so there's no need to define `PLUSPLUS ::= '++'` even though '++' occurs more than once in the grammar).

On the other hand, I believe that negative lookahead is considerably more esoteric than relying on the ordering of the lexical rules.

I'm keeping the rules as they are for now, please create an issue to get it changed if needed.

specification/dartLangSpec.tex
eernstg: Done!

PR #2207, opened 2022-04-19
Adjust wording about extension vs. static method conflicts

eernstg:
The section about extensions in the language specification had an imprecise wording about conflicts involving an extension member (static or not) and a member of the class `Object`. This PR clarifies the text: Such conflicts only arise when the member of `Object` is an instance member.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -5840,8 +5840,8 @@ Moreover, a compile-time error occurs in the following situations:
 \item $D$ declares a member whose basename is the name of a type parameter
   of $D$.
 \item $D$ declares an instance member or a static member whose basename is
-  \code{hashCode}, \code{noSuchMethod}, \code{runtimeType}, or \code{toString}.
-  Or $D$ declares an operator named \lit{==}.
+  \code{hashCode}, \code{noSuchMethod}, \code{runtimeType}, \code{toString},
+  or \lit{==}.
   \commentary{%
     That is, a member whose basename is also the name of
     an instance member that every object has.%
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -39,6 +39,8 @@
 % - Adjust and clarify simple string interpolation (to allow `'$this'`, which
 %   is already implemented and useful).
 % - Add several lexical rules about identifiers, clarifying different kinds.
+% - Clarify the conflicts between extension members and `Object` instance
+%   members.
 %
 % 2.14
 % - Add constraint on type of parameter which is covariant-by-declaration in
@@ -5837,8 +5839,8 @@ Moreover, a compile-time error occurs in the following situations:
 \item $D$ declares a type parameter named \code{E}.
 \item $D$ declares a member whose basename is the name of a type parameter
   of $D$.
-\item $D$ declares a member (static or not) with the same basename as a
-  member declared by the built-in class \code{Object}.
+\item $D$ declares a member (static or not) with the same basename as an
+  instance member declared by the built-in class \code{Object}.
 \item $D$ declares a constructor.
 \item $D$ declares an instance variable.
 \item $D$ declares an abstract member.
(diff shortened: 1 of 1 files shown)


eernstg:
@mit-mit, it looks like there's a need to fix the 'Failed to authenticate' failure again: https://github.com/dart-lang/language/runs/6075044920?check_suite_focus=true.

specification/dartLangSpec.tex
```diff
@@ -5837,8 +5839,8 @@ \section{Extensions}
 \item $D$ declares a type parameter named \code{E}.
 \item $D$ declares a member whose basename is the name of a type parameter
   of $D$.
-\item $D$ declares a member (static or not) with the same basename as a
-  member declared by the built-in class \code{Object}.
+\item $D$ declares a member (static or not) with the same basename as an
+  instance member declared by the built-in class \code{Object}.
```
lrhn, line 5843: Let's just list them:
```latex
\begin{itemize}
  \item{} \code{==}
  \item{} \code{hashCode}
  \item{} \code{noSuchMethod}
  \item{} \code{runtimeType}
  \item{} \code{toString}
\end{itemize}
```

Maybe phrase it as:

> the same basename as one of the instance members that are implemented by all objects:
> ... the list ...

since being declared by `Object` is really descriptive rather than defining for those members (`Null` implements them too, and even `dynamic` has special-treatment for those members.)

> Or just say, "the instance members declared by the type `Object?`", to not leave `Null` out.

lrhn (approved):

specification/dartLangSpec.tex
eernstg: We can't talk about `Object?` at this point (still waiting for reviews of null safety updates ;-), but it would indeed be nice to (1) introduce a regular name for the top type as a class (no special properties like `void` or `dynamic`), let's call it `Any`, and then talk about the members declared by `Any` where we currently talk about the members declared by `Object`.

But that's going to be handled in https://github.com/dart-lang/language/pull/2023 or some other PR which is doing part of the work done in the null safety update.

Listing the instance members directly seems to be a good choice for now, because that list would be very, very unlikely to change, and it will remain stable even if we introduce something like `Any`.

eernstg: Now using the listing.

eernstg (commented):
Thanks, now using the explicit listing.

specification/dartLangSpec.tex
```diff
@@ -5837,8 +5839,13 @@ \section{Extensions}
 \item $D$ declares a type parameter named \code{E}.
 \item $D$ declares a member whose basename is the name of a type parameter
   of $D$.
-\item $D$ declares a member (static or not) with the same basename as a
-  member declared by the built-in class \code{Object}.
+\item $D$ declares an instance member or a static member whose basename is
```
lrhn, line 5842: (And since extensions can't declare constructors at all, that covers every member).

specification/dartLangSpec.tex
```diff
@@ -5837,8 +5839,13 @@ \section{Extensions}
...
+  \code{hashCode}, \code{noSuchMethod}, \code{runtimeType}, or \code{toString}.
+  Or $D$ declares an operator named \lit{==}.
```
lrhn, line 5844: (Is this necessary, or could we just say "member whose basename is ... \lit{==}", because that is the name of the member introduced by the `operator==` declaration?)

lrhn (approved):

specification/dartLangSpec.tex
eernstg: Right, but using the word 'member' alone was too ambiguous, which would be the reason why we used to have '(static or not)'. So it seems reasonable to spell it out.

specification/dartLangSpec.tex
eernstg: It is true, we do define the basename of operators, and they are instance members. I just changed the text from a form that relied on this to a form where the operator is mentioned separately, because I thought it would be easier to read. However, taking another look, `\lit{==}` works fine, too. Done.

eernstg (commented):
Review response

specification/dartLangSpec.tex
lrhn: ACK.

lrhn (approved):

PR #2236, opened 2022-05-12
Clean-ups to horizontal inference feature specification.

stereotype441:


stereotype441:
See comments from @leafpetersen in https://github.com/dart-lang/language/pull/2199

eernstg (approved):
LGTM!

I can see that a number of questions that a reader might very well pose along the way have been answered, and I did not spot any discrepancies.

I do not have the algorithms described in the last two change blocks (lines 689++) so firmly in hand that I can review them properly. @leafpetersen?

resources/type-system/inference.md
```diff
@@ -680,6 +686,38 @@ The constraint solution for a type variable `X` with respect to a constraint set
 Note that the constraint solution is a type schema, and hence may contain
 occurences of the unknown type.
 
+#### Constraint solution for a set of type variables
+
+The constraint solution for a set of type variables `{X0, ..., Xn}` with respect
```
leafpetersen, line 691: I think I know what you mean, but you haven't quite set up what it means to be a previous solution here, as far as I can tell.  That is, what does it mean to be a "previous solution"?  I initially thought that you were carrying this across the implicit iteration in this definition, but I don't think that's right - this is an input to the algorithm here, right?  If so, I think maybe "partial solution" is a better term to make this clear.  That is, I think you are setting this up so that there can be an initial set of "frozen" things that are already fixed (this could maybe use commentary to make that clear if so).   So maybe something like:

"to a constraint set C and partial solution ..."

resources/type-system/inference.md
```diff
@@ -680,6 +686,38 @@ The constraint solution for a type variable `X` with respect to a constraint set
...
+to a constraint set `C`, with previous solution `{T0, ..., Tn}`, is defined to
+be the set of type schemas `{U0, ..., Un}` such that:
+  - If `Ti` is known (that is, does not contain `_`), then `Ui = Ti`.  _(Note
+    that the upcoming "variance" feature will relax this rule so that it only
+    applies to type variables without an explicitly declared variance.)_
+  - Otherwise, let `Vi` be the constraint solution for the type variable `Xi`
+    with respect to the constraint set `C`.
+  - If `Vi` is not known (that is, it contains `_`), then `Ui = Vi`.
+  - Otherwise, if `Xi` does not have an explicit bound, then `Ui = Vi`.
+  - Otherwise, let `Bi` be the bound of `Xi`.  Then, let `Bi'` be the type
```
leafpetersen, line 701: "be the type" -> "be the type schema" I think.  You are substituting in type schemas for type variables, so the result is a type schema.

resources/type-system/inference.md
```diff
@@ -696,6 +734,26 @@ constraint set `C` is define as follows:
 Note that the grounded constraint solution is a type, and hence may not contain
 occurences of the unknown type.
 
+#### Grounded constraint solution for a set of type variables
+
+The grounded constraint solution for a set of type variables `{X0, ..., Xn}`
+with respect to a constraint set `C`, with previous solution `{T0, ..., Tn}`, is
```
leafpetersen, line 740: As above, maybe say "and partial solution ..."?

resources/type-system/inference.md
```diff
@@ -696,6 +734,26 @@ constraint set `C` is define as follows:
...
+defined to be the set of types `{U0, ..., Un}` such that:
+  - If `Ti` is known (that is, does not contain `_`), then `Ui = Ti`.  _(Note
+    that the upcoming "variance" feature will relax this rule so that it only
+    applies to type variables without an explicitly declared variance.)_
+  - Otherwise, if `Xi` does not have an explicit bound, then `Ui` is the
+    grounded constraint solution for the type variable `Xi` with respect to the
+    constraint set `C`.
+  - Otherwise, let `Bi` be the bound of `Xi`.  Then, let `Bi'` be the type
```
leafpetersen, line 748: Something is missing here.  The `Ti` are still type schemas, so `Bi` is a type schema.  Either we need to take a least (greatest?) closure here or.... something.

accepted/future-releases/horizontal-inference/feature-specification.md
```diff
@@ -209,8 +220,9 @@ Performing type inference on an invocation consists of the following steps:
    try to match the return type of the target function type as a subtype of the
    invocation's type context.  This produces an initial set of type constraints.
    Then, using those constraints, find the constraint solution for the target
-   function type's type variables, considering bounds.  This produces a
-   preliminary mapping of type variables to type schemas.
+   function type's type variables, using a previous mapping that maps all type
```
leafpetersen, line 223: As below, I find the terminology "previous mapping" confusing here. I would say "using an initial partial solution that maps all type variables..."

accepted/future-releases/horizontal-inference/feature-specification.md
```diff
@@ -246,14 +258,16 @@ Performing type inference on an invocation consists of the following steps:
 
    * Horizontal inference: if generic inference is needed, and this is not the
      last stage, use all the type constraints gathered so far to find the
-     constraint solution for the target function's type variables, considering
-     bounds.  This produces an updated preliminary mapping of type parameters to
-     type schemas.
+     constraint solution for the target function's type variables, using the
+     mapping from the most recent previous execution of either this step or step
+     3 as the "previous mapping".  This produces an updated preliminary mapping
```
leafpetersen, line 263: I would propose uniformly using "partial solution" here instead of "mapping".  That is, talking about the "previous partial solution" and "produces an updated partial solution".  WDYT?

leafpetersen (commented):
Thanks!  Sorry to be slow with this.  Mostly LGTM with nits, but there's one substantive point around how we get to grounded type that I don't see yet.

resources/type-system/inference.md
stereotype441: That's much clearer, thanks!

resources/type-system/inference.md
stereotype441: Fixed, thanks

resources/type-system/inference.md
stereotype441: Done

accepted/future-releases/horizontal-inference/feature-specification.md
stereotype441: Done

leafpetersen (approved):
LGTM, thanks!

PR #2237, opened 2022-05-12
Add a section on Aspect Macros

jakemac53:
This is a very rough proposal for the "blackboard" api we have previously discussed. Basically, the goal is to reduce the code duplication risk of macros.

It allows for generating shared helper code across libraries in the program, based on Macro + Declaration pairs.

I plan on fleshing this out further at a later stage, but wanted to get some ideas written down.


working/macros/feature-specification.md
```diff
@@ -550,6 +550,116 @@ This does have two interesting and possibly unexpected consequences:
...
+
+This API does not promise to immediately invoke `macro` on `declaration`, it
+only returns an opaque `Identifier`, which refers to the declaration that will
+be produced by running `macro` on `declaration`. It returns a `Future` because
+all `Identifier` objects must come from the host compiler, so there is some
+async communication that has to happen.
+
+When a macro implementation receives a call to `applyAspect`, it must register
+that `macro` should be _eventually_ applied to `declaration`. Any two identical
+calls to `applyAspect` must always return an "equal" `Identifier` object,
```
munificent, line 616: By "identical calls" you mean where the parameters are the same, I assume? Do the AspectMacro instances need to be identical, or just instances of the same class?

It might be worth considering allowing some other kind of data or context parameters in order to parameterize what the macro does.

working/macros/feature-specification.md
```diff
@@ -550,6 +550,116 @@ This does have two interesting and possibly unexpected consequences:
...
+but they can't actually _see_ what each other has generated.
+
+In the worst case, this would result in them both outputting conflicting
+members, and in the best case they would end up generating duplicate code,
+causing unnecessary code bloat.
+
+Aspect Macros exist to solve this problem, by allowing macros to share these
+implementations. This is done by associating a single, canonical declaration
+with an Aspect Macro Instance + Declaration pair, and giving back an opaque
+identifier for the resulting declaration.
```
munificent, line 597: What if there is no logical declaration to attach the aspect macro to? Imagine you have a macro that needs some big blob of generated static data. You only want one copy of that data in the program, but it's not specific to any particular declaration, it's just shared by all applications of the macro.

Can the declaration be optional?

working/macros/feature-specification.md
```diff
@@ -550,6 +550,116 @@ This does have two interesting and possibly unexpected consequences:
...
+Both of these macro applications need to generate the same shared code for
+converting a Map into an instance of `A` (since `B` has a field of that type),
+but they can't actually _see_ what each other has generated.
+
+In the worst case, this would result in them both outputting conflicting
+members, and in the best case they would end up generating duplicate code,
+causing unnecessary code bloat.
+
+Aspect Macros exist to solve this problem, by allowing macros to share these
+implementations. This is done by associating a single, canonical declaration
```
munificent, line 595: This solves deduplication when two macro applications would generate the exact same code. But it doesn't solve wanting to *coalesce* information where you want multiple macro applications to build up one single data structure, like a global injector map for DI.

Is that a use case you want to think about for this?

munificent (approved):
This is really really cool.

working/macros/feature-specification.md
```diff
@@ -550,6 +550,116 @@ This does have two interesting and possibly unexpected consequences:
...
+extension BFromJson on Map<String, Object?> {
+  A toA(); // This also needs to generate a helper for converting `A`!
+
+  B toB();
+}
+```
+
+Both of these macro applications need to generate the same shared code for
+converting a Map into an instance of `A` (since `B` has a field of that type),
+but they can't actually _see_ what each other has generated.
```
scheglov, line 588: There might be another argument that requires this aspect macro approach, but specifically this example does not look convincing to me.

I was triggered by `A special library "next" to each library` statement below. So, we kind of pretend that something was generated while running phases 1-3, but actually we don't want to... why?

If `fromJsonExtension` runs on `B` and find that it needs something from `A`, then this something should have been generated while running on `A`. And `B` just knows that it was generated and uses it. If there is no required method (or whatever) for `A`, the macro reports an error.

jakemac53: > If `fromJsonExtension` runs on `B` and find that it needs something from `A`, then this something should have been generated while running on `A`. And `B` just knows that it was generated and uses it. If there is no required method (or whatever) for `A`, the macro reports an error.

The general use case is where you might want to generate something "on the side" of another type, without having to add it explicitly to the library where that type is defined. This could be serialization logic or anything else.

This pattern has several advantages:

- Users who do not need this functionality do not pay the cost for it.
- You don't have to try and get your  dependencies to support said functionality. You can add it yourself (as long as you can add it via new top level declarations of some kind).
- You can get all of this without causing code bloat/duplication when the same helper code is needed by multiple macros (using this feature at least).

For instance consider the use case of `mockito`. Classes generally should not all generate mocks and ship them as a part of their library. Instead, it is better if tests ask for the specific mocks they need, and that is how mockito works today. You can generate transitive mocks for a given type with an annotation like `@GenerateMocks([Cat])` on any top level declaration in your test file. This has to generate N classes (one for each user defined type that appears in the API of Cat basically). You can't actually do this at all today, you would have to manually specify all the transitive mocks to create in the annotation.

> I was triggered by `A special library "next" to each library` statement below. So, we kind of pretend that something was generated while running phases 1-3, but actually we don't want to... why?

We don't actually run it as a part of those phases in order to allow for multiple macros  to run the same "Aspect Macro", but only actually run it once. Also, these aspect macros have full access to introspect on the program (except the result of other aspect macros). Basically, it's a 4th phase which works very differently from the other phases, but the output of which is not introspectable beyond just getting an identifier. You can iteratively ask for more Aspect Macros to be applied in this phase as well, which is another key difference. This is safe to do because of the lack of introspection capability.

The particulars of where this code would end up in generally is tricky, I think conceptually the code is "next" to each library (this feature is inspired heavily by [bazel aspects](https://docs.bazel.build/versions/main/skylark/aspects.html)). But the code obviously can't actually live there. I also don't think we want to jam all of these aspects into a single top level library.

working/macros/feature-specification.md
jakemac53: Hmm, interesting question. A similar pattern could probably be used to do that (bazel aspects which this is inspired by do allow you to propagate information and merge it together).

I will have a think on that, and add a TODO section.

working/macros/feature-specification.md
jakemac53: This feels related to your other comment, in this case it is likely generating stuff based on some sort of input data, possibly produced by other macros, which is then likely aggregated together in some fashion?

working/macros/feature-specification.md
jakemac53: I see really two viable options here:

1.   Say that all `AspectMacro`s have to be _equal_.
    - We may want to enforce that they override `==` and `hashCode` if they have any fields?
2.   Don't allow `AspectMacro`s to have fields at all.

Fwiw, the later approach is basically the bazel approach, aspects are not configurable. If you want different behavior you have to create a different aspect.

Given that we also have to send these across the wire, they have to be serializable (and deserializable) as well. I am inclined to just go with approach 2 and say they can't have fields at all (but they can have getters, which allows for customized versions of the aspect).

In general we probably don't want to easily encourage customizable aspects because then they can't share code.

jakemac53:
Going to go ahead and land this, I moved it to a separate doc next to the feature specification, and added a disclaimer at the top that this is just a thought, not a concrete proposal to implement.

working/macros/feature-specification.md
```diff
@@ -550,6 +550,116 @@ This does have two interesting and possibly unexpected consequences:
...
+### Ordering of Aspect Macros
+
+Aspect Macros are conceptually applied in a 4th macro phase. All non-aspect
+macros must be invoked on a library prior to aspect macros being invoked on any
+declaration in that library.
+
+### Modular compilation
+
+Aspect Macros may be ran modularly, which may require merging of generated
+declarations at link time. Esssentially, the result of each module should have
```
johnniwinther: Esssentially -> Essentially

johnniwinther (approved):

working/macros/feature-specification.md
munificent: Even without aggregating, I can imagine that a macro might want to generate some helper function whose body maybe depends on some data known by the macro. but that doesn't really come from a particular *declaration*.

For example, imagine a `fieldHash` macro that generates a `hashCode` member on a class. It does so in terms of a series of helper functions like `hash1(Object? a)`, `hash2(Object? a, Object? b)`, etc. A `hash_n_()` helper with some given arity is only needed if the macro is applied to a class with that many fields. The macro author could hand author a helper library that contains `hash_n_()` methods up to some fixed arity and rely on the tree-shaker to discard all the unused ones. But they're essentially copy/paste of each other.

Instead, it might be nice if the `fieldHash` macro could just generate those helper functions on demand for the specific arities it needs. We still definitely want these functions to be deduplicated. But they aren't applied to any particular *declaration*.

Another way to think about it is that aspect macros here are essentially functions that take in some parameters and produce a declaration. They are memoized in that if you call the same aspect macro with the same parameters, the redundant declaration gets discarded. In the design here, an aspect macro can only be parameterized on one thing: a declaration. And it *must* be parameterized on that.

My suggestion is to generalize that to allow them to be parameterized on basically any set of data (as long as we can reliably send it over an isolate and test for equality, etc.). In the above (very contrived) example, it would just be parameterized on an int representing the arity.

jakemac53: Ah yeah the hash example is interesting, and I could see other similar uses as well

PR #2260, opened 2022-05-25
[patterns] Change named field shorthand syntax.

### LINKED ISSUE #2231: Should we extend the named field shorthand syntax to handle null-check ("?") patterns?
When destructuring a named field from an object, it's common to want to capture the resulting value in a variable with the same name:

```dart
var point = Point(1, 2);

var (x: x, y: y) = point;
```

To avoid the redundancy of having to say the same name twice (once for the field and once for the getter) many languages with name-based destructuring offer a shorthand syntax to let you elide one and infer it from the other. For example, in JavaScript:

```js
const user = {
    id: 42,
    isVerified: true
};

const {id, isVerified} = user; // Same thing as:
// const {id: id, isVerified: isVerified} = user;
```

The patterns proposal for Dart has a [similar shorthand](https://github.com/dart-lang/language/blob/master/working/0546-patterns/patterns-feature-specification.md#record-binder) (last paragraph). It looks like:

```dart
var (x:, y:) = point;
```

It's like a named field but with no subpattern. The `:` is needed to disambiguate it from a positional field.

However, using that shorthand doesn't work with what is likely to be a very common pattern:

1. Destructure a value from a getter on an object that returns a nullable type.
2. Test to see if the value is null.
3. If not, bind it to a variable with the same name but a non-nullable type.

Using the current explicit syntax, for example:

```dart
class MaybeStringBox {
  String? string;
}

test(MaybeStringBox box) {
  switch (box) {
    case (string: var string?): print('Actual $string');
  }
}
```

Here, because of the `?` null-check pattern, you can't use the shorthand syntax. Should we extend the shorthand syntax to allow that? One option is:

```dart
case (string:?): ...
```

So just allow a bare `?` as the subpattern and treat that as implicitly defining a null-check pattern containing a variable subpattern whose name is the same as the field.

Or maybe:

```dart
case (:string?): ...
```

So instead of eliding the variable subpattern and inferring it from the field, we elide the field name and infer it from the subpattern. Hence, the `:` appears at the beginning instead of the end. Then we allow that subpattern to be a variable pattern or a null-check pattern contaning a variable pattern.

This syntax would likely extend to other patterns more gracefully, like:

```dart
case (:string!): // Null-assert pattern.
case (:string as String): // Cast pattern.
```

Basically, we'd allow any subpattern as long as it contains a variable subpattern whose name could be used to infer the named field from.

munificent:
Instead of omitting the variable and inferring it from the field name,
we omit the field name and infer it from the variable. This allows all
types of variable patterns (var, final, typed), allows them to be
wrapped in null-check and null-assert patterns, and also allows
inferring a field name from a cast pattern.

Fix #2231.

### DIFF
--- a/working/0546-patterns/patterns-feature-specification.md
+++ b/working/0546-patterns/patterns-feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: In progress
 
-Version 1.8 (see [CHANGELOG](#CHANGELOG) at end)
+Version 1.9 (see [CHANGELOG](#CHANGELOG) at end)
 
 Note: This proposal is broken into a couple of separate documents. See also
 [records][] and [exhaustiveness][].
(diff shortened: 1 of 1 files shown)


munificent:
Closing this in favor of a revamped syntax.

Issue #2276, opened 2022-06-05
[Spec] The grammar does not have rules for dynamic types


kaby76:
I am testing the Spec. I programmatically scrape the [Spec](https://github.com/dart-lang/language/blob/master/specification/dartLangSpec.tex) for the rules of the grammar and generate an Antlr4 grammar. The grammar is then refactored via my Trash tool kit to fix various problems. Finally, the grammar is tested against the [sdk](https://github.com/dart-lang/sdk) library. The scraped grammar is better than the [grammar written by the Dart Language Team](https://github.com/dart-lang/sdk/blob/main/tools/spec_parser/Dart.g), but it is only as good as the Spec, which seems incomplete.

I am noticing missing rules in the [Spec](https://github.com/dart-lang/language/blob/8fb9f0ad4e8e5eabb1ec83abcd53f2f587d111fb/specification/dartLangSpec.tex) for dynamic types. For example, I cannot parse [dynamic methods](https://github.com/dart-lang/sdk/blob/09967d8c255e6d996bbc419fe3a2148aaff5e9e4/sdk/lib/_internal/wasm/lib/object_patch.dart#L43).

lrhn:
The word `dynamic` is not a reserved word, just a built-in identifier. That means that it *is* an `<identifier>`, just with rules against using it in particular places (and even those rules are loosened for `dynamic` and `Function`).
I think you are right and the grammar doesn't include that exception:
```ebnf
<typeIdentifier> ::= <IDENTIFIER>
  \alt <OTHER\_IDENTIFIER>
```
should probably be
```ebnf
<typeIdentifier> ::= <IDENTIFIER>
  \alt <OTHER\_IDENTIFIER>
  \alt \DYNAMIC{}
  \alt \FUNCTION{}
```

We should be a little careful since we use have
```dart
<qualifiedName> ::= <typeIdentifier> `.' <identifier>
  \alt <typeIdentifier> `.' <typeIdentifier> `.' <identifier>
```
and `dynamic` and `Function` are only allowed as *types*, not as *prefix names*. 
Still, even if the grammar allows it, there won't ever be a prefix names `dynamic`.

@eernstg

kaby76:
Thanks. I'll update my code to refactor the generated grammar for now.

kaby76:
Exploring some of the permutations:

1) "dynamic" as a field => compiles.
```
class Expression {
  String dynamic = '+';
  bool isPlusOrMinus(Expression expression) {
[...]

PR #2318, opened 2022-06-28
Correct `typeIdentifier` to include `dynamic`

eernstg:
The language specification currently has a grammar rule where `<typeIdentifier>` does not include `dynamic`, which is a mistake. This PR adds it. Thanks to @kaby76 for noticing this issue!

`Function` was mentioned in the same context, but `typeIdentifier` should actually not include `Function`. The word `Function` is treated specially in the rules about `<type>`, because it is used in ways that would otherwise give rise to ambiguities.

Resolves https://github.com/dart-lang/language/issues/2276.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -16567,6 +16567,7 @@ than it is to any single one of those grammar rules where it is used.%
 
 <typeIdentifier> ::= <IDENTIFIER>
   \alt <OTHER\_IDENTIFIER>
+  \alt \DYNAMIC{}
 
 <qualifiedName> ::= <typeIdentifier> `.' <identifier>
   \alt <typeIdentifier> `.' <typeIdentifier> `.' <identifier>


specification/dartLangSpec.tex
```diff
@@ -16567,6 +16567,7 @@ \subsection{Identifier Reference}
 
 <typeIdentifier> ::= <IDENTIFIER>
   \alt <OTHER\_IDENTIFIER>
+  \alt \DYNAMIC{}
```
lrhn, line 16570: Also needs `Function` (unless that's implicitly allowed as a special case of the function type syntax).

eernstg: As I mentioned, `type` [already derives](https://github.com/dart-lang/language/blob/ab893b49549bdd2bad750ca3a87fd0c309c21f85/specification/dartLangSpec.tex#L20047) `Function`. If `Function` is also added to `typeIdentifier` it is immediately an ambiguity.

@kaby76, why did you need to do that? Did you remove `Function` from the rule I linked above?

lrhn: Should `<typeNotVoidNotFunction>` really be named `<typeNotVoidNotFunctionNotDynamic>`?

Is there a good reason to treat `dynamic` and `Function` differently (say, because `?` is not allowed after `dynamic`?)

lrhn:
Sorry, accidentally pressed the "commit" button.

eernstg:
OK, if something comes up we'll create a new PR.

specification/dartLangSpec.tex
eernstg: `<typeNotVoidNotFunction>` is a type which is not the type `void` and not a function type (but it _does_ include `Function`). It does include `dynamic` as well: `<typeNotVoidNotFunction>` -> `<typeName>` -> `<typeIdentifier>` -> `dynamic`.

`Function` is tricky, because it's used in function types (`... Function(...)`), and also as a type in its own right.

Note that the spec grammar now says that `class dynamic {}` is OK, but the spec also says that [it is a compile-time error](https://github.com/dart-lang/language/blob/ab893b49549bdd2bad750ca3a87fd0c309c21f85/specification/dartLangSpec.tex#L16617) for a type declaration or a prefix to have a name which is a built-in identifier, so `class dynamic {}` is still an error after all.

But an extension could be named `dynamic` (e.g., `extension dynamic<X> on List<X> {...}`), which may or may not be a problem.

PR #2323, opened 2022-06-29
extension type spec: correct release version of extension methods

srawlins:


eernstg (approved):
I was a bit confused about this at first. I noticed the following in $SDK/tools/experimental_features.yaml:

```
  extension-methods:
    help: "Extension Methods"
    enabledIn: '2.6.0'
    expired: true
```

But $SDK/CHANGELOG.md has the following:

> **Extension methods** -- which we shipped in preview in 2.6.0 -- are no longer
in preview, and are now officially supported as of 2.7.0. Learn more about them
here:

So I guess 2.7 is it, LGTM!

PR #2333, opened 2022-07-07
Breaking change to await, update to flatten.

lrhn:
Changes **flatten** to be specified for a static type which is a type variable.

Changes `await` to throw when provided a `Future` value
of a type that could provide an unsound value if awaited.

This differes from the current specification, which instead treats the future as
a value, so `await someFuture` can evaluate to `someFuture`.

It also differs from the current implementation,
which awaits `someFuture` to its value,
whether the result is sound or not.

The result *can* be sound. Compare:
```dart
Object o = await (Future<Object?>.value(3) as FutureOr<Object));
```
which ends up not breaking soundness, with
```dart
Object o = await (Future<Object?>.value(null) as FutureOr<Object));
```
which currently does break soundness.

This change will make both into runtime errors.

(The alternative, and current specification,
to make both expression evaluate to the future itself,
is considered just as error prone and harder to debug,
since a type error is likely happen further downstream
when the `Object` is unexpectedly still a `Future`.)

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -11271,15 +11271,18 @@ which is used below and in other sections, as follows:
   or $T$ is an intersection type
   (\ref{intersectionTypes})
   \code{$X$\,\,\&\,\,$B$},
-  then $\flatten{T} = \flatten{B}$.
+  then \DefEquals{\flatten{$T$}}{\code{\flatten{$B$}}}.
 
-\item Otherwise if $T$ is \code{FutureOr<$S$>} for some $S$
-    then $\flatten{T} = S$.
+\item If $T$ is \code{$S$?}\ for some $S$
+  then \DefEquals{\flatten{$T$}}{\code{\flatten{$S$}?}}.
 
-\item Otherwise if $T$ implements \code{Future<$S$>} for some $S$
-    (\ref{interfaceSuperinterfaces}) then $\flatten{T} = S$.
+\item If $T$ is \code{FutureOr<$S$>} then \DefEquals{\flatten{$T$}}{S}.
 
-\item In any other circumstance, $\flatten{T} = T$.
+\item If $T$ implements \code{Future<$S$>}
+  (\ref{interfaceSuperinterfaces})
+  then \DefEquals{\flatten{$T$}}{S}.
+
+\item Otherwise, \DefEquals{\flatten{$T$}}{T}.
 \end{itemize}
 
 \commentary{%
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -15988,7 +15988,9 @@ Let $R$ be the run-time type of $o$.
 If $R$ implements \code{Future<$S$>} for some $S$
 (\ref{interfaceSuperinterfaces}) then
 it is a dynamic type error if $S$ is not a subtype of $T$.
-If $S$ is a subtype of $T$, let $f$ be $o$.
+Otherwise
+(\commentary{if no run-time error occurred}),
+let $f$ be $o$.
 
 Otherwise (when $R$ does not implement \code{Future}),
 let $f$ be a future object, implementing \code{Future<$T$>},
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -13748,7 +13748,7 @@ to an object---that would not even be possible for an extension.%
 }
 
 \commentary{%
-An member access on a type literal
+A member access on a type literal
 (e.g., \code{C.id()}, \code{C.id}, or \code{C?.id()}),
 always treats the declaration denoted by the literal as
 a namespace for accessing static members or constructors.
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -11267,22 +11267,22 @@ We define the auxiliary function
 which is used below and in other sections, as follows:
 
 \begin{itemize}
-\item If $T$ is a type variable $X$ with bound $B$,
+\item{} If $T$ is a type variable $X$ with bound $B$,
   or $T$ is an intersection type
   (\ref{intersectionTypes})
   \code{$X$\,\,\&\,\,$B$},
-  then \DefEquals{\flatten{$T$}}{\code{\flatten{$B$}}}.
+  then \DefEquals{\flatten{T}}{\flatten{B}}.
 
-\item If $T$ is \code{$S$?}\ for some $S$
-  then \DefEquals{\flatten{$T$}}{\code{\flatten{$S$}?}}.
+\item{} If $T$ is \code{$S$?}\ for some $S$
+  then \DefEquals{\flatten{T}}{\code{\flatten{S}?}}.
 
-\item If $T$ is \code{FutureOr<$S$>} then \DefEquals{\flatten{$T$}}{S}.
+\item{} If $T$ is \code{FutureOr<$S$>} then \DefEquals{\flatten{T}}{S}.
 
-\item If $T$ implements \code{Future<$S$>}
+\item{} If $T$ implements \code{Future<$S$>}
   (\ref{interfaceSuperinterfaces})
-  then \DefEquals{\flatten{$T$}}{S}.
+  then \DefEquals{\flatten{T}}{S}.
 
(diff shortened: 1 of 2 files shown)


specification/dartLangSpec.tex
```diff
@@ -11267,33 +11267,24 @@ \subsection{Function Expressions}
 which is used below and in other sections, as follows:
 
 \begin{itemize}
-\item If $T$ is \code{FutureOr<$S$>} for some $S$ then $\flatten{T} = S$.
+\item If $T$ is a type variable $X$ with bound $B$,
+  or $T$ is a promoted type variable $X & B$,
+  then $\flatten{T} = \flatten{B}$.
```
eernstg, line 11272: For consistency with other function definitions, we're using an equals sign with a triangle, defined by `\DefEquals{}{}`:
```suggestion
 then \DefEquals{\flatten{$T$}}{\code{\flatten{$B$}}}.
```

specification/dartLangSpec.tex
```diff
@@ -11267,33 +11267,24 @@ \subsection{Function Expressions}
 which is used below and in other sections, as follows:
 
 \begin{itemize}
-\item If $T$ is \code{FutureOr<$S$>} for some $S$ then $\flatten{T} = S$.
+\item If $T$ is a type variable $X$ with bound $B$,
```
eernstg, line 11270: OK, so we're updating flatten anyway! But in that case we would basically have to assume the null safety specification as the context, because the new definition of flatten includes `S?` as a type. This is a little bit inconvenient because we don't have all the sections (that is, `\ref{X}` is needed for some `X` that don't exist), but I guess it should be OK to have a spec with unresolved refs for a while.

In any case, we'll land the null safety update any day now, of course! ;-)

specification/dartLangSpec.tex
```diff
@@ -11267,33 +11267,24 @@ \subsection{Function Expressions}
...
+  then $\flatten{T} = \flatten{B}$.
 
-\item Otherwise if
-  \code{$T <:$ Future}
-  then let $S$ be a type such that
-  \code{$T <:$ Future<$S$>}
-  and for all $R$, if
-  \code{$T <:$ Future<$R$>}
-  then $S <: R$.
+\item Otherwise if $T$ is \code{FutureOr<$S$>} for some $S$
```
eernstg, line 11274: Here's an updated version of the rest of the definition of flatten (replacing lines 11274--11280, including both):

```
\item If $T$ is \code{$S$?}\ for some $S$
  then \DefEquals{\flatten{$T$}}{\code{\flatten{$S$}?}}.
\item If $T$ is \code{FutureOr<$S$>} then \DefEquals{\flatten{$T$}}{S}.
\item If $T$ implements \code{Future<$S$>}
  (\ref{interfaceSuperinterfaces})
  then \DefEquals{\flatten{$T$}}{S}.
\item Otherwise,
  \DefEquals{\flatten{$T$}}{T}.
```

specification/dartLangSpec.tex
```diff
@@ -13755,7 +13746,7 @@ \subsubsection{Ordinary Invocation}
 }
 
 \commentary{%
-An member access on a type literal 
+An member access on a type literal
```
eernstg, line 13749:
```suggestion
A member access on a type literal
```

specification/dartLangSpec.tex
```diff
@@ -15993,11 +15984,15 @@ \subsection{Await Expressions}
...
-If the run-time type of $o$ is a subtype of \code{Future<$T$>},
-then let $f$ be $o$;
-otherwise let $f$ be the result of creating
-a new object using the constructor \code{Future<$T$>.value()}
-with $o$ as its argument.
+Let $R$ be the run-time type of $o$.
+If $R$ implements \code{Future<$S$>} for some $S$
+(\ref{interfaceSuperinterfaces}) then
+it is a dynamic type error if $S$ is not a subtype of $T$.
+If $S$ is a subtype of $T$, let $f$ be $o$.
```
eernstg, line 15991: I usually handle descriptions of errors (run-time or compile-time) slightly differently:
```suggestion
Otherwise
(\commentary{if no run-time error occurred}),
let $f$ be $o$.
```

We could do that for consistency. However, I also think this style is a bit safer to read (a bit harder to misunderstand), because it puts the emphasis on the error: "One case is an error, all other cases proceed as follows".

With 'If $S$ is a subtype of $T$', there's more room to think that there could be even more cases.

specification/dartLangSpec.tex
```diff
@@ -15993,11 +15984,15 @@ \subsection{Await Expressions}
...
+
+Otherwise (when $R$ does not implement \code{Future}),
```
eernstg, line 15993:
```suggestion
Otherwise
(\commentary{when $R$ does not implement \code{Future}}),
```

specification/dartLangSpec.tex
```diff
@@ -15993,11 +15984,15 @@ \subsection{Await Expressions}
...
+let $f$ be a future object, implementing \code{Future<$T$>},
```
eernstg, line 15994:
```suggestion
let $f$ be an object whose run-time type implements \code{Future<$T$>},
```

specification/dartLangSpec.tex
eernstg: We'd need to add the following to `dart.sty`, (preferably, just before `% ......` at the end, such that it will merge cleanly with the dart.sty from the null safety update):

```
\newcommand{\DefEquals}[2]{\ensuremath{{#1}\stackrel{\vartriangle}{=}{#2}}}
\newcommand{\DefEqualsNewline}[2]{
  \ensuremath{{#1}\stackrel{\vartriangle}{=}}\\
  \ensuremath{{#2}}%
}
```

eernstg (approved):
LGTM, recommending some changes, for consistency with the rest of the language specification.

I also suggested several changes for the definition of `\flatten`, in order to make it complete for null safety (I don't think we should specify flatten in a new way for the legacy language, so the new definition of flatten should be the full null safety thing). But this creates some issues: We need to use a `\ref{...}` to a section that doesn't exist in the current specification, and we need to add a command to `dart.sty`.

specification/dartLangSpec.tex
lrhn: Of course, of course.

I decided to omit the `S?` case because I assumed it would be merged in with the rest of null safety,
so you only needed to merge the one line here into the patch.

Also why I didn't use `\DefEquals`, since the other existing cases didn't, and changing those would make rebasing the null safety PRs harder.

specification/dartLangSpec.tex
lrhn: The other cases don't use `\DefEquals`. I assumed the null safety PRs would make that change, and it was better to *not* change the other lines, so that rebasing got easier.

specification/dartLangSpec.tex
```diff
@@ -11267,33 +11267,24 @@ \subsection{Function Expressions}
 which is used below and in other sections, as follows:
 
 \begin{itemize}
-\item If $T$ is \code{FutureOr<$S$>} for some $S$ then $\flatten{T} = S$.
+\item If $T$ is a type variable $X$ with bound $B$,
+  or $T$ is a promoted type variable $X & B$,
```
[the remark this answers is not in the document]
lrhn: Physical formatting (`\ ,\,`) ... ICK.
But what works works.

specification/dartLangSpec.tex
lrhn: Done.

eernstg (approved):
LGTM! Just to make sure, does it LaTeX successfully? (A script on github will do that when it's landed.)

specification/dartLangSpec.tex
```diff
@@ -15993,11 +15989,18 @@ \subsection{Await Expressions}
...
+(\ref{interfaceSuperinterfaces}) then
+it is a dynamic type error if $S$ is not a subtype of $T$.
+Otherwise
+(\commentary{if no run-time error occurred}),
+let $f$ be $o$.
+
+Otherwise
+(\commentary{when $R$ does not implement \code{Future}}),
+let $f$ be an object whose run-time type implements \code{Future<$T$>},
+which is completed with the value $o$.
```
lrhn, line 16003: Nit. This doesn't actually say what behavior the *f* object has.

PR #2343, opened 2022-07-11
Change noSuchMethod example to be correct with null safety

eernstg:
Cf. https://github.com/dart-lang/language/pull/2052. This PR changes an example in the language specification section on `noSuchMethod` such that it is correct with null safety (it is correct without null safety as well, so we can do it now, and it will simply be there when we land the null safety updates).


eernstg:
@lrhn, friendly ping?

lrhn (approved):

eernstg:
Thanks!

PR #2352, opened 2022-07-21
Initial overview of extension structs

leafpetersen:
This is an initial sketch of an alternative direction for approaching the view/extension type problem.  Still in a rough state, putting this up here for some preliminary discussion.

### DIFF
--- a/working/extension_structs/overview.md
+++ b/working/extension_structs/overview.md
@@ -30,6 +30,7 @@ entirely static).
 This proposal builds on previous proposals in this space, including:
   - [Views](https://github.com/dart-lang/language/blob/master/working/1426-extension-types/feature-specification-views.md)
   - [Extension types](https://github.com/dart-lang/language/blob/master/working/1426-extension-types/feature-specification.md)
+  - [Protected extension types](https://github.com/dart-lang/language/issues/1467)
 
 
 This overview is not intended to be a feature specification: it is designed to
--- a/working/extension_structs/overview.md
+++ b/working/extension_structs/overview.md
@@ -342,9 +342,11 @@ struct GenericData<T>(T x, T y) extends Foo;
 
 Members are inherited from super-structs as usual.
 
-It is an error for a struct to extend a class.
+It is an error for a struct to extend a class, except Object.
 
 It is an error for a struct to be extended outside of the defining library.
+
+It is an error for a struct to be extended by a class.
 
 Extension is supported primarily to allow compact definition of algebraic
 datatypes.  We expect that structs would be incorporated into switches with
@@ -455,11 +457,11 @@ otherwise the result of calling "toString".
 
 **COMMENTARY(leafp):** *This needs a bit of work.  If we keep the dynamic check
   for "struct-ness", then we need do deal with the case that the user defines a
-  debutToString thing with the wrong type*.
+  debugToString thing with the wrong type*.
 
 **COMMENTARY(leafp):** *If we keep this, we may wish to specify that compilers
   may choose to make it a link time error to have an invocation of debugToString
-  in the program outside of asserts, and other debutToString methods*.
+  in the program outside of asserts, and other debugToString methods*.
 
 
 #### copyWith
@@ -617,6 +619,14 @@ extension struct AlternativeNat(int _x) implements Nat { ...}
 class MockNat implements Nat {...}
 ```
 
+Extension structs can define their own constructors as usual, replacing or
+adding to the generated constructors.
+
+```dart
+extension struct Window(JSObject o) {
+  Window.cons() : o = Window(js_util.callConstructor(...));
+}
+```
 
 ## Extension structs in more detail
 
@@ -841,10 +851,10 @@ The toString method delegates to the underlying representation.
 
 If no debugToString method is defined in the extension struct, then a
 debugToString method will be defined which returns a formatted description of
-the receiver, of the form `<type>(<f0>, ..., <f1>)` where `<type>` is the
-extension struct type via which the receiver is called, and the `fi` are the
-result of calling "debugToString" on the `i`th field if that field is
-(dynamically) a struct, and otherwise the result of calling "toString".
+the receiver, of the form `<type>(<f0>)` where `<type>` is the extension struct
+type via which the receiver is called, and `<f0>` is the result of calling
+"debugToString" on the unique field if that field is (dynamically) a struct, and
+otherwise the result of calling "toString".
 
 #### copyWith
 
@@ -865,6 +875,175 @@ It is an error to define any of the object members in the extension struct.
 
 TODO(leafp): This should work, write out the details.
 
+## Extensions to the core proposal
+
+### Allow abstract fields in abstract super-structs
+
+We could choose to allow abstract structs to define a subset of the fields in a
+primary constructor, interpreting them as abstract fields.  e.g.
+
+```dart
+abstract struct ColorPoint(int color);
+struct ColorPoint2D(int color, int x, int y) extends ColorPoint;
+```
+
+would be roughly equivalent to:
+
+```dart
+abstract class ColorPoint {
+  abstract int color;
+}
+
+class ColorPoint2D extends ColorPoint{
+  final int color;
+  final int x;
+  final int y;
+  ColorPoint2D(this.color, this.x, this.y);
+  // More generated methods here
+}
+```
+
+### Allow concrete super-structs
+
+We could allow extending concrete structs. 
+
+```dart
+abstract struct ColorPoint(int color);
+struct ColorPoint2D(int color, int x, int y) extends ColorPoint;
+struct ColorPoint3D(int z) extends ColorPoint2D;
+```
+
+which would be roughly equivalent to:
+
+```dart
+abstract class ColorPoint {
+  abstract int color;
+}
+
+class ColorPoint2D extends ColorPoint{
+  final int color;
+  final int x;
+  final int y;
+  ColorPoint2D(this.color, this.x, this.y);
+  // More generated methods here
+}
+
+class ColorPoint3D extends ColorPoint2D {
+  final int z;
+  ColorPoint3D(super.color, super.x, super.y, this.z);
+  // More generated methods here
+}
+```
+
+**COMMENTARY(leafp):** *I would prefer, at least as a starting point, to forbid
+  overriding of the fields in sub-structs, to make it easier to compile structs
+  to something with a predictable memory layout.  For the same reason, I would
+  propose to continue to treat these as non-extensible outside of the defining
+  library*
+
+### Allow extension structs to extend other extension structs
+
+To support non-trivial subtyping hierarchies using extension structs, we could
+choose to allow extension structs to extend other extension structs, subject to
+the requirement that the field type remains the same.
+
+```dart
+extension struct Nat(int _x) {
+  Nat(int x) : assert(x >= 0), _x = x;
+  Nat.zero() : _x = 0;
+
+  Nat get succ => Nat(_x+1);
+  Nat plus(Nat other) => Nat._x + other._x;
+
+}
+
+extension struct Pos extends Nat {
+  Pos(int x) : assert(x >= 1), super(x);
+}
+```
+
+The semantics of extension would be, as with `implements`, that all methods on
+the super-struct type are statically available on the sub-struct, but no
+overriding is allowed.
+
+### Allow extension structs to extend other extension structs and refine the
+    type.
+
+We could choose to allow extension structs to require a more specific type for
+the unique field.
+
+```dart
+extension struct Number(num _x);
+extension struct Integer(int _x) extends Number;
+```
+
+### Allow extension structs to provide overriding implementations of implemented
+    or extended members
+
+The core proposal forbids overriding, to avoid the surprising behavior where the
+same object resolves methods differently depending on the static type.  We could
+choose to relax this, at the cost of some surprising behavior.
+
+```dart
+extension struct EvenInteger(int x) {
+    bool get isEven => true;
+}
+// Truly odd integers.
+extension struct OddInteger(int x) extends EvenInteger {
+    bool get isEven => false;
+}
+
+void test() {
+  OddInteger i = OddInteger(2);
+  assert(!i.isEven); // Dispatch goes to OddInteger.isEven
+  EvenInteger e = i; // Ok
+  assert(i.isEven); // Dispatch goes to EvenInteger.isEven
+}
+```
+
+**COMMENTARY(leafp):** *We could in principle not enforce the usual subtyping
+  constraints on overriding, but in practice I think this should be done.*
+
+### Define an implicit boxed version of extension structs
+
+The close correspondence between structs and extension structs suggests the
+possibility of saying that every `extension struct` declaration implicitly
+defines a corresponding `struct` declaration, which behaves exactly as if the
+same declaration had been made except with the `extension` prefix removed.  For
+an extension struct `Foo`, we might choose to name these implicit types as
+`Foo.struct`.
+
+```dart
+extension struct Nat(int _x) {
+  Nat(int x) : assert(x >= 0), _x = x;
+  Nat.zero() : _x = 0;
+
+  Nat get succ => Nat(_x+1);
+  Nat plus(Nat other) => Nat._x + other._x;
+
+  // Override the underlying isNegative operation (incorrectly)
+  bool get isNegative => true;
+}
+
+void test() {
+  Nat n = Nat(2);
+  n.succ(); // Returns Nat(3)
+  // (n as dynamic).succ(); // noSuchMethod
+  assert(n.isNegative);
+
+  int i = n as int; // Succeeds
+  assert(!i.isNegative); // Dispatch goes to the integer method
+
+  // Nat.struct is the type which would be defined by the same struct
+  // definition above, with the extension prefix removed.
+  Nat.struct b = n.struct;
+  b.succ(); // Returns Nat(3)
+  (b as dynamic).succ(); // Returns Nat(3)
+  assert(b.isNegative);
+
+  // int i = n as int; // Case fails
+}
+```
 
 
 ## Changelog
--- a/working/extension_structs/overview.md
+++ b/working/extension_structs/overview.md
@@ -18,20 +18,21 @@ types", discussed among other places
   - The desire to have an accounting for data structures without identity as
     described [here](https://github.com/dart-lang/language/issues/2246).
 
-The proposal here takes two steps.  The first is to add a restricted kind of
-class (provisionally describe as a "struct" here), which gives up some of the
-affordances of general Dart classes in exchange for compact syntax and automatic
-generation of useful methods.  These restricted classes provide data class like
-functionality.  The second step is to support a further restriction on structs
-with a single field which eliminates the wrapper object, representing the struct
-entirely as the underlying object (at the cost of making the abstraction
-entirely static).
+The proposal here takes two steps.
+  - The first is to add a restricted kind of class (provisionally describe as a
+"struct" here), which gives up some of the affordances of general Dart classes
+in exchange for compact syntax and automatic generation of useful methods.
+These restricted classes provide data class like functionality: they are
+immutable, they have structural identity, and they get a number of conveniently
+auto-generated methods.
+  - The second is to support a further restriction on structs with a single
+field which eliminates the wrapper object, representing the struct entirely as
+the underlying object (at the cost of making the abstraction entirely static).
 
 This proposal builds on previous proposals in this space, including:
   - [Views](https://github.com/dart-lang/language/blob/master/working/1426-extension-types/feature-specification-views.md)
   - [Extension types](https://github.com/dart-lang/language/blob/master/working/1426-extension-types/feature-specification.md)
   - [Protected extension types](https://github.com/dart-lang/language/issues/1467)
-
 
 This overview is not intended to be a feature specification: it is designed to
 give a brief overview of the core proposal and set out some design points for
@@ -64,8 +65,8 @@ Larger extensions are postponed to a later section.
 
 ## Structs by example
 
-Structs allow you to define simple classes very compactly.  For example, we
-might model a component vector as follows:
+Structs allow you to define simple classes containing immutable data very
+compactly.  For example, we might model a component vector as follows:
 
 ```
 struct Component(int x, int y, int z, int w = 1);
@@ -84,7 +85,7 @@ void test() {
   assert(c1 == c2); // Equality is defined to be structural
   assert(c1.hashCode == c2.hashCode); // With correspondind hashCode
 
-  print(c1.debutToString()); // Prints "Component(0, 0, 0, 1)"
+  print(c1.debugToString()); // Prints "Component(0, 0, 0, 1)"
 
   print(c1.toString()); // Prints "struct"
 
@@ -95,7 +96,10 @@ void test() {
 ```
 
 Structs may explicitly define constructors, static members, and normal class
-members (except fields), and may implement interfaces.
+members (except fields), and may implement interfaces.  All fields must be
+declared in the header (the "primary constructor"), and are implicitly final
+(making all structs shallowly immutable).  Fields may be marked as private as
+usual by naming them with a leading `_` in their name.
 
 ```dart
 struct Component(int _x, int _y, int _z, int _w = 1)
@@ -180,6 +184,13 @@ view Data();
 view GenericData<T>();
 ```
 
+Perhaps record:
+
+```dart
+record Data();
+record GenericData<T>();
+```
+
 ##### Use a modifier on classes
 
 
@@ -208,6 +219,12 @@ initializer values.
 Each entry in the list is a field in the struct.  Every field is implicitly
 final.  They may not be late, and a type must be provided.
 
+**COMMENTARY(leafp):** *The restriction to final fields is both because that's
+  the common use case, and because structural identity doesn't make much sense
+  if we allow them to be mutable.  We could potentially box each mutable field
+  into heap allocated ref cell but that feels very unpleasant, and has perf
+  implications that don't match the intended use cases. *
+
 If an initializer is provided for an entry, all subsequent entries must also
 have an initializer provided (see generated members below).
 
@@ -235,12 +252,12 @@ following classes:
 
 ```dart
 class Data {
-  int x;
-  List<int> l = [3]
+  final int x;
+  final List<int> l = [3]
  }
 class GenericData<T> {
- T x;
- T y;
+ final T x;
+ final T y;
  }
 
 ```
@@ -249,7 +266,42 @@ class GenericData<T> {
 
 We could choose to make this look more directly like a constructor argument list
 by allowing "named" parameters inside of a brace delimited set.  These could
-then become named parameters in the default constructor.
+then become named parameters in the default constructor.  Example:
+
+```
+struct Component(int x, int y, int z, {int w = 1});
+
+void test() {
+  // Call the default constructor, using the default value for w
+  var c1 = Component(0, 0, 0);
+  // Call the default constructor, explicitly passing w
+  var c2 = Component(1, 1, 1, w: 1);
+...
+}
+```
+
+#### Alternative: no primary constructors
+
+We could choose to make the field declarations look exactly like classes instead
+of using a primary constructor syntax.
+
+
+```
+struct Component {
+  int x;
+  int y;
+  int z;
+  int w = 1;
+}
+```
+
+**COMMENTARY(leafp):** *Given the constraint that structs are immutable, I don't
+  like that `int x` here means a different thing than it does in a class (since
+  it is implicitly final).  We could therefore require each one to be written as
+  `final int x` and make it an error to have a non-final variable.  This feels
+  very boilerplate-heavy to me relative to the compact one line form.*
+
+
 
 ### Members of structs
 
@@ -353,6 +405,9 @@ parameter, the initializer value is assigned to that parameter in the
 initializer list of the constructor.  Note that this requires the ability to
 detect whether or not a parameter was passed, which is not expressible strictly
 as a de-sugaring.
+
+**Note that initializers are not required to be constant, so this de facto adds
+  non-const default values in a very limited case.**
 
 If no default constructor is defined in the class, a default constructor is
 generated which forwards to the generated primary constructor.
--- a/working/extension_structs/overview.md
+++ b/working/extension_structs/overview.md
@@ -41,7 +41,8 @@ them into an existing proposal, or add a new feature specification.
 ## Design principles
 
 We aim to minimize the differences between structs and classes.  As much as
-possible, structs should behave as restrictions of classes.  We specifically aim
+possible, structs should behave as restrictions of classes, and extension
+structs should behave as further restrictions of structs.  We specifically aim
 to avoid as much as possible having different behaviors for the same concept
 (e.g. differences in scoping).
 
@@ -59,6 +60,76 @@ describes the restriction of the full struct to support static wrappers.
 This section describes how structs look and behave, largely by example.  In many
 cases, alternatives or extensions to the core proposal are described in line.
 Larger extensions are postponed to a later section.
+
+## Structs by example
+
+Structs allow you to define simple classes very compactly.  For example, we
+might model a component vector as follows:
+
+```
+struct Component(int x, int y, int z, int w = 1);
+```
+
+This short definition essentially defines a class `Component`, with four fields,
+and provides a number of convenient methods.
+
+```dart
+void test() {
+  // Call the default constructor, using the default value for w
+  var c1 = Component(0, 0, 0);
+  // Call the default constructor, explicitly passing w
+  var c2 = Component(1, 1, 1, 1);
+
+  assert(c1 == c2); // Equality is defined to be structural
+  assert(c1.hashCode == c2.hashCode); // With correspondind hashCode
+
+  print(c1.debutToString()); // Prints "Component(0, 0, 0, 1)"
+
+  print(c1.toString()); // Prints "struct"
+
+  // Generated copy method
+  var c2 = c1.copyWith(x : 2);
+  assert(c2.x == 2);
+}
+```
+
+Structs may explicitly define constructors, static members, and normal class
+members (except fields), and may implement interfaces.
+
+```dart
+struct Component(int _x, int _y, int _z, int _w = 1)
+  implements Comparable<Component> {
+
+  double get x => _x/_w;
+  double get y => _x/_w;
+  double get z => _z/_w;
+
+  int compareTo(Component other) => throw "TODO: Unimplemented";
+
+  Component.zero() : _x = 0, _y = 0, _z = 0;
+}
+```
+
+With the addition of pattern matching and switches, structs will support closed
+families (algebraic datatypes).
+
+```dart
+abstract struct Operand;
+
+struct ConstantOperand(Value c) extends Operand;
+
+struct IdentifierOperand(Identifier i) extends Operand;
+
+Operand replaceIn(Operand oper, Identifier t, Value c) {
+  return switch (oper) {
+    case ConstantOperand(_) : oper;
+    case IdentifierOperand(var i) where i == t: ConstantOperand(c);
+    case IdentifierOperand(_) : oper;
+  }
+}
+```
+
+## Structs in more detail
 
 ### Introduction form
 
@@ -110,7 +181,8 @@ view GenericData<T>();
 
 ##### Use a modifier on classes
 
-We could, for example, use "data class":
+
+We could use "data class":
 
 ```dart
 data class Data();
@@ -138,9 +210,17 @@ final.  They may not be late, and a type must be provided.
 If an initializer is provided for an entry, all subsequent entries must also
 have an initializer provided (see generated members below).
 
+**COMMENTARY(leafp):** *This is to allow initializers to double as default
+  values for the generated constructor.  This may be too cute.  Alternatively we
+  could forbid initializers, or just make fields with initializers not available
+  to the constructor (as with a normal class)*
+
 It is an error if two entries have the same name, and it is an error if one
 entry consists of the name of another entry except prefixed with `_`.  That is,
 all entries must be uniquely named after ignoring privacy.
+
+**COMMENTARY(leafp):** *This is to allow us to use the non-private version of
+  the name as a named parameter in methods*
 
 Examples:
 
@@ -188,6 +268,12 @@ provided.  This is supported to allow families of structs sharing a common
 super-interface which provides clean support for algebraic data types (see the
 section on extension below).
 
+**COMMENTARY(leafp):** *We could possibly allow hierarchies of abstract
+  super-structs.  We could also potentially allow super-structs to define
+  fields, which would be treated as abstract fields which the sub-structs must
+  provide*
+
+
 ### Extending structs
 
 A struct may extend another struct from the same library.  It is an error if the
@@ -228,11 +314,12 @@ semantically exactly as if they were de-sugared into classes in the obvious way.
 
 ### Identity of structs
 
-The identity operator on structs *may* return `true` for structs which:
-  - Have the same runtime type
-  - Are pointwise identical on the fields in the primary constructor.
+The identity operator on structs *may* return `true` for structs such that:
+  - Both have the same runtime type
+  - For every pair of corresponding fields, identical could validly return true
+    on that pair.
 
-The identiity operator on structs may always return false.
+The identity operator on structs may always return false.
 
 In other words, the identity operator may be used as a "fast" cut-off for
 equality, but compilers are free to box and unbox structs at will without
@@ -269,6 +356,12 @@ as a de-sugaring.
 If no default constructor is defined in the class, a default constructor is
 generated which forwards to the generated primary constructor.
 
+**COMMENTARY(leafp):** *The treatment of initializers as default values here is
+  appealing, but it may be too cute.  It essentially adds non-const default
+  values only for this specific use case.  There's also a question of whether we
+  allow user defined constructors on the struct to also override the default
+  initializer, and if so via what syntax*.
+
 
 ##### Alternative: named parameters
 
@@ -304,6 +397,15 @@ runtime type of the receiver, and the `fi` are the result of calling
 "debugToString" on the `i`th field if that field is (dynamically) a struct, and
 otherwise the result of calling "toString".
 
+**COMMENTARY(leafp):** *This needs a bit of work.  If we keep the dynamic check
+  for "struct-ness", then we need do deal with the case that the user defines a
+  debutToString thing with the wrong type*.
+
+**COMMENTARY(leafp):** *If we keep this, we may wish to specify that compilers
+  may choose to make it a link time error to have an invocation of debugToString
+  in the program outside of asserts, and other debutToString methods*.
+
+
 #### copyWith
 
 If no copyWith method is defined in or inherited by the struct, then a copyWith
@@ -320,6 +422,12 @@ passed to the copyWith method invocation, the value of the corresponding field
 from the current instance is passed on as the argument to the corresponding
 parameter of the generated primary constructor.
 
+**COMMENTARY(leafp):** *As with constructors, this method cannot be generated as
+  a strict de-sugaring, since the ability to detect whether or not an argument
+  has been passed is not available in Dart.  Implementations already support
+  this for default values, so it is likely not problematic to implement*.
+
+
 #### Additional generated members
 
 We may wish to also define additional generated members.  For example, we may
@@ -333,6 +441,9 @@ It is an error to override the "runtimeType" method of a struct.
 
 It is an error to override the "noSuchMethod" method of a struct.
 
+### Const structs
+
+TODO(leafp): This should work, write out the details.
 
 ## Extension Structs
 
@@ -340,6 +451,118 @@ Extension structs are restrictions of the core struct feature, designed to
 support wrapper-less views on an object.  This section describes how extension
 structs look and behave, largely by example.  In many cases, alternatives or
 extensions to the core proposal are described in line.
+
+## Extension structs by example
+
+Extension structs are targeted at relatively niche uses where you wish to define
+a set of statically dispatched methods layered on top of an underlying
+representation, without introducing a wrapper object.  A canonical use case
+driving this design is to be able define a Dart typed interface for methods on a
+Javascript object, providing a wrapperless interoperation capability.
+
+
+```dart
+extension struct Window(JSObject o) {
+  // Signatures provided here for methods to be delegated to the underlying
+  // JSObject
+  external bool get closed;
+  // etc
+}
+
+void test(Window w) {
+  // Window methods can be called using Dart syntax
+  if (w.closed) {...}
+
+  // Windows are represented as the underlying object
+  assert(w is JSObject);
+
+  // Windows can be cast to the underlying object
+  JSObject o = w as JSObject;
+
+  // Extension struct types are reified as the underlying representation type.
+  List<Window> l = [w];
+  assert(l is List<JSObject>);
+}
+```
+
+Extension structs are also useful for providing a lightweight facade over an
+existing type.
+
+```dart
+// Natural numbers
+extension struct Nat(int _x) {
+  Nat(int x) : assert(x >= 0), _x = x;
+  Nat.zero() : _x = 0;
+
+  Nat get succ => Nat(_x+1);
+  Nat plus(Nat other) => Nat._x + other._x;
+
+  // Override the underlying isNegative operation (incorrectly)
+  bool get isNegative => true;
+}
+
+void test() {
+  var n1 = Nat(3);
+  var n2 = Nat.zero();
+  assert(n2.succ.succ.succ == n2);
+
+  //The underlying representation is still as an int
+  assert(n1 is int);
+
+  // The static type is used to dispatch the method calls
+  assert(n1.isNegative);
+
+  // If the static type is lost, dispatch goes to the underlying object.
+  dynamic d = n1;
+  assert(!d.isNegative);
+}
+```
+
+Extension structs may delegate members to the underlying field.  They may also
+implement interfaces, but only if the underlying field implements the interface.
+
+```dart
+// Natural numbers
+extension struct Nat(int _x) implements Comparable<num> {
+  // Constructors etc as above
+
+
+  bool get isEven;  // Abstract definition delegates to _x.isEven
+}
+
+void test() {
+  var n = Nat(2);
+  // Same as 2.isEven
+  assert(n.isEven);
+  // Comparable interface allows access to the Comparable methods on int
+  assert(n.compareTo(2) == 0);
+
+  // Since Nat implements Comparable<num>, it may be assigned to it
+  Comparable<num> c = n;
+  // The representation is still as an integer
+  assert(c is int);
+}
+```
+
+Extension structs have no inheritance.
+
+```dart
+// Static error
+extension struct PositiveNumber(int _x) extends Nat {...}
+```
+
+Extension structs do not define a signature, and hence cannot be implemented.
+
+```dart
+// Static error
+extension struct AlternativeNat(int _x) implements Nat { ...}
+
+// Static error
+class MockNat implements Nat {...}
+```
+
+
+## Extension structs in more detail
 
 
 ### Introduction form
@@ -408,6 +631,11 @@ extension struct Data(int x);
 extension struct GenericData<T>(T y);
 ```
 
+**COMMENTARY(leafp):** *The point here is that the actual runtime representation
+  of the extension struct will simply be the value of the single unique field,
+  with no wrapper object*.
+
+
 ### Members of extension structs
 
 An extension struct may contain static and instance member definitions in the
@@ -424,6 +652,10 @@ member of the same name and kind is available on the unique field of the
 extension struct, and the type of the abstract member is a supertype of the type
 of the corresponding member in the unique field.
 
+**COMMENTARY(leafp):** *Abstract members here allow delegation of methods
+  without having to write an explicit forward.  We could elide this*.
+
+
 ### Abstract extension structs
 
 Extension structs may not be be marked abstract.
@@ -431,6 +663,12 @@ Extension structs may not be be marked abstract.
 ### Extending extension structs
 
 An extension struct may not be extended.
+
+**COMMENTARY(leafp):** *This may be contentious.  If we end up needing some form
+  of inheritance, either to build up a subtype hierarchy or to allow code
+  re-use, there are probably paths we can take here, but it will be important
+  for the purposes of this design to keep this consistent with the behavior of
+  general structs*.
 
 ### Implementing extension structs
 
@@ -442,22 +680,46 @@ define interfaces.
 It is an error to use the type defined by an extension struct as a bound on a
 generic type parameter.
 
+**COMMENTARY(leafp):** *This is probably harmless, maybe we should allow it*.
+
 ### Extension structs implementing interfaces
 
 Extension structs may implement interfaces if and only if each implemented
-interface type is a subtype of the type of the unique field in the primary
+interface type is a supertype of the type of the unique field in the primary
 constructor list.  That is, implemented interfaces must be implemented *by the
 field*, which will serve as the underlying object representation.
+
+**COMMENTARY(leafp):** *The driving motivation for this design choice is keep
+  the behavior of extension structs consistent with general structs.  For
+  general structs and classes, implementing an interface means that the newly
+  defined type is both a subtype of that interface, and supports all of the
+  methods of that interface.  If we wish to preserve the former for extension
+  structs, then we must ensure that the underlying representation object also
+  implements the same interface, so that when we assign it, we do not break
+  soundness.  We could give up on fully subtyping and instead only allow
+  assignability, with conversion to the implemented interface requiring boxing,
+  but I have chosen not to do that, since it still makes subtyping unavailable.
+  That is, under this proposal, for an extension struct Foo that implements Bar,
+  `Foo` is assignable to `Bar` with no boxing, and `List<Foo>` is assignable to
+  `List<Bar>`.  If we auto-boxed on assignment to `Bar`, we could preserve the
+  former, but not the latter*.
 
 An extension struct is a subtype of each of its implemented super-interfaces.
 
 An extension struct implements Object.
 
 It is an error if any member of an implemented interface has a non-abstract
-definition in the body of the extension struct.  (This is to avoid the confusing
-behavior that would result from different dispatch behavior depending on whether
-the member is accessed via the extension struct interface, or via the
-implemented super-interface).
+definition in the body of the extension struct.
+
+**COMMENTARY(leafp):** *This restriction is to avoid the confusing behavior that
+  would result from different dispatch behavior depending on whether the member
+  is accessed via the extension struct interface, or via the implemented
+  super-interface.  That is `Foo` is an extension struct that implements
+  `Comparable<Foo>` and defines its own `compareTo` method, then accessing the
+  compareTo method on a value of type `Foo` will call a different method than
+  first assigning the value to a variable of type `Comparable<Foo>` and then
+  calling the method.  It may be that it is too important to support
+  "overriding" here though, and so we may need to relax this restriction.*
 
 
 
@@ -471,7 +733,7 @@ representation" of the extension struct.
 ### Identity of extension structs
 
 The identity operator on an instance of an extension struct is defined to return
-the same value as applying the identity operator to the underlying
+the same result as applying the identity operator to the underlying
 representation of the extension struct.
 
 ### Semantics of members of extension structs
@@ -536,7 +798,16 @@ No copyWith method is generated for an extension struct.
 
 It is an error to define any of the object members in the extension struct.
 
+**COMMENTARY(leafp):** *This restriction is to avoid the same confusing behavior
+  described above in the section on implementing interfaces.  Almost all uses of
+  `hashCode` will be done via the `Object` interface, and it seems dangerous to
+  allow users to define a `hashCode` getter that will be ignored when the value
+  is used as (e.g.) a key in a map.*
 
+
+### Const extension structs
+
+TODO(leafp): This should work, write out the details.
 
 
 
--- a/working/extension_structs/overview.md
+++ b/working/extension_structs/overview.md
@@ -1,0 +1,543 @@
+# Overview of a proposal for structs
+
+Author: Leaf Petersen
+
+Status: In progress
+
+Version 1.0 (see [CHANGELOG](#CHANGELOG) at end)
+
+## Summary
+
+This document is an overview of an approach to solving the following problems:
+  - The desire for a zero cost wrapper type (motivated largely by interop
+   concerns), described briefly
+   [here](https://github.com/dart-lang/language/issues/1474)
+  - The desire to have compact syntax for so called "data classes" or "value
+types", discussed among other places
+[here](https://github.com/dart-lang/language/issues/314)
+  - The desire to have an accounting for data structures without identity as
+    described [here](https://github.com/dart-lang/language/issues/2246).
+
+The proposal here takes two steps.  The first is to add a restricted kind of
+class (provisionally describe as a "struct" here), which gives up some of the
+affordances of general Dart classes in exchange for compact syntax and automatic
+generation of useful methods.  These restricted classes provide data class like
+functionality.  The second step is to support a further restriction on structs
+with a single field which eliminates the wrapper object, representing the struct
+entirely as the underlying object (at the cost of making the abstraction
+entirely static).
+
+This proposal builds on previous proposals in this space, including:
+  - [Views](https://github.com/dart-lang/language/blob/master/working/1426-extension-types/feature-specification-views.md)
+  - [Extension types](https://github.com/dart-lang/language/blob/master/working/1426-extension-types/feature-specification.md)
+
+
+This overview is not intended to be a feature specification: it is designed to
+give a brief overview of the core proposal and set out some design points for
+discussion.  If there is buy-in on these ideas, we may choose to incorporate
+them into an existing proposal, or add a new feature specification.
+
+
+## Design principles
+
+We aim to minimize the differences between structs and classes.  As much as
+possible, structs should behave as restrictions of classes.  We specifically aim
+to avoid as much as possible having different behaviors for the same concept
+(e.g. differences in scoping).
+
+We also aim to minimize the amount of new syntax required, and to maximize the
+amount of new functionality that we can provide relative to the syntactic real
+estate consumed, and the new cognitive load imposed on users.
+
+## Roadmap
+
+The first section describes the full struct feature.  The second section
+describes the restriction of the full struct to support static wrappers.
+
+## Structs
+
+This section describes how structs look and behave, largely by example.  In many
+cases, alternatives or extensions to the core proposal are described in line.
+Larger extensions are postponed to a later section.
+
+### Introduction form
+
+We add a new keyword "struct", which is used in place of "class".  The class
+name (plus generics if applicable) must be followed by a parenthesized list of
+field declarations (more on this below). There may optionally be an "extends"
+clause, and an "implements" list.  No mixins are permitted.  After the header,
+the usual brace delimited list of members may be provided subject to
+restrictions described below.  An empty set of members may be elided in favor of
+a semi-colon.  The simplest possible struct definitions then would look like:
+
+```dart
+struct Data();
+struct GenericData<T>();
+```
+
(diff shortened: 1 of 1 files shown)


working/extension_structs/overview.md
```diff
@@ -0,0 +1,543 @@
...
+
+### Using extension structs as a bound
+
+It is an error to use the type defined by an extension struct as a bound on a
+generic type parameter.
+
+### Extension structs implementing interfaces
+
+Extension structs may implement interfaces if and only if each implemented
+interface type is a subtype of the type of the unique field in the primary
```
chloestefantsova, line 448: DBC: should it say "supertype of the type of the unique field" instead?

leafpetersen: yep, thanks!

working/extension_structs/overview.md
```diff
@@ -0,0 +1,543 @@
...
+#### copyWith
+
+If no copyWith method is defined in or inherited by the struct, then a copyWith
+method will be defined with named parameters for every entry in the primary
+constructor.  For every entry, if the field name is not private, then the
+parameter name is the field name.  If the field name is private, then the
+parameter name is the field name with all leading `_`s removed.
+
+The body of the method calls the generated primary constructor.  For every
+argument which is explicitly passed to the copyWith method invocation, the
```
jakemac53, line 316: How should this method determine if an argument is "explicitly passed"?

leafpetersen: The same way that compilers currently determine this for filling in default values (I believe on the VM via an invocation object, and in dart2js by calling a different entry point).  In other words, compiler magic that users don't get.  I added commentary on this.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,815 @@
...
+void test() {
+  // Call the default constructor, using the default value for w
+  var c1 = Component(0, 0, 0);
+  // Call the default constructor, explicitly passing w
+  var c2 = Component(1, 1, 1, 1);
+
+  assert(c1 == c2); // Equality is defined to be structural
+  assert(c1.hashCode == c2.hashCode); // With correspondind hashCode
+
+  print(c1.debutToString()); // Prints "Component(0, 0, 0, 1)"
```
kevmoo, line 87:
```suggestion
  print(c1.debugToString()); // Prints "Component(0, 0, 0, 1)"
```

working/extension_structs/overview.md
```diff
@@ -0,0 +1,815 @@
...
+ T y;
+ }
+
+```
+
+#### Alternative: Split into positional and named
+
+We could choose to make this look more directly like a constructor argument list
+by allowing "named" parameters inside of a brace delimited set.  These could
+then become named parameters in the default constructor.
```
mit-mit, line 252: Yeah, I was expecting this when I read the introductory example, that I might have written as:

```dart
struct Component(int x, int y, int z, {int w = 1});

void test() {
  // Call the default constructor, using the default value for w
  var c1 = Component(0, 0, 0);
  // Call the default constructor, explicitly passing w
  var c2 = Component(1, 1, 1, w: 1);
...
}
```

working/extension_structs/overview.md
```diff
@@ -0,0 +1,815 @@
...
+Operand replaceIn(Operand oper, Identifier t, Value c) {
+  return switch (oper) {
+    case ConstantOperand(_) : oper;
+    case IdentifierOperand(var i) where i == t: ConstantOperand(c);
+    case IdentifierOperand(_) : oper;
+  }
+}
+```
+
+## Structs in more detail
```
mit-mit, line 133: Is there any support for declaring a struct that is immutable?

working/extension_structs/overview.md
```diff
@@ -0,0 +1,815 @@
...
+
+  // Generated copy method
+  var c2 = c1.copyWith(x : 2);
+  assert(c2.x == 2);
+}
+```
+
+Structs may explicitly define constructors, static members, and normal class
+members (except fields), and may implement interfaces.
+
```
mit-mit, line 99: Maybe add something similar to _"Struct fields can be made private by using a `_` prefix in their name"_ here.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,815 @@
...
+feature correspondingly.  Ignoring that, several alternative syntaxes for
+extension structs are on the table.
+
+##### Alternative modifiers
+
+The choice of the keyword "extension" is intended to leverage users existing
+intuitions about how extension methods work: that is, that they are largely
+statically dispatched.  It's not clear to me that this intuition actually
+carries over, however. There are various alternatives to "extension".
+
```
mit-mit, line 597: I personally find the term "extension struct" quite confusing...

I think my main issue is that I don't think of extension methods so much in terms of their static dispatch, but rather in their utility: the ability to extend the set of methods. As the Dart tour [says](https://dart.dev/guides/language/language-tour#classes) in it's introduction to extension methods: _"Extension methods are a way to add functionality to a class without changing the class or creating a subclass"_. This matches poorly with extension structs, which are much more subtractive in terms on the API.

While not perfect, I think "view" is the best term I've seen so far. I could also get used to "facade" (it matches the design pattern well in terms of intent: _"Provides a simplified interface"_, [source](https://en.wikipedia.org/wiki/Facade_pattern#Usage))).

I'm also a bit in doubt about whether to make the connection to structs or not. While similar in terms of some of the semantics, a regular struct and a view struct are very dissimilar in terms of what they are used for. So I think it might be worth considering just calling them views, even if we keep all the similarities between then suggested in the present doc.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,815 @@
...
+the usual brace delimited list of members may be provided subject to
+restrictions described below.  An empty set of members may be elided in favor of
+a semi-colon.  The simplest possible struct definitions then would look like:
+
+```dart
+struct Data();
+struct GenericData<T>();
+```
+
+#### Alternatives
```
mit-mit, line 150: I really like the term "struct"; it pair nicely with the structural identity. I think it's helpful to not call them classes as they are so different from classes, and views seems even more confusing.

working/extension_structs/overview.md
jakemac53: Ah yeah ok, that makes me sad. I would rather see us make a general user visible feature (its something people have to work around a fair amount today, especially in codegen, and macros will run into it as well).

working/extension_structs/overview.md
kevmoo: https://github.com/tc39/proposal-record-tuple – it'd be great to layer this on this feature in JS – if it's approved

working/extension_structs/overview.md
```diff
@@ -0,0 +1,815 @@
...
+cases, alternatives or extensions to the core proposal are described in line.
+Larger extensions are postponed to a later section.
+
+## Structs by example
+
+Structs allow you to define simple classes very compactly.  For example, we
+might model a component vector as follows:
+
+```
+struct Component(int x, int y, int z, int w = 1);
```
sigmundch, line 71: I like the compactness of the struct declaration, but to be honest I'm torn here.

Two thoughts come to mind:
* is there a more incremental way to introduce structs mapping to previous syntax?
* is the new syntax useful more generally beyond the concept of structs?

One incremental syntax I'd imagine is:

```dart
struct Component {
  int x;
  int y;
  int z;
  int w;

  Component(this.x, this.y, this.z, [this.w = 1]);
}
```

Another incremental step would be to make the constructor implicit (now instead of 0-args as we did with classes, we make struct constructors take all args).

Having a more incremental syntax like this, could potentially enable us to consider the new compact syntax as an alternative to redundant declaration/initialization in classes.  For example, we could make the two following programs equivalent:
```dart
class Foo(int x) extends Bar {
  foo() => x;
}
```

```dart
class Foo extends Bar {
  int x;
  Foo(this.x);
  foo() => x;
}
```

working/extension_structs/overview.md
```diff
@@ -0,0 +1,815 @@
...
+```
+
+Extension structs have no inheritance.
+
+```dart
+// Static error
+extension struct PositiveNumber(int _x) extends Nat {...}
+```
+
+Extension structs do not define a signature, and hence cannot be implemented.
```
sigmundch, line 555: :+1: We definitely want to disallow implementing an interface from a extension struct.

That said, for JSInterop, we are also interested in some form of conformance checking. This is especially useful with mocking. I'll share a few examples here, but I am not sure this is a problem we can solve at this time.

Consider this code:
```dart
extension struct Foo(JSObject o) {
  external bool m1(int x);
}
```
Or the equivalent expansion:
```dart
extension struct Foo(JSObject o) {
  bool m1(int x) => js_util.callMethod(o, 'm1', [x]);
}
```

We'd like to create mocks in Dart that get exported at the JS level, so they can be a valid JSObject to that extension struct (in this case, `Foo` expects the underlying `JSObject` to have certain JS properties and methods).

Our current proposal is to say:
```dart
class MyFooMock {
  bool m1(int y) => false;
}
...
Foo mock = mockify<MyFooMock, Foo>(MyFooMock());
```

The type-parameters to mockify give us the magic we need to take `MyFooMock` and shape it as a `JSObject` with the expectations of `Foo`. 

We'd like that connection to be more explicit in the syntax, and eventually have a static error if `MyFooMock` doesn't conform to the extensions declared in `Foo`. We don't want `MyFooMock implements Foo`, but if If `Foo` had an implied interface `ExtensionStructInterface<Foo>`, we would like to say `MyFooMock implements ExtensionStructureInterface<Foo>`.

What makes this harder, is that it's not true that Foo and the mock need to match, but how Foo uses the JSObject and the mock do. So to make this example more complicated:
```dart
extension struct Foo(JSObject o) {
  @JS('m2')
  external bool m1(int x);

  int m3() => 3;
}
```
Or the equivalent expansion:
```dart
extension struct Foo(JSObject o) {
  bool m1(int x) => js_util.callMethod(o, 'm2', [x]);
  int m3() => 3;
}
```

The ideal mock class should be:
```dart
class MyFooMock {
  bool m2(int x) => false;
}
```

:exploding_head: 

/cc @srujzs

working/extension_structs/overview.md
```diff
@@ -0,0 +1,815 @@
...
+
+It is an error to override the "runtimeType" method of a struct.
+
+It is an error to override the "noSuchMethod" method of a struct.
+
+### Const structs
+
+TODO(leafp): This should work, write out the details.
+
+## Extension Structs
```
sigmundch, line 449: One important constraint we'd like it to have explicitly specified is that we need these extensions to be "sticky", that is, they should be available as part of the type of the expression, and not just based on scoping rules.

Because of this, I'm also not sure that the word `extension` maps well with the existing use of that keyword with classes. 

An idea here could be to use the word `static` instead:

```dart
static struct F(x) {
  int get foo => x.bar;
}
```

`static` here means that the dispatch is static, like extension methods; but also that it is a static (compile-time only) concept and the struct and type name is erased at runtime.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,815 @@
...
+## Extension structs in more detail
+
+
+### Introduction form
+
+Extension structs are defined by adding the keyword "extension" before a struct
+definition.  The syntax for extension structs is otherwise exactly identical to
+that of general structs.  However, extension structs are subject to additional
+restrictions.  The most important of these is that extension structs may only
+contain a single entry in their primary constructor.
```
sigmundch, line 575: If I consider this proposal in isolation, I buy this restriction. However, what if we consider this together with the proposal on records, could we treat more than one entry in the constructor as a record (rather than multiple fields?) Even in the context of regular structs, is the distinction of 3 fields vs 1 record field relevant anymore?

working/extension_structs/overview.md
sigmundch: :heavy_plus_sign: I also find the term _"extension struct"_ confusing. I made a suggestion further below that we could potentially call these `static struct`.


If we didn't have structs, I could see us using `static class` as a way of declaring a zero-cost wrapper. A _static class_ would have practically the same limitations (all methods are statically dispatched, the type is erased and thus only available for static type checking, it has no interface and hence there is no interface to be implemented, etc)

working/extension_structs/overview.md
```diff
@@ -0,0 +1,815 @@
...
+
+## Design principles
+
+We aim to minimize the differences between structs and classes.  As much as
+possible, structs should behave as restrictions of classes, and extension
+structs should behave as further restrictions of structs.  We specifically aim
+to avoid as much as possible having different behaviors for the same concept
+(e.g. differences in scoping).
+
+We also aim to minimize the amount of new syntax required, and to maximize the
```
sigmundch, line 50: Sorry I missed this sentence on my earlier pass. I think this aligns with my other comment about the primary constructor syntax and whether we should consider that a second-tier syntax.

working/extension_structs/overview.md
timsneath: My opinion is worth less than a three day old lettuce leaf. But I like the `record` [name from Java](https://docs.oracle.com/en/java/javase/14/language/records.html), that said. Both because there's prior art from another popular language, and because the word gives me something useful to intuit from.

working/extension_structs/overview.md
leafpetersen: Agreed.

working/extension_structs/overview.md
leafpetersen: Fixed, thanks.

working/extension_structs/overview.md
leafpetersen: Added your example to this section.

working/extension_structs/overview.md
leafpetersen: All structs are shallow immutable.  There's no support for deep immutability in this proposal.

working/extension_structs/overview.md
leafpetersen: Done.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,815 @@
...
+view GenericData<T>();
+```
+
+##### Use a modifier on classes
+
+
+We could use "data class":
+
+```dart
+data class Data();
```
sigmundch, line 189: One thing that draws me to this syntax is that it aligns with the vision that this is a "restrictive" feature on top of what already exists.

Now I can map that a `data class` is a `class` with the following restrictions:
* it can only extend from an empty `data class` that acts as a tagging interface
* it can implement interfaces and override methods like regular classes
* unless otherwise specified, the fields by default structurally identify instances of the data class
* it doesn't export a default interface, and hence cannot be "implemented" by other classes in the program

Similarly a `static data class` (using the proposed `static` modifier i mentioned earlier) can be explained as a `data class` that:
* only supports having a single field (a wrapped object)
* always implements a single interface: the type of the wrapped object (called the wrapped type)
* doesn't support virtual method dispatch, except for the methods in the wrapped type
* cannot shadow/override methods of the wrapped type, but may statically expose a subset of them.
* has its type available at compile-time only, but later gets erased and replaced by the wrapped type at runtime.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,815 @@
...
+  assert(n.compareTo(2) == 0);
+
+  // Since Nat implements Comparable<num>, it may be assigned to it
+  Comparable<num> c = n;
+  // The representation is still as an integer
+  assert(c is int);
+}
+```
+
+Extension structs have no inheritance.
```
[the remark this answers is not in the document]
sigmundch: If the worry with extension is that we are adding fields, this is one restriction that would work for us:

structs can only extend other structs that have the exact same set of fields and default values. In the case of JS interop, it's the same JSObject after all.

working/extension_structs/overview.md
leafpetersen: I added `record` to the alternatives list.

I'm not opposed to breaking the syntactic tie between structs and extension structs (e.g using just `view`).  I continue to feel that the more we can keep the syntax and semantics consistent the better though.  I would like it to be the case that if a user clicks on the type name `Foo`, and it takes them to a definition that looks like

```
<something here> Foo(<maybe something here>) implements Bar {
  ....
}
```

then they can be fairly confident that they understand what `implements Bar` means, etc.  I'm less opinionated about what the `<something here>` is.  I think there's value in tying things together (I initially wanted to just use `class Foo wraps (...)` and `static class Foo wraps(...)` for structs and extension structs, respectively, and I'm still tempted by that.

It's interesting that `static` appeals to you as well @sigmundch - it was my top choice, but I have a hard time justifying it on first principles.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,815 @@
...
+class name + generics.  The primary constructor serves both to define the
+fields, and to define the signature of the default generated constructor.
+
+The primary constructor consists of a parenthesized list of comma separated
+`<type> <identifier> (= <expression>)?` entries.  That is, a list of variable
+declarations: with no modifiers; with types required; and with optional
+initializer values.
+
+Each entry in the list is a field in the struct.  Every field is implicitly
+final.  They may not be late, and a type must be provided.
```
sigmundch, line 209: nit: consider mentioning this earlier in the proposal (it may be too buried in the details that indeed structs are immutable) :smile:

working/extension_structs/overview.md
leafpetersen: > One incremental syntax I'd imagine is:

Yes, I considered that (and have added this to the alternatives section), though not with an explicit constructor - getting a default constructor is a big part of the deal here.   There are two things I don't like.  First, is that it feels very heavyweight relative to the terseness of the primary constructor.  I really do kind of feel that if we do this, we should get maximum bang for the buck, and I think the fact that you can literally just write a single line and get all the goodies is appealing.  Secondly, we need to require these to be final if we're going to use structural identity (I guess technically we could box each mutable field, but bleh), and I don't want to have to explicitly mark them as final.  But I'm also not super keen on `int x`  meaning one thing in one place and a different thing somewhere else.

> For example, we could make the two following programs equivalent:

Yeah, we could - I think  @munificent has a doc exploring that somewhere.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,815 @@
...
+as a de-sugaring.
+
+If no default constructor is defined in the class, a default constructor is
+generated which forwards to the generated primary constructor.
+
+**COMMENTARY(leafp):** *The treatment of initializers as default values here is
+  appealing, but it may be too cute.  It essentially adds non-const default
+  values only for this specific use case.  There's also a question of whether we
+  allow user defined constructors on the struct to also override the default
+  initializer, and if so via what syntax*.
```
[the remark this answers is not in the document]
leafpetersen: I'm not sure I follow you.  My proposal allows you to write an explicit constructor.  This comment is about the fact that if I have:

```
struct Foo(int x = factorial(3));
```

you get a generated default constructor where the default value for the `x` parameter is non-const.  You currently have no way of writing that explicit constructor:

```
struct Foo(int x = factorial(3)) {
  Foo([int x]) : // what to write here?
}
```

working/extension_structs/overview.md
```diff
@@ -0,0 +1,815 @@
...
+
+
+##### Alternative: named parameters
+
+These could be made named parameters.  This feels heavyweight, but has some
```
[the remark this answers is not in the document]
leafpetersen: It just makes the constructor call much more verbose.  Think `child` and `children` in Flutter: there is definitely documentation value to named parameters, but they also definitely add a lot of visual noise to the call site.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,815 @@
...
+advantages.  The restriction on names in the primary constructor would allow us
+to use the non-private versions of the field names as the parameter names.
+
+#### Equality
+
+If no equality method is defined in or inherited by the struct (except from
+Object), then an equality method will be defined which checks that its argument
+has the same runtime type as the receiver (actual runtime type, not the result
```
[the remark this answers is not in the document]
leafpetersen: I think of these roughly as nominal structs.  I noodled for a while with a version where every struct had only a single field (so then extension structs and general structs are closer), and you just use a tuple if you want multiple fields.  I didn't like that design for various reasons, but I could talk some more about it.

working/extension_structs/overview.md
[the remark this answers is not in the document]
leafpetersen: I think there's something that can hang together. I'll try to add a section proposing how this could work on Monday.  It would be useful to know how much is actually needed though.  Do you need mixins?  Do you need interfaces?  Do you need concrete superclasses, or are abstract superclasses enough?

working/extension_structs/overview.md
leafpetersen: Yeah, I don't really see a nice way to get this.  One thing you could do with my proposal (almost) is define `FooInterface` as an abstract class, and say that `Foo implements FooInterface`, and then you can also have `class MyFooMock implements FooInterface` .  The `m1` -> `m2` mapping you'd be on your own for... :) but maybe you could mark that in the interface.  To get this from my proposal though, you need to make two adjustments (I think)
  - You need to allow members from interfaces to be "overridden" in the extension struct (there's nothing problematic with this semantically, but as I discuss in the proposal it has slightly surprising semantics)
  - You need to say that JSObject implements all interfaces (or at least interfaces marked in certain ways), since I require that the "on type" implements the interface as well.

working/extension_structs/overview.md
sigmundch: > ... I really do kind of feel that if we do this, we should get maximum bang for the buck, and I think the fact that you can literally just write a single line and get all the goodies is appealing

Yeah, agree, that's very appealing.

Just to be clear: I'm not vouching for dropping all new syntax. Instead, I'd really like to have a layered approach. Just like UI-as-code: we'll recommend the new syntax, but we can use the desugared form to guide developers in how the new concepts work, highlight how we are mostly adding restrictions that can be leveraged for optimizations, and possibly reduce the implementation cost.

working/extension_structs/overview.md
leafpetersen: Yes, this is the semantics (I'm taking this as a given for this discussion).  I also like static quite a bit for the modifier.

working/extension_structs/overview.md
leafpetersen: Yes.  I hesitate about it a bit though.  There are three interpretations I considered for this:
  - `extension struct Foo(int x, int y)` is represented as a two-tuple.  When you call the constructor, there is an auto-boxing into the tuple representation.
    - I don't like this because auto-boxing.  I'd at least like something there as an opt in.
  - `extension struct Foo(int x, int y)` is represented as a two-tuple, and the constructor has a single tuple parameter
    - the boxing is still there, but at least the user knows it
    - the syntax ends up bleh `var x = Foo((3, 4))`
  - `extension struct Foo(int x, int y)` is represented as two objects, not one.  It is not a subtype of object.  You cannot use it polymorphically.  You can have variables of type Foo, and they have move semantics.
    - This is intriguing - I like the new functionality it gives you
    - This is weird, and it may surprise people writing APIs (they may innocently add a field, and be surprised that weird things happen)

working/extension_structs/overview.md
leafpetersen: Yeah, I could definitely get behind this story.

working/extension_structs/overview.md
leafpetersen: done

working/extension_structs/overview.md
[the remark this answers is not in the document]
sigmundch: I believe the main requirement we have is concrete superclasses. I don't believe we need mixins or interfaces, but @srujzs and @rileyporter may have more context here.

A common example is something like this:
```dart
static struct Node(JSObject o) {
  external get childNodes;
}

static struct Element(JSObject o) extend Node(o) {
  external get name;
}
```

(even though they have no body as declared, I'd default to reason about `external` as a concrete members, since we also have scenarios where we need to add method bodies)

Another common example is that we want to have some features layered:
'default.dart':
```dart
static struct Element(JSObject o) {
  external addEventListener;
}
```

'compat.dart':
```dart
import 'default.dart' as default;
static struct Element(JSObject o) extends default.Element(o) {
  Stream<Event> onClick => ....
}
```

`compat.dart` provides some backwards compatibility while we migrate from one API to another by using additional extensions.

This could be done with double wrapping too, but because the first wrapper has no interface, it makes it harder to "reexport" it's API.

working/extension_structs/overview.md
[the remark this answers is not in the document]
munificent: For what it's worth, when I last dug into it, about half of constructors use positional parameters and half use named parameters. Thanks to the Flutter's framework design, named parameters have become very common in constructors.

(This is one of the main reason I've found it hard to come up with a nice data class syntax, because neither positional nor named parameters have a clear majority and supporting both is a hard syntax problem.)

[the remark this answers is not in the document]
leafpetersen: I'm not actually strongly married to either choice here.  The pitch of this proposal is we be opinionated and ship a single default, but this was one where I only weakly came down on this side.  And we could allow named with fairly minimal boilerplate, I think (just use braces in the primary constructor).

working/extension_structs/overview.md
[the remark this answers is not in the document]
leafpetersen: The main problem that I have with concrete superclasses is actually not with extension structs, but with regular structs.  I was trying to avoid structs having virtual fields.  I can probably still cobble something together, I'll take a crack next week.

working/extension_structs/overview.md
munificent: > One thing that draws me to this syntax is that it aligns with the vision that this is a "restrictive" feature on top of what already exists.

I don't see much value in having a feature that users have to opt in to using and in return... it just prevents them from doing things. I think restrictive features makes sense when in return for fitting within some restrictions, you get other *addition* for free. If it's just "add this modifier and the compiler will yell at you more", I don't find that very compelling. I understand the general software engineering argument for restrictions making things easier to maintain, but I don't seem those really applying to data classes. A *user* of a data class shouldn't care whether it was implemented using a `data` modifier or by hand.

What I *do* find very compelling is, "Add this modifier and you'll get implementations of `==`, `hashCode`, and `copyWith()` for free." That's essentially what `data` does in Kotlin and `record` in Java and C#. Ideally, any time a user wants those methods with the behavior that the syntactic sugar would produce, then they should be able to use this modifier to get it. The fewer restrictions involved, the more often they'll be able to reach for this modifier and get nicer code.

So, in order to provide implementations of those methods, does the language really need to care if all the fields are final or if the class has an implicit interface? In other words, which of these restrictions are essential?

leafpetersen: I think you're misunderstanding the usage of "restrictive", which comes from me.  What I mean by that is not "I restrict you from doing things!", but rather "as much as possible, these things behave exactly like classes, which you are familiar with, except that there are some things you can't do anymore, in order that we may provide you with these benefits".  This is in contrast to some other proposals we've made in this area where "extends" or "implements" meant something fairly substantially different.

> If it's just "add this modifier and the compiler will yell at you more", I don't find that very compelling

I don't think there's anything that I'm doing in here where this is the motivation.

> The fewer restrictions involved, the more often they'll be able to reach for this modifier and get nicer code.

Agreed.  But also, the more benefits they get, they more often they'll reach for this modifier and get nicer code.  Essentially all of my choices are (at least intended to be) driven by benefits.

> So, in order to provide implementations of those methods, does the language really need to care if all the fields are final or if the class has an implicit interface? In other words, which of these restrictions are essential? (I mean, obviously, a mutable class with value-based == and hashCode is gonna behave weird if you mutate it. But if the user wants to do that because they know they'll never stuff it in a hash table, why not let them?)

It is very much a goal of this to provide/allow structural identity.  If you don't agree with that goal, then you won't see that as a benefit (but I do).  If you take that as a goal, the semantics get insane to a degree that I'm not comfortable with if you allow mutability.

For the implicit interface, there are at least three aspects that I see as the benefit: one is performance, another is that you get ADTs (sealed families) for free, another is that you get field promotion for free.  We could certainly make the ADT thing opt in.  I don't love that it requires yet more boilerplate, but I'm not intensely opposed.  For field promotion.... maybe we add something to interfaces that requires them to be "stable"?  But then there's more leakage.

working/extension_structs/overview.md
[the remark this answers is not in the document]
sigmundch: Thinking more about my last sentence above - I think we could go a long way with double-wrappers if we can find a way to forward the wrapper static interface:

```dart
static struct A(int x) {
  bool isFour => x == 4;
}
static struct B(A a) implements A {
  bool isFive => a == 5; // this implicitly means a.x == 5
}

main() {
  B b = B(A(generateInt()));
  print(b.isFour); // valid
```

I don't like the word `implements` here, since it seems to imply that `A` has an interface. But the notion that we can expose parts of the underlying wrapped API statically. But maybe keywords like `show` (if I recall I think Erik once proposed that) could work

munificent (commented):
Overall, there is a whole lot of good stuff in this proposal and I'm pretty excited about it.

I left a bunch of comments mainly to braindump stuff that came to mind as I read it so I don't forget before I get back from vacation, but don't feel any need to respond to them.

working/extension_structs/overview.md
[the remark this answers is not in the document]
leafpetersen: I think this is what I would make `extends` do, FWIW.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,870 @@
...
+```
+
+Structs may explicitly define constructors, static members, and normal class
+members (except fields), and may implement interfaces.  All fields must be
+declared in the header (the "primary constructor"), and are implicitly final
+(making all structs shallowly immutable).  Fields may be marked as private as
+usual by naming them with a leading `_` in their name.
+
+```dart
+struct Component(int _x, int _y, int _z, int _w = 1)
```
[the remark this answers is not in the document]
leafpetersen: @sigmundch proposed the same thing: make primary constructors a thing (in general), and just have them be available for structs.  I'm open to it.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,870 @@
...
+```
+
+##### Use a modifier on classes
+
+
+We could use "data class":
+
+```dart
+data class Data();
+data class GenericData<T>();
```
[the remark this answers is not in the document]
leafpetersen: @sigmundch had a similar proposal, but using `static data class` instead of your `struct` proposal.  I like his proposal a little better - `struct` on its own doesn't mean much to me for the `view` semantics.  I'm very open to alternative syntax here though.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,870 @@
...
+`<type> <identifier> (= <expression>)?` entries.  That is, a list of variable
+declarations: with no modifiers; with types required; and with optional
+initializer values.
+
+Each entry in the list is a field in the struct.  Every field is implicitly
+final.  They may not be late, and a type must be provided.
+
+**COMMENTARY(leafp):** *The restriction to final fields is both because that's
+  the common use case, and because structural identity doesn't make much sense
+  if we allow them to be mutable.  We could potentially box each mutable field
```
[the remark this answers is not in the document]
leafpetersen: > We'd probably want to loosen those to allow non-const values too.

shhhh, you'll give us away.  ;). Yes, we probably would.

> But that also would make it harder to generalize to a more broadly available "primary constructor" syntax that you could use for regular class declarations too.

Would it be too weird if we did generalize it, but only for final fields so we could continue to use `int x` to mean `final`?  Or did allow mutable fields in primary constructors, but made you write `var int x` or some nonsense to get mutable?

working/extension_structs/overview.md
```diff
@@ -0,0 +1,870 @@
...
+corresponding parameter is passed on as the argument to the corresponding
+parameter of the generated primary constructor.  For every argument which is not
+passed to the copyWith method invocation, the value of the corresponding field
+from the current instance is passed on as the argument to the corresponding
+parameter of the generated primary constructor.
+
+**COMMENTARY(leafp):** *As with constructors, this method cannot be generated as
+  a strict de-sugaring, since the ability to detect whether or not an argument
+  has been passed is not available in Dart.  Implementations already support
+  this for default values, so it is likely not problematic to implement*.
```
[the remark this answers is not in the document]
leafpetersen: Yup.  Someone should make a proposal for that... ;)

working/extension_structs/overview.md
```diff
@@ -0,0 +1,870 @@
...
+
+With the addition of pattern matching and switches, structs will support closed
+families (algebraic datatypes).
+
+```dart
+abstract struct Operand;
+
+struct ConstantOperand(Value c) extends Operand;
+
+struct IdentifierOperand(Identifier i) extends Operand;
```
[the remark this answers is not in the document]
leafpetersen: Yeah, this would be really nice.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,815 @@
...
+"debugToString" on the `i`th field if that field is (dynamically) a struct, and
+otherwise the result of calling "toString".
+
+**COMMENTARY(leafp):** *This needs a bit of work.  If we keep the dynamic check
+  for "struct-ness", then we need do deal with the case that the user defines a
+  debutToString thing with the wrong type*.
+
+**COMMENTARY(leafp):** *If we keep this, we may wish to specify that compilers
+  may choose to make it a link time error to have an invocation of debugToString
+  in the program outside of asserts, and other debutToString methods*.
```
rakudrama, line 462: debutToString -> debugToString.

We might not need a link-time restriction - one of the reasons that `toString` cannot be tree-shaken is that it is called in interopolation and various places like `print` and `StringBuffer.write`.

We might consider that `debugToString` might return something richer than a String that allows a linear or structured presentation.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,815 @@
...
+
+
+#### copyWith
+
+If no copyWith method is defined in or inherited by the struct, then a copyWith
```
rakudrama, line 467: What would an inherited `copyWith` method do that is 'correct'?

working/extension_structs/overview.md
rakudrama: Defaulting to an existing field would be something new. Today we can default only to a constant.
If computing the field value is really calling a getter that may have overrides, we might as well allow arbitrary expressions as default values.
I hope there is somewhere between constants and arbitrary expressions.

If we allow `copyWith` to be torn off, the tear-off can be passed to `Function.apply`.
`Function.apply` on dart2js does not use the multitude of entry points, which all match come call site, but a fixed entry point plus metadata.
Changing the defaulting to handle non-constants will make the metadata need to encode receiver-dependent defaults.

working/extension_structs/overview.md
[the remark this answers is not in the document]
rakudrama: Algebraic types tend to grow to be a discriminated union over many elements. dart2js's SSA HInstruction type has O(100) subclasses in a rich hierarchy. 'pattern matching' is done via a visitor pattern that has base class visitors that delegate to a 'visit' method for the abstract superclass. This is all tedious and error-prone.

This kind of hierarchy is often thousands of lines of code with each element of the hierarchy implementing dozens of methods.

It would be great if the class hierarchy + visitor patterns could be replaced by a hierarchical algebraic (but partially mutable) types.

What would be most disappointing is if a simple hierarchy could be written using something like the proposed syntax above but as the code-base evolves at some point it needs to be completely rewritten in the hierarchy+visitor pattern.

Can multilevel algebraic types with mutable fields be supported?

```dart
class Operand switch {
  bool get isSimple;  // abstract member that must be implemented by each 'case'
  case ConstantOperand switch {
    bool get isSimple => true;
    method() { ... }  // methodapplicable to all ConstantOperands
    case StringConstantOperand(String stringValue) {
      List<int> get utf8Bytes => ...;
    }
    case IntConstantOperand(int intValue) {
    }
    ...
  }
  ...
}
```

working/extension_structs/overview.md
srujzs: > What makes this harder, is that it's not true that Foo and the mock need to match, but how Foo uses the JSObject and the mock do. (etc.)

I mentioned one complication for this earlier on how JS members may have characters in their name that aren't supported in Dart, but maybe it doesn't happen often. Either way, you'd still need a `mockify` call to wrap it and all that, so it's just slightly cleaner. One reason I wanted to match the Dart names is that the interfacing we describe is actually possible (since `MyFooMock.m1` implements `Foo.m1`, whereas `MyFooMock.m2` does not implement `Foo.m1`, since the language does not care about our renaming annotations).

> One thing you could do with my proposal (almost) is define FooInterface as an abstract class...

Yeah, you'd still need to ensure that `MyFooMock` implements all the members in `Foo`, unless you start adding weird restrictions that `MyFooMock` doesn't add any new members.

P.S. I may be reading the proposal wrong, but wouldn't we need `JsObject` to implement the interfaces in order for an extension struct to implement those interfaces?

It's not *that bad* if we can't get interfaces for extension structs, it's just a *slightly* worse user experience for writing mocks.

working/extension_structs/overview.md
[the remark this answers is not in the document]
srujzs: We don't really support mixins today for interop and I think I've maybe seen a total of 2 bugs that tried to use a mixin with interop. They may work today in some cases, but I do not see a strong need for them to be included in the future. I talk about interfaces in the other comment chain.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,815 @@
...
+
+Extension structs are targeted at relatively niche uses where you wish to define
+a set of statically dispatched methods layered on top of an underlying
+representation, without introducing a wrapper object.  A canonical use case
+driving this design is to be able define a Dart typed interface for methods on a
+Javascript object, providing a wrapperless interoperation capability.
+
+
+```dart
+extension struct Window(JSObject o) {
```
srujzs, line 521: Can you illustrate what a non-default constructor might look like for extension structs? I ask because it might be tempting for a user to write something like the following:

```
extension struct Window(JSObject o) {
  Window.cons() => Window(js_util.callConstructor(...));
}
```

to avoid duplicating logic to construct `o` every time.

working/extension_structs/overview.md
mit-mit: Another option, borrowing from a discussion above, is to make these all kinds of classes:

* `class`: An abstraction over a collection of fields with associated behavoir
(methods). Defines a reference type.

* `data class`: A special class optimized for holding data. All fields are final. haskCode, copyWith, adn == (with structural identity) provided for free. Defines a value type.

* `view class` or `facade class`: A special class for creating an alternate API of another type. Limited to a single field; the underlying type. The view is lightweight, and invocations on the underlying type are dispatched statically.

`data class` is what this proposal currently calls `struct` and `view/facade class` is what is currently `extension struct`.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,870 @@
...
+```dart
+// Static error
+extension struct AlternativeNat(int _x) implements Nat { ...}
+
+// Static error
+class MockNat implements Nat {...}
+```
+
+
+## Extension structs in more detail
```
eernstg, line 621: I'm not really convinced that `extension structs` should be `structs`: The problem is that an extension struct may have an interface which is the entire interface of the underlying representation (the type of the unique instance variable), and then some added members, and there is no reason to assume that this is a proper interface for an immutable entity to have.

If we hadn't had this ability to skip the indirection (so we can do `myES.foo()` meaning `myES.uniqueField.foo()`) then the outermost object (existing or not) would be a proper (shallow-)immutable object, and everything would fit nicely. On the other hand, that object wouldn't be able to do _anything_ other than calling the getter `uniqueField` (plus, presumably the members of `Object`).

When we conflate the wrapped object with the wrapper (whether or not it exists), we get this conflict.

For instance, if we were to introduce some kind of boxing, I'd assume that the boxed class would simply emerge by taking the `extension struct` declaration and delete the word `extension`. We would then have a declaration that specifies a wrapping entity, wrapping a single object. It would be a regular, proper object, and it would have OO dispatch for its methods, and so on.

However, that's a struct, but it would now be a struct that contains forwarding methods to the representation object, including, possibly, some getters and setters which are forwarding to the implicitly induced getters and setters of the mutable instance variables that the representation object has.

This is a weird kind of `struct`, because it's simply a forwarding wrapper that exposes the entire interface of the wrapped object, including whatever mutability that object offers.

Why, then, would we even have the connection between the `struct` mechanism and the view mechanism known as `extension struct`? As far as I can see we could just as well use a normal class when we're boxing an extension struct.

eernstg (commented):
Commented on several things. The main thing is the comment on line 49, and the invariant in the comment on line 200 starting 'We can ensure this by enforcing'. Finally, I've commented in line 621 on the very idea that the `extension struct` mechanism should be related to the `struct` mechanism.

I'd love to have value classes, and the name could very well be `struct` (because they are more structural than other classes) or something with `value`.

working/extension_structs/overview.md
leafpetersen: Yes, this is definitely a possibility, and is growing on me a bit.  We could also split out the primary constructor syntax, and essentially ship three features (possibly separately):
  - Primary constructors
  - data classes (which can (or must?) use primary constructors)
  - view classes (which can (or must?) use primary constructors)

working/extension_structs/overview.md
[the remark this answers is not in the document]
leafpetersen: Yes, I think that's my initial take on adding `extends`.  We say that `extends` means that all methods on the superview are available on the subview, but you can't override them.  @sigmundch @srujzs do we need to allow override?  We can do so, but it has the obvious surprising behavior when you assign from a subview type to a superview type and then access the method (and get the superview method rather than the subview one).

@sigmundch @srujzs do you feel that we need to allow cross-library extension, or can we keep views "open in the library, sealed externally"?

working/extension_structs/overview.md
leafpetersen: > P.S. I may be reading the proposal wrong, but wouldn't we need JsObject to implement the interfaces in order for an extension struct to implement those interfaces?

@srujzs I'm not entirely sure I understand your question - are you asking about mocking, or in general?  In general, yes, this is a question that I wanted to get input from you all.  It seems "reasonable" to me, to say that JSObject implements all interfaces for the purposes of this conformance check.  So then you can write something like `view class MyJSArray(JSObject o) implements List<dynamic>;`.  It does of course allow you to do all sorts of nasty things: you can pass in any old JSObject to create a MyJSArray, and if you assign it to a `List<dynamic>` chaos ensues.  You can add validation in the constructor if you want (in fact, we could probably even specify this in general), but casts can still get around this.

working/extension_structs/overview.md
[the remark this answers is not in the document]
leafpetersen: Yeah, that's a way we could go.  We could say that:
  - Everything gets an optional primary constructor
  - For a `struct` (or whatever we call them) all fields are implicitly final, whether declared in the primary constructor or not.

working/extension_structs/overview.md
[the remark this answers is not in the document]
leafpetersen: > Can multilevel algebraic types with mutable fields be supported?

I think there are two separate pieces of your question.  I think the patterns/sealed family proposal from @munificent would completely support your use case: you get hierarchies of normal classes, with exhaustiveness checking.  The one restriction is that they need to be in the same library.

There's a separate question around a nice short (and possibly hierarchical) syntax for specifying a sealed family.  I think @munificent has some ideas around that as well, along the lines of what you sketch out above.

working/extension_structs/overview.md
leafpetersen: Nothing useful.  Probably should just disallow it.

working/extension_structs/overview.md
leafpetersen: My initial pitch would be that you just use the normal constructor syntax.  So your example would look like:

```dart
extension struct Window(JSObject o) {
  Window.cons() : o = Window(js_util.callConstructor(...));
}
```

Factory constructors are also available.  

If that's not enough, we could consider allowing the field to be initialized in the constructor body, with the restriction that `this` cannot be accessed before the field is definitely assigned.  That's more of an implementation lift though.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,870 @@
...
+  - The desire to have compact syntax for so called "data classes" or "value
+types", discussed among other places
+[here](https://github.com/dart-lang/language/issues/314)
+  - The desire to have an accounting for data structures without identity as
+    described [here](https://github.com/dart-lang/language/issues/2246).
+
+The proposal here takes two steps.
+  - The first is to add a restricted kind of class (provisionally describe as a
+"struct" here), which gives up some of the affordances of general Dart classes
+in exchange for compact syntax and automatic generation of useful methods.
```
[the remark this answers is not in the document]
leafpetersen: I'm definitely open as to what we do/don't build in.  I think there's probably a minimal set that should be built in though (e.g. equality, hashCode, and probably copyWith).  I'll file an issue on this.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,870 @@
...
+```
+
+## Structs in more detail
+
+### Introduction form
+
+We add a new keyword "struct", which is used in place of "class".  The class
+name (plus generics if applicable) must be followed by a parenthesized list of
+field declarations (more on this below). There may optionally be an "extends"
+clause, and an "implements" list.  No mixins are permitted.  After the header,
```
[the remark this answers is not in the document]
leafpetersen: > Perhaps it should not be allowed for a public struct to have private instance variables.

I'm not sure I see the motivation for this.  It seems very reasonable (even desirable) to me - you can have structs with public data (i.e 1:1 mapping between fields and getters) or structs where the backing data is intermediated with getters.  I do see value in this being transparent to the *compiler*.

> but shared state can be expressed in abstract structs used as common superclasses of other structs.

I don't currently allow this, but we could consider it, especially if we restrict it to a single library.

> I also assume that no struct can inherit a mutable instance variable.

As proposed so far, no struct can inherit any instance variables (a struct can implement something with a setter though).  If we allow super-structs to provide some of the fields, they would, of course, still be immutable.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,870 @@
...
+
+### Members of structs
+
+A struct may contain static and instance member definitions in the same way as a
+class, with exactly the same syntactic resolution.
+
+For scope resolution purposes, entries in the primary constructor list are
+treated exactly as if they were defined as instance members on the struct.
+
+It is an error for a struct to define a field as a member.
```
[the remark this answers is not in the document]
leafpetersen: Could be.  Feels a little overly restrictive to me, with no benefit, but I'll file a discussion issue.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,870 @@
...
+}
+struct Data(int x, List<int> l = [3]) extends Foo ;
+struct GenericData<T>(T x, T y) extends Foo;
+```
+
+Members are inherited from super-structs as usual.
+
+It is an error for a struct to extend a class.
+
+It is an error for a struct to be extended outside of the defining library.
```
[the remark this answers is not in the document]
leafpetersen: Yes.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,870 @@
...
+abstract struct Foo {
+  int foo() => 3;
+}
+struct Data(int x, List<int> l = [3]) extends Foo ;
+struct GenericData<T>(T x, T y) extends Foo;
+```
+
+Members are inherited from super-structs as usual.
+
+It is an error for a struct to extend a class.
```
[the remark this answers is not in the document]
leafpetersen: yes, probably.  Though I wonder if it would make more sense to say that structs implement Object?

working/extension_structs/overview.md
```diff
@@ -0,0 +1,870 @@
...
+entry in the primary constructor list, each with the obvious type.
+
+If any entry has an initializer value provided, then every entry after (and
+including) that entry is an optional parameter with no default.
+
+For every entry with an initializer value, if no argument is passed for that
+parameter, the initializer value is assigned to that parameter in the
+initializer list of the constructor.  Note that this requires the ability to
+detect whether or not a parameter was passed, which is not expressible strictly
+as a de-sugaring.
```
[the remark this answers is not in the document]
leafpetersen: Yes, it would be nice to generalize this.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,870 @@
...
+
+##### Alternative: named parameters
+
+These could be made named parameters.  This feels heavyweight, but has some
+advantages.  The restriction on names in the primary constructor would allow us
+to use the non-private versions of the field names as the parameter names.
+
+#### Equality
+
+If no equality method is defined in or inherited by the struct (except from
```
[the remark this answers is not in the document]
leafpetersen: I say above that generated methods are only generated for concrete structs.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,870 @@
...
+them into an existing proposal, or add a new feature specification.
+
+
+## Design principles
+
+We aim to minimize the differences between structs and classes.  As much as
+possible, structs should behave as restrictions of classes, and extension
+structs should behave as further restrictions of structs.  We specifically aim
+to avoid as much as possible having different behaviors for the same concept
+(e.g. differences in scoping).
```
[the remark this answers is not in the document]
leafpetersen: I think this is a good summary.  I think the question around `implements` is the key question.  I don't really *like* that the extension struct doesn't allow you to implement an interface that the representation type doesn't, but I think it leads to a more straightforward feature for the consumer of APIs, even if it is more limited for producers.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,870 @@
...
+We may wish to also define additional generated members.  For example, we may
+wish to have `parse` and `unParse` methods.  TODO(leafp): consider fleshing
+something out here.
+
+
+#### Restrictions
+
+It is an error to override the "runtimeType" method of a struct.
+
+It is an error to override the "noSuchMethod" method of a struct.
```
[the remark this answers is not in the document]
leafpetersen: In the core proposal, I'm not allowing any overriding, because of the surprising behavior on upcasts.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,870 @@
...
+The toString method delegates to the underlying representation.
+
+#### debugToString
+
+If no debugToString method is defined in the extension struct, then a
+debugToString method will be defined which returns a formatted description of
+the receiver, of the form `<type>(<f0>, ..., <f1>)` where `<type>` is the
+extension struct type via which the receiver is called, and the `fi` are the
+result of calling "debugToString" on the `i`th field if that field is
+(dynamically) a struct, and otherwise the result of calling "toString".
```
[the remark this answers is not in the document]
leafpetersen: yep.  clarified.

working/extension_structs/overview.md
leafpetersen: I'm open to splitting out extension structs and structs (e.g. into `data class` and `view class`), but I mostly think the mutability argument is a red-herring.  Structs aren't deeply immutable, and you can freely add forwarding methods that mutate the underlying fields.  Similarly, and extension struct is shallow (not deeply) immutable in the sense that it's not an lvalue, but it does allow delegating mutating members to the unique field.   It's true it makes it easier to do so in a way that somewhat obscures the difference though.

Put another way, any interface that can be used in the implements clause of an extension struct can also be used in the interface of a regular struct.

working/extension_structs/overview.md
leafpetersen: I wonder whether this should just be a static method?  As long as we keep extension within the library, it works out, and it makes "extends" work out better.

working/extension_structs/overview.md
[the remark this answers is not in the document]
srujzs: > We can do so, but it has the obvious surprising behavior when you assign from a subview type to a superview type and then access the method (and get the superview method rather than the subview one)

Maybe I'm coming at this weird, but this doesn't seem surprising to me. :) You "view" it differently, so you're going to get a different behavior/method. I personally would want this behavior (but maybe the deviation from classes is indeed confusing).

> do we need to allow override?

In the case where all the methods are external anyways, no, because JS will do the appropriate dynamic dispatch. Obviously, it gets a little more complicated in the cases where you want to do define non-external methods for whatever reason. One instance I can see in `dart:html` is different behavior for `setInnerHtml` for sanitization reasons (see `TemplateElement`).

working/extension_structs/overview.md
srujzs: Not sure how I missed your second bullet point earlier - you already answered it there, sorry. Having `JsObject` implement *all* interfaces is indeed chaos, and I think that would be worse than supporting the interfacing behavior.

It seems like it's fundamentally hard to get this behavior (where an extension struct can provide an interface that is used for another class/struct). Picking your brain here a bit I guess, but why do we want to disallow structs defining a signature?

working/extension_structs/overview.md
srujzs: Oh, this is great! External constructors might be a bit more complicated if we intend to support that (perhaps they just desugar to something like the above), but this should cover any cases we have today for interop constructors (and more).

> allowing the field to be initialized in the constructor body

This might be useful in the case where you have a complicated initializer, but I supposed you can always have a separate initialization method that you can call to instantiate `Window` e.g. `: o = Window(initializer())`.

leafpetersen:
Thanks for all the comments!  I added a short section on some possible extensions, including support for more extensive sub-struct relationships.  I'll leave this open for another day for any additional comments, and then plan on landing this and filing discussions issues for a number of open design points (including questions raised here).

working/extension_structs/overview.md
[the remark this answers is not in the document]
leafpetersen: > Maybe I'm coming at this weird, but this doesn't seem surprising to me. :)

It's not surprising to me either, but my goal with this proposal was that for *consumers* of the API, they should as much as possible be able to guess what things would do based solely on knowing what classes and objects do.  That is, I want developers to not really have to know what views are to be able to successfully use an API written using them.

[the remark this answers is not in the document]
sigmundch: Re: need for overrides. I believe we do, but to a lower degree. @rileyporter has examples from the html prototype that illustrate the need for us to shadow a property with a new implementation in a new static view. In our case it is because we want to use a static view to provide backward compatibility to an old API that happens to use the same property name as a new API.

Re: cross-library extension. We believe we need it too. Our approach for exposing html APIs will require a layering approach with an initial set of views defined in one library, and a separate library defining additional static views that extend the base ones.

A couple thoughts:
 * It may be worth revisiting the terminology. The `extends` word (and similarly `implements`) gives the illusion that we have virtual dispatch, and hence the possibility of `overrides`. I personally don't see this part of the feature as an "override", but as a shadowing (more below). So maybe using a different terminology may help our end users as well?
 *  I see a static view as a way of encapsulating access to an underlying object, so I'd like a static view to have a choice of what that static API will be. For example, if I don't want to expose anything from the underlying object I should be allowed to do so. With copy/paste we can achieve anything, but I'd hope we can make the language syntax support the common patterns.

Here are some of the scenarios I'd like to support:

1. Create opaque views:

```dart
// A1 is an opaque view of `int`
// the original methods from `int` are not exposed.
static struct A1(int x) {
  bool isFour => x == 4;
}
```

2. Create opaque views with methods that shadow what exist in the underlying `on` type.
```dart
// A2 can expose APIs with a name that conflict
// with existing APIs in the `on` type, no problem:
static struct A2(int x) {
  bool isEven => throw "WAT";
}
// Yes (A2(4) as int).isEven is true
```

3. Create transparent views, that expose methods from the underlying `on` type.
```dart
// A3 has the same methods as `int`.
// We could say that we always expose all methods on `Object` by default.
static struct A3(int x) show int.* {
}

// The "show" is syntactic sugar for writing:
static struct A3(int x) {
  bool isEven => x.isEven;
  bool isOdd => x.isOdd;
  ... // all other int methods here
}
```

4. Create static views exposing some underlying APIs and shadowing others:
```dart
// A4 hides `isEven` from int, and as result it's OK to shadow it:
static struct A4(int x) show int.* hide int.isEven {
 bool isEven => throw "what";
}
```

5. Detect when you are inadvertently shadowing an API:
```dart
static struct A5(int x) show int.* { // oops this shows isEven
 bool isEven => throw "what"; // ERROR: can't shadow it too!
 // I could be convinced however of some shadowing/scoping rules,  but no conflict
 // like we do with our import syntax seems like a good starting point
}
```

6. Derive static views from other static views:
```dart
// If we use composition as the syntax for deriving subviews, it may look as such:
// This exposes the static members in A1, but still hides anything from `int` (since int was opaquely hidden in A1)
static struct B6(A1 a) show A1.* {
  bool isNotEven => !(a as int).isEven; // cast needed because A1 was opaque and didn't expose isEven
  // but A1.isFour is available on B6 variables
}
```

Thoughts?

working/extension_structs/overview.md
leafpetersen: I don't really know what it means for an extension struct to define a proper interface.  How does a class implement that interface?  I can in principle allow extension structs to implement other extension structs, but I'm not sure that really helps.

We should talk about how you want to deal with JsObject.  It feels to me like we should be able to find a way to mark specific uses of JsObject as implementing specific interfaces, if we want to do that.  I can imagine something like:

working/extension_structs/overview.md
leafpetersen: You can't access `this`, so `intializer` can't be an instance method.  But otherwise yes.  And you do have factory constructors.  I think I should probably change this to make the generated constructor always be named `._`, so then you could do:

```
extension struct Window(JSObject o) {
  factory Window.cons() {
     var o = Window(...);
      return Window._(o);
  }
}
```

eernstg (approved):

working/extension_structs/overview.md
srujzs: I'm guessing there was supposed to be more to this comment from the last sentence.

> I don't really know what it means for an extension struct to define a proper interface. How does a class implement that interface?

I'd have to imagine it's some synthetic class that's created for that view that can be implemented (like Siggi mentions), but that seems heavy and ugly. Since classes can't implement structs, boxing (with your proposed `.struct`) doesn't help here either.

Going back a bit here to Siggi's comment:

> We don't want `MyFooMock implements Foo`

*Why* don't we want this, again? I'd expect the class to conform to every member signature (and if needed, the singular field's signature as well), just like with classes, but I feel like I'm missing something.

> I can in principle allow extension structs to implement other extension structs, but I'm not sure that really helps.

Yeah, the lack of state from extension structs (since we don't have fields) would make mocks less useful anyways, so I don't think this is needed.

working/extension_structs/overview.md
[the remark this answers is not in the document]
leafpetersen: This proposal supports all of your examples, 1 through 6, though in some cases with slightly more verbose syntax.

> It may be worth revisiting the terminology. The extends word (and similarly implements) gives the illusion that we have virtual dispatch, and hence the possibility of overrides. I personally don't see this part of the feature as an "override", but as a shadowing (more below). So maybe using a different terminology may help our end users as well?

So, this proposal is very specifically trying to avoid doing what you are proposing here.  It's almost the heart of the design principle here.  I don't think we should be designing views to be the way that Dart users design large amounts of new Dart code.  I think they are filling a fairly niche area.  As such, I would like to minimize the cognitive load on *consumers* of views.  If a user sees `static struct B6(A1 a) show A1.*`, they can... probably guess what it does?  But compare that to `view class B6(int x) extends A1` .  Sure, they need to learn what primary constructors are (but they learn that for data classes too).  But other than that, all of their existing intuitions carry over.  If Foo extends Bar, then Foo is a subtype of Bar, so B6 is a subtype of A1.  check.  If Foo extends Bar, then Foo has all of the methods of Bar.  check.  Now, of course, there are differences that will show through.  But the idea behind this design is to see how much we can leverage existing behavior to make the view feature be as small of a cognitive load on Dart users as possible.  Now, maybe that doesn't work out, in which case maybe we take some pieces of this and some pieces of the other proposal (e.g. your hybrid primary constructor + show hide examples above)?

> Re: need for overrides. I believe we do,

Ok.

> Re: cross-library extension. We believe we need it too. 

Ok.   My main concern with this is harmonizing it with regular structs, but that's tractable. Maybe this pushes in the direction of splitting apart structs and extension structs into data classes and view classes, so there is less of an imperative to make them consistent on things like this.

It's also not clear that making regular structs sealed is going to fly.  :)

working/extension_structs/overview.md
leafpetersen: > I'd have to imagine it's some synthetic class that's created for that view that can be implemented (like Siggi mentions), but that seems heavy and ugly.

Sorry, I still don't understand how this works.  views use static dispatch.  I can perfectly well allow you to say that `Foo` implements `MyStruct`, but when I assign a `Foo` to a `MyStruct`, the methods that get called will not be the `Foo` methods, but the `MyStruct` methods.  That's what static dispatch means.

> I'm guessing there was supposed to be more to this comment from the last sentence.

Sorry, I meant to delete that last bit.  But to close it, we could do something like:

```
// Magic marker typedef
typedef JSObjectImplements<T> = T;
struct Window(JSObject o) implements JSObjectImplements<List<int>> {}
```

or really any number of other things.  You could go the other way around and do:

```
typedef JSObjectImplements<T> = JSObject
struct Window(JSObjectImplements<List<int>> o) implements List<int> {}
```
though this one doesn't generalize well to the n-ary case.

There are other things we could brainstorm.

sigmundch: >> We don't want MyFooMock implements Foo

> _Why_ don't we want this, again? I'd expect the class to conform to every member signature (and if needed, the singular field's signature as well), just like with classes, but I feel like I'm missing something.

My point was mainly to make a distinction that static struct/views do not have a virtual interface. I fear that writing `MyFooMock implements Foo` can lead to confusion because developers will expect an assigment like `Foo x = MyFooMock()` to work. I'd be more comfortable if Foo's implied interface is distinguished from the type name syntactically somehow. For example, I'd be OK with `MyFooMock implement Foo.interface` or `MyFooMock implements ImpliedInterface<Foo>` (that's what I meant by `ExtensionStructureInterface<Foo>` in the earlier example)

working/extension_structs/overview.md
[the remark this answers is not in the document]
sigmundch: > ... So, this proposal is very specifically trying to avoid doing what you are proposing here. It's almost the heart of the design principle here. 

Totally makes sense - I was trying to balance that with the possible cognitive load of having overlapping concepts have different meanings 🙂 

As I think through these extension examples and how do we craft these features in terms of other simpler ones, one more thought comes to mind. Any thoughts about having a general notion of statically dispatched non-virtual methods in classes? We have that with extension methods, but what if they we allow them generally?

I am visualizing the complement of "virtual" in other languages:
```dart
class A1 {
  int m1() => 1;
  nonvirtual int m2() => 2;
}
```
We could consider extends also exposing nonvirtual methods (just like extension methods do):
```dart
class A2 extends A1 {
   // inherits both m1 and m2, but m2 is dispatched statically
}
```

And we could consider allowing overrides or not:

```dart
class A2 extends A1 {
   nonvirtual int m2() => 3; // TBD if we allow it
}

// If we do, nonvirtual methods are dispatched statically:
// (A2() as A1).m2() returns 2
```

I know we are trying to keep the new feature as small as possible, but can't help to connect how all these building blocks are themselves useful features independently. One reason that makes this specific feature attractive is that now static views are even less special. Extension is the same as in regular classes, nothing new here. That is, `static views` then are regular classes that only have a single field, only declare and inherit nonvirtual methods, don't expose an interface (seals them to some extend), etc.

Look forward to your thoughs!

working/extension_structs/overview.md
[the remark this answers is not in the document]
eernstg: Got it!

working/extension_structs/overview.md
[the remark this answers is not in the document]
leafpetersen: It's an interesting idea.  We've discussed non virtual methods before.  C# has what you describe, basically: methods which are dispatched based on type (so even if you "override" them in a subclass, you get the superclass method any time you access via the superclass type).  I agree that it would help a bit with the consistency story here, though possibly only be extending the cognitive load universally... :)

One tricky corner for us would be generics.  For views/extension types, we definitely have the semantics that generic parameters get filled in based on the static type.  We could do the same for non-virtual methods on classes, but it's not a forced move, and in some ways feels off.  Example:

```
class A<T> {
  nonvirtual void show() { print(T); };
}

void main() {
  A<num> a = A<int>();
  a.show();
}
```

Does this print `num`, or `int`?  The extension method semantics give you `num`.  Essentially, we're just giving you a short syntax for defining an extension method on `A`.  But if I'm just thinking of these as ordinary (but non-virtual) methods, I'd kind of expect to see `int`.

working/extension_structs/overview.md
srujzs: > I can perfectly well allow you to say that Foo implements MyStruct, but when I assign a Foo to a MyStruct, the methods that get called will not be the Foo methods, but the MyStruct methods.

Right, I guess the problem here is assignability as Siggi mentions. I guess `implements` is the wrong way to think about this, as we only what the signature conformance side of that feature, not assignability or subtyping.

And okay, I agree that explicitly mentioning `.interface` or something similar would avoid the confusion where users think they could subtype.

I do not see any plausible usage for this besides specifically for interop mocking (which we can check anyways, but just at a later stage unfortunately, unless we can maybe leverage the analyzer), so I'm inclined to agree that this is just a small very specific nice-to-have that I'm okay with dropping. After all, why would anyone want to make sure the signatures match if they never plan on using one type for the other? :)

leafpetersen: > unless we can maybe leverage the analyzer

Possibly a candidate for a lint.

working/extension_structs/overview.md
leafpetersen: Filed https://github.com/dart-lang/language/issues/2362 to discuss immutability.

working/extension_structs/overview.md
leafpetersen: Filed https://github.com/dart-lang/language/issues/2363 to discuss declaration syntax.

working/extension_structs/overview.md
leafpetersen: Filed https://github.com/dart-lang/language/issues/2364 to discuss adding general primary constructors.

working/extension_structs/overview.md
```diff
@@ -0,0 +1,1067 @@
...
+and provides a number of convenient methods.
+
+```dart
+void test() {
+  // Call the default constructor, using the default value for w
+  var c1 = Component(0, 0, 0);
+  // Call the default constructor, explicitly passing w
+  var c2 = Component(1, 1, 1, 1);
+
+  assert(c1 == c2); // Equality is defined to be structural
```
natebosch, line 103: Should this be `Component(0, 0, 0, 1)` for a structural equality to match `Component(0, 0, 0)`?

working/extension_structs/overview.md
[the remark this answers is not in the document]
leafpetersen: Filed https://github.com/dart-lang/language/issues/2365 to discuss named vs positional.

working/extension_structs/overview.md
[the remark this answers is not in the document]
leafpetersen: I filed #2368, #2369, and #2370 to discuss this.  Let's continue the discussion there.  Feel free to copy over anything you feel important to capture.

working/extension_structs/overview.md
leafpetersen: Filed #2363 to discuss syntax.

working/extension_structs/overview.md
leafpetersen: Filed https://github.com/dart-lang/language/issues/2371 to discuss this.

working/extension_structs/overview.md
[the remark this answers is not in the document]
leafpetersen: Filed https://github.com/dart-lang/language/issues/2364 to track the idea of generalizing primary constructors, let's take the discussion there.  If I've missed anything else in here, please file additional issues.

working/extension_structs/overview.md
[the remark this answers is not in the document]
leafpetersen: Filed #2363 to track the syntax discussion.

working/extension_structs/overview.md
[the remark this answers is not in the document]
leafpetersen: Filed #2362 to discuss immutability, and #2364 to discuss primary constructors.  Let's continue discussion there.

working/extension_structs/overview.md
leafpetersen: Filed https://github.com/dart-lang/language/issues/2372 to discuss generated methods.

working/extension_structs/overview.md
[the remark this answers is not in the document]
leafpetersen: Filed #2372 to track generated methods.

leafpetersen:
I've filed https://github.com/dart-lang/language/issues/2360 to track this general proposal, with linked sub-issues.  Let's continue discussion there and on the sub-issues.  Thanks for all the feedback!

PR #2411, opened 2022-08-18
[patterns] Revise record identical() to be faster.

### LINKED ISSUE #2390: Records: What are the semantics of identity?
The records proposal specifies identity [here](https://github.com/dart-lang/language/blob/master/working/0546-patterns/records-feature-specification.md#identity) to be structural.  In particular, identity is *required* to return `true` for any two records with the same shape for which the fields are pointwise identical.

We have discussed an alternative specification which would look more like the following:

> The identity operator on records *may* return true for records such that: 
>    - Both have the same shape.
>    - For every pair of corresponding fields, identical could validly return true on that pair.
> The identity operator on records may always return false.

In either of the two formulations compilers required neither to preserve pointer identity, nor to preserve pointer dis-identity.   That is, a compiler may unbox and re-allocate a record freely, or it may replace two source level records with identical components with a single allocation.

However, in the second formulation, compilers are free to implement identity as pointer identity, rather than requiring a slow recursive traversal.  This also has implications even for non-record based code, since all identity checks for which the compiler cannot prove that a record reaches neither argument must now include conditional logic to handle the case of records.

Should we change this specification?

cc @munificent @eernstg @lrhn @jakemac53 @natebosch @stereotype441 @rakudrama @mraleph @alexmarkov

munificent:
I think this matches the semantics in the struct proposal.

Fix #2390.

### DIFF
--- a/working/0546-patterns/records-feature-specification.md
+++ b/working/0546-patterns/records-feature-specification.md
@@ -340,8 +340,8 @@ offer loose guarantees. Calling `identical()` with a record argument returns:
 *   `false`, if the other argument is not a record.
 *   `false`, if the records do not have the same shape. *Since named field
     order is not part of a record's shape, this implies that named field order
-    does not affect identity either. `(a: 1, b: 2)` and `(b: 2, a: 1)` are
-    identical.*
+    does not affect identity either. `identical((a: 1, b: 2), (b: 2, a: 1))` is
+    not required to return false.*
 *   `false`, if any pair of corresponding fields are not identical.
 *   Otherwise it *may* return `true`, but is not required to.
 
--- a/working/0546-patterns/records-feature-specification.md
+++ b/working/0546-patterns/records-feature-specification.md
@@ -322,13 +322,20 @@ through calls to `identical()`, then optimizing away the creation of these
 objects is harder.
 
 Semantically, we do not want records to have unique identities distinct from
-their contents. A record *is* its contents in the same way that every value 3
-in a program is the "same" 3 whether it came from the number literal `3` or the
-result of `1 + 2`.
+their contents. A record *is* its contents in the same way that every value 3 in
+a program is the "same" 3 whether it came from the number literal `3` or the
+result of `1 + 2`. This is why `==` for records is defined in terms of their
+shape and fields. Two records with the same shape and equal fields are equal
+values.
 
-This is why `==` for records is defined in terms of their shape and fields. Two
-records with the same shape and fields are equivalent. Identity follows similar
-rules. Calling `identical()` with a record argument returns:
+At the same time, we want `identical()` to be fast because one of its primary
+uses is as a fast-path check for equality. An `identical()` that is obliged to
+iterate over the record's fields (transitively in the case where some fields
+are themselves records) might nullify the benefits of using `identical()` as a
+fast-path check before calling `==`.
+
+To balance those opposing goals, `identical()` on records is defined to only
+offer loose guarantees. Calling `identical()` with a record argument returns:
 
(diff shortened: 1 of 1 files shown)


working/0546-patterns/records-feature-specification.md
```diff
@@ -322,38 +322,39 @@ through calls to `identical()`, then optimizing away the creation of these
...
+
+To balance those opposing goals, `identical()` on records is defined to only
+offer loose guarantees. Calling `identical()` with a record argument returns:
 
 *   `false`, if the other argument is not a record.
 *   `false`, if the records do not have the same shape. *Since named field
     order is not part of a record's shape, this implies that named field order
     does not affect identity either. `(a: 1, b: 2)` and `(b: 2, a: 1)` are
     identical.*
 *   `false`, if any pair of corresponding fields are not identical.
```
lrhn, line 345: Is it worth saying "recursively identical" here, or otherwise allude to the fields being able to be records too?

working/0546-patterns/records-feature-specification.md
```diff
@@ -322,38 +322,39 @@ through calls to `identical()`, then optimizing away the creation of these
...
+*If an implementation can easily determine that two record arguments to
+`identical()` have the same shape and identical fields, then it should return
+`true`. Typically, this is because the two arguments to `identical()` are
+pointers with the same address to the same heap-allocated record object. But if
+an implementation would have to do a slower field-wise comparison to determine
+identity, it's probably better to return `false` quickly.*
+
+*In other words, if `identical()` returns `true`, then the records are
+definitely indistinguishable. But if it returns `false`, they may or may not
+be.*
```
lrhn, line 357: Maybe elaborate and say something like:

> Unlike object references, where being identical is deterministic and must be reported consistent over time, records being recognized as identical is an ephemeral occurrence that can change at any time. However, once two records have been recognized as identical, it's safe to keep treating them as such, since they will keep being indistinguishable whether or not `identical` will recognize it.

lrhn (approved):

working/0546-patterns/records-feature-specification.md
```diff
@@ -322,38 +322,39 @@ through calls to `identical()`, then optimizing away the creation of these
...
+fast-path check before calling `==`.
+
+To balance those opposing goals, `identical()` on records is defined to only
+offer loose guarantees. Calling `identical()` with a record argument returns:
 
 *   `false`, if the other argument is not a record.
 *   `false`, if the records do not have the same shape. *Since named field
     order is not part of a record's shape, this implies that named field order
     does not affect identity either. `(a: 1, b: 2)` and `(b: 2, a: 1)` are
     identical.*
```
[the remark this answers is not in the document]
leafpetersen: Not sure if it's necessary to do for this feature spec, but I agree that separating the two concepts might be a cleaner way to specify this, perhaps in the formal spec.  Though I would avoid confusion by choosing a new term: e.g. define a notion of "same object", and define the identical function in terms of "same object".  cc @eernstg

leafpetersen (approved):
LGTM

working/0546-patterns/records-feature-specification.md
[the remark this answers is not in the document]
munificent: I like the idea of defining "same object" as the optimal identical behavior and `identical()` as a fast-path version of that that may report false negatives. But I don't want to define that here since it should take into account objects of other types.

Reworded the non-normative text to not say "are identical".

working/0546-patterns/records-feature-specification.md
munificent: That's true, but I feel like it's maybe belaboring the point.

Issue #2219, opened 2022-04-28
Introduce additional primitive getters


eernstg:
Cf. https://github.com/dart-lang/language/issues/1296#issuecomment-1109428611, https://github.com/dart-lang/language/issues/1296#issuecomment-1110075607. Issue #1296 contains several other comments that are relevant to this discussion, but the topic is separate from that of #1296, so here is an issue where the addition of primitive getters is the main topic.

The language specification currently specifies what it means to have a _primitive_ operator `==`. Essentially, this means that the implementation is guaranteed to be provided by the system, which again makes it possible to guarantee that the corresponding behavior can also be provided during constant expression evaluation at compile time.

We could introduce the corresponding notion of being primitive for any getter whose implementation is guaranteed to be system provided, thus extending the set of constant expressions with things like `myList.length`, `mySet.isNotEmpty`, etc.

Here is a table showing the rather large number of cases where we could do this (taken from [this comment by @lrhn](https://github.com/dart-lang/language/issues/1296#issuecomment-1110075607)):

> The following instance member invocations are allowed in potentially constant and compile-time constant expressions,
> where the receiver and arguments must be compile-time constant expressions in order for the invocation to be so,
> and potentially constant or compile-time constant expressions for the invocation to be potentially constant.
> The invocation is not constant if the static type of the receiver expression is not assignable to one of the 
> required receiver static types, and it's a compile-time error if the run-time type of the receiver constant value
> is not an instance of one of the required receiver runtime types.
>
> | Member | Receiver static types | Receiver runtime types | 
> | ----------- | --------------------------- | ------------------------------ |
> | `+`, `*`    | `String`, `num`           | `String`, `int`, `double`  |
> | `-`, `%`, `/`, `remainder`, `~/`, `unary-` | `num` | `int`, `double` |
> | `\|`, `&`, `^` | `int`, `bool` | `int`, `bool` |
> | `~`, `>>`, `>>>`, `<<` | `int` | `int` |
> | `<`, `<=`, `>`, `>=` | `num` | `int`, `double` |
> | `==` | `Object?` | any type with primitive equality |
> | `length` | `String` | `String` |

We can then just add more entries to the table if we want to, without having to rewrite the entire "constants" section.

> | Member | Receiver static types | Receiver runtime types | 
> | ----------- | --------------------------- | ------------------------------ |
> | `length`, `isEmpty`, `isNotEmpty`, <br> `first`, `last`, `single`, `contains` | `List` | value of a constant `List` literal |
[...]

PR #2413, opened 2022-08-18
Add section on records being constant.

### LINKED ISSUE #2337: Can records be const?
The records spec does not currently use the word 'const' or 'constant'. Can a record be const? Something like

```dart
const x = 'one';
const y = 'two';

const r = (x, y);

// Separate, for patterns, which has an open TODO on const evaluation of destructuring.
const (d, e) = r;
```

lrhn:
A proposal for constant specifying records being constant.
Hope I covered all the corners. 

Basically, only two paragraphs are normative, the rest is elaboration.

Fixes #2337 

@mit-mit

### DIFF
--- a/working/0546-patterns/records-feature-specification.md
+++ b/working/0546-patterns/records-feature-specification.md
@@ -285,8 +285,10 @@ fields are) and collection literals.
 
 ### Constants
 
+_Record expressions can be constant and potentially constant expressions._
+
 A record expression is a compile-time constant expression
-if all its record field expressions are compile-time constant expressions. 
+if and only if all its record field expressions are compile-time constant expressions. 
 
 _This is true whether the expression occurs in a constant context or not,
 which means that a record expression can be used directly as a parameter default value 
@@ -294,22 +296,23 @@ if its record field expressions are constant expressions.
 Example: `f({(int, int) x = (1, 2)}) => ...`._
 
 A record expression is a potentially constant expression 
-if all its record field expressions are potentially constant or constant expressions.
+if and only iff all its record field expressions are potentially constant or constant expressions.
 
 _This means that a record expression can be used in the initializer list
-of a constant non-redirecting generative constructor._
+of a constant non-redirecting generative constructor, 
+and can depend on constructor parameters._
 
-Constant *object* instantiations create deeply immutable and canonicalied objects.
+_Constant *object* instantiations create deeply immutable and canonicalied objects.
 Records are always unmodifiable, and if their field values are deeply immutable,
 like constants values, the records are also deeply immutable.
 It's meaningless to consider whether record constants are canonicalized,
-since records do not have a persistent identity.
+since records do not have a persistent identity._
 
-Because of that, there is no need for a `const (1, 2)` syntax to force a record 
+_Because of that, there is no need for a `const (1, 2)` syntax to force a record 
 to be a constant, like there is for object creation expressions. 
 A record expression with field values that are constant-created values, 
 will be indistinguishable from a similar expression created in a constant 
-context, since identity cannot be used as a distinguishing trait.
+context, since identity cannot be used as a distinguishing trait._
 
 _(We could choose to promise that a compile-time constant `identical(c1, c2)`,
 where the expression occurs in a constant context and `c1` and `c2` are records, 
@@ -317,6 +320,39 @@ will evaluate to `true` iff a runtime evaluation of `identical`
 *can* return `true` for the same values. 
 That is, records would be canonicalized during compile-time constant evealuation,
 but may lose their identity at runtime. We will not make such a promise.)_
+
+For canonoicalization purposes, we update the definition of when to canonicalize
+the result of a constant object creation expression to not be dependent on 
+the `identical` function, since it does not behave predictably (or usefully)
+for records.
+
+We define two Dart values, *a* and *b*, to be _structurally equivalent_ as follows:
+* If *a* and *b* are both records, and they have the same shape, 
+  and for each field *f* of that shape, the records' values of that field, 
+  *a*<sub>*f*</sub> and *b*<sub>*f*</sub> are structurally equivalent, 
+  then *a* and *b* are structurally equivalent.
+* If *a* and *b* are non-record object references, 
+  and they refer to the same object, then *a* and *b* are structurally equivalent.
+  _So structural equivalence agrees with `identical` for non-records._
+* Otherwise *a* and *b* are not structurally equivalent.
+
+With that definition, the rules for object and collection canonicalization is changed
+from requiring that instance variable, list/set element and map key/value values are
+`identical` between the instances, to them being _structurally equivalent_.
+
+_This change allows a class like_
+```dart
+class C {
+  final (int, int) pair;
+  const C(int x, int y) : pair = (x, y);
+}
+```
+_to be properly canonicalized for objects with the same effective state, 
+independentlty of whether `identical` returns `true` or `false` on the `pair` value._
+
+_Notice that if the `identical`returns `true` on two records, they must be structurally equivalent,
+but unlike for non-records, the `identical` function can also return `false`
+for structurally equivalent records._
 
 ## Runtime semantics
 
--- a/working/0546-patterns/records-feature-specification.md
+++ b/working/0546-patterns/records-feature-specification.md
@@ -283,6 +283,41 @@ fields are) and collection literals.
 
 **TODO: Specify this more precisely.**
 
+### Constants
+
+A record expression is a compile-time constant expression
+if all its record field expressions are compile-time constant expressions. 
+
+_This is true whether the expression occurs in a constant context or not,
+which means that a record expression can be used directly as a parameter default value 
+if its record field expressions are constant expressions.
+Example: `f({(int, int) x = (1, 2)}) => ...`._
+
+A record expression is a potentially constant expression 
+if all its record field expressions are potentially constant or constant expressions.
+
+_This means that a record expression can be used in the initializer list
+of a constant non-redirecting generative constructor._
+
+Constant *object* instantiations create deeply immutable and canonicalied objects.
+Records are always unmodifiable, and if their field values are deeply immutable,
+like constants values, the records are also deeply immutable.
+It's meaningless to consider whether record constants are canonicalized,
+since records do not have a persistent identity.
+
+Because of that, there is no need for a `const (1, 2)` syntax to force a record 
+to be a constant, like there is for object creation expressions. 
+A record expression with field values that are constant-created values, 
+will be indistinguishable from a similar expression created in a constant 
+context, since identity cannot be used as a distinguishing trait.
+
+_(We could choose to promise that a compile-time constant `identical(c1, c2)`,
+where the expression occurs in a constant context and `c1` and `c2` are records, 
+will evaluate to `true` iff a runtime evaluation of `identical` 
+*can* return `true` for the same values. 
+That is, records would be canonicalized during compile-time constant evealuation,
+but may lose their identity at runtime. We will not make such a promise.)_
+
 ## Runtime semantics
 
 ### Records


working/0546-patterns/records-feature-specification.md
```diff
@@ -283,6 +283,41 @@ fields are) and collection literals.
...
+if all its record field expressions are potentially constant or constant expressions.
+
+_This means that a record expression can be used in the initializer list
+of a constant non-redirecting generative constructor._
+
+Constant *object* instantiations create deeply immutable and canonicalied objects.
+Records are always unmodifiable, and if their field values are deeply immutable,
+like constants values, the records are also deeply immutable.
+It's meaningless to consider whether record constants are canonicalized,
+since records do not have a persistent identity.
```
eernstg, line 306: The purpose of having explicitly constant records and specifying canonicalization would be that (1) this clarifies the intent, and makes it an error if some components of a `const` record literal are not constant expressions, and (2) this serves as a request for physical canonicalization. Granted, if there is a location in code where this kind of record is unboxed then we would _not_ promise to find the canonical object if and when it is re-boxed. Still, I'd expect the explicit request for canonicalization to be helpful, in order to make `identical` work as a fast path for testing equality.

working/0546-patterns/records-feature-specification.md
```diff
@@ -283,6 +283,41 @@ fields are) and collection literals.
...
+
+Because of that, there is no need for a `const (1, 2)` syntax to force a record 
+to be a constant, like there is for object creation expressions. 
+A record expression with field values that are constant-created values, 
+will be indistinguishable from a similar expression created in a constant 
+context, since identity cannot be used as a distinguishing trait.
```
eernstg, line 312: Identity cannot be used as a _negatively_ distinguishing trait, that is, we can't conclude non-sameness from `identical(r1, r2) == false`.

However, this doesn't imply that `identical(r1, r2) == true` is useless, and we might be able to use explicitly constant records to ensure that this occurs more frequently.

working/0546-patterns/records-feature-specification.md
```diff
@@ -283,6 +283,41 @@ fields are) and collection literals.
...
+
+_(We could choose to promise that a compile-time constant `identical(c1, c2)`,
+where the expression occurs in a constant context and `c1` and `c2` are records, 
+will evaluate to `true` iff a runtime evaluation of `identical` 
+*can* return `true` for the same values. 
+That is, records would be canonicalized during compile-time constant evealuation,
+but may lose their identity at runtime. We will not make such a promise.)_
```
eernstg, line 319: I think such a promise would be too big, anyway. We would not _prevent_ unboxing from occurring on a canonicalized record, we would just allow canonicalization of certain record values to improve performance whenever possible.

```dart
const r0 = (1, 2);

void f((int, int) r) {
  ... // Uses `r.$0` and `r.$1`.
}
// A customized variant of `f` is generated by the compiler:
void f_unboxed(int r1, int r2) {
  ... // Copy of body of `f`, using `r1` and `r2`.
}

bool b = true;

void main() {
  var r1 = (Datetime.now().millisecondsSinceEpoch, 2);
  var r = b ? r0 : r1;
  f(r); // May compile to `f_unboxed(r.$0, r.$1)`.
}
```

eernstg (commented):
For now, I've just commented on the situation in this review, but looking at https://github.com/dart-lang/language/issues/2337, I think the matter could benefit from some further discussion in that issue (or in some other issue, if needed).

working/0546-patterns/records-feature-specification.md
```diff
@@ -283,6 +283,41 @@ fields are) and collection literals.
 
 **TODO: Specify this more precisely.**
 
+### Constants
+
+A record expression is a compile-time constant expression
+if all its record field expressions are compile-time constant expressions.
```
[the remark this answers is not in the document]
lrhn: True. Nothing is a compile-time constant unless the specification says so, so I'm being inductive here.

working/0546-patterns/records-feature-specification.md
```diff
@@ -283,6 +283,41 @@ fields are) and collection literals.
...
+if its record field expressions are constant expressions.
+Example: `f({(int, int) x = (1, 2)}) => ...`._
+
+A record expression is a potentially constant expression 
+if all its record field expressions are potentially constant or constant expressions.
+
+_This means that a record expression can be used in the initializer list
+of a constant non-redirecting generative constructor._
+
+Constant *object* instantiations create deeply immutable and canonicalied objects.
```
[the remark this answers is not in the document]
lrhn: Looking at the rest of the document, it doesn't seem to be as strongly separating rationale from normative text. 
I tried making everything except paragraph 1 and 3 italic, and it looked weird.

working/0546-patterns/records-feature-specification.md
```diff
@@ -283,6 +283,41 @@ fields are) and collection literals.
...
+_This is true whether the expression occurs in a constant context or not,
+which means that a record expression can be used directly as a parameter default value 
+if its record field expressions are constant expressions.
+Example: `f({(int, int) x = (1, 2)}) => ...`._
+
+A record expression is a potentially constant expression 
+if all its record field expressions are potentially constant or constant expressions.
+
+_This means that a record expression can be used in the initializer list
+of a constant non-redirecting generative constructor._
```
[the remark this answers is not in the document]
lrhn: > It means that we would allow declarations like `const C(int i): this.myRecord = (1, i);`.

That's actually a consequence of the "potentially constant" paragraph below, not this one. Its true that we'd allow it.

We should definitely not allow `const [1, i]` (with or without the leading `const`). We use different implementation classes with different behavior for constant and non-constant lists.
A `C(1, i)` (class constant constructor invocation with constant/potentially constant arguments) is more likely, because that uses the same class for constant and non-constant invocations, but because constructors can execute (some) code, we can get into problems with recursive constructor calls. And it differs wrt. canonicalization, which records don't (at least you can't tell if they do when a single record can be non-identical to itself).

Records are special. They may or may not correspond to an object at all, and you can't tell if they do.
A record like `(c1, c2)` really is just the juxtaposition of its values. It's just two values, nothing more or less. If each value can occur, so can both.

A list has structure on top of its contents. We use different classes for const lists and non-const lists. Creating a list is an object creation expression.

Not so for records. Example:
```dart
class C {
  final int a, b;
  final (int, int) p1;
  const C(int a, int b, int c, int d) : a = a, b = b, p = (c, d);
  (int, int) get p2 => (a, b);
}
```
There is no fundamental difference between the initializers for `a`, `b`, `p1.$0` and `p1.$1`, and I don't want there to be one. The only difference is the *grouping* of the values, where `p.$0` and `p.$1` can be accessed together as `p1`, just as `a` and `b` can be accessed through `p2`. The `p1` variable still just (potentially) two adjacent storage locations in the class.

We *can* add `const` in front of a record, but it feels spurious. It doesn't cause canonicalization, because records don't have a consistent identity anyway. It ensures that the field values are constants, but either that doesn't matter, it's enforced anyway, or you can put `const` on each field expression if necessary, like `({x = (const C(), const [])})`.
(I'd be fine with allowing `const expression` for any expression, which then puts that expression into a constant context. The `const (e1, e2)` would just be shorthand for `(const e1, const e2)`, and `const 4` is valid and constant.


I don't know if we intend to do the same thing for structs. If we *can*, I'd probably be happy with it. A `StructName(arg1,...,argn)` would be constant (aka. is valid where a constant is required) if it invokes the primary constructor, and the argument expressions are constant.
However, since a struct can have non-primary constructors too, and those would presumably require `const` to invoke, it would be better for consistency, with other struct constructors and other const constructors in general, to require a `const` in front. So, "probably not" to doing the same for structs, or other constructs with constructors.

working/0546-patterns/records-feature-specification.md
lrhn: If you plan to depend on canonicalization, you're setting yourself up for failure.

I'd expect a significant percentage of record-typed instance variables to be inlined in the object structure. You lose canonicalization for those immediately at creation time. 

Asking for canonicalization is asking for something you should not expect getting, which is why I actually prefer *not* giving you the option.

working/0546-patterns/records-feature-specification.md
lrhn: I really, really hope people *won't* start trying to force canonicalization and depend on the identity of records for anything.
It's an optimization. If your code can't run well if `identical` returns `false` on all records, you should fix it. Or stop using records for something where identity matters.

working/0546-patterns/records-feature-specification.md
lrhn: It'd only be a guarantee during compilation. When we perform an `identical(v1, v2)` *as a constant expression*, we have (representations of) the values available, and we can choose to perform the slower structural comparions each time, while still using the fast-path `identical` at runtime.

You won't be calling `f_unboxed` at compile-time.

Now, bigger question, should `record.$0` or `record.name` be constant expressions if `record` is?

working/0546-patterns/records-feature-specification.md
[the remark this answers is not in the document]
eernstg: Ah, yes, it's encoded in the 'Constants' section where we say 'The .. constant expressions are the following', which implies that no additional expressions are constant. We do have `if{}f` in many other locations, but not in the section about constants.

working/0546-patterns/records-feature-specification.md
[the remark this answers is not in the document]
eernstg: I think that's OK. It can look weird in this context, but that's just because we ought to change the rest of this document similarly, and in the meantime there's nothing lost by doing it here.

working/0546-patterns/records-feature-specification.md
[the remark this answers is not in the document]
eernstg: > Records are special. They may or may not correspond to an object at all, and you can't tell if they do.

I can't see any way to support a reference of type `Object` whose value is assigned/initialized to a record, other than boxing (or using an existing boxed representation of) the record, unless that whole reference can be removed during optimization. The type could also be any top type, or it could be a type variable whose value is `Object` or a top type.

This means that records can be known to have a heap entity representation at some point. Any usage of that representation at the record type might unbox it, but the `Object` typed reference is not a weak reference and the fact that it is a reference to a heap entity makes it more object-like than the wording

> It's just two values, nothing more or less

implies.

> We can add const in front of a record, but it feels spurious. It doesn't cause canonicalization

It _could_ cause canonicalization.

> allowing `const expression` for any expression

Looks like that would be a thin layer of syntactic sugar: Putting `const freshVariableName = expression;` at the top level and then replacing `const expression` by `freshVariableName`.

working/0546-patterns/records-feature-specification.md
eernstg: > If you plan to depend on canonicalization, you're setting yourself up for failure.

That is true!

But it still clarifies the intent and allows for compile-time checking that we have a constant.

Of course, a top level `const freshVariableName = ...` would do that, too. So the question is just whether or not we want this syntactic sugar.

working/0546-patterns/records-feature-specification.md
eernstg: > I really, really hope people won't start trying to force canonicalization and
> depend on the identity of records for anything.

Right, we might very well want a lint against statically known records as arguments to `identical`, except for the use of exactly this approach in the implicitly induced implementation of `==` on record types.

But the point is that an explicit request for canonicalization could ensure that `==` runs faster, more frequently.

working/0546-patterns/records-feature-specification.md
eernstg: Part of https://github.com/dart-lang/language/issues/2219. ;-)

working/0546-patterns/records-feature-specification.md
lrhn: I'm strongly opposed to giving people the idea that they can affect canonicalization of records, and that it matters in any way.

You really cannot know that canonicalization will make anything run faster. It might. It's definitely true that more canonicalized objects gives more chances of a positive `identical` check.

But you also cannot know that *requesting* canonicalization actually will cause more canonicalization. It's perfectly reasonable for a compiler to canonicalize all the constant records expressions it can see anyway, and if the compiler chooses not to, perhaps it's for a good reason. And it still can choose not to, even if you request canonicalization.

If I was a compiler writer implementing records, I'd completely ignore any user request for canonicalization of records, because I can, and do whatever I think is best for the compilation.

So why give the user a button to press, if it's not connected to anything anyway?

working/0546-patterns/records-feature-specification.md
munificent: Yeah, I think canonicalization is a non-starter for records. The language has no formal notion of what it means to assign a value or bind it to a variable. So even if you *have* a canonicalized constant record, there's no guarantee that when you *mention* it that doing so doesn't copy it to a thing with a new identity. Even binding it to the parameter to the `identical()` function itself could cause the canonicalized record to be copied to a new stack-allocated storage and lose its original identity.

working/0546-patterns/records-feature-specification.md
munificent: Even if the compiler *does* canonicalize every single record, as soon as you *mention* it, the compiler could silently copy it to a new unboxed storage and you've lost the fact that it once referred to some canonical instance.

working/0546-patterns/records-feature-specification.md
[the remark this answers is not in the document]
munificent: I try to use italics for non-normative text when it's in the middle of otherwise clearly normative sections, mostly as a signal to implementers that they don't have to treat it as gospel. But in introductory sections that are higher level or motivational, almost the whole thing is non-normative, so I don't bother highlighting it.

I admit I could probably be clearer and more disciplined about this. :)

working/0546-patterns/records-feature-specification.md
```diff
@@ -283,6 +283,41 @@ fields are) and collection literals.
...
+
+Constant *object* instantiations create deeply immutable and canonicalied objects.
+Records are always unmodifiable, and if their field values are deeply immutable,
+like constants values, the records are also deeply immutable.
+It's meaningless to consider whether record constants are canonicalized,
+since records do not have a persistent identity.
+
+Because of that, there is no need for a `const (1, 2)` syntax to force a record 
+to be a constant, like there is for object creation expressions.
```
munificent, line 309: This is a really good insight! I like the idea of records being automagically const. They're immutable and have no persistent identity, so there's literally nothing preventing a compiler from treating any record expression with compile-time-known field values as a constant. And if that's the case, I see no compelling reason to require users to write `const` before it.

On the other hand, auto-const for records might lead users to want other immutable classes to be auto-constable too, and then get annoyed that they aren't. Perhaps something to think about with structs.

munificent (approved):
I really like this, better than the naive approach I took in #2422. If you want to merge it under me, I'm fine with it. I'll rebase my PR to handle it.

working/0546-patterns/records-feature-specification.md
lrhn: True. 
My perspective is that where `const Foo(..)` creates a new `Foo` object, `(a, b)` does not create anything. It just places `a` next to `b`. It's just *two values* instead of *one value*.
It's not that you create a constant pair, it's that a pair of constant values are ... well, two constant values. They can be used wherever two constant values are needed.
You never need "a constant pair", you need a pair of constant values.


We can allow an initializer list of the form:
```dart
  const C((int, int) pair) : (this.x, this.y) = pair;
```
(We should!)

We could even allow 
```dart
  (final x, const y) = (1, 2);
```
(We probably shouldn't.)

munificent (approved):

lrhn:
Landed. Your merge :)

working/0546-patterns/records-feature-specification.md
eernstg: I don't think there is anything particularly tricky about

```dart
class C {
  final int x, y;
  const C((int, int) pair) : (x, y) = pair;
}
```

All we need is the ability to desugar the initializer list to `x = pair.$0, y = pair.$1`. But the opposite direction is more tricky:

```dart
class D {
  final (int, int) pair;
  const D(int x, int y) : pair = (x, y);
}
```

And even more:

```dart
class E {
  final Object o;
  const E(int x, int y, bool b) : o = b ? (x, y) : x + y;
}
```

Canonicalization is currently based on a compile-time version of `identical` applied to the values of each instance variable, but `identical` applied to a pair of records isn't guaranteed to _ever_ return true, so we wouldn't have any guarantees about canonicalization of instances of `E`, unless we specify that this compile-time-`identical` behaves differently.

Perhaps it is enough to say that `identical` at compile-time will _always_ perform a complete traversal in case it is invoked on two record values (of course, bailing out early if there is a non-match, but traversing the entire structure in order to obtain the value `true`).

PR #2422, opened 2022-08-19
Address a bunch of records issues.

### LINKED ISSUE #2337: Can records be const?

### LINKED ISSUE #2386: Zero length and unary records
The records proposal forbids empty records and records with a single entry:

> A record may have only positional fields or only named fields, but cannot be totally empty. There is no "unit type". A record with no named fields must have at least two positional fields. This prevents confusion around whether a single positional element record is equivalent to its underlying value, and avoids a syntactic ambiguity with parenthesized expressions.

There has been discussion in the past of trying to exploit the symmetry between parameter lists and records.  I'm slightly worried that this restriction may come back to bite us, since it prevents (e.g.) uniformly reifying argument lists as records.  Is it possible to make this a restriction only on the *literal syntax*?  That is, semantically we still have zero length and unary records, we just have no literal syntax for them?

As a second step (or an alternative step) we could now or in the future add alternative syntax that generalized fully.  For example, I can imagine:
  - Adding a static constant `Record.unit` on `Record` which is the unique zero length record.
  - Adding a unary constructor `Record.single` on `Record` which produces a unary tuple
  - Or generalizing fully and say that there is a "magic" n-ary constructor on `Record` such that `Record(....)` produces the record corresponding to the literal record syntax `(...)`, with the additional generality that `...` may be empty or a single positional argument.

cc @munificent @lrhn @eernstg @natebosch @jakemac53 @stereotype441

### LINKED ISSUE #2388: Should there be an expression syntax for accessing positional record fields?
The current record proposal provides no notation for accessing a single positional field:

> Positional fields are not exposed as getters. Record patterns in pattern matching can be used to access a record's positional fields.

There are a couple of implications of this.  

First, I think this means that there's significant asymmetry between named and positional fields, in that you can access a named field directly on the object, but you can't do so for a positional field.  So you have a constant sized syntax for reading a single named field, but the only syntax for reading a positional field requires reading *all* of the fields (out into a pattern match) which is fairly verbose (even if you just use `_` for all of the fields you don't care about).  

Second, there's a semantic asymmetry in that you can presumably access named fields dynamically (and hence write code that is polymorphic over records with named fields) but not positional fields:

```dart
void printX(Record record) {
  print(record as dynamic.x);
}
```

Should we provide a way to read a single positional field?  e.g. `record.0` etc?

cc @munificent @lrhn @eernstg @stereotype441 @natebosch @jakemac53

### LINKED ISSUE #2389: What does `toString` on a record do?
The record proposal says:

```
The toString() method's behavior is unspecified.
```

This is all well and good, but we have to *implement* something.  So, what do we do here?

cc @munificent @eernstg @lrhn @jakemac53 @natebosch @stereotype441 @rakudrama @mraleph

### LINKED ISSUE #2406: Record types and try-catch-on
I've asked in the chat as well. Posting here to make it easier to find if needed later.

Currently

```dart
void foo() {
  try {
    ;
  } on Bar {
    ;
  }

  on(a, b, c) {
    ;
  }
  on(1, 2, 3);
}
```
is a try statement, then a local function declaration named `on` taking 3 parameters, then a call to that local function.

With record types this will instead parse as a try statement with a `on specificRecordType` clause, then a function call.

Any input?

/cc @lrhn @munificent @eernstg

munificent:
- Support constant records. Fix #2337.
- Support empty and one-positional-field records. Fix #2386.
- Re-add support for positional field getters Fix #2388.
- Specify the behavior of `toString()`. Fix #2389.
- Disambiguate record types in `on` clauses. Fix #2406.

cc @eernstg @stereotype441 @jakemac53 @natebosch @kallentu @mit-mit

### DIFF
--- a/working/0546-patterns/records-feature-specification.md
+++ b/working/0546-patterns/records-feature-specification.md
@@ -429,17 +429,23 @@ calling `toString()` or relying on it for end-user visible output.
 
 #### Equality
 
-The `==` method on record `r` with right operand `o` is defined as:
+Records have value equality, which means two records are equal if they have the
+same shape and the corresponding fields are equal. Since named field order is
+*not* part of a record's shape, that implies that the order of named fields
+does not affect equality:
 
-1.  If `o` is not a record with the same shape as `r` then return `false`.
+```dart
+var a = (x: 1, 2);
+var b = (2, x: 1);
+print(a == b); // true.
+```
 
-1.  For each pair of corresponding positional fields `rf` and `of` in position
-    order:
+More precisely, the `==` method on record `r` with right operand `o` is defined
+as:
 
-    1.  If `rf == of` is `false` then return `false`.
+1.  If `o` is not a record with the same shape as `r` then `false`.
 
-1.  For each pair of corresponding named fields `rf` and `of`, in unspecified
-    order:
+1.  For each pair of corresponding fields `rf` and `of` in unspecified order:
 
     1.  If `rf == of` is `false` then `false`.
 
@@ -447,15 +453,9 @@ The `==` method on record `r` with right operand `o` is defined as:
 
 *The order that fields are iterated is potentially user-visible since
 user-defined `==` methods can have side effects. Most well-behaved `==`
-implementations are pure. The order that named fields are visited is
-deliberately left unspecified so that implementations are free to canonicalize
-their order.*
-
-```dart
-var a = (x: 1, 2);
-var b = (2, x: 1);
-print(a == b); // true.
-```
+implementations are pure. The order that fields are visited is deliberately left
+unspecified so that implementations are free to reorder the field comparisons
+for performance.*
 
 The implementation of `hashCode` follows this. The hash code returned should
 depend on the field values such that two records that compare equal must have
--- a/working/0546-patterns/records-feature-specification.md
+++ b/working/0546-patterns/records-feature-specification.md
@@ -325,74 +325,74 @@ fields are) and collection literals.
 
 ### Constants
 
-_Record expressions can be constant and potentially constant expressions._
+Record expressions can be constant and potentially constant expressions. A
+record expression is a compile-time constant expression if and only if all its
+field expressions are compile-time constant expressions.
 
-A record expression is a compile-time constant expression
-if and only if all its record field expressions are compile-time constant expressions. 
+*This is true whether the expression occurs in a constant context or not, which
+means that a record expression can be used directly as a parameter default value
+if its record field expressions are constant expressions, as in:
 
-_This is true whether the expression occurs in a constant context or not,
-which means that a record expression can be used directly as a parameter default value 
-if its record field expressions are constant expressions.
-Example: `f({(int, int) x = (1, 2)}) => ...`._
+```dart
+void someFunction({(int, int) x = (1, 2)}) => ...`
+```
 
-A record expression is a potentially constant expression 
-if and only iff all its record field expressions are potentially constant or constant expressions.
+A record expression is a potentially constant expression if and only if all its
+field expressions are potentially constant or constant expressions. *This means
+that a record expression can be used in the initializer list of a constant
+non-redirecting generative constructor, and can depend on constructor
+parameters.*
 
-_This means that a record expression can be used in the initializer list
-of a constant non-redirecting generative constructor, 
-and can depend on constructor parameters._
+*Constant object instantiations (i.e. const constructor calls and const
+collection literals) create deeply immutable and canonicalized objects. Records
+are always unmodifiable. If a record's field values are also deeply immutable
+(which all constant values are), then the record is also deeply immutable. It's
+meaningless to consider whether record constants are canonicalized, since
+records do not have a persistent identity.*
 
-_Constant *object* instantiations create deeply immutable and canonicalied objects.
-Records are always unmodifiable, and if their field values are deeply immutable,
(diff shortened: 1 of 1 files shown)


munificent:
I just saw https://github.com/dart-lang/language/pull/2413, so I'll take a look at that and see if that makes more sense that my rudimentary const support in this PR.

working/0546-patterns/records-feature-specification.md
```diff
@@ -283,34 +323,88 @@ fields are) and collection literals.
...
 var a = (x: 1, 2);
 var b = (2, x: 1);
 print(a == b); // true.
 ```
 
-The implementation of `hashCode` follows this. Two records that are equal must
-have the same hash code.
+The implementation of `hashCode` follows this. The hash code returned should
+depend on the field values such that two records that compare equal must have
+the same hash code.
```
rakudrama, line 407: This implies that either (1) `hashCode` visits fields in some canonical order, or (2) the hash function is weakened so that it cannot distinguish permutations of named field values.

This is almost, but not quite, saying that the field order should be canonicalized. Perhaps we should go as far as to specify there is a canonicalization, and perhaps what it is.

working/0546-patterns/records-feature-specification.md
```diff
@@ -283,34 +323,88 @@ fields are) and collection literals.
...
-for equality by calling `==` on the corresponding field values in the same
-order that `==` was called on the records.
+The `==` method on record `r` with right operand `o` is defined as:
+
+1.  If `o` is not a record with the same shape as `r` then return `false`.
+
+1.  For each pair of corresponding positional fields `rf` and `of` in position
+    order:
+
+    1.  If `rf == of` is `false` then return `false`.
```
rakudrama, line 384: I would like a degree of freedom here that is not allowed.

Consider a record created where the static type is `(Uri, int)`.
This would allow there to be a specialized Record implementation that has a specialized `==` method that knows that `$0` and `$1` have types `Uri` and `int` respectively.

The specification requires `==` to be something like:

```
operator ==(other) =>
    other is (Object?, Object?) &&  // shape test
    this.$0 == other.$0 &&          // Uri.==
    this.$1 == other.$1;            // int.==
```

It would be beneficial to move the faster `int.==` forwards:

```
operator ==(other) =>
    other is (Object?, Object?) &&  // shape test
    this.$1 == other.$1 &&          // int.==
    this.$0 == other.$0;            // Uri.==
```

We can't tell that `Uri.==` is side-effect free - strictly speaking, it does since certain substructures are lazily initialized.
But we do know for sure `int.==` is side-effect free and super efficient.

Can the language be loosened to permit some comparisons to be moved forward provided they have no effects?

rakudrama: Simpler statement of my concern: this definition of record equality does not permit `r == r` to take a short-cut.

working/0546-patterns/records-feature-specification.md
lrhn: Not sure it implies that.
We do not allow equality between different record shapes, so as long as the order of fields used for the hashCode is consistent within the same shape, the order itself doesn't matter.
If two records are equal, the using `Object.hash(...)` on the field values in *any* order works, as long as it's the same order for all objects that has that shape. (This assumes that fields being equal implies their hash codes being equal to begin with.)

We do not promise that the `hashCode` is consistent between different executions of the same program, so precisely how we compute it shouldn't matter.

working/0546-patterns/records-feature-specification.md
lrhn: I'd be fine with allowing any order of comparisons. We're going to call *all* of the before answering `true` anyway.

We don't need to promise that we use the same ordering between calls, even on the same shape, because it doesn't matter for the result. And if your `==` has a side effect, ... you had it coming!

working/0546-patterns/records-feature-specification.md
munificent: Lasse has it right. It doesn't require hashing the fields in a canonical order across all record shapes, just that within a single shape, there is a consistent process for hashing them.

working/0546-patterns/records-feature-specification.md
munificent: I'm OK with leaving all field order unspecified. Done.

rakudrama: > I'd be fine with allowing any order of comparisons. We're going to call all of the `==`s before answering `true` anyway.

We often accelerate `==` in our own classes via `identical`:

    class Foo {
      bool operator==(other) => identical(this, other) || ... hard way ...;

This is not permitted for records since a field may have a side-effect-free but adversarial `==`.
It seems a shame to do structural equality for the very same allocated record, e.g. looking up an existing record in a Map.

lrhn: Nit-pickily, it's also not permitted/correct for non-record classes that contain `double`s that might be NaN.
We just choose to live with that.
(Records still should't.)

munificent:
I'm going to go ahead and land this because it has changes that the implementation teams will want to take into account. Please don't hesitate to file issues if there are pieces of this you don't agree with. I'm trying to land and iterate in order to make progress. :)

PR #2427, opened 2022-08-23
Introduce `ERASEINTERSECTION`

eernstg:
Cf. https://github.com/dart-lang/sdk/issues/49691.

The current definition of the **UP** function (computing the standard upper bound of two types) allows for the creation of a type of the form `(X & T)?`, even though such types should never exist.

This PR changes the definition of **UP** such that it will not return any such type. **DOWN** has been inspected as well, and it will not create any such type. The changed definition of **UP** uses a new type function, **ERASEINTERSECTION**, which is defined together with a handful of other helper functions in the null safety feature specification.

### DIFF
--- a/resources/type-system/upper-lower-bounds.md
+++ b/resources/type-system/upper-lower-bounds.md
@@ -52,6 +52,7 @@ which are specified in the
 **NULL**, which is true for types equivalent to `Null`;
 **MORETOP**, which is a total order on top and `Object` types; and
 **MOREBOTTOM**, which is an (almost) total order on bottom and `Null` types.
+**ERASEINTERSECTION**, which maps `X & T` to `X` and every other type to itself.
 
 ## Upper bounds
 
@@ -78,12 +79,12 @@ We define the upper bound of two types T1 and T2 to be **UP**(`T1`,`T2`) as foll
 - **UP**(`T1`, `T2`) where **NULL**(`T1`) =
   - `T2` if  `T2` is nullable
   - `T2*` if `Null <: T2` or `T1 <: Object` (that is, `T1` or `T2` is legacy)
-  - `T2?` otherwise
+  - `T3?` otherwise, where `T3` is **ERASEINTERSECTION**(`T2`).
 
 - **UP**(`T1`, `T2`) where **NULL**(`T2`) =
   - `T1` if  `T1` is nullable
   - `T1*` if `Null <: T1` or `T2 <: Object` (that is, `T1` or `T2` is legacy)
-  - `T1?` otherwise
+  - `T3?` otherwise, where `T3` is **ERASEINTERSECTION**(`T1`).
 
 - **UP**(`T1`, `T2`) where **OBJECT**(`T1`) and **OBJECT**(`T2`) =
   - `T1` if **MORETOP**(`T1`, `T2`)
@@ -92,12 +93,12 @@ We define the upper bound of two types T1 and T2 to be **UP**(`T1`,`T2`) as foll
 - **UP**(`T1`, `T2`) where **OBJECT**(`T1`) =
   - `T1` if `T2` is non-nullable
   - `T1*` if `Null <: T2` (that is, `T2` is legacy)
-  - `T1?` otherwise
+  - `T3?` otherwise, where `T3` is **ERASEINTERSECTION**(`T1`).
 
 - **UP**(`T1`, `T2`) where **OBJECT**(`T2`) =
   - `T2` if `T1` is non-nullable
   - `T2*` if `Null <: T1` (that is, `T1` is legacy)
-  - `T2?` otherwise
+  - `T3?` otherwise, where `T3` is **ERASEINTERSECTION**(`T2`).
 
 - **UP**(`T1*`, `T2*`) = `S*` where `S` is **UP**(`T1`, `T2`)
 - **UP**(`T1*`, `T2?`) = `S?` where `S` is **UP**(`T1`, `T2`)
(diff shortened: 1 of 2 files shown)


eernstg:
@leafpetersen, WDYT about this approach to https://github.com/dart-lang/sdk/issues/49691?

eernstg:
@leafpetersen, WDYT?

lrhn (approved):

eernstg:
Updated: The rules `UP(T1, T2)` where `OBJECT(T2)` and `UP(T1, T2)` where `OBJECT(T2)` were changed: There is no need to erase an intersection from a type which is known to be `OBJECT(_)`, because they are never intersections. Also, `ERASEINTERSECTION` is now used in the rules for `UP(T1?, T2)` and `UP(T1, T2?)`, because it is possible that `T2` respectively `T1` is an intersection type.

Note that this is one possible approach to handle the issue raised in https://github.com/dart-lang/sdk/issues/49691. Another one is presented in https://github.com/dart-lang/language/pull/3435. 

This proposal is the least disruptive, it simply reduces some types of the form `(X & B)?` to `X?`, and those types probably occur very rarely in the first place.

In contrast, #3435, changes the order of rules in `UP`, which may cause `UP` to yield different results in a more complex and perhaps more frequent manner.

eernstg:
Here's an example that shows a core difference between the models in play:

```dart
void f<X extends num?>(bool b, X x) {
  if (x is int) { // `x` promoted to `X & int`.
    var y = b ? x : null;
    // Type of `y`:
    //   Current behavior: `(X & int)?`.
    //   #2427: `X?`.
    //   #3435: `int?`.
  }
}
```

eernstg:
Here is another example that could be taken to imply that the #3435 treatment is more consistent with the treatment of all other types than `Null` than this PR:

```dart
class A {}
class B { get b => 1; } // No relation to `A`.
class ASubBSub1 extends B implements A {}
class BSub2 extends B {}

void f<X extends A>(bool b, X x) {
  if (x is ASubBSub1) { // `x` promoted to `X & ASubBSub1`.
    var y = b ? x : BSub2();
    // Type of `y`:
    //   Current specified behavior: `B`.
    //   #2427: `B`.
    //   #3435: `B`.

    // Verify that the implemented behavior uses `B`.
    y.expectStaticType<Exactly<B>>(); // OK.
  }
}

extension <X> on X {
}
```

As we can see, `UP(X & ASubBsub1, BSub2) == UP(ASubBsub1, BSub2) == B`, just like the #3435 treatment of `Null` in the first example: `UP(X & int, Null) == UP(int, Null) == int?`.

lrhn:
Just for the record:
```dart
y.expectStaticType<Exactly<B>>(); // OK.
```
doesn't prove that `y` has type `B`, only that it has a type which _erases_ to `B`, because the static type is captured in a type variable before it's tested.

The real check of the type is that `B b = y;` is valid and `X x = y;` is not. (It's the "not" part that's hard to test without a negative test.)

eernstg:
> .. add `X x = y;` and seeing that it's invalid would work. 

Confirmed, and the error message also refers to `y` as having "type 'B'".

eernstg:
This PR is likely to be abandoned: #3435 looks more promising at this time.

resources/type-system/upper-lower-bounds.md
```diff
@@ -102,22 +105,24 @@ We define the upper bound of two types T1 and T2 to be **UP**(`T1`,`T2`) as foll
...
+
+- **UP**(`T1*`, `T2*`) = `S*` where `S` is **UP**(`T1`, `T2`).
+- **UP**(`T1*`, `T2?`) = `S?` where `S` is **UP**(`T1`, `T2`).
+- **UP**(`T1?`, `T2*`) = `S?` where `S` is **UP**(`T1`, `T2`).
+- **UP**(`T1*`, `T2`) = `S*` where `S` is **UP**(`T1`, `T2`).
+- **UP**(`T1`, `T2*`) = `S*` where `S` is **UP**(`T1`, `T2`).
+
+- **UP**(`T1?`, `T2?`) = `S?` where `S` is **UP**(`T1`, `T2`).
+- **UP**(`T1?`, `T2`) = `S?` where `S` is **UP**(`T1`, `T3`)
+  where `T3` is **ERASEINTERSECTION**(`T2`).
```
lrhn, line 123: Could we make this

... where `S` is **ERASEINTERSECTION**(**UP**(`T1`,`T2`)).

Then **UP**(`int?`, `X&int`) could possibly be `int?`, which or can't when we erase too eagerly.

I don't know if that'll work today, because I haven't been able to memorize the rules, but even if not. maybe it will in the future (say of we improve **UP**(`X&T1`, `X&T3`) at some point).

And we won't get a different result for **UP**(`T1`,`T2?`) than for **UP**(`T1?`,`T2?`).
The only difference is that currently we know that the `T`of `T?` cannot be an intersection type, and the **UP** of two inon-ntersection types cannot be an intersection type, so we don't need to mention **ERASEINTERSECTION** in the  **UP**(`T1?`,`T2?`) case. But maybe we should, just to be consistent and future proof. "Always erase before embedding" into a composite type, even in the places where it's _currently_ never needed.

That is, generally, erase an intersection just before adding the `?`, but no sooner.
That should still be safe, and we'll get the benefit of any improvements we do wrt. Intersection types. And if we write it everywhere, we won't forget a case later if we start introducing intersection types, somehow.

eernstg:
Closing: The proposal in #3435 was accepted and this one will not be used.

PR #2437, opened 2022-08-26
Tweak `Records` class description

lrhn:
Opens up the opportunity to have static members, and is precise that there are no exposed constructors ("members inherited from `Object`" could be read as ambiguous about a default constructor).


munificent (approved):

munificent:
Good catch!

PR #2443, opened 2022-08-29
Revise the views feature specification substantially

eernstg:
This PR uses recent ideas (from the extension struct proposal and several other sources) to create a substantially revised version of the views proposal.

In particular:

- Views no longer have an `implements` clause.
- Views no longer support boxing. 
- Views allow for the use of a 'primary constructor' (a parenthesized list occurring right after the name and type parameters). Hence, the "wrapped" object is denoted by the name declared in this constructor, and `this` has the enclosing view as its type.
- Views use an `export` declaration in the body rather than a `show`/`hide` clause in the header to provide access to members of the underlying representation type.
- Views come in only two variants: Plain ones where the representation type and the view type are unrelated, and `implicit` ones where the representation type is a subtype of the view type.
- Views are enhanced with a new mechanism known as 'view extensions'. A view extension declaration allows for adding members to a view, in a manner which is much like that of a regular extension methods, but the added members are "sticky" (which means that they are available on any expression whose type is the view type, even in the case where the import path to the view extension declaration has more than one edge).

### DIFF
--- a/working/1426-extension-types/feature-specification-views.md
+++ b/working/1426-extension-types/feature-specification-views.md
@@ -14,7 +14,7 @@ Status: Draft
 2021.05.12
   - Initial version.
 
-[1](https://github.com/dart-lang/language/blob/master/working/extension_structs/overview.md)
+[1]: https://github.com/dart-lang/language/blob/master/working/extension_structs/overview.md
 
 
 ## Summary
--- a/working/1426-extension-types/feature-specification-views.md
+++ b/working/1426-extension-types/feature-specification-views.md
@@ -55,13 +55,13 @@ object will never exist at run time, and a reference whose type is the
 view will actually refer directly to the underlying wrapped
 object. Every member access (e.g., an invocation of a method or a
 getter) on an expression whose static type is a view will invoke a
-member of the view (with a few exceptions, as explained below), but
+member of the view (with some exceptions, as explained below), but
 this occurs because those member accesses are resolved statically,
 which means that the wrapper object is not actually needed.
 
-In this document, given that there is no wrapper object, we will refer
-to the "wrapped" object as the _representation object_ of the view, or
-just the _representation_.
+Given that there is no wrapper object, we will refer to the "wrapped"
+object as the _representation object_ of the view, or just the
+_representation_.
 
 Inside the view declaration, the keyword `this` is a reference to the
 representation whose static type is the enclosing view. A member
@@ -73,14 +73,21 @@ is, typed by a "normal" type for the representation) is available as a
 declared name, which is introduced by a new syntax similar to a
 parameter list declaration (for example `(int i)`) which follows the
 name of the view. (This syntax is intended to be a special case of an
-upcoming mechanism known as _primary constructors_.)
+upcoming mechanism known as _primary constructors_.) The
+representation type of the view (with `(int i)` that's `int`) is
+similar to the on-type of an extension declaration.
 
 All in all, a view allows us to replace the interface of a given
 representation object and specify how to implement the new interface
-in terms of the interface of the representation object itself. This is
-something that we could obviously do with a wrapper, but when it is
-done with a view there is no wrapper object, and hence there is no
-run-time performance cost.
+in terms of the interface of the representation object.
+
+This is something that we could obviously do with a wrapper, but when
+it is done with a view there is no wrapper object, and hence there is
+no run-time performance cost. In particular, in the case where we have
+a view `V` with representation type `T` we may be able to refer to a
+`List<T>` using the type `List<V>`, and this corresponds to "wrapping
+every element in the list", but it only takes time _O(1)_ and no
+space, no matter how many elements the list contains.
 
 
 ## Motivation
@@ -92,7 +99,9 @@ operations (the members declared by the given view type).
 
 It is zero-cost in the sense that the value denoted by an expression whose
 type is a view type is an object of a different type (known as the
-_representation type_ of the view type), and there is no wrapper object.
+_representation type_ of the view type), and there is no wrapper
+object, in spite of the fact that the view behaves similarly to a
+wrapping.
 
 The point is that the view type allows for a convenient and safe treatment
 of a given object `o` (and objects reachable from `o`) for a specialized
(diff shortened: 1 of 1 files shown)


working/1426-extension-types/feature-specification-views.md
```diff
@@ -58,12 +110,83 @@ requires a certain discipline in the use of `o`'s instance methods: We may
...
+example:
+
+```dart
+view IdNumber(int i) {
+  // Declare a few members.
+
+  // Assume that it makes sense to compare ID numbers
+  // because they are allocated with increasing values,
+  // so "smaller" means "older".
+  operator <(IdNumber other) => i < (other as int);
```
leafpetersen, line 123: But not `other.i`?

working/1426-extension-types/feature-specification-views.md
```diff
@@ -58,12 +110,83 @@ requires a certain discipline in the use of `o`'s instance methods: We may
...
+  // does not make sense for ID numbers.
+}
+
+void main() {
+  int myUnsafeId = 42424242;
+  myUnsafeId = myUnsafeId + 10; // No complaints.
+
+  var safeId = IdNumber(42424242);
+
+  safeId.verify; // OK, could be true.
```
leafpetersen, line 141: verify is a method in your example.

working/1426-extension-types/feature-specification-views.md
```diff
@@ -74,24 +197,24 @@ reasoning is required may be fragmented into many different locations, and
...
-view TinyJson on Object {
+view TinyJson(Object it) {
   Iterable<num> get leaves sync* {
-    var self = this;
-    if (self is num) {
-      yield self;
-    } else if (self is List<dynamic>) {
-      for (var element in self) {
-        yield* element.leaves;
+    if (it is num) {
```
leafpetersen, line 213: just mentioning in passing that there's an implicit assumption of promotability here.  Probably warranted, but requires we specify the special treatment here.

working/1426-extension-types/feature-specification-views.md
```diff
@@ -149,20 +282,19 @@ working on a wrapper object rather than accessing `o` and its methods
...
-  TinyJson(this.representation);
+  TinyJson(this.it);
 
   Iterable<num> get leaves sync* {
-    var self = representation;
-    if (self is num) {
-      yield self;
-    } else if (self is List<dynamic>) {
-      for (var element in self) {
+    if (it is num) {
```
leafpetersen, line 294: Technically doesn't work because of promotion.

working/1426-extension-types/feature-specification-views.md
```diff
@@ -286,6 +455,33 @@ The token `view` is made a built-in identifier.
...
 
+A few errors can be detected immediately from the syntax:
+
+If a view declaration named `View` includes a
+`<viewPrimaryConstructor>` then it is a compile-time error if the
+declaration includes a constructor declaration named `View`. (*But it
+can still contain other constructors.*)
+
+A compile-time error occurs if a view declaration declares a
+constructor which is redirecting, or a constructor which is
```
leafpetersen, line 466: Why no redirecting constructors?

working/1426-extension-types/feature-specification-views.md
```diff
@@ -286,6 +455,33 @@ The token `view` is made a built-in identifier.
...
+If a view declaration named `View` does not include a
+`<viewPrimaryConstructor>` then an error occurs unless it declares a
+factory constructor named `View`, that declares one required, positional
+formal parameter, and does not declare any other parameters.
+
+The _name of the representation_ in a view declaration that includes a
+`<viewPrimaryConstructor>` is the identifier specified in there. In a
+view declaration named `View` that does not include a
+`<viewPrimaryConstructor>`, the name of the representation is the name
+of the unique formal parameter of the factory constructor named `View`.
```
leafpetersen, line 478: This seems overly limiting to me.  It seems specifically useful to be able to construct the representation in the constructor.  Maybe use the unique field?  Or are you proposing a sort of hybrid model where the `View` constructor also declares the field (even when not a primary constructor)?  This seems unlikely to be consistent with what we do for regular classes.

working/1426-extension-types/feature-specification-views.md
```diff
@@ -319,31 +515,25 @@ but this is ambiguous since
...
-denotes an invocation of the view method `m` with the same bindings.*
-
-The static analysis of `invokeViewMethod` is that it takes exactly three
-positional arguments and must be the receiver in a member access. The first
-argument must be a type name that denotes a view declaration, the next
-argument must be a type argument list, together yielding a view type
-_V_. The third argument must be an expression whose static type is _V_ or
-the corresponding instantiated on-type (defined below). The member access
-must be a member of `V` or an associated member of a superview of `V`.
+is similar to a named constructor invocation, but that is also
```
leafpetersen, line 518: I don't really understand the motivation for this whole section.  Why do we need a syntax here?  Why not just specify the semantics?

working/1426-extension-types/feature-specification-views.md
```diff
@@ -414,130 +607,115 @@ view type_ `V<S1, .. Sk>`, and that its static type _is a view type_.
...
-If there is no declaration in scope whose basename is the basename of `m`,
-`m(args)` is treated as `this.m(args)`. *See a later section for the lookup
-rule when an `extends` clause is present.*
+This is just the same treatment of `this` as in the body of a class.*
 
 *For example:*
 
 ```dart
-extension E1 on int {
+extension E1(int it) {
```
leafpetersen, line 627: extension -> view?

leafpetersen: Or do you mean to not do the primary constructor change here?

working/1426-extension-types/feature-specification-views.md
```diff
@@ -414,130 +607,115 @@ view type_ `V<S1, .. Sk>`, and that its static type _is a view type_.
...
-instance methods on `this`, when `this` is implicit. Note that 
-associated members of superviews can be invoked implicitly as well, as
-specified in section 'Composing view types'.*
+- If `V` is a plain view type then `V` is a proper subtype of
+`Object?`.  *So the representation type and the view type are
+unrelated, and there is no assignability in either direction. In this
+case a view constructor may be used to obtain a value of the view type
+(see below).*
+- If `V` is an implicit view type then `V` is a proper subtype of
+`Object?`, and a proper supertype of `T`. *That is, an expression of
```
leafpetersen, line 690: I'm not sure proper is true.  What if `T` is `Object?`?

working/1426-extension-types/feature-specification-views.md
```diff
@@ -546,48 +724,127 @@ or
...
-`<viewShowHidePart>` in a view declaration.
-
-*The show/hide part provides access to a subset of the members of the
-interface of the on-type. For instance, if the intended purpose of the
-view type is to maintain a certain set of invariants about the state
-of the on-type instance, it is no problem to let clients invoke any methods
-that do not change the state. We could write forwarding members in the
-view body to enable those methods, but using show/hide can have the
-same effect, and it is much more concise and convenient.*
+*This means that the constructor can return an expression whose static type
```
leafpetersen, line 731: I'm not sure how I feel about this approach, relative to just saying that you need to initialize the unique field as normal, unless it's a factory.

working/1426-extension-types/feature-specification-views.md
eernstg: I hadn't thought about that, but I do think it would be useful to support that kind of access.

The developer who writes a view can choose a private name for the "wrapped object" (that is, the 'representation name') and disable this mechanism, but otherwise the representation name could be used as a getter that provides access to the representation object under its "original" type.

This is _better_ than casting, because there's no guarantee about a cast being correct (that is, we have no way to make `myIdNumber as int` pass but warn/err in case of `myIdNumber as String`).

The semantics is kind of funny, but it is all statically resolved, so I wouldn't expect the handling of `other.id` to introduce any special difficulties during code generation.

I adjusted the example to mention this.

working/1426-extension-types/feature-specification-views.md
eernstg: Indeed!

PR #2449, opened 2022-08-31
[patterns] Destructuring assignment, clarify patterns in for, etc.

### LINKED ISSUE #2438: Destructuring assignment
The [patterns proposal](https://github.com/dart-lang/language/blob/master/working/0546-patterns/patterns-feature-specification.md) currently supports destructuring *variable declarations*, but it has no way to use destructuring to assign to existing variables. Should it?

I think the answer is generally "yes", though it raises interesting questions about: 

* What kinds of patterns are allowed on the left-hand side of an `=`? Certainly not refutable ones. Can you also declare variables?

* Where a destructuring assignment expression is allowed to appear? Do we only allow this for assignments in "statement position" inside an expression statements, or anywhere an assignment can occur? If it declares variables, what is their scope?

* Are there any ambiguities with existing assignment target syntax?

* Does it work with any compound assignment operators? Probably not.

### Syntax

We extend `expression`:

```
expression  ::= patternAssignment
              | // Existing productions...

patternAssignment ::= pattern '=' expression
```

As with pattern variable declarations, we restrict what kind of pattern the outermost pattern can be. It must be one of:

* groupingPattern
* listPattern
* mapPattern
* recordPattern
* extractorPattern
* nullCheckPattern

Also:

* An assignment is an irrefutable context, so it's a compile error to have a refutable pattern in there.

* It's a compile-time error if a variable pattern has a type annotation.

Note that this syntax allows chaining pattern assignments (probably not very useful) but doesn't allow any compound assignment operators.

### Semantics

All variable and cast patterns are interpreted as assigning to an existing variable (or setter) with that name using the normal scope resolution.

You can't declare any new variables. Disallowing this allows pattern assignment expressions to appear anywhere expressions are allowed without raising tricky questions around the scope of any new variables.

It's a compile time error if a variable or cast pattern refers to a name that can't be resolved. In assignments, variable and cast patterns are always treated as irrefutable, and their static type is the type of the assigned variable or setter.

The runtime semantics are a little subtle because patterns (even irrefutable ones) can have side effects, as can assignment. Consider:

```dart
var a = 'before';
try {
  Object o = 'not int';
  [a, b as int] = ['after', o];
} catch (_) {
  // Ignore exception.
}
print(a);
```

Should this print "before" or "after"?

An even stranger one is:

```dart
var x = 'x';

class FunnyList implements List<String> {
  String operator[](int index) => '$x!';

  int get length => 2;
  // Other List methods...
}

main() {
  var y = 'y before';
  [x, y] = FunnyList();
  print('$x $y');
}
```

Does this print "x! x!!" or "x! x!"?

I think the right thing is to defer all assignments until after the destructuring is done, so:

1.   Evaluate the right-hand side expression to a value `v`.
2.   Match `v` against the pattern on the left. When matching a variable or cast pattern against a value `o`, record that `o` will be the new value for the corresponding variable/setter, but do not store the variable or invoke the setter.
3.   Once all destructuring and matching is done, store all of the assigned variables and invoke the setters with their corresponding values.

In other words, it's as if every variable or cast pattern in an assignment expression *is* a new variable declaration (with a hidden name). And then after the assignment expression and matching completes, those temporary variables are all then written to the corresponding real variables and setters.

### LINKED ISSUE #2441: Should empty list and map patterns be supported?
The [patterns proposal](https://github.com/dart-lang/language/blob/master/working/0546-patterns/patterns-feature-specification.md) defines `listPattern` and `mapPattern` as non-empty:

> ```
> listPattern       ::= typeArguments? '[' patterns ']'
> patterns          ::= pattern ( ',' pattern )* ','?
> mapPattern        ::= typeArguments? '{' mapPatternEntries '}'
> mapPatternEntries ::= mapPatternEntry ( ',' mapPatternEntry )* ','?
> ```

Is this intentional?  Since the runtime semantics of both list and map patterns includes a length check, it seems like allowing empty list and map patterns could be useful.

If we do decide that we want to allow empty list and/or map patterns, we'll also need to specify their type schemas.  Currenly we specify that a list pattern without an explicit type gets its type schema from the GLB of the type schemas of all the element subpatterns; that doesn't work if there are no element subpatterns at all.  Empty map patterns have a similar problem.

### LINKED ISSUE #2446: [Patterns] Question about pattern variable declaration grammar
The proposal, under "Pattern variable declaration" says:

> It is a compile-time error if the outermost pattern in a `patternDeclaration` is not one of:
>
> * `groupingPattern`
> * `listPattern`
> * `mapPattern`
> * `recordPattern`
> * `extractorPattern`
> * `nullCheckPattern`

Would it be better to have a separate production for these six kinds of patterns and use it in the production for `patternDeclaration`? Doing so would mean that it was a syntactic error to use any other kind of pattern in that location making it possible for the parser to report the problem.

Later, in the same section, it says

> But excludes other kinds of patterns to prohibit weird code like:
>
> `// Not allowed:`
> ...
> `var (pointless) = 'parentheses';  // Grouping.`

Even though a grouping pattern is explicitly allowed above. Which is correct?

### LINKED ISSUE #2447: [Patterns] Question about grouping pattern
Would it be possible to change `groupingPattern` to `parenthesizedPattern`? The phrase "grouping pattern" makes me think of match groups in regular expressions, and this construct doesn't have those semantics.

### LINKED ISSUE #2448: [Patterns] Question about patterns in a for loop
There doesn't appear to be any discussion about the use of a `patternDeclaration` in a for loop, such as what the semantics are. Does the expression after the `in` have to have the type `Iterator`? Are there actually two expressions? If so, what is the one following the equal sign for?

```
void f(List<(int, int)> listOfPairs) {
  for (var (int left, int right) = ? in listOfPairs) { ... }
}
```

munificent:
- Add destructuring assignment. Fix #2438.
- Specify context type for empty list and map patterns. Fix #2441.
- Define grammar for outermost patterns in a declaration. Fix #2446.
- Rename "grouping" to "parenthesized" patterns. Fix #2447.
- Specify behavior of patterns in for loops better. Fix #2448.

There are enough substantive changes that I thought a real review would be worth it instead of just landing. If any of you have the time, I'd appreciate it. If not, I can always just merge.

### DIFF
--- a/working/0546-patterns/patterns-feature-specification.md
+++ b/working/0546-patterns/patterns-feature-specification.md
@@ -1592,7 +1592,7 @@ main() {
 
 We define an *irrefutable context* as the pattern in a
 `localVariableDeclaration`, `forLoopParts`, or `patternAssignment`. A *refutable
-context* is the pattern in a `caseHead` or `ifCondition` or its subpatterns.
+context* is the pattern in a `caseHead` or `ifCondition`.
 
 Refutability is not just a property of the pattern itself. It also depends on
 the static type of the value being matched. Consider:
@@ -1613,30 +1613,30 @@ the second function, `obj` may fail to match because the value may not be a
 record. *This implies that we can't determine whether a pattern in a variable
 declaration is incorrectly refutable until after type checking.*
 
-Refutability of a pattern `p` matching a value of type `v` is:
+Refutability of a pattern `p` matching a value of type `V` is:
 
-*   **Logical-and**, **parenthesized**, **null-assert**, or **cast**: Always
-    irrefutable (though may contain refutable subpatterns).
+*   **Logical-and**, **parenthesized**, **null-assert**, or **cast**:
+    Irrefutable if and only if all subpatterns are irrefutable.
 
 *   **Logical-or**, **relational**, **null-check**, **literal**, or
     **constant**: Always refutable. *Logical-or patterns are refutable because
     there is no point in using one with an irrefutable left operand. We could
-    make null-check patterns irrefutable if `v` is assignable to its static
+    make null-check patterns irrefutable if `V` is assignable to its static
     type, but whenever that is true the pattern does nothing useful since its
     only behavior is a type test.*
 
 *   **variable**, **list**, **map**, **record**, or **extractor**: Irrefutable
-    if `v` is assignable to the static type of `p`. *If `p` is a variable
-    pattern with no type annotation, the type is inferred from `v`, so it is
-    never refutable.*
+    if and only if `V` is assignable to the static type of `p` and all
+    subpatterns are irrefutable. *If `p` is a variable pattern with no type
+    annotation, the type is inferred from `V`, so it is never refutable.*
 
 It is a compile-time error if a refutable pattern appears in an irrefutable
-context, either as the outermost pattern or a subpattern. *This means that the
-explicit predicate patterns like constants and literals can never appear in
-pattern variable declarations or pattern assignments. The patterns that do type
-tests directly or implicitly can appear in variable declarations or assignments
-only if the tested type is a supertype of the value type. In other words, any
-pattern that needs to "downcast" to match is refutable.*
+context. *This means that the explicit predicate patterns like constants and
+literals can never appear in pattern variable declarations or pattern
+assignments. The patterns that do type tests directly or implicitly can appear
+in variable declarations or assignments only if the tested type is assignable
+from the value type. In other words, any pattern that needs to "downcast" to
+match is refutable.*
 
 ### Variables and scope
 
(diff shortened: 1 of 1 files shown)


working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1564,10 +1618,10 @@ Refutability of a pattern `p` matching a value of type `v` is:
 It is a compile-time error if a refutable pattern appears in an irrefutable
 context, either as the outermost pattern or a subpattern. *This means that the
 explicit predicate patterns like constants and literals can never appear in
-pattern variable declarations. The patterns that do type tests directly or
-implicitly can appear in variable declarations only if the tested type is a
-supertype of the value type. In other words, any pattern that needs to
-"downcast" to match is refutable.*
+pattern variable declarations or pattern assignments. The patterns that do type
+tests directly or implicitly can appear in variable declarations or assignments
+only if the tested type is a supertype of the value type. In other words, any
```
leafpetersen, line 1623: This isn't true, because of dynamic?  The value type must be assignable to the value type, but not necessarily a subtype?

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1726,6 +1801,61 @@ behavior.
...
+
+```dart
+for (<keyword> <pattern> in <expression>) <statement>
+```
+
+Where `<keyword>` is `var` or `final` is interpreted as the following code,
+where `id1` and `id2` are fresh identifiers:
+
+```
+var id1 = <expression>;
```
leafpetersen, line 1851: Current spec says:

T id1 = e;
var id2 = id1.iterator;
while (id2.moveNext()) {
D id = id2.current;
{ S }
}
Dart Programming Language Specification 192
If the static type of e is a top type (15.2) then T is Iterable<dynamic>, otherwise T is the static type of e. It is a compile-time error if T is not assignable
to Iterable<dynamic>.

We probably want to keep the constraints around Iterable?

leafpetersen (approved):

PR #2454, opened 2022-09-01
[patterns] Tease apart type testing from refutability.

### LINKED ISSUE #2450: [Patterns] Refutability and assignability
The patterns proposal says:

> *   **Null-check**, **variable**, **list**, **map**, **record**, or
    **extractor**: Irrefutable if `v` is assignable to the static type of `p`.
    *If `p` is a variable pattern with no type annotation, the type is inferred
    from `v`, so it is never refutable.*

I don't understand why this is specified this way.  It seems to me that in a refutable context, you don't care (these are refutable, I think), and in an irrefutable context they are considered as irrefutable.  Separately, we should then say that it is an error if `v` is not assignable to the type associated with `p`, but that's not about refutability, that's about typing, right?  

Or is there something I'm missing here?  Either way, some explanation would help.

munificent:
Some patterns fail to match if the value doesn't have a certain type.
We still allow those patterns in irrefutable contexts if the matched
value is statically known to always have that type (or be dynamic).

Instead of specifying by having the refutability state of those patterns
depend on their type, make refutability a syntactic property and then
define the compile-time error for the types during type checking.

Fix #2450.

### DIFF
--- a/working/0546-patterns/patterns-feature-specification.md
+++ b/working/0546-patterns/patterns-feature-specification.md
@@ -1190,7 +1190,7 @@ fails to match. For example:
 
 ```dart
 main() {
-  var (a, b) = (1, 2);
+  var (a, 5) = (1, 2);
   print(a + b);
 }
 ```
@@ -1199,14 +1199,9 @@ What happens when `a` is printed in the example above?
 
 To avoid that, we restrict which patterns can be used in variable declarations
 and pattern assignments. Only *irrefutable* patterns that never fail to match
-are allowed in contexts where match failure can't be handled. For example, this
-is an error:
-
-```dart
-main() {
-  var (== 2, == 3) = (1, 2);
-}
-```
+are allowed in contexts where match failure can't be handled. Since a literal
+pattern like `5` in the above example might fail to match, it's prohibited and
+the above example has a compile error.
 
 We define an *irrefutable context* as the pattern in a
 `localVariableDeclaration`, `forLoopParts`, or `patternAssignment` or any of its
--- a/working/0546-patterns/patterns-feature-specification.md
+++ b/working/0546-patterns/patterns-feature-specification.md
@@ -1165,6 +1165,71 @@ That is less important here since the only other case is the else branch.
 
 ## Static semantics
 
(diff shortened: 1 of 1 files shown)


munificent:
Does this make more sense?

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1165,6 +1165,71 @@ That is less important here since the only other case is the else branch.
...
+  var (== 2, == 3) = (1, 2);
+}
+```
+
+We define an *irrefutable context* as the pattern in a
+`localVariableDeclaration`, `forLoopParts`, or `patternAssignment` or any of its
+subpatterns. It is a compile-time error if any of these patterns appear in an
+irrefutable context:
+
+*   Logical-or
```
lrhn, line 1216: (Doesn't *have* to be irrefutable, but it's probably hard to do anything useful in an `or` that requires at least one branch to match. If the last case always works anyway, why not just use it for everything.)

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1165,6 +1165,71 @@ That is less important here since the only other case is the else branch.
...
+*   Relational
+*   Null-check
+*   Literal
+*   Constant
+
+*Logical-or patterns are refutable because there is no point in using one with
+an irrefutable left operand. We could make null-check patterns irrefutable if
```
lrhn, line 1223: You could have an or-pattern with an irrefutable *right* operand, that would be irrefutable by itself.

It's probably not *useful*. If the second pattern always works anyway, why not just use it for everything. And they'd have to bind the same variables, if they bind anything at all (and if not, it's *really* not being useful).

lrhn (commented):
Drive-by comment. Looks good to me, but leave it to the type experts.

working/0546-patterns/patterns-feature-specification.md
munificent: Yeah, I think it's just not useful to allow them in irrefutable contexts.

leafpetersen (approved):
LGTM, thanks!  I find this clearer.

Issue #2473, opened 2022-09-09
Loosen the restriction on variables declared in case patterns that share a body


munificent:
When refactoring dart_style to try out some pattern syntax, I ran into this example:

```dart
switch (expression) {
  case NamedExpression(:var expression)
      when _isBlockFunction(expression):
    return true;

  case MethodInvocation(:var target, argumentList: [var arg])
      when _isValidWrappingTarget(target) && _isBlockFunction(arg):
    return true;

  case InstanceCreationExpression(argumentList: [var arg]):
      when _isBlockFunction(arg);
    return true;

  case FunctionExpressionInvocation(:var function, [_, ...])
      when _isBlockFunction(function):
    return true;

  case ParenthesizedExpression(:var expression)
      when _isBlockFunction(expression):
    return true;

  case FunctionExpression(body: BlockFunctionBody(:var block))
      when block.statements.isNotEmpty ||
          block.rightBracket.precedingComments != null:
    return true;

  default:
    return false;
}
```

Note that all of the cases have the same body. It would be nice to rely on fallthrough to let them actually just share one body:

```dart
switch (expression) {
  case NamedExpression(:var expression)
      when _isBlockFunction(expression):
  case MethodInvocation(:var target, argumentList: [var arg])
      when _isValidWrappingTarget(target) && _isBlockFunction(arg):
  case InstanceCreationExpression(argumentList: [var arg]):
      when _isBlockFunction(arg);
  case FunctionExpressionInvocation(:var function, [_, ...])
      when _isBlockFunction(function):
  case ParenthesizedExpression(:var expression)
      when _isBlockFunction(expression):
  case FunctionExpression(body: BlockFunctionBody(:var block))
      when block.statements.isNotEmpty ||
          block.rightBracket.precedingComments != null:
    return true;

  default:
    return false;
}
```

This is much shorter and there's nothing problematic about the code. But it violates the current proposal because some of those cases do bind variables that others don't. It's not a problem because the variables only end up being used in their respective guard clause.

I think we should allow code like this. To enable that, I suggest that we loosen the rule to say that it's it's a compile-time error if a variable *used inside a case body* is not declared by all case patterns with the same type. If the variable isn't used in the case body, it's fine if not all cases define it or it has different types in each.

Thoughts?

leafpetersen:
As an aside, are the cases there overlapping?  Otherwise why not:
```dart
return switch (expression) {
  case NamedExpression(:var expression): _isBlockFunction(expression);
[...]

PR #2475, opened 2022-09-10
Explicit binders in cases

munificent:
OK, I've revamped the patterns proposal to resolve (hopefully once and for all!) how we want to handle variable binders.

I think this is overall a significant improvement over the current proposal. It's roughly similar to the original proposal, but still retains the unified grammar in terms of how its specified and is, I think, easier to understand.

cc @stereotype441 @natebosch @jakemac53 @kallentu 

Note that there is a "CHANGELOG" section at the end of the proposal that goes into more detail about what actually changed.

### DIFF
--- a/working/0546-patterns/patterns-feature-specification.md
+++ b/working/0546-patterns/patterns-feature-specification.md
@@ -353,18 +353,12 @@ A cast pattern is similar to an extractor pattern in that it checks the matched
 value against a given type. But where an extractor pattern is *refuted* if the
 value doesn't have that type, a cast pattern *throws*. Like the null-assert
 pattern, this lets you forcibly assert the expected type of some destructured
-value. This isn't useful as the outermost pattern in a declaration since you can
-always move the `as` to the initializer expression:
+value.
 
-```dart
-num n = 1;
-var i as int = n; // Instead of this...
-var i = n as int; // ...do this.
-```
-
-But when destructuring, there is no place in the initializer to insert the cast.
-This pattern lets you insert the cast as values are being pulled out by the
-pattern:
+This isn't useful as the outermost pattern in a declaration since you can always
+move the `as` to the initializer expression, but when destructuring there is no
+place in the initializer to insert the cast. This pattern lets you insert the
+cast as values are being pulled out by the pattern:
 
 ```dart
 (num, Object) record = (1, "s");
@@ -485,33 +479,6 @@ valid constant pattern.*
 
 It is a compile-time error if a constant pattern's value is not a valid constant
 expression.
-
-**Breaking change**: Using patterns in switch cases means that a list or map
-literal in a switch case is now interpreted as a list or map pattern which
-destructures its elements at runtime. Before, it was simply treated as identity
-comparison.
-
-```dart
-const a = 1;
-const b = 2;
-var obj = [1, 2]; // Not const.
-
-switch (obj) {
-  case [a, b]: print("match"); break;
-  default: print("no match");
-}
-```
-
-In Dart today, this prints "no match". With this proposal, it changes to
-"match". However, looking at the 22,943 switch cases in 1,000 pub packages
-(10,599,303 lines in 34,917 files), I found zero case expressions that were
-collection literals.
-
-If you want the previous behavior, prefix the list or map literal with `const`:
-
-```dart
-case const [a, b]:
-```
 
 ### Variable pattern
 
@@ -778,6 +745,7 @@ The `outerPattern` rule defines a subset of the patterns that are allowed as the
 outermost pattern in a declaration. Subsetting allows useful code like:
 
 ```dart
+var ((a, b) & record) = (1, 2);           // Parentheses.
 var [a, b] = [1, 2];                      // List.
 var {1: a} = {1: 2};                      // Map.
 var (a, b, x: x) = (1, 2, x: 3);          // Record.
@@ -815,8 +783,8 @@ As with regular for-in loops, it is a compile-time error if the type of
 `expression` in a pattern-for-in loop is not assignable to `Iterable<dynamic>`.
 
 *We could potentially allow patterns in top-level variables and static fields
-but laziness makes that more complex. We could support patterns in instance
-field declarations, but constructor initializer lists make that harder.
+but lazy initialization makes that more complex. We could support patterns in
+instance field declarations, but constructor initializer lists make that harder.
 Parameter lists are a natural place to allow patterns, but the existing grammar
 complexity of parameter lists&mdash;optional parameters, named parameters,
 required parameters, default values, etc.&mdash;make that very hard. For the
@@ -845,14 +813,14 @@ are allowed while avoiding confusion about the scope of new variables.*
 
 It is a compile-time error if:
 
-*   An identifier in a variable or cast pattern does not resolve to a non-final
-    local variable. *We could allow assigning to other variables or setters, but
-    it seems strange to allow assigning to `foo` when `foo` is an instance field
-    on the surrounding class with an implicit `this.`, but not allowing to
-    assign to `this.foo` explicitly. In the future, we may expand pattern
-    assignment syntax to allow other selector expressions. For now, we restrict
-    assignment to local variables, which are also the only kind of variables
-    that can be declared by patterns.*
+*   An identifier in a variable pattern does not resolve to a non-final local
+    variable. *We could allow assigning to other variables or setters, but it
+    seems strange to allow assigning to `foo` when `foo` is an instance field on
+    the surrounding class with an implicit `this.`, but not allowing to assign
+    to `this.foo` explicitly. In the future, we may expand pattern assignment
+    syntax to allow other selector expressions. For now, we restrict assignment
+    to local variables, which are also the only kind of variables that can be
+    declared by patterns.*
 
 *   The matched value type for a variable or cast pattern is not assignable to
     the corresponding variable's type.
@@ -881,6 +849,103 @@ Allowing patterns in cases significantly increases the expressiveness of what
 properties a case can verify, including executing arbitrary user-defined code.
 This implies that the order that cases are checked is now potentially
 user-visible and an implementation must execute the *first* case that matches.
+
+#### Breaking existing switches
+
+Many constant expressions are subsumed by the new pattern syntax so most
+existing switch cases have the same semantics under this proposal. However,
+patterns are not a strict superset of constant expressions and some switches may
+be broken.
+
+To estimate how breaking these changes are, I analyzed 18,672,247 lines of code
+in 102,015 files across 2,000 Pub packages, a large collection of open source
+Flutter applications, and the Dart and Flutter repositories. I found a total of
+94,249 switch cases.
+
+The specific kinds of switches whose behavior changes are:
+
+*   **List and map patterns.** A list or map constant literal in a switch case
+    is now interpreted as a list or map *pattern* which destructures its
+    elements at runtime. Before, it was simply treated as identity comparison.
+
+    ```dart
+    const a = 1;
+    const b = 2;
+    var obj = [1, 2]; // Not const.
+
+    switch (obj) {
+      case [a, b]: print("match"); break;
+      default: print("no match");
+    }
+    ```
+
+    In Dart today, this prints "no match". With this proposal, it changes to
+    "match". I did not find any switch cases whose expression is a list or map
+    literal.
+
+*   **Wildcards.** A switch case containing the identifier `_` currently matches
+    if the matched value is equal to the constant named `_`. With this proposal,
+    it becomes a wildcard that always matches. I did not find any switch cases
+    whose expression is `_`.
+
+*   **Constant constructors.** A switch case can be a constant constructor call
+    with implicit `const`, like:
+
+    ```dart
+    case SomeClass(1, 2):
+    ```
+
+    With this proposal, that is interpreted as an extractor pattern whose
+    arguments are subpatterns. In cases where the matched value is also a
+    constant, this will *likely* behave the same but may not. I found 8 switch
+    cases of this form (0.008%).
+
+*   **Other constant expressions.** Constant patterns allow simple literals and
+    references to named constants to be used directly as patterns, which covers
+    the majority of all existing switch cases. Also a constant constructor
+    explicitly prefixed with `const` is a valid constant expression pattern. But
+    some more complex expressions are valid constant expressions but not valid
+    constant patterns. In the switch cases I analyzed, the exceptions are:
+
+    ```
+    case A + A:                                         // Infix "+".
+    case A + 'b':                                       // Infix "+".
+    case -ERR_LDS_ICAO_SIGNED_DATA_SIGNER_INFOS_EMPTY:  // Unary "-".
+    case -sigkill:                                      // Unary "-".
+    case List<RPChoice>:                                // Generic type literal.
+    case 720 * 1280:                                    // Infix "*".
+    case 1080 * 1920:                                   // Infix "*".
+    case 1440 * 2560:                                   // Infix "*".
+    case 2160 * 3840:                                   // Infix "*".
+    ```
+
+    These nine cases represent 0.009% of the cases found.
+
+For any switch case that is broken by this proposal, you can revert back to the
+original behavior by prefixing the case expression (now pattern) with `const`:
+
+```dart
+// List or map literal:
+case const [a, b]:
+
+// Const constructor call:
+case const SomeClass(1, 2):
+
+// Other constant expression:
+case const A + A:
+case const A + 'b':
+case const -ERR_LDS_ICAO_SIGNED_DATA_SIGNER_INFOS_EMPTY:
+case const -sigkill:
+case const List<RPChoice>:
+case const 720 * 1280:
+case const 1080 * 1920:
+case const 1440 * 2560:
+case const 2160 * 3840:
+```
+
+We can determine syntactically whether an existing switch case's behavior will
+be changed by this proposal, so this fix can be easily automated and applied
+mechanically.
 
 #### Guard clause
 
@@ -1169,9 +1234,9 @@ allowed and what their syntax is. The rules are:
 
     *The remaining patterns are allowed syntactically to appear in a refutable
     context. Patterns that do type tests like variables and lists produce a
-    compile-time error when used in an irrefutable context if the matched value
-    isn't assignable to their required type. This error is specified under type
-    checking.*
+    compile-time error when used in an irrefutable context if the static type of
+    the matched value isn't assignable to their required type. This error is
+    specified under type checking.*
 
 *   It is a compile-time error if a variable pattern in a declaration context is
     marked `var` or `final`. *A pattern declaration statement is already
@@ -1235,7 +1300,7 @@ allowed and what their syntax is. The rules are:
     ```dart
     // OK:
     switch (triple) {
-      case [_, y, _]: print('The middle element is $y');
+      case [_, var y, _]: print('The middle element is $y');
     }
     ```
 
@@ -1243,11 +1308,11 @@ allowed and what their syntax is. The rules are:
     context because it would require additional specification to explicitly
     forbid it, but doing so is discouraged.*
 
-*In short, don't use refutable patterns in places that don't do control flow.
-Use simple identifiers (optionally with type annotations) to declare variables
-in pattern declarations. Use simple identifiers to assign to variables in
-pattern assignments. Use explicitly marked identifiers to declare variables in
-cases. Use `_` anywhere for a wildcard.*
+*In short, you can't use refutable patterns in places that don't do control
+flow. Use simple identifiers (optionally with type annotations) to declare
+variables in pattern declarations. Use simple identifiers to assign to variables
+in pattern assignments. Use explicitly marked identifiers to declare variables
+in `case` patterns. Use `_` anywhere for a wildcard.*
 
 ## Static semantics
 
@@ -1493,19 +1558,29 @@ To type check a pattern `p` being matched against a value of type `M`:
 *   **Logical-or** and **logical-and**: Type check each branch using `M` as the
     matched value type.
 
-*   **Relational**: If the operator is a comparison (`<`, `<=`, `>`, or `>=`),
-    then it is a compile-time error if `M` does not define that operator, if the
-    type of the constant in the relational pattern is not assignable to the
-    operator's parameter type, or if the operator's return type is not
-    assignable to `bool`. *The `==` and `!=` operators are valid for all pairs
-    of types.*
+*   **Relational**:
+
+    1.  Let `C` be the static type of the right operand constant expression.
+
+    2.  If the operator is a comparison (`<`, `<=`, `>`, or `>=`), then it is a
+        compile-time error if:
+
+        *   `M` does not define that operator,
+        *   `C` is not assignable to the operator's parameter type,
+        *   or if the operator's return type is not assignable to `bool`.
+
+    3.  Else the operator is `==` or `!=`. It is a compile-time error if `C?` is
+        not assignable to `M`'s `==` method parameter type. *The language
+        screens out `null` before calling the underlying `==` method, which is
+        why `C?` is the allowed type. Since Object declares `==` to accept
+        `Object` on the right, this compile-time error can only happen if a
+        user-defined class has an override of `==` with a `covariant`
+        parameter.*
 
 *   **Cast**:
 
     1.  Resolve the type name to a type `X`. It is a compile-time error if
         the name does not refer to a type.
-
-        **TODO: Should we infer type arguments on `X`?**
 
     2.  Type-check the subpattern using `X` as the matched value type.
 
@@ -1533,8 +1608,8 @@ To type check a pattern `p` being matched against a value of type `M`:
 
 *   **Variable**:
 
-    1.  In an assignment context, the required type of `p` is the static type
-        of the variable that `p` resolves to.
+    1.  In an assignment context, the required type of `p` is the (unpromoted)
+        static type of the variable that `p` resolves to.
 
     2.  Else if the variable has a type annotation, the required type of `p` is
         that type, as is the static type of the variable introduced by `p`.
@@ -1643,12 +1718,13 @@ To type check a pattern `p` being matched against a value of type `M`:
 
 It is a compile-time error if:
 
-*   The type of an expression in a guard clause is not `bool` or `dynamic`.
+*   The type of an expression in a guard clause is not assignable to `bool`.
 
 *   `p` is in an irrefutable context, it has a required type `T`, and `M` is not
     assignable to `T`. *Destructuring and variable patterns can only be used in
     declarations and assignments if we can statically tell that the
-    destructuring and variable binding won't fail.*
+    destructuring and variable binding won't fail to match (though it might
+    throw a runtime exception from implicit downcasts from `dynamic`).*
 
 ### Pattern uses
 
@@ -1668,6 +1744,9 @@ The variables a patterns binds depend on what kind of pattern it is:
     that do. It is a compile-time error if the two subpatterns do not introduce
     the same variables with the same names and types.
 
+    **TODO: "Same name" is underspecified and also probably
+    [too conservative][2473].**
+
 *   **Logical-and**, **cast**, **null-check**, **null-assert**,
     **parenthesized**, **list**, **map**, **record**, or **extractor**: These do
     not introduce variables themselves but may contain subpatterns that do.
@@ -1680,8 +1759,10 @@ The variables a patterns binds depend on what kind of pattern it is:
     matching context, the variable is final if the variable pattern is marked
     `final` and is not otherwise.
 
-The scope where a pattern's variables are declared depends on the construct
-that contains the pattern:
+[2473]: https://github.com/dart-lang/language/issues/2473
+
+The scope where a pattern's variables are declared and available for use depends
+on the construct that contains the pattern:
 
 *   **Local pattern variable declaration**: The rest of the block following
     the declaration.
@@ -1783,7 +1864,7 @@ fail in some way.*
         other cases).
 
     2.  If there is a guard clause, evaluate it. If it does not evaluate to a
-        Boolean, throw a runtime exception. *This can happen if the guard
+        Boolean, throw a runtime error. *This can happen if the guard
         expression's type is `dynamic`.* If it evaluates to `false`, continue to
         the next case (or default or exit).
 
@@ -1798,7 +1879,7 @@ fail in some way.*
     statements after it.
 
 4.  If no case matches and there is no default clause, throw a runtime
-    exception. *This can only occur when `null` or a legacy typed value flows
+    error. *This can only occur when `null` or a legacy typed value flows
     into this switch statement from another library that hasn't migrated to
     [null safety][]. In fully migrated programs, exhaustiveness checking is
     sound and it isn't possible to reach this runtime error.*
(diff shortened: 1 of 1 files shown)


working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1885,14 +1900,27 @@ Where `<keyword>` is `var` or `final` is treated like so:
...
 
-#### Pattern-if statement
+#### If-case statement
 
 1.  Evaluate the `expression` producing `v`.
 
-2.  Match the `pattern` against `v`.
+2.  Match the `pattern` in the `caseHead` against `v`.
+
+3.  If the match succeeds:
```
lrhn, line 1909: Worth mentioning that new variables are available here?

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1885,14 +1900,27 @@ Where `<keyword>` is `var` or `final` is treated like so:
...
+
+    1.  If there is a guard clause:
+
+        1.  Evaluate it. If it does not evaluate to a Boolean, throw a runtime
+            exception. *This can happen if the guard expression's type is
```
lrhn, line 1914: exception -> error.

In practice, this should be a completely normal failing implicit downcast from `dynamic` to `bool`.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1885,14 +1900,27 @@ Where `<keyword>` is `var` or `final` is treated like so:
...
+            `dynamic`.*
 
-3.  If the match succeeds, evaluate the then `statement`. Otherwise, if there
-    is an `else` clause, evaluate the else `statement`.
+        1.  If the guard evaluates to `true`, evaluate the then `statement`.
```
lrhn, line 1917: Nit: evaluate -> execute (which is what we say we do for statements, evaluation is for expressions).

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1940,6 +1968,15 @@ To match a pattern `p` against a value `v`:
         method of the same name on the matched value, with `c` as the argument
         returns true.
 
+*   **Cast**:
+
+    1.  If the runtime type of `v` is not a subtype of the cast type of `p` then
+        throw a runtime exception. *Note that we throw even if this appears in a
```
lrhn, line 1974: exception -> error. (It'll be a `TypeError`)

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1949,31 +1986,27 @@ To match a pattern `p` against a value `v`:
 *   **Null-assert**:
 
     1.  If `v` is null then throw a runtime exception. *Note that we throw even
```
lrhn, line 1988: exceptionn -> error

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1949,31 +1986,27 @@ To match a pattern `p` against a value `v`:
...
 
-*   **Cast**:
+*   **Variable**:
 
-    1.  If the runtime type of `v` is not a subtype of the static type of `p`
-        then throw a runtime exception. *Note that we throw even if this appears
-        in a refutable context. The intent of this pattern is to assert that a
-        value *must* have some type.*
+    1.  If the runtime type of `v` is not a subtype of the static type of the
+        variable that `p` refers to then the match fails.
```
lrhn, line 2007: This is an assignment, it can only happen in an irrefutable context.
So, instead of the match failing, we should throw a runtime error (and it will always be a failed downcast from `dynamic`).

lrhn (approved):

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1133,35 +1101,153 @@ if (var [int x, int y] = json) {
...
+    const c = 1;
+    switch (2) {
+      case c: print('match $c');
+      default: print('no match');
+    }
+    ```
+
+    *This program prints "no match" and not "match 2".*
+
+*   A simple identifier in a matching context named `_` is treated as a wildcard
```
stereotype441, line 1231: Technically this is a breaking change, since in Dart today we can do:

```dart
main() {
  const _ = 1;
  switch (2) {
    case _:
      print('Matched _');
      break;
    default:
      print('Did not match _');
      break;
  }
}
```

(and this program prints `Did not match _`).

I think this is fine, because the chances of the breakage affecting a real user are so low, but we should probably mention in this doc that it's a breakage just for completeness.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1314,12 +1402,15 @@ The context type schema for a pattern `p` is:
...
 
-*   **Literal** or **constant**: The context type schema is the static type of
-    the pattern's constant value expression.
+*   **Constant**: The context type schema is the static type of the pattern's
+    value.
 
 *   **Variable**:
 
-    1.  If `p` has no type annotation, the context type schema is `?`.
+    1.  In an assignment context, the context type schema is the static type of
```
stereotype441, line 1410: What if the variable is promoted?  Does `static type` mean the promoted type or the unpromoted type of the variable?

I think it should mean the unpromoted type, so that demoting assignments are possible (as they are for ordinary non-pattern assignments).

stereotype441 (approved):
lgtm, though my review is nowhere near as thorough as @lrhn's, so if he contradicts me I defer to him :)

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -144,8 +144,8 @@ matching constructs to make working with that style enjoyable:
 ```dart
 double calculateArea(Shape shape) =>
   switch (shape) {
-    case Square(length: l) => l * l;
-    case Circle(radius: r) => math.pi * r * r;
+    case Square(length: var l) => l * l;
```
jakemac53, line 147: 🎉 this makes things way more clear imo

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -357,10 +343,38 @@ String asciiCharType(int char) {
...
+A cast pattern is similar to an extractor pattern in that it checks the matched
+value against a given type. But where an extractor pattern is *refuted* if the
+value doesn't have that type, a cast pattern *throws*. Like the null-assert
+pattern, this lets you forcibly assert the expected type of some destructured
+value. This isn't useful as the outermost pattern in a declaration since you can
+always move the `as` to the initializer expression:
+
+```dart
+num n = 1;
+var i as int = n; // Instead of this...
```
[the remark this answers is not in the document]
munificent: Just removed this example.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -671,31 +634,37 @@ Field subpatterns can be in one of three forms:
...
-    variable with the same name. As a convenience, the identifier can be omitted
-    on a named field. In that case, the name is inferred from `pattern`. The
-    subpattern must be a variable pattern or cast pattern, which may be wrapped
-    in any number of null-check or null-assert patterns.
+    variable with the same name.
 
-    The field name is then inferred from the name in the variable or cast
+    As a convenience, the identifier can be omitted and inferred from `pattern`.
+    The subpattern must be a variable pattern which may be wrapped in a unary
+    pattern. The field name is then inferred from the name in the variable
```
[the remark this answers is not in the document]
munificent: More variants might be useful, but I think they quickly get unreadable. I prefer to not support:

```dart
Foo(:var point & Point(:var x,:var y))
```

At some point, it's better to just be clear and write the field name, and I think using `&` and `|` patterns has crossed that threshold.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -357,10 +343,38 @@ String asciiCharType(int char) {
...
+value. This isn't useful as the outermost pattern in a declaration since you can
+always move the `as` to the initializer expression:
+
+```dart
+num n = 1;
+var i as int = n; // Instead of this...
+var i = n as int; // ...do this.
+```
+
+But when destructuring, there is no place in the initializer to insert the cast.
```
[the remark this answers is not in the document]
munificent: Removed the example so I think it flows better with "but" now. :)

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -799,8 +767,7 @@ Add this new rule:
 ```
 patternDeclaration  ::= ( 'final' | 'var' ) outerPattern '=' expression
 
-outerPattern        ::= nullCheckPattern
-                      | parenthesizedPattern
+outerPattern        ::= parenthesizedPattern
```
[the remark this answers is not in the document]
munificent: I don't know if I had a super concrete use case in mind, but yes, that's a valid one. Back when `outerPattern` was also used by `if (var <pattern> = <expr>)`, it was also an escape hatch to let you access refutable patterns at the outermost level.

Your example makes me wonder if we should just allow an unparenthesized `&` in `outerPattern` too. Filed: https://github.com/dart-lang/language/issues/2481

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -811,11 +778,10 @@ The `outerPattern` rule defines a subset of the patterns that are allowed as the
...
 ```dart
-var [a, b] = [1, 2];                  // List.
-var {1: a} = {1: 2};                  // Map.
-var (a, b, x: x) = (1, 2, x: 3);      // Record.
-var Point(x: x, y: y) = Point(1, 2);  // Extractor.
-if (var string? = nullableString) ... // Null-check.
+var [a, b] = [1, 2];                      // List.
+var {1: a} = {1: 2};                      // Map.
+var (a, b, x: x) = (1, 2, x: 3);          // Record.
+var Point(x: x, y: y) = Point(1, 2);      // Extractor.
```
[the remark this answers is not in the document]
munificent: Done.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -848,9 +814,9 @@ forLoopParts ::=
 As with regular for-in loops, it is a compile-time error if the type of
 `expression` in a pattern-for-in loop is not assignable to `Iterable<dynamic>`.
 
-*This could potentially be extended to allow patterns in top-level variables and
-static fields but laziness makes that more complex. We could support patterns in
-instance field declarations, but constructor initializer lists make that harder.
+*We could potentially allow patterns in top-level variables and static fields
+but laziness makes that more complex. We could support patterns in instance
```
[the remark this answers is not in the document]
munificent: Done. :)

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -183,12 +183,11 @@ Before introducing each pattern in detail, here is a summary with some examples:
...
 | [Relational][relationalPattern] | `== expression`<br>`< expression` |
+| [Cast][castPattern] | `foo as String` |
 | [Null-check][nullCheckPattern] | `subpattern?` |
 | [Null-assert][nullAssertPattern] | `subpattern!` |
-| [Literal][literalPattern] | `123`, `null`, `'string'` |
-| [Constant][constantPattern] | `math.pi`, `SomeClass.constant` |
-| [Variable][variablePattern] | `foo`, `String str`, `_`, `int _` |
-| [Cast][castPattern] | `foo as String` |
+| [Constant][constantPattern] | `123`, `null`, `'string'`<br>`math.pi`, `SomeClass.constant`<br>`const Thing(1, 2)`, `const (1 + 2)` |
+| [Variable][variablePattern] | `foo`, `var bar`, `String str`, `_`, `int _` |
```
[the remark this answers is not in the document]
munificent: I went back and forth on this several times, but ultimately I think having a single pattern that covers all of these was the simplest answer.

I could split out wildcards, but I still think it's important to support typed wildcards like `int _`. Those should behave like variable patterns in that they are only allowed in irrefutable contexts if the variable isn't a downcast (which makes them useless, but still), and do a type test when used in a refutable context. The only difference between them and variable patterns is that they don't bind. If I treated `type? '_'` as a separate pattern, I'd have to specify that ambiguity between those and typed variables is resolved to a wildcard (since `_` is a valid `identifier`). Also, `var _` should be non-binding to be consistent with *non*-pattern variable declarations, assuming we can pull off that change too. It just ended up being simpler to treat them as variable patterns.

Likewise, separating out identifiers into their own patterns doesn't buy much. If we use those to refer to named constants in refutable contexts, then the specification of them has to basically copy/paste the specification for constants but only when in a refutable context.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -872,26 +838,28 @@ patternAssignment ::= outerPattern '=' expression
 assignments, but does not allow patterns to the left of a compound assignment
 operator.*
 
-In a pattern assignment, all variable and cast patterns are interpreted as
-referring to existing variables or setters. You can't declare any new variables.
-*Disallowing new variables allows pattern assignment expressions to appear
-anywhere expressions are allowed while avoiding confusion about the scope of new
-variables.*
+In a pattern assignment, all variable patterns are interpreted as referring to
+existing variables. You can't declare any new variables. *Disallowing new
```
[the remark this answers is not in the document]
munificent: > That's a confusing phrasing when "variable pattern" includes `var x`.

It does include that, but that specific gets ruled out later when talking about context.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -872,26 +838,28 @@ patternAssignment ::= outerPattern '=' expression
...
+variables allows pattern assignment expressions to appear anywhere expressions
+are allowed while avoiding confusion about the scope of new variables.*
 
 It is a compile-time error if:
 
-*   A variable pattern has a type annotation. *Only simple identifiers are
-    allowed since you aren't declaring a new variable.*
-
 *   An identifier in a variable or cast pattern does not resolve to a non-final
```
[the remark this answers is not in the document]
munificent: I think the way it's specified makes sense here given the other spec choices where variable pattern also includes simple identifiers.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1081,11 +1049,11 @@ expression.
 ```dart
 main() {
   switch (some(extremely, long, expression, here)) {
-    case Some(Quite(var long, var pattern)) => expression();
+    case Some(Quite(:var long, :var pattern)) => expression();
   };
```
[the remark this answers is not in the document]
munificent: That's an interesting point. Lambdas already do too (modulo `await`/`yield`/`break`/`continue`). But, yes, switch expressions give you full expression locals even inside loops, async functions, and generators.

Whether that suggests we should do `let` expressions or some other syntax is a separate discussion. :)

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -994,6 +959,9 @@ switch (1) {
 }
 ```
 
+To have an empty case that does *not* fallthrough, use `break;` for its body as
+you would today.
+
 ### Switch expression
```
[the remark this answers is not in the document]
munificent: Probably. Filed: https://github.com/dart-lang/language/issues/2482

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1133,35 +1101,153 @@ if (var [int x, int y] = json) {
...
+
+    *Logical-or patterns are refutable because there is no point in using one
+    with an irrefutable left operand. We could make null-check patterns
+    irrefutable if `V` is assignable to its static type, but whenever that is
+    true the pattern does nothing useful since its only behavior is a type
+    test.*
+
+    *The remaining patterns are allowed syntactically to appear in a refutable
+    context. Patterns that do type tests like variables and lists produce a
+    compile-time error when used in an irrefutable context if the matched value
```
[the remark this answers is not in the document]
munificent: Good catch. :)

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1133,35 +1101,153 @@ if (var [int x, int y] = json) {
...
+    *This program prints "no match" and not "match 2".*
+
+*   A simple identifier in a matching context named `_` is treated as a wildcard
+    variable pattern. *A bare `_` is always treated as a wildcard regardless of
+    context, even though other variables in matching contexts require a marker.*
+
+    ```dart
+    // OK:
+    switch (triple) {
+      case [_, y, _]: print('The middle element is $y');
```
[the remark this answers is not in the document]
munificent: Oops, forgot to update this to the new syntax. :) Should be `var y`. Fixed.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1133,35 +1101,153 @@ if (var [int x, int y] = json) {
...
+    }
+    ```
+
+    *You can also use `var _` or `final _` to write a wildcard in a matching
+    context because it would require additional specification to explicitly
+    forbid it, but doing so is discouraged.*
+
+*In short, don't use refutable patterns in places that don't do control flow.
```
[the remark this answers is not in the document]
munificent: Done.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1133,35 +1101,153 @@ if (var [int x, int y] = json) {
...
+Use simple identifiers (optionally with type annotations) to declare variables
+in pattern declarations. Use simple identifiers to assign to variables in
+pattern assignments. Use explicitly marked identifiers to declare variables in
+cases. Use `_` anywhere for a wildcard.*
```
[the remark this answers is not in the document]
munificent: Done (well, "`case` patterns"`)

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1314,12 +1402,15 @@ The context type schema for a pattern `p` is:
...
-    the pattern's constant value expression.
+*   **Constant**: The context type schema is the static type of the pattern's
+    value.
 
 *   **Variable**:
 
-    1.  If `p` has no type annotation, the context type schema is `?`.
+    1.  In an assignment context, the context type schema is the static type of
+        the variable that `p` resolves to.
+
```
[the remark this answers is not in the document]
munificent: Yes. A `variablePattern` always refers to a variable and never a constant. A `constantPattern` always refers to a constant. Those two kinds of patterns are syntactically ambiguous since a short identifier is valid for both, but we resolve that syntactically using the surrounding pattern context which is known at parse time (which is why I moved the "pattern context" section up before the static semantics).

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1404,22 +1496,30 @@ To type check a pattern `p` being matched against a value of type `M`:
 *   **Relational**: If the operator is a comparison (`<`, `<=`, `>`, or `>=`),
     then it is a compile-time error if `M` does not define that operator, if the
     type of the constant in the relational pattern is not assignable to the
-    operator's parameter type, or if the operator's return type is not `bool` or
-    `dynamic`. *The `==` and `!=` operators are valid for all pairs of types.*
+    operator's parameter type, or if the operator's return type is not
+    assignable to `bool`. *The `==` and `!=` operators are valid for all pairs
+    of types.*
```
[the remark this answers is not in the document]
munificent: Nice catch. Done.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1404,22 +1496,30 @@ To type check a pattern `p` being matched against a value of type `M`:
...
+
+*   **Cast**:
+
+    1.  Resolve the type name to a type `X`. It is a compile-time error if
+        the name does not refer to a type.
+
+        **TODO: Should we infer type arguments on `X`?**
```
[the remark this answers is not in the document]
munificent: SGTM.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1264,14 +1350,16 @@ To orchestrate this, type inference on patterns proceeds in three phases:
     holes in the type schema. When that completes, we now have a full static
     type for the pattern and all of its subpatterns.
```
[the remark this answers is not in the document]
munificent: Exactly right, yes.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1537,130 +1639,46 @@ To type check a pattern `p` being matched against a value of type `M`:
...
-
-In the first function, the `(a, b)` pattern will always successfully destructure
-the record because `obj` is known to be a record type of the right shape. But in
-the second function, `obj` may fail to match because the value may not be a
-record. *This implies that we can't determine whether a pattern in a variable
-declaration is incorrectly refutable until after type checking.*
+It is a compile-time error if:
 
-Refutability of a pattern `p` matching a value of type `V` is:
+*   The type of an expression in a guard clause is not `bool` or `dynamic`.
```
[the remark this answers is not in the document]
munificent: Oops, yes. Good catch. :)

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1537,130 +1639,46 @@ To type check a pattern `p` being matched against a value of type `M`:
...
 
-*   **Logical-and**, **parenthesized**, **null-assert**, or **cast**:
-    Irrefutable if and only if all subpatterns are irrefutable.
+*   `p` is in an irrefutable context, it has a required type `T`, and `M` is not
+    assignable to `T`. *Destructuring and variable patterns can only be used in
+    declarations and assignments if we can statically tell that the
+    destructuring and variable binding won't fail.*
```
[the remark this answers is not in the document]
munificent: "fail" here means "fail to match" and not "complete successfully". Reworded and added a parenthetical to clarify.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1537,130 +1639,46 @@ To type check a pattern `p` being matched against a value of type `M`:
...
 Patterns often exist to introduce new variable bindings. A "wildcard" identifier
-named `_` in a variable or cast pattern never introduces a binding.
-
+named `_` in a variable pattern never introduces a binding.
 The variables a patterns binds depend on what kind of pattern it is:
 
 *   **Logical-or**: Does not introduce variables but may contain subpatterns
-    that do. If it a compile-time error if the two subpatterns do not introduce
+    that do. It is a compile-time error if the two subpatterns do not introduce
     the same variables with the same names and types.
```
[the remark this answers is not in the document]
munificent: Yeah, I agree this is definitely too vague. For now, I'm going to just put a "TODO" here, because I don't want to bother specifying something in detail that will then likely get replaced when I address https://github.com/dart-lang/language/issues/2473.

working/0546-patterns/patterns-feature-specification.md
munificent: I think the "Variables and scope" section covers this pretty well, but if you think it's better to explain it here I could do that.

working/0546-patterns/patterns-feature-specification.md
munificent: Done.

working/0546-patterns/patterns-feature-specification.md
munificent: I like that. Done.

working/0546-patterns/patterns-feature-specification.md
munificent: Done.

working/0546-patterns/patterns-feature-specification.md
munificent: Done (here and elsewhere).

working/0546-patterns/patterns-feature-specification.md
munificent: The sentence isn't clear, but the intent is that "the variable that `p` refers to" can also mean "the variable that `p` itself is declaring". So it can occur in a refutable context in which case it is a type test:

```dart
if ((123 as Object) case String s) { ... }
```

I'll reword this to make it clearer.

working/0546-patterns/patterns-feature-specification.md
munificent: Good point. I went ahead and consolidated all of the breaking stuff into a new section.

working/0546-patterns/patterns-feature-specification.md
munificent: I think so too, but I really appreciate the feedback. We went back and forth on this *so* much. :)

working/0546-patterns/patterns-feature-specification.md
munificent: Haha, good question! I'll defer to your expertise. Done. :)

working/0546-patterns/patterns-feature-specification.md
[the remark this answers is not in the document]
lrhn: If people can read it and understand which identifier means what where, then it works. Ship it.

It'll be Erik's problem to do the final formal specification anyway :smiling_imp:

working/0546-patterns/patterns-feature-specification.md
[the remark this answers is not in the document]
lrhn: ACK.

working/0546-patterns/patterns-feature-specification.md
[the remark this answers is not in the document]
lrhn: We need to be very clear on the scoping of such variables. Each evaluation must introduce a new binding since the variable can be final. That's *probably* ensured by putting the variable in the nearest enclosing scope, because there is no way to evaluate the same expression twice (aka. loop) without introducing a new scope per iteration. We just need to make sure we haven't forgotten (or introduced) any operation where that's not true.

For switch cases, we need a scope per `casehead`, because each can introduce independent final variables that can be read in the `when` clause, whether the case is then accepted or not.

Not a problem, we just need to be careful.

working/0546-patterns/patterns-feature-specification.md
lrhn: Do what we'd do on a plain assignment, `x = e`, for the same variable.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1374,12 +1467,15 @@ The context type schema for a pattern `p` is:
...
-*   **Literal** or **constant**: The context type schema is the static type of
-    the pattern's constant value expression.
+*   **Constant**: The context type schema is the static type of the pattern's
+    value.
 
 *   **Variable**:
 
-    1.  If `p` has no type annotation, the context type schema is `?`.
+    1.  In an assignment context, the context type schema is the static type of
+        the variable that `p` resolves to.
```
lrhn, line 1476: (Should we disallow assigning the same variable more than once in the same assignment pattern?)

working/0546-patterns/patterns-feature-specification.md
lrhn: \*\* goes to read "Variables and scope"\*\*

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1600,44 +1718,51 @@ To type check a pattern `p` being matched against a value of type `M`:
...
-    not final.*
+[2473]: https://github.com/dart-lang/language/issues/2473
 
-The scope where a pattern's variables are declared depends on the construct
-that contains the pattern:
+The scope where a pattern's variables are declared and available for use depends
+on the construct that contains the pattern:
 
 *   **Local pattern variable declaration**: The rest of the block following
     the declaration.
```
lrhn, line 1768: This is new. The current behavior is that the scope is the entire block, and access only allowed in the rest of the block.
It matters for 
```dart
var x = 42;
{
  var y = x;  // <- invalid access to inner x
  var x = y;
}
```
We should probably do the same thing for pattern and non-pattern declarations.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1600,44 +1718,51 @@ To type check a pattern `p` being matched against a value of type `M`:
...
 
-*   **Variable** or **cast**: May contain type argument patterns. Introduces a
-    variable whose name is the pattern's identifier. The variable is final if
-    the surrounding `patternDeclaration` has a `final` modifier. *Variables in
-    switch cases don't occur inside pattern variable declarations and thus are
-    not final.*
+[2473]: https://github.com/dart-lang/language/issues/2473
 
-The scope where a pattern's variables are declared depends on the construct
-that contains the pattern:
```
lrhn, line 1640: The scope where a local variable is declared, and where it's available for use, are not the same thing in the current specification. We need to define both.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1646,7 +1771,7 @@ that contains the pattern:
 *   **Switch statement case**: The guard clause and the statements of the
     subsequent non-empty case body.
```
lrhn, line 1772: If multiple `casehead`s can reach the same body, then we need some way to merge scopes.

Take:
```dart
 case Point(:int x, :int y):
 case (:int x, :int y):
    x += 1;
    y += 1;
    print(x + y);
```
Both branches introduce the same names with the same types. One of them reaches the body, and defines the variables that are modified. But *both* sets of variables can be in play:

```dart
late int Function() next;
...
 case Point(:int x, :int y) when (next = () => x++) != 0 && x < y:
 case Point(:int x, :int y): 
    print([x++, y++, next());
```
If given `Point(x:10, y: 5)` the first case will be rejected in the `when` clause, but still capture the `x` variable of that.
Then the next case succeeds, binding a new `x` and `y`.
It should print `[10, 5, 10]`.


Is this just an argument against allowing multiple `casehead`s for the same body? I think it is!

You can still capture variables from *other* cases, which were rejected in the `when` clauses, but those will definitely be *different* variables, but you only get one pattern scope and one `when` clause per body.

```dart
late int Function() next;
...
 case Point(:int x, :int y) when (next = () => x++) != 0 && false:
    print("not today");
 case Point(:int x, :int y): 
    print([x++, y++, next());
```
This still works and binds the variable of the first (matching!) pattern, but there is no way for those variables to ever flow into the same body as the second `case`.

We can allow `case c1: case c2: case c3: body` as a backwards compatible way to write `case c1 | c2 | c3: body`, but you can't add bindings or `when` clauses to any of those cases. To do that, you have to combine them to one pattern using `|`.

On the other hand, that would prohibit:
```dart
case Value(:var x) when x >= 0:
case NegatedValue(:var x) when x < 0:
  same body;
```
if that's something you want to do (multiple independent `when`s leading to the same body).

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1646,7 +1771,7 @@ that contains the pattern:
...
 *   **Switch expression case**: The guard clause and the case expression.
-*   **Pattern-if statement**: The then statement.
+*   **If-case statement**: The guard clause and then statement.
 
 Multiple switch case patterns may share the same variable scope if their case
 bodies are empty:
```
lrhn, line 1777: Nope. Just thought about this, and it *cannot* work.

```dart
late int Function() otther;
...
switch([1, 2]) {
  case [final int a, final int b] when (other = ()=>a) == 0:
  case [final int b, final int a]:
     print(a + b + other(); // 2 + 1 + 1
```

The scope of each `casehead` must be independent, because it can bind *final* variables to values even if then refuted by a `when` clause. The same final variable cannot be initialized more than once. 
And since a `when` clause is an expression, it can capture the final variable, and must see the correct value.

So each `case` pattern introduces its own bindings.

We then have to make *one* of those sets of bindings the current binding for the body, when we execute/evaluate that body. That's a little weird, but not much weirder than what we do for `for(var x = 1;x < 10;x++)` iterations. (Which isn't really saying much. That *is* weird.)

See more above.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1668,7 +1793,7 @@ However, it is a compile-time error if all switch case patterns that share a
 body do not all define the exact same variables with the exact same types.
 
 *Aside from this special case, note that since all variables declared by a
-pattern and its subpattern go into the same scope, it is an error if two
+pattern and its subpatterns go into the same scope, it is an error if two
 subpatterns declare a variable with the same name, unless the name is `_`.*
```
lrhn, line 1797: Even if those are sub-patterns of a `|` pattern?

working/0546-patterns/patterns-feature-specification.md
munificent: Yes, that's covered in the initial syntax section for pattern assignments.

PR #2487, opened 2022-09-14
Move records proposal to accepted

leafpetersen:
This has been accepted, moving to its own directory under future_releases.  cc @munificent


kallentu (approved):

bernaferrari:
Kind of unrelated, but, I guess you can move enhanced enum from future-releases to released?

PR #2493, opened 2022-09-15
[patterns] A handful of mostly small changes to patterns and records:

### LINKED ISSUE #2457: Do relational patterns/extractor patterns allow extension members
A pattern like `Foo(bar: < 4)` reads a `bar` getter and calls a `<` operator on the result.

Can those be extension members.

Say:
```dart
class Foo {
  String _value;
  Foo(String value) : _value = value;
  String toString() => value;
}

extension on Foo {
  String get bar => this.toString();
}
extension on String {
  bool operator< (String other) => this.compareTo(other) < 0;
}

// ...

switch (foo) {
  case Foo(bar: < "abc"): ...
```

Will this code work?

(I'd hope so, if we consider pattern matching as mainly syntactic sugar, and desugaring it at the place of writing would call the extension members. I think we should make sure to say this explicitly in the specification.)

### LINKED ISSUE #2461: [Patterns] Do relational patterns do a downcast?
The dynamic semantics of relational patterns in the pattern proposal says:

```
Relational:

Evaluate the right-hand constant expression to c.

A == c pattern matches if v == c evaluates to true. This takes into account the built-in semantics that null is only equal to null.

A != c pattern matches if v == e evaluates to false. This takes into account the built-in semantics that null is not equal to anything but null.

For any other operator, the pattern matches if calling the operator method of the same name on the matched value, with c as the argument returns true.
```

I believe that this implies that if calling the method in a relational pattern returns a non-boolean (which can happen, I think, if the return type is `dynamic`), then it is simply a failed match (rather than a runtime error).  Is this the intended semantics?  This is not consistent with how conditionals currently behave, just as a point of comparison.

```dart
class A {
  dynamic operator <(int other) => "hello";
}

void main() {
  A a = A();
  if (a < 3) {} // Throws a runtime cast error
  switch (a) {
    case < 3: print("hello"; // Does this throw a cast error, or just fail to match this case?
    default: print("world");
  
}
```

### LINKED ISSUE #2466: Is there any way to control the ordering of destructuring.
We have patterns with side effects, specifically the `!` and `as SomeType` patterns.
The current specification does not specify the order that fields of a record pattern are checked in.

If I know I have a pair of values, where the value of the first element determines the type of the second, can I write it as:
```dart
switch (pair) {
  case ("s", s as String): ...
  case ("i", i as int): ...
  ...
}
```
That seems like the immediate approach that someone might try, but if we don't actually specify the ordering of the tests, it's possible for the pattern matching algorithm to test `s as String` first.

I believe the `as`-pattern is currently defined only applies to a simple variable pattern and only does the cast when binding, after successfully matching, and the cast pattern is irrefutable, but I believe the cast pattern might need to be more general. Also the null-assert  `!` pattern is not defined the same way, so
```dart
  case ("int", int x!): 
```
would be able to throw on the scrutinee `("string", "value")`, even though the tag wouldn't match anyway.

Do we want to ensure a predictable order of testing? (Probably source order for the pattern.)
Or should we specify that the order is *not* guaranteed?

(For a situation where you would want/need to do a cast *before* binding to a single variable, consider:
```dart
enum Kind { int, string, bool, point };
...
switch (taggedValue) {
  case (Kind.int, x as int): ...
  case (Kind.string, s as String): ...
  case (Kind.bool, b as bool):  ...
  case (Kind.point, (:x, :y) as (int, int)): ...
} // exhaustive!
```
Here we don't want to just do `(:int x, :int y)` for the second element of a `point`, because then it's refutable, and then the switch is not exhaustive. So, to ensure exhaustiveness, we want to *cast* the second field of the `taggedValue` to `(int, int)` *before* doing an irrefutable pattern match on the result.)

### LINKED ISSUE #2474: Should switch statements be exhaustive?
The patterns proposal is currently silent about compile errors related to exhaustiveness checking on patterns, though it does talk about exhaustiveness checking as a *concept*. That needs to be filled in.

Switch *expressions* need to be exhaustive because otherwise there may not be a value that the switch expression can evaluate to. We could silently throw a runtime exception if no cases match, but I don't think anyone wants that.

It's less clear what the right answer for switch statements is. Currently, the language only reports static warnings for non-exhaustive cases when switching on an enum type. For all other types, it is happy to let you have unmatched values and no default case.

For switch statements *on sealed families*, I think users want it to be at least a warning if the cases are not exhaustive.

I suspect that switching on other unsealed types would be very annoying if we required users to write a default clause. It would certainly be extremely breaking.

Based on that, I propose:

* It is a compile-time *error* for a switch *expression* to not be exhaustive, regardless of the type being switched on. In practice, this means that for unsealed types, you'll need a `default` case, untyped variable, or wildcard that matches everything.
* It is a compile-time *warning* for a switch *statement* to not be exhaustive *when the static type of the matched value is a "sealed type".*
* For switch statements on any other type, it's fine for it to not be exhaustive.

A "sealed type" is:

* Bool
* Null
* Any enum type
* Any type marked as sealed using an as-yet-unspecified way
* A nullable type whose underlying type is sealed
* A record type whose fields are all sealed types

*TODO: Figure out how generic type arguments come into play.*

How does that sound?

### LINKED ISSUE #2486: Allow both "final" and a type in pattern variables
Elsewhere in the language, in variable declarations you can specify both `final` and a type annotation if you want the variable to have a specific type and not be mutable. The patterns proposal currently treats `final` and the type annotation is mutually exclusive. I don't think there's any strong reason for that and we should probably allow:

```dart
if (foo case final int x) ...
```

### LINKED ISSUE #2491: Naming request for a pattern specification
Current:
```
primary               ::= // Existing productions...
                        | switchExpression

switchExpression      ::= 'switch' '(' expression ')' '{'
                          switchExpressionCase* defaultExpressionCase? '}'
switchExpressionCase  ::= caseHead '=>' expression ';'
defaultExpressionCase ::= 'default' '=>' expression ';'
```
Can we rename `defaultExpressionCase` to `switchExpressionDefault`?

And similarly, renames for `switchStatement`. Currently:
```
switchStatement ::= 'switch' '(' expression ')' '{' switchCase* defaultCase? '}'
switchCase      ::= label* caseHead ':' statements
caseHead        ::= 'case' pattern ( 'when' expression )?
```
Proposal:
```
switchStatement ::= 'switch' '(' expression ')' '{' switchStatementCase* switchStatementDefault? '}'
switchStatementCase      ::= label* caseHead ':' statements
caseHead        ::= 'case' pattern ( 'when' expression )?
switchStatementDefault      ::= label* 'default' ':' statements
```

We would like to try naming in the analyzer close to what the specification uses, and sometimes the naming in the specification feels suboptimal. We used to say "oh, well" and name it as we like in the analyzer, but now we try to do the other way.

munificent:
- Clarify that patterns can call extension members. Fix #2457.
- Non-boolean results throw in relational patterns. Fix #2461.
- Maps and extractors are evaluated in source order. Fix #2466.
- Specify non-exhaustive switch errors and warnings. Fix #2474.
- Allow `final` before type annotated variables. Fix #2486.
- Rename some grammars to align with Analyzer AST names. Fix #2491.

### DIFF
--- a/accepted/future-releases/records/records-feature-specification.md
+++ b/accepted/future-releases/records/records-feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: Accepted
 
-Version 1.8 (see [CHANGELOG](#CHANGELOG) at end)
+Version 1.9 (see [CHANGELOG](#CHANGELOG) at end)
 
 ## Motivation
 
@@ -498,6 +498,18 @@ The implementation of `hashCode` follows this. The hash code returned should
 depend on the field values such that two records that compare equal must have
 the same hash code.
 
+### Primitive equality
+
+A record object has a primitive `==` operator if all of its field have primitive
+`==` operators.
+
+*Note that this is a dynamic property of a record object, not a static property
+of its type. Since primitive equality only comes into play in constants, the
+compiler can see the actual field values for a relevant record at compile time
+because it has the actual constant record value with all of its constant fields.
+This means records can be used in constant sets and maps keys, but only when the
+records' fields could be as well.*
+
 ### Identity
 
 We expect records to often be used for multiple return values. In that case, and
@@ -565,9 +577,13 @@ covariant in their field types.
 
 ## CHANGELOG
 
+### 1.9
+
+- Specify that a record has a primitive `==` when its fields all do.
+
 ### 1.8
 
-- Move to accepted
+- Move to `accepted/`.
 
 ### 1.7
 
--- a/working/0546-patterns/patterns-feature-specification.md
+++ b/working/0546-patterns/patterns-feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: In progress
 
-Version 2.6 (see [CHANGELOG](#CHANGELOG) at end)
+Version 2.7 (see [CHANGELOG](#CHANGELOG) at end)
 
 Note: This proposal is broken into a couple of separate documents. See also
 [records][] and [exhaustiveness][].
@@ -483,7 +483,7 @@ expression.
 ### Variable pattern
 
 ```
-variablePattern ::= ( 'var' | 'final' | type )? identifier
+variablePattern ::= ( 'var' | 'final' | 'final'? type )? identifier
 ```
 
 A variable pattern binds the matched value to a new variable. These usually
@@ -732,13 +732,13 @@ against the initializer's value.
 Add this new rule:
 
 ```
-patternDeclaration  ::= ( 'final' | 'var' ) outerPattern '=' expression
+patternVariableDeclaration  ::= ( 'final' | 'var' ) outerPattern '=' expression
 
-outerPattern        ::= parenthesizedPattern
-                      | listPattern
-                      | mapPattern
-                      | recordPattern
-                      | extractorPattern
+outerPattern                ::= parenthesizedPattern
+                              | listPattern
+                              | mapPattern
+                              | recordPattern
+                              | extractorPattern
 ```
 
 The `outerPattern` rule defines a subset of the patterns that are allowed as the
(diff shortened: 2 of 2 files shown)


leafpetersen (approved):
LGTM, thanks.

Issue #2179, opened 2022-03-30
Should promotion take sealed types into account?


munificent:
In order to support [exhaustiveness checks](https://github.com/dart-lang/language/blob/master/working/0546-patterns/exhaustiveness.md) in [algebraic datatype-style code](https://github.com/dart-lang/language/blob/master/working/0546-patterns/patterns-feature-specification.md#algebraic-datatypes) in [pattern matching](https://github.com/dart-lang/language/blob/master/working/0546-patterns/patterns-feature-specification.md), we want some notion of [sealed types](https://github.com/dart-lang/language/blob/master/working/0546-patterns/exhaustiveness.md#sealed-types).

That raises the question of whether sealed types should come into play with type promotion. For example:

```dart
sealed class Foo {} // Only subtypes are Bar and Baz
class Bar extends Foo {}
class Baz extends Foo {}

test(Foo foo) {
  if (foo is Bar) return;
  // Promote foo to Baz here?
}
```

I suspect that the answer is "yes", since this seems practically useful.

jakemac53:
One concern with this is it would make it more likely for users to be broken by a package adding a new sub-type of a sealed class, because the promotion could no longer work where it did before.

This was already breaking anywhere exhaustiveness checking applies, so maybe not a large issue, but it might be easier to accidentally rely on the exhaustiveness checking in this case than say a switch statement.

munificent:
That's true, but that is something a package author is deliberately signing themselves up for when they mark a supertype as sealed. If we treated "sealed" as a distinct concept from "final" (i.e. preventing subclassing/implementing/etc.) then the only reason to seal a class is to opt in to exhaustiveness checking. And once you do that, adding a new subtype is always a potentially breaking change, type promotion or not.

leafpetersen:
> There are other reasons as well, you might not support arbitrary subtypes of your class for a variety of reasons.

`sealed` as used in this context allows essentially arbitrary indirect subtypes.

jakemac53:
Of course on the flipside, this means you can implement `Dog` as a mock, so there are testing advantages to that approach.

munificent:
> So is this allowed?
> 
> ```dart
> sealed class Animal {}
> class Dog extends Animal {}
> ```
> 
> And then in some other library:
> 
> ```dart
> class Husky implements Dog {}
> ```
> 
[...]

PR #2494, opened 2022-09-16
Patterns caching

munificent:


working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2159,17 +2209,233 @@ To match a pattern `p` against a value `v`:
...
+        method resolved to, `op` is the name of the operator and `arg` is the
+        right operand value.
+
+*   **Cast**, **null-check**, **null-assert**, or **parenthesized**:
+
+    1.  Bind invocations in the subpattern using target `i`.
+
+*   **Constant**:
+
+    1.  Bind `i : ("==", [arg])` to the `==` method invocation where`arg` is the
```
leafpetersen, line 2377: I don't think this is correct, because it equates `case ==a` and `case a`, right?  You need to somehow distinguish the two cases (because they call different methods).

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2159,17 +2209,233 @@ To match a pattern `p` against a value `v`:
...
+    invocation key tree and represents the matched value itself.*
+
+2.  For each pattern `p` in `s` with parent invocation `i`, bind invocation keys
+    to it and its subpatterns using the following procedure:
+
+To bind invocation keys in a pattern `p` using parent invocation `i`:
+
+*   **Logical-or** or **logical-and**:
+
+    1.  Bind invocations in the left and right subpatterns using target `i`.
```
leafpetersen, line 2356: You switch to using "target" here instead of "parent invocation".  I think (but could be wrong), you mean the same thing?  If not, probably explain?  If so, maybe just use one term?

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2159,17 +2209,233 @@ To match a pattern `p` against a value `v`:
...
+*   **Extractor**:
+
+    1.  For each field `f` in `p`:
+
+        1.  Resolve the getter name to a member.
+
+        2.  If it resolves to an instance method, then let `f` be `i : (field,
+            [])` where `field` is the name of the getter.
+
+        3.  Else it resolves to an extension method. Let `f` be `i : type(field,
```
leafpetersen, line 2433: There's a subtlety here around what `type` is that should probably be specified.  I *think* (but am not sure), that what it should be is actually the name + type arguments of the extension declaration.  The resolved `on` type may not be sufficient, though maybe the fact that we can't use the explicit invocation syntax (`E(receiver).m`) here makes it sufficient? 
 An alternative would be the type of the receiver, but that may conflate more things than necessary.

We clearly need to make sure that things with different resolved `on` types are different.  Example:

```dart
extension E<T> on  List<T> {
  int get foo => (T == int) ? 3 : 4;
}

void test(Object x) {
  switch (x) {
    case List<int>(:var foo): ...
    case List<double>(:var foo):...
}
```

So it's not enough to use just `E` as the key.  Using `E<double>` and `E<int>` as the keys would work.

Using the type of the receiver is probably correct, because I think for a given name `foo`, two invocations of it with the same receiver type in the same import environment *must* always result in the same resolution.  This will be coarser than necessary though:

```dart
extension E on  num {
  int get foo => 3;
}

void test(Object x) {
  switch (x) {
    case int(:var foo): ...
    case double(:var foo):...
}
```

These two invocations are demonstrably equal, but if we use the receiver types as keys, they will be memoized separately.

Using the resolved `on` type would not be correct if we could somehow get an explicit invocation.  Consider this extension method declaration:

```dart
extension E<T> on  num {
  int get foo => (T == int) ? 3 : 4;
}
```

Clearly `E<int>(receiver).foo` and `E<double>(receiver).foo` will produce different results.  But I can't think offhand of any way to get this without using the explicit syntax, and I don't think we can use the explicit syntax here.

On balance, I *think* that what we want here is the extension name + type arguments as the key.  @lrhn and @eernstg might want to check my reasoning here though.

leafpetersen (approved):
Nice, thanks!  Raised a couple of questions, but otherwise LGTM.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2159,17 +2209,233 @@ To match a pattern `p` against a value `v`:
...
+2.  At runtime, whenever the runtime semantics say to call a member or access a
+    record field, if a previous call or access with that same invocation key has
+    already been evaluated, we reuse the result.
+
+3.  Otherwise, we invoke the member or field access now and associate the result
+    with that invocation key for future calls.
+
+Let an *invocation key* comprise:
+
+*   A possibly absent parent invocation key.
```
lrhn, line 2318: The word "parent" confuses me. Does it mean "receiver" (the target of a named lookup)?

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2159,17 +2209,233 @@ To match a pattern `p` against a value `v`:
...
+*   A possibly absent extension type. If the invocation represents an extension
```
lrhn, line 2319: "Extension type" also confuses me a little, because extensions are not types.
I assume it covers an extension declaration and the inferred type parameters to that extension. Could it be called "extension instantiation"?

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2159,17 +2209,233 @@ To match a pattern `p` against a value `v`:
...
+
+To bind invocation keys in a pattern `p` using parent invocation `i`:
+
+*   **Logical-or** or **logical-and**:
+
+    1.  Bind invocations in the left and right subpatterns using target `i`.
+
+*   **Relational**:
+
+    1.  Resolve the relational operator to a member.
```
lrhn, line 2360: This has to be explained in some place. There are a number of cases:
* The static type of `i`, *S*, is `dynamic` or `Never`. In that case, no lookup is performed.
* *S* is another type whose interface declares an `op` operator.
* *S* does not declare an `op` operator, but extension method resolution finds a solution for `op` on *S*, *E*\<*T*<sub>1</sub>,...,*T<sub>n</sub>*\>.

The first two cases is what the item below covers, but I'd not say that the `dynamic` case "resolves to an instance method".
(We could say that, `dynamic`/`Never` have all instance methods, but that gets weird when we start looking at the type of those methods.)

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2159,17 +2209,233 @@ To match a pattern `p` against a value `v`:
...
+
+    2.  If it resolves to an instance method, then bind `i : (op, [arg])` to
+        the operator method invocation where `op` is the name of the operator
+        and `arg` is the right operand value.
+
+    3.  Else it resolves to an extension method. Bind `i : type(op, [arg])` to
+        the operator method invocation where `type` is the type the extension
```
lrhn, line 2367: Not a type. I don't remember if we have a name for an "extension instantiation", but that phrase can be used.

working/0546-patterns/patterns-feature-specification.md
lrhn: Agree. I think you need a "base" invocation key for each constant which occurs, and the invocation has to be on that.

Which means that the *argument* is not necessarily a constant, but can be an invocation key.

So: An invocation key is either a four tuple:

   InvocationKey<sub>opt</sub> &times; ExtensionInstantiation<sub>opt</sub> &times; memberName &times; List\<InvocationKey\> 

*or* a constant value (which is equivalent to the same constant value.)
  
(In practice, the argument list is always zero or one argument, since all we invoke are binary operators and getters).

Maybe:
```abnf
invocationKey ::= 
      constant
    | this
    | invocationKey? : extensionInstantiation? ( memberName, invocationKey*)
```

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2159,17 +2209,233 @@ To match a pattern `p` against a value `v`:
...
+*   A possibly absent extension type. If the invocation represents an extension
+    member call, this tracks the type the call was resolved to.
+*   A member name.
+*   A possibly empty list of argument constant values.
+
+Two invocation keys are equivalent if and only if all of these are true:
+
+*   They both have parent invocation keys and the keys are equivalent or
+    neither of them have parent invocation keys.
+*   The extension types refer to the same type or are both absent.
```
lrhn, line 2328: -> refer to the same extension declaration with the same type parameters.

(Caveat: The usual "when are types the same"? Whatever we do in other places where the types have to be the same.)

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2159,17 +2209,233 @@ To match a pattern `p` against a value `v`:
...
+
+        2.  Bind `e` to the field accessor for this field.
+
+        3.  Bind invocations in the field subpattern using target `e`.
+
+*   **Extractor**:
+
+    1.  For each field `f` in `p`:
+
+        1.  Resolve the getter name to a member.
```
lrhn, line 2428: Same comment as above: At least need to say that it resolves against static type, and what it means to resolve against `dynamic`/`Never`

working/0546-patterns/patterns-feature-specification.md
lrhn: I've already commented on this above. You pretty certainly need to remember the entire corresponding explicit extension invocation of `Ext<T>(x).foo`, which means the chosen extension declaration plus the inferred type arguments.

The type arguments are inferred from the static type of the receiver, but it's more precise to use the inferred type arguments. Take:
```dart
extension F<T> on Iterable<T> {
  (T, Type)? get firstAndTypeOrNull => this.isEmpty ? null : (this.first, T);
}
```
It behaves differently on `List<num>` and `List<int>`, so the inferred type must be included.
It resolves to the extension `F<int>` on both `List<int>` and `Iterable<int>`, so the entire receiver type is not important. 

The most precise distinguisher for behavior is the `F<int>` extension instantiation.


(Secondary note: It's slightly worrying that you *cannot* do an explicit extension instantiation.)

lrhn (approved):
Nits on details, overall the approach seems sound.

The equivalent `late` variable behavior is very illustrative.

(An entire program, written using `late`, it feels like programming Mona/Miranda again!)

working/0546-patterns/patterns-feature-specification.md
leafpetersen: I can't find where you commented above, and I can't tell whether you're agreeing with me or not.  I think you are?  That is, that it's sufficient to use the extension name + type arguments as the token?

working/0546-patterns/patterns-feature-specification.md
munificent: I think Leaf is right. Done.

working/0546-patterns/patterns-feature-specification.md
munificent: Oops, yes. I went back and forth on the terminology and left it inconsistent. Fixed.

accepted/future-releases/records/records-feature-specification.md
```diff
@@ -498,6 +498,18 @@ The implementation of `hashCode` follows this. The hash code returned should
 depend on the field values such that two records that compare equal must have
 the same hash code.
 
+### Primitive equality
+
+A record object has a primitive `==` operator if all of its field have primitive
+`==` operators.
```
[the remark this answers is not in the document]
munificent: I checked, actually. :)

It says:

```latex
In order to specify these constraints just once we introduce the notion of a
% Neither \syntax nor \lit works, so we fall back to `\code{==}'.
\IndexCustom{primitive operator `\code{==}'}{%
  operator `\code{==}'!primitive}:

\begin{itemize}
\item Every instance of type \code{int} and \code{String}
  has a primitive operator \lit{==}.
```

accepted/future-releases/records/records-feature-specification.md
```diff
@@ -498,6 +498,18 @@ The implementation of `hashCode` follows this. The hash code returned should
...
+
+*Note that this is a dynamic property of a record object, not a static property
+of its type. Since primitive equality only comes into play in constants, the
```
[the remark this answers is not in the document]
munificent: Fixed.

accepted/future-releases/records/records-feature-specification.md
```diff
@@ -498,6 +498,18 @@ The implementation of `hashCode` follows this. The hash code returned should
...
+compiler can see the actual field values for a relevant record at compile time
+because it has the actual constant record value with all of its constant fields.
+This means records can be used in constant sets and maps keys, but only when the
+records' fields could be as well.*
```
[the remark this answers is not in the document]
munificent: Too many possessives. :) Reworded.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -483,7 +483,7 @@ expression.
 ### Variable pattern
 
 ```
-variablePattern ::= ( 'var' | 'final' | type )? identifier
+variablePattern ::= ( 'var' | 'final' | 'final'? type )? identifier
```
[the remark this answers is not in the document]
munificent: Oh, dear. I think when I merged the changes from https://github.com/dart-lang/language/pull/2493 into this PR (since that PR landed first), they showed up here in the review. But I guess that works since it means you can review those changes here now. :)

> Do we need to say, in prose somewhere, that you are not allowed to use the `final` in a declaration like `var (x, final int y) = ....` You can use the `int y`, but not the `final int y`.

That's covered in the pattern context section, which says:

> It is a compile-time error if a variable pattern in a declaration context is marked with `var` or `final`.

I reworded a bit to clarify. I think it works OK now.

working/0546-patterns/patterns-feature-specification.md
munificent: Ah, good catch! Fixed.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1803,21 +1811,44 @@ type.**
...
-patterns can be complex, so the proposal for that is in a [separate
-document][exhaustiveness].
+patterns is complex, so the proposal to define how it works is in a [separate
+document][exhaustiveness]. That tells us if the cases in a switch statement
+or expression are exhaustive or not. Given that:
+
+*   It is a compile-time error if the cases in a switch expression are not
+    exhaustive. *Since an expression must yield a value, the only other option
+    is to throw an error and most Dart users prefer to catch those kinds of
+    mistakes at compile time.*
```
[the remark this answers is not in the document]
munificent: > Users can still write those defaults manually, so it's just for convenience.

Yeah, I don't want to add `switch!` or layer any other syntactic sugar on top of this syntactic sugar. There's already a lot of that going on and at some point, we have to err on the side of being explicit. :)

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1803,21 +1811,44 @@ type.**
...
+
+*   It is a compile-time error if the static type of the matched value in a
+    switch statement is an *exhaustive type* and the cases are not exhaustive.
+    An exhaustive type is:
+
+    *   `bool`
+    *   `Null`
+    *   A type whose declaration is marked sealed
+    *   `T?` where `T` is exhaustive
+    *   `FutureOr<T>` for some type `T` that is exhaustive
```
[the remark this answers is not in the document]
munificent: I admit I haven't put a *lot* of thought into this case, but it felt like the right behavior to me.

> Would a general rule be that a union is exhaustive if any part of it is exhaustive?
That could be argued: This type must be exhausted. This type is part of the possible values you switch over. Ergo, this switch must be exhaustive.

Yes, I think so. One way to think of it is, when you mark a type exhaustive, you're saying, "I *want* you to yell at me whenever this type is not used exhaustively in a switch." Since you have opted *in* to that compile-time checking, the compiler should persist that preference as well as its able. That suggests that whenever it shows up in a union type, it taints the entire type. Otherwise, the alternative would be to silently drop the exhaustiveness checks on the floor, which is unlikely to be what the user wants.

> If so, could we say that a switch must "exhaust the exhaustive type", without needing to be exhaustive in general?

I think simplest is best and just say it must be exhaustive. Otherwise, you could promote but still not actually use the thing.

See also: https://github.com/dart-lang/language/issues/2179

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1803,21 +1811,44 @@ type.**
...
+    *   A record type whose fields are all exhaustive types
+
+    **TODO: Finalize the syntax for marking a class as a sealed family.**
+
+*   It is a compile-time warning if the static type of the matched value in a
+    switch statement is an enum type or a nullable enum type and the cases are
+    not exhaustive.
```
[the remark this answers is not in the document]
munificent: > Why not just put enum types into the list of exhaustive types?

Enums are a warning if non-exhaustive, while other types are an error. :-/ We could lump them together, but that would be a breaking change for any code that's currently ignoring the warning on enums.

> By separating them, it means that (EnumType, bool) is not exhaustive.

Oh, crap. :( Nice catch.

OK, redid this. Now it treats enum types as exhaustive types too, with respect to composing in record types. Then it carves out where the matched type is an enum type (or nullable one) specially as a warning for backwards compatibility.

I'm not sure if that's what we want, so I'll file an issue and see how Michael feels about possibly making it always an error and taking the breaking change.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2038,16 +2069,38 @@ To match a pattern `p` against a value `v`:
...
     1.  Evaluate the right-hand constant expression to `c`.
 
-    2.  A `== c` pattern matches if `v == c` evaluates to true. *This takes into
-        account the built-in semantics that `null` is only equal to `null`.*
+    2.  If the operator is `==`:
+
+        1.  Let `r` be the result of `v == c`.
+
+        2.  If `r` is not a Boolean then throw a runtime error. *This can
+            happen if operator `==` on `v`'s type returns `dynamic`.*
```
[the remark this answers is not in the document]
munificent: Done.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2038,16 +2069,38 @@ To match a pattern `p` against a value `v`:
 
     1.  Evaluate the right-hand constant expression to `c`.
 
-    2.  A `== c` pattern matches if `v == c` evaluates to true. *This takes into
-        account the built-in semantics that `null` is only equal to `null`.*
+    2.  If the operator is `==`:
+
+        1.  Let `r` be the result of `v == c`.
+
+        2.  If `r` is not a Boolean then throw a runtime error. *This can
```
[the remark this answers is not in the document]
munificent: I use "Boolean" to refer to the values `true` and `false`. <shrug>

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2038,16 +2069,38 @@ To match a pattern `p` against a value `v`:
...
-    3.  A `!= c` pattern matches if `v == e` evaluates to false. *This takes
-        into account the built-in semantics that `null` is not equal to anything
-        but `null`.*
+        1.  Let `r` be the result of calling `op` on `v` with argument `c`.
 
-    4.  For any other operator, the pattern matches if calling the operator
-        method of the same name on the matched value, with `c` as the argument
-        returns true.
+        2.  If `r` is not a Boolean then throw a runtime error. *This can happen
+            if the operator on `v`'s type returns `dynamic`.*
```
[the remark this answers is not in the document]
munificent: > Is that pattern allowed? (I can argue either way.)

It's not. Under type checking, we say:

> 1.  Type-check each of `f`'s positional field subpatterns using the corresponding positional field type on `M` as the matched value type **or `Object?` if `M` is not a record type with the corresponding field.**

(Emphasis added.) Since we don't know the types of the fields, we type check each field as `Object?`. Since `Object?` doesn't define `<` (unless you've got some extension floating around that adds it), this is a compile error.

> If no, could I do `dynamic _ & < "a"` instead to force the dynamic invocation?

We still need to figure out promotion but... probably not?

> (Would it change if the switch expression had type `dynamic` instead of `Object?`?)

No, strangely. It will still use `Object?`. That's a bug. Rewrote the type checking to fix.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2038,16 +2069,38 @@ To match a pattern `p` against a value `v`:
...
+
+        3.  The pattern matches if `r` is true and fails otherwise. *This takes
+            into account the built-in semantics that `null` is only equal to
+            `null`.*
```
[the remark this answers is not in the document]
munificent: Oops. Done.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2126,7 +2179,7 @@ To match a pattern `p` against a value `v`:
         becomes a runtime exception if the map pattern is in a variable
         declaration.*
```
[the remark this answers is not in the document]
munificent: Done.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2159,17 +2209,233 @@ To match a pattern `p` against a value `v`:
     1.  If the runtime type of `v` is not a subtype of the static type of `p`
         then the match fails.
 
-    3.  Otherwise, for each field `f` in `p`:
+    2.  Otherwise, for each field `f` in `p`, in source order:
 
         1.  Call the getter with the same name as `f` on `v` to a result `r`.
```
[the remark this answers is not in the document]
munificent: Done.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2137,9 +2190,6 @@ To match a pattern `p` against a value `v`:
 
     4.  The match succeeds if all entry subpatterns match.
 
-    *Note that, unlike with lists, a matched map may have additional entries
-    that are not checked by the pattern.*
-
 *   **Record**:
 
     1.  If the runtime type of `v` is not a record type with the same type as
```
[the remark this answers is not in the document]
munificent: Done.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2159,17 +2209,233 @@ To match a pattern `p` against a value `v`:
...
+    member call, this tracks the type the call was resolved to.
+*   A member name.
+*   A possibly empty list of argument constant values.
+
+Two invocation keys are equivalent if and only if all of these are true:
+
+*   They both have parent invocation keys and the keys are equivalent or
+    neither of them have parent invocation keys.
+*   The extension types refer to the same type or are both absent.
+*   The member names are equivalent.
```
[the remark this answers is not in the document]
munificent: Done.

working/0546-patterns/patterns-feature-specification.md
munificent: I went back and forth on nomenclature here. Yes, it semantically refers to the invocation that yielded the current object that this invocation is called on. But I tried to define invocation keys more abstractly as a pure data structure. And in that case, each invocation key is essentially just a linked list with a single "parent" pointer, so I went with that word.

working/0546-patterns/patterns-feature-specification.md
munificent: Reworded.

working/0546-patterns/patterns-feature-specification.md
munificent: Fixed.

working/0546-patterns/patterns-feature-specification.md
munificent: Done.

working/0546-patterns/patterns-feature-specification.md
munificent: Yeah, this was pretty hand-wavey. Changed it to something that I think/hope captures that.

working/0546-patterns/patterns-feature-specification.md
munificent: Done.

PR #2504, opened 2022-09-20
Adjust views to use `implements`, and to not have show/hide

eernstg:
This is a draft revision of the views proposal, changing it to use `implements` rather than `extends` to declare superviews, and eliminating the support for `show` and `hide` in superviews (so every conflict has to be handled by overriding).

Added changes discussed at the meeting today Wednesday Sep 21: Deleted everything about the `export` declaration and everything about `show` and `hide`. Changed `view` to `view class` at the beginning of a view declaration.

### DIFF
--- a/working/1426-extension-types/feature-specification-views.md
+++ b/working/1426-extension-types/feature-specification-views.md
@@ -7,6 +7,9 @@ Status: Draft
 
 ## Change Log
 
+2022.09.21
+  - Removed support for `export`, changed `view` to `view class`
+
 2022.09.20
   - Updated the inheritance mechanism to fit in with a potential non-virtual
     method mechanism for classes: Use `implements`, remove show/hide.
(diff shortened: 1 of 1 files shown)


working/1426-extension-types/feature-specification-views.md
```diff
@@ -114,7 +121,7 @@ type `R` or some supertype of `R` (which is what we normally do). For
 example:
```
mit-mit, line 121: There are a few places above here that also need updating, e.g. :
- line 29 "we call views" to "we call view classes"
- line 32 "declared by a new view decl." to "declared by a view class decl"
- line 33 "replacement or modification" to just "replacement"

mit-mit (changes requested):

working/1426-extension-types/feature-specification-views.md
eernstg: Done.

mit-mit (approved):

PR #2508, opened 2022-09-22
[Records] Fix typo. It's an error to have named field hashCode, etc..

sgrekhov:
Changelog reads:

> Allow positional fields in record types named hashCode, runtimeType, noSuchMethod, or toString.

So, I think, that it is a compile-time error to have a _named_ field with the name equal any of `Object` members


munificent (approved):

munificent:
Oops, yes, thank you! :D

PR #2514, opened 2022-09-23
Move the feature specification of super parameters into 'accepted'

eernstg:
The super-parameters feature was released with Dart 2.17. This PR moves the 'proposal.md' from 'working' into 'accepted/2.17/super-parameters' and renames it to 'feature-specification.md'. The existing 'proposal.md' is updated to have no other contents than a link to the new feature specification.

The phrase 'super-initializer parameters' in the title was replaced by 'super parameters' (there was only this single use of that phrase, and it seems rather universally accepted that they can be called 'super parameters'). No other changes, if any further changes are needed I think it can just as well be done in a separate PR.


leafpetersen (approved):
Thanks!

PR #2519, opened 2022-09-24
Remove record keyword

mnordine:
Now that it seems Records have been accepted, this option would be very confusing.


leafpetersen (commented):
This proposal has largely been superceded at this point I think (though I may pull pieces out of it), so I think I'd prefer to just leave this as is.

PR #2526, opened 2022-09-28
Views spec updates based on discussion Sep 28

eernstg:
Hi, @stereotype441, could you take a look? This update includes several changes motivated by your feedback, plus some symbol rename operations for consistency (such that the representation type is always called `R` and there are always `s` type parameters, etc).

### DIFF
--- a/working/1426-extension-types/feature-specification-views.md
+++ b/working/1426-extension-types/feature-specification-views.md
@@ -451,7 +451,7 @@ We use
 <code>invokeViewMethod(V, &lt;T<sub>1</sub>, .. T<sub>k</sub>&gt;, o).m(args)</code>
 where `V` is a type name denoting a view to denote the invocation of
 the view method `m` on `o` with arguments `args` and view type
-arguments 
+arguments
 <code>T<sub>1</sub>, .. T<sub>k</sub></code>.
 Similar constructs exist for invocation of getters, setters, and
 operators.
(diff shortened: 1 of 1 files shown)


working/1426-extension-types/feature-specification-views.md
```diff
@@ -718,23 +743,26 @@ A compile-time error occurs if `V1` is a type name or a parameterized type
 which occurs as a superview in a view declaration _DV_, but `V1` does not
 denote a view type.
 
+A compile-time error occurs if `V1` is a superview of `V1`, directly or
```
stereotype441, line 746: Since `V1` is a view *type*, not a view *declaration*, I think this permits stuff like this:

```dart
view class V<T> implements V<List<T>> {}
```

(because `V<T>`, `V<List<T>>`, `V<List<List<T>>>`, and so on are all distinct types).

I don't think we want to permit that.  Could we say something like this instead?

> A compile-time error occurs if any superview of _DV_ is the type `View` or `View<...>`. *As usual, subtype cycles are not allowed.*

working/1426-extension-types/feature-specification-views.md
```diff
@@ -746,12 +774,12 @@ A compile-time error occurs if a view _DV_ has two superviews `V1` and
...
-inheritance pattern can create the case where two superviews have an
-_m_, but they are both declared by the same declaration (so `V` is a
-subview of `V1` and `V2`, and both `V1` and `V2` are subviews of `V3`,
-and `V3` declares _m_, in which case there is no conflict in `V`).*
+from two superviews then the subview must resolve the conflict. The
+so-called diamond inheritance pattern can create the case where two
+superviews have an _m_, but they are both declared by the same
+declaration (so `V` is a subview of `V1` and `V2`, and both `V1` and
+`V2` are subviews of `V3`, and `V3` declares _m_, in which case there
+is no conflict in `V`).*
```
stereotype441, line 782: Hmm, in order for this to be sound, I think we also need to require that each transitive superview `V1` of _DV_ must have the same type arguments.  E.g. this is not allowed:

```dart
view class V1<T> {
  void m(T t) {}
}
view class V2 implements V1<int> {}
view class V3 implements V1<String> {}
view class DV implements V2, V3 {}
```

stereotype441 (changes requested):

working/1426-extension-types/feature-specification-views.md
eernstg: Certainly! Done.

eernstg (commented):
Review response.

stereotype441 (approved):
lgtm, thank you!

eernstg:
Thanks!

PR #2529, opened 2022-09-29
[records] Disambiguate metadata annotation arguments and record types.

### LINKED ISSUE #2469: [records] parse errors when trying to annotate a record type class field
The following code generates what look like parser errors.

image

```
error: Annotation must be either a const variable reference or const constructor invocation. (invalid_annotation at [sandbox] lib/public.dart:27)
error: Variables must be declared using the keywords 'const', 'final', 'var' or a type name. (missing_const_final_var_or_type at [sandbox] lib/public.dart:28)
```

/fyi @jensjoha

munificent:
Fix #2469.


jensjoha:
Can we please not do the

```
Comments do not count as whitespace,
nor do whitespace characters in comments. The newline at the end of a line
comment *does* count as whitespace.
```

part?

accepted/future-releases/records/records-feature-specification.md
```diff
@@ -257,6 +257,80 @@ avoid this ambiguity ([#2469][]).**
...
+
+```
+metadatum ::= identifier                                    // Existing rule.
+            | qualifiedName                                 // Existing rule.
+            | constructorDesignation ~WHITESPACE arguments  // Changed.
+```
+
+The `~WHITESPACE` lexical rule matches when there are no whitespace characters
+(according to the existing `WHITESPACE` lexical rule) between the
+`constructorDesignation` and `arguments`. Comments do not count as whitespace,
```
lrhn, line 297: I'd want both `~WHITESPACE` and `~COMMENT` here.

The must be *characters* between the end of the identifier and the following `(` or `<`.
That should be the easiest thing to implement (`idToken.end == nextToken.start` or something similar).

Anything more complicated is just asking for trouble.

I can see the rationale for allowing inline comments (with no surrounding whitespace), but I don't think it's actually worth it.

accepted/future-releases/records/records-feature-specification.md
```diff
@@ -257,6 +257,80 @@ avoid this ambiguity ([#2469][]).**
...
+@metadata (x,) a;
+
+@metadata
+(x,) a;
+
+@metadata // Comment.
+(x,) a;
+
+@metadata/* Comment with newline after. */
+(x,) a;
```
lrhn, line 320: Other examples that might be relevant
```dart
@metadata /* a */(int. int) Function() f;
@metadata/* a */ (int. int) Function() f;
@metadata/* a *//* b */(int. int) Function() f;  // *
@metadata/* a */ /* b */(int. int) Function() f;
```
Of these, I think only the one marked `*` would be valid.

Again, I think it's just too complicated to allow comments. 

Also, it's worth showing that whitespace after `>` is different:
```dart
@metadata<A> (int, int) Function() f; // *
@metadata<A> 
  (int, int) Function() f;  // *
```
(Both would be valid as *annotation arguments*.)

munificent: > From https://github.com/dart-lang/language/issues/2469#issuecomment-1255950332
>
>> I'm very tempted to say that if the next source token (ignoring comments and whitespace) after the identifier does not start right where the identifier ends, it's not part of the annotation. So no comments either.
>
> Can we have that instead please?

Yeah, I think that's reasonable. It's almost as if the `(` is actually part of the annotation's identifier *token*.

> Also, it's worth showing that whitespace after `>` is different:
> 
> ```dart
> @metadata<A> (int, int) Function() f; // *
> @metadata<A> 
>   (int, int) Function() f;  // *
> ```
> 
> (Both would be valid as _annotation arguments_.)

Ugh, that's super weird. What I specified in this PR is:

```dart
@metadata<A>(int, int) f; // `(int, int`) goes to annotation.
@metadata <A>(int, int) f; // `(int, int`) goes to annotation.
@metadata<A> (int, int) f; // `(int, int`) goes to variable type.
@metadata <A> (int, int) f; // `(int, int`) goes to variable type.
```

I wonder if we should disallow whitespace before `<` too. It's not ambiguous (now) but it feels very strange to not just ignore the whitespace before `<` but also have that mean that whitespace before the subsequent `(` is now allowed and ignored.

munificent: OK, I took a position on this. Let me know if you don't like it.

accepted/future-releases/records/records-feature-specification.md
munificent: SGTM.

accepted/future-releases/records/records-feature-specification.md
```diff
@@ -252,10 +252,100 @@ ambiguity, we disambiguate by treating `on` as a clause for `try` and not a
...
+```
+
+This is a syntax error because the `<T>` means there *must* be an argument list
+after it, but the `NO_SPACE` in `metadatum` prevents it from being parsed as
+such and the result is an error. *We could ignore whitespace after the `>` by
+tweaking the grammar, but we choose to require `NO_SPACE` even here since
+`@metadata<T>` appears to be a generic instantiation and could potentially be
+valid syntax in the future.*
+
+**Breaking change:** Existing metadata annotations with whitespace before their
```
leafpetersen, line 343: We should clarify whether this is a hard breaking change or language versioned.  @jensjoha I would imagine you would prefer not to language version this?

jensjoha: Yes.

lrhn: Would we want the `on`-after-`try`-`catch` to be a breaking change too?

jensjoha: That's what was suggested in https://github.com/dart-lang/language/issues/2406, right?

lrhn: Not sure I was being that precise, I was mainly thinking of whether code needed to be migrated, not whether it should be fixed. But since there likely are no code to fix/migrate, I'm fine with either choice.

munificent: I kind of assumed we would gate this behind a language version since the code it disambiguates can only appear in a later language version too. Is there a reason we wouldn't?

jensjoha: For one thing the parser doesn't know about language versioning.

jensjoha: Also if the parser _was_ to know about language versioning (and for instance not know about a feature that's not turned on), recovery is going to suffer.

lrhn: Have we really not made any breaking changes since 2.7? (Like we did with `f(a<b,c>(d))` in Dart 2.0.)

I don't think we can promise to not make such a breaking change in the future.

Parsing incremental features even in libraries with a language-version which doesn't support the feature, is fine, but if we have breaking changes, where the *same* syntax parses *differently* in a later version, we should be able to support that too. That means finding the language version before parsing the code, and parsing differently depending on the language version.

We designed the `//@dart=2.7` syntax, and placement of it, so that it should always be possible to read the language marker before any actual Dart code. As long as we don't change comment syntax based on language version, the parser can see the language version marker first. (It will have to look for it, obviously.)

munificent: OK, updated to clarify that the disambiguation is not gated on the language version.

jensjoha: The "parser" is split into two: the scanner and the parser.
Currently the scanner has some booleans about features, e.g. extension types which affect whether "extension" is marked as a keyword or an identifier. (to be honest I'm not sure that's necessary, but that's how it is) whereas the parser knows nothing (*).
The parser gets a token stream and then parses the language as it knows it. This also means that when getting an invalid program it can recover - try to make sense of it - with the full language. For instance if it sees `(a, b)` it will report it to the listener as a record (type or literal depending on where it is) and the listener will then say "you don't have records enabled. do this to enable it" if records aren't enabled.
If the parser instead was to recover without records it would spew lots of errors about "unexpected comma" or "expected end-paren" or whatever, which is a lot less useful imo.

Now, I understand that if it's the same syntax meaning different things, recovery (maybe) wouldn't suffer. I can't really tell what's worse, though, code that silently means something different when I upgrade my sdk, or code that silently means something different when I upgrade my min sdk version in my pubspec.yaml... Both are pretty bad (but I was also a fan of the `Record(a, b)` syntax that didn't have these problems).

*: because of the booleans in the scanner, triple shift will become two tokens when triple shift is not enabled and we had to hack better recovery around that in the parser which wasn't great (in that recovery we, as I recall, give an error message about language versions although we're really guessing).

jensjoha: Actually, in the case around `on` I probably do know which is worse.
If it's unconditionally the new thing and one upgrades the sdk but not the pubspec one will get an error saying you don't have records enabled.
If it's conditionally on the language version, upgrading won't give an error. Then changing the language version won't give an error, but it will mean something different.

munificent: OK, so there's two pieces here:

* How the parser handles record expressions in libraries whose language version doesn't support records. I think it makes perfect sense to parse that as a record expression and then report a graceful error about not having the feature enabled. The syntax isn't valid anyway, so the parser's going to have some kind of error either way and the most likely thing the user thought they were doing was writing a record.

* How the parser handles the way we disambiguated code to deal with records. Concretely, that means an `on` followed by a parenthesized thing after a `try` block and whitespace between a metadata annotation name and the parameter list. If the parser ignores language versioning and applies the new disambiguation rules even to old code, then code which used to be fine now has an error.

The latter feels unnecessarily breaking to me and normally I'd suggest we not do that. But given that this seems to not really break any actual code in the wild, I think it's probably fine too. That's what the latest version of this PR specifies. Is that OK?

jensjoha: I am certainly fine with it not being guarded by language versions --- that much was probably clear so the question was likely meant for someone else :)

munificent:
I've updated the proposal to clarify that the parser changes are not language versioned. Reviewer friends, does this look ready to land?

lrhn (approved):
LGTM

PR #2535, opened 2022-09-29
[records] Allow "()" as an empty record expression.

munificent:
I believe we decided to do this in a previous language meeting, just catching the proposal up to that.

### DIFF
--- a/accepted/future-releases/records/records-feature-specification.md
+++ b/accepted/future-releases/records/records-feature-specification.md
@@ -122,6 +122,8 @@ not captured by the grammar. It is a compile-time error if a record has any of:
 
 *   Only one positional field and no trailing comma.
 
+*   No fields and a trailing comma. *The expression `(,)` isn't allowed.*
+
 *   A field named `hashCode`, `runtimeType`, `noSuchMethod`, or `toString`.
 
 *   A field name that starts with an underscore.
@@ -138,8 +140,7 @@ var number = (1);  // The number 1.
 var record = (1,); // A record containing the number 1.
 ```
 
-There is no syntax for a zero-field record expression. Instead, there is a
-static constant `empty` on `Record` that returns the empty record.
+The expression `()` refers to the constant empty record with no fields.
 
 ### Record type annotations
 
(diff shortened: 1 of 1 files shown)


leafpetersen (approved):

leafpetersen:
@chloestefantsova @johnniwinther @scheglov FYI, `Record.empty` goes away, replaced with `()`.

munificent:
> FYI, `Record.empty` goes away, replaced with `()`.

That reminds me, I should update the CHANGELOG.

johnniwinther:
cc @jensjoha

accepted/future-releases/records/records-feature-specification.md
```diff
@@ -122,6 +122,8 @@ not captured by the grammar. It is a compile-time error if a record has any of:
 
 *   Only one positional field and no trailing comma.
 
+*   No fields and a trailing comma. *The expression `(,)` isn't allowed.*
```
lrhn, line 125: My guess is that this is (or should be) captured by the grammar. 
In the rest of our grammar, an optional trailing colon is usually handled as:
```dart
collection ::= startToken elements? endToken
elements ::= element (',' element)* ','?
```

The grammar above doesn't actually allow `()` as a record. And it allows `(e)`, which is ambiguous.
Having a prose section saying that it isn't allowed is technically enough, but not how we usually do it.

How about:
```abnf
record ::= 'const'? '(' recordFields? ')'

recordFields ::= expression (',' | (',' recordField)+ ','?)
               | namedRecordField ( ',' recordField )* ','?

recordField ::= expression | namedRecordField
namedRecordField ::= identifier ':' expression
```

An alternative phrasing would be:
```dart
recordFields ::= expression ','
               | namedRecordField ','?,
               | recordField (',' recordField)+ ','?
```

munificent: > My guess is that this is (or should be) captured by the grammar.

I thought about that.

> Having a prose section saying that it isn't allowed is technically enough, but not how we usually do it.

Yes, and the implementers tell me often that they don't like the way we currently do it. :)

Every time I try to sort out `type`, `typeNotVoid`, `typeNotFunction`, and `typeNotVoidNotFunction`, I get a little sadder.

PR #2538, opened 2022-09-30
Specify interactions with older language versions

leafpetersen:
Specify the errors for when record syntax is used in libraries with an older language version, and clarify the treatment of record types and values that flow into libraries on older language versions.


leafpetersen:
cc @srawlins @bwilkerson @johnniwinther @chloestefantsova FYI

accepted/future-releases/records/records-feature-specification.md
```diff
@@ -586,8 +586,51 @@ The runtime type of `pair` is `(int, double)`, not `(num, Object)`, However, the
...
+semantic interpretation of the typedef is as the underlying record type, just as
+it would be for any other type.  Similarly, type inference may introduce record
+types into a legacy library, and such types will be interpreted by the compiler
+as record types as usual (that is, there is no erasure implied to remove these
+inferred types).  Record values may flow into a legacy library via a reference
+to a member from another libraries, and a legacy library may freely call getters
+on record values (since there is no new syntax for calling a record getter).
+The reason for this choice is that the intent of language versioning (for an
+additive feature such as records) is to ensure that users do not accidentally
+use new features in a package without specifying an sdk constraint which ensures
```
munificent, line 620: Tiny grammar nit: "sdk" -> "SDK" here and on the next line.

munificent (approved):
LGTM!

accepted/future-releases/records/records-feature-specification.md
leafpetersen: done

accepted/future-releases/records/records-feature-specification.md
```diff
@@ -586,8 +586,51 @@ The runtime type of `pair` is `(int, double)`, not `(num, Object)`, However, the
...
+"legacy library") to include or reference the `Record` class, record types or
+record expressions directly or indirectly from another library whose language
+version is greater than or equal to `v`.  Such a legacy library may reference a
+typedef name which is bound to a record type in another library, and the
+semantic interpretation of the typedef is as the underlying record type, just as
+it would be for any other type.  Similarly, type inference may introduce record
+types into a legacy library, and such types will be interpreted by the compiler
+as record types as usual (that is, there is no erasure implied to remove these
+inferred types).  Record values may flow into a legacy library via a reference
+to a member from another libraries, and a legacy library may freely call getters
```
lrhn, line 616: "another library" or "other libraries"

accepted/future-releases/records/records-feature-specification.md
```diff
@@ -586,8 +586,51 @@ The runtime type of `pair` is `(int, double)`, not `(num, Object)`, However, the
...
+on record values (since there is no new syntax for calling a record getter).
+The reason for this choice is that the intent of language versioning (for an
```
lrhn, line 618: The "this choice" is a little underspecified (there are many small choices above, and normative and explanatory text is not distinguished).

Just elaborate: "The choice to not make it an error to reference record types in legacy libraries ...."

(Documentation always gets better when the implicit is made explicit. Prove me wrong!)

accepted/future-releases/records/records-feature-specification.md
```diff
@@ -586,8 +586,51 @@ The runtime type of `pair` is `(int, double)`, not `(num, Object)`, However, the
...
+### Interactions with libraries using older language versions
+
+The records feature is language versioned, as usual with new Dart features.
+This means that it will be an error to use the syntax for records in libraries
+which do not have a language version greater than or equal to the language
+version in which records are released.  More specifically, assuming that `v` is
+the language version in which records are released, the following errors apply.
+
+It is an error to reference the SDK class `Record` in a library whose language
+version is less than `v`.
```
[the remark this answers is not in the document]
leafpetersen: > Just drop this paragraph entirely, it's not a restriction we need.

I don't think this is true.  I can clarify this, I see no issues with specifying this more precisely if you feel that the intent is unclear, but this is an important error.  Without this error, you can write a library which uses the `Record` class (which will work with no errors or warnings, since you are running on an sdk that supports it) and then ship the code to pub without raising your sdk constraint, whereupon you break the world.

> The Record declaration will be marked with @Since("v"), so we can still give warnings if you use it in a package with min-SDK < v.
(And if your min SDK >= v anyway, there is no problem, even if the referencing library is held back by a //@dart=<v marker.)

Empirically, the `@Since` annotation doesn't seem to work this way.  I can write a package with sdk constraint   `sdk: ">=2.9.0 <3.0.0"` and use the `Object.hash` method, which is marked `@Since("2.14")` with  no errors or warnings when running with a bleeding edge SDK.  

> To disallow direct references to the Record declaration, we'd have to be explicit and, say, state that you are not allowed to have the identifier Record in a place where it resolves to/denotes the Record class declaration of dart:core.

This is more restrictive than we need (it's fine to use the identifier `Record`, which resolves to the `Record` class from `dart:core`, if it is re-exported via another library).  But I don't know that we care about this use case.  I'll update this with more precise wording, either using this restriction or something more fine grained.

accepted/future-releases/records/records-feature-specification.md
leafpetersen: More verbosity isn't always better?  This text is commentary - the set of errors is completely specified and circumscribed without it.  Do you really find this text confusing?  If we're trying to be 100% formally precise it's going to be a lot more wordy.  For example, it's not just "The choice to not make it an error to reference record types".  It's "The choice to make X, Y and Z and error, but not A, B, C, D, E, F is because" (because note that it's not just types, it's expressions, etc, etc).  On balance, unless you legitimately believe that this non-normative text is in some way confusing, I'd prefer to err on the side of legibility here.

accepted/future-releases/records/records-feature-specification.md
```diff
@@ -586,8 +586,51 @@ The runtime type of `pair` is `(int, double)`, not `(num, Object)`, However, the
...
+
+It is an error for the record literal syntax (e.g. `(3, 4)`) to be used
+syntactically in a library whose language version is less than `v`.
+
+It is an error for the record type syntax (e.g. `(int, int)`) to be used
+syntactically in a library whose language version is less than `v`.
```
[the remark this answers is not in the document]
leafpetersen: I think this should be covered elsewhere (e.g. by the section that @munificent is adding on the grammar).  This section describes the interactions between different language versions: it does not specify how the syntax of the the current language version is different from the syntax of the old language version.

accepted/future-releases/records/records-feature-specification.md
leafpetersen: Sharpened "this choice" slightly, PTAL.

accepted/future-releases/records/records-feature-specification.md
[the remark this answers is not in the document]
lrhn: I expecting to not restrict access to the type. We don't generally restrict access to new features in the platform libraries to libraries with language versions before the SDK where the feature was introduced.

We could, and then it should really apply to every `@Since`-annotated platform declaration, but so far we haven't done that. So, is there something which makes `Record` special in this regard, or is it something we would want to apply more generally?


As for phrasing, if we do want the restriction, maybe:
> ... must not contain the identifier `Record` in a place where the name denotes the `Record` class declaration of `dart:core`, and where that import scope name is only imported from platform libraries.

That should allow any reference which is also exported through a non-platform library.

accepted/future-releases/records/records-feature-specification.md
lrhn: Verbosity isn't better, but being explicit is better than being implicit :)

Phrasing is better now. I don't get confused about which choice "this choice" refers to.

Excellent :)

accepted/future-releases/records/records-feature-specification.md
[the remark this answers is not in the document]
leafpetersen: The point of this section is to describe the *interactions* between language versions.  It does so by circumscribing the set of errors that arise from using record features in older language versioned libraries (and noting that certain things are not in that set of errors, e.g. types that arise by inference).  The first of the two things you mention is already covered in this proposal, I don't see value in adding it here.  The second thing, I don't know the answer to, it's still under discussion.  

If you really believe that these should be covered here, I'd suggest following up with a PR to re-organize as you prefer.

accepted/future-releases/records/records-feature-specification.md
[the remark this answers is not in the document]
leafpetersen: > ... must not contain the identifier Record in a place where the name denotes the Record class declaration of dart:core, and where that import scope name is only imported from platform libraries.

I used something modeled after this.

accepted/future-releases/records/records-feature-specification.md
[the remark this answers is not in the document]
lrhn: ACK. If this section is simply on *records*, and not the entire feature around them (and we don't even know if those two changes will be rolled into the records flag, or just be released independently), then it makes sense to focus only on the actual record types.

I was considering those two changes as part of the records feature, at an equal level to record syntax, but that's probably premature.

So keep as is.

PR #2541, opened 2022-09-30
Clarify how implicit conversions happen in patterns.

### LINKED ISSUE #2488: [Records] Should pointwise implicit downcasts be allowed in records?

```dart
  dynamic d = "hello";
  (int, int) r = (d, 4);
```

There are two possible inferred types for the record literal.  
  - We might choose to infer this as `(dynamic, int)`
    - In this case this code is a static error, since neither type is `dynamic`, and the RHS type is not a subtype of the LHS type.
  - We might choose to infer this as `(int, int)`
    - In this case this code is a runtime error, since there must be an implicit cast added to cast `d` to `int` before constructing the record.

Which do we intend?

In general, I'm not keen on adding more implicit downcasts.  I'd rather the user make the cast explicit here.

We do push casts inwards for list literals however:

```
  dynamic d = "hello";
  List<int> l = [d];
```

fails at runtime, not statically (as it must, since we prefer the context type for nominal types, unlike the structural types).

cc @munificent @eernstg @lrhn @jakemac53 @natebosch @stereotype441 @kallentu @chloestefantsova @johnniwinther @scheglov @srawlins

munificent:
Fix #2488.


working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1404,30 +1404,66 @@ To orchestrate this, type inference on patterns proceeds in three phases:
...
 
-*It would be hard to apply inference from cases in a switch to the value since
-there are multiple cases and it's not clear how to unify that. Even in case-if
-statements, it's not clear that downwards inference is desirable, since the
-intent of the pattern is to ask a question about the matched object, and not
-necessarily to try to force a certain answer.*
+    *Since the right-hand is not a record literal, we can't use the pattern's
+    context type schema to insert coercions when the record is being created.
+    However, `(T Function<T>(T), dynamic)` is assignable to `(int Function(int),
+    String)`, so the declaration is valid. Instead, coercions are inserted after
```
lrhn, line 1463: It's not clear that the record types are assignable. This is a case where we need to destruct, coerce, and then recreate the record value, which is something I haven't been able to sell. (Probably a wise choice too.)

What is the case here is that `T Function<T>(T)` is assignable to `int Function(int)`, and `dynamic` to `String`, and we *don't* check the assignability of the record type itself here, only the of the actual values extracted by destructuring.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1738,6 +1796,32 @@ It is a compile-time error if:
     destructuring and variable binding won't fail to match (though it might
     throw a runtime exception from implicit downcasts from `dynamic`).*
 
+If `p` has required type `T` and `M` is not a subtype of `T` but is assignable,
+then an implicit coercion or cast is inserted immediately before the pattern
+first accesses the value. *Each pattern that requires a certain type can be
+thought of as an "assignment point" where an implicit coercion may happen when a
+value flows in during matching. Examples:*
```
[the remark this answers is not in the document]
munificent: Sure, for example:

```dart
var Foo(:x, :y) = Foo();
```

Here, the extractor pattern calls `.x` on the incoming value and then calls `.y` on it.

Reworded to be less specific, though.

working/0546-patterns/patterns-feature-specification.md
munificent: > It's not clear that the record types are assignable.

I actually can't find where assignability is defined _at all_. The language spec still seems to have the old implicit downcast verbiage. I didn't find any mention of assignability in "resources/type-system".

@leafpetersen, is there somewhere else I should be looking?

> This is a case where we need to destruct, coerce, and then recreate the record value,

I don't think so. We have a record object already. The record pattern will match it as long as it has the right shape and doesn't care about the field types at all. Then the record pattern destructures each field and sends them to the corresponding field subpatterns. It's when sending the destructured field values into the subpatterns that we insert the coercions. So we should never need to allocate a new coerced record.

lrhn: Assignability is defined (at least) here: https://github.com/dart-lang/language/blob/master/accepted/2.12/nnbd/feature-specification.md#assignability

Which is not useful, since it only covers `dynamic`. 

A more recent reference is from ["constructor tearoffs"](https://github.com/dart-lang/language/blob/master/accepted/2.15/constructor-tearoffs/feature-specification.md#:~:text=both%20implicit%20and%20explicit%20instantiation).
(which, btw., was where we specified that you can do `.call` and instantiation in the same implicit operation.)

It's not specified in terms of "assignability" though.
It is probably just assumed to be part of type inference, which ... is still not formally specified?

So, specification is lacking.
What the *effect* has so far been is the following:
If `e` has static type `T` in context type `C` then
* If `T` has a `call` method (instance or extension) and `C` is a function type or `Function`, then treat `e` as if it was `e.call` with a static type of the function signature of the `call` method.
* If `T` is a generic function type and `C` is a non-generic function type, try to infer type arguments `typeArgs` such that the instantiation `T<typeArgs>` is assignable to `C` (in the specific way we usually do that). If so, treat `e` as if it was `e<typeArgs>` with static type `T<typeArgs>`.
* If `T` is dynamic and `C` is a non-top-type, treat `e` as if it was `e as C` with static type `C`.

The first two can both apply if the `call` method is generic.
We then say that `T` is assignable to `C` if the conversion above gives a program which is well typed.

For records, we have to decide whether the record type `(A, B)` is assignable to `(B, C)` by any means other than those above.
By pushing in the context type, we get conversion on the individual fields of a record literal.
For a non-literal, the above conversions do not apply (unless someone extends a record type with a `call` method).

Which means that if we don't destructure and rebuild, then `(T Function<T>(T), dynamic)` cannot be assignable to `(int Function(int), String)`.

munificent: > For records, we have to decide whether the record type `(A, B)` is assignable to `(B, C)` by any means other than those above.
> ...
> Which means that if we don't destructure and rebuild, then `(T Function<T>(T), dynamic)` cannot be assignable to `(int Function(int), String)`.

I don't want to have to destructure *and rebuild*, but I think it's fine to apply the coercions at the point of destructuring. That suggests to me that we don't want to define assignability for records specially. A record type R is assignable to T only if R is a subtype of T, as usual. That way, this is forbidden:

```dart
(int, int) record = (1 as dynamic, 2 as dynamic);
```

Otherwise, we would have to destructure and then rebuild a new record. But I think allowing this is reasonable:

```dart
var (int x, int y) = (1 as dynamic, 2 as dynamic);
```

Because we don't need to actually build any new record here. I believe the current proposal draws this line. It forbids the former because there's no special assignability rules for records. It allows the latter because the required type of a record pattern always uses `Object?` for the field types. So in the second example, there is no compile time error because `(dynamic, dynamic)` is assignable to `(Object?, Object?)` and likewise `dynamic` is assignable to `Object?` for each field.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1606,15 +1649,33 @@ To type check a pattern `p` being matched against a value of type `M`:
     }
     ```
 
-    *Here, the `1` constant pattern in the case is inferred in a context type of
-    `double` to be `1.0` and so does match.*
+    *We don't do any int-to-double conversion on the case constant here and `1`
+    is not equal to `1.0` (outside of the web), so this doesn't match.*
```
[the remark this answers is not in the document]
munificent: Heh, today I learned. Changed to a different example.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1404,30 +1404,69 @@ To orchestrate this, type inference on patterns proceeds in three phases:
     a type *schema* and not a *type* because there may be holes where types
     aren't known yet.
 
+    We only calculate a pattern type schema for pattern variable declarations
+    and pattern assignments. In matching contexts (switch cases and if-case
+    statements), the pattern context type schema is not used, no downwards
+    inference is performed from the pattern to the matched value expression, and
+    no coercions or casts from `dynamic` are inserted in the matched value
+    expression.
```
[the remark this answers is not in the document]
leafpetersen: `scrutinee` is a relatively standard name for the object being matched on ("scrutinized").

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1404,30 +1404,69 @@ To orchestrate this, type inference on patterns proceeds in three phases:
     a type *schema* and not a *type* because there may be holes where types
     aren't known yet.
 
+    We only calculate a pattern type schema for pattern variable declarations
+    and pattern assignments. In matching contexts (switch cases and if-case
+    statements), the pattern context type schema is not used, no downwards
+    inference is performed from the pattern to the matched value expression, and
+    no coercions or casts from `dynamic` are inserted in the matched value
```
leafpetersen, line 1411: This is a breaking change from current behavior.  Probably not actually breaking in practice, but worth calling out, and definitely needs tests since otherwise the existing behavior will certainly carry over.

leafpetersen: Scratch this - I misread this as being scrutinee -> pattern, but you're talking about the other way around here.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1595,9 +1634,13 @@ To type check a pattern `p` being matched against a value of type `M`:
 
     [nonnull]: https://github.com/dart-lang/language/blob/master/accepted/2.12/nnbd/feature-specification.md#null-promotion
 
-*   **Constant**: Type check the pattern's value in context type `M`. *The
-    context type comes into play for things like type arguments and
-    int-to-double:*
+*   **Constant**: Nothing to do. When type-checking the constant expression,
+    use no context type.
+
+    *Constant patterns can only appear in refutable contexts, and we don't want
```
leafpetersen, line 1640: This is the breaking change that I referred to (incorrectly) above.  Currently, this code works, and prints "equal" followed by "not equal":

```dart
T id<T>(T x) => x;

void test(int Function(int) f) {
  switch(f) {
    case id: print("equal");break;
    default: print("not equal");break;
  }
  switch(f as Object) {
    case id: print("equal");break;
    default: print("not equal");break;
  }
  print("Done");
}


void main() {
  test(id);
}
```

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1595,9 +1634,13 @@ To type check a pattern `p` being matched against a value of type `M`:
...
+    the matched value type to cause a coercion that would change the constant
+    value when the intent of the pattern is to see if the matched value _is_
+    that value, not to _make_ it so. In other words, this does not match:*
```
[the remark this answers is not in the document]
leafpetersen: FWIW, we currently do do this conversion (it immediately results in an error, but we do it).

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1595,9 +1634,13 @@ To type check a pattern `p` being matched against a value of type `M`:
 
     [nonnull]: https://github.com/dart-lang/language/blob/master/accepted/2.12/nnbd/feature-specification.md#null-promotion
 
-*   **Constant**: Type check the pattern's value in context type `M`. *The
-    context type comes into play for things like type arguments and
-    int-to-double:*
+*   **Constant**: Nothing to do. When type-checking the constant expression,
```
leafpetersen, line 1637: Github reviews suck.  Somewhere above the point where this comment is (but which github does not allow me to comment on), there is a line that starts "It is a compile-time error if C? is not assignable to M's == method parameter type.".  I believe the assignability check here is backwards.  That is, you want to check that `C` is assignable to `T?` where `T` is the type of `M`'s parameter type.

working/0546-patterns/patterns-feature-specification.md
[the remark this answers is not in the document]
munificent: I'll follow up on [the bug](https://github.com/dart-lang/language/issues/2587), but I'm generally OK with relying on the existing equality behavior around doubles for matching as well. The set of users who actually care about NaN and distinguishing negative and positive zero is *very small* and they generally know that they have to be careful.

So I don't think numbers are a strong signal for deciding how we want to handle context types for constant patterns.

I know we've gone back and forth on this a number of times. My recollection is that after talking to @leafpetersen a bunch, I was pretty well convinced that *not* using the context type was the right choice, though I can't remember the arguments right now. Unless there's a compelling reason outside of numbers, I think it's probably best to leave the current behavior as specified.

working/0546-patterns/patterns-feature-specification.md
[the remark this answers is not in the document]
munificent: > Is "matched value expression" the "switch expression" (the value being switched on)?

Yeah, the proposal used "matched value", "matched value type" and "matched value expression" to refer to the thing that patterns are applied to. At the top level, it's the initializer of a pattern variable declaration, the RHS of a pattern assignment, the switch expression in a switch statement or expression, and the express before `case` in an if-case statement. In the runtime semantics section for each of the constructs, it defines what value gets sent to the pattern as the matched value.

When patterns destructure, each pattern specifies what the matched value is that gets passed to each subpattern.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1738,6 +1799,32 @@ It is a compile-time error if:
     destructuring and variable binding won't fail to match (though it might
     throw a runtime exception from implicit downcasts from `dynamic`).*
 
+If `p` has required type `T` and `M` is not a subtype of `T` but is assignable,
```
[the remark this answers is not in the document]
munificent: Yes, but that won't happen in this doc, right?

working/0546-patterns/patterns-feature-specification.md
munificent: Lasse and I are discussing this choice [here](https://github.com/dart-lang/language/pull/2541#discussion_r1004222265). Do you think we should perform downwards inference on constant patterns? I'm fairly agnostic and can't really remember most of the arguments for or against.

working/0546-patterns/patterns-feature-specification.md
munificent: Good catch.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1738,6 +1799,32 @@ It is a compile-time error if:
...
+then an implicit coercion or cast is inserted before the pattern accesses the
```
leafpetersen, line 1803: It's not clear to me what it means for a pattern to access the value.  Maybe it might be clearer to say "is inserted at the point that the pattern destructures, matches, or binds the matched value"?

working/0546-patterns/patterns-feature-specification.md
leafpetersen: I don't know.  We currently do (for more than just conversions):

```dart
class Foo<T> {
  final T x;
  const Foo(T this.x);
}


test(Foo<num> f) {
  switch(f) {
    case Foo(3): print("equal"); break;
    default: print("not equal"); break;
  }
}
  
void main() {
  test(const Foo(3));
}
```

That said, I don't suspect it's actually that useful in practice.  

One thing that we might want to think about is what we want to happen if we add user defined coercions.  That is, if we, say, added implicit constructors, and had an implicit constructor which constructed a `Foo` from an `int` if an `int` was used where a `Foo` was expected, what would we want:

```dart
test(Foo f) {
  switch(f) {
    case 3: ...
  }
}
```

to do?

In general, I don't love the coercions here.  I think my preference would probably be to say:

1) No coercions in refutable patterns
2) Yes downwards inference on constants in refutable patterns

leafpetersen (approved):

working/0546-patterns/patterns-feature-specification.md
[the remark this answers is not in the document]
lrhn: I'm not sure whether it's a good example, but consider:
```dart
switch (someConstantSet) {
  case const {1, 2}: ...
  case const {1}: ...
  case const {}: ...
}
```
Here the final `const{}` is actually a map. If we had used the matchee type as the context type of the constant, then it would be a set. (In practice, if you have a bunch of constant sets that you are looking for, they'll likely be named constants anyway.)

So, probably not going to happen in real code, but I see no *harm* in providing a context type. If we introduce more context-dependent literals in the future, we might get more examples.

It just feels *right* to provide the context type. Throwing away available information unnecessarily feels like something that will come back and haunt us later. Very much a gut feeling, nothing practical and concrete I can point to in the current language.

working/0546-patterns/patterns-feature-specification.md
munificent: Your preferences SGTM. Done.

working/0546-patterns/patterns-feature-specification.md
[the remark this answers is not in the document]
munificent: OK, using the context type SGTM.

working/0546-patterns/patterns-feature-specification.md
munificent: OK, so I as I understand it, this is no longer a breaking change because we're sticking with using the matched value's type as the context type for inference on the constant? This should still print "equal" followed by "not equal".

PR #2543, opened 2022-10-01
[pattern] Clarify variable scoping and shared case body rules.

### LINKED ISSUE #2473: Loosen the restriction on variables declared in case patterns that share a body

### LINKED ISSUE #2485: [Patterns] Do we need to specify scope overlap restrictions for patterns?
Patterns can bind and reference variables.  It's not entirely clear to me from the spec what the scoping is, e.g.:

```dart
const int x = 3;
const int y = 3;
void main() {
int q = 3;
switch (...) {
  case (x, var x): ... // Error?  Or ok?
  case (var x, x): ... // Error?  Or ok? 
  case (var y, var z): ... // Ok
  case (y, y): ... // Error (because declared in previous case)? Or ok?
  case(var p, var q): ... // Error, because q already declared in a local scope?  Or ok?
}
```  

Is this covered somewhere (e.g. by existing spec language, or by something that I missed), or do we need to clarify this?

cc @munificent

### LINKED ISSUE #2533: Patterns spec needs to specify an error for mismatch of "finality" within a pattern that declares variables:
There's nothing in the spec preventing this:

```dart
switch (...) {
  case (var x | final x): ...
}
```
We need some kind of compile-time error here, otherwise it's not clear whether the variable should be final or not.

munificent:
Fix #2473.
Fix #2485.
Fix #2533.

cc @johnniwinther @chloestefantsova @scheglov

### DIFF
--- a/working/0546-patterns/patterns-feature-specification.md
+++ b/working/0546-patterns/patterns-feature-specification.md
@@ -1751,19 +1751,21 @@ variables bound by a pattern can only be used when the pattern has matched,
 which means variables bound by refutable patterns must only be in scope in code
 that can't be reached when the match fails.
 
-Also, logical-or patterns and switch case fallthrough adds some complexity.
+Also, logical-or patterns and switch case fallthrough add some complexity.
 
 #### Pattern variable sets
 
 A *pattern variable set* specifies the set of variables declared by a pattern
 and its subpatterns when not in an assignment context. Each variable in the set
-has a unique name, a static type, and whether it is final or not. The pattern
-variable set for a pattern is:
+has a unique name, a static type (the declared or inferred type, but not its
+promoted type), and whether it is final or not. The pattern variable set for a
+pattern is:
 
 *   **Logical-or**: The pattern variable set of either branch. It is a
     compile-time error if the two branches do not have equal pattern variable
     sets. Two pattern variable sets are equal if they have the same set of names
-    and each corresponding pair of variables have the same type and finality.
+    and each corresponding pair of variables have the same finality and their
+    types are structurally equivalent after `NORM()`.
 
     *Since only one branch will match and we don't know which, for the pattern
     to have a stable set of variables with known types, the two branches must
@@ -1782,11 +1784,11 @@ variable set for a pattern is:
     with the same name. *A pattern can't declare the same variable more than
     once.*
 
-*   **Relational** or **constant**: An empty set.
+*   **Relational** or **constant**: The empty set.
 
 *   **Variable**:
 
-    1.  If the variable's identifier is `_` then an empty set.
+    1.  If the variable's identifier is `_` then the empty set.
 
     2.  Else a set containing a single variable whose name is the pattern's
         identifier and whose type is the pattern's required type (which may have
@@ -1859,6 +1861,9 @@ appears:
 
     The guard expression is evaluated in its case's case scope.
 
+    The then statement of an if-case statement is executed in the case's case
+    scope.
+
 #### Switch case fallthrough
 
 In a switch statement or expression, multiple cases may share the same body:
@@ -1875,13 +1880,11 @@ This would normally be a name collision, but we make an exception to allow it
 when it doesn't cause any problems.
 
 When compiling the body shared by a set of cases with pattern variable sets `vs`
-(where a default case has an empty pattern variable set), the enclosing scope is
-a synthesized scope `s` defined as:
+(where default cases and labels have empty pattern variable sets), the enclosing
+scope is a synthesized scope `s` defined as:
 
-1.  Let `n` be the union of the sets of names of all of the pattern variable
-    sets in `vs`.
-
-2.  For each name in `n`:
+1.  For each name in `n` appearing as a variable name in any of the pattern
+    variable sets in `vs`:
 
     1.  If `n` is defined in every pattern variable set in `vs` and has the same
         type and finality, then introduce `n` into `s` with that type and
@@ -1892,11 +1895,15 @@ a synthesized scope `s` defined as:
         types of `n` in the cases in the same way that promotions are merged at
         join points.
 
-    2.  Else `n` is inconsistently defined by the cases. Introduce `n` into `s`
-        with a sentinel type `Bad`.
+    2.  Else `n` is inconsistently defined by the cases. Introduce a variable
+        `n` into `s` (with arbitrary type and finality) and note that `n` is
+        an inconsistent variable.
 
-Compile the body using that scope. It is a compile-time error if any identifier
-in the body resolves to a variable in `s` whose type is `Bad`.
+Compile the body using static scope `s`. It is a compile-time error if any
+identifier in the body resolves to an inconsistent variable in `s`. *In other
+words, a inconsistently defined variable in in scope such that it shadows a
+valid variable with the same name in an outer scope, but it is an error to refer
+to it.*
 
 *Note that it is not a compile-time error for there to be inconsistently
 defined variables between cases. It's only an error to use them in the body.
@@ -1916,10 +1923,14 @@ switch (obj) {
 defined consistently by all cases.*
 
 At runtime, the enclosing scope is *only* the case scope of the case that
-matched. *This determines which specific variables in which case are actually
-bound at runtime. We can't use a single synthesized scope that is shared by all
-cases because closures in guard clauses should see each case as having its own
-variables:*
+matched. (If the body is reached from a `default` clause or by continuing to a
+label, the case scope is an empty scope whose enclosing scope is the scope
+surrounding the switch statement or expression.)
+
+*The runtime scope determines which specific variables in which case are
+actually bound at runtime. We can't use a single synthesized scope that is
+shared by all cases because closures in guard clauses should see each case as
+having its own variables:*
 
 ```dart
 bool capture(void Function() callback, bool result) {
--- a/working/0546-patterns/patterns-feature-specification.md
+++ b/working/0546-patterns/patterns-feature-specification.md
@@ -286,12 +286,8 @@ A logical-or pattern may match even if one of its branches does not. That means
 that any variables in the non-matching branch would not be initialized. To avoid
 problems stemming from that, the following restrictions apply:
 
-*   It is a compile-time error if one branch contains a variable pattern whose
-    name or type does not exactly match a corresponding variable pattern in the
-    other branch. These variable patterns can appear as subpatterns anywhere in
-    each branch, but in total both branches must contain the same variables with
-    the same types. This way, variables used inside the body covered by the
-    pattern will always be initialized to a known type.
+*   The two branches must define the same set of variables. This is specified
+    more precisely under "Variables and scope".
 
 *   If the left branch matches, the right branch is not evaluated. This
     determines *which* value the variable gets if both branches would have
@@ -1171,7 +1167,9 @@ We replace the existing `ifStatement` rule with:
 ifStatement ::= 'if' '(' expression caseHead? ')' statement ('else' statement)?
 ```
 
-**TODO: Allow patterns in if elements too.**
+**TODO: Allow patterns in if elements too ([#2542][]).**
+
+[#2542]: https://github.com/dart-lang/language/issues/2542
 
 When the `condition` has no `caseHead`, it behaves as it does today. If there is
 a `caseHead`, then the expression is evaluated and matched against the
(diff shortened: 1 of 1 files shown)


working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1748,65 +1746,196 @@ types of all of the case expressions.
...
-  case (String a): // Error.
-    break;
+This would normally be a name collision, but we make an exception to allow it
+when it doesn't cause any problems.
+
+When compiling the body shared by a set of cases with pattern variable sets `vs`
+(where a default case has an empty pattern variable set), the enclosing scope is
+a synthesized scope `s` defined as:
+
+1.  Let `n` be the union of the sets of names of all of the pattern variable
```
lrhn, line 1881: You used "union" above as disjoint union on "pattern variable sets", which may be different from "sets of names of pattern variables", but I think that distinction is probably so subtle it'll be lost on a reader.

Maybe introduce more precise terms: "pattern declaration union", "pattern declaration intersection".
Heck, even "join"/"merge" (probably too generic). As long is it doesn't overlap directly with set operations, when we do use set operations as themselves as well.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1748,65 +1746,196 @@ types of all of the case expressions.
...
+    sets in `vs`.
+
+2.  For each name in `n`:
+
+    1.  If `n` is defined in every pattern variable set in `vs` and has the same
+        type and finality, then introduce `n` into `s` with that type and
+        finality.
+
+        If any of the corresponding variables in `vs` are promoted, calculate
```
lrhn, line 1890: Do we have to say *where* they are promoted?

Like, the result of inference of a pattern is a set of variable declarations (name, declared type, finality) and some promotions (of those variables or existing ones).

Fx `if (y case int _ when x is int)` promotes a variable, `x`, which is not declared by the pattern. It may also promoote `y`, depending on how we specify it.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1748,65 +1746,196 @@ types of all of the case expressions.
...
+        the promoted type of the variable in `s` based on all of the promoted
+        types of `n` in the cases in the same way that promotions are merged at
+        join points.
+
+    2.  Else `n` is inconsistently defined by the cases. Introduce `n` into `s`
+        with a sentinel type `Bad`.
+
+Compile the body using that scope. It is a compile-time error if any identifier
```
lrhn, line 1898: that scope -> that static scope
(for emphasis).

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1748,65 +1746,196 @@ types of all of the case expressions.
...
+            when c == 2:
+        //       ^ OK.
+        print(c);
+        //    ^ OK.
+    }
+    ```
 
-The scope where a pattern's variables are declared and available for use depends
-on the construct that contains the pattern:
+    The guard expression is evaluated in its case's case scope.
```
lrhn, line 1860: This never actually says that the `then` branch of an `if`/`case` is evaluated in the case scope of the `case`.
(The switch case is handled in the next section, the `if` isn't.)

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1748,65 +1746,196 @@ types of all of the case expressions.
...
+  return result;
+}
+
+switch ([1]) {
+  case [int a] when capture(() => a++, false):
+  case [int a] when capture(() => a += 10, true):
+    print(n);
+}
+```
+
```
lrhn, line 1936: Mention that:
```dart
case [int a]:
default:
   break;
```
treats `default` as a case with zero declared variables.
(We can disallow combining `default` with anything else. Until now it was useless, but with `when` clauses with side effect, it might actually be useful now. Also, `default:` and `case _:` should probably just work the same.)

Same for labels:
```dart
label:
case [int a]: {
  ...
}
case [num a]: continue label;
```
which also allows a non-binding way into the case body.

lrhn (approved):

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1748,65 +1746,196 @@ types of all of the case expressions.
...
+    The union of a series of pattern variable sets is the union of their
+    corresponding sets of variable names. Each variable in the resulting set is
+    mapped to the corresponding variable's type and finality.
+
+    It is a compile-time error if any two sets being unioned have a variable
+    with the same name. *A pattern can't declare the same variable more than
+    once.*
+
+*   **Relational** or **constant**: An empty set.
```
stereotype441, line 1785: Nit: I would say "the empty set" here (and on line 1789) rather than "an empty set".  In math-speak, there is only one empty set.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1748,65 +1746,196 @@ types of all of the case expressions.
...
+```
 
-  case [int a]:
-  case (String a): // Error.
-    break;
+This would normally be a name collision, but we make an exception to allow it
+when it doesn't cause any problems.
+
+When compiling the body shared by a set of cases with pattern variable sets `vs`
+(where a default case has an empty pattern variable set), the enclosing scope is
```
stereotype441, line 1878: Should also mention that a label has an empty pattern variable set.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1748,65 +1746,196 @@ types of all of the case expressions.
...
+        type and finality, then introduce `n` into `s` with that type and
+        finality.
+
+        If any of the corresponding variables in `vs` are promoted, calculate
+        the promoted type of the variable in `s` based on all of the promoted
+        types of `n` in the cases in the same way that promotions are merged at
+        join points.
+
+    2.  Else `n` is inconsistently defined by the cases. Introduce `n` into `s`
+        with a sentinel type `Bad`.
```
stereotype441, line 1896: I would recommend checking with the implementation teams on the feasibility of this.  I believe the analyzer team has discussed introducing a sentinel type `Bad` in the past to aid in error recovery, and has thus far not had time to do so.

stereotype441 (approved):
lgtm, with some nits that I'm happy to leave to your judgment

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1748,65 +1746,196 @@ types of all of the case expressions.
...
 
-Patterns often exist to introduce new variable bindings. A "wildcard" identifier
-named `_` in a variable pattern never introduces a binding.
-The variables a patterns binds depend on what kind of pattern it is:
+Patterns often exist to bind new variables. The language must ensure that the
+variables bound by a pattern can only be used when the pattern has matched,
+which means variables bound by refutable patterns must only be in scope in code
+that can't be reached when the match fails.
+
+Also, logical-or patterns and switch case fallthrough adds some complexity.
```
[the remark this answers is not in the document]
munificent: Done.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1748,65 +1746,196 @@ types of all of the case expressions.
...
+
+#### Pattern variable sets
+
+A *pattern variable set* specifies the set of variables declared by a pattern
+and its subpatterns when not in an assignment context. Each variable in the set
+has a unique name, a static type, and whether it is final or not. The pattern
```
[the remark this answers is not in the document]
munificent: Done.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1748,65 +1746,196 @@ types of all of the case expressions.
...
+variable set for a pattern is:
 
-*   **Logical-or**: Does not introduce variables but may contain subpatterns
-    that do. It is a compile-time error if the two subpatterns do not introduce
-    the same variables with the same names and types.
+*   **Logical-or**: The pattern variable set of either branch. It is a
+    compile-time error if the two branches do not have equal pattern variable
+    sets. Two pattern variable sets are equal if they have the same set of names
+    and each corresponding pair of variables have the same type and finality.
```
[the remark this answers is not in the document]
munificent: Done.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1748,65 +1746,196 @@ types of all of the case expressions.
...
 
-    **TODO: "Same name" is underspecified and also probably
-    [too conservative][2473].**
+    *Since only one branch will match and we don't know which, for the pattern
+    to have a stable set of variables with known types, the two branches must
+    define the same variables. This way, uses of the variables later will have
+    a known type and finality regardless of which branch matched.*
```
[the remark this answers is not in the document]
munificent: Yeah, I think the intersection is feasible but too weird to be useful.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1748,65 +1746,196 @@ types of all of the case expressions.
...
 
 *   **Logical-and**, **cast**, **null-check**, **null-assert**,
-    **parenthesized**, **list**, **map**, **record**, or **extractor**: These do
-    not introduce variables themselves but may contain subpatterns that do.
+    **parenthesized**, **list**, **map**, **record**, or **extractor**: The
+    union of the pattern variable sets of all of the subpatterns.
```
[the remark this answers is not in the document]
munificent: I think the third paragraph clarifies this.

working/0546-patterns/patterns-feature-specification.md
munificent: Done.

working/0546-patterns/patterns-feature-specification.md
munificent: Heh, oops. Added.

working/0546-patterns/patterns-feature-specification.md
munificent: Ugh labels. Good catch.

working/0546-patterns/patterns-feature-specification.md
munificent: Reworded.

working/0546-patterns/patterns-feature-specification.md
munificent: I think it's implicit here that the variable is promoted in the scope where it gets added. But I strongly suspect that all of the type promotion stuff will end up needing more work anyway, so it's probably OK if this is a little vague for now?

working/0546-patterns/patterns-feature-specification.md
munificent: My intent wasn't to actually require implementers implement it this way (but it does certainly read that way). Reworded to try to be more abstract. The main goal was just to ensure that colliding variables still end up in the scope instead of disappearing and allowing an outer shadowed variable to shine through.

working/0546-patterns/patterns-feature-specification.md
munificent: Done.

working/0546-patterns/patterns-feature-specification.md
munificent: Done.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1748,65 +1746,207 @@ types of all of the case expressions.
...
+```dart
+bool capture(void Function() callback, bool result) {
+  callback();
+  return result;
+}
+
+switch ([1]) {
+  case [int a] when capture(() => a++, false):
+  case [int a] when capture(() => a += 10, true):
+    print(n);
```
leafpetersen, line 1944: `print(a)`?

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1748,65 +1746,207 @@ types of all of the case expressions.
...
+appears:
+
+*   **Pattern variable declaration**: The scope enclosing the variable
+    declaration statement. *This will be either a function body scope or a block
+    scope.*
 
-*   **Relational** or **constant**: These do not introduce any variables.
+    The *initializing expression* for every variable in the pattern is the
+    pattern variable declaration's initializer. *This means all variables
+    defined by a pattern are in scope beginning at the top of the surrounding
```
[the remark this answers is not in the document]
munificent: > I don't understand how the previous sentence means this.

The language spec says about local variables: "It is a compile-time error if a local variable is referenced at a source code location that is before the end of its initializing expression, if any, and otherwise before the declaring occurrence of the identifier which names the variable."

Since pattern variables are local variables, I defined it in terms of that.

> More generally, I continue to find this ("scope beginning at the top of the surrounding") formulation of scoping very odd, 

It's odd to me too, but I figure it's easiest on the implementers to be consistent with how the language already specifies local scopes.

> and it doesn't match up with the way that the implementations talk about scope when reporting errors.

Yeah, that is interesting. I don't know why the implementations present it differently from how the language does. But if we want to change this, it probably makes sense to change it consistently throughout the language spec and not just in this proposal.

working/0546-patterns/patterns-feature-specification.md
munificent: Oops, fixed.

working/0546-patterns/patterns-feature-specification.md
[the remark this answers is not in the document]
leafpetersen: > Since pattern variables are local variables, I defined it in terms of that.

Ack.

>  I don't know why the implementations present it differently from how the language does.

I mean, I think it's fairly obvious - the resulting errors would be incomprehensible.  If the compiler error for your example below talked about `c` being declared at the beginning of the test function, and that it was an error to refer to a variable before it was declared (where it was declared) but after it was declared (where the spec says it's declared) I'd throw up my hands and go find a new compiler.  :)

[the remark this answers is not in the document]
munificent: For what it's worth, if you throw this program at a C# compiler:

```c#
public class Program
{
	static String s = "field";
	
	public static void Main()
	{
		Console.WriteLine(s);
		var s = "local";
	}
}
```

It says:

```
Compilation error (line 9, col 21): Cannot use local variable 's' before it is declared. The declaration of the local variable hides the field 'Program.s'.
```

So it treats the `s` in `Console.WriteLine(s);` as referring to the one on the next line, like the Dart spec does. 🤷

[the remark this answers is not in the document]
leafpetersen: I don't object to making it an error (well, I actually do somewhat, but I'm basically fine with it).  And that error message is perfectly fine.  It says exactly what I would expect:
  - It's an error to use a variable before it is declared
  - Declaring a variable in a scope hides variables with the same name from the outer scope

But I don't understand why the Dart spec insists that the reason for this is that `s` is actually in scope for the whole block, but for the first part of the block up until its declaration point it's an error to refer to it.  What does it even mean to be in scope then?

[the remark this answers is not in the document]
stereotype441: My understanding (of both the spec and the analyzer implementation) is that the meaning of "in scope" is "candidate for a bare identifier to resolve to".  So if we changed the spec to say that a variable is not in scope until the point of its declaration (or after its declaration), then this program:
```dart
String s = 'top level';
void main() {
  print(s);
  var s = 'local';
}
```
would change from being an error to being valid, because at the point of `print(s)`, the only `s` that would be in scope would be the top level `s`.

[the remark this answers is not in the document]
leafpetersen: > would change from being an error to being valid

No it really wouldn't.  Unless you wanted it to, I guess.  Look, one way or the other, you have to specify where I am allowed to use variables.  And there's a very well established notion of lexical scope out there which you can lean on when defining a programming language.  But the Dart spec doesn't use it.  Instead, it says:
 -  a variable's scope doesn't start in the usual place, instead it begins at the start of the block, and therefore all references to a variable within the block refer to that variable!.
 -  But!  You can't use those variables, even though they're in scope, until you get to the point where they are declared! (generalize "point where they are declared" as needed).  
 Which works, I guess?  But it seems very convoluted to me.  Instead, if you want C#/Dart semantics, it seems to me much cleaner to say:
  - Scoping works as normal: it starts at the point you declare the variable, and continues to the end of the block (generalize as needed).
  - If a variable v is declared in a block, it is an error to refer to a variable with the same name declared in a scope outside the block.

I think that (maybe with a couple of tweaks) gives you everything you need to get the semantics you describe above, and makes the error messages given by the C# compiler and the Dart analyzer fall out very nicely (the CFE error messages for these examples are less than ideal).  It's much simpler to explain, and it's no more verbose.  So why the circumlocution?  It feels like a hold over from Javascript, where variables really are function scoped (with no restrictions around reference, IIRC correctly).  So it's like Dart kind of took the JS scoping, and then added a static error to get back to something that's basically like what you'd get if you'd just started with the standard scoping rules from the start.  It works, I guess, but I continue to find it weird.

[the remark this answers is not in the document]
stereotype441: Sure, we could totally change the spec in the way you suggest.  I could live with that.  I wasn't really trying to push for any particular position, just trying to answer the question you posed above ("What does it even mean to be in scope then?").  To me, personally, it makes sense.  And it aligns well with how the implementation currently works.

FWIW, the C# spec defines the scope of a local variable similarly to the way Dart currently does (from https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/basic-concepts#771-general):

> The scope of a local variable declared in a *local_variable_declaration* (§12.6.2) is the *block* in which the declaration occurs.

And then in https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/statements#1262-local-variable-declarations it says:

> The scope of a local variable declared in a *local_variable_declaration* is the block in which the declaration occurs. It is an error to refer to a local variable in a textual position that precedes the *local_variable_declarator* of the local variable.

Whereas the C++ spec defines the scope of local variables in the way that I gather you would prefer:

> The potential scope of a name declared in a block (compound statement) begins at the point of declaration and ends at the end of the block.

Notably, however, in C++, if a variable v is declared in a block, it is *not* an error to refer to a variable with the same name declared in a scope outside the block.  In other words, this is totally ok:

```c++
int main() {
  int i = 0;
  {
    std::cout << i << std::endl; // Ok, prints `0`
    int i = 1; // Scope of inner `i` begins here
    std::cout << i << std::endl; // Ok, prints `1`
  }
}
```

Whereas doing the equivalent thing in C#, Dart, or Javascript *is* an error.  (In Javascript it's a runtime error; in C# and Dart it's a compile-time error).

So I guess I would quibble with this statement:

> ...there's a very well established notion of lexical scope out there which you can lean on when defining a programming language. But the Dart spec doesn't use it.

IMHO, the fact that Dart, Javascript, and C# all use the same approach is enough to convince me that Dart's approach isn't crazy-pants insane 😄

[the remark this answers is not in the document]
leafpetersen: > Whereas the C++ spec defines the scope of local variables in the way that I gather you would prefer:

I do prefer these semantics (and by the way, it's not just C++, it's at least C, C++, Java, Kotlin, Swift, the ML family and others), but that's irrelevant to this discussion.  I wasn't complaining about the set of errors, I was complaining about the manner in which the set of errors were specified.

> So I guess I would quibble with this statement:

Well, then, I'm going to quibble back.  :). This is really off the track here, so I'm going to stop quibbling after this, but... I really think it's hard to argue that the interpretation of the word scope that Dart and C# are taking is standard.  It may not be crazypants, but it's not a standard interpretation of the word "scope" when referring to a variable.  Wikipedia certainly is not the last word on anything, but I'll quote the introductory paragraph from their entry on "Scope" as being somewhat indicative of the general understanding of the word:

> In [computer programming](https://en.wikipedia.org/wiki/Computer_programming), the scope of a [name binding](https://en.wikipedia.org/wiki/Name_binding) (an association of a name to an entity, such as a [variable](https://en.wikipedia.org/wiki/Variable_(programming))) is the part of a [program](https://en.wikipedia.org/wiki/Computer_program) where the name binding is valid; that is, where the name can be used to refer to the entity. In other parts of the program, the name may refer to a different entity (it may have a different binding), or to nothing at all (it may be unbound).

This is what I would call the standard understanding of what variable scope means.  And I would definitely say that it's a bit non-standard to say that the scope of a variable is not just the places where the name is valid and can be used to refer to the entity to which it is bound, but also some other places, where the variable is still considered in scope, but is not valid, and cannot be used to refer to the entity to which it is bound.  I get that it can be done technically.  I don't think it's the end of the world to abuse the word "scope" in this way.  But... it's hard for me not to  view it as somewhat of an abuse of terminology.  

Stepping aside from the terminology, and in all seriousness, I do think this is an example of the kind of "cleverness" that the Dart spec uses in ways that I don't think serves it well.  It's a confusing (but clever!) way of achieving a reasonable end goal.  That is, the designers presumably wanted to make it an error to refer to a variable from an outer scope in a block where the same variable name was later declared.  Rather than just say that (clear, simple, easy to understand), they did something clever.  They said "ah!", we will just say that the scope of a variable *is* the block, and therefore any references to that variable anywhere in the block refer to the locally declared variable (and then we will separately say that it is an error to refer to a variable before it is declared).  It's fine, it gets us to the same place.  But it's such a more convoluted way to end up at the same place, and it relies on redefining terminology to mean something other than the usual meaning.  If I were implementing Dart based on this specification, I think I could very easily mis-interpret this ("oh yeah, block scope, I know what that means, wait why are my conformance tests failing?").  Again, it's not the end of the world, and I'm certainly not planning (yet) to go rewrite the Dart spec, but yes, it is one of the many things in the Dart spec that make me want to take a year to just go rewrite the whole damn thing from scratch.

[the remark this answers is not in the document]
lrhn: I'm pretty sure the Dart "scope" specification is an amalgam of copying JavaScript/C# behavior and some amount of laziness. Catching shadowing is probably at most the third rationale for it.

The definitions of "scopes" have always been somewhat vague (what is a static scope, what is a runtime scope, how do they interact), and by only introducing new scopes at block boundaries, at least there were *fewer* scopes to care about. (And they were entirely block structured and nested.)

It is simpler to specify. Maybe not much (but I honestly don't know).
It may or may not be simpler to implement (most likely it is exactly the same, but maybe it requires slightly fewer "Scope" objects internally in the compiler).
It rarely affects users, and if it does, it can save them from weird mistakes, say moving the declaration to after a use, and then suddenly having it refer to something outside the block. But that feels like an argument added after the choice was made, not a guiding principle.

Doing variable scopes like Java does (even without their rule against redeclaring local variables) would be more complicated. I'm sure the original Dart spec authors would know *precisely* how complicated.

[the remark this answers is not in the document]
munificent: For what it's worth, I think many programmers coming from C tradition languages have an intuition that a block forms a single scope. That was coherently true back in the early C days where all variable declarations had to be at the beginning of a block. When C++ (and later C) let you declare variables in the middle of a block, that created a new question as to whether the parts before and after that variable declaration are the "same" scope.

I interpret Dart (and languages with similar specs) as trying to hew to that original understanding of scope = entire block while avoiding weirdness with shadowing.

The ML family comes from any entirely different tradition where `let v = e in ...` makes it explicitly clear that the thing after `in` is its own scope nested inside the scope preceding it.

I have no horse in this race. I think either way of specifying it is reasonable. It's low on the list of things I would change if I were rewriting the spec. :)

PR #2560, opened 2022-10-07
[views] Add rule that R is a subtype of V inside view V(R r)

eernstg:
This PR makes the representation type (`R` in `view V(R r) {}`)  a subtype of the view type (`V`) for code in the body of the view declaration. The rationale is that it is useful for _some_ code to be able to perform "mass adoption" of the view (like `List<V> views = <R>[...];`) safely (without a cast). The danger associated with this mechanism is that it allows for any number of objects to implicitly be given the view type, without invoking any constructors. That's the reason why the subtype relation is not enabled for all code. It is up to the maintainers of the view class to ensure that such mass adoptions of the view are appropriate.

@stereotype441, @natebosch, @jakemac53, @munifient, @lrhn, @kallentu, Leaf and I discussed this yesterday, and it seems to be a useful change.

### DIFF
--- a/working/1426-extension-types/feature-specification-views.md
+++ b/working/1426-extension-types/feature-specification-views.md
@@ -7,6 +7,12 @@ Status: Draft
 
 ## Change Log
 
+2022.10.07
+  - Made a view class whose representation type is a top type a proper
+    subtype of `Object?`.
+  - Changed subtyping inside a view class `V` such that the
+    representation type is a subtype of the enclosing view type.
+
 2022.09.22
   - Removed support for `export`, changed `view` to `view class`.
 
@@ -641,17 +647,30 @@ Let `V` be a view type of the form
 <code>View&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;</code>,
 and let `R` be the corresponding instantiated representation type.
 `V` is a proper subtype of `Object?`. If `R` is non-nullable then `V`
-is a proper subtype of `Object` as well.
+is a proper subtype of `Object` as well. 
 
 *That is, an expression of a view type can be assigned to a top type
 (like all other expressions), and if the representation type is
 non-nullable then it can also be assigned to `Object`. Non-view types
-(except bottom types) cannot be assigned to view types without a cast.*
+(except bottom types) cannot be assigned to view types without a cast,
+except in the case mentioned below.*
 
-In the body of a member of a view declaration _DV_ named `View`
-and declaring the type parameters
-<code>X<sub>1</sub>, .. X<sub>s</sub></code>,
-the static type of `this` is
+Let _DV_ be a view class declaration named `View` with type parameters
+<code>X<sub>1</sub>, .. X<sub>s</sub></code>.
+In the body of _DV_, the representation type is a subtype of the view type
+<code>View&lt;X<sub>1</sub>, .. X<sub>s</sub>&gt;</code>.
+
+*This means that the body of the view class itself has an extra power,
+compared to code outside the view class: It can assign values of the
+representation type `R` to a variable of the view type `V` without a
+cast. "Lifted" versions of this power follow from the normal subtype
+rules, e.g., it can also assign a value whose type is `List<R>` to a
+variable of type `List<V>`. In other words, the body of the view class
+declaration has the ability to "wrap the entire list in the view" in
+time O(1) and without a cast. It is up to the maintainers of the view
+class to make sure that this is an appropriate thing to do.*
+
+In the body of a member of _DV_, the static type of `this` is
 <code>View&lt;X<sub>1</sub> .. X<sub>s</sub>&gt;</code>.
 The static type of the name of the representation name is the
 representation type.


working/1426-extension-types/feature-specification-views.md
```diff
@@ -641,17 +647,30 @@ Let `V` be a view type of the form
 <code>View&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;</code>,
 and let `R` be the corresponding instantiated representation type.
 `V` is a proper subtype of `Object?`. If `R` is non-nullable then `V`
-is a proper subtype of `Object` as well.
+is a proper subtype of `Object` as well.
```
lrhn, line 650: Trailing space :)

working/1426-extension-types/feature-specification-views.md
```diff
@@ -641,17 +647,30 @@ Let `V` be a view type of the form
...
-and declaring the type parameters
-<code>X<sub>1</sub>, .. X<sub>s</sub></code>,
-the static type of `this` is
+(except bottom types) cannot be assigned to view types without a cast,
+except in the case mentioned below.*
+
+Let _DV_ be a view class declaration named `View` with type parameters
+<code>X<sub>1</sub>, .. X<sub>s</sub></code>.
+In the body of _DV_, the representation type is a subtype of the view type
+<code>View&lt;X<sub>1</sub>, .. X<sub>s</sub>&gt;</code>.
```
lrhn, line 661: Hint: You can use `\<` instead of `&lt;` to escape the tag-start meaning of `<` in markdown.

working/1426-extension-types/feature-specification-views.md
```diff
@@ -641,17 +647,30 @@ Let `V` be a view type of the form
...
-In the body of a member of a view declaration _DV_ named `View`
-and declaring the type parameters
-<code>X<sub>1</sub>, .. X<sub>s</sub></code>,
-the static type of `this` is
+(except bottom types) cannot be assigned to view types without a cast,
+except in the case mentioned below.*
+
+Let _DV_ be a view class declaration named `View` with type parameters
+<code>X<sub>1</sub>, .. X<sub>s</sub></code>.
+In the body of _DV_, the representation type is a subtype of the view type
```
lrhn, line 660: What does "body" mean? Does it affect type parameters on the view itself, or type arguments in the `extends`/`implements` clauses? Or is it only the member declarations? 
For the member declarations, does it affect their public API, or is it just inside *their* bodies?

working/1426-extension-types/feature-specification-views.md
lrhn, line 661: Our type system has so far not been defined to depend on "where" the test is made. Which mans that we do not have a terminology or methodology which assigns a location to type tests in general.

Because of that, I'm wary about just writing this and letting implementors figure out what it means.

For example:
```dart
typedef F<X extends ViewType> = ...;
view class ViewType(int Object) {
   F<Object> foo() => ...; // Is this valid? (What if it was inside the function body instead)?
   void bar(ViewType v) => ...;
}
Object o = Object();
ViewType(o).bar(o); // Is this valid?
```
That is, is the type check of the type argument application performed at the application or at the declaration?
Does the argument type check apply the subtype relation of the call site or of the declaration site?

Also, by picking `Object` as the representation type, I've created two mutual subtypes which are not equal, and a new supertype of `Object`. (That's not unique, `FutureOr<Object>` is another example.)
We should make absolutely sure our type inference implementation can handle that. (It's probably fine).

A formal specification approach could be that we have a notion of "type system" which includes the subtype relation, and the static and runtime semantics of Dart every construct is parameterized by such a type system. Then the semantics of the a view type would pass down a modified type system to the body. The specification would then be explicit about which type system it uses for which checks. Without a specification at that level of precision, I'd be very, very concerned about saying that subtyping depends on "location".

working/1426-extension-types/feature-specification-views.md
```diff
@@ -641,17 +647,30 @@ Let `V` be a view type of the form
...
+compared to code outside the view class: It can assign values of the
+representation type `R` to a variable of the view type `V` without a
+cast. "Lifted" versions of this power follow from the normal subtype
+rules, e.g., it can also assign a value whose type is `List<R>` to a
+variable of type `List<V>`. In other words, the body of the view class
+declaration has the ability to "wrap the entire list in the view" in
+time O(1) and without a cast. It is up to the maintainers of the view
+class to make sure that this is an appropriate thing to do.*
+
+In the body of a member of _DV_, the static type of `this` is
```
lrhn, line 673: _Unpopular opinion: We should just make `ViewType(representationValue)` always do the zero-cost, static-checked conversion to the view type, by forcing you to use the primary constructor notation to declare it, instead of allowing it to be a generalized constructor that does things on top. That also makes it obvious that any value can be come a member of the view._

working/1426-extension-types/feature-specification-views.md
eernstg: This subtype relation should hold for the bodies of member declarations. Clarified that.

working/1426-extension-types/feature-specification-views.md
eernstg: OK, thanks!

eernstg: > Our type system has so far not been defined to depend on "where" the test is made

Right, this is a good place to take care. ;-)

We are checking subtype relationships for constructs (like expressions, if-statements, variable declarations with initializers, etc.) that occur at a specific location. When a specific subtype relationship is required to hold, we check the location and decide whether this particular rule in the subtype relation is enabled.

The subtype relationship between the representation type and the view type allows certain constructs to be type correct in the body of a view member declaration, even though it may be a compile-time error if it occurred elsewhere. Each construct is type checked according to the location where it occurs. For instance:

```dart
...
// This was outside the declaration of `View`, hence an error.
ViewType(o).bar(o);
```
A member signature like foo is an error, too:

```dart
typedef F<X extends ViewType> = ...;
view class ViewType(int Object) {
   F<Object> foo() => ...; // Is this valid? (What if it was inside the function body instead)?
   void bar(ViewType v) => ...;
}
```

The application of `F` to `Object` is an error because `Object <\: ViewType` at that location (where `<\:` means 'is not a subtype'), so the declaration of `foo` is an error.

stereotype441: For the record, I'm also concerned about adding location-awareness to our type system.  I worry partly because it causes extra work for the implementation teams, and partly because I don't have a good mental model on which attributes of our type system are necessary for soundness.  Are we creating a risk, for example, that "instantiate to bounds" might produce a type that is valid in one code location, but when picked up by code elsewhere (e.g. via type inference) is no longer a valid type?  I feel like it would take a really big benefit to justify complicating the type system in this way, and I'm not convinced that the ability to do stuff like `List<V> views = <R>[...];` is that big a benefit.

And even if you convince me that being able to do things like `List<V> views = <R>[...];` is *really* important, I still wonder if there are other ways to do it that aren't so dangerous to the type system.  Like, for example, we could say that within the body of a view, at the same places where we perform other implicit conversions like `.call` tearoff and downcast from `dynamic`, we also allow an implicit conversion where one or more `R`s in covariant positions are replaced with `V`s.  Implicit conversions are way less scary to me than changing the definition of subtyping, because they don't affect any of the soundness guarantees of Dart.  Effectively we would just be saying "here is a situation where we've decided not to require you to write an explicit cast."

working/1426-extension-types/feature-specification-views.md
stereotype441: For the record, this "unpopular opinion" is very similar to what I was (unsuccessfully) arguing for in Wednesday's language meeting.  So maybe it's not as unpopular as you think 😃

working/1426-extension-types/feature-specification-views.md
lrhn: Implicit coercion makes much more sense.

Say we instead state that at any place where we check for assignability between a subexpression and a context type inside the body of a member of a view declaration, if the subexpression's static type is *not* a subtype of the context type, but it *would be* if the representation type of the surrounding view type was a subtype of the surrounding view type, then we implicitly side-cast the static type to the context type.

That's still worrisome in other ways. A composite type would lose precision, say:
```dart
Map<ViewType, num>? m = <RepresentationType, int>{};
```
If we just cast to the *context type*, we lose the non-nullability, so we forget to promote.

The "more proper" way would be to go through the static type's structure and replace individual representation types occurring in it with the view type. Even that can get weird, and might fail:
```dart
class C<S, T> {}
class D<R> extends C<R, R> {}
view class V(Object int) { 
  foo() {
    C<V, int>? v1 = D<int>();  // Can't do `as D<V>`, because that's also not valid.
  }
}
```

I guess the solution is simply to not cast, simply say that it *is* assignable, at its original static type.
But then we get into how that type affects inference further up the evaluation stack.

We really do need to have a local type system which *consistently* treats the representation as a subtype of the view type, and make sure we apply that type system to any check applied by type inference inside the view method bodies.

There is also generics. What is the "view type" and what is the "view type" when one or both is a generic type.
Can we have subtype relations between only some instantiations of a type:
```dart
view class B(List<int> f) { 
  void foo() { B x = f; }
}
```
Here we say that `List<int>` is a subtype of `B`, but `List<num>` is not, even if `B` is not generic. I'm not sure our type system is ready for that. It's as if `List<int>` implements an interface that `List<num>` does not.
(But if we allowed subtype relations between the view and the representation type in general, we'd have the same problem, that's not just because it's a *local* subtype extension.)

leafpetersen:
I'm generally a bit skeptical of the bang for the buck here.  It's a *lot* of complexity to avoid a cast or two.

natebosch:
I'd prefer to err on the side of safety and explicit syntax. I don't see the advantage in making some usage in view definitions less safe.

Issue #2554, opened 2022-10-05
[Patterns] Inferred type of variable patterns in list pattern


scheglov:
Why do we infer `var a` as `Object? a`, not `int a`?
We said that we want `List<int>` with `<int>[var a]`.

```dart
  test_matchObject_withTypeArguments_variable_untyped() async {
    await assertNoErrorsInCode(r'''
void f(Object x) {
  switch (x) {
    case <int>[var a]:
      break;
  }
}
''');
    final node = findNode.switchPatternCase('case').pattern;
    assertResolvedNodeText(node, r'''
ListPattern
  typeArguments: TypeArgumentList
    leftBracket: <
    arguments
      NamedType
        name: SimpleIdentifier
          token: int
          staticElement: dart:core::@class::int
          staticType: null
        type: int
    rightBracket: >
  leftBracket: [
  elements
    VariablePattern
      keyword: var
      name: a
      declaredElement: hasImplicitType a@53
        type: Object?
  rightBracket: ]
  requiredType: List<int>
''');
  }
```

lrhn:
Looks like a bug. I hink it should give `int`.

If there is no type argument on the list pattern, we should use the element type of the static type of the matched value as the matchee type used to do type inference on the element patterns.
If there is a type argument, we should probably use it instad.

@munificent

lrhn:
Found this in the specification:

> ```dart
> var <int>[x] = ...
> ```
> 
> Here, we infer `int` for the inner `x` subpattern based on the type of the
> surrounding list pattern.

so that behavior is at least documented as being intended.

munificent:
Fixed by #2569.

PR #2569, opened 2022-10-11
Update type-checking of patterns to use type arguments.

lrhn:
Context: https://github.com/dart-lang/language/issues/2554

If a list/map pattern has an explicit type argument, use that as the element/key/value types used to match sub-patterns against, not the element type of the matched value.

This makes a difference for something like:
```dart
var <num>[x] = <int>[1]; /*1*/
var <num>[y] = <int>[1] as dynamic; /*2*/
```
where the declared type of `x` will be taken from the `<num>` instead of being `int` for `/*1*/`, and the declared type of`y` will also be `num` instead of `dynamic`.

Some wording tweaks.

### DIFF
--- a/working/0546-patterns/patterns-feature-specification.md
+++ b/working/0546-patterns/patterns-feature-specification.md
@@ -45,7 +45,12 @@ Records are sort of like "first class argument lists" and give you a natural
 way to return multiple values:
 
 ```dart
-(double, double) geoCode(String city) {  var lat = // Calculate...  var long = // Calculate...  return (lat, long); // Wrap in record and return.}
+(double, double) geoCode(String city) {
+  var lat = // Calculate...
+  var long = // Calculate...
+
+  return (lat, long); // Wrap in record and return.
+}
 ```
 
 ### Destructuring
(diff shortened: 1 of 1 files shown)


working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1684,30 +1677,25 @@ To type check a pattern `p` being matched against a value of type `M`:
...
-        3.  Else let `F` be `Object?`. *The field subpattern will only be
-            matched at runtime if the value does turn out to be a record with
-            the right shape where the field is present, so it's safe to just
-            assume the field exists when type checking here.*
+      4.  Type-check `s` using `F` as the matched value type, and find its required type.
 
-        4.  Type-check `f` using `F` as the matched value type.
-
-    3.  The required type of `p` is a record type with the same shape as `p` and
+  2.           The required type of `p` is a record type with the same shape as `p` and
```
munificent, line 1698: The spacing after "2." here is... weird.

As much as possible, I try to format the Markdown in the proposal consistently (using +4 indentation for blocks) and it might be best to try to stick with that.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1718,20 +1706,19 @@ To type check a pattern `p` being matched against a value of type `M`:
...
+        2.  Type check `f` with `G` as the matched value type, to find its required type.
+3.  The required type of `p` is `X`.
 
 It is a compile-time error if:
 
 *   The type of an expression in a guard clause is not assignable to `bool`.
-
-*   `p` is in an irrefutable context, it has a required type `T`, and `M` is not
-    assignable to `T`. *Destructuring and variable patterns can only be used in
+*   A pattern `p` is in an irrefutable context (directly or nested as a subpattern),
```
munificent, line 1719: The proposal does specify that subpatterns are already in the same context as the outermost pattern so this parenthetical is redundant, but not harmful:

> *   **Declaration context.** The pattern in `localVariableDeclaration`,
    `forLoopParts`, or any of its subpatterns. ...
> *   **Assignment context.** The pattern in a `patternAssignment` or any of its
    subpatterns. ...
> *   **Matching context.** The pattern in a `caseHead` or any of its subpatterns. ...

munificent (approved):
Thanks!

working/0546-patterns/patterns-feature-specification.md
lrhn: Might as well remove this to be consistent.

(I'm trying to find a way to nest a pattern inside another pattern, without it being a subpattern, but we don't allow general expressions inside a pattern, only constants which cannot contain patterns, so it probably can't happen.)

PR #2589, opened 2022-10-27
[patterns] Support "..." rest patterns in list and map patterns.

### LINKED ISSUE #2453: Support "..." in list and map patterns
The patterns proposal has a TODO to allow `...` in list and map patterns. Currently, both patterns only match if the object has exactly the same length as the number of element/entry subpatterns. Adding `...` would loosen that to match if the object has *at least* that many elements/entries.

Also, `...` could have element patterns after it in a list pattern in order to extract trailing elements from the list while ignoring elements in the middle.

munificent:
Fix #2453.

### DIFF
--- a/working/0546-patterns/patterns-feature-specification.md
+++ b/working/0546-patterns/patterns-feature-specification.md
@@ -594,8 +594,7 @@ It is a compile-time error if:
     don't have primitive equality, it is possible to have redundant keys and the
     compiler won't detect it.*
 
-*   There is more than one `...` element in the map pattern. *It can
-    appear anywhere in the map, but there can only be zero or one.*
+*   There is more than one `...` element in the map pattern.
 
 *   The `...` element is not the last element in the map pattern.
 
@@ -1565,7 +1564,7 @@ The context type schema for a pattern `p` is:
         //                                 ^- Infers List<int>.
         ```
 
-*   **Relational** or **cast**: The context type schema is `Object?`.
+*   **Relational** or **cast**: The context type schema is `?`.
 
 *   **Parenthesized**: The context type schema of the inner subpattern.
 
@@ -1573,9 +1572,7 @@ The context type schema for a pattern `p` is:
 
     1.  If `p` has a type argument, then `E` is the type argument.
 
-    2.  Else if `p` has no elements then `E` is `Object?`. *If the pattern
-        doesn't destructure anything, it matches any list, so it is permissive
-        with the context type.*
+    2.  Else if `p` has no elements then `E` is `?`.
 
     3.  Else, infer the type schema from the subpatterns:
 
@@ -1584,18 +1581,18 @@ The context type schema for a pattern `p` is:
         2.  For each subpattern `e` in `p`:
 
             1.  If `e` is a rest element pattern with a subpattern `s` and the
-                context type schema of `s` is a `List<T>` for some type `T`,
-                then add `T` to `es`.
+                context type schema of `s` is an `Iterable<T>` for some type
+                `T`, then add `T` to `es`.
 
             2.  Else if `e` is not a rest element pattern, add the context
                 type schema of `e` to `es`.
 
-            *Else, `e` is a rest element with no subpattern, so it doesn't
-            contribute to inference.*
+            *Else, `e` is a rest element without an iterable element type, so it
+            doesn't contribute to inference.*
 
-        3.  If `es` is empty, then `E` is `Object?`. *This can happen if the
-            list pattern contains only a rest element which doesn't have a
-            context type schema that is known to be a `List<T>` for some `T`,
+        3.  If `es` is empty, then `E` is `?`. *This can happen if the list
+            pattern contains only a rest element which doesn't have a context
+            type schema that is known to be an `Iterable<T>` for some `T`,
             like:*
 
             ```dart
@@ -1620,9 +1617,7 @@ The context type schema for a pattern `p` is:
 
     1.  If `p` has type arguments then `K`, and `V` are those type arguments.
 
-    2.  Else if `p` has no entries, then `K` and `V` are `Object?`. *If the
-        pattern doesn't destructure anything, it matches any map, so it is
-        permissive with the context type.*
+    2.  Else if `p` has no entries, then `K` and `V` are `?`.
 
     3.  Else `K` is the least upper bound of the types of all key expressions
         and `V` is the greatest lower bound of the context type schemas of all
@@ -2906,6 +2901,10 @@ Here is one way it could be broken down into separate pieces:
 ### 2.12
 
 -   Add `...` rest patterns in list and map patterns (#2453).
+
+-   Change context type schema to consistently use `?` in patterns where the
+    type isn't known instead of `?` for unannotated variable patterns and
+    `Object?` for other patterns.
 
 ### 2.11
 
--- a/working/0546-patterns/patterns-feature-specification.md
+++ b/working/0546-patterns/patterns-feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: In progress
 
-Version 2.10 (see [CHANGELOG](#CHANGELOG) at end)
+Version 2.11 (see [CHANGELOG](#CHANGELOG) at end)
 
 Note: This proposal is broken into a couple of separate documents. See also
 [records][] and [exhaustiveness][].
@@ -212,7 +212,6 @@ Here is the overall grammar for the different kinds of patterns:
 
 ```
 pattern               ::= logicalOrPattern
-patterns              ::= pattern ( ',' pattern )* ','?
 
 logicalOrPattern      ::= ( logicalOrPattern '|' )? logicalAndPattern
 logicalAndPattern     ::= ( logicalAndPattern '&' )? relationalPattern
@@ -541,24 +540,40 @@ is expected.
 ### List pattern
 
 ```
-listPattern ::= typeArguments? '[' patterns? ']'
+listPattern         ::= typeArguments? '[' listPatternElements? ']'
+listPatternElements ::= listPatternElement ( ',' listPatternElement )* ','?
+listPatternElement  ::= pattern | restPattern
+restPattern         ::= '...' pattern?
 ```
 
 A list pattern matches an object that implements `List` and extracts elements by
 position from it.
 
-It is a compile-time error if `typeArguments` is present and has more than one
-type argument.
+It is a compile-time error if:
 
-**TODO: Allow a `...` element in order to match suffixes or ignore extra
-elements. Allow capturing the rest in a variable.**
+*   `typeArguments` is present and has more than one type argument.
+
+*   There is more than one `restPattern` element in the list pattern. *It can
+    appear anywhere in the list, but there can only be zero or one.*
+
+#### Rest patterns
+
+A list pattern may contain a *rest element* which allows matching lists of
+arbitrary lengths. If a rest element is present and has a subpattern, all of the
+elements not matched by other subpatterns are collected into a new list and that
+list is matched against the rest subpattern.
+
+```dart
+var [a, b, ...rest, c, d] = [1, 2, 3, 4, 5, 6, 7];
+print('$a $b $rest $c $d'); // Prints "1 2 [3, 4, 5] 6 7".
+```
 
 ### Map pattern
 
 ```
 mapPattern        ::= typeArguments? '{' mapPatternEntries? '}'
 mapPatternEntries ::= mapPatternEntry ( ',' mapPatternEntry )* ','?
-mapPatternEntry   ::= expression ':' pattern
+mapPatternEntry   ::= expression ':' pattern | '...'
 ```
 
 A map pattern matches values that implement `Map` and accesses values by key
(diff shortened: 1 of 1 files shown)


working/0546-patterns/patterns-feature-specification.md
```diff
@@ -541,24 +540,40 @@ is expected.
...
+*   There is more than one `restPattern` element in the list pattern. *It can
+    appear anywhere in the list, but there can only be zero or one.*
 
-**TODO: Allow a `...` element in order to match suffixes or ignore extra
-elements. Allow capturing the rest in a variable.**
+#### Rest patterns
+
+A list pattern may contain a *rest element* which allows matching lists of
+arbitrary lengths. If a rest element is present and has a subpattern, all of the
+elements not matched by other subpatterns are collected into a new list and that
```
leafpetersen, line 563: What is the runtime type of the new list?

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -541,24 +540,40 @@ is expected.
...
+list is matched against the rest subpattern.
```
leafpetersen, line 564: Do we really want to specify that a new list is allocated?  Should we consider a using "slice" of the underlying list instead?

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -541,24 +540,40 @@ is expected.
...
+
+```dart
+var [a, b, ...rest, c, d] = [1, 2, 3, 4, 5, 6, 7];
```
leafpetersen, line 567: Should the pattern have a required type of `List<S>` for some kind of `S`?  That is, I think in a refutable context we would otherwise allow `var [... String rest]` which can't ever match, right?  Or is that an error for other reasons?

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1501,17 +1538,44 @@ The context type schema for a pattern `p` is:
...
+            1.  If `e` is a rest element pattern with a subpattern `s` and the
+                context type schema of `s` is a `List<T>` for some type `T`,
+                then add `T` to `es`.
+
+            2.  Else if `e` is not a rest element pattern, add the context
+                type schema of `e` to `es`.
 
-        *Here, the GLB of `int` and `num` is `int`, which ensures that neither
-        `int a` nor `num b` need to downcast their respective fields.*
+            *Else, `e` is a rest element with no subpattern, so it doesn't
```
leafpetersen, line 1554: No, `e` can be a rest element with a subpattern with some other type or no type, right?  I think this should probably not be an error: `var [...rest] = <int>[]`, but `rest` has no context type schema here.  I'm not sure whether `case [...String rest]` is intended to be an error or not - I asked about that above.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2540,12 +2639,54 @@ To bind invocation keys in a pattern `p` using parent invocation `i`:
 
     1.  Bind `i : ("length", [])` to the `length` getter invocation.
 
-    2.  For each element subpattern:
+    2.  For each element subpattern `s`:
+
+        1.  If `s` is a rest element:
```
leafpetersen, line 2644: Don't we need to distinguish between rest elements with a pattern (which call sublist) and those that don't?  That is, `[..., a]` doesn't cache anything for the sublist, but `[...rest, a]` does, right?

working/0546-patterns/patterns-feature-specification.md
leafpetersen: I think you answer this later implicitly, via the specification of this as a call to `sublist`, right?

leafpetersen (approved):

leafpetersen:
Is this something we need for the initial implementation?  If not, we may want to coordinate with the front end teams to see how they feel about increasing the scope now.

eernstg:
We could consider putting the 'rest' patterns into `patterns-later` entirely. But we could also consider adding a plain `...` (without a subpattern) that allows the enclosing pattern to ignore the rest of a list/record only (so `...` must be the rightmost element of the list/record).

The full generality seems to give rise to many new questions. For instance, matching `[1, 2, 3, 4]` with `[a, ..., b, ..., c]`, do we bind `b` to 2 or 3?

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -579,6 +594,28 @@ It is a compile-time error if:
     don't have primitive equality, it is possible to have redundant keys and the
     compiler won't detect it.*
 
+*   There is more than one `...` element in the map pattern. *It can
+    appear anywhere in the map, but there can only be zero or one.*
```
stereotype441, line 598: The italicized text here contradicts the bullet below.  Is the `...` allowed to appear anywhere in the map, or not?

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1501,17 +1538,44 @@ The context type schema for a pattern `p` is:
...
+
+            2.  Else if `e` is not a rest element pattern, add the context
+                type schema of `e` to `es`.
 
-        *Here, the GLB of `int` and `num` is `int`, which ensures that neither
-        `int a` nor `num b` need to downcast their respective fields.*
+            *Else, `e` is a rest element with no subpattern, so it doesn't
+            contribute to inference.*
+
+        3.  If `es` is empty, then `E` is `Object?`. *This can happen if the
```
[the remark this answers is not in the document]
stereotype441: I don't have an opinion about whether `Object?` or the "unknown type" is a better choice, but whatever we do here, we should do the same for bullet point 2 on line 1537.

Or maybe it would be better to just drop that bullet point entirely.  (We previously needed it because the GLB of zero types was undefined.  But now that we're explicitly defining what to do if no types get accumulated into `es`, there's no need to separately handle the case of an empty list).

working/0546-patterns/patterns-feature-specification.md
stereotype441: Personally I'm ok with `sublist()`.  It's simple and well-defined, and reasonably performant for the cases that I believe are going to arise in practice.

I might have been ok with a slice if slices were a thing that were commonly used in Dart, but they aren't.  And because of that, I think people are going to assume sublist-like semantics with regard to mutation (e.g. `var [x, ...rest] = y; rest.add(0);` doesn't modify the original list).

I don't want to go to extra implementation effort to optimize a list pattern appearing to the right of `...` because I don't think anyone is going to write code like that.  Our time is better spent on other language features.  If someone really does write `[var [var a, ...[var b, ...var r, var c], var d] = [1,2,3,4,5];` and they get sub-optimal performance, well, they asked for it.

stereotype441 (approved):
lgtm with a few nits

working/0546-patterns/patterns-feature-specification.md
munificent: Yeah, the runtime type falls out of that.

working/0546-patterns/patterns-feature-specification.md
munificent: I agree with everything Paul said. It felt a little arbitrary when I first thought to use `sublist()`, but the longer I thought about it, the more I like it. It *is* part of the `List` interface, so it's right there for the taking. It makes it easier to explain to users what kind of object they get if the rest pattern captures it and what they can do with it. We just say: it's whatever `sublist()` on the matched list returns.

I can't imagine any reasonable person doing nested list patterns inside a rest pattern, which in turn contains a rest pattern. If that's marginally slower than it could be, I won't lose sleep.

I do strongly like that using `sublist()` makes the specification of this and the semantics fairly short and simple. It's essentially just another subpattern that calls `sublist()` instead of `[]`.

working/0546-patterns/patterns-feature-specification.md
munificent: > I think in a refutable context we would otherwise allow `var [... String rest]` which can't ever match, right?

That's true, but only reliably true because we happen to know that String is sealed. I think this is similar to the general property where we don't report an error when a subpattern requires a type unrelated to the matched value type because there could be some unknown third type that is a subtype of both. It might be weird to have:

```dart
var [...Future foo] = ...
```

But some particularly irascible Dart programmer *could* define:

```dart
class Wat<T> implements List<T>, Future<T> { ... }
```

So even though `[...String rest]` is weird and useless, I don't know if it needs to be an error.

working/0546-patterns/patterns-feature-specification.md
munificent: Done (except for the `FutureOr` bit which seems a little gratuitous).

working/0546-patterns/patterns-feature-specification.md
munificent: Yeah, we don't have to worry about caching in this section. The point of this part of the proposal is to statically define a key to identify each cached object—or perhaps more precisely, to define which patterns *share* cached values— and we can harmlessly define keys for `sublist()` even for rest patterns that don't actually end up calling `sublist()`. It just means that *if* the rest pattern were to call `sublist()`, we know which other `sublist()` calls elsewhere in the set of cases would share the same cached value.

munificent:
> Is this something we need for the initial implementation?

I think we at least need `...` for map patterns because otherwise there's no way to match a map that has some keys you care about while ignoring others and that seems like a useful thing to support. `...` in list patterns is also useful. We could conceivably support just trailing `...` without any subpatterns if that helps reduce the scope, but I hope that the full behavior isn't too much additional work beyond that. I'll reach out to the implementation teams and let them know that the whole enchilada is negotiable if they're feeling up against the wall.

working/0546-patterns/patterns-feature-specification.md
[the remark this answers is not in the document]
munificent: Good point. Changed to `?` here and elsewhere (constant patterns, unknown maps, etc.).

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2304,19 +2372,47 @@ To match a pattern `p` against a value `v`:
...
+    4.  Match the head elements. For `i` from `0` to `h - 1`, inclusive:
+
+        1.  Extract the element value `e` by calling `[]` on `v` with index `i`.
+
+        2.  Match the `i`th element subpattern against `e`.
+
+    5.  If there is a rest element and it has a subpattern:
+
+        1.  Let `r` be the result of calling `sublist()` on `v` with arguments
+            `h`, and `l - t`.
```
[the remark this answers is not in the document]
munificent: We could do that, but it seems like a lot of new machinery to spec and implement for marginal benefit. I think just calling `sublist()` has a lot of value:

1. @leafpetersen pointed out that it preserves the runtime type of the original list, which could be useful.
2. It lets a custom `List` implementation control the returned type. For example, a `LoggingList` that logs every element access could return a sublist that also logs element access.
3. It's easy to explain to users because the semantics are defined in terms of an API they already know.

working/0546-patterns/patterns-feature-specification.md
munificent: Oops, copy-pasted from list and didn't fix. :) Fixed.

munificent:
I'm going to go ahead and merge this in but I'm definitely up for discussing the choice to call `sublist()` and we can iterate on this more if we want.

working/0546-patterns/patterns-feature-specification.md
leafpetersen: I'm pretty confused now.  What is your intended semantics of this?

```
var [...String x] = <int>[3];
```

I had expected that you would say that the required type for `String x` is `List<int>`, and that this would be an error.  But you seem to be saying that no, this is fine, there's no required type for the pattern.  In which case.... this pattern is refutable?  Or is there some other static error that applies here that I'm missing?

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1528,41 +1564,65 @@ The context type schema for a pattern `p` is:
...
-        ```dart
-        var [int a, num b] = [1, 2];
-        ```
+        1.  Let `es` be an empty list of types.
+
+        2.  For each subpattern `e` in `p`:
+
+            1.  If `e` is a rest element pattern with a subpattern `s` and the
+                context type schema of `s` is an `Iterable<T>` for some type
+                `T`, then add `T` to `es`.
```
lrhn, line 1585: Do we want to mention that `T` can be a type schema, since the schema of `s` can be `List<?>`.
In that case `T` is not a type, and isn't added to the list of types, and that's fine.
Might just want to be explicit about it.

working/0546-patterns/patterns-feature-specification.md
lrhn: The required type of `String x` is `String`. The matched value type is `List<int>`. Since `List<int>` is not assignable to `String`, that should be a compile-time error.
The required type of a pattern is the type that it can actually match. It only depends on the matched value type when the pattern is underspecified, and we infer its type through type inference from the matched value type.

In a refutable context, the `[...String x]` will just never match, but neither will `[Never x]`, and `someList is String` will always be false. We don't generally prevent you from doing impossible checks (but the analyzer is free to warn you if you do).

munificent: Lasse is correct. Sorry, I got my example wrong. In your previous comment, you mentioned "in a refutable context" (which is correct), but then my example was an irrefutable use of a pattern. I should have written:

```dart
case [...Future foo]: ...
```

This doesn't have a compile-time error. It's theoretically possible that the `List` returned by the `sublist()` call on the destructured list happens to return some class that both implements `List` and `Future`, so we don't prohibit it outright. But this *is* a compile-time error:

```dart
var [...Future foo] = ...
```

Because we don't know that it *will* match, in an irrefutable context, we prohibit this at compile-time since pattern variable declarations should never fail to match.

There is an interesting wrinkle here. Someone could in theory define:

```dart
class FunnyList implements List<int> {
  ListAndFuture sublist() { ... }
  ...
}

class ListAndFuture implements List<int>, Future { ... }
```

Then if you were to do:

```dart
var [...Future f] = FunnyList();
```

We *could* allow this. The question is whether the matched value type of a rest subpattern is just `List<E>` where `E` is the surrounding list pattern's required type's element type, or we could say that the matched value type is the return type of the `sublist()` method on the matched value of the surrounding list. The proposal says the former, but it would potentially be more precise to pick the latter because then `sublist()` methods with covariant returns could return a more precise type.

In practice... I don't think anyone will do that, so I'm OK with the proposal where it is.

lrhn: If we define the matched value type of a list spread pattern to be the same list type as the outer list pattern, and don't try to be clever and take the return type of `sublist`, then the irrefutable example will be invalid.

It won't stop the same example in a refutable context from matching, though. I guess that's OK. 
I expect a lint warning about pattern match between unrelated types, and the analyzer warnings about non-list types in spread patterns.

working/0546-patterns/patterns-feature-specification.md
munificent: Ah, good catch. Done: https://github.com/dart-lang/language/commit/a93b4b14776eb0e802065c8eb39dd9c11d514fa5

Issue #2451, opened 2022-09-01
`base` class modifier loophole


stereotype441:
Regarding the `base` class modifier defined [here](https://github.com/dart-lang/language/blob/master/working/type-modifiers/feature-specification.md#prohibiting-implementing), it seems to me that the purpose of marking class `A` as a "base" class is to ensure that any instance of a class that subclasses `A` in the program is guaranteed to have `A` in its superclass chain.  (This, for example, makes it safe to put private members in `A`, and access them via a target other than `this`).

But as I understand the current spec proposal, there's a loophole.  If my library contains:
```dart
abstract base class A {
  _foo() {}
}
```

Then what's to stop some other library from doing this?
```dart
abstract class B extends A {}
class C implements B {}
```

To close this loophole I think we should say that if a class `A` in library `L` has the `base` class modifier, then any subclass of `A` outside of `L` must have `A` somewhere in its supertype chain.

stereotype441:
Note: it's tempting to fix this proposal by instead saying that if a class `C` extends a class `A`, and `A` has the `base` modifier, then `C` also has the `base` modifier (whether it's specified explicitly or not).  But that doesn't close the loophole, because restrictions are ignored inside a library, so the problematic example above would be accepted.

stereotype441:
CC @lrhn @munificent @leafpetersen @kallentu @eernstg @mit-mit @jakemac53 @natebosch

lrhn:
Prohibition against implementation would need to be inherited, otherwise it doesn't work.

Which means it's not a property of a class, it's a property of an inheritance tree, so the "base class" name doesn't really match.

That can be implemented either by implicitly inheriting the `base`-ness, or by requiring any subclass to declare itself as `base` as well.

stereotype441:
> That can be implemented either by implicitly inheriting the `base`-ness, or by requiring any subclass to declare itself as `base` as well.

That's not sufficient, because we allow you to violate the prohibition within the same library.  Which means that if we did it by requiring the subclass to declare itself as `base`, or implicitly inheriting the `base`-ness, this would be allowed:

```dart
// lib1.dart
abstract base class A {
  _foo() {}
}
```

```dart
// lib2.dart
import 'lib1.dart';

abstract base class B extends A {} // allowed because also "base"
class C implements B {} // allowed because same library as B
```

lrhn:
[...]

PR #2592, opened 2022-10-28
Standalone proposal for "closed" and "base" type modifiers.

munificent:
This is a subset of the larger "type modifiers" proposal just covering the "closed" and "base" modifiers. I carved it out from type modifiers because the latter also proposes disallowing using classes as mixins, which is a much bigger change. It's separate from sealed types because "closed" and "base" aren't necessary for exhaustiveness.

### DIFF
--- a/working/closed-and-base/feature-specification.md
+++ b/working/closed-and-base/feature-specification.md
@@ -48,7 +48,7 @@ break much code.
 This proposal will likely build on top of the [sealed types][] proposal, in
 which case the full grammar is:
 
-[sealed types]: https://github.com/dart-lang/language/blob/master/working/type-modifiers/feature-specification.md
+[sealed types]: https://github.com/dart-lang/language/blob/master/working/sealed-types/feature-specification.md
 
 ```
 classDeclaration ::=
--- a/working/closed-and-base/feature-specification.md
+++ b/working/closed-and-base/feature-specification.md
@@ -38,27 +38,57 @@ mixinDeclaration ::= 'base'? 'mixin' identifier typeParameters?
   '{' (metadata classMemberDeclaration)* '}'
 ```
 
-This proposal will likely build on top of the [sealed types][] proposal, in
-which case the full grammar is:
-
-```
-classDeclaration ::=
-  'closed'? ('sealed' | 'abstract')? 'base'? 'class' identifier typeParameters?
-  superclass? interfaces?
-  '{' (metadata classMemberDeclaration)* '}'
-  | 'closed'? ('sealed' | 'abstract')? 'base'? 'class' mixinApplicationClass
-
-mixinDeclaration ::= sealed? 'base'? 'mixin' identifier typeParameters?
-  ('on' typeNotVoidList)? interfaces?
-  '{' (metadata classMemberDeclaration)* '}'
-```
-
-[sealed types]: https://github.com/dart-lang/language/blob/master/working/type-modifiers/feature-specification.md
-
 **Breaking change:** Treating `closed` and `base` as built-in identifiers means
 that existing code that uses those the names of type will no longer compile.
 Since almost all types have capitalized names in Dart, this is unlikely to be
 break much code.
+
+### With sealed types
+
+This proposal will likely build on top of the [sealed types][] proposal, in
+which case the full grammar is:
+
+[sealed types]: https://github.com/dart-lang/language/blob/master/working/type-modifiers/feature-specification.md
+
+```
+classDeclaration ::=
+  classModifiers 'class' identifier typeParameters?
+  superclass? interfaces?
+  '{' (metadata classMemberDeclaration)* '}'
+  | classModifiers 'class' mixinApplicationClass
+
+classModifiers ::= 'sealed' | 'closed'? 'abstract'? 'base'?
+
+mixinDeclaration ::= ('sealed' | 'base')? 'mixin' identifier typeParameters?
+  ('on' typeNotVoidList)? interfaces?
+  '{' (metadata classMemberDeclaration)* '}'
+```
+
+Note that the grammar disallows combining `sealed` with `closed` or `base` since
+a sealed type is already prohibited from being extended or implemented outside
+of the current library. We *do* allow a class to be marked `closed abstract
+base` instead of treating `sealed` as a synonym for that because there are
+subtle differences if the class has subtypes in the same library:
+
+```dart
+// lib.dart
+sealed class A {}
+class B extends A {}
+class C extends A {}
+
+closed abstract base class D {}
+class E extends D {}
+class F extends D {}
+```
+
+Here, pattern matching on `B` and `C` exhaustively covers `A`, but matching on
+`E` and `F` does not cover `D`. Marking `D` as `closed abstract base` means that
+users are prevented from extending, implementing, or constructing it, but the
+maintainer of `lib.dart` can freely add new subtypes of `D` without breaking
+users by causing what were exhaustive pattern matches to no longer be
+exhaustive. In other words, it gives the supertype author the ability to opt
+out of exhaustiveness checks while still defining a type that is otherwise as
+restricted as `sealed`.
 
 ### Static semantics
 
--- a/working/closed-and-base/feature-specification.md
+++ b/working/closed-and-base/feature-specification.md
@@ -17,6 +17,124 @@ larger breaking change. This proposal is non-breaking.
 
 [type modifiers]: https://github.com/dart-lang/language/blob/master/working/type-modifiers/feature-specification.md
 
+## Motivation
+
+Why might a class or mixin author want to *remove* capabilities? Doesn't that
+just make the type less useful? The type does end up more restricted, but in
+return, there are more invariants about the type that you can rely on being
+true. Those invariants may make the type easier to understand, maintain, evolve,
+or even just to use.
+
+### Preventing subclassing
+
+Most types contain methods that invoke other methods on `this`, for example:
+
+```dart
+class Account {
+  int _balance = 0;
+
+  bool canWithdraw(int amount) => amount <= _balance;
+
+  bool tryWithdraw(int amount) {
+    if (amount <= 0) throw ArgumentError("Amount must be positive.");
+
+    if (!canWithdraw(amount)) return false;
+    _balance -= amount;
+    return true;
+  }
+
+  // ...
+}
+```
+
+The intent is that `_balance` should never be negative. There may be other code
+in this class that breaks if that isn't true. However, there's nothing
+preventing a subclass from doing:
+
+```dart
+class BustedAccount extends Account {
+  // YOLO.
+  bool canWithdraw(int amount) => true;
+}
+```
+
+Extending a class gives you free rein to override whatever methods you want,
+while also inheriting concrete implementations of other methods that may
+assume you haven't.
+
+If we can prevent Account from being subclassed, we ensure that when
+`tryWithdraw()` calls `canWithdraw()`, it calls the actual `canWithdraw()`
+method we expect.
+
+Note that it's *not* necessary to prevent *implementing* `Account` for this use
+case. If you implement `Account`, you inherit *none* of its concrete
+implementation, so you don't end up with methods like `tryWithdraw()` whose
+behavior is broken.
+
+### Preventing implementing
+
+Consider:
+
+```dart
+class Account {
+  int _balance = 0;
+
+  bool tryTransfer(int amount, Account destination) {
+    if (amount > _balance) return false;
+
+    _balance -= amount;
+    destination._balance += amount;
+  }
+
+  // ...
+}
+```
+
+This code may fail with a `NoSuchMethodException` at runtime. Spot the bug?
+There's nothing preventing someone from passing in their own implementation of
+`Account` defined in another library that doesn't have a `_balance` field. In
+general, it's not safe to assume any object coming in to your library actually
+has the private members you expect, because it could be an outside
+implementation of the class's interface.
+
+Here's another example:
+
+```dart
+/// Assigns a unique ID to each instance.
+class Handle {
+  static int _nextID = 0;
+
+  final int id;
+
+  Handle() : id = _nextID++;
+}
+
+class Cache {
+  final Map<int, Handle> _handles = {};
+
+  void add(Handle handle) {
+    _handles[handle.id] = handle;
+  }
+
+  Handle? find(int id) => _handles[id];
+}
+```
+
+The `Cache` class assumes each `Handle` has a unique `id` field. The `Handle`
+class's constructor ensures that (ignoring integer overflow for the moment).
+But if an unrelated type could implement `Handle`'s interface, then there's
+no guarantee that every instance of `Handle` has actually gone through that
+constructor.
+
+If the constructor is doing validation or caching, you might want to assume that
+all instances of the type must run it. But if the class's interface can be
+implemented, then it's possible to route around the constructor and break the
+class's invariants.
+
+By preventing a class from having its interface implemented, you ensure that
+every instance of the class you will ever see has all of the private members you
+defined on it and has gone through the constructors you defined.
+
 ## Syntax
 
 A class declaration may be preceded with the built-in identifiers
--- a/working/closed-and-base/feature-specification.md
+++ b/working/closed-and-base/feature-specification.md
@@ -1,0 +1,84 @@
+# "Closed" and "base" types
+
+Author: Bob Nystrom
+
+Status: In-progress
+
+Version 1.0
+
+This proposal specifies `closed` and `base` modifiers on classes and mixins,
+which allow an author to prohibit the type being extended or implemented,
+respectively.
+
+This proposal is a subset of the [type modifiers][] strawman, which also
+contains most of the motivation. It is split out here because the type modifiers
+strawman also proposes prohibiting classes being used as mixins, which is a
+larger breaking change. This proposal is non-breaking.
+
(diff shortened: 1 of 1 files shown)


munificent:
In here, I took a very simple approach to trying to address #2451: I basically say that `base` is poisonous and affects all downstream classes, even ones in the same library. Seems pretty simple, but I'm not sure at all if that's the best approach. I just wanted to get something down so we can iterate on it.

working/closed-and-base/feature-specification.md
```diff
@@ -0,0 +1,84 @@
...
+  ('on' typeNotVoidList)? interfaces?
+  '{' (metadata classMemberDeclaration)* '}'
+```
+
+This proposal will likely build on top of the [sealed types][] proposal, in
+which case the full grammar is:
+
+```
+classDeclaration ::=
+  'closed'? ('sealed' | 'abstract')? 'base'? 'class' identifier typeParameters?
```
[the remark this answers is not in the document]
munificent: Ah, good point! If we *don't* have `closed` and `base`, then it's useful to allow `sealed` even on types with no subtypes because it gives you a backdoor way to express `closed base`. But if *do* have those, as this proposes, then it's redundant to put `closed` and `base` on a class also marked `sealed`.

It doesn't conflict because even when you mark a class `closed` and/or `base` you can still extend/implement it *in the same library. But it is redundant, and we should probably prohibit it. Updated.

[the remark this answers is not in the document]
leafpetersen: I don't think it's redundant, because of the transitiveness of base, right?

working/closed-and-base/feature-specification.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Static semantics
+
+It is a compile-time error to:
+
+*   Extend a class marked `closed` outside of the library where it is defined.
+
+*   Implement a type marked `base` outside of the library where it is defined.
+
+Any type that extends or mixes in a type marked `base` is implicitly treated as
```
leafpetersen, line 101: Why is base applied transitively, but closed not?  That feels weird to me.

leafpetersen:
I'm fine with landing this as a proposal, but I'm not really sold on it yet.  :)

leafpetersen:
Added a concrete counter-proposal [here](https://github.com/dart-lang/language/issues/2595)

working/closed-and-base/feature-specification.md
munificent: I should probably fill in a little more motivation and context in this doc. :)

It's worth thinking about these capabilities in terms of the invariants they define and the tools they offer to class maintainers. In other words, *why* a user might want to *remove* a capability from a type and what positive value they provides in return.

### Base

By preventing a class C from being implemented, you get the invariant that if you have an instance of C, you know you have an object that concretely inherits from C. That means it is safe to cast it to private interfaces that C might implement, call private methods on C, access private fields, etc. You know that any validation or initializing in the constructor of C has been executed on this instance. It's a real C and (except for the things that can be broken by overriding), you know it has all of the state, validation, and behavior that C itself defines.

That invariant is only maintained if `base` is transitive. If a subclass D of C can in turn expose its own interface and that interface has C as a superinterface, then you could have an unrelated class that implements D but does not extend C. Now you have an object that is an instance of C but doesn't inherit from it.

Another possible alternative is to say that `base` isn't transitive. D can still extend C, and it can an interface, but that interface does *not* have C as a superinterface. So any type that *implements* D is not a subtype of C, but a type that *extends* D is. That feels pretty shaky to me, which is why I proposed instead we make `base` transitive.

Another option is to not make `base` transitive but make it *required* in subclasses. So you don't get it for free, but it's a compile error if the subclasses don't also write it.

### Closed

I think the main useful invariant that disabling subclassing gives you is freedom from overriding. If you have a class C and some of its methods call other methods on `this`, then you know it will invoke your own methods right on C. You don't have to worry about "does this class get into a weird broken state if this method is overridden?" (This *is* something Dart class designers should worry about but people mostly just don't and rely on a friendly agreement not to override methods that aren't explicitly allowed to be.)

That invariant doesn't require `closed` to be transitive. If I have a closed class C and some class D that implements C, I don't care if people extend D or not. Either way, they aren't using *any* of C's implementation, so the invariant I care about is preserved. D has already severed itself from C's implementation, so whether or not D is extended is D's problem and D's problem alone.

leafpetersen: > That invariant doesn't require `closed` to be transitive. If I have a closed class C and some class D that implements C, I don't care if people extend D or not.

I don't understand your example - you seem to be talking about implements.  The example I have in mind is that `C` is closed, `D` extends `C` (it's my library, I know what I'm doing) but I don't want things out side of my library extending `D`.  Sure, I can mark all of the subclasses of `C` as closed as well, but it feels to me that that's probably not the right default.

lrhn:
My concrete problem with the names is that `base` and `interface` are positive names, used as a restriction.

The default is that you can extend and implement.
That's the same as writing `base interface` (you can do both).
Writing only one of those is effectively the same as not allowing the *other* one, `base class Foo` means not implementable, `interface class Bar` means not extensible, but `base interface class Foo` allows both, and we need `closed` to prevent both.
It's just feels backwards to me.

It is sort-of consistent in the sense that if there is *any* modifier, then we remove all the capabilities and add back the ones we wrote, with `closed` meaning adding none (which is different from writing nothing, because that just inherits capabilities instead).
| capability | modifiers |
|-|-|
| -extends -implements | `closed` |
| -extends +implements | `interface` |
| +extends -implements | `base` |
| +extends +implements | `base interface` |
| - inherit - | |

It's inconsistent in that we say that the default is permissive, but then I'd expect that you should be removing capabilities, not adding them back.

munificent:
Heads up that I added a motivation section just to try to make sure we're on the same page about why we might want to do this and what kinds of behavior that might suggest for the feature.

munificent:
> My concrete problem with the names is that `base` and `interface` are positive names, used as a restriction.

Are you responding to [Leaf's proposal](https://github.com/dart-lang/language/issues/2595) or this one? If the former, it might be good to move this comment over to that issue.

working/closed-and-base/feature-specification.md
eernstg: Can we maintain a 'private correctness' property for a `base` class? If we can say that it is safe for a leaking `base` class to have private properties (but a leaking non-`base` class with private members may cause noSuchMethod failures that subtypes can't avoid, so we can lint against leaking non-`base` classes). Surely that's only possible if `base` is enforced for all subtypes outside the declaring library.

Another property that a transitive `base class A` enforces is that the subtype hierarchy is a tree (unless this property is violated in the library of `A`). When the subtype hierarchy is a tree, reasoning about types may be stronger. (E.g., it is known that `o is B` and `o is C` cannot both be true if `B` and `C` are unrelated subtypes of `A`.) Similarly, when the subtype hierarchy is a tree and every subtype relation is the subclass relation, it may be possible to use faster implementations of some operations. (I believe some very performant approaches to subtype tests and object layout are only applicable in this situation.)

I'd prefer to require `base` explicitly on every subtype of a `base`  class, because it is useful for a reader of the code of a `class B extends A` to know that it's a `base` class, even in the case where that's the only available choice because the superclass `A` is a `base` class.

I agree with @munificent that `closed` does not need to be transitive: When we enforce that subtypes can't inherit anything, it's trivially maintained for further subtypes that they also can't inherit anything.

The obvious metaphor could be the following: To break a chain, it's enough to break one link. To have a working chain, we have to transitively ensure that none of the links are broken. (So with `closed` we're breaking the chain at the first link; with `base` we're transitively ensuring that every subtype link is a subclass link, so we know that all of them _will_ inherit these private members).

eernstg (commented):
Looks quite good! I suggested a couple of changes, in particular that the ability to derive a mixin should be omitted from a `closed class`, and perhaps even from a `base class`.

working/closed-and-base/feature-specification.md
leafpetersen: > I agree with @munificent that closed does not need to be transitive: When we enforce that subtypes can't inherit anything, it's trivially maintained for further subtypes that they also can't inherit anything.

This is not true, and I give an example of why not in the comment immediately above.

eernstg: I believe the example is
```dart
// Same library.
closed class C {}
class D extends C {}

// Some other library.
class E extends D {} // Allowed if `closed` isn't implied on `D`.
```
I was never very happy about the rule "everything is allowed as long as it is in the same library", and the property 'subtypes can't inherit anything' is of course not true when we use that loophole.

But if we want to make `closed` transitive then it seems to be motivated by the need to close this loophole (so we might want `closed` to be implied on `D`), but it is not needed for any classes in other libraries. Presumably, a class `E` (in some other library) that implements `C` or `D` could allow subtypes of `E` to use `extends` (and `E` would be designed with that usage in mind), so it could even be really inconvenient if `closed` is implied on `E`. Those subtypes would inherit from `E` (not from `C` or `D`), so `C` and `D` wouldn't suffer any harm.

So maybe `closed` should be "transitive in the current library only"?

leafpetersen: > So maybe closed should be "transitive in the current library only"?

It doesn't matter outside of the current library.  Outside of the current library, you can't extend it, so you can't "inherit it".  It's only inside the current library that transitivity or non-transitivity is observable.  And it seems to me to be surprising to users if `base` is transitive (inside the currently library) but `closed` is not.  Users have to remember that putting `base` on a subclass inside the library is redundant, but putting `closed` on a subclass inside of a library is not redundant.  That seems like a foot gun to me.

eernstg: > It doesn't matter outside of the current library.

I would certainly expect global transitivity vs. transitivity in the current library (let's call it 'library transitivity') to make a difference:

```dart
// Library L1.
closed class C {}
class D extends C {} // Means `closed class D extends C {}`.

// Library L2.
import L1;
class E1 extends D {} // Error because `D` is closed.
class E2 implements D {} // OK.

// Library L3.
import L2;
class F2 extends E2 {} // Error?
```

With global transitivity, `E2` is implicitly closed because it has `C` in its superinterface graph and `C` is closed (libraries don't matter here). With library transitivity, `E2` is not `closed`. So `F2` is an error in the former case and OK in the latter case.

The fact that `C` is a closed class provides some guarantees (I haven't quite found any essential ones, yet, but those guarantees must be the motivation for having the `closed` modifier in the first place). I can't see how any reasonable guarantees would be maintained or violated by restricting whether or not `F2` is allowed to inherit member implementations from `E2`. Basically, I'm thinking that it's none of `C`'s business whether `F2` inherits from `E2` or not.

That's the reason why I do not think it's obviously desirable to make `closed` globally transitive. But it may very well be a good idea to make it library transitive.

leafpetersen: > With global transitivity, E2 is implicitly closed because it has C in its superinterface graph and C is closed (libraries don't matter here). 

Correct me if I'm wrong @munificent but I don't believe that there is any proposal to make access capabilities be carried via `implements` (super-interface) relations, only `extends/with` (superclass ) relations.  So both `E2` and `F2` are fine, in any of these proposals, as best I know.

eernstg:
```dart
// Library L1.
some modifiers class C {}
class D implements C {}
```

So you're saying that class `D` could be used to allow every possible capability for out-of-library subtypes of `A`? Can we then have useful guarantees at all?

munificent: > Can we maintain a 'private correctness' property for a base class? 

Yes. The motivation section I added later talks about this. It's one of the motivations for requiring that if a supertype can't be implemented, then all subtypes can't be implemented either. See #2451.

The other correctness property you get by preventing implements is that it means every instance of the class must have invoked one of the constructors defined on that class.

> I'd prefer to require base explicitly on every subtype of a base class, because it is useful for a reader of the code of a class B extends A to know that it's a base class, even in the case where that's the only available choice because the superclass A is a base class.

Good call. Done.

> Correct me if I'm wrong @munificent but I don't believe that there is any proposal to make access capabilities be carried via implements (super-interface) relations, only extends/with (superclass ) relations. 

The *only* thing I proposed being propagated in any way or form was `base`. And I'm changing that to say that it's not carried either, and instead the subtypes are required to explicitly state it.

```dart
// Library L1.
closed class C {}
class D extends C {} // Means `closed class D extends C {}`.
```

According to this proposal, `class D extends C {}` just means `class D extends C {}`. It inherits no capability restrictions from `C`. Outside of L1, a user can't extend `C` but they could extend `D`. If the author of L1 doesn't want that, they can write:

```dart
// Library L1.
closed class C {}
closed class D extends C {} // Means `closed class D extends C {}`.
```

I am definitely worried about the footgun that users might assume the restriction *is* propagated down and will be surprised that it's not. But aside from that, I like the semantics of not propagating down any capabilities or restrictions. It means:

1. You can see the capabilities of a class by looking only at its own class declaration.
2. We only need a set of modifiers to toggle the defaults in one direction. We don't also need `unclosed` and `unbased` for when a subclass inherits a restriction but the library author wants to remove the restriction in a subclass.
3. It's simple. The specification in this document is, like, three sentences. (And one of them is to close the implements loophole in #2451.)

working/closed-and-base/feature-specification.md
```diff
@@ -0,0 +1,232 @@
...
+A class declaration may be preceded with the built-in identifiers
+`closed` and/or `base`:
+
+```
+classDeclaration ::=
+  'closed'? 'abstract'? 'base'? 'class' identifier typeParameters?
+  superclass? interfaces?
+  '{' (metadata classMemberDeclaration)* '}'
+  | 'closed'? 'abstract'? 'base'? 'class' mixinApplicationClass
+```
```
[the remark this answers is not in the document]
munificent: Good call! Done.

working/closed-and-base/feature-specification.md
```diff
@@ -0,0 +1,232 @@
...
+mixinDeclaration ::= 'base'? 'mixin' identifier typeParameters?
+  ('on' typeNotVoidList)? interfaces?
+  '{' (metadata classMemberDeclaration)* '}'
+```
+
+**Breaking change:** Treating `closed` and `base` as built-in identifiers means
+that existing code that uses those the names of type will no longer compile.
+Since almost all types have capitalized names in Dart, this is unlikely to be
+break much code.
```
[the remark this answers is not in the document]
munificent: Good point! Done.

working/closed-and-base/feature-specification.md
```diff
@@ -0,0 +1,232 @@
...
+[sealed types]: https://github.com/dart-lang/language/blob/master/working/sealed-types/feature-specification.md
+
+```
+classDeclaration ::=
+  classModifiers 'class' identifier typeParameters?
+  superclass? interfaces?
+  '{' (metadata classMemberDeclaration)* '}'
+  | classModifiers 'class' mixinApplicationClass
+
+classModifiers ::= 'sealed' | 'closed'? 'abstract'? 'base'?
```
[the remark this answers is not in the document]
munificent: Done.

working/closed-and-base/feature-specification.md
```diff
@@ -0,0 +1,232 @@
...
+
+```
+mixinDeclaration ::= 'base'? 'mixin' identifier typeParameters?
+  ('on' typeNotVoidList)? interfaces?
+  '{' (metadata classMemberDeclaration)* '}'
+```
+
+**Breaking change:** Treating `closed` and `base` as built-in identifiers means
+that existing code that uses those the names of type will no longer compile.
+Since almost all types have capitalized names in Dart, this is unlikely to be
```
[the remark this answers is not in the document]
munificent: Deleted since I took your suggestion to not make these built-in identifiers.

working/closed-and-base/feature-specification.md
```diff
@@ -0,0 +1,232 @@
...
+```
+
+Here, pattern matching on `B` and `C` exhaustively covers `A`, but matching on
+`E` and `F` does not cover `D`. Marking `D` as `closed abstract base` means that
+users are prevented from extending, implementing, or constructing it, but the
+maintainer of `lib.dart` can freely add new subtypes of `D` without breaking
+users by causing what were exhaustive pattern matches to no longer be
+exhaustive. In other words, it gives the supertype author the ability to opt
+out of exhaustiveness checks while still defining a type that is otherwise as
+restricted as `sealed`.
```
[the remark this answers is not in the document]
munificent: > I'm not really convinced that this distinction is particularly useful.

I don't think it will be *super* common, but I do think it's useful for an API author to be able to express:

* Here is some class.
* Here is a set of subtypes *I* have defined for it which you can use.
* You may not define your own subtypes of it. I expect to know about them all.
* But *I* reserve the right to add new subtypes myself in future versions of the library without it being a breaking change.

In other words, it's a closed set of subtypes when it comes to user extension, but it doesn't promise to be closed *over time*.

working/closed-and-base/feature-specification.md
```diff
@@ -0,0 +1,232 @@
...
+
+### Static semantics
+
+It is a compile-time error to:
+
+*   Extend a class marked `closed` outside of the library where it is defined.
```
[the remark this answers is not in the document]
munificent: Ah, good idea. Done.

working/closed-and-base/feature-specification.md
```diff
@@ -0,0 +1,232 @@
...
+
+*   Implement a type marked `base` outside of the library where it is defined.
```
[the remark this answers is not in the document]
munificent: Great idea! Done.

munificent:
Closing this in favor of #2646.

PR #2605, opened 2022-11-02
Specify null safety

eernstg:
This is the full PR that integrates the null-safety feature into the language specification. It is not intended to be landed as such: Parts of it are currently being reviewed, and further parts will be reviewed later on, such that all of this gets landed, but avoiding a review which is too large to get done in one fell swoop.

Parts under review (as of Nov 2 2022): https://github.com/dart-lang/language/pull/2052, https://github.com/dart-lang/language/pull/2023.

This PR has been uploaded at this point in order to make it more convenient to refer to specific locations in the proposed new text. When each of the parts lands it is expected that this PR is rebased, and piece by piece of it will disappear because some of the changes in this PR have already been made by those partial PRs.


eernstg:
Woohoo! Part of this PR has now been landed: https://github.com/dart-lang/language/pull/2052.

PR #2606, opened 2022-11-03
[patterns] Refine variable scoping rules for cases that share a body.

### LINKED ISSUE #2553: Scoping of cases that share a body is still weird
I recently [landed a change](https://github.com/dart-lang/language/commit/05e110b3b80d9a9f9b9b1e05165eb86a993cbc45) that tries to clarify how scoping works around patterns and also loosens some restrictions on variables declared in switch case patterns that share a body.

Previously, this was an error:

```dart
switch (obj) {
  case [int i] when i > 0:
  case [String s] when s.isNotEmpty:
    print('matched');
}
```

The old rules were that since those two cases share a body but don't define the exact same set of variables, it's an error. But there's nothing actually problematic about this example because the variables each case declares are only used in their respective guards and not in the body.

The latest update allows this. Basically, it says that when a set of cases share a body, *only the variables used in the body* must be declared by every case and have the same type. That allows the above example, and also this:

```dart
switch (obj) {
  case [int sameName] when sameName > 0:
  case [String sameName] when sameName.isNotEmpty:
    print('matched');
}
```

Both cases declare a variable named `sameName`, but whose type is different in each case. That's fine because the variable isn't used in the body, so the inconsistency doesn't matter.

However, actually specifying this in a coherent way gets pretty complex, especially when you consider that guard expressions can *capture variables in closures*. Consider this gnarly example:

```dart
final callbacks = <Function()>[];

bool capture2(void Function() callback, bool result) {
  callbacks.add(callback);
  return result;
}

switch(['']) {
  case [var s] when capture(() => s += 'guard1', false):
  case [var s] when capture(() => s += 'guard2', true):
    callbacks[0]();
    callbacks[1]();
    print(s);
}
```

Does this print "", "guard2", or "guard1guard2"? The proposal currently says "guard2". This means that *at runtime*, the body runs in a scope where the enclosing scope is *only* the one case that matched. In theory, if "go to definition" was a meaningful thing at runtime, then go to definition on the `s` in `print(s)` would take you to the `var s` in the second case. There is only one `s` and it's only defined by one case.

But that gets very confusing for static analysis, which doesn't know which case will match and must analyze the body in the context of all of the cases. The proposal tries to handle that by analyzing the body in a scope that is the conservative intersection of only the things that are true coming from all cases.

It's *sort of* as if the code sees the above example like:

```dart
String s;
switch(['']) {
  case [s] when capture(() => s += 'guard1', false):
  case [s] when capture(() => s += 'guard2', true):
    callbacks[0]();
    callbacks[1]();
    print(s);
}
```

Statically, there is a single scope where all of the variables shared across the pattern are declared. And then each case just initializes that one variable. But the runtime semantics don't follow that because this would imply that the program prints "guard1guard2".

The core weirdness is that statically, a variable used in the body appears to have *multiple declarations*. But at runtime, it only has one. Having a variable in a body that can statically resolve to multiple possible declarations is a very weird concept for our implementations. The analyzer assumes every variable reference resolves to a single Element, and CFE assumes every variable reference resolves to a single variable declaration AST node.

Implementing flow analysis and type promotion correctly for this is likely possible, but probably a lot of work since it breaks the assumption that every variable is declared at one point.

### Hoisting only shared variables

One option would be to say that when a set of cases share a body, whenever multiple cases declare a variable with the same name, you get *one* variable shared across all of the cases.

In other words, we decide that, yes, the above program prints "guard1guard2". I think this is a *little* surprising because syntactically the two cases do look like they're declaring separate variables. But the distinction is only visible when a guard captures the variable in a closure and that will likely be extremely rare.

The other problem with this model is code like the `sameName` example:

```dart
switch (obj) {
  case [int sameName] when sameName > 0:
  case [String sameName] when sameName.isNotEmpty:
    print('matched');
}
```

The current proposal allows this. If we hoist all variables declared by a set of shared cases up, how do we create a single declaration that represents `sameName`? What is its type?

We could say that we *only* hoist variables that are used in the case body. So if a variable is only mentioned in the guard expression, it gets declared right in its one case pattern and can have its own type independent of other cases.

But that leads to a chicken-and-end problem: We don't know which variables are used in the body until we analyze and resolve it. But we can't analyze it until we've defined the enclosing scopes. And we don't know which variables go into the hoisted shared scope and which are per case until we know which ones the body uses.

This is conceptually tractable. We could just do what the proposal says and eagerly synthesize a scope that contains the union of all variables in all cases and marks the ones with conflicting types as unusable. Then we analyze the body in that scope.

But then we would have to go *back* and re-resolve all of the guard expressions. Because whether or not a variable gets hoisted and shared across cases is user-visible when closures capture variables in guards. Currently, our implementations only do a single analysis pass over the code, and it would be a lot of work to have to do multiple passes.

### Hoisting all variables

A simpler solution is to say that *all* variables declared by a set of cases sharing a body get hoisted. We treat them as all being declared above the first case (or perhaps the first case where a variable with that name appears). Then in each case, we treat the variable pattern as *initializing* the variable and making it available for use. This lets you read every variable a case pattern declares in its own guard clause.

The case body is treated as a join point for flow analysis and definite assignment analysis. A variable is only usable in the body if *all* cases match that variable and mark it initialized. The desired behavior that you can only use a variable in the body if all cases define it falls out naturally from definite assignment analysis.

There are two problems, one minor and one less minor. The minor one: It may be surprising that the above example prints "guard1guard2" because it syntactically looks like two unrelated variable declarations. I think almost no users will ever observe this and if they do, we can explain it.

The larger problem is determining what type the hoisted variable should have. In the `sameName` example, the two patterns declare it with different types. To merge that to a single hoisted declaration, we need a single type. We could try something like `Object?` and then treat each pattern as *promoting* the variable to its matched type, but I think that feels pretty weird. The difference between a declared type and promoted type is user visible. I think it would be surprising, for example, if you could do:

```dart
switch (obj) {
  case [int sameName] when sameName > 0:
  case [String sameName] when sameName.isNotEmpty:
    // Assign a bool, allowed since declared type is Object?:
    sameName = true;
}
```

The only other option I can think of is that when multiple cases declare a variable with the same name, it *must* have the same type in all cases *even if the variable is never used in the body*.

That's a restriction from the current proposal, and could be annoying in examples like:

```dart
switch (obj) {
  case SomeClass(:int someProp) when someProp > 1:
  case OtherClass(:String someProp) when someProp.isNotEmpty:
    print('Not using someProp here.');
}
```

You wouldn't be able to use the field shorthand and instead have to do:

```dart
switch (obj) {
  case SomeClass(someProp: int intProp) when intProp > 1:
  case OtherClass(someProp: String stringProp) when someProp.isNotEmpty:
    print('Not using someProp here.');
}
```

It feels like the previous example should just work, but the language won't let you.

I think I'm OK with that restriction because a simpler model around flow analysis is generally a good thing for users too. There's maybe an argument that if you have cases sharing a body, you really *shouldn't* have variables with the same name with different types just because it's kind of confusing to read.

Thoughts?

munificent:
Fix #2553.

I think this captures the discussion on #2553 and what we talked about today, but let me know if I missed something. One restriction I ended up *not* doing is making it an error to have the same variable in different cases with different types. It's only an error to try to *use* it in the body. I didn't see any need to make it an error since we treat the variables in each case as separate declarations anyway.

Instead, we synthesize a scope for the body that has all variables declared by any of the cases. Then we determine at compile-time which ones do appear in all cases with the same type and finality. Those are the only ones that you can use inside the body. This feels fairly natural to me, but let me know if it seems hard to implement.

cc @leafpetersen @lrhn @eernstg @jakemac53 @natebosch @kallentu

### DIFF
--- a/working/0546-patterns/patterns-feature-specification.md
+++ b/working/0546-patterns/patterns-feature-specification.md
@@ -1863,13 +1863,15 @@ appears:
 
     The guard expression is evaluated in its case's case scope.
 
-    The body of a switch statement or expression case is executed in the case
-    body scope, defined below.
+    If the body of a switch statement or expression is reached through only a
+    single case, then it is executed in a new scope whose enclosing scope is the
+    case scope of that case. Otherwise, the body is executed in a new scope
+    whose enclosing scope is the shared case scope, defined below.
 
-    The then statement of an if-case statement is executed in the case's case
-    scope.
+    The then statement of an if-case statement is executed in a new scope whose
+    enclosing scope is the case's case scope.
 
-#### Case body scope and shared case bodies
+#### Shared case scope
 
 In a switch statement, multiple cases may share the same body. This introduces
 complexity when those cases declare variables which may or may not overlap and
@@ -1888,29 +1890,22 @@ Somehow, in the body, `a` refers to *both* of those pattern variables.
 Conversely, only the first case declares `n` which is used in that case's guard
 but not the body.
 
-We specify how this behaves by defining the *case body scope* for the body. If
-there is only a single case, then the case body scope is that case's case scope
-and we're done. *If there is only a single case for the body, then any pattern
-variables used in the body have a single declaration in that case's pattern and
-need no other special handling.*
-
-Otherwise, the case body scope `s` of a body used by a set of cases with pattern
+We specify how this behaves by creating a new *shared case scope* that contains
+all variables from all of the cases and then report errors from invalid uses of
+them. The shared case scope `s` of a body used by a set of cases with pattern
 variable sets `vs` (where default cases and labels have empty pattern variable
 sets) is:
 
 1.  Create a new empty scope `s` whose enclosing scope is the scope surrounding
-    the switch statement or expression. *Note that the body scope doesn't
-    include any of the case scopes as its enclosing scope. Instead, we
-    synthesize a scope that has new variables derived from the union of all of
-    the cases' variables.*
+    the switch statement or expression.
 
 2.  For each name `n` appearing as a variable name in any of the pattern
     variable sets in `vs`:
 
     1.  If `n` is defined in every pattern variable set in `vs` and has the same
-        type and finality, then introduce `n` into `s` with that type and
-        finality. This is a *consistent variable* and is available for use in
-        the body.
+        type and finality, then introduce `n` into `s` with the same type and
+        finality. This is a *shared variable* and is available for use in the
+        body.
 
         If any of the corresponding variables in `vs` are promoted, calculate
         the promoted type of the variable in `s` based on all of the promoted
@@ -1943,13 +1938,19 @@ sets) is:
         introduces a variable in the initializer list distinct from the
         initialized field.*
 
-    2.  Else `n` is inconsistently defined by the cases. Introduce a new
-        variable `n` into `s` with arbitrary type and finality.
+        *Note that we only create a shared case scope with its own variables
+        when there are multiple cases sharing a body. If there is only a single
+        case, the body uses that case's scope as the enclosing scope directly.
+        If you delete the second case in the above example, it prints "after".*
+
+    2.  Else `n` is not consistently defined by all cases and thus isn't safe to
+        use in the body. Introduce a new variable `n` into `s` with unspecified
+        type and finality.
 
 3.  Compile the body in `s`. It is a compile-time error if any identifier in the
-    body resolves to a variable in `s` that isn't consistent. *In other words,
-    all variables declared by any of the case patterns shadow outer variables,
-    but only the consistent ones can actually be used:*
+    body resolves to a variable in `s` that isn't shared. *In other words, a
+    variable declared by any of the case patterns shadows an outer variable, but
+    only the shared ones can actually be used:*
 
     ```dart
     var c = 'outer';
@@ -1961,12 +1962,12 @@ sets) is:
     ```
 
     *This has a compile-time error instead of printing "outer" because `c` in
-    the body resolves to an inconsistent variable declared by one of the cases.*
+    the body resolves to a non-shared variable declared by one of the cases.*
 
-*Note that it is not a compile-time error for there to _be_ inconsistently
-defined variables between cases. It's only an error to _use_ them in the body.
-This enables patterns to define inconsistent variables that are only used by
-their respective guards:*
+*Note that it is not a compile-time error for there to _be_ non-shared defined
+variables between cases. It's only an error to _use_ them in the body. This
+enables patterns to define non-shared variables that are only used by their
+respective guards:*
 
 ```dart
 switch (obj) {
@@ -1980,8 +1981,16 @@ switch (obj) {
 *This example has no errors because the only variable used in the body, `a`, is
 defined consistently by all cases.*
 
-At runtime, we initialize all of the consistent variables in the body of the
-case with the values of the corresponding case variables from the matched case.
+It is a compile-time error to assign to a variable declared by a case in that
+case's guard expression. *This helps avoid users running into the confusing
+behavior where the body sees a different variable than the guard when cases
+share a body. We make this an error even when the body only has a single case to
+keep the rule simpler for users to understand. This is similar to the
+restriction that you can't assign to the variable introduced by an initializing
+formal inside the initializer list.*
+
+At runtime, we initialize all of the shared variables in the body of the case
+with the values of the corresponding case variables from the matched case.
 
 ### Type promotion
 
@@ -2091,15 +2100,19 @@ fail in some way.*
         allowed to have multiple empty cases where all preceding ones share the
         same body with the last case.*
 
-    4.  Initialize any consistent variables declared in the body scope with the
-        values of the corresponding variables from the case scope. *There will
-        be no consistent variables and the body scope will be the case scope if
-        the body is only used by a single case.*
+    4.  If the enclosing scope for the body is a shared case scope, then
+        initialize all shared variables the values of the corresponding
+        variables from the case scope. *There will be no shared case scope and
+        nothing to copy if the body is only used by a single case.*
 
     5.  Execute the body statement.
 
-    6.  Exit the switch statement. *An explicit `break` is no longer
-        required.*
+    6.  If execution of the body statement continues with a label, and that
+        label is labeling a switch case of this switch, go to step 3 and
+        continue from that label.
+
+    7.  Otherwise the switch statement completes normally. *An explicit `break`
+        is no longer required.*
 
 3.  If no case pattern matched and there is a default clause, execute the
     statements after it.
--- a/working/0546-patterns/patterns-feature-specification.md
+++ b/working/0546-patterns/patterns-feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: In progress
 
-Version 2.10 (see [CHANGELOG](#CHANGELOG) at end)
+Version 2.11 (see [CHANGELOG](#CHANGELOG) at end)
 
 Note: This proposal is broken into a couple of separate documents. See also
 [records][] and [exhaustiveness][].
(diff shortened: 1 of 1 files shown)


scheglov (approved):

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1863,52 +1863,108 @@ appears:
...
+        }
+        ```
+
+        *This prints "before", not "after". In practice, users will rarely
+        notice this, the same way they rarely notice that an initializing formal
+        introduces a variable in the initializer list distinct from the
+        initialized field.*
+
+    2.  Else `n` is inconsistently defined by the cases. Introduce a new
+        variable `n` into `s` with arbitrary type and finality.
```
lrhn, line 1947: I'd prefer "unspecified" to "arbitrary". It really has no type or finality.

That way, if any part of the specification mistakenly refers to the type or finality of an inconsistent variable, it cannot consistently continue.

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2054,11 +2087,18 @@ fail in some way.*
...
-    3.  Execute the nearest non-empty case body at or following this case.
-        *You're allowed to have multiple empty cases where all preceding
-        ones share the same body with the last case.*
+    3.  Find the nearest non-empty case body at or following this case. *You're
+        allowed to have multiple empty cases where all preceding ones share the
+        same body with the last case.*
+
+    4.  Initialize any consistent variables declared in the body scope with the
+        values of the corresponding variables from the case scope. *There will
+        be no consistent variables and the body scope will be the case scope if
```
lrhn, line 2096: "be no inconsistent variables"?

Can we *not* make it the case scope, because if we do, we can't make the variables final in the case scope (which I want to).

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -2054,11 +2087,18 @@ fail in some way.*
...
+        the body is only used by a single case.*
 
-    4.  Exit the switch statement. *An explicit `break` is no longer
+    5.  Execute the body statement.
+
+    6.  Exit the switch statement. *An explicit `break` is no longer
```
lrhn, line 2101: "The switch statement completes normally."

(Using spec-terminology for "exiting".)

Actually, might need:

> 6. If execution of the body statement breaks without a label, then the switch statement completes normally.
> 7. If execution of the body statement continues with a label, and that label is labeling a
>     switch case of this `switch`, go to step 3 and continue from that label.
> 8. Otherwise the switch statement completes in the same way as the body statement.

(Or just leave this for the formal specification, and just be clear an understandable.)

working/0546-patterns/patterns-feature-specification.md
```diff
@@ -1863,52 +1863,108 @@ appears:
 
     The guard expression is evaluated in its case's case scope.
 
+    The body of a switch statement or expression case is executed in the case
+    body scope, defined below.
+
     The then statement of an if-case statement is executed in the case's case
     scope.
```
lrhn, line 1870: I'd prefer to do it in the body scope too, and make the variables of the case scope always be final.
(It's not important for `if` and single-`case` switch cases, but I like the consistency of doing it the same way everywhere. Having 
```dart
switch (e) { 
  case int e when ++e > 0: print(e); 
  case int e when --e < 0: print(e); 
}
```
work one way and
```dart
switch (e) { 
  case int e when ++e > 0:
  case int e when --e < 0: print(e); 
}
```
work another is error-prone.
So, don't reuse the case scope, even when you can.

lrhn (approved):

stereotype441 (approved):
lgtm since this definitely seems to be going in the right direction.  I feel like there are still some issues open, though.  I'll make a follow-up comment with the things I think we may still need to adress.

stereotype441:
Ok, thinking about it more I think there are two issues still open in my mind:

1. In our meeting yesterday we talked about the idea of prohibiting a guard from modifying a variable that's bound by the corresponding pattern.  I don't see anything in this PR about this; do we still want to do it?  (It sounds like @lrhn does 😃).  If so, do we want to do it everywhere (switch statements, switch expressions, if-case statements, if-case in collections)?  Just in switch statements where multiple cases share a body?  Something in between?  Also, how do we want to spec it?  Informally, we've talked about the idea that the variables are considered final inside the guard but non-final in the case body, but the analyzer doesn't have the ability to treat finality as a context-dependent property, and this is baked into its public API.  So it would be a lot of work to implement it that way.  I would favor instead having a separate rule in the spec saying that within a guard, it is a compile-time error to write to a variable that's bound anywhere in the corresponding pattern.

2. In our meeting yesterday we talked about the idea of specifying or-patterns similarly, so for instance the pattern `(int x, 1) | (0, int x)` would be considered to bind 3 separate variables: one for the first `x`, one for the second `x`, and one for an `x` that comes into existence at the "join point".  Do you have a preference about this?  There would be no user-visible difference in terms of program behavior or errors and warnings, but it insofar as the analyzer's API tries to match the spec, there would be a user-visible difference for users of the analyzer API.  And since the analyzer API to some degree dictates the implementation internals, it would be good to make a decision on this sooner rather than later, so that we can implement the underlying just once.  If you don't have a preference about this, let me know and I can convene a meeting with the analyzer and CFE folks to make a decision.  (CC @scheglov @johnniwinther @chloestefantsova in case any of them want to weigh in on the discussion here)

I'm totally fine landing this PR as is and then making follow-up adjusments later once we've decided on these two topics.

bwilkerson (commented):
This looks good to me, module the point Paul raised about assignability. Thanks!

working/0546-patterns/patterns-feature-specification.md
munificent: No, I meant "no consistent". If there aren't shared cases, the there are no consistent variables that are declared as distinct from the variables in the case pattern itself and that need to get initialized. Instead, the body refers directly to the actual variables from the pattern.

> Can we not make it the case scope, because if we do, we can't make the variables final in the case scope (which I want to).

Done. I did this deliberately because I intended to match parameter/function body scope. But it turns out I misremembered how function body scopes worked. :)

I want these two to behave the same:

```dart
test(int x) {
  var x = 2;
}

switch (0) {
  case int x: var x = 2;
}
```

Either both of these should be an error, or neither should be. I thought the former was an error so specified that the body uses the same scope as the scope where the pattern variables are declared, but I was wrong. There's no error in the former so there should be no error in the latter. Tweaked the proposal to state that the body has its own scope whose enclosing scope is the case body scope.

working/0546-patterns/patterns-feature-specification.md
munificent: Done, but combined 6 with 8.

working/0546-patterns/patterns-feature-specification.md
munificent: Made it an error to assign to case variables in the guard, so those examples are prohibited directly. Even so, introducing new variables for the body is still visible because the guard can capture the variable (which is different from the body's variable). Since that's visible and pretty weird (but will almost never be seen), I'd prefer to only do that when we need to because there's a shared case.

I think I could reasonably explain to a user why:

```dart
        Function captured;

        bool capture(Function closure) {
          captured = closure;
          return true;
        }

        switch (['before']) {
          case [String a] when capture(() => print(a)):
          case [_, String a]:
            a = 'after';
            captured();
        }
```

Prints "before". The `a` in the body *can't* be the "same" `a` is in the guard because there's clearly *two* `a` variable declarations.

But I'd have a real hard time explaining or justifying why this would print "before":

```dart
        Function captured;

        bool capture(Function closure) {
          captured = closure;
          return true;
        }

        switch (['before']) {
          case [String a] when capture(() => print(a)):
            a = 'after';
            captured();
        }
```

There's absolutely no need for the `a` in the body to be different. So since we can, I think it's best to actually make it the same variable. There's only one variable declaration. There's nothing to merge.

lrhn:
About `or`-patterns introducing three variables: I'd be fine with whatever is easier to specify and/or implement (and if there is no user-visible difference, I don't even care whether the spec and implementation agrees, the specification is satisfied if the visible behavior is indistinguishable from what is specified).

I can see that it's easier to make two variable declarations introduce two different variables, instead of two declarations defining 
the same variable. That's odd, but probably only if you think about it. So don't do that. 
(How do we handle `for(;;)` loops where the same variable is introduced by two different *paths*, even if it's not two different declarations?)

From a specification perspective, I can easily accept that the result of matching a pattern against a value is either a rejection, or a set of variable bindings (where the names and static/declared types are always known statically). Then *each subpattern* introduces its own binding "scope", and each compositional patter combines those into a new one, not just `or` patterns. 
That's a proper *compositional* definition of pattern matching.
That way, `var x & var y` sub-patterns would each introduce either a binding for `x` or a binding for `y` and the `&` pattern would combine those to a binding for both.
Even a sequence of `case` clauses just combine their resulting bindings in a specific way (and expose an environment based on the individual bindings to the `when` clauses). It's not *special* any more, every pattern does that.

Many more "scopes" and "bindings', but to the same final effect anyway.
(To put it differently: If that was the specification, would it worry you to implement the visible behavior of that?)

munificent:
> 1. In our meeting yesterday we talked about the idea of prohibiting a guard from modifying a variable that's bound by the corresponding pattern.  I don't see anything in this PR about this; do we still want to do it?

Yes, I just, uh, forgot to add it. Done. 🙃 

> 1. If so, do we want to do it everywhere (switch statements, switch expressions, if-case statements, if-case in collections)?

I think it's best to just do it unconditionally. I think it's essentially an anti-pattern anyway, so may as well forbid it everywhere.

> I would favor instead having a separate rule in the spec saying that within a guard, it is a compile-time error to write to a variable that's bound anywhere in the corresponding pattern.

Me too, that's what I added.

> 2\. In our meeting yesterday we talked about the idea of specifying or-patterns similarly, so for instance the pattern `(int x, 1) | (0, int x)` would be considered to bind 3 separate variables: one for the first `x`, one for the second `x`, and one for an `x` that comes into existence at the "join point".  Do you have a preference about this?

I don't have a preference. Since, as far as I know, the choice here doesn't affect any visible semantics, I'm content to have the proposal be silent on this and leave it up to implementations as to the best way to model it.

Revised. PTAL, thanks!

Issue #2611, opened 2022-11-07
Definition for assignable expression needs clarification


jmewes:
## Observed

In section "17.36 Assignable Expressions" of [DartLangSpecDraft.pdf](https://spec.dart.dev/DartLangSpecDraft.pdf) it is written that:

> Assignable expressions are expressions that can appear on the left hand side of an assignment.

Further, it is mentioned that an _assignable expression_ can be:

> An invocation (possibly conditional) of a getter (10.3) or list access operator on an expression e.

I understand the term "on the left hand side of an assignment" as left to the assignment operator:

```
<expression> ::= <assignableExpression> <assignmentOperator> <expression>
  | ...
```

Here is a code snippet that tries to provide an example for the definitions above:

```dart
class Qux {
  int _q = 0;
  
  int get quux => _q;
}

void main() {
   var qux = Qux();
  
   qux.quux = 42; // Error: The setter 'quux' isn't defined for the class 'Qux'.
// ^      ^
// |------|-------  invocation of a getter, used as an assignable expression, as defined above
}
```

(https://dartpad.dev/?id=7c25b19d3197a554b382cf50263b4c2a)

I understand `qux.quux` as an invocation of a getter that - according to the quotes above - should be usable on the left hand side of an assignment. However, the code fails to compile since a setter is expected instead of the invocation of a getter.

## Expected

~For me, section "17.36 Assignable Expressions" would make sense if it would mention "setter" instead of "getter". The updated definition would then look like this:~

> ~An assignable expression is either: (...) An invocation (possibly conditional) of a **setter** or list access operator on an expression e. (...)~

Here is a code snippet that tries to provide an example for this updated definition:

```dart
class Qux {
  int _q = 0;
  
  set quux(int i) {
    _q = i;
  }
}

void main() {
   var qux = Qux();
  
   qux.quux = 42;
// ^      ^
// |------|-------  invocation of a setter, used as an assignable expression, as defined above
  
   print(qux._q); // 42
}

```

(https://dartpad.dev/?id=c0a1bd1036ad850e4d456ab1378f4561)

## Steps to reproduce

See https://github.com/dart-lang/language/blob/6727010/specification/dartLangSpec.tex#L16282-L16283

## References

- Also see the following discussion in the `#language` channel of [The Unofficial Dart Community Discord](https://discord.gg/Qt6DgfAWWx): https://discord.com/channels/823048554532700191/870318216747024414/1038373764674633780
[...]

PR #2615, opened 2022-11-08
Clarify section 'Assignable Expressions'

eernstg:
Cf. https://github.com/dart-lang/language/issues/2611. The section 'Assignable Expressions' in the language specification is rather confusing. This PR makes it more explicit how we obtain the subterm of any given assignable expression that needs to be evaluated, if any (in `e.x = 42`, we need to evaluate `e`). It no longer specifies bottom-up how to evaluate forms like `e[v]`: It notes that when this subterm is an expression, it is evaluated following the same rules as any other expression.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -5947,7 +5947,7 @@ We define the
 \IndexCustom{instantiated \ON{} type}{extension!instantiated \ON{} type}
 of $a$ as $[T_1/X_1, \ldots, T_s/X_s]T$.
 We define the
-\IndexCustom{instantiation-to-bound \ON{} type}{
+\IndexCustom{instantiation-to-bound \ON{} type}{%
   extension!instantiation-to-bound \ON{} type}
 of $a$ as $[U_1/X_1, \ldots, U_s/X_s]T$,
 where \List{U}{1}{s} is the result of instantiation to bound
@@ -6022,7 +6022,7 @@ $s_1 = [T_1/X_1, \ldots, T_k/X_k]s$:
 
 \LMHash{}%
 The member signature $s_1$ is called the
-\IndexCustom{invocation member signature}{
+\IndexCustom{invocation member signature}{%
   extension!invocation member signature}
 of $i$.
 The static type of $i$ is the return type of
@@ -13372,7 +13372,7 @@ on a
 expression $r$ is an expression of
 one of the forms shown in Fig.~\ref{fig:memberInvocations}.
 Each member invocation has a
-\IndexCustom{corresponding member name}{
+\IndexCustom{corresponding member name}{%
   member invocation!corresponding member name}
 as shown in the figure.
 
@@ -13713,7 +13713,7 @@ and if the method lookup succeeded then let $F$ be the static type of $d$.
 Otherwise, let $d$ be the result of getter lookup
 for $m$ in $T$ with respect to $L$,
 and let $F$ be the return type of $d$.
-(\commentary{
+(\commentary{%
 Since \code{$T$.$m$} exists we cannot have a failure in both lookups.%
 })
 If the getter return type $F$ is an interface type
@@ -13737,7 +13737,7 @@ and the static type of $i$ is as specified there.
 \LMHash{}%
 It is a compile-time error to invoke an instance method on a type literal
 that is immediately followed by the token `.' (a period).
-\commentary{
+\commentary{%
 For instance, \code{int.toString()} is an error.%
 }
 
@@ -14010,7 +14010,7 @@ as well as when it does not exist at all.%
 
 \LMHash{}%
 An
-\IndexCustom{implicit \CALL{} superinvocation}{
+\IndexCustom{implicit \CALL{} superinvocation}{%
   method superinvocation!implicit \CALL}
 has the form
 
@@ -20325,13 +20325,13 @@ $X_j <: B_j$, for all $j \in 1 .. s$,
 it is a compile-time error if $T$ is not regular-bounded,
 and it is a compile-time error if any type occurring in $T$ is not well-bounded.
 
-\commentary{
+\commentary{%
 This means that the bounds declared for
 the formal type parameters of a generic type alias
 must be such that when they are satisfied,
 the bounds that pertain to the body are also satisfied,
 and a type occurring as a subterm of the body can violate its bounds,
-but only if it is a correct super-bounded type.
+but only if it is a correct super-bounded type.%
 }
 
 \LMHash{}%
@@ -20469,7 +20469,7 @@ When $D$ is a type alias declaration of the form
 we say that the parameterized type $U$ of the form
 \code{$F$<\List{U}{1}{s}>}
 in a scope where $F$ resolves to $D$
-\IndexCustom{alias expands in one step to}{
+\IndexCustom{alias expands in one step to}{%
   type alias!alias expands in one step}
 $[U_1/X_1, \ldots, U_s/X_s]T$.
 
@@ -20484,7 +20484,7 @@ which is a parameterized type applying a type alias to some type arguments
 by its alias expansion in one step
 (\commentary{including the non-generic case where there are no type arguments}).
 When no further steps are possible we say that the resulting type is the
-\IndexCustom{transitive alias expansion}{
+\IndexCustom{transitive alias expansion}{%
   type alias!transitive alias expansion}
 of $U$.
 
@@ -20756,7 +20756,7 @@ and the subtype relationship is always determined in the same way.
   \begin{minipage}[c]{0.49\textwidth}
     \RuleTwo{\SrnLeftFutureOr}{Left FutureOr}{S}{T}{%
       \code{Future<$S$>}}{T}{\code{FutureOr<$S$>}}{T}
-    \RuleTwo{\SrnRightPromotedVariable}{Right Promoted Variable}{S}{X}{S}{T}{
+    \RuleTwo{\SrnRightPromotedVariable}{Right Promoted Variable}{S}{X}{S}{T}{%
       S}{X \& T}
     \Rule{\SrnRightFutureOrB}{Right FutureOr B}{S}{T}{S}{\code{FutureOr<$T$>}}
     \Rule{\SrnLeftVariableBound}{Left Variable Bound}{\Gamma(X)}{T}{X}{T}
@@ -20766,7 +20766,7 @@ and the subtype relationship is always determined in the same way.
     \Rule{\SrnRightFutureOrA}{Right FutureOr A}{S}{\code{Future<$T$>}}{%
       S}{\code{FutureOr<$T$>}}
     \Rule{\SrnLeftPromotedVariable}{Left Promoted Variable B}{S}{T}{X \& S}{T}
-    \RuleRaw{\SrnRightFunction}{Right Function}{T\mbox{ is a function type}}{
+    \RuleRaw{\SrnRightFunction}{Right Function}{T\mbox{ is a function type}}{%
       T}{\FUNCTION}
   \end{minipage}
   %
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -16245,18 +16246,20 @@ \subsection{Assignable Expressions}
 \LMLabel{assignableExpressions}
 
 \LMHash{}%
-Assignable expressions are expressions
+\IndexCustom{Assignable expressions}{assignable expression} are terms
 that can appear on the left hand side of an assignment.
-This section describes how to evaluate these expressions
+This section describes how to evaluate these terms
 when they do not constitute the complete left hand side of an assignment.
```
lrhn, line 16252: Maybe add: "\commentary{That is, when they are being evaluated for their value, not being assigned to.}"
Or just add "to a value" to the "evaluated" below. Just for clarity.

specification/dartLangSpec.tex
```diff
@@ -16245,18 +16246,20 @@ \subsection{Assignable Expressions}
...
-one would have no need for their value,
-and the rules for evaluating them would be unnecessary.
-However, assignable expressions can be subexpressions of other expressions
-and therefore must be evaluated.%
+\commentary{%
+Assignable expressions can be subexpressions of other expressions,
+in which case they need to be evaluated.
+For example, if we are assigning a value to an instance variable \code{x}
+of a receiver $e$ by means of the expression \code{$e$.x\,\,=\,\,42},
+we need to evaluate $e$,
```
lrhn, line 16261: ... to an object,

But it's not actually clear why this is relevant, because we haven't said that `e` might be an assignable expression. This is just about expressions, not assignable expressions.

specification/dartLangSpec.tex
```diff
@@ -16245,18 +16246,20 @@ \subsection{Assignable Expressions}
...
+such that we can update \code{x} in the correct object.%
```
lrhn, line 16262: in -> of

(`x` is a property _of_ the object, not a value _in_ the object.)

specification/dartLangSpec.tex
```diff
@@ -16245,18 +16246,20 @@ \subsection{Assignable Expressions}
...
-always appeared \emph{as} the left hand side,
-one would have no need for their value,
-and the rules for evaluating them would be unnecessary.
-However, assignable expressions can be subexpressions of other expressions
-and therefore must be evaluated.%
+\commentary{%
+Assignable expressions can be subexpressions of other expressions,
+in which case they need to be evaluated.
+For example, if we are assigning a value to an instance variable \code{x}
+of a receiver $e$ by means of the expression \code{$e$.x\,\,=\,\,42},
```
lrhn, line 16260: receiver -> receiver expression

specification/dartLangSpec.tex
```diff
@@ -16276,38 +16279,29 @@ \subsection{Assignable Expressions}
...
+\commentary{%
+In short, we obtain $t$ by cutting off an assignable selector
+from the end of $e$.
+It is easy to see that every \synt{assignableExpression}
+except the ones that are simply an \synt{identifier}
+ends in an \synt{assignableSelector}
+(an \synt{unconditionalAssignableSelector}
+is also an \synt{assignableSelector}).
+Note that $t$ is an \synt{expression}:
+Every possible value for $t$ can also be derived from \synt{expression}.%
```
lrhn, line 16299: Except if `t` is `super`.
And if `t` is an identifier denoting an import prefix, then it's *technically* an expression, because it's an identifier, but not one which can be *evaluated*.
And if `t` denotes a class/mixin/extension declaration, it's also not evaluated (unless we still define static lookup as members on the `Type` object.)

specification/dartLangSpec.tex
```diff
@@ -16276,38 +16279,29 @@ \subsection{Assignable Expressions}
...
+}
 
 \LMHash{}%
-Evaluation of an assignable expression of the form \code{\SUPER{}[$e_2$]}
-is equivalent to
-evaluation of the method invocation \code{\SUPER.[]($e_2$)}.
+Evaluation of the receiver expression $t$ of a given assignable expression $e$
+proceeds in the same way as evaluation of any other expression.
```
lrhn, line 16304: Except when you cannot evaluate `t`.

specification/dartLangSpec.tex
```diff
@@ -16276,38 +16279,29 @@ \subsection{Assignable Expressions}
...
-An assignable expression of the form \code{$e$.\id} or \code{$e$?.\id}
-is evaluated as a property extraction (\ref{propertyExtraction}).
-
-\LMHash{}%
-An assignable expression of the form \code{$e_1$[$e_2$]}
-is evaluated as a method invocation of
-the operator method \code{[]} on $e_1$ with argument $e_2$.
-
-\LMHash{}%
-An assignable expression of the form \code{\SUPER.\id} is evaluated as
```
lrhn, line 16304: We had special cases for `super` before, I think we still need them.
(Until we do something like https://github.com/dart-lang/language/blob/master/working/specification/compositional%20semantics.md).

specification/dartLangSpec.tex
eernstg: It is about an assignable expression that contains an expression as the subterm `$e$`: We can't have `\code{$e$.x = 42}` unless `$e$.x` is an assignable expression.

Given that this is commentary, I'm jumping to the claim that `$e$` is an expression (that's often true, and that is indeed the case where expression evaluation is needed). Given that `$e$` could be arbitrarily complex, we certainly want to rely on expression evaluation (to avoid repeating every rule about expression evaluation here, just because the grammar rule doesn't parse `$e$` as an expression).

specification/dartLangSpec.tex
eernstg: True! Fixed elsewhere.

specification/dartLangSpec.tex
eernstg: Yes, we certainly need to mention those situations where `t` isn't an expression. Done.

specification/dartLangSpec.tex

specification/dartLangSpec.tex
eernstg: I think it's considerably more correct now! ;-)

eernstg (commented):
Review response.

specification/dartLangSpec.tex
```diff
@@ -16276,38 +16279,40 @@ \subsection{Assignable Expressions}
...
-a property extraction.
+\BlindDefineSymbol{e, t}%
+Let $e$ be an \synt{assignableExpression}.
+Let $t$ be a term such that $e$ can be derived from
+\syntax{$t$ <assignableSelector>}.
+We say that $t$ is the
+\IndexCustom{receiver term}{assignable expression!receiver term}
+of $e$.
+When $t$ is an expression, we say that $t$ is the
+\IndexCustom{receiver expression}{assignable expression!receiver expression}
```
lrhn, line 16290: (Can't see it in this change, so just asking instead of trying to find the text:
Do we include `prefix` as a receiver expression? That is, do we differentiate between receiver terms which are grammatically an expression or not, or between receiver terms which are *semantically* evaluable expressions or not? Should we? It might make things easier in other places.)

specification/dartLangSpec.tex
```diff
@@ -16245,18 +16246,20 @@ \subsection{Assignable Expressions}
...
-and the rules for evaluating them would be unnecessary.
-However, assignable expressions can be subexpressions of other expressions
-and therefore must be evaluated.%
+\commentary{%
+An assignable expression can be a subterm of another term,
+in which case it may need to be evaluated to an object.
+For example, if we are assigning a value to an instance variable \code{x}
+of a receiver expression $e$ by means of the expression \code{$e$.x\,\,=\,\,42},
+we need to evaluate $e$
+such that we can update the \code{x} of the correct object.%
```
lrhn, line 16262: Unclear that `e` is the assignable expression.
We are talking about assignable expressions, and the first line of the example introduces a variable `x`, which is the quintessential assignable expression, so the reader may easily jump to the conclusion that `x` is the assignable expression of this example. In fact it's `e`, which is never said to be an assignable expression.

(Would also like to expand on the "may" above, with an example of an assignable expression which is not evaluated.)

Maybe:
```latex
\commentary{%
An expression which is grammatically an assignable expression
can occur in a position where it's not assigned to, but is instead
used to refer to a value or scope, and may be evaluated to a value.

Because the grammar of assignable expressions includes the very general
single and qualified identifiers, an assignable expression can
have many different meanings, depending on binding of those identifiers
in the context.

For example, an expression \code{x.y} is grammatically an assignable expression.
That expression can occur as a subexpression of a larger expression, 
for example as code{x.y.z = 42}. 
In that position \code{x.y} may refer to a getter \code{y} on an object referenced by a variable \code{x},
in which case \code{x.y} will be evaluated to an object before accessing the `z` member of that object.
The code\{x.y} can also denote a class \code{y} referenced through an import prefix \code{x},
with \code{z} denoting a static variable of that class, 
in which case \code{x.y} will not be evaluated to a value.
}
```

specification/dartLangSpec.tex
```diff
@@ -16276,38 +16279,40 @@ \subsection{Assignable Expressions}
...
+of $e$.
+
+\commentary{%
+In short, we obtain $t$ by cutting off an assignable selector
+from the end of $e$.
+It is easy to see that every \synt{assignableExpression}
+except one which is simply an \synt{identifier}
+ends in an \synt{assignableSelector}
+(an \synt{unconditionalAssignableSelector}
+is also an \synt{assignableSelector}).
```
lrhn, line 16300: Parenthetical comment is probably not needed.
Maybe just say
> ends in an (either null-aware or unconditional) selector.

specification/dartLangSpec.tex
```diff
@@ -16276,38 +16279,40 @@ \subsection{Assignable Expressions}
...
+In other words, every assignable expression except \synt{identifier}
+has a receiver term.
+
+Note that the receiver term $t$ may or may not be an \synt{expression}.
+For example, if $t$ is \SUPER{} then $e$ does not have a receiver expression.
```
lrhn, line 16305: You say "for example". Are there any other cases?
If not, use "In particular" instead.

specification/dartLangSpec.tex
```diff
@@ -16276,38 +16279,40 @@ \subsection{Assignable Expressions}
...
+}
 
 \LMHash{}%
-Evaluation of an assignable expression of the form \code{\SUPER{}[$e_2$]}
-is equivalent to
-evaluation of the method invocation \code{\SUPER.[]($e_2$)}.
+Let $e$ be an assignable expression.
+Assume that $e$ has a receiver expression $t$.
+Evaluation of $t$ proceeds in the same way as
+evaluation of any other expression.
```
lrhn, line 16315: Yup! This fails if `t` is a prefix, but an identifier _is_ (grammatically) an expression.

specification/dartLangSpec.tex
eernstg: The text now states explicitly that the cases in section Assignment are applicable when the side conditions are satisfied (e.g., `C.v = e` is applicable when `C` is a type literal), and that the forms containing expressions (such as `e1.v = e2`) are only applicable when the more specific forms do not apply.

specification/dartLangSpec.tex
eernstg: Done.

specification/dartLangSpec.tex
```diff
@@ -16276,38 +16279,40 @@ \subsection{Assignable Expressions}
...
-\LMHash{}%
-An assignable expression of the form \code{$e_1$[$e_2$]}
-is evaluated as a method invocation of
-the operator method \code{[]} on $e_1$ with argument $e_2$.
-
-\LMHash{}%
-An assignable expression of the form \code{\SUPER.\id} is evaluated as
-a property extraction.
+\BlindDefineSymbol{e, t}%
+Let $e$ be an \synt{assignableExpression}.
```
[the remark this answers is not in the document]
eernstg: Done.

specification/dartLangSpec.tex
eernstg: This text is gone today.

specification/dartLangSpec.tex
eernstg: This text is gone today.

specification/dartLangSpec.tex
eernstg: The text now explicitly states that cases with side conditions other than 'the subterm is an expression' are more specific. So that case is handled by the case that explicitly talks about import prefixes.

eernstg (commented):
Rewritten extensively. The text now specifies explicitly that each case in section Assignment which requires specific subterms (like `p.v = e` where `p` must be an import prefix) is applied when possible, and the cases where subterms are denoted as expressions are given lower priority.

Also the notions of 'the receiver term' and 'the receiver expression' are specified with an explicit mention of the possibility that a given assignable expression doesn't have those subterms. No problem: The cases where a receiver expression is needed are just not applicable when there is no receiver expression.

Note that the assignment section needs a revision, too. For instance, there is no case for `prefix.ClassName.staticVariable = e`. Cf. https://github.com/dart-lang/language/issues/2670.

lrhn (approved):

PR #2626, opened 2022-11-11
Add feature specification for inline classes

eernstg:
This PR adds a feature specification of 'inline classes', which was obtained by taking the current version of the feature specification of 'views' and rename the construct from 'view class' to 'inline class'. It also removes the support for primary constructors (such that we get more time to get that feature just right and introduce it in general, not just for inline classes).

Finally, this PR adjusts the text accordingly. In particular, it changes the word 'superview' to 'superinterface', and it specifies that it is a compile-time error for an inline class to have a type `T` in the `implements` clause unless `T` is an inline class.

The rules about name clashes among "inherited" names are unchanged: It is a compile-time error if an inline class `V` has superinterfaces `V1` and `V2`, and both of them has a member named `m`, and it is not the same declaration (e.g., in a diamond superinterface structure). The error is eliminated by adding a declaration named `m` to `V`.

### DIFF
--- a/accepted/future-releases/inline-classes/feature-specification.md
+++ b/accepted/future-releases/inline-classes/feature-specification.md
@@ -741,16 +741,18 @@ type.
 `R`, and `this` has type `V`.*
 
 Let _DV_ be an inline class declaration named `V` with representation
-type `R`. Assuming that all types have been fully alias expanded,
-we say that _DV_ is raw-dependent on an inline class declaration
-_DV2_ if `R` contains an identifier `id` (possibly qualified) that
-resolves to _DV2_, or `id` resolves to an inline class declaration
-_DV3_ and _DV3_ is raw-dependent on _DV2_.
+type `R`. Assuming that all types have been fully alias expanded, we
+say that _DV_ has a representation dependcy on an inline class
+declaration _DV2_ if `R` contains an identifier `id` (possibly
+qualified) that resolves to _DV2_, or `id` resolves to an inline class
+declaration _DV3_ and _DV3_ has a representation dependency on _DV2_.
 
-It is a compile-time error if an inline class declaration is
-raw-dependent on itself.
+It is a compile-time error if an inline class declaration has a
+representation dependency on itself.
 
-*In other words, cycles are not allowed.*
+*In other words, cycles are not allowed. This ensures that it is
+always possible to find a non-inline type which is the ultimate
+representation type of any given inline type.*
 
 An inline class declaration _DV_ named `Inline` may declare one or
 more constructors. A constructor which is declared in an inline class
--- a/accepted/future-releases/inline-classes/feature-specification.md
+++ b/accepted/future-releases/inline-classes/feature-specification.md
@@ -434,11 +434,11 @@ T<sub>s</sub></code> are implicit in the actual syntax.*
 We need to introduce a concept that is similar to existing concepts
 for regular classes.
 
-We say that an inline class `V` _has_ a member named `n` in the case
-where `V` declares a member named `n`, and in the case where `V` has
-no such declaration, but `V` has a superinterface `Vs` that has a
-member named `n`. In both cases,
-_the member declaration named `n` that `V` has_ is said declaration.
+We say that an inline class declaration _DV_ _has_ a member named `n`
+in the case where _DV_ declares a member named `n`, and in the case
+where _DV_ has no such declaration, but _DV_ has a direct
+superinterface `V` that has a member named `n`. In both cases,
+_the member declaration named `n` that DV has_ is said declaration.
 
 *This definition is unambiguous for an inline class that has no
 compile-time errors, because name clashes must be resolved by `V`.*
@@ -719,11 +719,11 @@ inline class. *For non-generic inline classes, the representation type
 is the same in either case.*
 
 Let `V` be an inline type of the form
-<code>Inline&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;</code>,
-and let `R` be the corresponding instantiated representation type.
-`V` is a proper subtype of `Object?`, and potentially non-nullable. If
-`R` is non-nullable then `V` is a proper subtype of `Object` as well,
-and non-nullable.
+<code>Inline&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;</code>, and let
+`R` be the corresponding instantiated representation type.  If `R` is
+non-nullable then `V` is a proper subtype of `Object`, and `V` is
+non-nullable.  Otherwise, `V` is a proper subtype of `Object?`, and
+`V` is potentially nullable.
 
 *That is, an expression of an inline type can be assigned to a top
 type (like all other expressions), and if the representation type is
@@ -749,7 +749,7 @@ type.
 
 Let _DV_ be an inline class declaration named `V` with representation
 type `R`. Assuming that all types have been fully alias expanded, we
-say that _DV_ has a representation dependcy on an inline class
+say that _DV_ has a representation dependency on an inline class
 declaration _DV2_ if `R` contains an identifier `id` (possibly
 qualified) that resolves to _DV2_, or `id` resolves to an inline class
 declaration _DV3_ and _DV3_ has a representation dependency on _DV2_.
--- a/accepted/future-releases/inline-classes/feature-specification.md
+++ b/accepted/future-releases/inline-classes/feature-specification.md
@@ -387,11 +387,12 @@ the inline class declaration with a fixed lookahead.*
 
 A few errors can be detected immediately from the syntax:
 
-A compile-time error occurs unless the inline class declares exactly
-one instance variable. Let `v` be the name of said instance
-variable. The declaration of `v` must have the modifier `final`
-and it must not have the modifier `late`; otherwise a compile-time
-error occurs.
+A compile-time error occurs if the inline class does not declare any
+instance variables, and if it declares two or more instance
+variables. Let `v` be the name of unique instance variable that it
+declares. The declaration of `v` must have the modifier `final`, and it
+can not have the modifier `late`; otherwise a compile-time error
+occurs.
 
 The _name of the representation_ in an inline class declaration is the
 name `id` of the unique final instance variable that it declares, and
@@ -739,11 +740,11 @@ Assume that _DV_ has two direct or indirect superinterface of the form
 respectively
 <code>W&lt;S<sub>1</sub>, .. S<sub>k</sub>&gt;</code>.
 A compile-time error
-occurs unless
+occurs if
 <code>T<sub>j</sub></code>
-is equal to
+is not equal to
 <code>S<sub>j</sub></code>
-for each _j_ in _1 .. k_. The notion of equality used here is the same
+for any _j_ in _1 .. k_. The notion of equality used here is the same
 as with the corresponding rule about superinterfaces of classes.
 
 Assume that an inline class declaration _DV_ named `Inline` has
@@ -751,7 +752,7 @@ representation type `R`, and that the inline type `V1` with
 declaration _DV1_ is a superinterface of _DV_ (*note that `V1` may
 have some actual type arguments*).  Assume that `S` is the
 instantiated representation type corresponding to `V1`. A compile-time
-error occurs unless `R` is a subtype of `S`.
+error occurs if `R` is not a subtype of `S`.
 
 *This ensures that it is sound to bind the value of `id` in _DV_ to `id1`
 in `V1` when invoking members of `V1`, where `id` is the representation
--- a/accepted/future-releases/inline-classes/feature-specification.md
+++ b/accepted/future-releases/inline-classes/feature-specification.md
@@ -590,6 +590,9 @@ the case where the values of the type variables do not satisfy their
 declared bounds, and those values will be obtained directly from the static
 type of the receiver in each member invocation on `V`.*
 
+A compile-time error occurs if a type parameter of an inline class
+occurs in a non-covariant position in the representation type.
+
 When `s` is zero,
 <code>V&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;</code>
 simply stands for `V`, a non-generic inline type.
@@ -641,6 +644,13 @@ and no actual argument part.
 
 *Setter invocations are treated as invocations of methods with a
 single argument.*
+
+If `e` is an expression whose static type `V` is the inline type
+<code>Inline&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;</code>
+and `V` has no member whose basename is the basename of `m`, a member
+access like `e.m(args)` may be an extension member access, following
+the normal rules about applicability and accessibility of extensions,
+in particular that `V` must match the on-type of the extension.
 
 *In the body of an inline class declaration _DV_ with name `Inline`
 and type parameters
@@ -704,15 +714,15 @@ is `R`, and the _instantiated representation type_ corresponding to
 
 We will omit 'declared' and 'instantiated' from the phrase when it is
 clear from the context whether we are talking about the inline class
-itself, or we're talking a particular instantiation of a generic
-inline. *For non-generic inline classes, the representation type is
-the same in either case.*
+itself, or we're talking about a particular instantiation of a generic
+inline class. *For non-generic inline classes, the representation type
+is the same in either case.*
 
 Let `V` be an inline type of the form
 <code>Inline&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;</code>,
 and let `R` be the corresponding instantiated representation type.
 `V` is a proper subtype of `Object?`. If `R` is non-nullable then `V`
-is a proper subtype of `Object` as well.
+is a proper subtype of `Object` as well, and non-nullable.
 
 *That is, an expression of an inline type can be assigned to a top type
 (like all other expressions), and if the representation type is
@@ -768,10 +778,13 @@ A compile-time error occurs if an inline class constructor includes a
 superinitializer. *That is, a term of the form `super(...)` or
 `super.id(...)` as the last element of the initializer list.*
 
+A compile-time error occurs if an inline class constructor declares a
+super parameter. *For instance, `Inline(super.x);`.*
+
 *In the body of a generative inline class constructor, the static type
-of `this` is the same as it is in any instance member of the inline,
-that is, `Inline<X1 .. Xk>`, where `X1 .. Xk` are the type parameters
-declared by `Inline`.*
+of `this` is the same as it is in any instance member of the inline
+class, that is, `Inline<X1 .. Xk>`, where `X1 .. Xk` are the type
+parameters declared by `Inline`.*
 
 An instance creation expression of the form
 <code>Inline&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;(...)</code>
@@ -787,6 +800,9 @@ variable, which is initialized according to the normal rules for
 constructors (in particular, it can occur by means of `this.id`, or in
 an initializer list, or by an initializing expression in the
 declaration itself, but it is an error if it does not occur at all).*
+
+An inline type `V` used as an expression (*a type literal*) is allowed
+and has static type `Type`.
 
 
 ### Composing Inline Classes
@@ -987,9 +1003,14 @@ use a cast to introduce or discard the inline type, as the static type
 of an instance, or as a type argument in the static type of a data
 structure or function involving the inline type.*
 
-A type test, `o is U` or `o is! U`, and a type cast, `o as U`, where `U` is
-or contains an inline type, is performed at run time as a type test and type
-cast on the run-time representation of the inline type as described above.
+A type test, `o is U` or `o is! U`, and a type cast, `o as U`, where
+`U` is or contains an inline type, is performed at run time as a type
+test and type cast on the run-time representation of the inline type
+as described above.
+
+An inline type `V` used as an expression (*a type literal*) evaluates
+to the value of the corresponding instantiated representation type
+used as an expression.
 
 
 ### Summary of Typing Relationships
--- a/accepted/future-releases/inline-classes/feature-specification.md
+++ b/accepted/future-releases/inline-classes/feature-specification.md
@@ -84,10 +84,6 @@ representation) is available as a declared name: The inline class must
 have exactly one instance variable whose type is the representation
 type, and it must be `final`.
 
-The representation type of the inline class (when the unique field of
-the inline class is `final int i;`, that's `int`) is similar to the
-on-type of an extension declaration.
-
 All in all, an inline class allows us to replace the interface of a given
 representation object and specify how to implement the new interface
 in terms of the interface of the representation object.
@@ -389,8 +385,8 @@ A few errors can be detected immediately from the syntax:
 
 A compile-time error occurs if the inline class does not declare any
 instance variables, and if it declares two or more instance
-variables. Let `v` be the name of unique instance variable that it
-declares. The declaration of `v` must have the modifier `final`, and it
+variables. Let `id` be the name of unique instance variable that it
+declares. The declaration of `id` must have the modifier `final`, and it
 can not have the modifier `late`; otherwise a compile-time error
 occurs.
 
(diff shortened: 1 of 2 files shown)


accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,948 @@
...
+provides an interface which is chosen independently of the interface
+of the wrapped object, and it provides implementations of the members
+of the interface of the inline class, and those implementations can use
+the members of the wrapped object as needed.
+
+However, even though an inline class behaves like a wrapping, the wrapper
+object will never exist at run time, and a reference whose type is the
+inline class will actually refer directly to the underlying wrapped
+object. Every member access (e.g., an invocation of a method or a
+getter) on an expression whose static type is an inline type will invoke
```
lrhn, line 60: Comma before "will"?

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,948 @@
...
+Given that there is no wrapper object, we will refer to the "wrapped"
+object as the _representation object_ of the inline class, or just the
+_representation_.
+
+Inside the inline class declaration, the keyword `this` is a reference
+to the representation whose static type is the enclosing inline
+class. A member access to a member of the enclosing inline class may
+rely on `this` being induced implicitly (for example, `foo()` means
+`this.foo()` if the inline class contains a method declaration named
+`foo`).
```
lrhn, line 74: , and there is no other `foo` in the lexical scope.

(If we have inheritance of inline class members, the `foo` of `this` might not be in the lexical scope, and another top-level declaration might be.)

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,948 @@
...
+
+A reference to the representation typed by its run-time type or a
+supertype thereof (that is, typed by a "normal" type for the
+representation) is available as a declared name: The inline class must
+have exactly one instance variable whose type is the representation
+type, and it must be `final`.
+
+The representation type of the inline class (with `final int i` that's
```
lrhn, line 82: Put a `;` after `final int i`?

Not *clear* that `final int i;` is the instance variable of the inline class. Maybe say that explicitly.

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,948 @@
...
+*The token `inline` is not made a built-in identifier: the reserved
+word `class` that occurs right after `inline` serves to disambiguate
+the inline class declaration with a fixed lookahead.*
+
+A few errors can be detected immediately from the syntax:
+
+A compile-time error occurs unless the inline class declares exactly
+one instance variable. Let `v` be the name of said instance
+variable. A compile-time error occurs unless the declaration of `v`
+has the modifier `final` or the modifier `late`.
```
lrhn, line 386: Do you mean ", or if it has the modifier `late`"? 
Here the "or" is an alternative to having the modifier `final`, so `late int i` is valid.

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,948 @@
...
+
+Again, let `V` be an inline type of the form
+<code>Inline&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;</code>,
+and let `R` be the corresponding instantiated representation type.
+If `R` is not an inline type then we say that `V` is an inline type
+at level zero. If `R` is an inline type at level _k_ then we say that
+`V` is an inline type at level _k + 1_.
+A compile-time error occurs if the level of `V` is undefined.
+
+*In other words, cycles are not allowed.*
```
lrhn, line 647: Nor is:
```dart
inline class C<T> {
  final T value;
}
```
I assume?

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,948 @@
...
+an approach for building an instance of the representation type of an
+inline declaration _DV_ with _DV_ itself, which makes it easy to
+recognize that this is a way to obtain a value of that inline type. It
+can also be used to verify that an existing object (provided as an
+actual argument to the constructor) satisfies the requirements for
+having that inline type.*
+
+A compile-time error occurs if an inline class constructor includes a
+superinitializer. *That is, a term of the form `super(...)` or
+`super.id(...)` as the last element of the initializer list.*
```
lrhn, line 663: Could we allow it? If I *want* to run the constructor of the super-class? 
It must necessarily pass the representation object to the super-inline-class, which is hard to specify, so maybe it's better to just make you call the super-constructor in the body.

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,948 @@
...
+
+All in all, an inline class allows us to replace the interface of a given
+representation object and specify how to implement the new interface
+in terms of the interface of the representation object.
+
+This is something that we could obviously do with a regular class used
+as a wrapper, but when it is done with an inline class there is no
+wrapper object, and hence there is no run-time performance cost. In
+particular, in the case where we have an inline type `V` with
+representation type `R` we may be able to refer to a `List<R>` using
```
stereotype441, line 93: "may be able" sounds like it implies that we sometimes *won't* be able to refer to a `List<R>` using the type `List<V>`.  Is that true?  What are the circumstances in which we can't?

accepted/future-releases/inline-classes/feature-specification.md
stereotype441: Consider rewriting to something like "The declaration of `v` must have the modifier `final` and must not have the modifier `late`; otherwise a compile-time error occurs."

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,948 @@
...
+
+*There are no special rules for static members in inline classes. They
+can be declared and called or torn off as usual, e.g.,
+`Inline.myStaticMethod(42)`.*
+
+
+## Primitives
+
+This document needs to refer to explicit inline method invocations, so we
+will add a special primitive, `invokeInlineMethod`, to denote invocations of
```
stereotype441, line 403: I've commented on this in previous PRs so I'm going to try to keep this brief, but I still think it's a bad idea to define the static analysis in terms of desugaring to a notation that's invalid Dart.  The actual implementation of the static analysis (at least in the analyzer) will not use this desugaring, so although we might gain extra precision in the specification by specifying it in this way, we pay for that precision by complicating the job of the implementers, who will have to mentally translate every line of the text that follows to figure out what it means in terms of the code the user can actually write.  IMHO it would be better for the spec to be written in terms of the code the user can actually write, because it would reduce the risk of implementation mistakes.

accepted/future-releases/inline-classes/feature-specification.md
stereotype441: Considering that we're going with `implements` rather than `extends` for composing inline classes, I think we are doing the right thing by not allowing `super(...)` or `super.id(...)`.

stereotype441 (approved):

accepted/future-releases/inline-classes/feature-specification.md
mit-mit: I believe the CFE team has given similar feedback in the past.

accepted/future-releases/inline-classes/feature-specification.md
lrhn: This. I dislike the tabloid headline style of "Disaster Will Happen ... unless very likely caveat".

Say what must be true first. Then say what happens if it isn't. If you only read the first part, your code will still be fine.

accepted/future-releases/inline-classes/feature-specification.md
eernstg: I would recommend that we make it an error now. A future enhancement could include the ability to run these constructors using special syntax. It is of course already possible to simply invoke them (and, probably, ignore the value of the expression).

accepted/future-releases/inline-classes/feature-specification.md
eernstg: The cycle error is associated with an inline type, not the inline class declaration. So we're considering `C<S>` where `S` is a type in the current scope (where `C<S>` occurs).

The badness that we're avoiding here is that a given type involving an inline class declaration, say, `C<S>` for a given `S`, does not have a representation type that we can determine (so we don't know how to find the run-time value of the inline type because we're in an infinite loop). So we can't generate code for `e is C<S>`, etc.

`C<int>` does not have a cycle error, and that matches the expectation that it is represented as `int` at run time.

In the body of `C`, `C<T>` has instantiated representation type `T`. For every possible invocation of a member of `C`, `T` will have a particular run-time value which is not an inline type (because we're in the middle of ensuring that no call site can violate that).

Let's create a cycle:

```dart
inline class C {
  final D d;
  C(this.d);
}

inline class D {
  final C c;
  D(this.c);
}
```

With this one we wouldn't know how to compile `e is C`, but that's also a cycle error, so we're still safe.

I added a paragraph saying that we must check the type introduced by an inline class declaration, even in the case where it is not used (such that we will report an error for `C` and `D` independently of expressions like `e is C`).

eernstg (commented):
Review response.

accepted/future-releases/inline-classes/feature-specification.md
eernstg: > This. I dislike...

Ah, I did not respond to this comment. (It was not shown when I responded to the review comments yesterday).

We do have an impedance mismatch here, though.

I recognize that it may read quite well to say "Do this and your program is fine" and then something like "Otherwise, an error occurs".

However, that style is somewhat counterproductive because it requires a specification of "good" programs, and we otherwise never do that (at least in the language specification). We specify the conditions that give rise to a compile-time error, and we specify the properties and behavior of specific shapes of program elements like expressions, statements, and declarations (and this behavior may include some dynamic errors). It is always understood that the latter is only applicable to programs with no compile-time errors.

The point is that the set of programs with no compile-time errors is determined by taking all syntactically correct programs and removing some of them (because they have non-syntax, compile-time errors). The remaining set is complex, and it consists of contributions from many different parts of the specification (e.g., for expressions we have many different sections, and each of them contribute a set of correct expressions with a specific syntactic form).

One thing which would be possible, and which wouldn't trigger this "good programs" problem, would be to always specify the error conditions positively (for instance, we wouldn't use 'unless').

Did that.

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,969 @@
...
+without code duplication.*
+
+*The reason why this mechanism uses the keyword `implements` rather
+than `extends` to declare a relation that involves "inheritance" is
+that it has a similar semantics as that of extension members (in that
+they are statically resolved).*
+
+Assume that _DV_ is an inline class declaration named `Inline`, and
+`V1` occurs as one of the `<type>`s in the `<interfaces>` of _DV_. In
+this case we say that `V1` is a _superinterface_ of _DV_.
```
srawlins, line 728: I don't think I see specified: what is the behavior when no `implements` clause is supplied? Is `Object` the singular superinterface?

I'm also asking whether an inline class without an `implements` clause has members like `runtimeType` and `operator ==` and `toString`.

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,969 @@
...
+    qux(); // Prints 'qux'.
+  }
+}
+```
+
+*That is, when the static type of an expression is an inline type `V`
+with representation type `R`, each method invocation on that
+expression will invoke an instance method declared by `V` or inherited
+from a superinterface (or it could be an extension method with on-type
+`V`).  Similarly for other member accesses.*
```
leafpetersen, line 607: Is the behavior of Object members called out anywhere?

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,969 @@
...
+the inline class declaration with a fixed lookahead.*
+
+A few errors can be detected immediately from the syntax:
+
+A compile-time error occurs if the inline class does not declare any
+instance variables, and if it declares two or more instance
+variables. Let `v` be the name of unique instance variable that it
+declares. The declaration of `v` must have the modifier `final`, and it
+can not have the modifier `late`; otherwise a compile-time error
+occurs.
```
leafpetersen, line 395: May it have an initializer?

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,969 @@
...
+<code>Inline&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;.name(...)</code>
+is used to invoke these constructors, and the type of such an expression is
+<code>Inline&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;</code>.
+
+*In short, inline class constructors appear to be very similar to
+constructors in regular classes, and they correspond to the situation
+where the enclosing class has a single, non-late, final instance
+variable, which is initialized according to the normal rules for
+constructors (in particular, it must occur by means of `this.id` or in
+an initializer list).*
```
leafpetersen, line 703: Should we allow it to be initialized via an explicit initializer instead?

```
inline class MyList {
    final List<int> l = [];
    MyList();
}
```

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,969 @@
...
+
+
```
leafpetersen, line 705: Is there a missing error around declaring the Object methods?

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,969 @@
...
+among distinct "inherited" members is an error, but it can be
+eliminated by redeclaring the clashing name.*
+
+The effect of having an inline class declaration _DV_ with
+superinterfaces `V1, .. Vk` is that the members declared by _DV_ as
+well as all members of `V1, .. Vk` that are not redeclared by a
+declaration in _DV_ can be invoked on a receiver of the type
+introduced by _DV_.
+
+In the body of _DV_, a superinvocation syntax similar to an explicit
```
leafpetersen, line 832: I think I'm somewhat opposed to adding this. It feels pretty superfluous - you can always access a method on a super-interface `V` by doing `V v = this; v.method(...)`, or `(this as V).method(...)` or `V(r).method()` where `r` is the representation field.

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,969 @@
...
+
+### Summary of Typing Relationships
+
+*Here is an overview of the subtype relationships of an inline type
+`V0` with representation type `R` and superinterfaces `V1 .. Vk`, as
+well as other typing relationships involving `V0`:*
+
+- *`V0` is a subtype of `Object?`.*
+- *`V0` is a supertype of `Never`.*
+- *If `R` is a top type then `V0` is a top type,
```
leafpetersen, line 916: I don't follow this.  Why would `V0` be a top type here?  I would not expect that given:

```dart
inline class V0 {
  final Object? x;
  V0(this.x);
}
```

that `String` would be a subtype of `V0`.

accepted/future-releases/inline-classes/feature-specification.md
eernstg: Very good. I'll conclude that it is an error to have a superinitializer and resolve this thread.

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,969 @@
...
+This document needs to refer to explicit inline method invocations, so we
+will add a special primitive, `invokeInlineMethod`, to denote invocations of
+inline methods.
+
+`invokeInlineMethod` is used as a specification device and it cannot occur in
+Dart source code. (*As a reminder of this fact, it uses syntax which is not
+derivable in the Dart grammar.*)
+
+
+### Static Analysis of invokeInlineMethod
```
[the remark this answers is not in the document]
lrhn: We may want to do that by introducing a generalization of extension methods and inline class methods, say "statically dispatched instance methods" (methods with a `this`, as opposed to, necessarily statically dispatched, static methods). The resolution of extension methods and inline class methods differ, but the invocation and tear-off behavior is the same (closes over actual receiver which `this` is bound to, at some type, plus type arguments provided at the invocation.)
Rather than saying that inline class methods work like extension methods, define how they both work, and say that both work like that, not like each other.

accepted/future-releases/inline-classes/feature-specification.md
lrhn: Asked Erik about that in person. Answer was "yes", because it's not a problem to do so. Field initializers are evaluated as part of constructor invocation (the `new`/`const` operators).
The initializer can be non-trivial/const, like `= DateTime.now()` or `++_counter`, so we don't necessarily get the same value each time. Doing `final field = expr;` is really not significantly different than `final field; Foo() : field = expr;`, which we will allow.

Constructors can also be `const`, there is also no problem with that. 

(In practice I'll probably just have something like `const Foo._(this._x);` as a constructor on all my inline classes, and forward all other constructors to that, but you can choose other approaches, just as you can for classes.)

accepted/future-releases/inline-classes/feature-specification.md
lrhn: We should. This text is overly restrictive.

The final field must be initialized no matter which non-redirecting generative constructor is invoked, but that can happen through an initializer expression on the field declaration, an initializing formal, or an initializer list entry.

accepted/future-releases/inline-classes/feature-specification.md
lrhn: I like it.
We present the `inline class` as similar to a class, and it inherits implementation from super-views.

The obvious way to write a method in a sub-inline-class which does something more than the inherited method, is to use `before...;super.method(); after...`. If we can make it work for the common cases, I think our uses will appreciate it.

If there is a conflict on the member among the super-inline-classes (which can be the reason you wrote an overriding method to begin with, to resolve that conflict by forwarding to one of the super-classes), then you can't just use `super.method()`. That's sad, but unavoidable for now.
I don't want to introduce a way to choose a _specific_ super (at least not just for this, we can also use such a feature it in classes with multiple mixins), so in that case you will have to up-cast/inject into a super-interface before calling.

accepted/future-releases/inline-classes/feature-specification.md
lrhn: Agree. Being a subtype of `Object?` does not make it a *supertype* of `Object?`.

This text may be a hold-over from when the views specification made the representation type a subtype of the view type?

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,968 @@
...
+
+A compile-time error occurs unless the inline class declares exactly
+one instance variable. Let `v` be the name of said instance
+variable. The declaration of `v` must have the modifier `final`
+and it must not have the modifier `late`; otherwise a compile-time
+error occurs.
+
+The _name of the representation_ in an inline class declaration is the
+name `id` of the unique final instance variable that it declares, and
+the _type of the representation_ is the declared type of `id`.
```
johnniwinther, line 399: What about untyped instance fields? Do we disallow them or just type them as dynamic? Do we disallow instance fields with initializers?

lrhn: There are no untyped instance variables in Dart, at least not after type inference. So yes, we likely type them as `dynamic` or `Object?`, like we would any similar instance variable declaration in a normal `class`.

No need to disallow fields with initializers (and type inference for the variable should work as normal for that too).
Generally it should behave just like the similar `class`, except that you must have precisely one final instance variable, and can't declare the class abstract, and therefore cannot have abstract members.

johnniwinther: Having to infer the types of the field means that the representation type of the inline class will not be available until very late in the compilation process. This will give a complex dependency between inline class declarations and top level inference.

If the field doesn't have an initializer we can short-cut this by saying that the representation type is `dynamic`, which we would have otherwise inferred. For fields with initializers there is no shortcut and it could even lead to cyclic dependencies:
```
inline class A {
  final it = field;
}
var field = A().it;
```
So to me the question is, do we have a good use case for fields _with_ initializers that justifies the added complexity?

accepted/future-releases/inline-classes/feature-specification.md
eernstg: This has been made explicit in the updated explanation about inline method invocations. The members of `Object` can be invoked on any expression whose static type is an inline class type.

(The intention is that this can be generalized to allow an inline class to provide access to any superinterface of the representation type. I'm keeping 'superinterfaces' of the inline class separate because they are all inline classes, and we do not wish to imply that they carry constraints like 'correct override'. So the relationship to `Object` is different than the relationship between two inline classes.)

I added an extra paragraph here specifying that the set of superinterfaces is empty when there is no `implements` clause.

The result is that an expression whose static type is an inline class _does_ admit invocations of `toString()` etc.

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,969 @@
...
+
+A reference to the representation typed by its run-time type or a
+supertype thereof (that is, typed by a "normal" type for the
+representation) is available as a declared name: The inline class must
+have exactly one instance variable whose type is the representation
+type, and it must be `final`.
+
+The representation type of the inline class (when the unique field of
+the inline class is `final int i;`, that's `int`) is similar to the
+on-type of an extension declaration.
```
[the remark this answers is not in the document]
eernstg: Done.

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,969 @@
...
+}
+
+void main() {
+  var tiny = TinyJson(<dynamic>[<dynamic>[1, 2], 3, <dynamic>[]]);
+  print(tiny.leaves);
+  tiny.add("Hello!"); // Error.
+}
+```
+
+Note that `it` is subject to promotion in the above example. This is safe
```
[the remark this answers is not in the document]
stereotype441: I'm ok with this as written.  I suppose you could make an argument that for consistency with the rest of the field promotion work, we shouldn't allow promotion of the representation "field" if it has a public name.  But I'd prefer to go ahead and make representation fields always promotable, because there's no soundness problem AFAICT, and it's easy to do implementation-wise.

accepted/future-releases/inline-classes/feature-specification.md
leafpetersen: I can see either saying "there can be no initializer", or "there must be an explicit type on the representation field".  I might lean a bit towards the latter.  Just say that we don't do field inference for inline class fields.

accepted/future-releases/inline-classes/feature-specification.md
leafpetersen: I'm fine with allowing `super.foo`, but I'm reluctant to add the new `super.I.foo` syntax, so it sounds like we're on the same page here.

accepted/future-releases/inline-classes/feature-specification.md
eernstg: News: Leaf and I discussed the treatment of `Object` interface names (`toString` etc.) and agreed that it's a safer bet to make them errors for now: With this change, an inline class can _not_ declare a member named `toString`.

The point is that it is very unlikely that an inline member shadowing an `Object` member will work as intended. If operator `==` is shadowed then sets and maps will still call the instance member; shadowing `noSuchMethod` is completely useless because basically (or strictly?) every single, meaningful invocation of `noSuchMethod` is done by code that doesn't know the static type of the receiver; and so on. Even with `toString`, where a custom version in an inline class might be useful, it's not likely to work as intended. Just consider the fact that _every_ string interpolation would have to use `'${x.toString()}'` rather than `'$x'`, and we would have to ensure that the static type of `x` is actually the given inline class in order to get the inline method when calling `x.toString()`.

accepted/future-releases/inline-classes/feature-specification.md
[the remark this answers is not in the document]
eernstg: We could rely on the fact that inline member resolution is static. The reason why promotion cannot soundly be applied to getter invocations that resolve statically to a final instance variable is that it could be overridden by an unstable getter. Not so here. We could immediately do the same thing for a final instance variable in a final class.

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,969 @@
...
+
+A compile-time error occurs if the inline class does not declare any
+instance variables, and if it declares two or more instance
+variables. Let `v` be the name of unique instance variable that it
+declares. The declaration of `v` must have the modifier `final`, and it
+can not have the modifier `late`; otherwise a compile-time error
+occurs.
+
+The _name of the representation_ in an inline class declaration is the
+name `id` of the unique final instance variable that it declares, and
```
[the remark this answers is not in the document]
eernstg: Of course, done!

accepted/future-releases/inline-classes/feature-specification.md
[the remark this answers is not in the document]
eernstg: > This doesn't cover tear offs, similarly for the dynamic semantics below.

Tear offs are covered in the new descriptions where `invokeInlineMethod` has been replaced by a natural language phrase.

> Perhaps it would be simpler to explain these directly

Did that (in terms of their semantics, no references to extension methods).

> define how they both work

This is the approach which is very nearly used already. This document specifies the semantics of tear offs of inline class methods directly, the language specification specifies the semantics of tear offs of extension methods directly, and then we just need to merge them when we integrate inline classes into the language specification.

I'll resolve this thread: I think there's no doubt that this issue has been resolved.

accepted/future-releases/inline-classes/feature-specification.md
eernstg: Very good! I'll conclude that there is no need to prohibit these initializers. I haven't added further text about this: The inline class static analysis already relies on the analysis of regular classes in many ways (e.g., how to type check the initializer list, the fact that a constructor can be `const` and so on).

accepted/future-releases/inline-classes/feature-specification.md
eernstg: Clarified!

accepted/future-releases/inline-classes/feature-specification.md
eernstg: The previous proposals actually allowed shadowing `Object` methods. I've added a rule to make it an error, as agreed yesterday.

accepted/future-releases/inline-classes/feature-specification.md
eernstg: It's the usual trade-off: If we support `super.I.foo` then we have a reasonably meaningful way to write forwarding methods that developers may need in order to resolve name clashes. If we do not support that syntax then we're encouraging developers to write code where type casts occur (and conflict resolution like this might not be so uncommon).

```dart
inline class V2 {
  final Object id;
  V2(this.id);
  void foo() { print('V2.foo()'); }
  void bar() { print('V2.bar()'); }
}

inline class V3 {
  final Object id;
  V3(this.id);
  void foo() { print('V3.foo()'); }
  void baz() { print('V3.baz()'); }
}

inline class V1 implements V2, V3 {
  final Object id;
  V1(this.id);
  void qux() {
    super.bar(); // 'V2.bar()'.
    super.baz(); // 'V3.baz()'.
    // super.foo(); // Compile-time error, do this instead:
    (this as V2).foo();
  }
}
```

Are you happy about telling developers to write `as`? Or do you have some other approach in mind?

accepted/future-releases/inline-classes/feature-specification.md
eernstg: Exactly, that bit came from earlier versions. Removed it.

accepted/future-releases/inline-classes/feature-specification.md
lrhn: I'd like `I.super.foo` as a language feature. It makes sense for classes with multiple mixins.
Which means that when we add it, it should work there too, which means it's probably out-of-scope/feature-creep for the inline classes feature. 

(I'd recommend `V2(id).foo()` instead of using `as`, because `as` is unsafe - it doens't check that you're doing an upcast or valid cast at all. It also makes it very clear that the call will not have access to any members of `V1`.)

accepted/future-releases/inline-classes/feature-specification.md
eernstg: > Just say that we don't do field inference for inline class fields.

Sounds good, done!

eernstg (commented):
Thanks for a lot of useful input! I believe the main remaining issue is the syntax `super.V2.foo()`, used to specify a "superinvocation" of `foo` in the superinterface (and inline class) `V2`. This would be needed, in particular, when writing a forwarding member in order to resolve a name clash:

```dart
inline class V1 { void foo() {}}
inline class V2 { void foo() {}}
inline class V3 implements V1, V2 { void foo() => super.V2.foo(); }
```

If we drop that syntax (and do not replace it by something else), I think the only easily expressible way to do the same thing is a cast. Any cast is unsafe, and it is harder to understand the purpose and effect of the code with the cast.

```dart
inline class V3 implements V1, V2 { void foo() => (this as V2).foo(); }}
```

leafpetersen:
> If we drop that syntax (and do not replace it by something else), I think the only easily expressible way to do the same thing is a cast. Any cast is unsafe, and it is harder to understand the purpose and effect of the code with the cast.

I don't really think this is true.  Assigning it to a variable of the super-interface type before calling the method works fine and is perfectly safe.

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,969 @@
...
+A compile-time error occurs if an inline type is used as a
+superinterface of a class or a mixin, or if an inline type is used to
+derive a mixin.
+
+*In other words, an inline type cannot occur as a superinterface in an
+`extends`, `with`, `implements`, or `on` clause of a class or mixin.
+On the other hand, it can occur in other ways, e.g., as a type
+argument of a superinterface of a class etc.*
+
+If `e` is an expression whose static type `V` is the inline type
```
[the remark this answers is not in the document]
eernstg: I'd expect extension methods whose on-type matches `V` to have lower priority than members that `V` has, just like extension methods with on-type `int` have lower priority than instance members of `int`. Added a paragraph to clarify that.

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,969 @@
...
+  final T id;
+  V(this.id);
+
+  ... // Other members.
+}
+```
+
+It is then allowed to use
+<code>V&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;</code>
+as a type.
```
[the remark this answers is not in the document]
eernstg: Added a paragraph about inline types as type literals to the static analysis section, and to the dynamic semantics section.

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,969 @@
...
+type `R`.
+
+We say that the _declared representation type_ of `Inline`
+is `R`, and the _instantiated representation type_ corresponding to
+<code>Inline&lt;T<sub>1</sub>,.. T<sub>s</sub>&gt;</code> is
+<code>[T<sub>1</sub>/X<sub>1</sub>, .. T<sub>s</sub>/X<sub>s</sub>]R</code>.
+
+We will omit 'declared' and 'instantiated' from the phrase when it is
+clear from the context whether we are talking about the inline class
+itself, or we're talking a particular instantiation of a generic
```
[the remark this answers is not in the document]
eernstg: Done.

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,969 @@
...
+Let `V` be an inline type of the form
+<code>Inline&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;</code>,
+and let `R` be the corresponding instantiated representation type.
+`V` is a proper subtype of `Object?`. If `R` is non-nullable then `V`
+is a proper subtype of `Object` as well.
+
+*That is, an expression of an inline type can be assigned to a top type
+(like all other expressions), and if the representation type is
+non-nullable then it can also be assigned to `Object`. Non-inline types
+(except bottom types) cannot be assigned to inline types without a cast.*
```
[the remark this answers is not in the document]
eernstg: Of course, done!

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,969 @@
...
+can also be used to verify that an existing object (provided as an
+actual argument to the constructor) satisfies the requirements for
+having that inline type.*
+
+A compile-time error occurs if an inline class constructor includes a
+superinitializer. *That is, a term of the form `super(...)` or
+`super.id(...)` as the last element of the initializer list.*
+
+*In the body of a generative inline class constructor, the static type
+of `this` is the same as it is in any instance member of the inline,
```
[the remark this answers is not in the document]
eernstg: Done

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,969 @@
...
+*The purpose of having an inline class constructor is that it bundles
+an approach for building an instance of the representation type of an
+inline declaration _DV_ with _DV_ itself, which makes it easy to
+recognize that this is a way to obtain a value of that inline type. It
+can also be used to verify that an existing object (provided as an
+actual argument to the constructor) satisfies the requirements for
+having that inline type.*
+
+A compile-time error occurs if an inline class constructor includes a
+superinitializer. *That is, a term of the form `super(...)` or
```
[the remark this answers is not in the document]
eernstg: Yes, done.

accepted/future-releases/inline-classes/feature-specification.md
eernstg: Yes.

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,969 @@
...
+`R`, and `this` has type `V`.*
+
+Again, let `V` be an inline type of the form
+<code>Inline&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;</code>,
+and let `R` be the corresponding instantiated representation type.
+If `R` is not an inline type then we say that `V` is an inline type
+at level zero. If `R` is an inline type at level _k_ then we say that
+`V` is an inline type at level _k + 1_.
+A compile-time error occurs if the level of `V` is undefined.
+
```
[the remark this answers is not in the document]
eernstg: Definitely finicky. But perhaps `C` and `Foo` are safe after all?

The erasure step is applied to an inline type, not to an inline class declaration. For any given occurrence of `C<T>` for some `T`, we would be able to check separately that `T` has an inline erasure, and `C<T>` would then have the same inline erasure (because `C` happens to use `T` as the representation type).

In the body of `C`, there is no need to inline erase `T`: At run time, `T` is available in the current scopes, and its value is already inline erased.

If we consider a raw `Foo` then it just means `Foo<C<dynamic>>`, and the inline erasure of that type argument is `dynamic`. If we consider `Foo<S>` for any `S`, then we'd again ("by the induction hypothesis") ensure separately that `S` has an inline erasure.

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,969 @@
...
+is undefined.
+
+*In other words, we check for this kind of cycles on every non-generic
+inline class declaration using the type directly (even in the case
+where nobody uses that type), and we check for this kind of cycles on
+the instantiation-to-bound of every generic inline class declaration
+(again, even when nobody uses it). Finally, we check for this kind of
+cycles with every parameterized type `V<T1..Tk>` that actually
+occurs in the code which is being analyzed/compiled.*
+
```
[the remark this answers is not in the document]
eernstg: Yes! Alternatively, we could handle variance of the type arguments of an inline class in the same way as we handle variance for a `typedef` with type arguments: Deriving it from the occurrences of the type parameters in the body of the `typedef` (respectively: in the representation type of the inline class).

For now, I've added a compile-time error for the non-covariant occurrences, which allows us to do the `typedef` thing as an enhancement (and it will be possible to do the same thing with declaration-site variance, anyway).

eernstg:
@leafpetersen wrote:
> Assigning it to a variable of the super-interface type before calling the method works fine and is perfectly safe.

Right, I was just considering expression forms (because it enables `=>`, which is probably considered desirable).

eernstg:
A couple of extra comments came up. In particular, the ability to do inline-erasure (computing a non-inline type that corresponds to any given inline type, by going repeatedly to the representation type) is finicky, and may need more scrutiny.

Another thing is the fact that the representation object may be null if the representation type is potentially nullable, but this doesn't mean that we can't call the inline class methods. So should the inline type be considered potentially nullable?

accepted/future-releases/inline-classes/feature-specification.md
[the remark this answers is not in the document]
leafpetersen: You can call extension methods on potentially nullable types, and these are extension methods, so I don't think there's any issue.  But in some sense I guess it's unnecessary: we can't say that these are non-nullable, and the definition of potentially nullable is "We say that a type T is potentially nullable if T is not non-nullable.", so whether we say it or not, these things are potentially nullable.

The main thing is to make it clear that these things are **not** non-nullable.

accepted/future-releases/inline-classes/feature-specification.md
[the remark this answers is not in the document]
leafpetersen: I don't think this flies in general.  It might work out for classes, but even there it makes me nervous.  But with generic methods, these things exist at runtime.  For example, `void Function<T extends C<T>>()`.

accepted/future-releases/inline-classes/feature-specification.md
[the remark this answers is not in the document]
eernstg: No objections were raised here. Resolving the thread.

accepted/future-releases/inline-classes/feature-specification.md
eernstg: Took the feature out. Resolving.

eernstg (commented):
Removed the `I.super.foo` syntax, adjusted other locations based on comments. PTAL.

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,1041 @@
...
+We need to introduce a concept that is similar to existing concepts
+for regular classes.
+
+We say that an inline class `V` _has_ a member named `n` in the case
+where `V` declares a member named `n`, and in the case where `V` has
+no such declaration, but `V` has a superinterface `Vs` that has a
+member named `n`. In both cases,
+_the member declaration named `n` that `V` has_ is said declaration.
+
+*This definition is unambiguous for an inline class that has no
```
leafpetersen, line 443: Mmm, not entirely right?  If `A` and `B` declare `foo`, `C` implements `A` and `B` and re-defines `foo`, and `D` implements `C`, then there are three such `Vs`.

accepted/future-releases/inline-classes/feature-specification.md
```diff
@@ -0,0 +1,1041 @@
...
+<code>Inline&lt;X<sub>1</sub> .. X<sub>s</sub>&gt;</code>.
+The static type of the representation name is the representation
+type.
+
+*For example, in `inline class V { final R id; ...}`, `id` has type
+`R`, and `this` has type `V`.*
+
+Let _DV_ be an inline class declaration named `V` with representation
+type `R`. Assuming that all types have been fully alias expanded, we
+say that _DV_ has a representation dependcy on an inline class
```
leafpetersen, line 752: dependency

accepted/future-releases/inline-classes/feature-specification.md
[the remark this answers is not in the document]
leafpetersen: Did this get handled somewhere that I missed?

leafpetersen (approved):

accepted/future-releases/inline-classes/feature-specification.md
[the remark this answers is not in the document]
eernstg: The current wording ensures that if `V` is an inline type with instantiated representation type `R` then:
- If `R` is non-nullable then `V` is non-nullable.
- If `R` is potentially nullable then `V` is potentially nullable.

Hence, inline types are potentially nullable in general, but in the case where the representation type is non-nullable, the inline type is non-nullable as well.

This is consistent with the null-safety specification: In the case where `V` is non-nullable according to this document, it satisfies the requirement in the null-safety specification (`V <: Object`). Conversely, in the case where `V` is potentially nullable, i.e., it is not non-nullable, it is not true that `V <: Object`.

It is also sound: In the case where `V` is non-nullable, the value of an expression of type `V` cannot be null.

accepted/future-releases/inline-classes/feature-specification.md
eernstg: Added the word 'direct' just before 'superinterface'.

accepted/future-releases/inline-classes/feature-specification.md
eernstg: Done!

accepted/future-releases/inline-classes/feature-specification.md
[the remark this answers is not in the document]
eernstg: No, but I did not understand it correctly the first time around. I think this is the single really serious issue that remains in this proposal.

Would it be overkill to require the following?:

Let `X1 extends T1, .. Xs extends Ts` be a type parameter list declaration. A compile-time error occurs if `X1 extends S1, .. Xs extends Ss` has any compile-time errors, where `Sj` is obtained by recursively replacing every occurrence of an inline type in `Tj` by the corresponding instantiated representation type.

In other words, for every type variable declaration, recursively perform an inline-erasure on every inline type in its bound, and then run the normal static analysis checks on it again.

I put this rule into the document.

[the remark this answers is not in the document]
eernstg: It might be unnecessary to check the inline-erased type parameter bounds of classes/mixins/typedefs/inline_classes, because instance creations are always statically resolved, and hence we can just check the specified bounds at compile time, and trust all type arguments to satisfy the bounds at run time.

However, this won't work for generic functions, because they can be invoked dynamically.

So it's probably not even worth the trouble to make class type parameters an exception.

Another perspective on this is that there are no (real) compile-time errors for type parameter bound declarations: It's an error to have a cycle like `X extends Y, Y extends X`, but we could also just allow it. Similarly for the union type bounds like `X extends FutureOr<X>`: We currently do allow them, but we might make them an error. As long as every actual type parameter satisfies the bounds, it doesn't really matter whether or not there is a cycle among several type parameters.

PR #2627, opened 2022-11-11
Tweak the switch expression syntax.

### LINKED ISSUE #2126: Using => in switch statements is confusing
https://github.com/dart-lang/language/blob/master/working/0546-patterns/patterns-feature-specification.md proposes this syntax:
```dart
Color shiftHue(Color color) {
  return switch (color) {
    case Color.red => Color.orange;
    case Color.orange => Color.yellow;
    case Color.yellow => Color.green;
    case Color.green => Color.blue;
    case Color.blue => Color.purple;
    case Color.purple => Color.red;
  };
}
```

```dart
bool xor(bool a, bool b) =>
    switch ((a, b)) {
      case (true, true) => false;
      case (true, false) => true;
      case (false, true) => true;
      case (false, false) => false;
    };
```

I've been trying to come to terms with this re-use of `=>` for a few days now and I am still finding it very confusing. I understand the logic path that leads to this design (`=>` is sort of "define the thing on the left as returning the expression on the right") but I can't help but read it as "define a function with the arguments on the left and the return value on the right". Seeing `(true, true) => false` not be a lambda is quite disorienting IMHO.

It seems to be to be somewhat unnecessary, too. `switch` could be defined as it is now, modulo removing the requirements for `break`, and allowed anywhere an expression is allowed, with the exception that if any of the cases have more than one statement, or if any of the cases evaluate to a value that isn't compatible in the expression where the `switch` is used, then the `switch` is invalid. (The error messages can be carefully designed to make this intuitive, IMHO.)

```dart
Color shiftHue(Color color) {
  return switch (color) {
    case Color.red: Color.orange;
    case Color.orange: Color.yellow;
    case Color.yellow: Color.green;
    case Color.green: Color.blue;
    case Color.blue: Color.purple;
    case Color.purple: Color.red;
  };
}
```

```dart
bool xor(bool a, bool b) =>
    switch ((a, b)) {
      case (true, true): false;
      case (true, false): true;
      case (false, true): true;
      case (false, false): false;
    };
```

munificent:
Get rid of "case" and "default" and use "," to separate.

Fix #2126.

### DIFF
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -145,7 +145,7 @@ matching constructs to make working with that style enjoyable:
 double calculateArea(Shape shape) =>
   switch (shape) {
     Square(length: var l) => l * l,
-    Circle(radius: var r) => math.pi * r * r,
+    Circle(radius: var r) => math.pi * r * r
   };
 ```
 
@@ -253,7 +253,7 @@ share a body:
 ```dart
 var isPrimary = switch (color) {
   Color.red | Color.yellow | Color.blue => true,
-  _ => false,
+  _ => false
 };
 ```
 
@@ -333,7 +333,7 @@ String asciiCharType(int char) {
     < space => 'control',
     == space => 'space',
     > space & < zero => 'punctuation',
-    >= zero & <= nine => 'digit',
+    >= zero & <= nine => 'digit'
     // Etc...
   }
 }
@@ -1117,7 +1117,7 @@ Color shiftHue(Color color) {
     Color.yellow => Color.green,
     Color.green => Color.blue,
     Color.blue => Color.purple,
-    Color.purple => Color.red,
+    Color.purple => Color.red
   };
 }
 ```
@@ -1156,13 +1156,13 @@ Making it high precedence allows useful patterns like:
 await switch (n) {
   1 => aFuture,
   2 => anotherFuture,
-  _ => otherwiseFuture,
+  _ => otherwiseFuture
 };
 
 var x = switch (n) {
   1 => obj,
   2 => another,
-  _ => otherwise,
+  _ => otherwise
 }.someMethod();
 ```
 
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -2,9 +2,9 @@
 
 Author: Bob Nystrom
 
-Status: In progress
+Status: Accepted
 
-Version 2.15 (see [CHANGELOG](#CHANGELOG) at end)
+Version 2.16 (see [CHANGELOG](#CHANGELOG) at end)
 
 Note: This proposal is broken into a couple of separate documents. See also
 [records][] and [exhaustiveness][].
@@ -144,12 +144,13 @@ matching constructs to make working with that style enjoyable:
 ```dart
 double calculateArea(Shape shape) =>
   switch (shape) {
-    case Square(length: var l) => l * l;
-    case Circle(radius: var r) => math.pi * r * r;
+    Square(length: var l) => l * l,
+    Circle(radius: var r) => math.pi * r * r,
   };
 ```
 
-As you can see, it also adds an expression form for `switch`.
+As you can see, it also adds an expression form for `switch` that doesn't
+require `case`.
 
 ## Patterns
 
@@ -251,8 +252,8 @@ share a body:
 
 ```dart
 var isPrimary = switch (color) {
-  case Color.red | Color.yellow | Color.blue => true;
-  default => false;
+  Color.red | Color.yellow | Color.blue => true,
+  _ => false,
 };
 ```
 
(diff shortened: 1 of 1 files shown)


accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -144,12 +144,13 @@ matching constructs to make working with that style enjoyable:
 ```dart
 double calculateArea(Shape shape) =>
   switch (shape) {
-    case Square(length: var l) => l * l;
-    case Circle(radius: var r) => math.pi * r * r;
+    Square(length: var l) => l * l,
+    Circle(radius: var r) => math.pi * r * r,
```
stereotype441, line 148: Nit: Usually in the spec when we're giving examples of lists, map literals, enums, or invocation arguments we don't include the optional final `,`.  I'd prefer for us to be consistent with that practice and leave off the optional final `,` from our examples here.

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -1162,17 +1174,17 @@ evidence this will be useful.
 
 Thanks to expression statements, a switch expression could appear in the same
 position as a switch statement. This isn't technically ambiguous, but requires
-unbounded lookahead to tell if a switch in statement position is a statement or
-expression.
+unbounded lookahead to read past the value expression to the first `case` in
```
stereotype441, line 1177: Technically this isn't a problem anymore, because the scanner tags every `(` with a pointer to the matching `)`, so at parse time we *could* check for the presence of `case` or `default` in a bounded amount of time.

It might still be reasonable to prohibit switch expressions at statement level just to avoid user confusion, though.

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -894,8 +895,8 @@ We extend switch statements to allow patterns in cases:
 ```
 switchStatement         ::= 'switch' '(' expression ')'
                             '{' switchStatementCase* switchStatementDefault? '}'
-switchStatementCase     ::= label* caseHead ':' statements
-caseHead                ::= 'case' pattern ( 'when' expression )?
+switchStatementCase     ::= label* 'case' guardedPattern ':' statements
+guardedPattern          ::= pattern ( 'when' expression )?
```
stereotype441, line 899: I like this rename.  👍

stereotype441 (approved):
lgtm with some minor nits

accepted/future-releases/0546-patterns/feature-specification.md
munificent: Done.

accepted/future-releases/0546-patterns/feature-specification.md
munificent: Our scanner might work fine, but there are other Dart scanners and parsers out there for things like syntax highlighters that may find it challenging to handle correctly.

munificent:
I feel pretty solid about this and don't want to leave the implementers hanging, so I'm going to go ahead and merge this in but I'm happy to discuss it more if other folks aren't sold yet.

leafpetersen (commented):
LGTM

DmitrySboychakov:
Might be better to call in "match" instead of "switch" now?

munificent:
We've discussed `match` several types. We can't make it a reserved word because that's a massively breaking change, in particular because there is a type in the core library named `Match`, which means `match` is a very common variable name.

Treating `match` as a contextual keyword *might* work, but that's always sort of dubious and risks running into ambiguity. It also makes it harder for tools like syntax highlighters which don't usually have the context to understand where an identifier should be treated like a keyword and where it shouldn't. (This is why many Dart syntax highlighters always highlight words like `show` and `hide` as if they were reserved words even in contexts where they are just used as a normal identifier.)

mraleph:
Just noticed this: this makes `if (x case pattern)` really confusing because we are getting rid of `case` in `switch` - this weakens the whole argument around the syntax. 

Any plans to shift to `if (var pattern = x)` instead?

lrhn:
I'm fairly strongly against the `if (var pattern = e)` syntax because it looks like an irrefutable pattern assignment, but is actually a refutable pattern. That means we have to define what `var refutablePattern` means, where we currently only have `var irrefutablePattern` and `case refutablePattern`. It's a third kind of pattern.

Should there be inference from the pattern to the expression, like there is in a `var p = e;` declaration? If so, we need to extend the pattern type schema algorithm to refutable patterns too, which it currently doesn't need to (and which has some annoying edge cases). If not, it's even further from what it looks like.

I'm not sold on this change at all. I'd prefer to keep `case` for expression switches too.
Even if not, I'd prefer to do the same thing for expression switches and element switches (which this change doesn't).

munificent:
> Just noticed this: this makes `if (x case pattern)` really confusing because we are getting rid of `case` in `switch` - this weakens the whole argument around the syntax.

I realized that too when I made this change. But `if (x case pattern)` is still the least bad syntax I've been able to come up with.

> Any plans to shift to `if (var pattern = x)` instead?

As Lasse says, this would be even *more* confusing, because the static semantics and type inference story between refutable and irrefutable patterns is quite different.

> I'm not sold on this change at all. I'd prefer to keep `case` for expression switches too.
> Even if not, I'd prefer to do the same thing for expression switches and element switches (which this change doesn't).

Oh no. I didn't realize you didn't like it. :(

I think it makes sense to switch elements to mirror switch statements, since the other elements mirror statement forms.

We could add `case` back for switch expressions, but I've tried the `case`-less syntax out in a bunch of examples and I have to say that I have grown to like it. If you haven't spent some time trying it out, it might be worth doing that to see if you warm up to it. I think there's real value in having an expression form aim for more brevity than a statement form does. (By analogy, consider how much more terse `?:` is compared to `if () else `.)

PR #2628, opened 2022-11-12
[patterns] Specify if-case elements.

### LINKED ISSUE #2542: If-case collection elements
The patterns proposal defines a new if-case statement:

```dart
if (json case [int x, int y]) {
  print('Was coordinate array $x,$y');
} else {
  throw FormatException('Invalid JSON.');
}
```

There is a TODO in the proposal to extend that to support if-case collection elements:

```dart
var list = [
  if (json case [int x, int y])
    Point(x, y)
  else
    'Invalid'
];
```

This issue tracks that.

munificent:
The syntax and semantics are obvious and I think at least some of the implementers were assuming we would do this, but it's not actually in the spec. Now it is.

Fix #2542.

### DIFF
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -2,9 +2,9 @@
 
 Author: Bob Nystrom
 
-Status: In progress
+Status: Accepted
 
-Version 2.15 (see [CHANGELOG](#CHANGELOG) at end)
+Version 2.16 (see [CHANGELOG](#CHANGELOG) at end)
 
 Note: This proposal is broken into a couple of separate documents. See also
 [records][] and [exhaustiveness][].
@@ -1187,7 +1187,7 @@ there will be an actual ambiguity. In that case, reword the above section.**
 
 [2126]: https://github.com/dart-lang/language/issues/2126
 
-### If-case statement
+### If-case statement and element
 
 Often you want to conditionally match and destructure some data, but you only
 want to test a value against a single pattern. A `switch` statement works but is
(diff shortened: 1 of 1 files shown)


accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -1243,6 +1243,21 @@ if (json case [int x, int y] when x == y) {
...
+if-case elements. We replace the existing `ifElement` rule with:
+
+```
+ifElement ::= 'if' '(' expression caseHead? ')' element ('else' element)?
+```
+
+The semantics follow the statement form. If there is no `caseHead`, then it
+behaves as before. When there is a `caseHead`, if the `expression` matches the
+pattern (and the guard returns `true`) then we evaluate and yield the then
+element into the surrounding collection. Otherwies, we evaluate and yield the
```
bwilkerson, line 1258: nit: 'Otherwies' --> 'Otherwise'

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -2414,6 +2432,32 @@ Where `<keyword>` is `var` or `final` is treated like so:
...
+        1.  If the guard evaluates to `true`, evaluate the then `element` and
+            yield the result into the collection.
+
+        2.  Else, evaluate the else `element` if there is one and yield the
+            result into the collection.
+
+    2.  Else there is no guard clause. Evaluate the then `element` and yield the
+        result into the collection.
+
+4.  Else the match failed. Else, evaluate the else `element` if there is one and
```
bwilkerson, line 2458: nit: There might be a good reason for the current wording, but the repeated "Else" sounds odd.

accepted/future-releases/0546-patterns/feature-specification.md
munificent: Thanks, done.

accepted/future-releases/0546-patterns/feature-specification.md
munificent: There is no good reason except for a copy/paste error.

leafpetersen (approved):
LGTM

PR #2639, opened 2022-11-16
[Patterns] remove irrelevant specification of pattern type schemas.

stereotype441:
The pattern type schemas for logical-or, null-check, constant, and relational patterns were not meaningful because pattern type schemas are only used in irrefutable contexts, and these patterns are not allowed to appear in irrefutable contexts.

In particular, the definition of pattern type schema for constant patterns was causing confusion because it appeared to cause a circularity: it defined the pattern type schema for a constant pattern to be the inferred type of the constant; however text elsewhere specifies that the type of the constant is inferred based on the type of the scrutinee, and the pattern type schema is used to infer the type of the scrutinee.  However this wasn't a real problem, because constant patterns are illegal in precisely the circumstances in whch pattern type schemas are used.

### DIFF
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -1570,21 +1570,15 @@ var (a, int b) = ... // Schema is `(?, int)`.
 
 The context type schema for a pattern `p` is:
 
-*   **Logical-or**: The least upper bound of the context type schemas of the
-    branches.
-
 *   **Logical-and**: The greatest lower bound of the context type schemas of the
     branches.
 
-*   **Null-check** or **null-assert**: A context type schema `E?` where `E` is
-    the context type schema of the inner pattern. *For example:*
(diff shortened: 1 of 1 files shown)


accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -1675,6 +1669,11 @@ The context type schema for a pattern `p` is:
     //                  ^-- Infer Foo<num>.
     ```
 
+The pattern type schema for logical-or, null-check, constant, and relational
+patterns is not defined, because logical-or patterns are only allowed in
```
lrhn, line 1673: "because logical-or patterns" -> "those patterns"

lrhn (approved):

accepted/future-releases/0546-patterns/feature-specification.md
stereotype441: Fixed, thanks!

munificent (commented):
Thanks!

Issue #2595, opened 2022-10-28
[Access Modifiers] Access modifiers using closed, sealed, open and interface


leafpetersen:
In #2592 @munificent proposes a set of modifiers for classes to handle exhaustiveness, as well as control over whether a class can be used as an interface or as a superclass.  This issue proposes a slightly different approach to achieve the same essential outcomes (with some differences in what is expressible).

The idea is to add two negative capabilities (`sealed` and `closed`), and two positive capabilities (`open`, and `interface`) which can be combined to achieve the results we want.

In all cases, restrictions only apply outside of the defining library (as with #2592).  

I ignore `abstract` here, since I believe it is orthogonal.

## Proposal

A class which is marked as `sealed` cannot be extended or implemented outside of the defining library.  All subclasses of a class which is marked as `sealed` are also considered implicitly sealed, but they may override this.  Sealed classes are not considered for exhaustiveness checking.
```dart
sealed class A {} // May not be extended or implemented outside of the current library
class B extends A {} // Likewise
class C implements A {} // no implicit restrictions
```

A class which is marked as `closed` cannot be extended or implemented outside of the defining library, and is considered implicitly abstract.  Subclasses of a class which is marked as `closed` are not considered implicitly sealed (they inherit no negative capabilities).  Closed classes are considered for exhaustiveness checking.
```dart
closed class A {} // May not be extended or implemented outside of the current library
class B extends A {} // May be extended or implemented outside of the current library
sealed class C extends A {} // May not be extended or implemented outside of the current library 
```

A class which is marked as `open` may be subclassed, and may not be implemented unless it is explicitly marked with `interface`.  All subclasses of a class marked as `open` are implicitly `open`.
```dart
open class A {} // May be extended outside of the library, but not implemented outside of the current library
class B extends A {} // Likewise
sealed class C {} // 
open class D extends C {} // May be extended outside of the library, but not implemented
open interface class E extends C {} // May be extended or implemented outside of the library
```

A class which is marked as `interface` may be implemented, and may not be extended unless it is explicitly marked with `open`.  All subclasses of a class marked with `interface` are implicitly marked with `interface`.  A class may not be marked with `interface` unless either its superclass is in the same library or its superclass also has the `interface` capability (implicitly or explicitly).  *We wish to prevent classes outside of a library from adding back in the interface capability.  However, a class may be marked `interface` to remove the  subclass capability that it inherits from its superclass*.
```dart
[...]

PR #2646, opened 2022-11-22
Proposal for "base", "interface" and "final" type modifiers.

munificent:
Here's a new proposal for a set of capability modifiers on types. It supersedes my earlier "closed" and "base" proposal (#2592), which I'll close. It's heavily based on @leafpetersen's proposal (#2595), incorporates a suggestion from @natebosch to use a metadata annotation for `@reopen` and Erik's discussion of invariants you get from removing capabilities (which I hope I captured OK here).

Let me know what you think!

@mit-mit @stereotype441 @jakemac53 @kallentu

### DIFF
--- a/working/base-interface-final/feature-specification.md
+++ b/working/base-interface-final/feature-specification.md
@@ -11,7 +11,8 @@ allow an author to control whether the type allows being implemented, extended,
 both, or neither. The proposed modifiers are:
 
 *   No modifier: As today, the type has no restrictions.
-*   `base`: The type can be extended or mixed in but not implemented.
+*   `base`: The type can be extended (if a class) or mixed in (if a mixin) but
+    not implemented.
 *   `interface`: The type can be implemented but not extended or mixed in.
 *   `final`: The type can't be extended, mixed in, or implemented.
 
@@ -55,8 +56,8 @@ is actually defined and will succeed. It helps you in case you forget to
 implement something.
 
 But it also means that if a *new* member is added to a class then every single
-class implementing that class's interface now has a new compile error since none
-of them have that member.
+class implementing that class's interface now has a new compile-time error since
+none of them have that member.
 
 This makes it hard to add new members to existing public types in packages.
 Since anyone could be implementing that type's interface, any new member is
@@ -84,7 +85,9 @@ class Account {
   bool canWithdraw(int amount) => amount <= _balance;
 
   bool tryWithdraw(int amount) {
-    if (amount <= 0) throw ArgumentError("Amount must be positive.");
+    if (amount <= 0) {
+      throw ArgumentError.value(amount, "amount", "Must be positive");
+    }
 
     if (!canWithdraw(amount)) return false;
     _balance -= amount;
@@ -236,11 +239,11 @@ programming style][fp] in Dart.
 
 [fp]: https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#algebraic-datatypes
 
-The `sealed` modifier prevents subtyping from outside of the library where the
-sealed type is defined. But it doesn't prevent you from subtyping within the
-same library. In fact, the *whole point* of `sealed` is to define subtypes
-within the same library so that you can pattern match on those to cover the
-supertype.
+The `sealed` modifier prevents direct subtyping from outside of the library
+where the sealed type is defined. But it doesn't prevent you from subtyping
+within the same library. In fact, the *whole point* of `sealed` is to define
+subtypes within the same library so that you can pattern match on those to cover
+the supertype.
 
 ### Extending non-extensible classes in the same library
 
@@ -268,7 +271,7 @@ can define `Square` and `Circle`, but disallow others from doing so.
 A key invariant you get by preventing a type from being implemented is that it
 becomes safe to access private members defined on that type without risking a
 runtime exception. You are ensured that any instance of the type is an instance
-of a type from your library that inclines all of its private members.
+of a type from your library that includes all of its private members.
 
 This invariant is still preserved if we allow you to implement the type from
 within the same library. When you implement a type inside its library, the
@@ -321,7 +324,45 @@ does `MySubclass` now expose externally? We have a few options:
     its own public one. That subclass is now externally extensible and the
     language quietly lets you do that.
 
-*   **Trust but verify.** In the above example, it's not clear what the author
+*   **Disallow removing restrictions.** We could say that you can ignore a
+    type's restrictions within the same library, but any types that do that
+    *must* have the same restrictions as the type they extend or implement. So
+    if you implement a class marked `base` in the same library, that
+    implementing class must also be marked `base` or `final`.
+
+    This avoids any confusion about whether a subtype removes a restriction. But
+    it comes at the expense of flexiblity. If a user *wants* to remove a
+    restriction, they have no ability to.
+
+    This would constrast with `sealed` where you can have subtypes of a sealed
+    type that are not themselves sealed. This is a deliberate choice because
+    there's no *need* for the direct subtypes of a sealed to be sealed in order
+    for exhaustiveness checking to be sound. Since exhaustiveness is the goal
+    and Dart is permissive by default, we allow subtypes of sealed types to be
+    unsealed.
+
+    It also prevents API designs that seem reasonable and useful to me. Imagine
+    a library for transportion with classes like:
+
+    ```dart
+    abstract final class Vehicle {}
+
+    class LandVehicle extends Vehicle {}
+    class AquaticVehicle extends Vehicle {}
+    class FlyingVehicle extends Vehicle {}
+    ```
+
+    It allows you to define new subclasses of the various modalities. You can
+    add cars, bikes, canoes, and gliders to it. But it deliberately does not
+    want to support adding entire new modalities by extending `Vehicle`
+    directly. You can't add vehicles that, say, fly through space because the
+    library isn't designed to support that.
+
+    If we require subclasses to have the same restrictions, then there's no way
+    to make `Vehicle` `final` while allowing `LandVehicle` and friends to be
+    extended.
+
+*   **Trust but verify.** In the earlier example, it's not clear what the author
     *intends*. Maybe they deliberately didn't put any modifiers on `MySubclass`
     because they *want* to re-add the capability that its superclass removed.
     But maybe they just didn't notice that `NoExtend` removed them, or they
@@ -359,7 +400,9 @@ does `MySubclass` now expose externally? We have a few options:
     @reopen class MySubclass extends NoExtend {}
     ```
 
-    This proposal takes that option.
+This proposal takes the last option where types have exactly the restrictions
+they declare but a lint can be turned on for users who want to be reminded if
+they re-add a capability in a subtype.
 
 Finally, to the actual proposal...
 
@@ -397,8 +440,6 @@ abstract final class
 mixin
 sealed mixin
 base mixin
-interface mixin
-final mixin
 ```
 
 The grammar is:
@@ -410,13 +451,13 @@ classDeclaration ::=
   '{' (metadata classMemberDeclaration)* '}'
   | classModifiers 'class' mixinApplicationClass
 
-mixinDeclaration ::= mixinModifiers 'mixin' identifier typeParameters?
+classModifiers ::= 'sealed' | 'abstract'? ('base' | 'interface' | 'final')?
+
+mixinDeclaration ::= mixinModifiers? 'mixin' identifier typeParameters?
   ('on' typeNotVoidList)? interfaces?
   '{' (metadata classMemberDeclaration)* '}'
 
-classModifiers ::= 'sealed' | 'abstract'? typeModifiers?
-mixinModifiers ::= 'sealed' | typeModifiers?
-typeModifiers  ::= 'base' | 'interface' | 'final'
+mixinModifiers ::= 'sealed' | 'base'
 ```
 
 ### Static semantics
@@ -460,7 +501,7 @@ warning is reported if a class or mixin is not annotated `@reopen` and it:
     marked `interface` or `final`.
 
 *   Extends, implements, or mixes in a type marked `base` or `final` and is
-    not itself marked `base` or `final`.
+    not itself marked `base`, `final`, or `sealed`.
 
 [meta]: https://pub.dev/packages/meta
 [linter]: https://dart.dev/guides/language/analysis-options#enabling-linter-rules
--- a/working/base-interface-final/feature-specification.md
+++ b/working/base-interface-final/feature-specification.md
@@ -6,9 +6,9 @@ Status: In-progress
 
 Version 1.0
 
-This proposal specifies three modifiers that can be placed on classes and mixins to
-allow an author to control whether the type allows being implemented, extended,
-both, or neither. The proposed modifiers are:
+This proposal specifies three modifiers that can be placed on classes and mixins
+to allow an author to control whether the type allows being implemented,
+extended, both, or neither. The proposed modifiers are:
 
 *   No modifier: As today, the type has no restrictions.
 *   `base`: The type can be extended (if a class) or mixed in (if a mixin) but
@@ -49,15 +49,15 @@ software:
 
 ### Adding methods
 
-It's a compile-time error to have an `implements` clause and not contain
-definitions of every member in the type that you claim to implement. This is a
-*useful* error because it ensures that any member someone can access on a type
-is actually defined and will succeed. It helps you in case you forget to
-implement something.
+It's a compile-time error to have an `implements` clause on a non-`abstract`
+class unless it contains definitions of every member in the type that you claim
+to implement. This is a *useful* error because it ensures that any member
+someone can access on a type is actually defined and will succeed. It helps you
+in case you forget to implement something.
 
 But it also means that if a *new* member is added to a class then every single
 class implementing that class's interface now has a new compile-time error since
-none of them have that member.
+they are very unlikely to coincidentally already have that member.
 
 This makes it hard to add new members to existing public types in packages.
 Since anyone could be implementing that type's interface, any new member is
@@ -248,10 +248,10 @@ the supertype.
 ### Extending non-extensible classes in the same library
 
 Preventing a class from being extended gives you an important invariant: Calls
-to members on that class won't end up in overrides you don't control. This
-invariant remains even if we let you extend the class in the same library. Calls
-to those members may end up in overrides, but they will be overrides you
-yourself wrote in that same library.
+to members on `this` from within that class won't end up in overrides you don't
+control. This invariant remains even if we let you extend the class in the same
+library. Calls to those members may end up in overrides, but they will be
+overrides you yourself wrote in that same library.
 
 Extending non-extensible classes is also really *useful* in API design. It lets
 you offer a class *hierarchy* to users that is closed to further extension.
@@ -264,7 +264,9 @@ and circle. That means it would be hard to correctly support users adding their
 own new subclasses of `Shape` and passing them to the library.
 
 As the shape library author, you want to subclass `Shape` yourself so that you
-can define `Square` and `Circle`, but disallow others from doing so.
+can define `Square` and `Circle`, but disallow others from doing so. (In this
+specific example, you probably also want to prohibit `Shape` from being
+implemented too.)
 
 ### Implementing non-implementable types in the same library
 
@@ -420,9 +422,6 @@ combinations don't make sense:
     `final`.
 *   `base`, `interface`, and `final` all control the same two capabilities so
     are mutually exclusive.
-*   `final` and `interface` imply that the type can't be used as a superclass,
-    but the reason for defining a mixin is so that it can be mixed in and become
-    a superclass, so we don't allow those on mixins.
 
 The remaining valid combinations are:
 
@@ -440,6 +439,8 @@ abstract final class
 mixin
 sealed mixin
 base mixin
+interface mixin
+final mixin
 ```
 
 The grammar is:
@@ -457,7 +458,7 @@ mixinDeclaration ::= mixinModifiers? 'mixin' identifier typeParameters?
   ('on' typeNotVoidList)? interfaces?
   '{' (metadata classMemberDeclaration)* '}'
 
-mixinModifiers ::= 'sealed' | 'base'
+mixinModifiers ::= 'sealed' | 'base' | 'interface' | 'final'
 ```
 
 ### Static semantics
--- a/working/base-interface-final/feature-specification.md
+++ b/working/base-interface-final/feature-specification.md
@@ -1,0 +1,470 @@
+# "Base", "interface", and "final" types
+
+Author: Bob Nystrom
+
+Status: In-progress
+
+Version 1.0
+
+This proposal specifies three modifiers that can be placed classes and mixins to
+allow an author to control whether the type allows being implemented, extended,
+both, or neither. The proposed modifiers are:
+
+*   No modifier: As today, the type has no restrictions.
+*   `base`: The type can be extended or mixed in but not implemented.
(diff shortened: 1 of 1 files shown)


working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+Status: In-progress
+
+Version 1.0
+
+This proposal specifies three modifiers that can be placed classes and mixins to
+allow an author to control whether the type allows being implemented, extended,
+both, or neither. The proposed modifiers are:
+
+*   No modifier: As today, the type has no restrictions.
+*   `base`: The type can be extended or mixed in but not implemented.
```
lrhn, line 14: Could we remove the "mixed in" and require you to write `mixin` to get that.

Then `mixin` could combine with `base class` or plain `class`, to allow the class (with only `Object` as superclass and no declared constructors) to be used as a mixin.
Then we can finally stop assuming that *any* class satisfying those random requirements can potentially be mixed in.

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+*   `interface`: The type can be implemented but not extended or mixed in.
```
lrhn, line 15: The `interface class` can only be implemented. It cannot be combined with `base` or `mixin`.
Could we allow omitting the `class` and just write `interface Foo`?

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+### Adding methods
+
+It's a compile-time error to have an `implements` clause and not contain
+definitions of every member in the type that you claim to implement. This is a
+*useful* error because it ensures that any member someone can access on a type
+is actually defined and will succeed. It helps you in case you forget to
+implement something.
+
+But it also means that if a *new* member is added to a class then every single
+class implementing that class's interface now has a new compile error since none
```
lrhn, line 58: compile error -> compile-time error
(Or "compilation error")

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+Most types contain methods that invoke other methods on `this`, for example:
+
+```dart
+class Account {
+  int _balance = 0;
+
+  bool canWithdraw(int amount) => amount <= _balance;
+
+  bool tryWithdraw(int amount) {
+    if (amount <= 0) throw ArgumentError("Amount must be positive.");
```
lrhn, line 87: Show a good example, never use the `ArgumentError` constructor with just a message :wink: 

```dart
  RangeError.checkNotNegative(amount, "amount");
```
or
```dart
  if (amount <= 0) throw ArgumentError.value(amount, "amount", "Must be positive");
```

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+
+    if (!canWithdraw(amount)) return false;
```
lrhn, line 89: This can also be seen as an example of not wanting a virtual invocation.
What if you could write
```dart
 if (!Account.this.canWithdraw(amount)) return false;
```
and get a non-virtual this-call (like a `super`-call, just on this class)?

(But not relevant to this document, just saying that `final` isn't necessarily the *only* solution to this particular problem.)

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+  // YOLO.
+  bool canWithdraw(int amount) => true;
+}
+```
+
+Extending a class gives you free rein to override whatever methods you want,
+while *also* inheriting concrete implementations of other methods that may
+assume you *haven't* done that. If we prevent `Account` from being subclassed,
+we ensure that when `tryWithdraw()` calls `canWithdraw()`, it calls the actual
+`canWithdraw()` method we expect.
```
lrhn, line 113: Or we could make `canWithdraw` itself `final`, which is what I'd go for instead of locking down the entire class.

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+In return for that restriction, in a switch, if you cover all of those subtypes,
+then the compiler knows that you have [exhaustively][exhaustive] covered all
+possible instances of the supertype. This is a big part of enabling a [functional
+programming style][fp] in Dart.
+
+[exhaustive]: https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/exhaustiveness.md
+
+[fp]: https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#algebraic-datatypes
+
+The `sealed` modifier prevents subtyping from outside of the library where the
```
lrhn, line 239: ... prevents direct subtyping ....

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+
+As the shape library author, you want to subclass `Shape` yourself so that you
+can define `Square` and `Circle`, but disallow others from doing so.
+
+### Implementing non-implementable types in the same library
+
+A key invariant you get by preventing a type from being implemented is that it
+becomes safe to access private members defined on that type without risking a
+runtime exception. You are ensured that any instance of the type is an instance
+of a type from your library that inclines all of its private members.
```
lrhn, line 271: "inclines"?

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+*   `sealed` implies non-extensibility, so can't be combined with `interface`
+    or `final`.
+*   `sealed` implies non-implementability, so can't be combined with `base` or
+    `final`.
+*   `base`, `interface`, and `final` all control the same two capabilities so
+    are mutually exclusive.
+*   `final` and `interface` imply that the type can't be used as a superclass,
+    but the reason for defining a mixin is so that it can be mixed in and become
+    a superclass, so we don't allow those on mixins.
+
```
lrhn, line 383: I'd add `mixin` to the mix.

* `mixin` can be combined with `base`, `class`, and with `abstract` if combined with `class`.

(as any of `base mixin class`, `mixin class`, `base mixin` or `mixin`, the first two can also be combined with `abstract`).

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+final class
+abstract class
+abstract base class
+abstract interface class
+abstract final class
+
+mixin
+sealed mixin
+base mixin
+interface mixin
```
lrhn, line 400: Not valid.

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+final mixin
```
lrhn, line 401: Not valid.

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+```
+class
+sealed class
+base class
+interface class
+final class
+abstract class
+abstract base class
+abstract interface class
+abstract final class
```
lrhn, line 395: So I'd add:

* `mixin class`
* `sealed mixin class`
* `base mixin class`
* `abstract mixin class`
* `abstract base mixin class`

as well. Those are the only classes which can be mixed in. (Possibly only outside of the library.)

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+  superclass? interfaces?
+  '{' (metadata classMemberDeclaration)* '}'
+  | classModifiers 'class' mixinApplicationClass
+
+mixinDeclaration ::= mixinModifiers 'mixin' identifier typeParameters?
+  ('on' typeNotVoidList)? interfaces?
+  '{' (metadata classMemberDeclaration)* '}'
+
+classModifiers ::= 'sealed' | 'abstract'? typeModifiers?
+mixinModifiers ::= 'sealed' | typeModifiers?
```
lrhn, line 418: can just be
```ebnf
mixinModifiers ::= 'sealed' | `base`
```

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+  classModifiers 'class' identifier typeParameters?
+  superclass? interfaces?
+  '{' (metadata classMemberDeclaration)* '}'
+  | classModifiers 'class' mixinApplicationClass
+
+mixinDeclaration ::= mixinModifiers 'mixin' identifier typeParameters?
+  ('on' typeNotVoidList)? interfaces?
+  '{' (metadata classMemberDeclaration)* '}'
+
+classModifiers ::= 'sealed' | 'abstract'? typeModifiers?
```
lrhn, line 417: I suggest adding 
```enbf
::'sealed' | 'abstract'? typeModifiers? 'mixin`?
```

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+mixinModifiers ::= 'sealed' | typeModifiers?
+typeModifiers  ::= 'base' | 'interface' | 'final'
+```
+
+### Static semantics
+
+It is a compile-time error to:
+
+*   Extend a class marked `interface` or `final` outside of the library where it
```
lrhn, line 426: or `sealed`

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+    is defined.
+
+*   Implement a type marked `base` or `final` outside of the library where it is
+    defined.
```
lrhn, line 430: or `sealed`

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+    defined without also being marked `base` or `final`. *This ensures that a
+    subtype can't escape the `base` restriction of its supertype by offering its
+    _own_ interface that could then be implemented without inheriting the
+    concrete implementation from the supertype.*
+
+*   Mix in a class marked `sealed`, `base`, `interface`, or `final`. *We want to
+    eventually move away from classes as mixins. We don't want to break existing
+    uses of classes as mixins but since no existing code is using these
+    modifiers, we can prevent classes using those modifiers from also being used
+    as mixins.*
```
lrhn, line 442: Ahhh, there it is.

If we don't think we can get there, I'd be fine with just marking `class`es that can be mixed in as `mixin class`.

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+    But I do think it could be useful to offer this as a lint with a metadata
+    annotation for users that are more cautious, like:
+
+    ```dart
+    interface class NoExtend {}
+    @reopen class MySubclass extends NoExtend {}
+    ```
+
+    This proposal takes that option.
+
```
lrhn, line 363: Missing the option of *not allowing* you to reopen, and requiring you to repeat the annotation in subclasses inside the library, just as you are required to outside. You can ignore the restrictions, but you cannot remove them. 
If you implement a `base` class, the implementation must still be `base` or `final`, even inside the library.

I know that the options that allow you to reopen also allow you to not do so, so not allowing you to reopen is more restrictive on users. I think it might be worth it, to avoid accidental re-opening.

(I guess the annotation and lint pushes users towards not reopening, so if you never use the annotation/get the lint, you are effectively getting the no-reopening effect. And the compiler can always see whether you reopen or not, and disable any optimizations that don't apply then. I'd just want to consider whether silently disabling optimizations will be a trip-wire for users.)

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+A metadata annotation `@reopen` is added to package [meta][] and a lint
+"require_reopen" is added to the [linter][]. When the lint is enabled, a lint
+warning is reported if a class or mixin is not annotated `@reopen` and it:
+
+*   Extends or mixes in a type marked `interface` or `final` and is not itself
+    marked `interface` or `final`.
+
+*   Extends, implements, or mixes in a type marked `base` or `final` and is
+    not itself marked `base` or `final`.
```
lrhn, line 463: Unless the class or mixin is itself marked `sealed`, in which case you _cannot_ add `base` or `final`, but the transitive subclasses must anyway.

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+Instead, it's a suggested part of the overall user experience of the feature.
+
+A metadata annotation `@reopen` is added to package [meta][] and a lint
+"require_reopen" is added to the [linter][]. When the lint is enabled, a lint
+warning is reported if a class or mixin is not annotated `@reopen` and it:
+
+*   Extends or mixes in a type marked `interface` or `final` and is not itself
+    marked `interface` or `final`.
+
+*   Extends, implements, or mixes in a type marked `base` or `final` and is
```
lrhn, line 462: Is this directly extends, implements or mixes in, or transitive?

If transitive, we need to be concerned about paths going through other reopened classes, possibly through other libraries. If not transitive ...

natebosch: I think directly.

working/base-interface-final/feature-specification.md
munificent: Done.

working/base-interface-final/feature-specification.md
munificent: > Could we allow writing just interface to mean the same as abstract interface class?

I've toyed with that idea as I've noodled on these various proposal over the years, but I've generally leaned away from it because I suspect it might be confusing that a thing labeled only `interface` can still have static methods, constructors, and even concrete instance methods.

I figure the best path is to just have a handful of relatively simple modifiers and let them be composed in obvious ways without offering too many shorthands and sugared combinations.

working/base-interface-final/feature-specification.md
munificent: Done.

working/base-interface-final/feature-specification.md
munificent: Done.

working/base-interface-final/feature-specification.md
munificent: Yeah, there's a whole other much larger discussion to have around non-virtual members that also touches protected, final members (virtual but not overridable), and maybe even overloading.

For now I just want to worry about modifiers on types. Modifiers on members can come later.

working/base-interface-final/feature-specification.md
munificent: True, but that's a different proposal. :)

working/base-interface-final/feature-specification.md
munificent: Done.

working/base-interface-final/feature-specification.md
munificent: Oops.

working/base-interface-final/feature-specification.md
munificent: Oops, fixed.

working/base-interface-final/feature-specification.md
munificent: Done.

working/base-interface-final/feature-specification.md
munificent: Thanks! I had toyed with whether it made sense to allow `interface` and `final` on mixins before ultimately deciding no and accidentally left things in a weird state. Fixed.

working/base-interface-final/feature-specification.md
munificent: Yes, but that's already specified under the `sealed` proposal. :)

This one doesn't respecify that, it just inherits the grammar.

working/base-interface-final/feature-specification.md
munificent: > Maybe we should embrace that some types are both mixins and classes, and allow you to declare them as mixin class.

We have talked about that before in earlier discussions around all of the various combinations of capabilities, but I don't think we ever found any real use cases for it. Given how confusing mixins already are, I think it makes the most sense to just keep them separate.

You can always define a mixin and then a mixin application class of it if you want to give users both affordances.

working/base-interface-final/feature-specification.md
munificent: Ah, good point. I'm not in favor of that option, but it's a good one to discuss. Added a section.

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,470 @@
...
+    a class marked `reopen` means the same thing as a class not marked with it.
+    But I do think it could be useful to offer this as a lint with a metadata
+    annotation for users that are more cautious, like:
+
+    ```dart
+    interface class NoExtend {}
+    @reopen class MySubclass extends NoExtend {}
+    ```
+
+    This proposal takes that option.
```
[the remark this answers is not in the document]
munificent: > Most of the cases are simple, but what should it do if you implement a `base` class. Should the new class be base too? The use-cases above suggests that it's not necessary, because base is about the concrete implementation, and the class using implements won't have anything to do with that implementation.

According to the next section, I do have the lint fire in this case (or require you to add `@reopen` to silence it). The `base` modifier is about removing the ability to implement the type. That guarantees that every instance of the type is an instance of that class or a subclass. If you have:

```dart
base class A {}
class B implements A {}
```

Then an instance of `B` is a subtype of `A` but doesn't inherit any of `A` methods and doesn't go through `A`'s generative constructor. Those are invariants you might care about, so the lint fires here.

The other side is `interface`:

```dart
interface class A {}
class B implements A {}
```

The lint does *not* fire here. The `interface` modifier's useful invariants are about preventing overriding of all methods and disallowing inheritance. Since `B` doesn't inherit from or allow anyone to override any methods in `A`, it doesn't break that invariant. Even if you extend `B`, you don't inherit anything from `A`, so there's no warning here.

working/base-interface-final/feature-specification.md
munificent: Ah good catch. Done.

working/base-interface-final/feature-specification.md
munificent: I haven't thought deeply about it, but I don't think it should be transitive? If you've already deliberately re-added a capability and put `@reopen` on that the type that does it, at that point further subtypes of the reopened type shouldn't need to restate that they also have the capability.

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,511 @@
...
+final class
+abstract class
+abstract base class
+abstract interface class
+abstract final class
+
+mixin
+sealed mixin
+base mixin
+```
```
eernstg, line 443: I suspect that the same bits make sense here, except `new`.

| Bits | Syntax | Properties outside owning library |
|----------------|-----------------------------------|----------------------|
| `00` | `final mixin`, `sealed mixin` | Just usable as a type |
| `01` | `interface mixin` | Can be implemented, but not mixed in |
| `10` | `base mixin` | Can be mixed in, but not implemented |
| `11` | `mixin` | Can be mixed in and implemented |

Again, the `final mixin` and the `sealed mixin` would only differ in that subtypes of the former are linted if they aren't final.

eernstg (commented):
This is great!

Added a handful of comments.

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,511 @@
...
+class
+sealed class
+base class
+interface class
+final class
+abstract class
+abstract base class
+abstract interface class
+abstract final class
```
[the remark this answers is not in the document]
natebosch: > it's not obvious to me that `sealed` needs to differ from `abstract final`.

With `abstract final` I can still add a new subclass in the same library at any time and it doesn't break consumers. With `sealed` I need to promise that I will never add another direct subclass after I publish.

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,511 @@
...
+subtypes within the same library so that you can pattern match on those to cover
+the supertype.
+
+### Extending non-extensible classes in the same library
+
+Preventing a class from being extended gives you an important invariant: Calls
+to members on that class won't end up in overrides you don't control. This
+invariant remains even if we let you extend the class in the same library. Calls
+to those members may end up in overrides, but they will be overrides you
+yourself wrote in that same library.
```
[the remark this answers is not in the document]
munificent: Good point, updated to clarify "on `this`", which is what this invariant described earlier is about.

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,511 @@
...
+
+Consider the earlier example where you have a `Shape` base class and a couple of
+subclasses. Let's say you also have code in that library for performing
+intersection tests on pairs of shapes. That intersection code needs special
+support for each *pair* of types: square and square, square and circle, circle
+and circle. That means it would be hard to correctly support users adding their
+own new subclasses of `Shape` and passing them to the library.
+
+As the shape library author, you want to subclass `Shape` yourself so that you
+can define `Square` and `Circle`, but disallow others from doing so.
```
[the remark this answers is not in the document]
munificent: Good point. Added a parenthetical.

working/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,511 @@
...
+*   `sealed` implies `abstract`, so they can't be combined.
+*   `sealed` implies non-extensibility, so can't be combined with `interface`
+    or `final`.
+*   `sealed` implies non-implementability, so can't be combined with `base` or
+    `final`.
+*   `base`, `interface`, and `final` all control the same two capabilities so
+    are mutually exclusive.
+*   `final` and `interface` imply that the type can't be used as a superclass,
+    but the reason for defining a mixin is so that it can be mixed in and become
+    a superclass, so we don't allow those on mixins.
```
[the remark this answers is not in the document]
munificent: You're exactly right. It does make sense to allow `final` and `interface` on mixin. Updated the proposal.

working/base-interface-final/feature-specification.md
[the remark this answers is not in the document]
munificent: > However, `sealed class` and `abstract final class` is the same thing according to these bits, and it's not obvious to me that `sealed` needs to differ from `abstract final`.

There's an extra fourth bit: exhaustiveness checks. Marking a class `sealed` opts that class into exhaustiveness checks in switches. Marking it `abstract final class` does not:

```dart
sealed class A {}
class A2 extends A {}
class A3 extends A {}

abstract final class B {}
class B2 extends B {}
class B3 extends B {}

test(A a, B b) {
  print(switch (a) {
    A1() => 'A1',
    A2() => 'A2'
  }); // No error here.

  print(switch (b) {
    B1() => 'B1',
    B2() => 'B2'
  }); // Error, switch isn't (known to be) exhaustive.
}
```

In principle, yes, the compiler *could* tell that switches over subclasses of `abstract final` classes are not exhaustive. But using `abstract final` instead of `sealed` explicitly opts *out* of that checking. That forces users to write:

```dart
test(B b) {
  print(switch (b) {
    B1() => 'B1',
    B2() => 'B2',
    _ => 'some default clause'
  });
}
```

That in turn means that if the author of `B` later adds another subclass, then the switch (which may be in a different library or package) still compiles without error. You can think of `abstract final` as sort of `sealed late` in that it turns exhaustiveness compile time errors into runtime errors (or at least runtime *handled*) by forcing users to add default cases to their switches.

It's a subtle distinction, but one that could be useful for some package maintainers who want the ability to ship new subclasses of a class without it being a breaking change to users of the API.

working/base-interface-final/feature-specification.md
munificent: Yup, done.

working/base-interface-final/feature-specification.md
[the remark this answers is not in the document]
eernstg: Makes sense!

munificent:
I'm going to go ahead and land this, because it's just a proposal either way. I'd like to talk about moving it forward soon, though, since I think it will be valuable to have these modifiers at the same time that we introduce `sealed`.

PR #2655, opened 2022-11-24
[Patterns] Small clarification added

sgrekhov:
The subpattern must be a variable pattern not in all cases but in case of `: pattern` only. Without this change it can be read as if subpattern always must be a variable pattern


munificent (approved):
Thanks! I agree this reads better.

PR #2674, opened 2022-11-30
Add "mixin class" to the "base-interface-final" proposal.

munificent:
I went ahead and took the more aggressive step of also saying that you can't mixin a class unless it has the "mixin" modifier. However, that rule is gated by a language version so that it's not breaking.

Since "mixin class" allows you to preserve the previous behavior of class+mixin, I think we can get away with doing this change in one step. Basically, whenever you upgrade your library to the latest version, you add "mixin" to whatever classes you want to support being used as mixins and you're done.

If you think that's too aggressive, let me know and I can change this. If we do that, though, it ends up meaning that the "mixin" modifier doesn't actually *do* anything (except yell at you if you apply it to a class with a generative constructor or non-Object superclass).

Lasse, I really like your suggestion for "mixin class". I think this lets us close the loose end of ChangeNotifier henceforth and forever without a lengthy migration and lets us support a rare but potentially useful combination of capabilities.

@mit-mit @eernstg @stereotype441 @jakemac53 @natebosch @kallentu

### DIFF
--- a/working/base-interface-final/feature-specification.md
+++ b/working/base-interface-final/feature-specification.md
@@ -412,11 +412,11 @@ they re-add a capability in a subtype.
 
 ## Mixin classes
 
-Inline with Dart's permissive default nature, Dart allows any class declaration
+In line with Dart's permissive default nature, Dart allows any class declaration
 to also be used as a mixin (in spec parlance, it allows a mixin to be "derived
 from a class declaration"), provided the class meets the restrictions that
-mixins require: it must extend `Object` and not define any generative
-constructors.
+mixins require: Its immediate superclass must be `Object` and it must not
+declare any generative constructors.
 
 In practice, mixins are quite different from classes and it's uncommon for users
 to deliberately define a type that is used as both. It's easy to define a class
@@ -426,12 +426,12 @@ breaking change to any downstream user that had that class in a `with` clause.
 
 Eventually, we would like classes to not be usable as a mixin by default. Being
 extensible and implementable seems to be the right default for classes based on
-Dart's history and talking to users. But being miscible by default is rarely
-(but not *never*) useful and often confusing. It makes classes brittle for
-little upside.
+Dart's history and talking to users. But being usable as a mixin by default is
+rarely (but not *never*) useful and often confusing. It makes classes brittle
+for little upside.
 
-Instead, we under this proposal authors to explicitly opt in to allowing the
-class to be used as a mixin by adding a `mixin` modifier to the class:
+Instead, under this proposal we require authors to explicitly opt in to allowing
+the class to be used as a mixin by adding a `mixin` modifier to the class:
 
 ```dart
 class OnlyClass {}
@@ -447,9 +447,9 @@ class UsesAsMixin extends OtherSuperclass with Both {} // OK.
 
 This change is guarded by a language version. Any class declaration in a library
 whose language version is before the one this feature ships in can be used as a
-mixin as long as the class meets the mixin restrictions. Libraries whose version
-is later than that must be explicitly marked `mixin class` to allow the class to
-be used in a `with` clause.
+mixin as long as the class meets the mixin restrictions. Classes in libraries
+whose version is later than that must be explicitly marked `mixin class` to
+allow the class to be used in a `with` clause.
 
 When upgrading your library to the new language version, you can preserve the
 existing behavior by adding `mixin` to every class declaration that can be used
@@ -471,7 +471,8 @@ Some combinations don't make sense:
     `final`.
 *   `base`, `interface`, and `final` all control the same two capabilities so
     are mutually exclusive.
-*   `mixin` as a modifier can obviously only be applied to a class.
+*   `mixin` as a modifier can obviously only be applied to a class. It can be
+    combined with any other modifiers that can be applied to a class.
 
 The remaining valid combinations are:
 
@@ -538,27 +539,33 @@ typeModifier ::= 'base' | 'interface' | 'final'
 It is a compile-time error to:
 
 *   Extend a class marked `interface` or `final` outside of the library where it
-    is defined.
+    is declared.
 
 *   Implement a type marked `base` or `final` outside of the library where it is
-    defined.
+    declared.
 
 *   Extend or mix in a type marked `base` outside of the library where it is
-    defined without also being marked `base` or `final`. *This ensures that a
+    declared without also being marked `base` or `final`. *This ensures that a
     subtype can't escape the `base` restriction of its supertype by offering its
     _own_ interface that could then be implemented without inheriting the
     concrete implementation from the supertype.*
 
 *   Mix in a class not marked `mixin` outside of the library where it is
-    defined, unless the class declaration is in a library whose language version
-    is older than the version this feature ships in.
+    declared, unless the class declaration is in a library whose language
+    version is older than the version this feature ships in.
 
 *   Apply the `mixin` modifier to a class whose superclass is not `Object` or
-    that defines a generative constructor. *The `mixin` modifier states that
+    that declares a generative constructor. *The `mixin` modifier states that
     you intend the class to be mixed in, which is inconsistent with defining a
     class that can't be used as a mixin. Note that this means that the `mixin`
     modifier becomes a helpful reminder to ensure that you don't inadvertently
     break your class's ability to be used as a mixin.*
+
+*   Mix in a class whose superclass is not `Object` or that declares a
+    generative constructor. *Because of the previous rule, this rule only comes
+    into play when you use a class not marked `mixin` as a mixin within the
+    library where it's declared. When you do that, the existing restriction
+    still applies that the class being used as a mixin must be valid to do so.*
 
 A typedef can't be used to subvert these restrictions. When extending,
 implementing, or mixing in a typedef, we look at the library where type the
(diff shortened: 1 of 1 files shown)


natebosch:
Do we expect that authors of classes used as mixins are reliably aware of that usage?

I think this is worth the risk as long as it's not breaking to add `mixin` to an existing class so that we can safely republish if we find there is a critical use case.

lrhn:
I expect authors of classes that are *intended* to be used as mixins to be aware of that (because they have documented it on the class, otherwise they probably didn't intend it).

It should not be breaking to add `mixin` to a class declaration. Adding `mixin` puts restrictions on the class itself (no generative constructors declared, no superclass other than `Object`), but if the class was used as a mixin before, it already satisfied those constraints. From the outside, adding `mixin` just allows you to mix in the class, it doesn't prevent anything. 

Adding `base` and *not* `mixin` may break code that depended on mixing in the class, because they no longer can.
That means that forgetting to add the `mixin` is the breaking change. If the class is intended as a mixin, your own tests should catch that.

Changing a class that was not intended as a mixin into a class that *cannot* be used as a mixin might break users who used it as a mixin anyway.
Preventing such unintended use is the point of restricting you to mixing in classes that are declared as `mixin class`, so it'd be working as intended in breaking those classes.

working/base-interface-final/feature-specification.md
```diff
@@ -406,14 +410,59 @@ This proposal takes the last option where types have exactly the restrictions
 they declare but a lint can be turned on for users who want to be reminded if
 they re-add a capability in a subtype.
 
-Finally, to the actual proposal...
+## Mixin classes
+
+Inline with Dart's permissive default nature, Dart allows any class declaration
+to also be used as a mixin (in spec parlance, it allows a mixin to be "derived
+from a class declaration"), provided the class meets the restrictions that
+mixins require: it must extend `Object` and not define any generative
```
lrhn, line 418: "it must extend `Object`" -> "Its immediate superclass must be `Object`"

(The word "extend" is sometimes used transitively, and if taken so every non-`Null` class extends `Object`. The word "superclass" usually isn't transitive. I added "immediate" just to be sure, but I don't think it's technically necessary.)

Also: define -> declare

(Might be O., @eernstg, do we use "define" interchangeably with "declare"? 
I fear that "define" might include implicit members, which would include the default constructor, which is the one constructor a mixin class is allowed to have. With `declare`, we're safer, because you definitely don't declare the default constructor. But we might want to reevaluate phrasing if we introduce default constructor syntax - is it a declaration or not? But that's a future-me's problem.)

working/base-interface-final/feature-specification.md
```diff
@@ -406,14 +410,59 @@ This proposal takes the last option where types have exactly the restrictions
...
+
+In practice, mixins are quite different from classes and it's uncommon for users
+to deliberately define a type that is used as both. It's easy to define a class
+without *intending* it to be used as a mixin and then accidentally forbid that
+usage by adding a generative constructor or superclass to the class. That is a
+breaking change to any downstream user that had that class in a `with` clause.
+
+Eventually, we would like classes to not be usable as a mixin by default. Being
+extensible and implementable seems to be the right default for classes based on
+Dart's history and talking to users. But being miscible by default is rarely
```
lrhn, line 429: "Miscible". TIL. Great word. 
Wouldn't know what it meant without context (and basically knowing what we're trying to say here before reading it), so maybe just use "usable as a mixin".

working/base-interface-final/feature-specification.md
```diff
@@ -406,14 +410,59 @@ This proposal takes the last option where types have exactly the restrictions
...
+(but not *never*) useful and often confusing. It makes classes brittle for
+little upside.
+
+Instead, we under this proposal authors to explicitly opt in to allowing the
```
lrhn, line 433: At least one word missing here.

eernstg:
I hope we can do this! However, it could be helpful to get some input from the Flutter team: Which and how many classes do they expect to turn into mixin classes? Do they know, or would they be worried about finding them? Luckily, lots of classes just become an error if `mixin` is added, which means that they can be ruled out immediately (that is, they _aren't_ in use as mixins).

working/base-interface-final/feature-specification.md
```diff
@@ -406,14 +410,59 @@ This proposal takes the last option where types have exactly the restrictions
...
+
+class UsesAsSuperclass extends Both {}
+
+class UsesAsMixin extends OtherSuperclass with Both {} // OK.
+```
+
+This change is guarded by a language version. Any class declaration in a library
+whose language version is before the one this feature ships in can be used as a
+mixin as long as the class meets the mixin restrictions. Libraries whose version
+is later than that must be explicitly marked `mixin class` to allow the class to
```
lrhn, line 451: As phrased, it's the libraries which are marked `mixin class`.

working/base-interface-final/feature-specification.md
```diff
@@ -406,14 +410,59 @@ This proposal takes the last option where types have exactly the restrictions
 they declare but a lint can be turned on for users who want to be reminded if
 they re-add a capability in a subtype.
 
-Finally, to the actual proposal...
+## Mixin classes
+
+Inline with Dart's permissive default nature, Dart allows any class declaration
```
eernstg, line 415:
```suggestion
In line with Dart's permissive default nature, Dart allows any class declaration
```

working/base-interface-final/feature-specification.md
```diff
@@ -422,9 +471,12 @@ combinations don't make sense:
     `final`.
 *   `base`, `interface`, and `final` all control the same two capabilities so
     are mutually exclusive.
+*   `mixin` as a modifier can obviously only be applied to a class.
```
lrhn, line 474: Maybe say: It can be combined with any of the other modifiers.

working/base-interface-final/feature-specification.md
eernstg, line 433:
```suggestion
Instead, under this proposal authors can explicitly opt in to allowing the
```

working/base-interface-final/feature-specification.md
```diff
@@ -477,11 +549,16 @@ It is a compile-time error to:
...
-    eventually move away from classes as mixins. We don't want to break existing
-    uses of classes as mixins but since no existing code is using these
-    modifiers, we can prevent classes using those modifiers from also being used
-    as mixins.*
+*   Mix in a class not marked `mixin` outside of the library where it is
+    defined, unless the class declaration is in a library whose language version
+    is older than the version this feature ships in.
+
+*   Apply the `mixin` modifier to a class whose superclass is not `Object` or
+    that defines a generative constructor. *The `mixin` modifier states that
```
lrhn, line 557: defines -> declares

working/base-interface-final/feature-specification.md
```diff
@@ -477,11 +549,16 @@ It is a compile-time error to:
...
 
-*   Mix in a class marked `sealed`, `base`, `interface`, or `final`. *We want to
-    eventually move away from classes as mixins. We don't want to break existing
-    uses of classes as mixins but since no existing code is using these
-    modifiers, we can prevent classes using those modifiers from also being used
-    as mixins.*
+*   Mix in a class not marked `mixin` outside of the library where it is
+    defined, unless the class declaration is in a library whose language version
+    is older than the version this feature ships in.
+
```
lrhn, line 555: Should we add

  * Mix in a class with a superclass other than `Object`, or which declares a generative constructor.

(which only applies to mixin applications of a non-`mixin` `class` inside the same library. Or do we consider that covered by the existing spec.)

lrhn (approved):
LGTM, ship it!

eernstg (approved):
Suggested a couple of typo-level changes

working/base-interface-final/feature-specification.md
munificent: Sometimes I write these things just to delight myself. I'll reword it. :)

working/base-interface-final/feature-specification.md
munificent: Done.

working/base-interface-final/feature-specification.md
munificent: I accidentally a word.

working/base-interface-final/feature-specification.md
munificent: Oops, fixed.

working/base-interface-final/feature-specification.md
munificent: Done.

working/base-interface-final/feature-specification.md
munificent: Done.

working/base-interface-final/feature-specification.md
munificent: Added a parenthetical to clarify.

working/base-interface-final/feature-specification.md
munificent: Thanks! Fixed locally.

working/base-interface-final/feature-specification.md
munificent: Fixed this locally too.

munificent:
> However, it could be helpful to get some input from the Flutter team: Which and how many classes do they expect to turn into mixin classes?

Excellent point. I'll reach out to them.

PR #2675, opened 2022-12-01
[patterns] More small fixes.

### LINKED ISSUE #2657: Ambiguous map patterns a dynamic error?
Thanks to @sgrekhov for bringing up this topic. Consider the following program:

```dart
void main() {
  switch ({1.5: true, 2.5: false}) {
    case {1.5: var x, 1.5: var y}: break;
  }
}
```

The class `double` does not have primitive equality. This means that it is not a compile-time error to have two keys in a map pattern that are equal (basically, we say that this can't be known at compile-time, and that is indeed true in the general case). The run-time behavior is to check that the given scrutinee is a map with suitable type arguments (there are a few possible choices, but they don't matter here), and the `length` of the map matches the number of subpatterns, and that each subpattern matches.

So the match succeeds (and binds `x` and `y` to true).

However, the 'success' is misleading: The `length` requirement certainly doesn't establish any useful guarantees about the scrutinee, and it is confusing (and non-obvious in the general case) that we're binding both `x` and `y` to the value of the same key.

A similar scenario can be created for any key type whose equality isn't primitive. In particular, we could use any key type which is a class `C` where `Object.==` has been overridden by `C` or any of its superclasses. For example:

```dart
class C {
  final int i;
  final String comment;
  const C(this.i, [this.comment = '']);
  bool operator ==(Object other) {
    print('    (invoked ==)');
    if (other.runtimeType != C) return false;
    return (other as C).i == i;
  }
  int get hashCode => classHash ^ i.hashCode;
  static const classHash = 0x123456789;
}

void main() {
  const c1 = C(1, 'Pattern key 1');
  const c2 = C(1, 'Pattern key 2');
  var cs1 = C(1, 'In scrutinee, matched twice');
  var cs2 = C(2, 'In scrutinee, ignored');
  var map = <C, int>{cs1: 42, cs2: 0};

  switch (map) {
    case {c1: var a, c2: var b}: print('$a, $b'); // '42, 42'.
  }
}
```

**This matching behavior seems confusing and bug-prone**, so we might want to report a problem at some point. However, it is an undecidable problem in general whether or not there is a match, because it runs a developer-written function body.

Would it be better to raise a dynamic error in the case where multiple keys in a map pattern are identical or equal? (We can't make it an error to get the same value twice, because there is nothing wrong with `{1: true, 2: true}`).

The 'correct' solution seems to be that we detect the situation where the _k_ key constant expressions of a map pattern do not look up every distinct key of the scrutinee (even though the `length` equality test seems to imply that we're getting all of them). However, there's no easy and performant implementation of "looked up every distinct key of a map". Alternatively, we could check that the matching process obtains every member of the `values.toSet()` of the given scrutinee (in the example above, the pattern matching process never retrieves the value `0`), but that's of course an imprecise check because accidental key duplication could have caused us to omit some values that are duplicates (again `{C(1): true, C(2): true}` could match `C(1)` twice and never match `C(2)`, and still get all values).

@munificent, @stereotype441, @kallentu, @natebosch, @leafpetersen, @jakemac53, @lrhn, WDYT? Should we report a dynamic error because the "precise `length` required, but multiple matches of one key and some other keys ignored" situation is a likely bug? Alternatively, should we just let it pass (and perhaps lint against using map keys with non-primitive equality)?

### LINKED ISSUE #2672: Parsing ambiguity in switch expressions: `=>` might be part of guard expression
The following code is ambiguous:
```dart
f(x, y, z) => switch (x) { _ when y + () => () => z };
```
It could either mean:
```dart
f(x, y, z) => switch (x) { _ when (y + ()) => (() => z) };
```
Or:
```dart
f(x, y, z) => switch (x) { _ when (y + () => ()) => z };
```
(Either of which would successfully type check because `y` has type `dynamic`.)

More generally, in the parser we have the rule that when an expression is expected, if `(...) =>` is found, the parser assumes it's looking at a function expression and greedily consumes the `=>`.  Which means that we currently can't parse this, even though technically it's unambiguous:
```dart
f(x, y, z) => switch (x) { _ when y + () => z };
```
Because the parser interprets it as:
```dart
f(x, y, z) => switch (x) { _ when (y + () => z) };
```
(Which is missing the `=> expression` part of the `switchExpressionCase`.)

There's one exception: if `(...) =>` appears in a constructor initializer, and it's not inside any grouping constructs, then the parser assumes it's looking at a parenthesized expression or a record, so it doesn't consume the `=>`.  This means that the `=>` winds up getting interpreted as the beginning of the constructor body (which is kind of weird because non-factory constructor bodies aren't allowed to use `=>` syntax, but who am I to judge?)

I propose that we address this ambiguity in a similar fashion: if `(...) =>` appears in a guard expression inside a switch expression, and it's not inside any grouping constructs, then let's specify that the construct should be interpreted as a parenthesized expression or a record, thus allowing the `=>` to be interpreted as terminating the `guardedPattern`.  Which means that
```dart
f(x, y, z) => switch (x) { _ when y + () => () => z };
```
will be interpreted as:
```dart
f(x, y, z) => switch (x) { _ when (y + ()) => (() => z) };
```
And
```dart
f(x, y, z) => switch (x) { _ when y + () => z };
```
will be interpreted as
```dart
f(x, y, z) => switch (x) { _ when (y + ()) => z };
```
However, this will be rejected:
```dart
f(x, y, z) => switch (x) { _ when y + () => z => z };
```
Because the parser will interpret it as:
```dart
f(x, y, z) => switch (x) { _ when (y + ()) => (z => z) };
```
(Which is illegal because `z => z` isn't a valid expression.)

munificent:
- Specify exhaustiveness checking of switch elements.
- Resolve ambiguity with `=>` in switch expression guards.
- Compile error if map pattern has identical keys.

Fix #2672.
Fix #2657.

### DIFF
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: Accepted
 
-Version 2.18 (see [CHANGELOG](#CHANGELOG) at end)
+Version 2.19 (see [CHANGELOG](#CHANGELOG) at end)
 
 Note: This proposal is broken into a couple of separate documents. See also
 [records][] and [exhaustiveness][].
@@ -590,6 +590,15 @@ It is a compile-time error if:
     arguments.
 
 *   Any of the entry key expressions are not constant expressions.
+
+*   If any two keys in the map are identical. *Map patterns that don't have a
+    rest element only match if the `length` of the map is equal to the number of
+    map entries. If a map pattern has multiple identical key entries, they will
+    increase the required length for the pattern to match but in all but the
+    most perverse `Map` implementations will represent the same key. Thus, it's
+    very unlikely that any map pattern containing identical keys (and no rest
+    element) will ever match. Duplicate keys are most likely a typo in the
+    code.*
 
 *   If any two keys in the map both have primitive `==` methods, then it is a
     compile-time error if they are equal according to their `==` operator. *In
@@ -1226,6 +1235,59 @@ To avoid that, we disallow a switch expression from appearing at the beginning
 of an expression statement. This is similar to existing restrictions on map
 literals appearing in expression statements. In the rare case where a user
 really wants one there, they can parenthesize it.
+
+#### Function expression in guard ambiguity
+
+Function expressions also use `=>`, which leads to a potential ambiguity:
+
+```dart
+switch (obj) {
+  _ when a + (b) => (c) => body
+};
+```
+
+This could be interpreted as either:
+
+```dart
+switch (obj) {
+  _ when (a + (b)) => ((c) => body)
+  //     ---------    -------------
+};
+
+switch (obj) {
+  _ when (a + (b) => (c)) => (body)
+  //     ----------------    ------
+};
+```
+
+A similar ambiguity exists with function expressions in initializer lists, if
+the constructor happens to be a factory constructor with `=>` for its body. We
+resolve the ambiguity similarly here:  When `=>` is encountered after `when`, if
+the code between forms a valid expression, then it is interpreted as such and
+the `=>` is treated as the separator between the guard and case body. *In the
+above example, we take the first interpretation.*
+
+We do this unconditionally even if the code after `=>` is not a valid body
+expression, as in:
+
+```dart
+switch (obj) {
+  _ when (a) => b => c
+};
+```
+
+Here, we treat the guard expression as `(a)`, which leads the body to be `b =>
+c` which isn't a valid expression and produces a compile-time error.
+
+If you want a guard expression that ends in a function expression (which is
+quite unlikely), you can avoid the `=>` being captured as the case separator by
+parenthesizing the function:
+
+```dart
+switch (obj) {
+  _ when ((a) => b) => c
+};
+```
 
 ### If-case statement and element
 
@@ -2261,8 +2323,9 @@ All other types are not exhaustive. Then:
     is to throw an error and most Dart users prefer to catch those kinds of
     mistakes at compile time.*
 
-*   It is a compile-time error if the cases in a switch statement are not
-    exhaustive and the static type of the matched value is an exhaustive type.
+*   It is a compile-time error if the cases in a switch statement or switch
+    collection element are not exhaustive and the static type of the matched
+    value is an exhaustive type.
 
 [exhaustiveness]: https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/exhaustiveness.md
 
@@ -3119,6 +3182,14 @@ Here is one way it could be broken down into separate pieces:
     *   Parenthesized patterns
 
 ## Changelog
+
+### 2.19
+
+-   Specify exhaustiveness checking of switch elements.
+
+-   Resolve ambiguity with `=>` in switch expression guards (#2672).
+
+-   Compile error if map pattern has identical keys (#2657).
 
 ### 2.18
 
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -1241,7 +1241,7 @@ really wants one there, they can parenthesize it.
 Function expressions also use `=>`, which leads to a potential ambiguity:
 
 ```dart
-switch (obj) {
+var x = switch (obj) {
   _ when a + (b) => (c) => body
 };
 ```
@@ -1249,12 +1249,12 @@ switch (obj) {
 This could be interpreted as either:
 
 ```dart
-switch (obj) {
+var x = switch (obj) {
   _ when (a + (b)) => ((c) => body)
   //     ---------    -------------
 };
 
-switch (obj) {
+var x = switch (obj) {
   _ when (a + (b) => (c)) => (body)
   //     ----------------    ------
 };
@@ -1262,16 +1262,24 @@ switch (obj) {
 
 A similar ambiguity exists with function expressions in initializer lists, if
 the constructor happens to be a factory constructor with `=>` for its body. We
-resolve the ambiguity similarly here:  When `=>` is encountered after `when`, if
-the code between forms a valid expression, then it is interpreted as such and
-the `=>` is treated as the separator between the guard and case body. *In the
-above example, we take the first interpretation.*
+resolve the ambiguity similarly here:  When `=>` is encountered after `when` in
+a guard, if the code between forms a valid expression, then it is interpreted as
+such and the `=>` is treated as the separator between the guard and case body.
+*In the above example, we take the first interpretation.*
 
-We do this unconditionally even if the code after `=>` is not a valid body
-expression, as in:
+This rules applies in all contexts where a guard can appear: switch statements,
+switch elements, switch expressions, and if-case statements. *We could restrict
+this rule to guards only in switch expressions where the ambiguity arises, but
+that leads to a syntactic restriction that is context-sensitive and harder to
+learn. Since the rule is unusual enough as it is, we apply it as consistently as
+possible. Note that the related restriction on `=>` in constructor initializers
+applies even though generative constructors can't use `=>` for their body.*
+
+The rule is applied unconditionally even if the code after `=>` is not a valid
+body expression, as in:
 
 ```dart
-switch (obj) {
+var x = switch (obj) {
   _ when (a) => b => c
 };
 ```
@@ -1284,7 +1292,7 @@ quite unlikely), you can avoid the `=>` being captured as the case separator by
 parenthesizing the function:
 
 ```dart
-switch (obj) {
+var x = switch (obj) {
   _ when ((a) => b) => c
 };
 ```


accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -591,6 +591,15 @@ It is a compile-time error if:
 
 *   Any of the entry key expressions are not constant expressions.
 
+*   If any two keys in the map are identical. *Map patterns that don't have a
```
stereotype441, line 594: I just found out from @johnniwinther that this is difficult to do in the CFE, because the CFE does constant evaluation as a separate step, after lowering everything to a kernel representation, and the kernel representation doesn't preserve patterns (because they get lowered to imperative code).

Could we make this an analyzer hint rather than a warning (kind of like we do for non-const maps)?  That way the user still finds out about the problem, but we don't have to pay the CFE implementation cost.

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -1227,6 +1236,59 @@ of an expression statement. This is similar to existing restrictions on map
...
+
+switch (obj) {
+  _ when (a + (b) => (c)) => (body)
+  //     ----------------    ------
+};
+```
+
+A similar ambiguity exists with function expressions in initializer lists, if
+the constructor happens to be a factory constructor with `=>` for its body. We
+resolve the ambiguity similarly here:  When `=>` is encountered after `when`, if
```
stereotype441, line 1265: Can you add a sentence or two clarifying whether this applies to just `when` clauses inside switch expressions, or to `when` clauses in general?  E.g. is this allowed?

```dart
f(x, y) {
  switch (x) {
    case _ when y + () => z:
      break;
  }
}
```
(and a similar question for if-case).

Note: See discussion at https://github.com/dart-lang/language/issues/2672#issuecomment-1333968549

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -1227,6 +1236,59 @@ of an expression statement. This is similar to existing restrictions on map
 literals appearing in expression statements. In the rare case where a user
 really wants one there, they can parenthesize it.
 
+#### Function expression in guard ambiguity
+
+Function expressions also use `=>`, which leads to a potential ambiguity:
+
+```dart
+switch (obj) {
```
stereotype441, line 1244: This example looks like it introduces a switch statement rather than a switch expression, due to the "Expression statement ambiguity" resolution above.  How about if we make the example:

```dart
var x = switch(object) {
  ...
};
```
just so that it's clear that we're talking about a switch expression?

accepted/future-releases/0546-patterns/feature-specification.md
munificent: The spec already states that it's an error if two map keys both have primitive equality and their `==` returns `true`, which also involves constant evaluation. Does that one worry them too?

The language also already has a similar check for constant map literals. Is there a reason this one is harder?

I'm not attached to it either way, I'm just a little surprised it's a problem given that the very next bullet point also requires it and so far no one has said it was a challenge.

accepted/future-releases/0546-patterns/feature-specification.md
munificent: Done.

accepted/future-releases/0546-patterns/feature-specification.md
munificent: Good catch, done.

accepted/future-releases/0546-patterns/feature-specification.md
lrhn: If the pattern lowering (desugaring) removes the information that these constants were map pattern keys of the same pattern, then it's harder than for a map literal, which is not desugared.

So the main problem is that desugaring happens before constant evaluation, but comparing keys for same-ness happens after constant evaluation, at which point it's no longer clearly visible which things must be non-seme.

(I'm a little impressed that we can lose information like that, and still be able to determine exhaustiveness. I guess the only constants needed for exhaustiveness are enum constants, which can be handled symbolically.)

stereotype441: > (I'm a little impressed that we can lose information like that, and still be able to determine exhaustiveness. I guess the only constants needed for exhaustiveness are enum constants, which can be handled symbolically.)

Yeah, the existing (pre-patterns) exhaustiveness logic only recognizes direct references to enum values.  If your switch case refers to a constant, and that constant evaluates to the enum value, exhaustiveness checking doesn't handle it properly.  So maybe it's not so impressive as you thought 🥲.

leafpetersen: This is a bit concerning for general pattern exhaustiveness checking.  It seems to me that we'll need constants as part of exhaustiveness checking, and I don't imagine we'll do exhaustiveness post-lowering.

leafpetersen (approved):

accepted/future-releases/0546-patterns/feature-specification.md
stereotype441: I've added an item to the agenda to discuss this at the next shared patterns infrastructure meeting.

stereotype441 (approved):
My concerns about implementation feasibility remain, but I think those can be addressed in follow-up discussions, so lgtm.

accepted/future-releases/0546-patterns/feature-specification.md
johnniwinther: The problem is not only the lowering but the fact that identity of constants is known very late in the analysis/compilation process. To support the keys check, and more importantly the exhaustiveness check on constants, we need to ensure that such a time even exists. For this I propose the following:

1. The value of constants (including identities) must never effect type inference/checking.
2. Unknown constants (`.fromEnvironment` constants) must be defined to have an unknown constant value for type inference/checking, "redundancy checking" (like this check for disjoint keys) and exhaustiveness checking.

1) is needed to allow the handling of constant evaluation to be performed fully after inference and I believe that it is already the case; for instance the type of a switch expression does not take the reachability of cases into account. 2) is needed to ensure that redundancy and exhaustiveness is well-defined for unknown constants for both the analyzer (which never knows their value) and the CFE (which only sometimes knows their value).

lrhn: If we _require_ exhaustiveness of enum value switches, then we can assume it during inference, and check it later. 
Might only works for straight switches on the enum type, not, e.g., on `(EnumType, int)`. (Do we require them to be exhausted if the rest of the pattern isn't exhaustive?)

The only other constants that we might do exhaustiveness checking on are booleans and `null`. I don't think we have other guaranteed finite types.

It's still worrisome that we can't prove exhaustiveness of:
```dart
const hasFoo = bool.hasEnvironment("foo");
const noFoo = !hasFoo;
...
switch (boolXFoo) {
  case (hasFoo, var x): ...
  case (noFoo, var x): ...
}
```
I can probably live with "environment"-values being considered too dynamic to be known in switches, but it is a special-case, and it can propagate through multiple constant expressions.

If we fail to show exhustiveness, end it leads to an error, and a "too dynamic" constant is involved in that, we should see if we can mention that in the error message.

munificent: > The value of constants (including identities) must never effect type inference/checking.

Yeah, I don't think that's a problem. As Lasse says, inference/checking can basically assume that a switch *is* exhaustive if it must be and then we just report a compile-time exhaustiveness error later if that turns out to not be the case.

> Unknown constants (.fromEnvironment constants) must be defined to have an unknown constant value for type inference/checking, "redundancy checking" (like this check for disjoint keys) and exhaustiveness checking.

That's definitely an odd corner of the language, but it should be doable. We already need to deal with guards, which is another place where a pattern may not fully participate in exhaustiveness checking because we can't be sure at compile time that it will match.

> (Do we require them to be exhausted if the rest of the pattern isn't exhaustive?)

No:

> We define an *exhaustive type* to be:
>
> ...
>
> *   A record type whose fields are all exhaustive types

So a record type that has any non-exhaustive fields is not required to be exhaustively covered in a switch statement or switch element. Switch expressions must always be exhaustive.

> It's still worrisome that we can't prove exhaustiveness of:
>
> ```dart
> const hasFoo = bool.hasEnvironment("foo");
> const noFoo = !hasFoo;
> ...
> switch (boolXFoo) {
>   case (hasFoo, var x): ...
>   case (noFoo, var x): ...
> }

It's weird, but people don't use environment flags often anyway. You can think of the above example as more or less the same as:

```dart
const hasFoo = bool.hasEnvironment("foo");
const noFoo = !hasFoo;
...
switch (boolXFoo) {
  case (var y, var x) when y == hasFoo: ...
  case (var y, var x) when y == noFoo: ...
}
```

The environment flags are sort of not really constants.

So... how should I resolve this PR? I'm not attached to any particular choice here but I do think we should at least be internally consistent. If primitive `==` keys in a map pattern is a compile-time error then I don't see why identical keys shouldn't be too. If it's too hard to make identical keys and error, then we probably shouldn't make primitive `==` keys an error either.

munificent: OK, I see that the PR has LGTMs, so I'm going to go ahead and merge this in. I'm happy to iterate on it more if needed afterwards. :)

PR #2682, opened 2022-12-02
Minor fixes to the patterns proposal

leafpetersen:
Some small fixes I found in going over the text, and replace `?` with `_` for type schemas to be consistent with the current syntax in `inference.md`.


munificent (approved):
Thanks!

PR #2687, opened 2022-12-06
[patterns] Fix soundness hole in map patterns.

### LINKED ISSUE #2685: Unsoundness in map pattern matching: values can unexpectedly be `null`
Possibly related to https://github.com/dart-lang/language/issues/2671.

Consider the following code:
```dart
void g(Object y) { ... }
void f(Map<String, int> m) {
  swich (m) {
    case {'foo': Object x}:
      g(x);
  }
}
```
Under the current patterns spec, the static analysis of `case {'foo': Object x}` operates as follows:
- The matched type `M` is `Map<String, int>`.
- Since no explicit type arguments are given for the map pattern, the key, value, and key context types are inferred based on the matched type, so they are `K=String`, `V=int`, and `C=String`.  So the required type of the map pattern is `Map<String, int>`.
- The subpattern `Object x` is type checked using `V=int` as the matched value type.
- Since the subpattern `Object x` has a type annotation, its required type is that type, namely `Object`.

And the runtime semantics of the match are:
1. *No type test on `m` since the matched value's static type is a subtype of the pattern's required type (see [Pointless type tests and legacy types](https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#pointless-type-tests-and-legacy-types))*
2. Let `l` be the length of the map `m`, determined by calling `m.length`.
3. If `l != 1` then the match fails.
4. Let `k` be the result of evaluating the key `'foo'`
5. Call `m.containsKey(k)`.  If this returns `false`, the map does not match.
6. Otherwise, let `v = m[k]` and proceed to match `Object x` against `v`.
7. *No type test on `v` since the matched value's static type (`int`) is a subtype of the variable pattern's required type (`Object`).*
8. Let `x = v`.

Note that at no point was `x` checked to see if it was `null`.  Which means that if we inject the following nefarious class into `f`:
```dart
class Nefarious implements Map<String, int> {
  noSuchMethod(invocation) => super.noSuchMethod(invocation);
  operator[](key) => null;
  containsKey(key) => true;
}
main() {
  print(f(Nefarious()));
}
```
Then `g` will be called with an argument of `null`, in violation of soundness.

Side note: as discussed in https://github.com/dart-lang/language/issues/2671, the spec says that the compiler may perform steps 5 and 6 in any order, and may elide either if it determines that doing so will produce the same result.  It is also allowed to assume that

> ...if the map's value type is non-nullable, then when `v[k]` returns `null`, the compiler can assume that the key is absent and `containsKey(k)` would return `false` too.

So the steps above are allowed (but not required) to be optimized to this (differences in bold):
1. *No type test on `m` since the matched value's static type is a subtype of the pattern's required type.*
2. Let `l` be the length of the map `m`, determined by calling `m.length`.
3. If `l != 1` then the match fails.
4. Let `k` be the result of evaluating the key `'foo'`
5. **Let `v = m[k]`.  If `v == null`, the map does not match.**
6. **Otherwise, proceed to match `Object x` against `v`.**
7. *No type test on `v` since the matched value's static type (`int`) is a subtype of the variable pattern's required type (`Object`).*
8. Let `x = v`.

Interestingly, this "optimized" set of steps doesn't have the same soundness problem.  That seems bad.  We shouldn't rely on optional optimizations to keep programs sound.

munificent:
Fix #2685.

### DIFF
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -2750,32 +2750,42 @@ To match a pattern `p` against a value `v`:
 
         2.  Otherwise, evaluate `v[k]` to `r`.
 
-        3.  If `V` is non-nullable and `r` is `null` then the map does not
-            match.
+        3.  If `r != null || (null is V) && v.containsKey(k)` evaluates to
+            `false` then the map does not match.
 
-            *The `[]` operator on `Map` is always nullable. If we don't
-            explicitly check for `null` here, then a bad `Map` implementation
-            that returns `true` from `containsKey()` and then `null` from `v[k]`
-            for the same key would lead to a soundness hole when `V` is a
-            non-nullable type.*
+            *Note:*
+
+            *   *When `v[k]` returns a non-null value, we know the key is
+                present and we short-circuit the `containsKey()` call.*
+
+            *   *If `V` is known to be a non-nullable type, then `null is V` is
+                always `false` and the expression simplifies to:*
+
+                ```dart
+                r != null
+                ```
+
+            *   *Conversely, if `V` is known to be a nullable type, then `null
+                is V` is always `true` and the expression simplifies to:*
+
+                ```dart
+                r != null || v.containsKey(k)
+                ```
+
+            *   *When `V` is a potentially nullable type, the `null is V` test
+                must be performed but can be hoisted out and shared across all
+                entries since it doesn't depend on `k`.*
+
+            *   *If `v` is a poorly-behaved `Map` whose `v[k]` and
+                `containsKey(k)` results don't agree (i.e. a non-`null` `v[k]`
+                and `false` `containsKey(k)` or vice versa) we do not detect
+                that mismatch. Since badly behaved maps are rare, this is
+                allowed. Even if `v` is poorly-behaved, a `null` value won't be
+                passed to the subpattern unless `V` is a nullable or potentially
+                nullable type, so soundness is preserved.*
 
         4.  Else, match `r` against this entry's value subpattern. If it does
             not match, the map does not match.
-
-        A compiler is free to call `v[k]` and `containsKey()` in either order,
-        or to elide calling one or both if it determines that doing so will
-        produce the same result. It may assume that the map adheres to the
-        following protocol:
-
-        *   When `containsKey(k)` returns `false` for some key, then `v[k]` will
-            return `null`.
-
-        *   When `containsKey(k)` returns `true` for some key, then `v[k]`
-            returns an instance of the map's value type.
-
-        *In particular, if the map's value type is non-nullable, then when
-        `v[k]` returns `null`, the compiler can assume that the key is absent
-        and `containsKey(k)` would return `false` too.*
 
     5.  The match succeeds if all entry subpatterns match.
 
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -2837,13 +2837,13 @@ To match a pattern `p` against a value `v`:
                 must be performed but can be hoisted out and shared across all
                 entries since it doesn't depend on `k`.*
 
-            *   *If `v` is a poorly-behaved `Map` whose `v[k]` and
+            *   *If `v` is a poorly behaved `Map` whose `v[k]` and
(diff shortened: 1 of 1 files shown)


accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -2748,20 +2748,30 @@ To match a pattern `p` against a value `v`:
         1.  Evaluate the key `expression` to `k` and call `containsKey(k)` on
             the value. If this returns `false`, the map does not match.
 
-        2.  Otherwise, evaluate `v[k]` and match the resulting value against
-            this entry's value subpattern. If it does not match, the map does
-            not match.
+        2.  Otherwise, evaluate `v[k]` to `r`.
+
+        3.  If `V` is non-nullable and `r` is `null` then the map does not
```
leafpetersen, line 2753: This isn't enough, I think, because `V` may be neither non-nullable nor nullable (it may be unknown).

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -2748,20 +2748,30 @@ To match a pattern `p` against a value `v`:
...
         produce the same result. It may assume that the map adheres to the
         following protocol:
 
-        *   If `containsKey(k)` returns `false` for some key, then `v[k]` will
+        *   When `containsKey(k)` returns `false` for some key, then `v[k]` will
             return `null`.
 
-        *   If `containsKey(k)` returns `true` for some key, then `v[k]` returns
-            an instance of the map's value type.
+        *   When `containsKey(k)` returns `true` for some key, then `v[k]`
```
leafpetersen, line 2773: I think we're talking about getting rid of this, but note that this is incoherent with the above: if the compiler can assume this, it is free to elide the checks that you are specifying above.

stereotype441 (commented):
Given the discussion at https://github.com/dart-lang/language/issues/2685, I think this needs to be reworked.

accepted/future-releases/0546-patterns/feature-specification.md
munificent: You're right. Fixed.

accepted/future-releases/0546-patterns/feature-specification.md
munificent: Yeah, I got this wrong. Fixed.

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -2809,24 +2809,44 @@ To match a pattern `p` against a value `v`:
         1.  Evaluate the key `expression` to `k` and call `containsKey(k)` on
```
stereotype441, line 2809: This part needs to be changed to just:

1. Evaluate the key `expression` to `k`.

(We shouldn't call `containsKey(k)` pre-emptively)

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -2809,24 +2809,44 @@ To match a pattern `p` against a value `v`:
...
+                must be performed but can be hoisted out and shared across all
+                entries since it doesn't depend on `k`.*
+
+            *   *If `v` is a poorly-behaved `Map` whose `v[k]` and
+                `containsKey(k)` results don't agree (i.e. a non-`null` `v[k]`
+                and `false` `containsKey(k)` or vice versa) we do not detect
+                that mismatch. Since badly behaved maps are rare, this is
+                allowed. Even if `v` is poorly-behaved, a `null` value won't be
+                passed to the subpattern unless `V` is a nullable or potentially
+                nullable type, so soundness is preserved.*
```
stereotype441, line 2846: I agree that soundness is preserved, but I don't think this explanation really captures why it's preserved, because if `V` is a potentially nullable type, it's only safe to pass `null` to the subpattern if `null is V` is `true`.

I'd recommend changing this last sentence to something like:

*Even if `v` is poorly-behaved, a `null` value will only be passed to the subpattern if `null is V`, so soundness is preserved.*

stereotype441 (approved):
lgtm assuming you agree with my suggested changes

leafpetersen (approved):

accepted/future-releases/0546-patterns/feature-specification.md
munificent: Done.

PR #2691, opened 2022-12-06
[patterns] Clarify when map key primitive `==` comes into play.

### LINKED ISSUE #2690: Map patterns error only applicable in one case: Records
Thanks to @sgrekhov for bringing up this issue. Consider the error specified [here](https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#:~:text=if%20any%20two%20keys%20in%20the%20map%20both%20have%20primitive%20%3D%3D%20methods%2C%20then%20it%20is%20a%20compile-time%20error%20if%20they%20are%20equal%20according%20to%20their%20%3D%3D%20operator.):

> If any two keys in the map both have primitive `==` methods, then it is a
> compile-time error if they are equal according to their `==` operator.

I think this rule is subsumed by the previous one:

> [It's a compile-time error] If any two keys in the map are identical.

The keys must have primitive `==`, so they are instances of `int`, `String`, `Symbol`, reified `Type` instances, instances of `Object`, `null`, instances of a class that inherits `==` from `Object`, or a record where every non-`Object` getter has a type that has a primitive `==`.

Also, they are map pattern keys, so the expressions denoting them must be constant. For every one of these, it is then guaranteed that `identical` returns true when `==` returns true (`null` is unique; it's specified directly for `int` and `String`; it is specified in terms of canonicalization for `Symbol`, reified types, and instances of the classes that inherit `==` from `Object`).

This means that the only case with map pattern keys where it isn't true that "`==` implies `identical`" is records.

So I believe we should change the second bullet such that it becomes a rule about records, and nothing else (for clarity).

[Edit: Added records.]

munificent:
Fix #2690.

### DIFF
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: Accepted
 
-Version 2.19 (see [CHANGELOG](#CHANGELOG) at end)
+Version 2.20 (see [CHANGELOG](#CHANGELOG) at end)
 
 Note: This proposal is broken into a couple of separate documents. See also
 [records][] and [exhaustiveness][].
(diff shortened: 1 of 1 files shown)


accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -595,27 +595,29 @@ It is a compile-time error if:
...
+    records don't have defined identity, we can't use the previous rule to
+    detect identical records. But records do support an equality test known at
+    compile time if all of their fields do, so we use that.*
 
 *   There is more than one `...` element in the map pattern.
 
 *   The `...` element is not the last element in the map pattern.
 
+*Note that we don't require map keys to have primitive `==` methods to enable
+more flexibility in key typed. If the keys have user-defined `==` methods, then
```
eernstg, line 617: Maybe?:
```suggestion
more flexibility in key types. If the keys have user-defined `==` methods, then
```

eernstg (approved):
LGTM, just one likely typo!

PR #2696, opened 2022-12-08
Adjust definition of flatten

eernstg:
Adjust definition of flatten to ensure that `flatten(X) == X` when `X` is a type variable which is unrelated to future types. It deletes the case which is explicitly dealing with type variables, and relies on the existing cases where the given type is related to `FutureOr` or `Future`, but adds one thing: The `FutureOr` case is applicable to types that are _bounded_ by `FutureOr<S>` for some `S`, not just the ones that _are_ `FutureOr<S>`, and this includes `X extends FutureOr<S>`.


lrhn (approved):
Ship it!

PR #2702, opened 2022-12-09
Ensure list patterns are exhaustive even with negative length lists.

### LINKED ISSUE #2669: Invocation key of list subpatterns regarding rest potential mistake
Link:
https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md

> Let `e` be `i : ("tail[]", [index])` where `index` is the zero-based index of this element subpattern.
> 
> *Note the "tail" in the invocation key name. This is to distinguish elements after a rest element at some position from elements at the same position but not following a rest element, as in:*
> 
> ````dart
> switch (list) {
>   case [var a, ..., var c]: ...
>   case [var a, _,   var d]: ...
> }
> ````
> *Here, `c` and `d` may have different values and `d` should not use the previously cached value of `c` even though they are both the third element of the same list. So we use an invocation key of "[]" for `c` and "tail[]" for `d`*.

If I understand this correctly, then shouldn't the last sentence be reversed? Since `c` proceeds a rest, it should have "tail[]" and `d` be the regular "[]"?

### LINKED ISSUE #2701: Exhaustiveness on list patterns with negative lengths
@stereotype441 mentioned this in a meeting and I wanted to file an issue so we don't lose track. This appears like it should be an exhaustive switch:

```dart
String test(List<int> list) => switch (list) {
  [] => 'empty',
  [_, ...] => 'not empty'
};
```

However, it's possible for neither case to match if you pass it a misbehaving implementation of `List` that returns a *negative* number for `length`.

Exhaustiveness needs to be sound. At the same time, I don't want users to have to write clearly pointless default cases when switching on lists just to handle misbehaving list implementations that will never actually reach their code. I can think of a couple of ways to handle this:

* When a list pattern calls `length` on a list, if a negative number is returned, throw a "bad list protocol" exception.
* Treat any length of 0 or less as equivalent to zero. That means an empty list pattern, or a list pattern containing only a rest element, will match a list of length 0 or less.

I lean slightly towards the first option because I like letting users known when something fishy is going on. But I could be convinced to do the latter if it's more efficient or has other desirable properties. In practice, almost no users will ever hit this case.

Thoughts?

munificent:
Fix #2701.
Fix #2669.

### DIFF
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -2757,14 +2757,14 @@ To match a pattern `p` against a value `v`:
         1.  Let `l` be the length of the list determined by calling `length` on
             `v`. *We only call `length` on the list if needed.*
 
-        2.  If `p` has a rest element and `h + t > 0`:
+        2.  If `p` has both a rest element and `h + t > 0`:
 
             1.  If `l < h + t` then the match fails.
 
             *When there are non-rest elements and a rest element, the list must
             be at least long enough to match the non-rest elements.*
 
-        3.  Else if `h + v > 0` *(and `p` has no rest element)*:
+        3.  Else if `h + t > 0` *(and `p` has no rest element)*:
 
             1.  If `l != h + t` then the match fails.
 
@@ -2776,7 +2776,7 @@ To match a pattern `p` against a value `v`:
             1.  If `l > 0` then the match fails.
 
             *An empty list pattern can match only empty lists. Note that this
-            treats misbehaving a list whose `length` is negative as an empty
+            treats a misbehaving list whose `length` is negative as an empty
             list. This is important so that a set of list patterns that is
             clearly exhaustive over well-behaving lists will also cover a
             misbehaving one.*
@@ -2826,15 +2826,41 @@ To match a pattern `p` against a value `v`:
         *This type test may get elided. See "Pointless type tests and legacy
         types" below.*
 
-    2.  Let `l` be the length of the map determined by calling `length` on `v`.
+    2.  Let `n` be the number of non-rest elements.
 
-    3.  If `p` has no rest element and `l` is not equal to the number of
-        subpatterns then the match fails. If `p` has a rest element and `l` is
-        less than the number of non-rest entry subpatterns, then the match
-        fails. *These match failures become runtime exceptions if the map
-        pattern is in an irrefutable context.*
+    3.  Check the length. If `p` is empty or has any non-rest elements:
 
-    4.  Otherwise, for each (non-rest) entry in `p`, in source order:
+        1.  Let `l` be the length of the map determined by calling `length` on
+            `v`. *We only call `length` on the map if needed.*
+
+        2.  If `p` has both a rest element and `n > 0`:
+
+            1.  If `l < n` then the match fails.
+
+            *When there are non-rest elements and a rest element, the map must
+            be at least long enough to match the non-rest elements.*
+
+        3.  Else if `n > 0` *(and `p` has no rest element)*:
+
+            1.  If `l != n` then the match fails.
+
+            *If there are only non-rest elements, then the map must have exactly
+            the same number of elements.*
+
+        4.  Else `p` is empty:
+
+            1.  If `l > 0` then the match fails.
+
+            *An empty map pattern can match only empty maps. Note that this
+            treats a misbehaving map whose `length` is negative as an empty map.
+            This is important so that a set of map patterns that is clearly
+            exhaustive over well-behaving maps will also cover a misbehaving
+            one.*
+
+        *These match failures become runtime exceptions if the map pattern is
+        in an irrefutable context.*
+
+    4.  For each non-rest entry in `p`, in source order:
 
         1.  Evaluate the key `expression` to `k` and call `containsKey(k)` on
             the value. If this returns `false`, the map does not match.
@@ -3286,7 +3312,7 @@ Here is one way it could be broken down into separate pieces:
 
 ### 2.21
 
--   Handle negative length lists (#2701).
+-   Handle negative length lists and maps (#2701).
 
 ### 2.20
 
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -2746,50 +2746,75 @@ To match a pattern `p` against a value `v`:
         *This type test may get elided. See "Pointless type tests and legacy
         types" below.*
 
-    2.  Let `l` be the length of the list determined by calling `length` on `v`
-        or `0` if `length` returns a negative number. *We treat misbehaving
-        `List` implementations that yield negative lengths as equivalent to
-        empty lists. That way, a set of list patterns that covers all
-        non-negative lengths is soundly exhaustive even when the list
-        misbehaves.*
-
-    3.  Let `h` be the number of non-rest elements preceding the rest element if
+    2.  Let `h` be the number of non-rest elements preceding the rest element if
         there is one, or the number of elements if there is no rest element.
 
-    4.  Let `t` be the number of non-rest elements following the rest element if
+    3.  Let `t` be the number of non-rest elements following the rest element if
         there is one, or `0` otherwise.
 
-    5.  If `p` has no rest element and `l` is not equal to `h` then the match
-        fails. If `p` has a rest element and `l` is less than `h + t` then the
-        match fails. *These match failures become runtime exceptions if the list
-        pattern is in an irrefutable context.*
+    4.  Check the length. If `p` is empty or has any non-rest elements:
 
-    6.  Match the head elements. For `i` from `0` to `h - 1`, inclusive:
+        1.  Let `l` be the length of the list determined by calling `length` on
+            `v`. *We only call `length` on the list if needed.*
 
-        1.  Extract the element value `e` by calling `[]` on `v` with index `i`.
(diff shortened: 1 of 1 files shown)


accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -2746,26 +2746,31 @@ To match a pattern `p` against a value `v`:
         *This type test may get elided. See "Pointless type tests and legacy
         types" below.*
 
-    2.  Let `l` be the length of the list determined by calling `length` on `v`.
+    2.  Let `l` be the length of the list determined by calling `length` on `v`
+        or `0` if `length` returns a negative number. *We treat misbehaving
```
stereotype441, line 2750: This works but I think we can do better in terms of runtime efficiency.  Personally I'd rather leave this step as:

2. Let `l` be the length of the list determined by calling `length` on `v`.

And then insert before the next step:

3. If the list pattern has no elements, and `l <= 0`, the match succeeds.  Otherwise...

(and leave the remaining text unchanged, other than renumbering of course).

This is indistinguishable from what you specified, and more efficient because it just replaces an `== 0` check with a `<= 0` check when the list pattern is `[]`, rather than adding a `< 0` check to *every* list pattern.

(Yes, I realize that implementers are free to implement something different from what we specify if the results are indistinguishable, but I think we should give them a good starting point.)

leafpetersen (approved):

accepted/future-releases/0546-patterns/feature-specification.md
munificent: > 3\. If the list pattern has no elements, and `l <= 0`, the match succeeds.

What about:

```dart
class NegativeLengthList<E> implements List<E> { ... }

main() {
  if (NegativeLengthList<String>() case [...]) {
    print('?');
  }
}
```

Does that match? An even more fun one:

```dart
main() {
  if (NegativeLengthList<String>() case [...var rest]) {
    print(rest);
  }
}
```

Should that throw an exception because it passes an invalid negative argument to `sublist()`, or should it truncate the negative length to zero and then call `sublist(0, 0)`?

lrhn: Should `NegativeLengthList() case [...]` match? Yes.
There should be no length check on the `[...]` pattern at all.

A list pattern *p*'s check could start by something like:

* Let *n* be the number of non-spread patterns in *p*
* If *p* contains a spread pattern, then:
    * If *n* > 0, check that *listLength* >= *n*
* Otherwise (no spread)
    * if *n* > 0, check that *listLength* == *n*
    * otherwise check that *listLength* >= 0

A `[...]` pattern would go through that without needing to check `listLength` at all, and `[]` would check length >= 0.

Also, for a tailing spread pattern with a match pattern, `[p1, p2, ... p3]`,
the `sublist` call should not pass an `end` argument, because it goes to the end of the list.
That makes `case [...l] => l` (list copy, the weird way!) not need the list length at all.

I don't care if we specify *listLength* as `let tmp = list.length in tmp < 0 ? 0 : tmp`.
A compiler can optimize that into the behavior above, if we make sure there is no way to see that value if it's not greater than zero anyway. 
(Which is why `[...p]` should do `list.sublist(0)`, whereas `[...p, q]` requires length > 0 anyway, so `sublist(0, list.length - 1)` will be fine.)

stereotype441: > > 3. If the list pattern has no elements, and `l <= 0`, the match succeeds.
> 
> What about:
> 
> ```dart
> class NegativeLengthList<E> implements List<E> { ... }
> 
> main() {
>   if (NegativeLengthList<String>() case [...]) {
>     print('?');
>   }
> }
> ```

Oops, you're right.

> Does that match? An even more fun one:
> 
> ```dart
> main() {
>   if (NegativeLengthList<String>() case [...var rest]) {
>     print(rest);
>   }
> }
> ```
> 
> Should that throw an exception because it passes an invalid negative argument to `sublist()`, or should it truncate the negative length to zero and then call `sublist(0, 0)`?

Yeah, good corner cases.  I'm glad you didn't blindly accept my suggestion 😃 

I like the approach @lrhn is suggesting.  We need to spell out all the details, but as I understand it, it basically boils down to:
- Assume that the list is well-behaved when deciding what operations to perform on it (including not calling `length` when handling a final `...`). 
- When we actually perform those operations, instead of computing `x.length`, compute `max(0, x.length)`.  *(this is equivalent to @lrhn's suggestion `let tmp = list.length in tmp < 0 ? 0 : tmp`)*.
- Optimize where appropriate (e.g. `max(0, x.length) == 0` becomes `x.length <= 0` and `max(0, x.length) == 3` becomes `x.length == 3`).

This feels very consistent with what we decided to do for ill-behaved maps.  The magic there was coming up with a definition for the `containsKey` operation that had all the consistency properties we wanted and was easily optimizable; here we are doing the same magic on `length`.  Not surprising that @lrhn had both ideas 😃

munificent: OK, how does the latest commit look?

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -2746,45 +2746,75 @@ To match a pattern `p` against a value `v`:
...
+            `v`. *We only call `length` on the list if needed.*
+
+        2.  If `p` has a rest element and `h + t > 0`:
+
+            1.  If `l < h + t` then the match fails.
+
+            *When there are non-rest elements and a rest element, the list must
+            be at least long enough to match the non-rest elements.*
+
+        3.  Else if `h + v > 0` *(and `p` has no rest element)*:
```
lrhn, line 2767: v -> t

lrhn (approved):
LGTM.

Now do maps :stuck_out_tongue_closed_eyes:

munificent:
Did maps too, heh. :) PTAL.

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -2746,45 +2746,75 @@ To match a pattern `p` against a value `v`:
...
-        there is one, or zero otherwise.
+    3.  Let `t` be the number of non-rest elements following the rest element if
+        there is one, or `0` otherwise.
+
+    4.  Check the length. If `p` is empty or has any non-rest elements:
+
+        1.  Let `l` be the length of the list determined by calling `length` on
+            `v`. *We only call `length` on the list if needed.*
+
+        2.  If `p` has both a rest element and `h + t > 0`:
```
stereotype441, line 2760: Nit: to me, the word "both" here is confusing.  It seems to imply that I'm supposed to read what follows as two things that `p` should have.  As in, "If `p` has a rest element and `p` has `h + t > 0`", which doesn't make any sense.

I think it reads better as just "If `p` has a rest element and `h + t > 0`"

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -2796,15 +2826,41 @@ To match a pattern `p` against a value `v`:
...
-        less than the number of non-rest entry subpatterns, then the match
-        fails. *These match failures become runtime exceptions if the map
-        pattern is in an irrefutable context.*
+    3.  Check the length. If `p` is empty or has any non-rest elements:
 
-    4.  Otherwise, for each (non-rest) entry in `p`, in source order:
+        1.  Let `l` be the length of the map determined by calling `length` on
+            `v`. *We only call `length` on the map if needed.*
+
+        2.  If `p` has both a rest element and `n > 0`:
```
stereotype441, line 2836: Same comment about the use of "both" here.

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -2796,15 +2826,41 @@ To match a pattern `p` against a value `v`:
...
+            be at least long enough to match the non-rest elements.*
+
+        3.  Else if `n > 0` *(and `p` has no rest element)*:
+
+            1.  If `l != n` then the match fails.
+
+            *If there are only non-rest elements, then the map must have exactly
+            the same number of elements.*
+
+        4.  Else `p` is empty:
```
stereotype441, line 2850: Similar to the issue above, we need to cover the case where `p` is `{...}`.

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -2746,45 +2746,75 @@ To match a pattern `p` against a value `v`:
...
-        match fails. *These match failures become runtime exceptions if the list
-        pattern is in an irrefutable context.*
+            1.  If `l != h + t` then the match fails.
 
-    4.  Match the head elements. For `i` from `0` to `h - 1`, inclusive:
+            *If there are only non-rest elements, then the list must have
+            exactly the same number of elements.*
 
-        1.  Extract the element value `e` by calling `[]` on `v` with index `i`.
+        4.  Else `p` is empty:
```
stereotype441, line 2774: Whoops, I tried to comment on this line earlier but I guess I forgot to click the button and my draft got deleted.

We're not covering the case where `p` is `[...]`.  I think we could fix this by saying:

4.  Else *(`p` is empty or contains just a single "rest" pattern)*:
    1. If `l > 0` and `p` is empty, then the match fails.

stereotype441: Better still, swap the order and make it "If `p` is empty and `l > 0`", that way it's clear that if `p` is `[...]`, we don't do a length check.

stereotype441 (approved):
lgtm

PR #2705, opened 2022-12-09
Add note about renaming to "inline class"

eernstg:
Hi Michael, I think it is important to include a note into the Views feature specification about the renaming to 'inline classes'. In particular, the Flutter@Google newsletter included a link to the views feature specification. This PR does that, and adds a link to the PR (to be updated to the landed document when landed). WDYT?


mit-mit (approved):

eernstg:
Thanks!

PR #2713, opened 2022-12-11
Update flatten definition

eernstg:
The definition of flatten needed yet another revision in order to cover intersection types properly.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -11276,34 +11276,55 @@ where inferred types have already been added.%
 }
 
 \LMHash{}%
+We define the
+\IndexCustom{future type of a type}{type!future type of}
+$T$ as follows:
+
+\begin{itemize}
+\item If $T$ implements \code{Future<$S$>}
+  (\ref{interfaceSuperinterfaces})
+  then the future type of $T$ is \code{Future<$S$>}.
+\item Otherwise, if $T$ is \code{FutureOr<$S$>} bounded
+  (\ref{bindingActualsToFormals})
+  then the future type of $T$ is \code{FutureOr<$S$>}.
+\item Otherwise, if $T$ is \code{Future<$S$>?} bounded
+  or \code{FutureOr<$S$>?} bounded
+  then the future type of $T$ is \code{Future<$S$>?}
+  respectively \code{FutureOr<$S$>?}.
+\item Otherwise, $T$ has no future type.
+\end{itemize}
+
+\LMHash{}%
 We define the auxiliary function
 \IndexCustom{\flatten{T}}{flatten(t)@\emph{flatten}$(T)$},
 which is used below and in other sections, as follows:
 
 \begin{itemize}
-\item{} If $T$ is \code{$S$?}\ bounded
-  (\ref{bindingActualsToFormals})
-  for some $S$ then \DefEquals{\flatten{T}}{\code{\flatten{S}?}}.
+\item If $T$ is \code{$S$?}\ for some $S$
+  then \DefEquals{\flatten{T}}{\code{\flatten{S}?}}.
 
-\item{} If $T$ is \code{FutureOr<$S$>} bounded
+\item Otherwise, if $T$ is \code{$X$\,\,\&\,\,$S$}
+  for some type variable $X$ and type $S$ then
+  \begin{itemize}
+  \item if $S$ has future type $U$
+    then \DefEquals{\flatten{T}}{\code{\flatten{U}}}.
+  \item otherwise,
+    \DefEquals{\flatten{T}}{\code{\flatten{X}}}.
+  \end{itemize}
+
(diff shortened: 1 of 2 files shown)


specification/dartLangSpec.tex
```diff
@@ -11275,35 +11275,56 @@ \subsection{Function Expressions}
...
 
-\item{} If $T$ implements \code{Future<$S$>}
-  (\ref{interfaceSuperinterfaces})
+\item Otherwise, if $T$ has future type \code{Future<$S$>}
+  or \code{FutureOr<$S$>}
   then \DefEquals{\flatten{T}}{S}.
 
-\item{} Otherwise, \DefEquals{\flatten{T}}{T}.
+\item Otherwise, if $T$ has future type \code{Future<$S$>?}\ or
+  \code{FutureOr<$S$>?}\ then \DefEquals{\flatten{T}}{\code{$S$?}}.
```
lrhn, line 11320: Can these two be combined into;

* If `T` implements `Future<S>` for some `S`, _flatten_(`T`) = `S`.
* Otherwise, if `T` has future type `F`, then _flatten_(`T`) = _flatten_(`F`).

and then we rely on all the possible kinds of future types being base-cases for _flatten_?

But that's not much shorter, because what you have here is basically (if we don't use `FutureOr` for future types):

* If `T` has future type `F`:
    * if `F` is `Future<S>` for some `S`, _flatten_(`T`) = `S`
    * if `F` is `Future<S>?` for some `S`, _flatten_(`T`) = `S?`

and that's short too.

Either should be correct, and give the same result.

specification/dartLangSpec.tex
```diff
@@ -16023,8 +16042,33 @@ \subsection{Await Expressions}
 (\ref{expressionEvaluation}).
 If $f$ completes with an object $v$, $a$ evaluates to $v$.
 
-% Otherwise, the value of $a$ is the value of $e$. If evaluation of
-% $e$ raises an exception $x$, $a$ raises $x$.
+\rationale{%
+This use of \flattenName{} implies that we await a future
+in every case where this choice is sound.%
```
lrhn, line 16047: Not clear which use of _flatten_ this refers to.

It's actually the use of the type `T`, derived from _flatten_, in the "is a subtype of `Future<T>`" which implies that we await as much as possible.

So, maybe:

> Using `Future<T>` as the check for which futures to await implies that
> we await every future that we can soundly await.

Or add it below in the commentary instead.

lrhn (approved):

specification/dartLangSpec.tex
eernstg: The kind of simplicity I've strived for is the following: These rules specify a particular type which is being searched in a nominal-supertype hierarchy ('implements' in general, 'bounded' in the case where we can't have any interfacetype-superinterface edges in the graph), and then it directly gives the result of the search.

The simplification that I'd very much like to have is to say "if `T` has nominal-supertype `T0` then the future type is `T0`" without duplicating `T0`.

specification/dartLangSpec.tex
eernstg: I think this is clearer now.

specification/dartLangSpec.tex
```diff
@@ -16023,8 +16042,33 @@ \subsection{Await Expressions}
...
+so it \emph{must} be considered to be in the second group.
+Nevertheless, \flatten{\code{FutureOr<Object>?}} is \code{Object?},
+so we \emph{will} await a \code{Future<Object?>}.
+We have chosen this semantics because it was the smallest breaking change
+relative to the semantics in earlier versions of Dart,
+and also because it allows for a simple rule:
+The type of \code{\AWAIT\,\,$e$} is used to decide whether or not
+the future (if any) is awaited, and there are no exceptions---even
+in cases like this example, where the type seems to imply that
+a \code{Future<Object?>} should not be awaited.%
```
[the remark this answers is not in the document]

specification/dartLangSpec.tex
```diff
@@ -11275,35 +11275,56 @@ \subsection{Function Expressions}
...
+\IndexCustom{future type of a type}{type!future type of}
+$T$ as follows:
+
+\begin{itemize}
+\item If $T$ implements \code{Future<$S$>}
+  (\ref{interfaceSuperinterfaces})
+  then the future type of $T$ is \code{Future<$S$>}.
+\item Otherwise, if $T$ is \code{FutureOr<$S$>} bounded
+  (\ref{bindingActualsToFormals})
+  then the future type of $T$ is \code{FutureOr<$S$>}.
```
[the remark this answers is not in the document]
eernstg: I managed to make the definition of 'the future type of a type' considerably more compact and non-redundant.

specification/dartLangSpec.tex
eernstg: Did that!

eernstg:
@lrhn, I landed this. Let's take another iteration in a new PR, if needed.

PR #2715, opened 2022-12-12
[Patterns] Fixed typo in the specification

sgrekhov:
Object pattern grammar is
```
objectPattern ::= typeName typeArguments? '(' patternFields? ')'
```
So, there should be a `typeName`, not an `objectName`


eernstg (commented):
LGTM. The grammar rule for `objectPattern` already uses `typeName`.

@munificent, do you have further comments?

munificent (approved):
Thanks!

PR #2722, opened 2022-12-14
Moved sealed-types and base-interface-final to accepted/

munificent:
Also move competing packaged-libraries and type-modifiers proposals to inactive/.


leafpetersen (approved):

accepted/future-releases/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,596 @@
...
+
+Status: Accepted
+
+Version 1.0
+
+This proposal specifies three modifiers that can be placed on classes and mixins
+to allow an author to control whether the type allows being implemented,
+extended, both, or neither. The proposed modifiers are:
+
+*   No modifier: As today, the type has no restrictions.
```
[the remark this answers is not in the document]
lrhn: It says "three modifiers" above, which are `base`, `interface` and `final`, so presumably `mixin` is not a modifier.

accepted/future-releases/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,596 @@
...
+The [type modifiers][] document has some motivation and discussion around
+defaults and keyword choice which may be a useful reference. Unlike that
+proposal, this proposal is non-breaking.
+
+[type modifiers]: https://github.com/dart-lang/language/blob/master/working/type-modifiers/feature-specification.md
+["closed" and "base"]: https://github.com/dart-lang/language/issues/2595
+https://github.com/dart-lang/language/issues/2595
+[leaf proposal]: https://github.com/dart-lang/language/issues/2595
+
+This proposal also allows `mixin` to be used as a modifier on `class` to
```
lrhn, line 34: I would change that to not call `mixin` a modifier at all.
Instead:

> This proposal also allows a declaration to be both a `mixin` and a `class` declaration at the same time.
> A `mixin class` declaration is explicitly intended t o
> be used as both a class and a mixin,
> and must be valid as both.

accepted/future-releases/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,596 @@
...
+  '{' (metadata classMemberDeclaration)* '}'
+
+classModifiers ::= ( 'sealed' | 'abstract'? typeModifier? ) 'mixin'?
+mixinModifier ::= 'sealed' | typeModifier
+typeModifier ::= 'base' | 'interface' | 'final'
+```
+
+### Static semantics
+
+It is a compile-time error to:
```
leafpetersen, line 539: The syntax allows the modifiers to be applied to a `mixin` declaration, but there are no compile time errors specified for mixins below.  Presumably there is something intended here, but I don't actually know what.

accepted/future-releases/type-modifiers/feature-specification.md
```diff
@@ -1,11 +1,13 @@
-# "Base", "interface", and "final" types
+# Type modifiers
```
leafpetersen, line 1: Mmm.  I'd really prefer we not use this terminology.  Classes are not types.  Classes *produce* types (among other things).  That is, a class declaration `class C` introduces a type named `C`, and it also introduces a bundle of code, with a namespace, some semantics, etc.  When we say `class C extends D`, `D` here is not the *type* `D`, it is the bundle of stuff that we also happen to name (in the surface syntax) `D`.   "Type modifiers" suggest things that I can use to modify types (e.g. `sealed int x = 3`), which here declares a variable `x` of type `sealed x`, whatever that might be.  Perhaps "declaration modifiers"?  Or just stick with "class"?

accepted/future-releases/base-interface-final/feature-specification.md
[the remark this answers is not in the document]
leafpetersen: It is a something, and it is not the case that `class` works "as today" after this proposal.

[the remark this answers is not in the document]
munificent: I just didn't do a thorough job of going through the proposal and updating other stuff when I added `mixin`. Fixed.

accepted/future-releases/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,596 @@
...
+
+This proposal specifies three modifiers that can be placed on classes and mixins
+to allow an author to control whether the type allows being implemented,
+extended, both, or neither. The proposed modifiers are:
+
+*   No modifier: As today, the type has no restrictions.
+*   `base`: The type can be extended (if a class) or mixed in (if a mixin) but
+    not implemented.
+*   `interface`: The type can be implemented but not extended or mixed in.
+*   `final`: The type can't be extended, mixed in, or implemented.
```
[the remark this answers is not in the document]
munificent: Done (and rewrote).

accepted/future-releases/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,596 @@
...
+
+This proposal is a blend of three earlier proposals:
+
+* [Type modifiers][]
+* ["Closed" and "base"][]
+* [Access modifiers using closed, sealed, open and interface][leaf proposal]
```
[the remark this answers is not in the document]
munificent: Weird. Tweaked the links to try to fix.

accepted/future-releases/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,596 @@
...
+proposal, this proposal is non-breaking.
+
+[type modifiers]: https://github.com/dart-lang/language/blob/master/working/type-modifiers/feature-specification.md
+["closed" and "base"]: https://github.com/dart-lang/language/issues/2595
+https://github.com/dart-lang/language/issues/2595
+[leaf proposal]: https://github.com/dart-lang/language/issues/2595
+
+This proposal also allows `mixin` to be used as a modifier on `class` to
+indicate that a type is explicitly intended to be used as both a class and a
+mixin.
```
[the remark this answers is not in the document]
munificent: Done.

accepted/future-releases/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,596 @@
...
+class UsesAsSuperclass extends Both {}
+
+class UsesAsMixin extends OtherSuperclass with Both {} // OK.
+```
+
+This change is guarded by a language version. Any class declaration in a library
+whose language version is before the one this feature ships in can be used as a
+mixin as long as the class meets the mixin restrictions. Classes in libraries
+whose version is later than that must be explicitly marked `mixin class` to
+allow the class to be used in a `with` clause.
```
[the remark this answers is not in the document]
munificent: Took a stab at clarifying this. (I haven't checked the issues yet, but I'll circle back to those next.)

accepted/future-releases/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,596 @@
...
+Some combinations don't make sense:
+
+*   `sealed` implies `abstract`, so they can't be combined.
+*   `sealed` implies non-extensibility, so can't be combined with `interface`
+    or `final`.
+*   `sealed` implies non-implementability, so can't be combined with `base` or
+    `final`.
+*   `base`, `interface`, and `final` all control the same two capabilities so
+    are mutually exclusive.
+*   `mixin` as a modifier can obviously only be applied to a class. It can be
```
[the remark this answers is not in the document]
munificent: OK, I went through all of the combinations and tried to think through what they mean and whether they're useful. 

* I agree `final mixin class` doesn't make sense. The `mixin` means "you can mix me in externally" but the `final` means "you can't use me as a superclass" and those are in conflict. Also agreed that `interface mixin class` is pointless for the same reason.

* `base mixin class` works out, I think. It doesn't have an interface you can implement, so it carries the same restrictions (now fully specified) that any class mixing it in or extending it also has to be marked `base` or `final`.
*

accepted/future-releases/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,596 @@
...
+abstract final mixin class
+```
+
+</td><td>
+
+```
+mixin
+sealed mixin
+base mixin
+interface mixin
```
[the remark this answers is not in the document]
munificent: We've been thinking of the two capabilities that `base`/`interface`/`final` controls as:

* Can be extended 
* Can be implemented

After thinking through the invariants we care about and the combinations users might want to express, I think it makes more sense to think of the two capabilities as:

* Can be used as a superclass
* Can be implemented

When a type is marked `final` or `interface`, it loses the ability to be used as a superclass. That means:

* For a `class`, it can't be extended.
* For a `mixin`, it can't be mixed in.

At first, I thought it was stupid to have a mixin declaration that you can't mixin, but it still be mixed in *inside* the library, so I think it's useful to be able to express "mixin that can be implemented but not mixed in" and "mixin that can't be implemented or mixed in". Treating the capability as "can be used as a superclass" lets us express that with the existing modifiers:

```
mixin // Can mix in and implement.
base mixin // Can mix in but not implement.
interface mixin // Can implement but not mix in.
final mixin // Can't implement or mix in.
```

Then there's the question of `mixin class`. For those, I think it simply doesn't make sense to combine a modifier that takes away subclassing or mixin in. `final mixin class` doesn't make sense because, as you note, you could just do `extends Object with` instead. And if you want to prevent both extending and using it as a mixin... then you shouldn't have put `mixin` on it.

accepted/future-releases/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,596 @@
...
+  ('on' typeNotVoidList)? interfaces?
+  '{' (metadata classMemberDeclaration)* '}'
+
+classModifiers ::= ( 'sealed' | 'abstract'? typeModifier? ) 'mixin'?
+mixinModifier ::= 'sealed' | typeModifier
+typeModifier ::= 'base' | 'interface' | 'final'
+```
+
+### Static semantics
+
```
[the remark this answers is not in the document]
munificent: I'll follow up on that issue next, but I wanted to go through the proposal and the various combinations to try to rationalize it all in my head first.

accepted/future-releases/base-interface-final/feature-specification.md
munificent: Fixed.

accepted/future-releases/type-modifiers/feature-specification.md
munificent: I'm a little worried that it might be confusing that the proposal applies to mixins too if we call it "class modifiers". "Declaration modifiers" is a little vague: I think of variables and methods as declarations too and I specifically didn't call this proposal "capability modifiers" to scope it to just type declarations and not other kinds of declarations.

"Class and mixin modifiers" is a mouthful. I wanted something relative short so we can also use it for the experiment flag. "Class modifiers" is probably the least bad.

@kallentu @bwilkerson, thoughts on the name?

accepted/future-releases/type-modifiers/feature-specification.md
```diff
@@ -426,13 +431,10 @@ without *intending* it to be used as a mixin and then accidentally forbid that
...
 breaking change to any downstream user that had that class in a `with` clause.
 
-Eventually, we would like classes to not be usable as a mixin by default. Being
-extensible and implementable seems to be the right default for classes based on
-Dart's history and talking to users. But being usable as a mixin by default is
-rarely (but not *never*) useful and often confusing. It makes classes brittle
-for little upside.
-
-Instead, under this proposal we require authors to explicitly opt in to allowing
+Using a class as a mixin rarely useful, but it is sometimes, so we don't want to
```
leafpetersen, line 434: "is rarely"

accepted/future-releases/type-modifiers/feature-specification.md
```diff
@@ -447,114 +449,108 @@ class UsesAsSuperclass extends Both {}
...
+|`sealed class`             |No     |No     |No     |No     |**Yes**|
+|`abstract class`           |No     |**Yes**|**Yes**|No     |No     |
+|`abstract base class`      |No     |**Yes**|No     |No     |No     |
+|`abstract interface class` |No     |No     |**Yes**|No     |No     |
+|`abstract final class`     |No     |No     |No     |No     |No     |
+|`mixin class`              |**Yes**|**Yes**|**Yes**|**Yes**|No     |
+|`base mixin class`         |**Yes**|**Yes**|No     |**Yes**|No     |
+|`abstract mixin class`     |No     |**Yes**|**Yes**|**Yes**|No     |
+|`abstract base mixin class`|No     |**Yes**|No     |**Yes**|No     |
+|`mixin`                    |No     |No     |**Yes**|**Yes**|No     |
```
leafpetersen, line 497: This is pretty much my proposal at the bottom of [this comment](https://github.com/dart-lang/language/issues/2723#issuecomment-1354093842).  If we allow extending a normal `mixin`, then the two `abstract ... mixin class` combinations also become redundant.

accepted/future-releases/type-modifiers/feature-specification.md
```diff
@@ -447,114 +449,108 @@ class UsesAsSuperclass extends Both {}
...
+*   Mix in a mixin marked `interface` or `final` outside of the library where it
+    is declared.
+
+*   Extend a class marked `base` outside of the library where it is declared
+    unless the extending class is marked `base` or `final`. *This ensures that a
     subtype can't escape the `base` restriction of its supertype by offering its
     _own_ interface that could then be implemented without inheriting the
     concrete implementation from the supertype.*
 
+*   Mix in a mixin marked `base` or `final` outside of the library where it is
```
leafpetersen, line 541: You can't mix in a mixin marked `final` outside of the library where it is declared, right?

accepted/future-releases/type-modifiers/feature-specification.md
```diff
@@ -447,114 +449,108 @@ class UsesAsSuperclass extends Both {}
...
+    declared unless the class mixing it in is marked `base` or `final`. *As with
+    the previous rule, ensures you can't get a backdoor interface on a mixin
+    that doesn't want to expose one.*
+
+*   Mix in a mixin class marked `base` outside of the library where it is
```
leafpetersen, line 546: This is the same as the previous?

leafpetersen: Never mind, `mixin` vs `mixin class`.

accepted/future-releases/base-interface-final/feature-specification.md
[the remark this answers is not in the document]
lrhn: Here I'd say that the `final` applies to both the `mixin` declaration and the `class` declaration of the `mixin class`. It can neither be implemented or used as a superclass outside of the library.
It can be used as either inside the library.

That only makes sense if you cannot mix in a non-`mixin` `class` inside the same library, which I'm also proposing.

accepted/future-releases/base-interface-final/feature-specification.md
```diff
@@ -0,0 +1,596 @@
...
+abstract base class
+abstract interface class
+abstract final class
+```
+
+</td><td>
+
+```
+mixin class
+sealed mixin class
```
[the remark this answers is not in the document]
lrhn: The restriction stands. You cannot *subtype* a sealed class outside of the library.
Inside the library, you can extend, mix in or implement freely.

It looks a little weird that people outside of the library can even see that it's a `class` or `mixin`, if they can't use it. That goes for a `sealed class` too, but we are just so used to seeing `class` being used as a way to introduce a type (also because we can't avoid exposing an interface today), that we think of it as a neutral type declaration.
If we had `interface Foo { ... }` to produce only a type and an interface, we'd probably use that instead, and use  `sealed interface Foo` to only introduce the type. We probably won't have `sealed type Foo;` by itself.

accepted/future-releases/type-modifiers/feature-specification.md
lrhn: I can live with "class modifiers", since mixins are only used to build classes anyway,.

accepted/future-releases/base-interface-final/feature-specification.md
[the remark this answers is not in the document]
leafpetersen: I agree that if we need `mixin` in order to mix things in (your proposed semantics) within a library this is needed.

accepted/future-releases/type-modifiers/feature-specification.md
kallentu: I originally called them 'Class modifiers' so that works for me.

accepted/future-releases/type-modifiers/feature-specification.md
munificent: Yup! I wrote a comment on that issue comparing your proposal and Lasse's with this one. I think we're converging modulo a couple of questions about mixins.

accepted/future-releases/type-modifiers/feature-specification.md
munificent: Good catch, fixed.

munificent:
I'm going to go ahead and land this because having a long-running PR that moves files around is a nightmare. I don't love having it land in "accepted" while also knowing more tweaks are coming, but I think it's better than trying to cram all of those fixes into this PR.

Cat-sushi:
Built in classes will have class modifiers?
For example, 

- `bool`, `Null` will be `sealed class`s?
- `int`, `double` will be `final class`s?

Seemingly, user defined enums will be `sealed classe`s?

Cat-sushi:
Do you plan to deprecate `mixin` modifier of `class`?
I hope so, because [the matrix](https://github.com/dart-lang/language/blob/master/accepted/future-releases/class-modifiers/feature-specification.md#syntax "language/feature-specification.md at master · dart-lang/language") is very complex to me.

Isn't 3.0, the introducing version of `mixin` modifier, a good chance to deprecate that?
It means, `mixin` modifier would be a innately deprecated thing.

eernstg:
> Built in classes will have class modifiers?

Check out https://github.com/dart-lang/sdk/issues/50736.

> Do you plan to deprecate `mixin` modifier of `class`?

I'm not sure what you mean here. We can't actually deprecate that, because it doesn't exist today.

The proposal will _introduce_ the notion of a `mixin class`. The idea is that a mixin class can be used as a mixin (so you can mix it in) and it can be used as a class (so you can extend it, and if it isn't abstract then you can create instances of it). With this proposal, an entity declared as a `class` (not `mixin class`, not `mixin`) can _not_ be mixed in any more.

This is a breaking change, but you can add the keyword `mixin` to certain classes in order to eliminate the breakage:

If you mark an existing class which is already being used as a mixin such that it is a `mixin class`, then (1) existing code will not break, and (2) the maintainer of the mixin class knows, directly from the declaration, that this class must remain usable as a mixin and usable as a class. For instance, we can't add a generative constructor declaration to it, and we can't give it a superclass different from `Object`.

This is better than the situation today where a small subset of the existing classes can be used as a mixin (because they satisfy those constraints, perhaps by accident), but it is not documented at the declaration that it can be used as a mixin. We might well change the class such that it can't be used as a mixin any more, and nobody will know that this is a breaking change until the issues start coming in, "I'm mixing in this class, and it doesn't work any more".

With this proposal it _is_ documented at the class that it is intended to be usable as a mixin, and we'll get a compile-time error if we change the class in such a way that it cannot be used as a mixin any more. Conversely, no class can be mixed in if it does not have the keyword `mixin`.

eernstg:
> [the matrix](https://github.com/dart-lang/language/blob/master/accepted/future-releases/class-modifiers/feature-specification.md#syntax) is very complex

Here is one way to look at that matrix which makes it simpler: Consider the 'kind' of declaration (`class`, `mixin`, `mixin class`) as one dimension, the derivation capabilities (allowing `extends`/`with` and/or `implements`) as another dimension, and abstractness as the third dimension, cf. https://github.com/dart-lang/language/issues/2723#issuecomment-1355096266.

This means that you just need to familiarize yourself with 3 kinds, 4 derivation capabilities (two booleans), and 2 abstractness capabilities (one boolean).

Cat-sushi:
> Check out https://github.com/dart-lang/sdk/issues/50736.

 Thank you for your information.

Cat-sushi:
https://github.com/dart-lang/language/issues/2723#issuecomment-1355096266. is comprehensible.

The problem is that **Kind** is not orthogonal to **Derivation** and **Instances**.
I understand that `mixin class` is necessary to migration target of current `class`.
But, I don't feel importance of instanciatable/ extendable mixins.

eernstg:
> The problem is that Kind is not orthogonal to Derivation and Instances.

Instances is not: Every `mixin` is abstract (and the main effect of this is that they can have abstract members). A `mixin` declaration doesn't admit the keyword `abstract` (it's always implicit). So we only have the two other dimensions with a `mixin` declaration.

For `class` and `mixin class` declarations, `abstract` can be used, hence 3 dimensions.

Derivation comes in 4 variants (00: final, 01: base, 10: interface, 11: [no keywords]), where the bits mean "can be extended/mixed-in" respectively "can be implemented". All 4 variants of derivation can be used with each of the 3 kinds of declarations, and `abstract` can also be used or omitted with the kinds `class` and `mixin class`.

I'd say that's not very far from orthogonal... ;-)

> But, I don't feel importance of instanciatable/ extendable mixins.

Well, some combinations may be much more common than others. However, I tend to think that a full, orthogonal set of variants is a useful choice because we don't need to think about irregularities. You think about those dimensions, choose what you want, and you're done.

A `mixin class` allows the same declaration to support mixing in a set of member declarations, and it also supports usages as a regular class (so you can create instances, if it isn't abstract). This might be useful, so why not?

Cat-sushi:
> it's always implicit

I expressed such situation as "not orthogonal".
My wording would not be correct, sorry.

The point is that, simply the spec is not intuitive for me, and I'm afraid that is not intuitive for most developers.
I understand historical reason, so, now, I just hope documentation will be comprehensible.

eernstg:
But it's true, it is not strictly orthogonal (and the details are still being discussed, so I'm just talking about my preferred outcome): A strictly orthogonal model would allow the following cartesian product: `Abstractness * Derivation * Kind`. My preferred model is that we have `Abstractness * Derivation * {'class', 'mixin class'}` and `Derivation * {'mixin'}`. And the reason why we don't have an abstractness dimension in the 'mixin' part is that a mixin is always abstract (and the motivation for making them abstract was that we can't create an instance of a mixin anyway, so we might as well allow for abstract methods without additional keywords).

I hope it will all be comprehensible! ;-)

Cat-sushi:
By the way, in `final class`es, fields will be non-nullable promoted?
In non-final `class`es, `final class` will be recommended as a quick fix of nullable fields access without `!`?

eernstg:
Possibly. Note that all the constraints introduced by type modifiers are enforced in different libraries, but not in the declaring library. So if you have a `final class A` in library _L_ then you can also have a `final class B extends A`, and the declarations in `B` could invalidate the promotion. It's still possible to check whether the situation is known to guarantee soundness of promotions. I don't think adding `final` to a class is going to be a reasonable quick fix for non-promotion, however, because that's a very disruptive change as seen from many other points of view (for instance, it could break a lot of subtypes, via `extends`, `implements`, or `with`) in other libraries.

My preferred approach would be final getters, #1518, but that proposal is being debated and there's no guarantee that the language team as a whole will support it. The point is that a final getter is promotable, no matter which kind of class/mixin it occurs in.

Cat-sushi:
>  So if you have a final class A in library L then you can also have a final class B extends A, and the declarations in B could invalidate the promotion.

Yes, I know.
Anyway, I'm happy to know that the most frequently asked question could be solved.

Shahidul783:
touhidislam783

Shahidul783:
touhidislam783

PR #2728, opened 2022-12-17
[class mod] Clarify how language versioning affects modifiers.

### LINKED ISSUE #2725: [Class Modifiers] Specify capability modifier interactions with older language versions.
The current version of the [capability modifiers proposal](https://github.com/dart-lang/language/pull/2722) addresses the interactions of the `mixin` modifier with older language versions explicitly (see https://github.com/dart-lang/language/issues/2724) but is silent on the behavior of the other modifiers.  By implication, it is likely intended that these modifiers are not *available* in libraries with older language versions, but that such libraries must nevertheless respect the restrictions imposed on classes declared with capability modifiers.    That is, if the language version of this feature is `N`, then a class declared as `interface` in a library from language version `N` or greater may not be extended outside of its own library, even in a library with language version less than `N`.  Assuming this is the behavior we want (I think it is) we should make this explicit at least in commentary, and ensure that it gets tested.

cc @munificent @lrhn @eernstg @kallentu @stereotype441 @jakemac53 @natebosch @mit-mit @srawlins

munificent:
Fix #2725.

### DIFF
--- a/accepted/future-releases/class-modifiers/feature-specification.md
+++ b/accepted/future-releases/class-modifiers/feature-specification.md
@@ -611,6 +611,9 @@ Let `n` be the language version this proposal ships in. Then:
     to all other libraries, regardless of the versions of those libraries.
     "Ignorance of the law is no defense."*
 
+    **TODO:** Decide if we want to carve out an exception to this rule for the
+    SDK core libraries.
+
 *   A class declaration in a library whose language version is `< n` can be used
     as a mixin as long as the class meets the mixin restrictions. This is is
     true even if the library where the class is being used as a mixin is `>=
--- a/accepted/future-releases/class-modifiers/feature-specification.md
+++ b/accepted/future-releases/class-modifiers/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: Accepted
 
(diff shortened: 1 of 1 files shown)


accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -599,18 +599,37 @@ non-breaking.
 
 Let `n` be the language version this proposal ships in. Then:
 
+*   `base`, `interface`, `final`, and `mixin` can only be applied to classes and
+    mixins in libraries whose language version is `>= n`.
+
+*   When the `base`, `interface`, `final`, `mixin`, or `sealed` modifiers are
```
leafpetersen, line 605: I think we're still discussing whether we want to carve out an exception for the SDK core libraries.  I'm fine with landing this now though if you want - I don't expect us to change this for non-SDK core libraries.

leafpetersen (approved):

accepted/future-releases/class-modifiers/feature-specification.md
munificent: Added a TODO so that if an implementer reads the proposal they can see this might get tweaked.

PR #2729, opened 2022-12-17
[patterns] Fix "||" vs "&&" in rationale for no repeated named fields

gnprice:
If this syntax were allowed, one would expect it to bind the variables from both patterns.  The way to do that is with "&&".

Also fix a minor typo, in a separate commit.


munificent:
These are great, thank you! :D

PR #2736, opened 2022-12-20
Add support for expressions `List<int>` and `myFunction<int>` in the grammar

eernstg:
The ability to use a parameterized type as an expression and the ability to perform an explicit generic instantiation of a given generic function were added to Dart in version 2.15. The grammar change was not specified at the time. This PR does it.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -38,6 +38,8 @@
 % Dec 2022
 % - Change the definition of the type function 'flatten' to resolve soundness
 %   issue, cf. SDK issue #49396.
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -16117,6 +16119,7 @@ \subsection{Postfix Expressions}
 <selector> ::= `!'
   \alt <assignableSelector>
   \alt <argumentPart>
+  \alt <typeArguments>
```
lrhn, line 16122: Needs the look-ahead restriction.

PR #2767, opened 2023-01-11
Allow trivial constructors on mixin classes.

lrhn:
Specifies that a `class C = S with M;` declaration cannot  be made a `mixin class`, but can use the other modifiers. Specifies that a `mixin class` (or a non-`mixin` `class` used as a mixin inside the same library) can declare *trivial* generative constructors.

Small tweaks.

### DIFF
--- a/accepted/future-releases/class-modifiers/feature-specification.md
+++ b/accepted/future-releases/class-modifiers/feature-specification.md
@@ -558,14 +558,15 @@ It is a compile-time error to:
     or no `extends` clause. It cannot have any `with` clause.*
 
     A _trivial generative constructor_ is a non-redirecting generative constructor 
-    with
+    which has
 
     *   an empty parameter list,
-    *   no initializer list (no `: ...`), and 
-    *   no constructor body (only `;`).
+    *   no initializer list (no `: ...`),  
+    *   no constructor body (only `;`),
+    *   and is not marker `external`.
 
     Any other generative constructor is non-trivial.
-    A trivial generative constructor may be `const` and may be a named   constructor.
+    A trivial generative constructor may be `const` and may be a named constructor.
 
     *Declaring a trivial generative constructor allows the class to be used as
     both a mixin and as a superclass, even if it also declares other factory constructors
--- a/accepted/future-releases/class-modifiers/feature-specification.md
+++ b/accepted/future-releases/class-modifiers/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: Accepted
 
-Version 1.2
+Version 1.3
 
 Experiment flag: class-modifiers
 
@@ -19,8 +19,9 @@ Informally, the new syntax is:
 
 *   `base`: As a modifier on a class, allows the class to be extended but not
     implemented. As a modifier on a mixin, allows it to be mixed in but not
-    implemented. In other words, it takes away implementation.
-
+    implemented. In other words, it takes away being able to implement
+    the interface of the declaration.
+    
 *   `interface`: As a modifier on a class or mixin, allows the type to be
     implemented but not extended or mixed in. In other words, it takes away
     being used as a subclass through extension or mixing in.
@@ -361,7 +362,7 @@ does `MySubclass` now expose externally? We have a few options:
 
     ```dart
     abstract final class Vehicle {}
-
+    
     class LandVehicle extends Vehicle {}
     class AquaticVehicle extends Vehicle {}
     class FlyingVehicle extends Vehicle {}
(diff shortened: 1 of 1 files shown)


accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -525,37 +533,44 @@ mixinModifier     ::= 'sealed' | 'base' | 'interface' | 'final'
...
-
 *   Mix in a class not marked `mixin` outside of the library where it is
     declared, unless the class declaration being used as a mixin is in a library
     whose language version is older than the version this feature ships in.
-
+*   Mix in a class not marked `mixin` inside the library where it is declared,
+    if the class has a superclass other than `Object` or the class declares a 
+    non-trivial generative constructor. If the library has a language version
+    older than the version this feature ships int, the class must not declare
+    any generative constructor.
```
munificent, line 558: I think this is mostly covered by the bullet point on line 574 and should be merged with that.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -525,37 +533,44 @@ mixinModifier     ::= 'sealed' | 'base' | 'interface' | 'final'
...
     mixin that doesn't want to expose one.*
-
 *   Mix in a class not marked `mixin` outside of the library where it is
     declared, unless the class declaration being used as a mixin is in a library
     whose language version is older than the version this feature ships in.
-
+*   Mix in a class not marked `mixin` inside the library where it is declared,
+    if the class has a superclass other than `Object` or the class declares a 
+    non-trivial generative constructor. If the library has a language version
+    older than the version this feature ships int, the class must not declare
```
munificent, line 557: "ships int" -> "ships in".

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -525,37 +533,44 @@ mixinModifier     ::= 'sealed' | 'base' | 'interface' | 'final'
...
+    any generative constructor.
 *   Apply `mixin` to a class whose superclass is not `Object` or that declares a
-    generative constructor. *A `mixin class` states that you intend the class to
-    be mixed in, which is inconsistent with defining a class that can't be used
+    _non-trivial generative constructor_. 
+    *A `mixin class` states that you intend the class to
+    be mixed in, which is inconsistent with defining a class that cannot be used
     as a mixin. Note that this means that `mixin` on a class becomes a helpful
     reminder to ensure that you don't inadvertently break your class's ability
     to be used as a mixin.*
```
munificent, line 565: I'm finding it pretty hard to parse the increasingly complex set of restrictions around when you can apply a mixin. Maybe it would be easier to specify when it's *allowed* and say that it's a compile-time error otherwise.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -525,37 +533,44 @@ mixinModifier     ::= 'sealed' | 'base' | 'interface' | 'final'
...
-
+    *Such a class can have an `extends` clause of the form `extends Object`,
+    or no `extends` clause. It cannot have a `with` clause. It can be used as a `mixin`
+    declaration, but also cannot have an `on` clause.* 
+    A trivial generative constructor is a non-redirecting generative constructor 
+    with an empty parameter list, no initializer list (no `: ...`) and no body (only `;`). 
+    Any other generative constructor is non-trivial.
+    A trivial generative constructor *may* be `const` and may be a named constructor.
```
munificent, line 572: Probably worth adding some non-normative text like:

*A trivial generative constructor allows the class to continue to be used as a mixin even if it declares other factory constructors.*

munificent (approved):

accepted/future-releases/class-modifiers/feature-specification.md
lrhn: Applying a *mixin* is easy. It's using a non-mixin class as a mixin that's complicated (so why do we even allow it in new code? I'm not sold on allowing you to mix in plain classes from your own library, because I don't see `mixin` as a modifier on the `class` that other classes lack. I see `mixin class` as two declarations in one, and other `class` declarations are just not defining mixins, so mixing them in makes as much sense as extending mixins.)

Anyway, you can mix in a (mixin derived from a) class when:
* The class is in a library with a language version before this feature, and
    * The class has `Object` as superclass and declares no (generative?) constructors.
* The class is in the same library, which has a language version including this feature, and
    * The class has `Object` as superclass and declares no non-trivial generative constructors.
    
A `mixin class` must satisfy the latter requirements.

I think the items I'm converging on say the same thing as negatives, in about the same verbosity.

lrhn:
Updated, PTAL

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -455,7 +456,11 @@ class UsesAsMixin extends OtherSuperclass with Both {} // OK.
 
 This proposal builds on the existing sealed types proposal so the grammar
 includes those changes. The full set of modifiers that can appear before a class
-or mixin are `abstract`, `sealed`, `base`, `interface`, `final`, and `mixin`.
+or mixin declaration are `abstract`, `sealed`, `base`, `interface`, `final`, and `mixin`.
+
+*The modifiers do not apply to other declarations. This includes `enum` declarations.
+**TODO**: Do they apply to inline class declarations?*
```
munificent, line 462: I believe @eernstg updated the inline class proposal to discuss the modifiers there, so we don't need to here.

munificent (approved):
Still LGTM!

accepted/future-releases/class-modifiers/feature-specification.md
lrhn: OK, removing it here.

PR #2771, opened 2023-01-12
Start record positional field getters at 1.

### LINKED ISSUE #2638: Should record fields start at `$0` or `$1`.
Last minute bike-shedding, I know.

We have defined positional record fields to be accessible as `$0`, `$1` etc. getters.
While starting at zero is good for integer indices (like arguments to `operator[]`), I'm not sure it is the best choice here. The getters are not integers, you won't have to do arithmetic on them, so starting at zero is not a benefit in that regard.

I worry that users will expect think of them as "1st" field, "2nd" field, etc., and therefore expect them to start at 1.
(I worry about that, because that's what I catch myself thinking.)

If it's not too late, I'd like to suggest we start at `$1` instead.
(I promise I'll fix our test files if we make the change!)

munificent:
Fix #2638.

### DIFF
--- a/accepted/future-releases/records/records-feature-specification.md
+++ b/accepted/future-releases/records/records-feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: Accepted
 
-Version 1.18 (see [CHANGELOG](#CHANGELOG) at end)
+Version 1.20 (see [CHANGELOG](#CHANGELOG) at end)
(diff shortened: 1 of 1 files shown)


lrhn (approved):
LGTM.

I fixed the implementations too. It's convenient that now a positional record field name cannot start with `$0`, which rules out both leading zeros and `$0` in one step.

PR #2773, opened 2023-01-13
[enhanced enums] Add support for NEW in enumEntry

eernstg:
The specified grammar does not support an `<enumEntry>` of the form `name.new(...)` or `name.new<...>(...)`. We do have a test for this case, and it seems reasonable to allow it. This PR adds support for `new` to occur in that position to the feature spec.

(There is no associated implementation work because the test already runs without failure.)

### DIFF
--- a/accepted/2.17/enhanced-enums/feature-specification.md
+++ b/accepted/2.17/enhanced-enums/feature-specification.md
@@ -43,7 +43,8 @@ The grammar of the `enum` declaration becomes:
   `}'
 
 <enumEntry> ::= <metadata> <identifier> <argumentPart>?
-  | <metadata> <identifier> <typeArguments>? `.' <identifier> <arguments>
+  | <metadata> <identifier> <typeArguments>? `.' (<identifier> | `new')
+    <arguments>
 ```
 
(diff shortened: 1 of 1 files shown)


accepted/2.17/enhanced-enums/feature-specification.md
```diff
@@ -43,7 +43,8 @@ The grammar of the `enum` declaration becomes:
   `}'
 
 <enumEntry> ::= <metadata> <identifier> <argumentPart>?
-  | <metadata> <identifier> <typeArguments>? `.' <identifier> <arguments>
+  | <metadata> <identifier> <typeArguments>? `.' (<identifier> | `new')
```
lrhn, line 46: Should we have a ``<constructorName> ::= <identifier> | `new'`` production somewhere, so we can reuse it everywhere we refer to the name of a named constructor?

That should avoid forgetting a place if we ever change it again.

lrhn (approved):

accepted/2.17/enhanced-enums/feature-specification.md
eernstg: I added it to Dart.g with a 'TODO', then it'll go into the language specification next.

PR #2777, opened 2023-01-18
[patterns] Specify the behavior of pattern-for collection elements.

### LINKED ISSUE #2769: Semantics for pattern-for-in and pattern-for elements
The patterns proposal weaves patterns into for loops by amending `forLoopParts` and `localVariableDeclaration`. That allows patterns in both for *statements* and for collection *elements.* That's good because we do want to support that. But the proposal fails to actually state that patterns are supported in for elements and doesn't define the semantics.

The semantics should be pretty obvious: If you know how patterns in a for statement work, and you know how for elements work, it's just connecting the dots. But connecting those dots is the proposal's job, so it should do that.

munificent:
Fix #2769.

### DIFF
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -3334,6 +3334,9 @@ Here is one way it could be broken down into separate pieces:
 
 -   Specify the runtime behavior of pattern-for collection elements (#2769).
 
+-   Clarify the inference context for the iterator expression in pattern for-in
+    statements (#2775).
+
 ### 2.20
 
 -   Fix soundness hole in map patterns with badly behaved maps (#2685).
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -2535,11 +2535,11 @@ A statement of the form:
 for (<patternVariableDeclaration>; <condition>; <increment>) <statement>
 ```
 
-Is executed similar to a traditional for loop except that multiple variables may
-be declared by the pattern instead of just one. As with a normal for loop, those
-variables are freshly bound to new values at each iteration so that if a
-function in the body closes over a variable, it captures the value at the
-current iteration and is not affected by later iteration.
+Is executed like a traditional for loop though is more likely to declare
+multiple variables. As with a normal for loop, those variables are freshly bound
+to new values at each iteration so that if a function in the body closes over a
+variable, it captures the value at the current iteration and is not affected by
+later iteration.
 
 The increment clause is evaluated in a scope where all variables declared in the
 pattern are freshly bound to new variables holding the current iteration's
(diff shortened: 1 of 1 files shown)


munificent:
Friendly ping, @lrhn. :)

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -2591,6 +2614,42 @@ Where `<keyword>` is `var` or `final` is treated like so:
...
+
+A collection element of the form:
+
+```dart
+for (<keyword> <pattern> in <expression>) <element>
+```
+
+Where `<keyword>` is `var` or `final` is treated like so:
+
+1.  Let `I` be the type of `<expression>`, inferred using the context type
```
lrhn, line 2627: -> be the static type of

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -2591,6 +2614,42 @@ Where `<keyword>` is `var` or `final` is treated like so:
...
+
+    2.  Else if `I` is `dynamic` then `E` is `dynamic`.
+
+    3.  Else it is a compile-time error.
+
+3.  Type check `<pattern>` with matched value type `E`.
+
+4.  If there are no compile-time errors, then execution proceeds as the
+    following code, where `id1` and `id2` are fresh identifiers and `append()`
+    is an operation to add an element to the surrounding collection being built:
```
lrhn, line 2642: We can't use `var` here, because this "rewrite" happens after type inference has already been performed on the expressions.
If we do type analysis again, then we get issues like `<expression>` having two, possibly different, static types. If not, it's undefined what declared type `var id1 = ...` has.
(In general, I prefer to not use desugaring as a specification device if it can be avoidable, but if we do, always give fully typed programs.)

Fx:

Let *E* be the static type of `<element>`.

```dart
  I id1 = <expression>;
  Iterator<T> id2 = id1.iterator;
  while (id2.moveNext()) {
    <keyword> <pattern> = id2.current;
    E id3 = <element>;
    append(id3);
  }
```

Maybe we should just introduce `let` as a specification device, and say that 
```dart
  for (<pattern> in <expression>) <element>
```
desugars (still post inference) to:
```dart
  for (T id in <expression>) let <pattern> = id in <expression>
```

lrhn (approved):
Whoa, you have to actually "Finish your review" to pose comments. Who'd have thunk.
(Sorry :disappointed:).

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -2540,6 +2545,23 @@ for (var fn in fns) {
...
+#### Pattern-for element
+
+Likewise, a collection element of the form:
+
+```dart
+for (<patternVariableDeclaration>; <condition>; <increment>) <element>
+```
+
+Is executed similar to a non-pattern for element except that the pattern may
+declare multiple variables. As with pattern for statements, all of the variables
```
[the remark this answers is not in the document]
munificent: TIL!

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -2540,6 +2545,23 @@ for (var fn in fns) {
...
+are declared fresh every iteration. Inside the `<increment>` expression, reads
+of those variables reflect their current values. Writes to them write to the
+values the newly bound variables will have on the next iteration.
```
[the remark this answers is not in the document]
munificent: Went with vague.

accepted/future-releases/0546-patterns/feature-specification.md
munificent: > (In general, I prefer to not use desugaring as a specification device if it can be avoidable, but if we do, always give fully typed programs.)

I'm OK with specifying in terms of desugaring as long as we only do that after all type analysis and type errors are done. I think desugaring works pretty well for defining runtime semantics.

I'll go with your first fix, thanks.

accepted/future-releases/0546-patterns/feature-specification.md
munificent: Done.

PR #2790, opened 2023-01-25
[Patterns] Fixed typo in the variable declaration scope example

sgrekhov:
Before the change there was another error `The value of type 'int' is not assignable to the required type 'List<int>'`


munificent:
Thanks!

PR #2796, opened 2023-01-26
[patterns] Fix spec for context type schema of variable and identifiers.

### LINKED ISSUE #2795: [patterns] Specification of required type for variablePattern needs to be split into variable/identifier
Thanks to @sgrekhov for noticing this. [This case](https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#:~:text=Variable%3A-,In%20an%20assignment%20context,-%2C%20the%20context%20type) should be moved from the 'Variable' paragraph to the 'Identifier' paragraph, because a variable pattern is an error if it occurs in an assignment context.

For the remaining cases, I believe it's fine to say in the 'Identifier' paragraph that the treatment is as described in the 'Variable' paragraph.

munificent:
This got slightly broken when I moved identifiers out to a separate pattern grammar. This PR puts it back to right.

Fix #2795.

I took the opportunity to slightly clean things up too. Thanks for noticing, @sgrekhov!

### DIFF
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -1769,25 +1769,25 @@ The context type schema for a pattern `p` is:
 
 *   **Variable**:
 
-    1.  In an assignment context, the context type schema is the static type of
-        the variable that `p` resolves to.
-
-    1.  Else if `p` has no type annotation, the context type schema is `_`.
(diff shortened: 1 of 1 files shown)


munificent:
Friendly ping, @eernstg. :)

eernstg:
Oops, didn't see this one!

eernstg (approved):
LGTM!

PR #2797, opened 2023-01-26
[patterns] In lists, don't call v[e] if the subpattern is a wildcard.

### LINKED ISSUE #2671: Wildcards in map and list patterns lead to "unnecessary" operations
As currently specified, matching an expression `e` against a pattern like `{'foo': _}` causes three map operations to be performed:
- check if `e.length` is equal to 1
- check if `e.containsKey('foo')` returns `true`
- evaluate `e['foo']` and discard the result

For a standard implementation of `Map`, the third bullet has no effect.

Similarly, matching an expression `e` against a pattern like `[_]` causes two list operations to be performed:
- check if `e.length` is equal to 1
- evaluate `e[0]` and discard the result

For a standard implementation of `List`, the second bullet has no effect.

Since the user may have their own implementations of `Map` and `List`, these "unnecessary" operations could actually have observable effects in contrived programs.  Because of that, it may be difficult for the compiler to optimize out the unnecessary operations even in programs where they truly have no effect.

We could fix this by specifying that when a map or list pattern contains a wildcard pattern, the call to `operator[]` is not performed.  Should we?

munificent:
Fix #2671.

cc @eernstg @stereotype441

### DIFF
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: Accepted
 
-Version 2.21 (see [CHANGELOG](#CHANGELOG) at end)
+Version 2.22 (see [CHANGELOG](#CHANGELOG) at end)
 
 Note: This proposal is broken into a couple of separate documents. See also
 [records][] and [exhaustiveness][].
@@ -2852,9 +2852,20 @@ To match a pattern `p` against a value `v`:
 
     5.  Match the head elements. For `i` from `0` to `h - 1`, inclusive:
 
-        1.  Extract the element value `e` by calling `v[i]`.
+        1.  Let `s` be the `i`th element subpattern.
 
-        2.  Match the `i`th element subpattern against `e`.
+        2.  If `s` is an identifier pattern whose name is `_` then do nothing
+            for this element.
+
+            *Wildcards are useful in list patterns to control the index that
+            latter element subpatterns access and to affect the length that the
+            pattern checks. When a user does that, we don't want them to pay a
+            runtime penalty for accessing list elements that the pattern won't
+            use anyway.*
+
+        3.  Else extract the element value `e` by calling `v[i]`.
+
+        4.  Match `s` against `e`.
 
     6.  If there is a matching rest element:
 
@@ -2875,10 +2886,14 @@ To match a pattern `p` against a value `v`:
     7.  Match the tail elements. If `t > 0`, then for `i` from `0` to `t - 1`,
         inclusive:
 
-        1.  Extract the element value `e` by calling `v[l - t + i]`.
+        1.  Let `s` be the subpattern `i` elements after the rest element.
 
-        2.  Match the subpattern `i` elements after the rest element against
-            `e`.
+        2.  If `s` is an identifier pattern whose name is `_` then do nothing
+            for this element.
+
+        3.  Else extract the element value `e` by calling `v[l - t + i]`.
+
+        4.  Match `s` against `e`.
 
     8.  The match succeeds if all subpatterns match.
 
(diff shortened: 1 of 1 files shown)


eernstg (approved):
LGTM

(Gave up on consistency, developers will probably be happier about avoiding the cost even though it's an yet another exception to remember ;-).

munificent:
> (Gave up on consistency, developers will probably be happier about avoiding the cost even though it's an yet another exception to remember ;-).

Yeah, the inconsistency bothers me too. The thing that convinced me is that I think it exceedingly unlikely that developers *will* have to remember this rule. I think almost no one will ever notice this. It's just that their code will be somewhat faster without them having to do any additional work.

eernstg:
You're probably right. ;-)

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -2875,10 +2886,14 @@ To match a pattern `p` against a value `v`:
     7.  Match the tail elements. If `t > 0`, then for `i` from `0` to `t - 1`,
         inclusive:
 
-        1.  Extract the element value `e` by calling `v[l - t + i]`.
+        1.  Let `s` be the subpattern `i` elements after the rest element.
+
+        2.  If `s` is an identifier pattern whose name is `_` then do nothing
+            for this element.
```
lrhn, line 2892: Probably wise. Even if you can write `[var x, ...]`, I bet someone will write `[var x, ... _]` anyway. No need to make that extra painful.

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -2852,9 +2852,20 @@ To match a pattern `p` against a value `v`:
...
+            latter element subpatterns access and to affect the length that the
+            pattern checks. When a user does that, we don't want them to pay a
+            runtime penalty for accessing list elements that the pattern won't
+            use anyway.*
+
+        3.  Else extract the element value `e` by calling `v[i]`.
+
+        4.  Match `s` against `e`.
 
     6.  If there is a matching rest element:
```
lrhn, line 2870: Should this be "If there is a rest element:"? We don't know if it's matching before we do the thing below.

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -2852,9 +2852,20 @@ To match a pattern `p` against a value `v`:
...
```
lrhn, line 2871: I think we should avoid calling `sublist` if the rest element has no pattern, like `[var x, ..., var y]`. No need to create the sublist here, the `...` is just for filling out positional positions.

And because someone will probably end up writing `[var x, ... _, var y]` to mean the same thing, we may want to not create the sublist in that case either.

So.

>    1. If the rest element has no pattern, then do nothing for this element.
>    2. Otherwise let *s* be the rest element pattern after the `...`.
>    3. If *s* is an identifier pattern whose name is `_` then do nothing
>        for this element.
>    4. Otherwise ...

lrhn (approved):

accepted/future-releases/0546-patterns/feature-specification.md
munificent: Earlier in the proposal, we define:

> We refer to a rest element with a subpattern as a *matching rest element*, and a rest element with no subpattern as a *non-matching rest element*.

So it just means "rest element with a subpattern", not necessarily "rest element whose subpattern has successfully matched."

accepted/future-releases/0546-patterns/feature-specification.md
munificent: > I think we should avoid calling sublist if the rest element has no pattern, like [var x, ..., var y]. No need to create the sublist here, the ... is just for filling out positional positions.

Agreed. That's why the proposal says "If there is a matching rest element" and not just "If there is a rest element".

> And because someone will probably end up writing [var x, ... _, var y] to mean the same thing, we may want to not create the sublist in that case either.

If, for some reason, the user chooses to write that, the only possible reason to do so over just `...` is because you _want_ it to call `sublist()`, so I think it makes sense to leave it as currently specced.

PR #2798, opened 2023-01-27
Inherited restrictions

munificent:
OK, I think I incorporated everything from [your Gist](https://gist.github.com/lrhn/981d4528fbb995c23afa0f9436209537) into the proposal and removed the stuff that it replaces. The rules you propose make sense to me. Let me know what I got wrong in the merge. :)

cc @kallentu

### DIFF
--- a/accepted/future-releases/class-modifiers/feature-specification.md
+++ b/accepted/future-releases/class-modifiers/feature-specification.md
@@ -776,7 +776,7 @@ needed.
 1.4
 
 - Update rules to close loopholes on classes that don't want to expose
-  interfaces.
+  interfaces (#2755, #2757).
 
 1.3
 
--- a/accepted/future-releases/class-modifiers/feature-specification.md
+++ b/accepted/future-releases/class-modifiers/feature-specification.md
@@ -1,10 +1,10 @@
 # Class modifiers
 
-Author: Bob Nystrom
+Author: Bob Nystrom, Lasse Nielsen
 
 Status: Accepted
 
-Version 1.3
+Version 1.4
 
 Experiment flag: class-modifiers
 
@@ -24,7 +24,7 @@ Informally, the new syntax is:
     
 *   `interface`: As a modifier on a class or mixin, allows the type to be
     implemented but not extended or mixed in. In other words, it takes away
-    being used as a subclass through extension or mixing in.
+    being able to inherit from the type.
 
 *   `final`: As a modifier on a class or mixin, prohibits extending,
     implementing, or mixing in.
@@ -420,6 +420,51 @@ This proposal takes the last option where types have exactly the restrictions
 they declare but a lint can be turned on for users who want to be reminded if
 they re-add a capability in a subtype.
 
+### Inherited restrictions
+
+Allowing you to ignore restrictions on your own types allows some useful
+architectural patterns, but it's important that doing so doesn't let you ignore
+restrictions on types from *other* libraries because then you could break the
+invariants the library expects. In particular, consider:
+
+```dart
+// lib_a.dart
+base class A {
+  void _private() {
+    print('Got it.');
+  }
+}
+
+callPrivateMethod(A a) {
+  a._private();
+}
+```
+
+This library declares a class and marks it `base` to ensure that every instance
+of `A` in the program must be an `A` or a class that inherits from it. That in
+turn ensures that the call to `_private()` in `callPrivateMethod()` is always
+safe.
+
+Now consider:
+
+```
+// lib_b.dart
+import 'lib_a.dart';
+
+base class B extends A {} // OK: Inheriting.
+
+class C implements B {} // OK: Ignoring restriction on own type B.
+```
+
+These two class declarations each seem to be fine. But put together, the result
+is a class `C` that is a subtype of `A` but doesn't inherit from it and doesn't
+have the `_private()` method that lib_a.dart expects.
+
+So we want to allow libraries to ignore restrictions on their own types, but we
+need to be careful that doing so doesn't break invariants in *other* libraries.
+In practice, this means that when a class opts out of being implemented using
+`base` or `final`, then that particularl restriction can't be ignored.
+
 ## Mixin classes
 
 In line with Dart's permissive default nature, Dart allows any class declaration
@@ -458,7 +503,8 @@ This proposal builds on the existing sealed types proposal so the grammar
 includes those changes. The full set of modifiers that can appear before a class
 or mixin declaration are `abstract`, `sealed`, `base`, `interface`, `final`, and `mixin`.
 
-*The modifiers do not apply to other declarations. This includes `enum` declarations.
+*The modifiers do not apply to other declarations like `enum`, `typedef`, or
+`extension`.*
 
 Many combinations don't make sense:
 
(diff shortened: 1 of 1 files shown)


accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -530,62 +576,142 @@ mixinModifier     ::= 'sealed' | 'base' | 'interface' | 'final'
...
-mixin the typedef resolves to is defined to determine if the behavior is
-allowed. *Note that the library where the _typedef_ is defined does not come
-into play. Typedefs cannot be marked with any of the new modifiers.*
+There are a few changes around mixins to support `mixin class` and disallow
+using normal `class` declarations as mixins while dealing with language
+versioning and backwards compatibility.
+
+Currently, a class may only be used as a mixin if it has a default constructor.
+This prevents the class from defining a `const` constructor or any factory
+constructors. We loosen this somewhat. Define a *trivial generative constructor*
```
kallentu, line 653: So to confirm:
```
mixin class M {
  ...
  M(); // 1. This is OK
  const M(); // 2. This is OK
  M(this.x); // 3. This is an error
  M() { ... } // 4. This is an error
  M(): super(); // 5. This is OK
  factory M = _M; // 6. for some _M constructor, this is an error.
}
```

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -530,62 +576,142 @@ mixinModifier     ::= 'sealed' | 'base' | 'interface' | 'final'
...
-    mixin that doesn't want to expose one.*
-    
-*   Apply `mixin` to a class whose superclass is not `Object` or that declares a
-    _non-trivial generative constructor_.
-    *Such a class can have an `extends` clause of the form `extends Object`,
-    or no `extends` clause. It cannot have any `with` clause.*
+*   Mix in a mixin or mixin class marked `interface`, `final` or `sealed`
+    outside of the library where it is declared.
+
+There is no direct restriction on types allowed in `on` clauses of mixin
```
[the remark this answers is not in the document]
lrhn: Yes, `B` is an error because it's not a `base` declaration, and `C` is not an error.

This ensures that no subtype of `A` has an implementable interface.
Another rule ensures, effectively, that other libraries also cannot ignore that `base` within their own library, when it was forced by another library.

Examples is a good idea.

accepted/future-releases/class-modifiers/feature-specification.md
lrhn: `factory M` is OK, it's not a generative constructor. The restriction is only on generative constructors, which must all be trivial.

accepted/future-releases/class-modifiers/feature-specification.md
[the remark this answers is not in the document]
munificent: Added examples. Just removed the sentence since I think it distracts instead of clarifying.

It's like telling someone there's no alligator in the pantry. They probably weren't even thinking about alligators and now they're wondering if they should be.

accepted/future-releases/class-modifiers/feature-specification.md
munificent: Added examples.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -530,62 +577,282 @@ mixinModifier     ::= 'sealed' | 'base' | 'interface' | 'final'
...
+
+    // b.dart
+    import 'a.dart';
+
+    base class S extends B {} // Can't be implemented locally but OK.
+
+    class D extends S {} // Error, needs `base`, `final`, or `sealed`.
+    ```
+
+*   A class or mixin declaration `D` implements the interface of a class or
```
munificent, line 751: By "implements the interface" here, you mean literally "with an `implements` clause" and not just "is a subtype of" right? This error is about inducing a subtype relation that does *not* inherit. Any subtype that does inherit (`extends`, `with`, or `on`) is OK?

lrhn: That is what I mean. I think that in the original I had something early saying "yadday implement (has the type in the `implements` clause) yadda ...".

However, I don't think it's consistent with how the language specification uses `implements X`, which means "have X as a superinterface", which can be both direct or indirect/transitive.

The spec says, about direct super-interfaces,
> ```latex
> A class has a set of \Index{direct superinterfaces}.
> This set contains the interface of its superclass
> and the interfaces of the classes specified in
> the \IMPLEMENTS{} clause of the class.
> ```

So, to be consistent, we should change 
> ... implements the interface of a class or mixin declaration `S`, declared in ...

to 

> ... specifies a class or mixin `S` as an element in the type list of its `implements` clause", `S` is declared in ...

@eernstg How to phrase?

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -530,62 +577,282 @@ mixinModifier     ::= 'sealed' | 'base' | 'interface' | 'final'
...
+
+*   A `mixin class` declaration declares any non-trivial generative constructor.
+    *It may declare no constructors, in which case it gets a default
+    constructor, or it can declare factory constructors and/or trivial
+    generative constructors.*
+
+These rules ensure that when you mark a `class` with `mixin` that it *can* be
+used as one.
+
+A class not marked `mixin` can still be used as a mixins when the class's
```
lrhn, line 826: as a mixins => as a mixin

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -530,62 +577,282 @@ mixinModifier     ::= 'sealed' | 'base' | 'interface' | 'final'
...
+
+    *   `D` is a mixin-application class declaration (has the form `class D =
+        ...;`), or
+
+    *   `D` declares any non-trivial generative constructors.
+
+    *When a class is in a library where it possible to distinguish between
+    whether the class is intended to use it as a mixin or not, the author is
+    obligated to document the intent, and that intent applies to all other
+    libraries regardless of their version.*
```
lrhn, line 855: Maybe worth saying that A class _can_ mix in a non-mixin class from the same library, if it has `Object` as supertype and no non-trivial generative constructors. Adding `mixin` to a class declaration only changes whether *other* libraries can do.

(And as always, that's because you only risk breaking your own library when changing the class, and you should know immediately and be in a position to fix it.)

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -616,44 +883,26 @@ The changes in this proposal are guarded by a language version. This makes the
...
-*   A class declaration in a library whose version is `>= n` must be explicitly
-    marked `mixin class` to allow the class to be used as a mixin from another
-    library. This is true even if the library where the class is being used as 
-    a mixin is `< n`.
-    
-    *When a class is in a library where it possible to distinguish between
-    whether the class is intended to use it as a mixin or not, the author is
-    obliged to document the intent, and that intent applies to all other libraries 
-    regardless of their version.*
+    This is a special case behavior only available to platform libraries.
```
lrhn, line 905: I'd add:
> Package libraries should use versioning to to introduce breaking restrictions instead,
> and those libraries can then rely on the restrictions being enforced.

lrhn (approved):

accepted/future-releases/class-modifiers/feature-specification.md
eernstg: I really should have studied this PR carefully, but I did not manage to find the time.

Anyway, '`T` implements `S`' in spec-ese means that `T` is `T0` bounded for some `T0` (e.g., `X extends int` is `int` bounded, `int` is `int` bounded), and there is a superinterface path from `T0` to to `S` (not just a subtype relation). So `X extends List<int>` implements `Iterable<int>`, but not `Iterable<num>`.

I think the '`S` implements `T`' terminology works fine in this case, because we're generally talking about relationships that are created using clauses like `extends` and `implements`, and they're all covered.

If we want to specify that a given type `S` occurs in an `extends`, `with`, `implements`, or `on` clause in the declaration of `D` then we would simply say that `S` is a direct superinterface of `D`.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -530,62 +577,282 @@ mixinModifier     ::= 'sealed' | 'base' | 'interface' | 'final'
...
+in a subtype. If a class is marked `base` or `final`, you may still implement
+the class's interface inside the same library, but the implementing class must
+again be marked `base`, `final`, or `sealed` to avoid it exposing an
+implementable interface.*
+
+Further, while you can ignore some restrictions on declarations within the same
+library, you can't use that to ignore restrictions inherited from other
+libraries.
+
+We say a class or mixin declaration `D` *can't be implemented locally* if it
```
eernstg, line 684: Trying to grasp what this means, my current hypothesis is that: if `D` can't be implemented locally then it is a compile-time error for a `D2` to be declared in the same library as `D` if `D2` has a clause of the form `implements ... D ...` (and most likely also if it has `on ... D ...`).

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -530,62 +577,282 @@ mixinModifier     ::= 'sealed' | 'base' | 'interface' | 'final'
...
+extends, mixes in, implements, or has as on type any declaration `S` where:
+
+*   `S` is from another library than `D`, and `S` has the modifier `base`,
+    `final` or `sealed`, or
```
eernstg, line 688: This is surprising to me, because I would assume that it is already an error for `D` to have `S` as a direct superinterface if `S` is declared in a different library and has `final` or `sealed`.

So why would we define the notion of 'can't be implemented locally' for a declaration that can't exist in the first place?

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -530,62 +577,282 @@ mixinModifier     ::= 'sealed' | 'base' | 'interface' | 'final'
...
-    *   and is not marker `external`.
+    ```dart
+    // a.dart
+    base class B {}
+
+    // b.dart
+    import 'a.dart';
+
+    // This can't be implemented locally (from the previous rule):
+    base mixin class S extends B {}
```
eernstg, line 713: Can a mixin class extend `B`?

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -530,62 +577,282 @@ mixinModifier     ::= 'sealed' | 'base' | 'interface' | 'final'
...
+
+    // And thus these also can't be implemented locally (from this rule):
+    base class DE extends S {}
+    base class DM with S {}
+    base class DI implements S {}
```
eernstg, line 718: Again, it's confusing if `DI` is given as an example of a class that receives a certain treatment when used from other declarations, if `DI` itself is a compile-time error.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -530,62 +577,282 @@ mixinModifier     ::= 'sealed' | 'base' | 'interface' | 'final'
...
+    base mixin MO on S {}
```
eernstg, line 719: Same issue as in line 718.

eernstg (commented):
Added a few comments..

accepted/future-releases/class-modifiers/feature-specification.md
lrhn: Because of the lack of restriction on `on` types, those are possible for the "has as on type" case.

accepted/future-releases/class-modifiers/feature-specification.md
lrhn: Correct for `implements ...D...`. No to the `on ... D ...`, as usual it's not a problem. 

The `on ... D ...` declared `mixin` will need to be `base`, `final` or `sealed` itself, and it too can't be implemented locally, so it cannot be implemented at all.

It can only come into play as the type of an *object* if it's applied to another type which implements `D`, in which case it inherits members from that class, which is what we want to ensure.

The crux is that we only do exhaustiveness on objects, not types.

```dart
sealed class A {}
// Assume these are only immediate subtypes of A
final class B extends A {}
final class C extends A {}
...
   List<A> l = ...;
   return switch (l) {
         List<B> lb => ...
         List<C> lb => ...
   }
```
This expression switch is not exhaustive, because `B` | `C` only exhaust *instances* of `A`.
It's possible to have a list of exactly type `List<A>`, even if `A` is abstract and cannot have any own instances.
We only care about exhaustiveness on instances, which is why `A` must be abstract and unimplementable, so all instance are instances of `B` or `C`.
We can allow other immediate subtypes, as long as those types are also abstract and uninhabitable by values that are not also `B` or `C`.
And mixins satisfy that.

An other-library declaration of
```dart
mixin M on A {
  ...
}
```
introduces a new direct subtype of `A`, but it doesn't introduce any possible instances that are not also subtypes of `B` or `C`. The mixin must be applied to a class which is *extensible* (`base` or no modifier) before an instance can be created.

Because of the other constraints, every extensible class which implements `A` is either declared in the same library as `A` and subtypes either `B` or `C` (if it directly subtypes `A`, it becomes part of the sealed set of subtypes, which we assume here to be just `B` and `C`).
`A` itself is not extensible from outside the library, and `M` is not implementable or instantiable.

accepted/future-releases/class-modifiers/feature-specification.md
lrhn: Yes, the `DI` class is a compile-time error.

eernstg (commented):
Looks good! I added some comments. I believe we should use direct references to the clauses `extends`, `implements` etc in order to specify rules about those clauses. This is because other spec documents use "`T` implements `S`" to indicate a transitive relation.

munificent:
I have to admit that I can't keep track of all of the subtleties and the combinations of capabilities in this PR. I'm having trouble even understanding what many of the review comments mean. Part of this is because I'm also trying to get patterns supported in dart_style right now, and switching between those tasks involves loading and unloading a lot of context.

Since the formatting stuff is time-critical, @lrhn or @eernstg, is there any chance I could ask one of you to take over this PR and get it landed? I don't feel like I'm much of a value add.

eernstg:
It would be great to get the formatter done! I'll try to create a new revision of this one, and @lrhn will review.

accepted/future-releases/class-modifiers/feature-specification.md
eernstg: With the decision to treat `on` as a benign propogator-rather-than-initiator of leaks, we won't do anything to prevent them. We can of course still lint them.

accepted/future-releases/class-modifiers/feature-specification.md
eernstg: But we just took `on` away from all these constraints, so we just can't create the mixin case.

accepted/future-releases/class-modifiers/feature-specification.md
eernstg: Removed it.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -530,62 +577,282 @@ mixinModifier     ::= 'sealed' | 'base' | 'interface' | 'final'
...
+    ```
+
+Otherwise, `D` can be implemented locally. It is a compile-time error if:
+
+*   A class or mixin declaration `D` can't be implemented locally, and `D` is
```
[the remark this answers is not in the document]
eernstg: Removed this rule.

accepted/future-releases/class-modifiers/feature-specification.md
eernstg: I did not add that, based on the expectation that this will in fact be an error.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -616,44 +883,26 @@ The changes in this proposal are guarded by a language version. This makes the
 restriction on not allowing classes to be used as mixins by default
```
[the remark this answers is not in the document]
eernstg: Done.

eernstg (commented):
Review response --- @lrhn, PTAL.

Please search for !!!TODO!!!, which occurs in two locations: I think we need to include `on` types in the "can't implement locally" rule as well.

accepted/future-releases/class-modifiers/feature-specification.md
eernstg: Done.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -616,44 +863,29 @@ The changes in this proposal are guarded by a language version. This makes the
...
-    *When a class is in a library where it possible to distinguish between
-    whether the class is intended to use it as a mixin or not, the author is
-    obliged to document the intent, and that intent applies to all other libraries 
-    regardless of their version.*
+*   We would like to add modifiers to some classes in platform (i.e. `dart:`)
+    libraries when this feature ships. But we would also like to not immediately
+    break existing code. To avoid forcing users to immediately migrate,
+    declarations in pre-feature libraries can ignore modifiers on some
+    declarations in platform libraries. Instead, users will only have to abide
+    by those restrictions when they upgrade their library's language version.
```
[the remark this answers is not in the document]
kallentu: Also, if we wanted an annotation, it'd be useful to put the details of that here so I can reference it.

accepted/future-releases/class-modifiers/feature-specification.md
eernstg: This excursion (consistently avoid constrants on `on` clauses) turned out to reveal some loopholes. So the `on` clause will now be taken into account when computing 'cannot be implemented locally', which closes the loopholes.

eernstg:
@lrhn, PTAL: We now have the notion of 'direct declared superinterface', and it is used to determine which classes cannot be implemented locally.

About the terminology: A 'direct declared superinterface' is a superinterface which is 'direct' among the 'declared' ones. It is reached in the superinterface graph via a path that contains exactly one edge whose destination is a declared type (that is, not a mixin application).

We could of course consider variants like 'declared direct superinterface', or completely different phrases, but 'direct declared superinterface' seems to make sense, to me at least. ;-)

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -530,62 +578,264 @@ mixinModifier     ::= 'sealed' | 'base' | 'interface' | 'final'
...
+
+    *For pre-feature libraries, we cannot tell if the intent of `class` was
+    "just a class" or "both a class and a mixin". For compatibility, we assume
+    the latter, even if the class is being used as a mixin in a post-feature
+    library where it does happen to be possible to distinguish those two
+    intents.*
+
+*   `K` is a post-feature library, and any of:
+
+    *   `K` is not the same library as `L`,
```
kallentu, line 828: Can we change the spec to talk about how only mixin classes and mixins can be mixed in regardless of library since this mornings meeting? And that its an error for non-mixin classes to be mixed in?
Not sure if I got that right...

lrhn: Should be done now.

Basically, it's a compile-time error to mix in a non-`mixin` `class` unless:
    * that class is declared in a pre-feature library,
    * it has `Object` as superclass, and
    * it declares no constructors.

(The same rules that apply to mixing in a `class` declaration in 2.19, where all libraries are pre-feature.)

There are now no other exceptions, even inside the same library as the class.

PR #2811, opened 2023-02-01
Warn if a switch case is unreachable because prior cases cover it.

munificent:
The exhaustiveness algorithm defines when a case is unreachable, but the proposal didn't actually do anything with that information. This closes that gap.

cc @lrhn @johnniwinther

### DIFF
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -2460,14 +2460,17 @@ All other types are not always-exhaustive. Then:
     other option is to throw an error and most Dart users prefer to catch those
     kinds of mistakes at compile time.*
 
-*   It is a compile-time warning if a case in a switch statement, collection
+*   The language doesn't specify or mandate this, but implementations are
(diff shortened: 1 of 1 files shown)


accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -2458,7 +2460,14 @@ All other types are not always-exhaustive. Then:
     other option is to throw an error and most Dart users prefer to catch those
     kinds of mistakes at compile time.*
 
-[exhaustiveness]: https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/exhaustiveness.md
+*   It is a compile-time warning if a case in a switch statement, collection
```
lrhn, line 2463: I (still) don't think we should specify warnings.

I'd prefer to tell the analyzer devs that they are free, and encouraged, to give a warning in this case, but don't make it part of the language specification.

This is just dead code. We have lots of dead code that we don't specify warnings for.

(Otherwise, phrasing is fine!)

munificent: Good point, done.

PR #2813, opened 2023-02-03
Move warnings to a separate section and suggest some more.

munificent:
cc @scheglov

Follow-up on: https://github.com/dart-lang/co19/issues/1734#issuecomment-1386774555

### DIFF
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: Accepted
 
-Version 2.22 (see [CHANGELOG](#CHANGELOG) at end)
+Version 2.23 (see [CHANGELOG](#CHANGELOG) at end)
 
 Note: This proposal is broken into a couple of separate documents. See also
 [records][] and [exhaustiveness][].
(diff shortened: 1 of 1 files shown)


accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -2460,10 +2460,39 @@ All other types are not always-exhaustive. Then:
...
+would only apply to code that has opted into the language version where this
+ships.
+
+### Warnings
+
+We don't want to mandate warnings in the language specification, but the user
+experience of a language feature is holistic and we try to design it thinking
+about the entire tooling experience. With that in mind, implementations are
+encouraged to report a static warning when:
+
```
eernstg, line 2476: Perhaps introduce the notion of a 'pattern which is guaranteed to match', then mention that `_` and `T _` are always guaranteed to match, and so is a variable pattern or an identifier pattern whose type is a supertype of the static type of the matched value. With those atomic cases in place, the bullet list below can be considered as the "induction step", and they'd just talk about subpatterns as being guaranteed to match.

eernstg (commented):
LGTM, suggesting a small reorganization along the way.

accepted/future-releases/0546-patterns/feature-specification.md
munificent: I don't want to get too far into the business of specifying all of the details of exactly which patterns are or are not useless. :)

PR #2819, opened 2023-02-04
[patterns] Small fixes and clarifications

gnprice:
This is a series of 6 small changes, in separate commits for ease of review.

Three of them fix small errors. The other three are rewordings that I'd like to think make things clearer; perhaps you'll agree.


munificent:
These are excellent, thank you! :D

PR #2825, opened 2023-02-07
[patterns] Clarify when a set of map patterns is exhaustive

### LINKED ISSUE #2818: Exhaustiveness on map patterns
In the patterns spec, at the runtime semantics for a map pattern ([permalink](https://github.com/dart-lang/language/blob/92fa381e5e9913b644561912ffa90d39804d02c8/accepted/future-releases/0546-patterns/feature-specification.md#matching-refuting-and-destructuring)), there's this explanatory remark:

> *An empty map pattern can match only empty maps. Note that this treats a misbehaving map whose `length` is negative as an empty map. This is important so that a set of map patterns that is clearly exhaustive over well-behaving maps will also cover a misbehaving one.*

The wording is very similar to a remark in the list section above it, where it's a reference to #2701. But I think the current spec doesn't actually accomplish this for maps (unlike lists). On the other hand, I'm not sure this property matters much for maps, because (unlike lists) I think there's no common pattern that would lead to a set of map patterns that has any hope of being exhaustive, except using `{...}`.

Here's a set of map patterns that is clearly exhaustive over well-behaving maps:
```dart
String test(Map<bool, int> map) => switch (map) {
    {} => 'unknown',
    {false: _} => 'false',
    {true: _} => 'true',
    {false: _, true: _} => 'contradiction',    
};
```

But if we pass this a map whose `length` reports 1 but whose `[]` and `containsKey` deny containing any keys ([dartpad](https://dartpad.dev/?channel=master&id=55a5691a6d2ae00eaf87a76b8f6a3fd1)), it will not match. The patterns `{false: _}` and `{true: _}` will get past the length check, but then find that the respective keys are absent.

For that matter, if we pass a map whose `length` reports 3, or any number ≥3, this will not match.

It also won't match if the map's `length` reports 2, but its `[]` and `containsKey` only accept one key: e.g. `map[false] != null` but `map[true] == null`. The only patterns that could match such a map would be patterns with duplicate keys:
```dart
    {false: _, false: _} => 'double',
```
but the spec already rightly forbids those.

---

The basic options here are:
1. Accept that no sets of map patterns without `{...}` are exhaustive; users will have to write unnecessary default or `{...}` cases.
2. Come up with a semantics that accepts at least some legitimate cases as exhaustive, while being sound in the face of misbehaving maps.

I think (2) will be hard, and there aren't a lot of legitimate cases it would benefit, so that the right answer will be (1). But perhaps someone will see a clever solution.

The basic reason it's hard while lists are doable is that when a list tells us its length `n`, that also tells us exactly which keys should be valid: 0, 1, 2, …, n-1. We'll go probing some of those keys according to the pattern, and if the list changes its mind and says they're absent then it can only do so by throwing an exception. But when a map tells us its length, then except in unusual cases we still don't know of any key that must be present. So the list could keep denying every key we ask it about, and unless we keep careful track we'll never catch it in a contradiction.

gnprice:
For lists, one can use the empty list pattern `[]` to make an exhaustive set of list patterns.

But for maps, the only way to make an exhaustive set of map patterns is the trivial way, using the pattern `{...}` that's exhaustive (on maps) all by itself.

So move the remark about exhaustive sets of map patterns out of the context of the empty map pattern `{}`, and up to the discussion of `{...}`.  Also say that these are the only exhaustive sets of map patterns, and why.

Fix #2818.

### DIFF
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -3013,6 +3013,12 @@ To match a pattern `p` against a value `v`:
             `{...}`, then any length is allowed, so we don't even ask the map
             for it.*
 
+            *The pattern `{...}` will match all maps (that pass the type test).
+            The only way to make a set of map patterns exhaustive is to include
+            `{...}`, because some misbehaving maps will not match any other map
+            pattern: for example, the map `v` could have a positive `length` but
+            `v[k] == null && !v.containsKey(k)` for every possible key `k`.*
+
         2.  Else let `l` be the length of the map determined by calling `length`
             on `v`.
 
@@ -3035,10 +3041,7 @@ To match a pattern `p` against a value `v`:
             1.  If `l > 0` then the match fails.
 
             *An empty map pattern can match only empty maps. Note that this
-            treats a misbehaving map whose `length` is negative as an empty map.
-            This is important so that a set of map patterns that is clearly
-            exhaustive over well-behaving maps will also cover a misbehaving
-            one.*
+            treats a misbehaving map whose `length` is negative as an empty map.*
 
         *These match failures become runtime exceptions if the map pattern is
         in an irrefutable context.*


lrhn:
An empty list pattern is not exhaustive for lists, you need `[...]` there too.

gnprice:
The empty list pattern on its own isn't exhaustive for lists, but by using it you can make an exhaustive set of list patterns without the pattern `[...]`. Like this:
```dart
T? first<T>(List<T> list) {
  return switch (list) {
    [] => null,
    [var x, ...] => x,
  };
}
```

Letting a set of patterns like that one be exhaustive is what #2701 was about, and why it's important that a misbehaving list with negative length matches `[]`.

That's different from maps, where to make a set of map patterns exhaustive for maps you do need `{...}`.

munificent:
Circling back to this. (Sorry for the delay. My GitHub inbox got away from me.)

We recently decided to change map patterns to [not check length](https://github.com/dart-lang/language/issues/2861). That eliminates `...` from map patterns entirely. We also no longer support empty map patterns (because it's confusing and likely not that useful). [This comment](https://github.com/dart-lang/language/issues/2861#issuecomment-1487789760) has all the details.

Given that, this PR in its current form can't be landed. I'm going to go ahead and close this, but I appreciate effort you put into it.

PR #2847, opened 2023-02-15
[class mods] Fix grammar to prohibit "mixin" on mixin applications.

munificent:
It looks like the prose (which I think is correct) got out of sync with the grammar. This fixes that.

I also tweaked the Markdown formatting a bit because I'm a little obsessive. I'm sorry.

cc @eernstg @kallentu

### DIFF
--- a/accepted/future-releases/class-modifiers/feature-specification.md
+++ b/accepted/future-releases/class-modifiers/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom, Lasse Nielsen
 
 Status: Accepted
 
-Version 1.4
+Version 1.5
 
 Experiment flag: class-modifiers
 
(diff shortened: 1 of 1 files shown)


eernstg (approved):
LGTM, suggesting a few changes to the grammar rules. Generation of a new parser went smoothly, and I'm currently running the spec parser on everything in $SDK/tests/language to see if there are any unintended failures on existing code.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -770,15 +771,16 @@ Define a *trivial generative constructor* to be a generative constructor that:
...
+    no super constructor invocation),
 
 *   has no body (only `;`), and
 
 *   is not `external`. *An `external` constructor is considered to have an
     externally provided initializer list and/or body.*
 
-A trivial constructor may be named or unnamed, and `const` or non-`const`.
-A *non-trivial generative constructor* is a generative constructor which is not a
+A trivial constructor may be named or unnamed, and `const` or non-`const`. A
```
lrhn, line 782: This seemingly trivial change is bad actually a negative for both readability and writeability.

Reflowing markdown is no more necessary than reflowing HTML, the lines will be combined when displayed anwyay, so keeping line breaks at semantically meaningful positions makes everything easier when working with the source.

(Aka: Automatic reflow is bad, m'kay!)

munificent: I know you are logically correct, but it pains my soul to not have the lines nicely wrapped to 80 columns.

I'll undo this change.

PR #2853, opened 2023-02-16
[Patterns] Typo fixed. Boolean replaced by bool

sgrekhov:
There is no `Boolean` in Dart


eernstg (approved):
LGTM

munificent:
These are fine. I use "Boolean" (capitalized but not in `code font`) to refer to the [general concept of a Boolean data type](https://en.wikipedia.org/wiki/Boolean_data_type) in the same way that the proposal uses "number" to refer to instances of numeric types.

I capitalize it because most mathematical terms based on someone's name retain their capitalization in English: Platonic solid, Euclidean geometry, etc. Most references I can find online currently seem to lean towards capitalizing it. (I researched this back when I was writing "Crafting Interpreters".)

But changing some uses of "Boolean" to "`bool`" is probably clearer, so I think this is a net win.

PR #2877, opened 2023-03-01
Fix typo in URI

eernstg:
Turns out that there was a stray Control-T at some point, swapping two characters and thus messing up a URI. This PR fixes it.


eernstg:
Hello Bill, this is just a typo.

kevmoo (approved):

Issue #2723, opened 2022-12-15
[Class Modifiers] Adding `mixin` confuses positive and negative capabilities again


leafpetersen:
The [most recent proposal](https://github.com/dart-lang/language/pull/2722) for capability modifiers adds a `mixin` keyword which allows using a class as a mixin.  Prior to that, we were able to finesse the issue of whether `interface` and `base` were positive or negative capabilities, since you never had to specify more than one of them.  This was convenient, because they are positive sounding keywords, and it is very confusing to have to reason "backwards" from negative capabilities (see discussion in #2595).  The proposal in that state had the property that **if any capabilities were listed, then all of the available capabilities were listed**.

Adding `mixin` as currently specified breaks this model, because the specification makes `interface` and `base` negative capabilities, and makes `mixin` a positive capability.  That means that combining `interface` or `base` with `mixin` requires reasoning about "residual" capabilities.  That is, one says that `base class` is class which can be extended, and `interface class` is one which can be used as an interface.  But a `mixin class` is a class which can be extended, implemented, or mixed in.  So it no longer lists all of its capabilities.  When you add a modifier, you therefore have to start reasoning negatively: adding `interface` **takes away** the ability to extend.  So `interface mixin class` is a class which cannot be extended (but can be used as a mixin or an interface).  This is confusing to start with (switching from positive to negative) but what seems worse is that if I start thinking of these as negative capabilities, then I naturally would expect that since I can have a `base mixin class` and an `interface mixin class`, then I should be able to have a `base interface mixin class` which can only be mixed in or constructed.   This follows naturally: I have a `mixin class`, I can remove the ability implement by adding `interface` and I can remove the ability to extend by adding `base`, so surely I can add both?  The specification currently does not allow this combination syntactically.

I see three possible ways out of this.  

First, we could choose to make `interface`, `mixin` and `base` behave uniformly as monotonic positive capabilities.  That is, a `mixin class` can *only* be mixed in, and you have to say `interface mixin class` if you want to be able to use it as an interface.  You then need to be able to say `interface base mixin class` if you want to allow everything, because plain `class` doesn't allow mixing in anymore. 

Second, we could just choose not to allow `mixin` to be combined with anything else.  Note that:
  - `interface mixin class` (and similarly `interface mixin`) is not very useful.  Stopping someone from extending something while allowing them to still mix it in (which they can do trivially on `Object`) doesn't seem to add anything.
[...]

PR #2889, opened 2023-03-06
Update class modifiers specification.

lrhn:
* Tighten the modifiers implied on anonymous mixin class applications.
* Claim that `enum` declarations are `final`.
* Allow pre-feature libraries to mix in platform no-mixin classes if they could before.
* A little rephrasing, to allow reusing a concept.

But do not allow `interface mixin class` etc. anyway.

@munificent @kallentu

### DIFF
--- a/accepted/future-releases/class-modifiers/feature-specification.md
+++ b/accepted/future-releases/class-modifiers/feature-specification.md
@@ -517,9 +517,7 @@ Many combinations don't make sense:
     so it's redundant to combine with `final`, `base`, or `interface`.
 =======
 *   `sealed` types cannot be extended or implemented, so it's redundant to
-    combine with `final`.
-*   `sealed` types cannot be extended so it contradicts `base`.
-*   `sealed` types cannot be implemented, so it contradicts `interface`.
+    combine with `final`, `base`, or `interface`.
 *   `sealed` types cannot be mixed in outside of their library, so it
     contradicts `mixin` on a class. *It's useful to allow `sealed` on a mixin
     declaration because the mixin can be applied within the same library.
@@ -855,11 +853,14 @@ intents.*
 
 ### Enum classes
 
-The class introduced by an `enum` declaration is considered `final`.
+The class introduced by an `enum` declaration is considered `final` for any purpose
+where a class modifier is required.
 
-Since the class cannot have any subclasses, the modifier does not prevent
-any otherwise allowed operation, and adding it ensures that the enum class
-satisfies any requirements introduced by its super-interfaces.
+The behavior of `enum` declarations is unchanged. Since an `enum` class cannot have
+any subclasses, the modifier would not prevent any otherwise allowed operation.
+The implicit `final` also does not imply fewer restrictions than `enum` declarations
+would otherwise have. The modifier is applied only to automatically satisfy any
+requirements introduced by super-interfaces.
 
 ### Anonymous mixin applications
 
@@ -915,8 +916,8 @@ the anonymous class can mostly be ignored, since it satisfies all possible requi
 of its superclasses, while still propagating those requirements to its subclass.
 Treating the anonymous class as having these modifiers allows the algorithms
 used to check that restrictions are satisfied, and for the “reopen” lint described below,
-to treat the anonymouse mixin application class as any other class, without needing
-special cases.
+to treat the anonymous mixin application class as any other class, without needing
+special cases, and without adding any new restrictions..
 
 ### `@reopen` lint
 
@@ -962,7 +963,7 @@ non-breaking.
     declarations in pre-feature libraries can ignore *some*
     `base`, `interface` and `final` modifiers on *some* declarations
     in platform libraries, and to mix in non-`mixin` classes from platform libraries,
-    as long as those classes has `Object` as superclass and declares no constructors.
+    as long as such a class has `Object` as superclass and declares no constructors.
     Instead, users will only have to abide by those restrictions
     when they upgrade their library's language version.
     _It will still not be possible to, e.g., extend or implement the `int` class,
--- a/accepted/future-releases/class-modifiers/feature-specification.md
+++ b/accepted/future-releases/class-modifiers/feature-specification.md
@@ -500,37 +500,30 @@ class UsesAsMixin extends OtherSuperclass with Both {} // OK.
 ## Syntax
 
 This proposal builds on the existing sealed types proposal so the grammar
-includes those changes. The full set of modifiers that can appear before a class
-or mixin declaration are `abstract`, `sealed`, `base`, `interface`, `final`, and
-`mixin`.
+includes those changes. The full set of modifiers that can appear before a class declaration are `abstract`, `sealed`, `base`, `interface`, `final`, and
+`mixin`. Only the `base` modifier can appear before a `mixin` declaration.
 
 *The modifiers do not apply to other declarations like `enum`, `typedef`, or
 `extension`.*
 
-Many combinations don't make sense:
+Some combinations don't make sense:
 
 *   `base`, `interface`, and `final` all control the same two capabilities so
     are mutually exclusive.
 *   `sealed` types cannot be constructed so it's redundant to combine with
     `abstract`.
-*   `sealed` types cannot be extended or implemented, so it's redundant to
-    combine with `final`, `base`, or `interface`.
-*   `sealed` types cannot be mixed in outside of their library, so it
-    contradicts `mixin` on a class. *It's useful to allow `sealed` on a mixin
-    declaration because the mixin can be applied within the same library.
-    A `sealed mixin class`  does not provide any significant extra
-    functionality over a `sealed mixin`, you can replace `extends MixinClass`
-    with `with Mixin`, so a `sealed mixin class` is not allowed.*
-*   `interface` and `final` classes would prevent a mixin class from being used
-    as a superclass or mixin outside of its library. *Like for `sealed`, an
-    `interface mixin class` and `final mixin class` are not allowed, and
-    `interface mixin` and `final mixin` declaration are recommended instead.*
+*   `sealed` types cannot be mixed in, extended or implemented,
+    so it's redundant to combine with `final`, `base`, or `interface`.
 *   `mixin` as a modifier can obviously only be applied to a `class`
-    declaration, which makes it also a `mixin` declaration.
+    declaration, which makes it also a introduce a mixin declaration.
 *   `mixin` as a modifier cannot be applied to a mixin-application `class`
     declaration (the `class C = S with M;` syntax for declaring a class). The
     remaining modifiers can.
+*   A `mixin` is intended to be mixed in, so it cannot be combined with an
+    `interface`, `final` or `sealed` modifier.
 *   Mixin declarations cannot be constructed, so `abstract` is redundant.
+
+_An `interface mixin class` seems redundant, because it disallows mixing in the declaration, so the `mixin` makes no difference for the AP. However, a `mixin` is required in order to mix in the declaration inside the same library, where the `interface` modifier’s restriction can be ignored. From outside the library, there is no distinction between `interface mixin class` and `interface class`, or even between `abstract interface mixin class`, `abstract interface class` and `interface mixin` declarations, as the following table also shows._
 
 The remaining valid combinations and their capabilities are:
 
@@ -547,8 +540,13 @@ The remaining valid combinations and their capabilities are:
 |`abstract final class`     |No     |No     |No     |No     |No     |
 |`mixin class`              |**Yes**|**Yes**|**Yes**|**Yes**|No     |
 |`base mixin class`         |**Yes**|**Yes**|No     |**Yes**|No     |
+|`interface mixin class`    |**Yes**|No     |**Yes**|No     |No     |
+|`final mixin class`        |**Yes**|No     |No     |No     |No     |
+|`sealed mixin class`       |No     |No     |No     |No     |**Yes**|
 |`abstract mixin class`     |No     |**Yes**|**Yes**|**Yes**|No     |
 |`abstract base mixin class`|No     |**Yes**|No     |**Yes**|No     |
+|`abstract interface mixin class`|No |No    |**Yes**|No     |No     |
+|`abstract final mixin class`|No    |No     |No     |No     |No     |
 |`mixin`                    |No     |No     |**Yes**|**Yes**|No     |
 |`base mixin`               |No     |No     |No     |**Yes**|No     |
 |`interface mixin`          |No     |No     |**Yes**|No     |No     |
@@ -656,20 +654,29 @@ It is a compile-time error to:
     class C3 with SM {} // Error.
     ```
 
-A typedef cannot be used to subvert these restrictions or any of the
-restrictions below. When extending, implementing, or mixing in a typedef, we
-look at the library where class or mixin the typedef resolves to is defined to
-determine if the behavior is allowed. *Note that the library where the _typedef_
-is defined does not come into play. Typedefs cannot be marked with any of the
+A type alias (`typedef`) cannot be used to subvert these restrictions or any of the
+restrictions below. When extending, implementing, or mixing in a type alias, we
+look at the library where class or mixin the type alias resolves to is defined to
+determine if the behavior is allowed. *Note that the library where the _type alias_
+is defined does not come into play. Type aliases cannot be marked with any of the
 new modifiers.*
 
 ### Disallowing implementation
 
-It is a compile-time error if a subtype of a declaration marked `base` or
+We say that an interface _prevents implementation_ if its declaration is marked
+`base` or `final`, or if it has any (immediate) superinterface which prevents
+implementation.
+_(This definition is inherently transitive, so it has no effect to add or remove_
+_the “immediate”.)_
+
+It’s a compile-time error if a declaration’s interface prevents implementation,
+and the declaration is not marked `base`, `final` or `sealed`.
+
+_Effectively, it is a compile-time error if any subtype of a declaration marked `base` or
 `final` is not marked `base`, `final`, or `sealed`. This restriction applies to
 both direct and indirect subtypes and along all paths that introduce subtypes:
 `implements` clauses, `extends` clauses, `with` clauses, and `on` clauses. This
-restriction applies even to types within the same library.
+restriction applies even to types within the same library._
 
 *Once the ability to use as an interface is removed, it cannot be reintroduced
 in a subtype. If a class is marked `base` or `final`, you may still implement
@@ -681,15 +688,15 @@ Further, while you can ignore some restrictions on declarations within the same
 library, you cannot use that to ignore restrictions inherited from other
 libraries.
 
-We say that `S` is a _direct declared superinterface_ of a class, mixin, or
+We say that a declaration `S` is a _direct declared superdeclaration_ of a class, mixin, or
 mixin class declaration `D` if `D` has a superclass clause of the form
 `C with M1 .. Mk` (where `k` may be zero when there is no `with` clause)
-and `S` is `C`, or `S` is `Mj` for some `j` in 1 .. k,
-or if `D` has an `implements` or `on` clause and `S` occurs as one of the operands of
-that clause.
+and `S` is is the declaration denoted by `C`, or by `Mj` for some `j` in 1 .. k,
+or if `D` has an `implements` or `on` clause and `S` is the declaration denoted by
+one of the operands of such a clause.
 
 We then say that a class or mixin declaration `D` *cannot be implemented locally* if it
-has a direct declared superinterface `S` such that:
+has a direct declared superdeclaration `S` such that:
 
 *   `S` is from another library than `D`, and `S` has the modifier `base`,
     `final` or `sealed`, or
@@ -732,7 +739,7 @@ Otherwise, `D` can be implemented locally.
 It is a compile-time error if:
 
 *   A class, mixin, or mixin class declaration `D` has an `implements` clause
-    where `S` is an operand, and `S` is a class, mixin, or mixin class
+    where `S` is an operand, and `S` denotes  a class, mixin, or mixin class
     declaration declared in the same library as `D`, and `S` cannot be
     implemented locally.
 
@@ -836,6 +843,14 @@ the latter, even if the class is being used as a mixin in a post-feature
 library where it does happen to be possible to distinguish those two
 intents.*
 
+### Enum classes
+
+The class introduced by an `enum` declaration is considered `final`.
+
+Since the class cannot have any subclasses, so the modifier does not prevent
+any otherwise allowed operation, and adding it ensures that the enum class
+satisfies any requirements introduced by its super-interfaces.
+
 ### Anonymous mixin applications
 
 An *anonymous mixin application* class is a class resulting from a mixin application
@@ -848,26 +863,50 @@ An anonymous mixin application class cannot be referenced anywhere except in
 the context where the application occurs, so its only role is to be a superclass of
 another class in the same library.
 
-To ensure reasonable and correct behavior, we infer class modifiers on anonymous
-mixin application classes as follows.
+To ensure reasonable and correct behavior, without having to special-case such
+anonymous mixin application classes elsewhere, we infer class modifiers as follows.
 
-Let *C* be an anonymous mixin application with superclass *S* and mixin *M*. Then:
+We define the following transitive property on declarations:
 
-* If any of *S* or *M* has a `sealed` modifier, *C* is has a `sealed` modifier.
-* Otherwise:
-  * *C* is `abstract`, and
-  * If either of *S* or *M* has a `base` or `final` modifier, then *C* has a `final` modifier.
+* A declaration *prohibits inheritance* _(of its implementation)_ if:
 
-_We do not distinguish whether *S* or *M* has `base` or `final` modifiers.
-The modifier on *C* is there to satisfy the requirement that a subtype of a `base` or `final`
-declaration is itself `base`, `final` or `sealed`. The anonymous mixin application class will
-be immediately extended inside the same library, which is allowed by both `base` and `final`.,
-and will not be used for anything else._
+  * the declaration is marked `interface` or `final`, or
+  * the declaration extends or mixes in a declaration which prohibits inheritance.
 
-Adding `sealed` to an anonymous mixin application class with only one subclass ensures
-that the subclass extending the mixin application class can be used
-in exhaustiveness checking of the sealed superclass.
-This is necessary since the anonymous mixin application class itself cannot be referenced.
+Let then *C* be an anonymous mixin application occurring in a post-feature library, with
+superclass *S* and mixin *M*.
+
+* If either *S* or *M* has a `sealed` modifier, then *C* has an implicit `sealed` modifier.
+
+* Otherwise *C* is `abstract`, and
+
+  * If *C* *prevents implementation* and *prohibits inheritance*,
+    then *C* has an implicit `final` modifier.
+
+  * Otherwise if *C* *prevents implementation*, then *C* has an implicit `base` modifier.
+
+  * Otherwise if *C* *prohibits inheritance*, then *C* has an implicit `interface` modifier.
+
+  * Otherwise *C* has no modifier.
+
+Adding `sealed` to an anonymous mixin application class, which always has precisely
+one subclass, ensures that the subclass can be used in exhaustiveness checking
+of the sealed superclass.
+
+Adding `final` or `base` satisfies the requirement that a subtype of a `base` or `final`
+declaration is itself `base`, `final` or `sealed`.
+
+Adding `interface` ensures that the mixin application class doesn’t remove a restriction
+on the implementation inherited from *S* or *M*. This makes the “reopen” lint described below
+easier to implement.
+
+Adding these modifiers on the anonymous mixin application class ensures that
+the anonymous class can mostly be ignored, since it satisfies all possible requirements
+of its superclasses, while still propagating those requirements to its subclass.
+Treating the anonymous class as having these modifiers allows the algorithms
+used to check that restrictions are satisfied, and for the “reopen” lint described below,
+to treat the anonymouse mixin application class as any other class, without needing
+special cases.
 
 ### `@reopen` lint
 
@@ -917,8 +956,14 @@ non-breaking.
     when they upgrade their library's language version.
     _It will still not be possible to, e.g., extend or implement the `int` class,
     even if will now have a `final` modifier._
+    Going through a pre-feature library does not remove transitive restrictions
+    for code in post-feature libraries. Any post-feature library declaration which has a
+    platform library class marked `base` or `final` as a superinterface must be marked
+    `base`, `final` or `sealed`, and cannot be implemented locally,
+    even if the superinterface chain goes through a pre-feature library declaration,
+    and even if that declaration ignores the `base` modifier.
 
-    This is a special case behavior only available to platform libraries.
+    This is special case behavior only available to platform libraries.
     Package libraries should use versioning to to introduce breaking
     restrictions instead, and those libraries can then rely on the restrictions
     being enforced.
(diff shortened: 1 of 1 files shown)


accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -850,26 +864,50 @@ An anonymous mixin application class cannot be referenced anywhere except in
...
+on the implementation inherited from *S* or *M*. This makes the “reopen” lint described below
+easier to implement.
+
+Adding these modifiers on the anonymous mixin application class ensures that 
+the anonymous class can mostly be ignored, since it satisfies all possible requirements
+of its superclasses, while still propagating those requirements to its subclass. 
+Treating the anonymous class as having these modifiers allows the algorithms
+used to check that restrictions are satisfied, and for the “reopen” lint described below,
+to treat the anonymouse mixin application class as any other class, without needing
+special cases.
```
eernstg, line 910: I like to think about this treatment as the one which will implicitly imply a set of modifiers on an anonymous mixin application and on a `sealed` declaration such that they would not incur a compile-time error and would also not cause an `implicit_reopen` lint, in the case where they had been regular (named, non-sealed) declarations.

In other words, the implicit class modifiers serve to propagate the constraints faithfully. This means that it is easy to check for re-opening declarations by looking at their direct superinterfaces. There is no need to search the entire superinterface graph.

I think the text above says most of this, but it doesn't mention that the set of implicit class modifiers are specifying the _most permissive set of restrictions that will not re-open anything_.

Wouldn't it be helpful to say that?

eernstg (approved):
LGTM! A couple of comments added.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -549,8 +541,13 @@ The remaining valid combinations and their capabilities are:
 |`abstract final class`     |No     |No     |No     |No     |No     |
 |`mixin class`              |**Yes**|**Yes**|**Yes**|**Yes**|No     |
 |`base mixin class`         |**Yes**|**Yes**|No     |**Yes**|No     |
+|`interface mixin class`    |**Yes**|No     |**Yes**|No     |No     |
```
kallentu, line 544: Could you help update the grammar below?

leafpetersen:
For the record, I continue to think this is a mistake, for all the reasons laid out in https://github.com/dart-lang/language/issues/2723 .

leafpetersen (changes requested):
I am very much not on board with this change.  If we actually feel that we need to ability to have variants of these that can be mixed in internally but not externally, we should instead remove the restriction that you cannot mix in a class which is not marked as `mixin` inside of your own library.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -683,26 +689,26 @@ Further, while you can ignore some restrictions on declarations within the same
 library, you cannot use that to ignore restrictions inherited from other
 libraries.
 
-We say that `S` is a _direct declared superinterface_ of a class, mixin, or
+We say that a declaration `S` is a _direct declared superdeclaration_ of a class, mixin, or
```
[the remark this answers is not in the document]
lrhn: Done.

accepted/future-releases/class-modifiers/feature-specification.md
lrhn: Probably would, yes. "It propagates the restrictions of its super-declarations, and odes not add any further", or similar.



I'd generally like to rephrase everything in terms of semantic properties of classes, rather than adding syntactic modifiers that aren't there in the original program.
(I'm a denotational semanticist at heart.)

Say "prevents implementation". A declaration prevents implementation if:
* It's marked `final` or `base`.
* It's an `enum` declaration.
* It has an immediate superinterface which prevents implementation.

(This is the easy one).
Then we say that it's a compile-time error to implement a declaration which prevents implementation.
And that it's a compile-time error if a `class` or `mixin` declaration prevents implementation,
is declared in a post-feature library, and is not marked `base`, `final` or `sealed`.

But we don't have to actually add `final` or `base`, or even figure out which one an anonymous 
mixin application should get.

Likewise, we can say that a class declaration prevents instantiation/cannot have instances if
* it's marked `abstract`,
* it's marked `sealed`, or
* it's an anonymous mixin application class

Rather than adding `abstract` to `sealed` classes, we lift it to the semantic domain and add the semantic property to both `abstract`-declared classes and sealed classes.

It'll just require rewriting everything. Noooo problem. :)

lrhn: But done.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -909,18 +951,25 @@ non-breaking.
...
     "Ignorance of the law is no defense."*
 
-*   We would like to add modifiers to some classes in platform (i.e. `dart:`)
+*   We would like to add modifiers to some classes in platform (i.e., `dart:`)
     libraries when this feature ships. But we would also like to not immediately
     break existing code. To avoid forcing users to immediately migrate,
     declarations in pre-feature libraries can ignore *some*
     `base`, `interface` and `final` modifiers on *some* declarations
-    in platform libraries.
+    in platform libraries, and can mix in non-`mixin` classes from platform libraries,
```
leafpetersen, line 959: Discussion of interactions between language versions are split between the section titled "Mixin restrictions" and here.  Can they be unified?  It took me a long time to realize that the things addressed in the other section were in fact addressed, and then an even longer period of flipping back and forth between the two areas to convince myself that all of the combinations were covered.  Combining them in one place would make it much clearer.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -935,6 +984,15 @@ needed.
 
 ## Changelog
 
+1.6
+
+* Update the modifiers applied to anonymous mixin applications to closer
+  match the superclass/mixin modifiers.
+* State that `enum` declarations count as `final`.
+* Some rephrasing to allow concept reuse.
+* Say that pre-feature libraries can mix in  non-`mixin` platform library classes
```
leafpetersen, line 993: Maybe add link to https://dart-review.googlesource.com/c/sdk/+/287665 here?

leafpetersen (approved):

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -631,127 +624,82 @@ It is a compile-time error to:
...
-restrictions below. When extending, implementing, or mixing in a typedef, we
-look at the library where class or mixin the typedef resolves to is defined to
-determine if the behavior is allowed. *Note that the library where the _typedef_
-is defined does not come into play. Typedefs cannot be marked with any of the
-new modifiers.*
+We say that an interface _prevents implementation_ if it has *any*
+superinterface whose declaration is marked `base` or `final`.
 
-### Disallowing implementation
+It is a compile-time error if a the interface of a
```
leafpetersen, line 645: "a the" -> "the"

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -631,127 +624,82 @@ It is a compile-time error to:
...
-    where `S` is an operand, and `S` is a class, mixin, or mixin class
-    declaration declared in the same library as `D`, and `S` cannot be
-    implemented locally.
-
-    ```dart
-    // a.dart
-    base class B {}
+It's a compile-time error if a `class`, `mixin`, `mixin class` or `enum`
+declaration in library *L* has an `implements` clause with an entry
+denoting the interface *S*, and *S* prevents implementation locally in *L*.
```
leafpetersen, line 674: I think by "prevents implementation locally" you mean "cannot be implemented locally"?  Or do you meant to change the definition above to define "prevents implementation locally"?

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -631,127 +624,82 @@ It is a compile-time error to:
...
+
+// b.dart
+import 'a.dart';
+
+// These are valid, but cannot be implemented locally:
+sealed class DE extends S {}
+final class DM with S {}
+base mixin MO on S {}
+
+// None of these interface can be implemented locally,
```
leafpetersen, line 695: "these interface" -> "these interfaces"

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -761,25 +709,26 @@ versioning and backwards compatibility.
 
 Currently, a class may only be used as a mixin if it has a default constructor.
 This prevents the class from defining a `const` constructor or any factory
-constructors. We loosen this somewhat.
+constructors. We loosen this restriction for `mixin class`es.
```
leafpetersen, line 712: `mixin class`es renders oddly, maybe reword to `mixin class` declarations?

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -836,38 +785,82 @@ the latter, even if the class is being used as a mixin in a post-feature
...
+purpose where a class modifier is required.
+
+The behavior of `enum` declarations is unchanged. Since an `enum` class cannot
+have any subclasses, the implicit `final` modifier would not prevent any
+otherwise allowed operation.
+The implicit `final` also does not imply fewer restrictions than `enum`
+declarations would otherwise have. The modifier is applied only to automatically
+satisfy any requirements introduced by super-interfaces.
+
 ### Anonymous mixin applications
```
leafpetersen, line 800: This section is super confusing.  I *believe* that this has no relevance to end users (or even implementers per se, at least for the static semantics) because it purely is about how the de-sugaring of mixins works, correct?  That is, there is no information an end user needs to know about here, and no errors to check for here.  Assuming that is correct, I think it would really help readers of this spec (i.e. our implementation teams) to have that context.  I'd say something at the start of this section like: "The specification of the semantics of mixins is given in terms of a rewrite in terms of a sequence of anonymous mixin classes.  The following section describes how to do that rewrite in a way that satisfies the restrictions described in this feature spec.  This section is purely informational, and specifies no new user facing errors or warnings".

leafpetersen (approved):
This looks great, thank you!

accepted/future-releases/class-modifiers/feature-specification.md
lrhn: I think we can remove the section, if we are *very careful* to not say something like "Every subclass of a `base` class is marked `base` or ...", because an anonymous mixin application is not a declaration and is therefore not marked anything.
And for `sealed`, it's even more important to get things just right.

What I really want is to to introduce a semantic abstraction layer, so our semantics don't have to talk directly about syntax.
I want to say something like: 

> An interface is *sealed* if it's the interface of a declaration marked `sealed` or if it's an anonymous mixin application whose superclass or mixin's interface is *sealed*.

and then, in the definition of exhaustiveness, never ask about the `sealed` modifier, only the *sealed* semantic property.

Also:

> An interface _prevents implementation_ if it is the interface of a declaration marked `base` or `final`,
> or if it has any superinterface which _prevents implementation_.
> An interface _prevents local implementation in library L_ if it has a superinterface which prevents 
> implementation, and that superinterface's declaration is in a library other than *L*.
> A class or mixin _prohibits inheritance_ if its declaration is marked `final` or `base`.

Then the errors become:
* It's a compile-time error if a class, mixin, or mixinclass declaration has 
   an interface which prevents implementation, and the declaration is not marked `base`, `final` or `sealed`.
* It's a compile-time error to extend a class or mix-in a mixin which prohibits inheritance,
   or whose interface is sealed, in another library than where the class or mixin is declared.
* It's a compile-time error to implement an interface in library *L* if
    * the interface prevents implementation in *L*, or if
    * the interface is sealed and not declared in *L*.
* It's a compile-time error for a mixin to have an `on` type interface which is sealed and
   declared in another library that the mixin.

This all works without having to consider whether an anonymous mixin application has a modifier 
or not. It has a declaration, but its interface is not "the interface" of that declaration.
All these little distinctions between declaration, declaration of, interface of a (class or declaration),
I want to get them *right*, because otherwise the specification risks being subtly ambiguous.


This section is ... doing something intermediate, which is probably why it's confusing.
It introduces the semantic properties, then it uses that to add synthetic modifiers to things that are not really declarations, because the rest of the specification *is* talking about things in terms of, say, classes having modifiers (they don't, their declaration might, but anonymous mixin applications don't have the modifiers of their declaration), and I want to be compatible with the rest.

(We cannot just say that the anonymous mixin application have the modifiers of the declaration it's in, because
then the following would not make `S&M` be `sealed`, and then it's not obvious that `case A()` would help exhausting `S`.
```dart
sealed class S {}
class A extends S with M {}
```
so ... need to be precise!)

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -836,38 +785,82 @@ the latter, even if the class is being used as a mixin in a post-feature
...
+the context where the application occurs, so its only role is
+to be a superclass of another class in the same library.
+
+To ensure reasonable and correct behavior, without having to special-case such
+anonymous mixin application classes elsewhere, we infer class modifiers as
+follows.
+
+First we define the following transitive property on declarations:
+
+* A declaration *prohibits inheritance* _(of its implementation)_ if:
```
eernstg, line 820: Does 'transitive' mean that it is possible for an anonymous mixin application to prohibit inheritance?

In that case, an anonymous mixin application is considered to be a 'declaration', and it 'is marked' with the implicitly induced modifier.

Otherwise, it's impossible for *C* in line 834 to 'prohibit inheritance'.

eernstg (commented):
I Indicated a handful of typo level issues (syntax errors in examples and such).

More importantly, the definition of 'prohibits inheritance' talks about 'declarations', and the actual wording of subsequent lines does not make sense unless we count each anonymous mixin application as a 'declaration' and consider that declaration to be 'marked with' each of its implicitly induced modifiers.

However, I still think the rules about anonymous mixin applications are simpler to understand if they are expressed directly in terms of the class modifiers (`sealed` as well as `final`/`base`/`interface`/none, explicitly specified as well as implicitly induced), so I inserted a 'suggestion' about how that could be done.

accepted/future-releases/class-modifiers/feature-specification.md
leafpetersen: I'm more lost than I started.  I honestly have no idea what you're both trying to accomplish with this section.  I completely believe that there is a reason for it, but ... I don't know what it is, and so I'd suggest that probably no other readers do either.  I *think* (but am no longer sure?) that this section is intended to describe (in an indirect way) what modifiers you would put on the implicitly generated classes that you would generate if you were to follow the desugaring semantics of mixins, instead of implementing a direct (static or runtime) semantics.  I suspect however that neither of our implementations use the desugaring for the static semantics, and I believe that backends can choose whether or not to use the desugaring for the dynamic semantics (and in any case, this is irrelevant for the dynamic semantics).  So I believe that this section is entirely irrelevant except to make the desugaring semantics used elsewhere in the spec correct.  Is that belief true?  If not, it's *really important* that we clarify how it is not.  Because if this is intended to *add* some static errors, I don't know what they are (and hence I suspect none of the folks responsible for implementing this know either).

leafpetersen: If my hypothesis above is correct, that this is just about the desugaring, why not just explain it that way directly?  That is, give the desugaring, and say something like:

```
In either case, let $K$ be a class declaration with
the same constructors, superclass, interfaces and instance members as
the defined class.  Then $K$ has the class modifier <foo> if <bar>
```

??

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -631,127 +624,82 @@ It is a compile-time error to:
...
-It is a compile-time error if a subtype of a declaration marked `base` or
-`final` is not marked `base`, `final`, or `sealed`. This restriction applies to
-both direct and indirect subtypes and along all paths that introduce subtypes:
+_Effectively, it is a compile-time error if any subtype of a declaration marked
+`base` or `final` is not also marked `base`, `final`, or `sealed`.
+This restriction applies to both direct and indirect subtypes
+and along all paths that introduce subtypes:
 `implements` clauses, `extends` clauses, `with` clauses, and `on` clauses. This
-restriction applies even to types within the same library.
+restriction applies even to types within the same library._
```
munificent, line 656: After several readings, I get what three previous paragraphs are saying, but I find the name and the use of "interface" pretty confusing. An "interface" that you're prevented implement is... what how is that an "interface" at all?

The fact that the third paragraph is needed to translate it to something comprehensible is probably telling. :)

How about:

> We say that a type is *unimplementable* if any of its supertypes are marked `base` or `final`.
>
> It is a compile-time error if a `class`, `mixin`, or `mixin class` declaration is unimplementable and doesn't itself have a `base`, `final` or `sealed` modifier.  _(Which must be a `base` modifier for `mixin` and `mixin class` declarations.)_
>
> _This restriction applies along all paths that introduce subtypes: `implements` clauses, `extends` clauses, `with` clauses, and `on` clauses. This restriction applies even to types within the same library._

?

I understand that "interface" is more spec-like terminology for "static type-y thing that has declarations", but it's a really confusing term in this context and something less formal but still accurate might be clearer, especially in a proposal that is literally adding an `interface` modifier.

accepted/future-releases/class-modifiers/feature-specification.md
munificent: If there's a superinterface *S* whose declaration is marked `base` or `final`, then according to the *prevents implementation* rule, every subtype of *S* will also be marked `base` or `final`.

Given that, I think it's sufficient to say:

> It's a compile-time error if a `class`, `mixin`, `mixin class` or `enum` declaration has an `implements` clause with an entry referring to a type declared in another library and that type is marked `base` or `final`.

I don't think we need to define any transitive "cannot be implemented locally" rule at all. The fact that `base` or `final` is already required to percolate down the entire subtype graph means we can rely on it being in any immediate supertype.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -631,127 +624,82 @@ It is a compile-time error to:
...
+and `D` does not have a `base`, `final` or `sealed` modifier.
+_(Which must be a `base` modifier for `mixin` and `mixin class`
+declarations.)_
 
-It is a compile-time error if a subtype of a declaration marked `base` or
-`final` is not marked `base`, `final`, or `sealed`. This restriction applies to
-both direct and indirect subtypes and along all paths that introduce subtypes:
+_Effectively, it is a compile-time error if any subtype of a declaration marked
+`base` or `final` is not also marked `base`, `final`, or `sealed`.
+This restriction applies to both direct and indirect subtypes
```
munificent, line 653: I don't think you need to look at indirect paths. If any indirect supertype of your type prevents implementation, then all of the subtypes between that supertype and you also have to be marked `base` or `final` according to the rule. Therefore, you'll always have an immediate supertype marked `base` or `final` telling you what to do.

I guess there's a question of whether an implementation is required to report an error at every subtype if an intermediate one is missing the modifier, but even then, I think the best user experience is to *only* report the error at the first supertype that is missing the modifier.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -805,9 +754,9 @@ class C {
 
 It's a compile-time error if:
 
-*   A `mixin class` declaration has a superclass other than `Object`. *The
-    declaration is limited to an `extends Object` clause or no `extends` clause,
-    and no `with` clauses. The class grammar prohibits `on` clauses.*
+*   A `mixin class` declaration has a superclass clause.
```
munificent, line 757: What is a "superclass clause"? Do you mean "`extends` clause"?

accepted/future-releases/class-modifiers/feature-specification.md
munificent: My radical opinion is that the entire notion of "anonymous mixin application classes" is an unnecessary attempt to cram a trait-like behavior square peg into a Smalltalk everything-is-a-class single inheritance round hole and causes more trouble than its worth in the spec.

I'm fairly certain there is a simpler way to incorporate mixins into the language without resorting to conjuring up a synthetic invisible class every time you use one. (And, also, mixin application classes are a useless feature as far as users are concerned.)

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -836,38 +785,82 @@ the latter, even if the class is being used as a mixin in a post-feature
...
+`base` or `final` declaration is itself `base`, `final` or `sealed`.
+
+Adding `interface` ensures that the mixin application class doesn’t
+remove a restriction on the implementation inherited from *S* or *M*.
+This makes the “reopen” lint described below easier to implement.
+
+Adding these modifiers on the anonymous mixin application class ensures that
+the anonymous class can mostly be ignored, since it satisfies all possible
+requirements of its superclasses, while still propagating those requirements
+to its subclass.
```
munificent, line 859: That's a whole lot of specification just to reach the point where anonymous mixin applications can now disappear back into the shadows where they belong. :)

accepted/future-releases/class-modifiers/feature-specification.md
eernstg: I don't see any particular reason to connect the rules about implicit class modifiers of mixin applications with the desugaring semantics (that's the kind of semantics we actually have in the language specification; surely we could have others, but I don't think that would make any difference here).

The way I see it could be illustrated by the following two graphs. First, the underlying code:

```dart
interface class A {}
interface class B {}
mixin M implements A {}
sealed class C extends B with M {}
class D extends C {}
```

Next comes the superinterface graph, showing the explicit annotations (the dotted arrow is `implements`, the black arrow is `extends`):

```mermaid
graph BT;
  D --> C[sealed C]
  C --> BwM[B with M]
  BwM --> B[interface B]
  BwM -.-> M
  M -.-> A[interface A]
```

Actually, we had `A`, `B`, `C`, and `M` already, and I'm currently writing `D`. So I'm wondering, can I do it as shown above (`class D extends C {}`), or would that be an implicit reopening? In order to be able to reason about the code that I'm writing (as opposed to "try all combinations of keywords until there is no error"), I'd like to explain what's going on.

The answer is that it _is_ an implicit reopening, so we'll get the `implicit_reopen` lint.

Let's just say that we don't want that. Would it help to remove `interface` from `A`? How about `B`?

Here is the superinterface graph again, with implicitly induced modifiers:

```mermaid
graph BT;
  D --> C[sealed (interface) C]
  C --> BwM[(interface) B with M]
  BwM --> B[interface B]
  BwM -.-> M
  M -.-> A[interface A]
```

I'd argue that we can now use some simple rules to reason about the situation: Implicitly induced modifiers are always sufficient to avoid reopening, so we don't need to worry about them. However, `interface` should always be propagated across black arrows (`extends`).

Inspecting the graph, we can see that `interface` is not propagated from `C` to `D`, and that's the reason why we get the lint. I'd argue that the same thing is not entirely as obvious if we look at the previous graph.

Returning to the question: Would it help if we remove `interface` from `A`? In this case we can immediately see that no `interface` modifier is present on `M` (we don't need to propagate `interface` across a dotted line). This implies that we can't solve the problem by removing `interface` from `A`, the graph stays the same everywhere else.

However, if we remove `interface` from `B` then we get the following:

```mermaid
graph BT;
  D --> C[sealed C]
  C --> BwM[B with M]
  BwM --> B
  BwM -.-> M
  M -.-> A[interface A]
```

We can see that there is no reopening any more, because `C` isn't implicitly `interface` any more&mdash;so there's no need to propagate it, and hence `D` can be a plain class.

I find it easier to understand how the classes and their modifiers interact if we allow ourselves to show all of them (explicit + implicit), such that we can answer every question about possible usages of the given declaration, and questions about reopening, using local information only.

If we insist that only the information in the first graph can be shown to the world, and developers out there can't even learn the rules about how the implicit modifiers are computed for mixin applications, then I honestly think it gets harder to answer questions like the ones I mentioned above.

lrhn: > My radical opinion is that the entire notion of "anonymous mixin application classes" is an unnecessary attempt to cram a trait-like behavior square peg into a Smalltalk everything-is-a-class single inheritance round hole and causes more trouble than its worth in the spec.

That may be, but it's what we have today, and changing it is potentially breaking (because our naive LUB depends on type hierarchy depth). 
At least our current model is *simple*.

Now, about how to phrase things.

We should decide on the goal for this specification:
* Is it about being precise and formal, or
* Is it intended for end-users to read?

I'm going for formal. It should still be readable and understandable to implementors (and it may be failing that), but end-users should read a carefully hand-crafted description of the feature instead.

That also means that the phrasing should be aimed at implementors, so saying up front how every section applies to implementations is a good idea. That's missing here, where the specification is more like a free-floating abstract specification. It may be mathematically sound, but not very actionable. 
Note taken!



I want to avoid desugaring semantics. Desugaring semantics always leaves *me* scratching my head about what they mean, because it's unclear what status the desugared syntax has, and it seems that's often unclear to the author who wrote it too. (Even if that was me.) 
Is the original code removed from the program, and the desugaring put in its place? Does this happen before, during or after type inference? Does the new code have properly assigned context types and static types? (If so, which?) Very often running type inference on the desugared code again would give a different result.
What if the desugared code has an error, does that mean the original had one too?
Or can the desugared code ignore some requirements, _because_ it's actually the desugaring of the feature introducing that requirement? (If we desugar `enum` classes into normal classes with normal `const` constructor invocations, those classes _do_ extend `Enum` and are instantiable, which is otherwise an error.)

Lifting things to a semantic level before (or while) doing "rewrites" allows us to ignore syntactic restrictions.

lrhn: But also, yes, I believe the section is unnecessary if we are careful about how we specify things.
We cannot just say "if the superclass has the modifier ..." when the superclass might not have any modifiers, because it's an anonymous mixin application class.

I'm willing to try!

accepted/future-releases/class-modifiers/feature-specification.md
lrhn: Yep. But maybe necessary.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -513,23 +515,16 @@ Many combinations don't make sense:
...
-    declaration because the mixin can be applied within the same library.
-    A `sealed mixin class`  does not provide any significant extra
-    functionality over a `sealed mixin`, you can replace `extends MixinClass`
-    with `with Mixin`, so a `sealed mixin class` is not allowed.*
-*   `interface` and `final` classes would prevent a mixin class from being used
-    as a superclass or mixin outside of its library. *Like for `sealed`, an
-    `interface mixin class` and `final mixin class` are not allowed, and
-    `interface mixin` and `final mixin` declaration are recommended instead.*
+*   `sealed` types cannot be mixed in, extended or implemented,
+    so it's redundant to combine with `final`, `base`, or `interface`.
```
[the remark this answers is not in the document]
eernstg: Indeed, that's better. Edited the suggestion to do that.

(I don't think it's redundant, of course, but I don't expect language team support for making all properties explicit. Actually, it would probably be too noisy, and we can always take another dog fight about whether or not it should be shown on hovering on `sealed` and on each mixin name in an anonymous mixin application ;-)

eernstg (commented):
Couple of comments.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -909,18 +951,25 @@ non-breaking.
...
-
-    This is a special case behavior only available to platform libraries.
+    Going through a pre-feature library does not remove transitive restrictions
+    for code in post-feature libraries. Any post-feature library declaration which has a
+    platform library class marked `base` or `final` as a superinterface must be marked
+    `base`, `final` or `sealed`, and cannot be implemented locally,
+    even if the superinterface chain goes through a pre-feature library declaration,
+    and even if that declaration ignores the `base` modifier.
+
+    This is special case behavior only available to platform libraries.
```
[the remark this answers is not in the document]
lrhn: It does refer to the "SDK-exception" described in the previous paragraph, so I think its where it should be. Will consider whether it's clear enough about that.

accepted/future-releases/class-modifiers/feature-specification.md
lrhn: I think I have all the discussion of *why* platform libraries are treated specially here, and the actual rules integrating that behavior directly in their specification, not as an appended exception.

Hope it works.

accepted/future-releases/class-modifiers/feature-specification.md
lrhn: And done. Things are now defined entirely in terms of super-declarations, not classes or interfaces (semantic entities, different from the syntactic declarations introducing them). Because anonymous mixin applications are just part of another declaration, they don't need further rules.

accepted/future-releases/class-modifiers/feature-specification.md
lrhn: Did so in the section describing the exceptions.

accepted/future-releases/class-modifiers/feature-specification.md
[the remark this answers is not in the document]
lrhn: Done. (Including the "already".)

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -566,13 +558,13 @@ classDeclaration  ::= (classModifiers | mixinClassModifiers) 'class' typeIdentif
...
 
-mixinClassModifiers ::= 'abstract'? 'base'? 'mixin'
+mixinClassModifiers ::= 'abstract'? mixinModifier? 'mixin'
 
 mixinDeclaration  ::= mixinModifier? 'mixin' typeIdentifier typeParameters?
                       ('on' typeNotVoidList)? interfaces?
                       '{' (metadata classMemberDeclaration)* '}'
 
-mixinModifier     ::= 'sealed' | 'base' | 'interface' | 'final'
+mixinModifier     ::= 'base'
```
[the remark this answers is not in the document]
lrhn: True. :+1:

accepted/future-releases/class-modifiers/feature-specification.md
lrhn: But you can ignore modifiers in your own library. You cannot ignore transitive `base` or `final` modifiers from another library.

accepted/future-releases/class-modifiers/feature-specification.md
lrhn: Narrator voice: They weren't.

(Huzzah!)

accepted/future-releases/class-modifiers/feature-specification.md
lrhn: The entire
```ebnf
<superclass> ::= \EXTENDS{} <typeNotVoid> <mixins>?
    \alt <mixins>
```
clause, so named in the specification.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -761,25 +709,26 @@ versioning and backwards compatibility.
...
 *   has no body (only `;`), and
 
 *   is not `external`. *An `external` constructor is considered to have an
     externally provided initializer list and/or body.*
 
-A trivial constructor may be named or unnamed, and `const` or non-`const`.
-A *non-trivial generative constructor* is a generative constructor which is not a
-trivial generative constructor.
+A trivial generative constructor may be named or unnamed,
+and maybe be `const` or non-`const`.
```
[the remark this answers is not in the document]
lrhn: Fixed.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -836,38 +785,82 @@ the latter, even if the class is being used as a mixin in a post-feature
...
+
+The class introduced by an `enum` declaration is considered `final` for any
+purpose where a class modifier is required.
+
+The behavior of `enum` declarations is unchanged. Since an `enum` class cannot
+have any subclasses, the implicit `final` modifier would not prevent any
+otherwise allowed operation.
+The implicit `final` also does not imply fewer restrictions than `enum`
+declarations would otherwise have. The modifier is applied only to automatically
+satisfy any requirements introduced by super-interfaces.
```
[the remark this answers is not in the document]
lrhn: Removed.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -551,9 +549,6 @@ The remaining valid combinations and their capabilities are:
 |`abstract base mixin class`|No     |**Yes**|No     |**Yes**|No     |
```
oprypin, line 549: This is not a new issue for this pull request, but I wanted to comment anyway:

Is it correct that `sealed class` has "Construct?" as "No" in this table?

PR #2893, opened 2023-03-08
Thoughts about errors, error recovery, and fixups for class modifiers

stereotype441:
I sent this in an email to some folks for comment, and the response was positive enough that it seemed worth including in the spec.

### DIFF
--- a/accepted/future-releases/class-modifiers/feature-specification.md
+++ b/accepted/future-releases/class-modifiers/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom, Lasse Nielsen
 
 Status: Accepted
 
-Version 1.5
+Version 1.6
 
 Experiment flag: class-modifiers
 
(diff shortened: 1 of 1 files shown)


munificent (approved):

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -933,8 +931,139 @@ as a mixin. If the class defines a generative constructor or extends anything
...
+
+- `sealed` means "this type has a known set of direct subtypes, so switching on
+  it will require the switch to be exhaustive".
+
+- `abstract` means "this type can't be constructed directly", but you already
+  knew that.  It's only included in the list to help clarify that `abstract` is
+  one of the seven keywords users should try combining together at the top of
+  your declaration.
+
+- `interface` means "this type can't be extended from outside this library".
```
leafpetersen, line 1035: I think it is clearer to present this in terms of what the user *can* do.  That is `interface` means "this type can only be implemented (not extended) outside of its defining library".  Likewise for the `base` below.  It seems small, but I think it's much clearer - if you tell people what they *can't do*, then they need to subtract from the set of all possible things in order to find out what they *can do*, which is usually what they want to know.

PR #2916, opened 2023-03-15
Remove the support for coercions during pattern matching

eernstg:
**Just in case we decide to do this**, here's a proposal for a spec update that removes the support for coercions that occur during pattern matching. It explicitly mentions that some coercions continue to be supported (and they are also working today, as far as I know&mdash;certainly, this example works today):

```dart
X id<X>(X x) => x;

void main() {
  dynamic d = 'Hello';
  var (int Function(int) f, String s) = (id, d); // OK, yields `(id<int>, d as String)`.

  (X Function<X>(X), dynamic) record = (id, d);
  var (int Function(int) f, String s) = record; // Compile-time error, not assignable.

  (dynamic, dynamic) record2 = (id, d);
  var (X Function<X>(X) f2, String s2) = record2; // OK, also at run time.
}
```

### DIFF
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: Accepted
 
-Version 2.24 (see [CHANGELOG](#CHANGELOG) at end)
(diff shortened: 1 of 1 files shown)


eernstg:
Same as #2926.

munificent:
Oops, sorry I didn't see this!

eernstg:
NP!

PR #2917, opened 2023-03-15
Update augmentation library spec for class modifiers

jakemac53:
I removed the line about `abstract` and instead chose to say that all class modifiers must be present on both the augmentation type and the original type. This is I think consistent with the requirement to duplicate the type parameters, and it ensures people won't be confused about the modifiers on a class when looking at only the augmentation.

Open for debate certainly though! We could go the opposite direction and drop the `class/mixin/enum` keywords from augmentation libraries, as well as duplicate type parameters, thus eliminating all redundant information. So you could augment any class/mixin/enum by doing just:

```
augment <typename> {
  <members>
}
```

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -313,14 +313,15 @@ It is a compile-time error if:
 *   The augmenting type and corresponding type are not the same kind: class,
     mixin, enum, or extension. You can't augment a class with a mixin, etc.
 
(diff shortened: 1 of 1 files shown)


johnniwinther (approved):

munificent (approved):

scheglov (approved):

kevmoo:
So glad we're keeping this fresh. 🤞

PR #2923, opened 2023-03-16
Change `<extensionDeclaration>` to use `<typeIdentifier>`

eernstg:
The language specification currently uses an `<identifier>` as the name of an `<extensionDeclaration>`. Implementations require this name to be a `<typeIdentifier>`, just like class and mixin declarations.

I see no value in forcing implementations to accept a built-in identifier as the name of an extension. This PR changes the language specification accordingly, which ensures that it is in sync with the implementations. This is arguably more consistent, and yields more readable code.

Added another small cleanup, cf. https://dart-review.googlesource.com/c/sdk/+/288981: Moves `?` from `typeNotVoidNotFunction` rule to usages, such that we do not (syntactically) allow constructs like `extends C?`, which are errors anyway.

Cf. https://github.com/dart-lang/sdk/issues/51751.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -6064,7 +6064,7 @@ and whether the invocation satisfies several other requirements.
 
 \begin{grammar}
 <extensionDeclaration> ::=
-  \gnewline{} \EXTENSION{} <identifier>? <typeParameters>? \ON{} <type>
+  \gnewline{} \EXTENSION{} <typeIdentifier>? <typeParameters>? \ON{} <type>
   \gnewline{} `\{' (<metadata> <classMemberDeclaration>)* `\}'
 \end{grammar}
 
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -20688,11 +20688,6 @@ that is,
 it denotes an \synt{IDENTIFIER} which is not a \synt{BUILT\_IN\_IDENTIFIER}
 (\ref{identifierReference}).%
 }
-%% TODO(eernst): The following non-terminals are currently unused (they will
-%% be used when we transfer more grammar rules from Dart.g): <typeNotVoid>
-%% and <typeNotVoidNotFunctionList>. They are used in the syntax for
-%% \EXTENDS, \WITH, \IMPLEMENTS{} syntax and for mixin applications
-%% in Dart.g, and it seems likely that we will use them here as well.
 
 \commentary{%
 Non-terminals with names of the form \synt{\ldots{}NotFunction}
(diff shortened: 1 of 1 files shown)


modulovalue:
https://github.com/dart-lang/language/blob/abca38c39e9bea0af2ff47fce55d6686207178e7/specification/dartLangSpec.tex#L5845-L5846

I think this part of the spec would become redundant and could be removed.

eernstg:
Thanks, done!

eernstg:
@lrhn, friendly ping, this one is tiny. ;-)

specification/dartLangSpec.tex
```diff
@@ -20710,13 +20706,13 @@ \subsection{Static Types}
...
+  \alt <typeNotVoidNotFunction> `?'?
 
 <typeNotFunction> ::= \VOID{}
-  \alt <typeNotVoidNotFunction>
+  \alt <typeNotVoidNotFunction> `?'?
 
-<typeNotVoidNotFunction> ::= <typeName> <typeArguments>? `?'?
-  \alt \FUNCTION{} `?'?
+<typeNotVoidNotFunction> ::= <typeName> <typeArguments>?
+  \alt \FUNCTION{}
```
lrhn, line 20715: Is `Function` not already a `<typeName>`?

It's not a built-in identifier.

specification/dartLangSpec.tex
```diff
@@ -20710,13 +20706,13 @@ \subsection{Static Types}
...
-  \alt <typeNotVoidNotFunction>
+  \alt <typeNotVoidNotFunction> `?'?
 
 <typeNotFunction> ::= \VOID{}
-  \alt <typeNotVoidNotFunction>
+  \alt <typeNotVoidNotFunction> `?'?
 
-<typeNotVoidNotFunction> ::= <typeName> <typeArguments>? `?'?
-  \alt \FUNCTION{} `?'?
+<typeNotVoidNotFunction> ::= <typeName> <typeArguments>?
```
lrhn, line 20714: What do we use this production for?
It's out the types allowed in extends, implements, with and mixin on types?

We could also just allow any type there, and make it an error if it's not one of the bespoke declaration kinds that you can extend. We allow you to try to extend those grammatically anyway, because of type aliases, so it might give better front-end errors if we don't make it a parse error to write `extends void`. 
It also feels arbitrary that we don't prevent `Never`.

And if we ever decide to allow extension types to implement function types, we won't need to change the grammar.

lrhn (approved):

specification/dartLangSpec.tex
eernstg: `Function` is actually a built-in identifier (since https://dart-review.googlesource.com/c/sdk/+/41360).

specification/dartLangSpec.tex
eernstg: It is used for all superinterfaces (`implements`, `extends`, `on`, and on the left hand side of `with` in mixin applications).

We could indeed allow any type, and make it a non-syntax compile-time error to have the kinds of types that we're currently ruling out in the grammar.

Most proposals have been in the opposite direction: "Adjust the grammar such that this-or-that wrong thing becomes a syntax error". Arguably, the grammar is easier to read if it describes the language properties more directly, and harder to read if it has more rules and more detail, so there is no obvious direction which is always the correct/useful one.

It is worth noting that various constructs (including type aliases and extension types) can be used to introduce types that are not allowed as superinterfaces, and yet they're syntactically identical to other types that are allowed.

However, it isn't obvious to me that it's helpful to indicate in the grammar that a function type can be a superinterface, and yet every single function type used as a superinterface is a compile-time error.

So I'll keep the grammar rule for now. Please create an issue if you wish to dive deeper into this topic.

eernstg:
Thanks! I did not eliminate `<typeNotVoidNotFunction>` in this PR after all. I think it's better to do that in a separate PR (where we might make several changes of that nature).

PR #2926, opened 2023-03-16
[patterns] Specify that coercions don't happen during destructuring.

### LINKED ISSUE #2914: Don't perform value-transforming coercions during destructuring
There are four implicit-coercion-ish things in Dart:

1. Int-to-double.
2. Implicit casts from dynamic.
3. Implicit call tear-offs.
4. Implicit generic function instantiation.

In the current proposal, any of these coercions may be inserted in code using patterns. In particular:

1. In pattern declarations and assignments, the context type of the pattern is used for downwards inference on the matched value. For example:

    ```dart
    T id<T>(T t) => t;

    var (int Function(int) f,) = (id,);
    ```

    Here, we calculate context type `(int Function(int),)` for the pattern and push that over to the initializer expression. That gets pushed into the field and `id` is coerced to `int Function(int)`, as if the user had written:

    ```dart
    T id<T>(T t) => t;

    var (int Function(int) f,) = (id<int>,);
    //                              ^^^^^
    ```

2. When a value is destructured by some pattern and passed to a subpattern. For example:

    ```dart
    T id<T>(T t) => t;

    var r = (id,);
    var (int Function(int) f,) = r;
    ```

    Here, the downwards inference context has no effect since `r` is just a variable lookup. Instead, the coercion of `id` to `int Function(int)` happens *after the record pattern extracts the field*. It's as if the user had written:

    ```dart
    T id<T>(T t) => t;

    var r = (id,);
    var (f_temp,) = r;
    int Function(int) f = f_temp;
    ```

Coercions from 1 are already implemented because the implementations already understand how to insert coercions based on context type. We get that "for free". In fact, the implementations don't even know or care if a context type applied to an expression came from a declared variable's type annotation, a pattern, or something else.

For scenario 2, int-to-double doesn't apply. That "coercion" only happens to integer *literals* based on context type.

Casts from `dynamic` do apply but don't require the compiler to insert an actual coercion. Instead, they are handled the same way other pattern type tests work where the pattern is refuted if the value isn't of that type. In an irrefutable context, that refutation becomes a runtime exception.

The other two coercions, call tear-offs and implicit generic function instantiation are not implemented for destructuring yet. What I've heard from @eernst and @johnniwinther is that doing so will be a significant amount of work.

It's not clear that it's useful to support either of these coercions during destructuring. We are considering removing support for call tear-offs from the language entirely because it causes a lot of implementation complexity for very little user value. If we do that, not supporting them in destructuring now will give us less to migrate if we remove support for them later.

I don't know how often implicit generic function instantiation is used, but I feel fairly confident we can live without it in destructuring. If it ends up being something users really want, we can always add it later.

Therefore, I suggest that we change the proposal to *not* insert value-transforming coercions during destructuring. This means that we do support:

1. All coercions when applied to a matched value expression based on the context type of the pattern, as already works today. So this works:

    ```dart
    class Callable {
      int call(int x) => x;
    }

    T id<T>(T t) => t;

    main() {
      var (double a,) = (123,);
      var (int b,) = (123 as dynamic,);
      var (int Function(int) c,) = (Callable(),);
      var (int Function(int) d,) = (id,);
    }
    ```

2. Implicit casts from `dynamic` during destructuring. This is allowed and works:

    ```dart
    var r = (123 as dynamic,);
    var (int f,) = r;
    ```

    And this is allowed but throws at runtime:

    ```dart
    var r = ('not int' as dynamic,);
    var (int f,) = r;
    ```

But we do not support:

1. Inserting call tear-offs when destructuring a value. So this is a compile-time error:

    ```dart
    class Callable {
      int call(int x) => x;
    }

    main() {
      var r = (Callable(),);
      var (int Function(int) c,) = r;
      // Error because Callaback is not a subtype of `int Function(int)`.
    }
    ```

    And this prints "No match":

    ```dart
    var r = (Callable(),);
    switch (r) {
      case (int Function(int) c,): print('Function record');
      case _: print('No match');
    }
    ```

2. Inserting implicit generic function instantiation when destructuring a value. So this is a compile-time error:

    ```dart
    T id<T>(T t) => t;

    main() {
      var r = (id,);
      var (int Function(int) d,) = r;
    }
    ```

    And this prints "No match":

    ```dart
    var r = (id,);
    switch (r) {
      case (int Function(int) c,): print('Function record');
      case _: print('No match');
    }
    ```

How does sound, @dart-lang/language-team, @johnniwinther, and @scheglov?

munificent:
Fix #2914 based on our discussion at this week's language meeting.

cc @dart-lang/language-team

### DIFF
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: Accepted
 
-Version 2.24 (see [CHANGELOG](#CHANGELOG) at end)
+Version 2.25 (see [CHANGELOG](#CHANGELOG) at end)
 
 Note: This proposal is broken into a couple of separate documents. See also
 [records][] and [exhaustiveness][].
@@ -1687,8 +1687,9 @@ To orchestrate this, type inference on patterns proceeds in three phases:
     Using the pattern's type schema as a context type (if not in a matching
     context), infer missing types on the value expression. This is the existing
     type inference rules on expressions. It yields a complete static type for
-    the matched value. This process may also insert implicit coercions and casts
-    from `dynamic` in the matched value expression.
+    the matched value. As usual, when a context type is applied to an
+    expression, the process may also insert implicit coercions and casts from
+    `dynamic` in the matched value expression.
 
     *For example:*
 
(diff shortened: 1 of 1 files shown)


munificent:
I have some language tests to cover this here: https://dart-review.googlesource.com/c/sdk/+/289404

cc @johnniwinther @scheglov

eernstg:
Ah, not that different from https://github.com/dart-lang/language/pull/2916. ;-)

accepted/future-releases/0546-patterns/feature-specification.md
```diff
@@ -1708,28 +1709,36 @@ To orchestrate this, type inference on patterns proceeds in three phases:
...
-    However, the matched value type `(T Function<T>(T), dynamic)` is allowed by
-    the record pattern's required type `(Object?, Object?)`, the field matched
-    value type `T Function<T>(T)` is allowed by the field required type `int
-    Function(int)`, and the field matched value type `dynamic` is allowed by the
-    field required type `String)`. So the declaration is valid. Coercions are
-    inserted after destructuring each record field before passing them to the
-    field subpatterns. At runtime, when the record is destructured during
-    matching, the coercions are applied. This is specified below.*
+    However, the matched value type `(dynamic, dynamic)` is allowed by the
+    record pattern's required type `(Object?, Object?)`, so the declaration is
```
eernstg, line 1725: We might also mention the type checking of each record field, in order to establish all the parts needed for the conclusion 'so the declaration is valid':
```suggestion
    record pattern's required type `(Object?, Object?)`, and the matched
    value type `dynamic` for each field is allowed by the required types
    of the fields, i.e., `int` and `String`, so the declaration is
```

eernstg (approved):
LGTM!

PR #2944, opened 2023-03-23
Proposal: specific bit-size num types

xzzulz:
Add a proposal for adding additional num types to Dart. These num types will have specific bit-size.  
Making the language useful in many use cases. And providing better interop with other languages.


lrhn:
I think the scope of this document fits well into an issue instead.
The existing issue https://github.com/dart-lang/language/issues/2100 might cover it.

PR #2947, opened 2023-03-24
Change the disambiguation rule for function literals inside guards.

stereotype441:
We now unconditionally prohibit function literals inside guards, regardless of whether their bodies take the form `=> expression` or a block.  This matches what was implemented in the parser.

See discussion in https://github.com/dart-lang/sdk/issues/51482.

### DIFF
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: Accepted
 
-Version 2.25 (see [CHANGELOG](#CHANGELOG) at end)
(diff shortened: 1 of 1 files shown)


munificent (approved):
Thanks!

Would you mind also adding a little line about this to the changelog in the doc and bumping the version number? It probably doesn't matter in this case since you're the implementer too, but it's nice to keep track of the changes.

stereotype441:
> Thanks!
> 
> Would you mind also adding a little line about this to the changelog in the doc and bumping the version number? It probably doesn't matter in this case since you're the implementer too, but it's nice to keep track of the changes.

Of course!  Thanks for the reminder.

munificent (commented):
LGTM!

PR #2948, opened 2023-03-25
Rewrite the exhaustiveness specification.

munificent:
This is based on Johnni's excellent implementation. I tried to follow the semantics of the implementation as closely as I could, and mostly stuck to the terminology too. The main change is that the implementation uses "type" to also sort of do some restriction stuff, and I tried to keep that separate here such that a "type" on a space always refers to a static type.

I'm sure there are some mistakes here (but I of course don't know where they are). I imagine it will take some iteration between the implementation, the spec, and the tests before they settle on some kind of fixpoint. But this should hopefully be enough to start writing some co19 tests.

cc @sgrekhov

### DIFF
--- a/accepted/future-releases/0546-patterns/exhaustiveness.md
+++ b/accepted/future-releases/0546-patterns/exhaustiveness.md
@@ -190,9 +190,10 @@ we would break that principle.
 ### Sealing supertypes
 
 Instead, we [extend the language to let a user explicitly "seal"][sealed] a
-supertype with a specified closed set of subtypes. No other code is allowed to
-define a new subtype of the sealed supertype (using `extends`, `implements`, or
-`with`). The supertype is also implicitly made abstract.
+supertype with a specified closed set of subtypes. No code outside of the
+library where the sealed type is declared is allowed to define a new subtype of
+the sealed supertype (using `extends`, `implements`, `with`, or `on`). The
+supertype is also implicitly made abstract.
 
 [sealed]: https://github.com/dart-lang/language/blob/master/accepted/future-releases/sealed-types/feature-specification.md
 
@@ -203,20 +204,31 @@ enum Suit { club, diamond, heart, spade }
 
 sealed class Card {
   final Suit suit;
+
+  Card(this.suit);
 }
 
 class Pip extends Card {
   final int pips;
+
+  Pip(this.pips, super.suit);
 }
 
 sealed class Face extends Card {}
 
 class Jack extends Face {
   final bool oneEyed;
+
+  Jack(super.suit, {this.oneEyed = false});
 }
 
-class Queen extends Face {}
-class King extends Face {}
+class Queen extends Face {
+  Queen(super.suit);
+}
+
+class King extends Face {
+  King(super.suit);
+}
 ```
 
 The above class declarations create the following class hierarchy:
@@ -240,8 +252,12 @@ sealing does *not* mean:
     fact, we use this to turn `Face` into a sealed supertype of its own set of
     subtypes.
 
-*   The subtypes do not have to be disjoint. We could define a `Royalty` class
-    that implements `Jack`, `Queen`, and `King`.
+*   The subtypes do not have to be disjoint. Another library that wanted to
+    depose the royalty could define a `Democracy` class that implements `Jack`,
+    `Queen`, and `King` and takes over their functionality. This doesn't cause
+    any problems for exhaustiveness even though it means that a single instance
+    of `Democracy` would match more than one of the sealed subtypes of `Card`
+    and `Face`.
 
 *   The subtypes can have other supertypes in addition to the sealed one. We
     could have a `Monarch` interface that `Queen` and `King` implement.
@@ -253,26 +269,27 @@ subtypes, then we have exhaustively covered all instances of the supertype.
 
 ## Spaces
 
-The two questions exhaustiveness answers are defined based on some notion of a
-"set of values". A series of cases is exhaustive if the set of all possible
+Determining whether a switch is exhaustive requires reasoning about some notion
+of a "set of values". A series of cases is exhaustive if the set of all possible
 values entering the switch is covered by the sets of values matched by the case
 patterns.
 
 Thus the algorithm needs a way to model a (potentially infinite) set of values.
 In a statically typed language, the first obvious answer is a static type. A
 type does represent a set of values. But a plain static type isn't precise
-enough things like the set of values matched by the pattern `DateTime(day: 1)`
-which matches not just values of some type, but also ones whose fields have
-certain values.
+enough for things like the set of values matched by the pattern `DateTime(day:
+1)` which matches not just all values of some type, but only ones whose fields
+have certain values.
 
-A pattern is obviously the natural way to describe a set of values of some type
-filtered by some arbitrary set of predicates on their properties. However, the
-patterns proposal defines a rich set of patterns to make the feature user
-friendly and expressive. It would require unnecessary complexity in the
-exhaustiveness algorithm to handle every single kind of pattern.
+A pattern is the natural way to describe a set of values of some type filtered
+by some arbitrary set of predicates on their propertiesl, and we will use
+somethig similar here. However, the patterns proposal defines a rich set of
+patterns to make the feature user friendly and expressive. It would require
+unnecessary complexity in the exhaustiveness algorithm to handle every single
+kind of pattern.
 
 Instead, following Liu's paper, we use a data structure called a *space*. Spaces
-are essentially a desugared, simpler representation of a pattern.
+are essentially a simpler unified abstration over patterns and static types.
 
 A single space has a *type*, a *restriction*, and zero or more *properties*. All
 of these are used to determine which values match the space and which don't. The
@@ -342,6 +359,18 @@ named getter accessors. There are a few kinds of keys:
     `i` from the *end* of the list by calling `list[length - i]` on the matched
     list.
 
+Every key has a corresponding static type that can be looked up given the
+type of the space it is being accessed on:
+
+*   For a getter or field key, the type is the type of the corresponding member
+    or record field.
+
+*   For an element or tail key, the type is the return type of the corresponding
+    `[]` operator declared on the space's type.
+
+*   For a rest key, the type is the same as the type of the space. That will
+    always be `List<T>` for some `T`.
+
 ## Lifting types and patterns to space unions
 
 The exhaustiveness algorithm works on spaces and space unions, but it is invoked
@@ -370,7 +399,7 @@ The lifted space union for a pattern with matched value type `M` is:
     matters it that execution can't flow out of a switch without matching at
     least one case. But if a pattern _throws_ a runtime exception, then
     execution also doesn't flow out of it. So we treat throwing as essentially
-    another kind of matching.*
+    another kind of matching for exhaustiveness.*
 
     *In practice, this normally isn't very helpful. Consider:*
 
@@ -405,7 +434,8 @@ The lifted space union for a pattern with matched value type `M` is:
 
     *After the first case, we know that if `a` is a `B` or `D` then we will have
     thrown an exception. But if `a` is a `C`, it may not have matched if the
-    field isn't `0`. So the space for the first case is `B|C(field: 0)|D'.*
+    field isn't `0`. So the space for the first case is a union of `B|C(field:
+    0)|D'.*
 
     *Then the second case covers the rest of `C` and this is exhaustive.*
 
@@ -436,8 +466,6 @@ The lifted space union for a pattern with matched value type `M` is:
     example:*
 
     ```dart
-    typedef NullableObject = Object?;
-
     test(Object? obj) => switch (obj) {
       case _?:
       case null:
@@ -449,8 +477,6 @@ The lifted space union for a pattern with matched value type `M` is:
     nullable, then step 2 can be involved:*
 
     ```dart
-    typedef NullableObject = Object?;
-
     test(Object? obj) => switch (obj) {
       case Object? _?:
     };
@@ -465,8 +491,18 @@ The lifted space union for a pattern with matched value type `M` is:
     *As with cast patterns, a null-assert pattern "matches" `null` by throwing
     an exception, which is sufficient for exhaustiveness.*
 
-*   **Constant pattern:** A space whose type is the type of the constant and
-    with a constant restriction for the given constant value.
+*   **Constant pattern:**
+
+    1.  If the constant has primitive equality, then a space whose type is the
+        type of the constant and with a constant restriction for the given
+        constant value.
+
+    2.  Else the empty space union.
+
+    *If the constant has a user-defined `==` method, then we can't rely on its
+    behavior for exhaustiveness checking. Fortunately, the constants that most
+    often come into play for exhaustiveness are enum values, booleans, and
+    `null`, and those all have primitive equality.*
 
 *   **Variable pattern** or **identifier pattern:** The lifted space union of
     the static type of the corresponding variable.
@@ -556,7 +592,7 @@ union that contains only values contained by both of its operands.
 Intersection is approximate and pessimistic: There may be values that are
 matched by both operands that are not matched by the resulting intersection.
 Since intersection is only invoked when lifting patterns to spaces (and not
-value types), which is sound, though it may lead to the compiler not recognizing
+value types), it is sound, though it may lead to the compiler not recognizing
 that a switch is actually exhaustive when it is or not recognizing a case as
 unreachable when it can't be reached.
 
@@ -568,7 +604,8 @@ Space intersection is defined as:
     each operand of the union with the other space.
 
     *In other words, distribute the intersection into the branches. So `(A|B) ^
-    C` is `(A^C)|(B^C)`.
+    C` (where `|` is "union" and `^` is "intersection") results in `(A^C)|(B^C)`
+    and then calculate the resulting intersections.*
 
 3.  Else (both sides are single unions):
 
@@ -827,7 +864,7 @@ The resulting spaces are:
             ```
 
         2.  Else (not a trivial substitution) let `S` be an overapproximation
-            of `C`.
+            (see below) of `C`.
 
             **TODO:** Is the overapproximation part here correct?
 
@@ -864,7 +901,7 @@ The resulting spaces are:
     1.  For each element in the enum:
 
         1.  If the element's type is a subtype of the overapproximation of the
-            enum type:
+            enum declaration:
 
             1.  A space with the type of the element and a constant restriction
                 with the element value.
@@ -975,19 +1012,20 @@ null
 
 ### Overapproximation
 
-The *overapproximation* of a type `T` is `T` with all type variables replaced
-with their *defaults*:
+The *overapproximation* of a declaration `D` is a type `T` with all type
+variables replaced with their *defaults*:
 
-1.  If `T` is in a contravariant position, then the default is `Never`.
+1.  If the type variable is in a contravariant position, then the default is
+    `Never`.
 
 2.  Otherwise, the default is the bound if there is one.
 
 3.  Otherwise, the default is `Object?`.
 
-The overapproximation of a type reliably contains all values that any possible
-instantiation of the type could contain. This lets us calculate exhaustiveness
-soundly even in the context of parameter types whose concrete instantiations
-aren't known.
+The overapproximation of a declaration reliably contains all values that any
+possible instantiation of its type parameters could contain. This lets us
+calculate exhaustiveness soundly even in the context of parameter types whose
+concrete instantiations aren't known.
 
 ### Filtering by a space
 
@@ -1006,13 +1044,13 @@ exhaustiveness, so we remove it.
 
     1.  Dequeue the first case union in `worklist` to `caseUnion`.
 
-    1.  For each `case` space in `caseUnion`:
+    2.  For each `case` space in `caseUnion`:
 
         1.  If `valueSpace` is a subset of `case`, then:
 
             1.  Add `case` to `caseFirstSpaces`.
 
-            1.  Add `worklist` to `remaining`.
+            2.  Add `worklist` to `remaining`.
 
             *If every value in `valueSpace` is also in `case`, then this case
             may still help with exhaustiveness, so keep it.*
@@ -1073,10 +1111,12 @@ To unpack a `space` given a set of `keys`:
     1.  If `space` has a property with this key, then append that property's
         subspace union to `result`.
 
-    2.  Else, append a `space` to `result` whose type is `Object?`. *If a
-        pattern doesn't have a subpattern for some field or getter, then it
-        matches all values that the field or getter could return, which is
-        equivalent to it matching it with type `Object?`.*
+    2.  Else, append a `space` to `result` whose type is the static type of the
+        key declared on the type of `space`.
+
+        *If a pattern doesn't have a subpattern for some field or getter, then
+        it matches all values that the field or getter could return, which is
+        equivalent to it matching it with its type.*
 
 3.  Return `result`.
 
--- a/accepted/future-releases/0546-patterns/exhaustiveness.md
+++ b/accepted/future-releases/0546-patterns/exhaustiveness.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: In progress
 
-Version 1.1 (see [CHANGELOG](#CHANGELOG) at end)
+Version 2.0 (see [CHANGELOG](#CHANGELOG) at end)
 
 ## Summary
 
@@ -52,7 +52,9 @@ Exhaustiveness checking is about answering two questions:
     ```
 
     Here, the third case can never be matched because any value it matches would
-    also already be matched by one of the preceding two cases.
+    also already be matched by one of the preceding two cases. This is not a
+    concern for soundness or correctness, but is dead code that an
+    implementation may want to warn on.
 
 Dart already supports exhaustiveness and reachability warnings in switch
 statements on `bool` and enum types. This document extends that to handle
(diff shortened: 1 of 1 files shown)


accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
...
-A a = ...
-switch (a) {
-  case B b: ...
-  case F f: ...
-}
-```
+    2.  Else, append a `space` to `result` whose type is `Object?`. *If a
+        pattern doesn't have a subpattern for some field or getter, then it
+        matches all values that the field or getter could return, which is
+        equivalent to it matching it with type `Object?`.*
```
johnniwinther, line 1079: While this is true for the case spaces it isn't precise enough for the value space since it might lead to a space larger that what the type of the property can contain. For instance, if the value space is `Object()` and we have a case space `Object(hashCode: int())` then unpacking the key `hashCode` for `Object()` to `Object?()` would mean that we would conclude that `int()` does not exhaust the `hashCode` property. Instead we need to use the type corresponding to the property of the type of the space when it exists. Currently the algorithm uses `Object?()` as a fallback when the property doesn't exist, which (I think) can only happen for erroneous code. (Given that we have checked that the type of the value space is a subtype of the case space, so if the property exists on the case space type it must exist on the value space type.)

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
...
-y:    club │(x: ♣︎, y: ♣︎)││(x: ♦︎, y: ♣︎)  (x: ♥︎, y: ♣︎)  (x: ♠︎, y: ♣︎)│
-           │            ││                                        │
-   diamond │(x: ♣︎, y: ♦︎)││(x: ♦︎, y: ♦︎)  (x: ♥︎, y: ♦︎)  (x: ♠︎, y: ♦︎)│
-           │            ││                                        │
-     heart │(x: ♣︎, y: ♥︎)││(x: ♦︎, y: ♥︎)  (x: ♥︎, y: ♥︎)  (x: ♠︎, y: ♥︎)│
-           └────────────┘│                                        │
-     spade    <hole>     │(x: ♦︎, y: ♠︎)  (x: ♥︎, y: ♠︎)  (x: ♠︎, y: ♠︎)│
-                         └────────────────────────────────────────┘
-```
+### Are any switch cases unreachable?
```
johnniwinther, line 675: To put this section in context we should probably put at the end of the specification more as a side note and mention that this is an unrequired extension of the exhaustiveness checking that can be used by tools to provide hints.

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
...
+3.  For each `worklist` in `cases`:
 
-## Next steps
+    1.  Dequeue the first case union in `worklist` to `caseUnion`.
 
-I haven't proven that the algorithm is correct. I also haven't proven any bounds
-on its performance. A fairly large test covering every combination of four
-fields seems to be snappy (and certainly was not with earlier versions of this
-algorithm).
+    1.  For each `case` space in `caseUnion`:
```
johnniwinther, line 1009: The item number is off here and in the sublist below (not that it matters to the markdown renderer).

johnniwinther (approved):

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
...
 
-```
-FutureOr<int> - int     expands to:  Future<int>|int - int = Future<int>
-FutureOr<int> - Future  expands to:  Future<int>|int - Future = int
-```
+*   **The declaration `D` of `T` is a `sealed` type:**
 
-### Subtraction
+    1.  For each declaration `C` in the library of `D` that has `D` in an
+        `implements`, `extends`, or `with` clause:
```
lrhn, line 816: Here I want to restrict it to:

... declaration `C` in the library of `D` that has `D` as a direct superdeclaration through an `implements`, `extends` or `with` clause _(I agree with excluding `on` types too)_
and which does not have any other superdeclaration *P* which has *D* as a proper superdeclaration.

This avoids weird behavior for unnecessary implements clauses:
```dart
sealed class S {}
class C extend S {}
class D extends S {}
class E extends D implements S {}
```
Here class `E` should *not* be included, because it's a proper subtype of a proper subtype of `S` (if `D` was not `sealed`, it would be included transitiviely in exhausting `S`, but that's not important here).

It's not necessary to include `E` in the set of spaces for `S` here (exhaustiveness of `S` requires exhausting `D` anyway), so it just complicates everything.

(That is: The *longest* path from the subclass to the sealed class in the class declaration hierarchy has length 1. Internal anonymous mixin applications don't count since we use "declaration" here.
Or as you say below, "a direct subtype" - ignoring anonymous mixin types as usual.)

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
...
+        2.  Else (not a trivial substitution) let `S` be an overapproximation
+            of `C`.
 
-    test(Animal a) {
-      switch (a) {
-        case Mammal m: ...
-      }
-    }
-    ```
+            **TODO:** Is the overapproximation part here correct?
```
lrhn, line 832: What *is* an "overapproximation"?

(Ah, found definition below. Add: `(see definition in "Overapproximation" section below)`)

The overapproximation is defined as a mapping from type to a type, but is probably really from a declaration to a type (types are instantiated, declarations are not), and `S` is a space, not a type. 
Should `S` be the space for that type?

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
...
+    `Null`.
 
-    1.  Let `Lf` be the field in `L` and `Rf` be the field in `R` (either of
-        which may be inferred if not present).
+*   **`T` is `FutureOr<F>` for some `F`**: A space of type `Future<F>` and a
+    space of type `F`.
 
-    2.  Calculate `D = Lf - Rf`.
+*   **`T` is `bool`:** Two spaces of type `bool` with constant restrictions
+    `true` and `false`.
```
lrhn, line 918: And

*    **`T` is `Null`:** One space of type `bool` with the constant restriction `null`.

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
...
 
-        ```dart
-        String s = ...
-        switch (s) {
-          case 'a string': ...
-          case 'a string': ...
-        }
-        ```
+*   **Constant pattern:** A space whose type is the type of the constant and
+    with a constant restriction for the given constant value.
```
lrhn, line 469: (Do we require the type of a constant pattern to have primitive equality? If not, should this only apply if it does?)

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
...
+null
+```
 
-2.  If `L` or `R` is a union, then `I` is empty if the intersection of every
-    arm of the union with the other operand is empty.
+### Overapproximation
 
-3.  Otherwise, we're intersecting two object spaces.
+The *overapproximation* of a type `T` is `T` with all type variables replaced
+with their *defaults*:
```
lrhn, line 979: I think this is slightly underdefined.

I think we only apply "overapproximation" to `class`, `mixin` and `enum` declarations.  So say that.
(It gets weird if we try to replace type variables of a generic function type.)
We can always extend it to apply to the type of such a declaration by going from the type to the declaration.

Also, we should probably say that we apply it to the *declaration*, not the type.
(Types do not have type variables or bounds, their declarations - and, depending on mood, their generic classes and interfaces, which are still different from the *types* - do. The types have type *arguments*, but the variables are long gone.)

So, given a generic enum, mixin or class declaration, instantiate its type with the default types for its declaration's type parameters.
But that's basically what "instantiate to bounds" (I2B) does, giving you a valid super-bounded supertype for all valid instantiations, so maybe just use that?
(I2B also accounts for bound expressions that aren't straight-up types, up-to and including F-bounded type parameters. Can't just use a bound as a type directly.)

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
...
 
-    1.  If neither object's type is a subtype of the other then `I` is empty.
-        They are unrelated types with no (known) intersection.
+1.  If `T` is in a contravariant position, then the default is `Never`.
 
-    2.  Otherwise, go through the fields. If the intersection of any
-        corresponding pair of fields is empty, then `I` is empty.
+2.  Otherwise, the default is the bound if there is one.
```
lrhn, line 983: `class C<T, S extends T, R extends List<T>> {}` has bound declarations, but not bound types.
Use I2B, it's good for you.

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
...
-    arm of the union with the other operand is empty.
+### Overapproximation
 
-3.  Otherwise, we're intersecting two object spaces.
+The *overapproximation* of a type `T` is `T` with all type variables replaced
+with their *defaults*:
 
-    1.  If neither object's type is a subtype of the other then `I` is empty.
-        They are unrelated types with no (known) intersection.
+1.  If `T` is in a contravariant position, then the default is `Never`.
```
lrhn, line 981: (Can a type parameter of an `extends`, `with` or `implements` type variable occur contravariantly - at least after alias expansion? Or is this just in preparation for variance annotations? In any case I2B should do the right thing.)

lrhn (commented):
Selectively reviewed the spec-like details, not the algorithm.

accepted/future-releases/0546-patterns/exhaustiveness.md
johnniwinther: I tend to think of this as "recursively instantiate to bounds" and it is slightly different that I2B. I2B works on raw types whereas the overapproximation works on instantiated types.

For instance the overapproximation of `Map<K, Map<V, K Function(K)>>>` in:
```
method<K extends num, V extends Foo<V>>(Map<K, Map<V, K Function(K)>>> map) => switch (map) { ... };
```
is `Map<num, Map<dynamic, num Function(Never)>>>`, so that each type variable occurring in the type has been replaced with the I2B default type modulo variance.

accepted/future-releases/0546-patterns/exhaustiveness.md
munificent: Ah, you're right. I felt like I was getting something wrong here. Fixed. (And added a section up on properties defining how to get the type of a key.)

accepted/future-releases/0546-patterns/exhaustiveness.md
munificent: I can move this if you want, but I think this is the natural place to stitch together how reachability is defined in terms of the core is-exhaustive function.

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -185,13 +189,14 @@ we would break that principle.
...
 ### Sealing supertypes
 
-Instead, we extend the language to let a user explicitly "seal" a supertype with
-a specified closed set of subtypes. No other code is allowed to define a new
-subtype of the sealed supertype (using `extends`, `implements`, or `with`). The
-supertype is also implicitly made abstract.
+Instead, we [extend the language to let a user explicitly "seal"][sealed] a
+supertype with a specified closed set of subtypes. No other code is allowed to
+define a new subtype of the sealed supertype (using `extends`, `implements`, or
+`with`). The supertype is also implicitly made abstract.
```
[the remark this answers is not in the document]
munificent: Done.

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -185,13 +189,14 @@ we would break that principle.
 
 ### Sealing supertypes
 
-Instead, we extend the language to let a user explicitly "seal" a supertype with
-a specified closed set of subtypes. No other code is allowed to define a new
-subtype of the sealed supertype (using `extends`, `implements`, or `with`). The
-supertype is also implicitly made abstract.
+Instead, we [extend the language to let a user explicitly "seal"][sealed] a
+supertype with a specified closed set of subtypes. No other code is allowed to
```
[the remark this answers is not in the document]
munificent: Done.

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -185,13 +189,14 @@ we would break that principle.
...
+`with`). The supertype is also implicitly made abstract.
 
-This document does *not* propose a specific syntax for sealing, but we'll use
-this hypothetical syntax for now to show an example:
+[sealed]: https://github.com/dart-lang/language/blob/master/accepted/future-releases/sealed-types/feature-specification.md
+
+Here is an example:
 
 ```dart
 enum Suit { club, diamond, heart, spade }
```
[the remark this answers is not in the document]
munificent: Done.

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -230,10 +235,10 @@ A parenthesized class name means "sealed". It's important to understand what
...
-    extension or implementation. In the example here, anyone can extend,
-    implement or even mixin, say, `Pip` or `Queen`. This doesn't cause any
-    problems. And in fact, we use this to turn `Face` into a sealed supertype
-    of its own set of subtypes.
+    extension or implementation. In the example here, anyone can extend or
+    implement, say, `Pip` or `Queen`. This doesn't cause any problems. And in
+    fact, we use this to turn `Face` into a sealed supertype of its own set of
+    subtypes.
 
 *   The subtypes do not have to be disjoint. We could define a `Royalty` class
```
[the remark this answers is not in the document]
munificent: Yeah, `Royalty` isn't a good example. Came up with another one.

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
 critical invariant that if we have matched against all instances of those
 subtypes, then we have exhaustively covered all instances of the supertype.
 
-## Types, patterns, and spaces
+## Spaces
 
 The two questions exhaustiveness answers are defined based on some notion of a
```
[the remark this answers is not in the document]
munificent: At the very beginning of the doc, it says:

> Exhaustiveness checking is about answering two questions...

But that's certainly far enough away to be worth clarifying here. Reworded.

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
...
+enough things like the set of values matched by the pattern `DateTime(day: 1)`
+which matches not just values of some type, but also ones whose fields have
+certain values.
 
-### Static types
+A pattern is obviously the natural way to describe a set of values of some type
+filtered by some arbitrary set of predicates on their properties. However, the
+patterns proposal defines a rich set of patterns to make the feature user
+friendly and expressive. It would require unnecessary complexity in the
+exhaustiveness algorithm to handle every single kind of pattern.
```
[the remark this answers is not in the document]
munificent: Reworded a bit.

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
...
 
-In a statically typed language, the first obvious answer is a static type. A
-type does represent a set of values. The type `bool` represents the set `true`
-and `false`. The type `DateTime` represents the set of all instances of the
-`DateTime` class or any of its subtypes.
+Instead, following Liu's paper, we use a data structure called a *space*. Spaces
+are essentially a desugared, simpler representation of a pattern.
```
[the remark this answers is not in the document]
munificent: Reworded a bit.

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
...
-    union pattern to represent "`1` or `2`".
+*   An *open restriction*, the default, matches all values.
 
-2.  On the other hand, patterns are too complex. Because the proposal is
-    intended to be richly expressive, it defines a large number of different
-    kinds of patterns each with its own semantics. Handling every one of those
-    (and every pairwise combination of them) would add a lot of complexity to
-    the algorithm.
+*   A *constant restriction* matches only values that are identical to a given
+    constant value. It's used for enum values, `true`, and `false`.
```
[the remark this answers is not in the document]
munificent: I used a space whose type is `Null` to handle `null`. Every space has a type anyway, so that's sufficient.

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
...
+*   **Relational pattern:** The empty space union. *Relational patterns don't
+    reliably match any values, so don't help with exhaustiveness.*
+
+*   **Cast pattern:** *A cast pattern matches if the value is of the casted type
+    and the inner pattern matches it. But a cast pattern also "handles" a value
+    when the cast _fails_. From the perspective of exhaustiveness checking, what
+    matters it that execution can't flow out of a switch without matching at
+    least one case. But if a pattern _throws_ a runtime exception, then
+    execution also doesn't flow out of it. So we treat throwing as essentially
+    another kind of matching.*
```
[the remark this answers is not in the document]
munificent: Done.

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
...
+    test(A a) => switch (a) {
+      C(field: 0) as C => 0,
+      C _ => 1
+    };
     ```
 
-### Types in spaces
+    *After the first case, we know that if `a` is a `B` or `D` then we will have
+    thrown an exception. But if `a` is a `C`, it may not have matched if the
+    field isn't `0`. So the space for the first case is `B|C(field: 0)|D'.*
```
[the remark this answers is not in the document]
munificent: Good catch. There used to be a lot more in here about notation, but I don't think it's very useful for the new algorithm so I discarded most of it. Add a "union of" to clarify. The notation is informal now and hopefully self evident.

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
...
-        and the case is unreachable. Otherwise, the case is reachable and
-        `remaining` describes the values it might match.
+2.  Else, if either side is a union, then a space union of the intersection of
+    each operand of the union with the other space.
 
-Note that we can calculate reachability even for cases with guards. It's useful
-to tell if a case with a guard is completely unreachable. It's just that we
-ignore guards on *preceding* cases when determining if each case is reachable.
+    *In other words, distribute the intersection into the branches. So `(A|B) ^
+    C` is `(A^C)|(B^C)`.
```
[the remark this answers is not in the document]
munificent: Added some prose.

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
...
+3.  Else if both are constant restrictions and the constants are identical then
+    `true`.
 
-```
-Jack(suit: heart|club) - Card(suit: club) = Jack(suit: heart)
-```
+4.  Else if both are arity restrictions:
 
-The objects have different types now, but it still works out. Now consider:
+    1.  If `b.hasRest` then `true` if `a.size >= b.size`. *Since `b` has no
```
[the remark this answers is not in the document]
munificent: I try to be clear about that. In this case, the omission is deliberate. If `b.hasRest` is `false`, it falls through to the next cases.

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
...
+the second switch is clearly exhaustive because every value is a `Face` and the
+second case matches all instances of `Face`.
 
-```
-Face? - Face  expands to:  Face|Null - Face = Null
-Jack? - Null  expands to:  Jack|Null - Null = Jack
-```
+In the algorithm, when it looks at a space, it first *expands* it. If the type
+of a space is a sealed type, enum type, or `bool` (which is basically an enum
+type), then it replaces the space with a set of more precise spaces that cover
```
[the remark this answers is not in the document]
munificent: A little more like a sealed type than an enum type. A nullable type expands out to the underlying type and a `Null`-typed space. We could treat `null` like an enum value instead but I think the underlying algorithm would behave the same either way.

accepted/future-releases/0546-patterns/exhaustiveness.md
munificent: I don't have a strong opinion about this, but I think it's harmless to include redundant direct subtypes and I think the proposal's a little simpler and clearer to not make an exception for them.

accepted/future-releases/0546-patterns/exhaustiveness.md
munificent: Reworded to be from a declaration to a type.

`S` is a type, not a space (may be a confusing choice of letter). We then create a space with type `S`.

accepted/future-releases/0546-patterns/exhaustiveness.md
munificent: I don't think that's necessary, but I could be wrong. The proposal tries to consistently use a space of type `Null` with no constant restriction to deal with `null`.

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
...
+type* which is determined during type checking and is known for each pattern.
+The lifted space union for a pattern with matched value type `M` is:
+
+*   **Logical-or pattern:** A union of the lifted spaces of the two branches.
+
+*   **Logical-and pattern:** The intersection (defined below) of the lifted
+    space unions of the two branches.
+
+*   **Relational pattern:** The empty space union. *Relational patterns don't
+    reliably match any values, so don't help with exhaustiveness.*
```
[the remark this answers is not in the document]
munificent: I suppose, yes, it does reliably match `null` when the operator is `==` or `!=`. But the current implementation doesn't seem track that for exhaustiveness, and I'm fine with it not doing so. So the proposal is consistent with that.

Today, this is not considered exhaustive:

```dart
test(int? n) => switch (n) {
  == null => 'null',
  int _ => 'int'
};
```

But this is:

```dart
test(int? n) => switch (n) {
  null => 'null',
  int _ => 'int'
};
```

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +251,854 @@ also be an instance of one of the known set of subtypes. That gives us the
...
+    List(length: 0, hasRest: false)
+    List(length: 1, hasRest: false)
+    List(length: 2, hasRest: false)
+    List(length: 3, hasRest: false)
+    List(length: 4, hasRest: false)
+    List(length: 4, hasRest: true)
     ```
 
-## Space intersection
+*   Otherwise, `s` does not expand and the result is just `s`.
```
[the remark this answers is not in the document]
munificent: No, they will get unpacked incrementally as the algorithm does its depth-first traversal through the spaces.

accepted/future-releases/0546-patterns/exhaustiveness.md
munificent: Reworded to make it clearer that it applies to declarations.

accepted/future-releases/0546-patterns/exhaustiveness.md
munificent: Good question. No, we don't require constants that appear in constant patterns to have primitive equality. In practice, the only constants that actually usefully participate in exhaustiveness checking are bools and enum values, which do have primitive equality. So it mostly doesn't matter.

But it does potentially matter for reachability. Filed: https://github.com/dart-lang/sdk/issues/51873

Updated the spec here to match what that issue suggests.

accepted/future-releases/0546-patterns/exhaustiveness.md
munificent: I don't understand the distinction you're making here, sorry. Can you explain, or did Johhni's explanation of how overapproximation differs from I2B address this?

accepted/future-releases/0546-patterns/exhaustiveness.md
munificent: You know, I don't know. I just know that the implementation does handle this.

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -246,772 +267,878 @@ also be an instance of one of the known set of subtypes. That gives us the
...
     ```
-    (x: false, y: bool, z: bool) // Set x to D.
-    (x: bool, y: true, z: bool) // Set y to D.
-    (x: bool, y: bool, z: false) // Set z to D.
+    List(length: 0, hasRest: false)
+    List(length: 1, hasRest: false)
+    List(length: 2, hasRest: false)
+    List(length: 3, hasRest: false)
+    List(length: 4, hasRest: false)
+    List(length: 4, hasRest: true)
```
sgrekhov, line 996: Shouldn't it be `List(length: 5, hasRest: true)` here? `headMax` is `3` and `tailMax` is `2`, so `n==5` here

accepted/future-releases/0546-patterns/exhaustiveness.md
```diff
@@ -185,33 +189,46 @@ we would break that principle.
...
+  Card(this.suit);
 }
 
 class Pip extends Card {
   final int pips;
+
+  Pip(this.pips, super.suit);
 }
 
 sealed class Face extends Card {}
```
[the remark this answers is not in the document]
munificent: Nice catch. :)

accepted/future-releases/0546-patterns/exhaustiveness.md
munificent: Oops, yes.

munificent:
I'm going to go ahead and land this but I'm happy to iterate on it more if there are other changes to make.

accepted/future-releases/0546-patterns/exhaustiveness.md
lrhn: I'm not sure I follow it either, but that could be because I don't actually *understand* I2B.
Or what is being attempted here.

So let me try again:

Given a *static type* (which may contain free type variables referencing surrounding declarations),
take each such occurrence of a free type variable in the type, and replace it with its default type which is:
* If the referenced type variable has a bound, the default is the overapproximation of that bound.
* Otherwise if the free type variable occurs covariantly, the default is `Object?`.
* Otherwise the default is `Never`

The resulting type has no free type variables and will be a valid super-bounded type.

... But it's not valid super-bounded. If the variable has a bound with other free type variables, the first variable occurs contravariantly, we need an *underapproximation* to use instead.

And what if it occurs invariantly, like `List<void Function<X extends T>(X)>`?


I think it can make sense to talk about type variables in static types (I'm just not used to doing so), but I think here are more details to handle, if the goal is to get a valid super-bounded type as result.

PR #2956, opened 2023-03-29
Make it an error to have `on F` when `F` is final in another library.

lrhn:
Disallows a `mixin` declaration to have an other-library `final` type as `on` type.

A little rephrasing and refactoring to the rules, but that's the only intended change,
anything else is intended as clarification.

The structure of the restrictions on other libraries are now:
* Sealed - cannot subtype, no exceptions
* Final - cannot subtype, SDK exceptions.
* Interface - cannot extend/with, SDK exceptions
* Base/Final - transitively cannot implement, SDK exception.

Also same library:
* Base/Final - transitively must be base/final/sealed. No exceptions.

### DIFF
--- a/accepted/future-releases/class-modifiers/feature-specification.md
+++ b/accepted/future-releases/class-modifiers/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom, Lasse Nielsen
 
 Status: Accepted
 
-Version 1.7
+Version 1.8
 
 Experiment flag: class-modifiers
 
@@ -662,10 +662,10 @@ It's a compile-time error if:
 *   A declaration depends directly on a `sealed` declaration from another
     library. _No exceptions, not even for platform libraries._
 
-    More formally:
-    A declaration *D* from library *L* has a direct superdeclaration *S*
-    marked `sealed` (so necessarily a `class` declaration) in a library
-    different from *L*.
+    More formally, it's a compile-time error if:
+    A `class`, `mixin class`, `mixin` or `enum` declaration from library *L*
+    has a direct superdeclaration marked `sealed` _(so necessarily a `class`
+    declaration)_ from a library other than *L*.
 
     ```dart
     // a.dart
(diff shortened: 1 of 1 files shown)


accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -680,18 +680,48 @@ It's a compile-time error if:
...
+    library _(with some exceptions for platform libraries)_.
+
+    _(You cannot subtype a `final` declaration, except inside the same library.
+    Unless you are in a pre-feature library and the super-declaration is from
+    a platform library.)_
+
+    More formally, it's a compile-time error if:
+    A `class`, `mixin class`, `mixin` or `enum` declaration from library *L*
+    has a direct superdeclaration *S* marked `final` _(so necessarily a `class`
+    declaration)_ in a library *K*, neither:
```
munificent, line 693: I would do "neither" -> "and neither of the following is true".

And then remove the "nor," from the first bullet.

Likewise further below.

munificent (approved):
LGTM. I would get @kallentu to review this too.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -715,13 +745,13 @@ It's a compile-time error if:
...
-    A declaration *C* in library *L* has a declared interface *P*,
-    and *P* has any superdeclaration *S*, from a library *K*,
-    which is marked `base` or `final` _(including *S* being *P* itself)_,
-    and neither:
-    * *K* and *L* is the same library, mor
-    * *K* is a platform library and *L* is a pre-feature library.
+    More formally, it's a compile-time error if:
+    A `class`, `mixin-class`, `mixin` or `enum` declaration in library *L* has
+    a declared interface *P*, and *P* has any superdeclaration *S*,
+    from a library *K*, which is marked `base` or `final`
```
kallentu, line 751: Also, since you specify behaviour for `final` declaration subtyping on line 683, maybe move these `final` examples (that aren't already in the `final` section) over to there and remove talking about both `base` and `final` in this section?

It's nice to read about each modifier one at a time.

kallentu (approved):
Thank you for updating the spec!

accepted/future-releases/class-modifiers/feature-specification.md
lrhn: This rule might just be about `base`, because it's impossible to have a super-declaration marked `final` in another library without *also* having a super-declaration marked `base` in that other library, because that's the only possible way to have a direct super-declaration from that library with the `final` declaration as super-declaration.

(I think. It's always worrisome to define rules that depend on *other* rules to be complete. If we remove or tweak the other rule, then this rule might start allowing something we didn't intend.
So, be complete in every rule, as if it was stand-alone, or be minimal and non-redundant? I'm aiming for the former.)

PR #2957, opened 2023-03-29
modernize the specification script

jakemac53:
This isn't runnable at all on Dart 3.0, updated it to Dart 2.12 so it is opted in to null safety and removed the `utf8` dep which was never migrated and is discontinued.

### DIFF
--- a/specification/pubspec.yaml
+++ b/specification/pubspec.yaml
@@ -1,8 +1,7 @@
 name: scripts
 publish_to: none
 environment:
-  sdk: ">=2.10.0 <3.0.0"
-dependencies: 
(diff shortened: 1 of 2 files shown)


jakemac53:
@eernstg not sure if you have some idea about the failing jobs here, or they can just be ignored?

eernstg:
The failing checks are most likely occurring because a firebase authentication token (or something like that) needs to be recreated. This happens about once every 3 months.

@athomas, I'm afraid we haven't found a way to do this without finding someone with certain rights. Can you do this again?

eernstg (approved):
LGTM, and thanks!

PR #2964, opened 2023-03-30
Add a github actions workflow for dart analysis and tests

jakemac53:
This may be controversial, but I wanted to send it out anyways to see what kind of reaction it gets :).

This also migrates an old benchmark to null safety so it can be ran with Dart 3. I will separate out that change if we don't land this PR. I already submitted other fixes upstream, I just missed that one :).

## Why is it worth enforcing clean analysis?

This is helpful because otherwise all the diagnostics become pointless/unusable, as you are drowned in a sea of them (there were well over 100 until yesterday, and I previously cleaned up a similar amount). It also just looks unprofessional IMO if our repo has a bunch of diagnostics when you clone it.

This repo is constantly collecting new diagnostics today, so the problem just gets worse and worse. I suspect most of the language team are not heavy users of IDEs which report diagnostics for the entire workspace, and so it just isn't as present for most of us (if you only see the ones for open files it is less annoying)? But I see them all :), and I suspect most external users cloning the repo would also see them.

Another argument here is this will ensure that any example code we have is _actually valid_, at least once such validation is possible. And it won't trigger diagnostics when copied into users repos (for the most part, they could enable extra lints though).

### How can I create examples for code that doesn't work yet? 

It is definitely intended and supported that we will have invalid Dart code in this repo, but we should simply exclude that code from analysis using the `analysis_options.yaml` file. This is a one liner change when adding a new proposal that has example code, and thus not a heavy burden.

## What about formatting?

I am not requiring formatted code right now because the formatter [doesn't support excluding files](https://github.com/dart-lang/dart_style/issues/864). Otherwise I would. I do have a wrapper around the formatter which does support that if we were willing to take a dependency on it, but I presume that would be a bridge too far.

## Won't this be a maintenance burden?

Yes, some amount of ongoing work will be needed here, but it will prevent us from accumulating larger amounts of technical debt. I would rather deal with small bits of maintenance over time (adding excludes where necessary mostly).

We could also decide to use different lint sets if we don't think the existing lints are providing value. Or no lints. I don't really care, I just don't want lints enabled that are being ignored :).

### DIFF
--- a/accepted/2.3/spread-collections/benchmarks/bin/profile.dart
+++ b/accepted/2.3/spread-collections/benchmarks/bin/profile.dart
@@ -13,8 +13,8 @@ final lists = <List<String>>[];
 final maps = <Map<String, int>>[];
 final mapsWithSplay = <Map<String, int>>[];
 
-String _configuration;
-int _length;
+late String _configuration;
+late int _length;
 bool _csvOutput = false;
 bool _warmingUp = true;
 
@@ -136,7 +136,7 @@ void benchmarkMaps(String collection, List<Map<String, int>> maps) {
 
 double benchmarkIterable(
     String method, void Function(Iterable<String>, List<String>) action,
-    [double overhead]) {
+    [double? overhead]) {
   var time = bench(() {
     for (var i = 0; i < lists.length; i++) {
       var from = lists[i];
@@ -155,7 +155,7 @@ double benchmarkIterable(
 
 double benchmarkList(
     String method, void Function(List<String>, List<String>) action,
-    [double overhead]) {
+    [double? overhead]) {
   var time = bench(() {
     for (var i = 0; i < lists.length; i++) {
       var from = lists[i];
@@ -177,7 +177,7 @@ double benchmarkMap(
     String method,
     List<Map<String, int>> maps,
     void Function(Map<String, int>, Map<String, int>) action,
-    [double overhead]) {
+    [double? overhead]) {
   var time = bench(() {
     for (var i = 0; i < maps.length; i++) {
       var from = maps[i];
@@ -250,7 +250,7 @@ void show(String collection, String method, double overhead, double baseline) {
 
 double bestTime(String collection, String method) {
   var runs = _runs["$collection $method"];
-  var min = runs.first;
+  var min = runs!.first;
   for (var i = 1; i < runs.length; i++) {
     if (runs[i] < min) min = runs[i];
   }
@@ -259,7 +259,7 @@ double bestTime(String collection, String method) {
 }
 
 double standardDeviation(String collection, String method) {
-  var runs = _runs["$collection $method"];
+  var runs = _runs["$collection $method"]!;
 
   var mean = runs.fold<double>(0.0, (a, b) => a + b) / runs.length;
 
@@ -281,9 +281,9 @@ void iterableIterator(Iterable<String> from, List<String> to) {
 }
 
 void iterableForEach(Iterable<String> from, List<String> to) {
-  var temp = to;
+  List<String>? temp = to;
   from.forEach((s) {
-    temp.add(s);
+    temp!.add(s);
   });
   temp = null;
 }
@@ -317,9 +317,9 @@ void listSubscript(List<String> from, List<String> to) {
 }
 
 void listForEach(List<String> from, List<String> to) {
-  var temp = to;
+  List<String>? temp = to;
   from.forEach((s) {
-    temp.add(s);
+    temp!.add(s);
   });
   temp = null;
 }
@@ -334,14 +334,14 @@ void mapEntries(Map<String, int> from, Map<String, int> to) {
 
 void mapKeys(Map<String, int> from, Map<String, int> to) {
   for (var key in from.keys) {
-    to[key] = from[key];
+    to[key] = from[key]!;
   }
 }
 
 void mapForEach(Map<String, int> from, Map<String, int> to) {
-  var temp = to;
+  Map<String, int>? temp = to;
   from.forEach((key, value) {
-    temp[key] = value;
+    temp![key] = value;
   });
   temp = null;
 }
--- a/accepted/2.3/spread-collections/benchmarks/bin/profile.dart
+++ b/accepted/2.3/spread-collections/benchmarks/bin/profile.dart
@@ -249,8 +249,8 @@ void show(String collection, String method, double overhead, double baseline) {
 }
 
 double bestTime(String collection, String method) {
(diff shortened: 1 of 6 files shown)


munificent:
> I suspect most of the language team are not heavy users of IDEs which report diagnostics for the entire workspace

I think the language team *should* be IDE users because that's how our users use Dart and we need to understand their experience to do our job well.

Whether we should load the entire language repo into a workspace is a separate question. I've never done that. I'm not sure why I *would* do that. I didn't even really consider that it was a thing you could do. :)

> It is definitely intended and supported that we will have invalid Dart code in this repo, but we should simply exclude that code from analysis using the `analysis_options.yaml` file.

In practice, I think most code for proposed language features lives in Markdown files or the issue tracker, so I'm not particularly worried about this.

jakemac53:
> Whether we should load the entire language repo into a workspace is a separate question. I've never done that. I'm not sure why I _would_ do that. I didn't even really consider that it was a thing you could do. :)

I think it is pretty common for workspaces to map to whole repos - except for exceptionally large repos. The language repo is easily handled by the analyzer.

natebosch:
> I think the language team _should_ be IDE users because that's how our users use Dart and we need to understand their experience to do our job well.

If you have strong preferences for any particular editor there is a very strong chance you can wire up the analyzer and see diagnostics as you edit. I do think that it helpful to have a feel for what it's like editing dart with the diagnostics and autocomplete that our users tend to rely on.

lrhn:
> The language repo is easily handled by the analyzer.

I'm using VSCode, and the Dart language server is *incredibly slow*, and almost never gives me anything useful when working with the platform libraries (probably to be expected, it's not clear how it should recognize that `sdk/lib/core/core.dart` is the same as `dart:core`).
I'm very often using "Save without Formatting" because I don't want to wait for formatting, if it ever happens.
(But asking for issues in all files, not just the current file, shows me 45K+ issues, so maybe including all our negative tests is the real problem here.)

accepted/2.3/spread-collections/benchmarks/bin/profile.dart
```diff
@@ -250,7 +250,7 @@ void show(String collection, String method, double overhead, double baseline) {
 
 double bestTime(String collection, String method) {
   var runs = _runs["$collection $method"];
-  var min = runs.first;
+  var min = runs!.first;
```
lrhn, line 253: I'd move the `!` to the previous line, like below.

accepted/2.3/spread-collections/benchmarks/bin/profile.dart
```diff
@@ -281,9 +281,9 @@ void iterableIterator(Iterable<String> from, List<String> to) {
 }
 
 void iterableForEach(Iterable<String> from, List<String> to) {
-  var temp = to;
+  List<String>? temp = to;
   from.forEach((s) {
-    temp.add(s);
+    temp!.add(s);
```
lrhn, line 286: Same here, move it to `= to!;`

leafpetersen:
> > I suspect most of the language team are not heavy users of IDEs which report diagnostics for the entire workspace
> 
> I think the language team _should_ be IDE users because that's how our users use Dart and we need to understand their experience to do our job well.

I try this, periodically, but the problem is that our use case is not well supported, and probably shouldn't be.  The analyzer team would like to make editing the core libraries work well in the IDE, but it's a P3, which means it's never going to happen.  You can sort of somewhat get it to work in some cases, but it requires manual reconfiguration of your IDE every time you switch between editing core library code and non-core library code, and it doesn't help you with patch files, etc, etc.  

Editing tests also hasn't worked well for me in the past, for different reasons - the IDEs often make assumptions about workspaces that don't map well to standalone files.  I've had discussions with the team about this in the past, and I know they've fixed some issues, but at the end of the day, it's not really a high priority use case.

Editing anything in the sdk directory with an IDE is actually fraught.  There was just a thread about this somewhere recently.  I think Nick knows how to set up the IDE to ignore test directories and a bunch of other things, but if you don't know how to do that, the analyzer slows to a crawl.  

TL;DR we aren't our users.

jakemac53:
> I try this, periodically, but the problem is that our use case is not well supported, and probably shouldn't be. The analyzer team would like to make editing the core libraries work well in the IDE, but it's a P3, which means it's never going to happen.

This is only really true for the core libraries and patch files. I do all SDK work in VsCode and it works just fine (as long as you are working from a remote machine with 512GB of ram at least 😄). But more seriously, just put this analysis_options.yaml file at the root of your SDK:

```
analyzer:
  exclude:
    - out/
    - benchmarks/ # has legacy benchmarks that aren't valid in 3.0, not sure the plan for these
    - runtime/
    - sdk/
    - tests/
    - third_party/
    - utils/peg/ # can be removed soon
    - utils/tests/ # can be removed soon
```

jakemac53:
> I try this, periodically, but the problem is that our use case is not well supported, and probably shouldn't be. The analyzer team would like to make editing the core libraries work well in the IDE, but it's a P3, which means it's never going to happen.

This is only really true for the core libraries and patch files. I do all SDK work in VsCode and it works just fine (as long as you are working from a remote machine with 512GB of ram at least 😄). But more seriously, just put this analysis_options.yaml file at the root of your SDK:

```
analyzer:
  exclude:
    - out/
    - benchmarks/ # has legacy benchmarks that aren't valid in 3.0, not sure the plan for these
    - runtime/
    - sdk/
    - tests/
    - third_party/
    - utils/peg/ # can be removed soon
    - utils/tests/ # can be removed soon
```

jakemac53:
> I try this, periodically, but the problem is that our use case is not well supported, and probably shouldn't be. The analyzer team would like to make editing the core libraries work well in the IDE, but it's a P3, which means it's never going to happen.

This is only really true for the core libraries and patch files. I do all SDK work in VsCode and it works just fine (as long as you are working from a remote machine with 512GB of ram at least 😄). But more seriously, just put this analysis_options.yaml file at the root of your SDK:

```
analyzer:
  exclude:
    - out/
    - benchmarks/ # has legacy benchmarks that aren't valid in 3.0, not sure the plan for these
    - runtime/
    - sdk/
    - tests/
    - third_party/
    - utils/peg/ # can be removed soon
    - utils/tests/ # can be removed soon
```

accepted/2.3/spread-collections/benchmarks/bin/profile.dart
lrhn: ACK. Not great code to begin with. 
`from.forEach(to!.add);` would be better, but even that's really just `to!.addAll(from)`.

PR #2965, opened 2023-03-30
Remove "mixin" from some positions in the @reopen text.

srawlins:
Regarding reopened mixins and reopening mixins, I think @eernstg [worded it well](https://github.com/dart-lang/linter/pull/4235#discussion_r1152164227):

> It is not possible to reopen a mixin or a mixin class: It has the modifier `base` or no modifier, and no subtype of a `base` declaration can exist without a modifier which is at least as strong as `base`.

Based on this, I made these changes:

* removed "or mixin" from the things which can be annotated with `@reopen`
* removed "or mixin" from the things which can be extended or mixed in and be modified with `interface` or `final`
* spread a sentence which has 7 "or"s in it 🤣 across two bullet points.

CC @pq @kallentu @munificent

### DIFF
--- a/accepted/future-releases/class-modifiers/feature-specification.md
+++ b/accepted/future-releases/class-modifiers/feature-specification.md
@@ -905,8 +905,8 @@ warning is reported if a class is not annotated `@reopen` and it:
 
 *   extends a class marked `interface` or `final`
     and is not itself marked `interface` or `final`, or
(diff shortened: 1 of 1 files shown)


srawlins:
Haha, I accepted your suggestions, and now there is a CLA complaint :/

kallentu (approved):

Issue #1930, opened 2021-10-21
Define the macro API for introspecting on metadata annotations (and enum values)


munificent:
The proposal has a section discussing this, but it's written like an open-ended design discussion. We should figure out what we actually want and then pin it down in the proposal. The key question is how arguments in metadata annotations are seen by the macro. Are they actual values, or unevaluated syntax? The former might not be possible (since it may refer to code produced by macros). The latter might not be useful.

GregoryConrad:
From a user's perspective, it would be very helpful to access annotations that depend on constant macro-produced types and declarations. 

Say I have:

```dart
class InfoAnnotation {
  const InfoAnnotation(Object obj);
}

@someFunctionDeclarationMacro
void fn(
  @InfoAnnotation(macroProducedObj) int someIntParameter,
) {}
```

In the above example, `someFunctionDeclarationMacro` could do something interesting with the `macroProducedObj` and I think that it is reasonable that it should be able to get a `Declaration` object from it.

I think I see why this would be problematic as-is: there is no way to know for certain whether this `macroProducedObj` has been declared yet in phase 2 (if I am wrong here, please do correct me).

Would it be possible to hint at the order in which some phase 2 macros are evaluated to allow for this? I would hate to have to resort to build_runner to handle something like this.

I have what I think is a good solution/compromise: a macro can directly specify what other macros it "depends on." I.e., In the example above, `someFunctionDeclarationMacro` "depends on" the macro that produces the `macroProducedObj`. For my particular use case, and I imagine most others, this would be sufficient.

Seeing as a macro that that depends on declarations produced by another macro would (in most cases) have access to the other macro, the above solution makes a lot of sense to me. It would act just like dependency resolution, giving a deterministic way to run macros in phase 2.

An _anti-solution_ would be for a macro class itself to ask to be run at a certain point in phase 2, perhaps even a "phase 2.5." Obviously, nothing like this should be supported.

GregoryConrad:
I actually just realized my own use case relies on having the macro depend on other instances where it was applied; which is an issue. But, I know exactly what I want to declare in phase 1 given the `FunctionDeclaration` the macro is applied to, so then in phase 2 my macro can use those previous declarations. Will it be possible to contribute more information in a phase than is required? That seems reasonable. I have some other ideas in which this would ease implementation.

GregoryConrad:
> Is the `macroProducedObj` generated from a macro in the same library, or a dependency?

[...]

PR #2976, opened 2023-04-03
ban type parameters for macros

jakemac53:
This came up in https://github.com/dart-lang/language/issues/1930.

I also removed the changelog section for now, it hasn't been updated consistently and has only one very old entry. Will bring it back once the proposal moves to accepted.


rrousselGit:
It's a bit tangential to generic parameters but does this limitation imply that macros' annotated element cannot contain generated types?

Like:

```dart
@myMacro
int fn(ClassGeneratedByMyMacro param) => 0;
```

munificent (approved):

PR #2983, opened 2023-04-06
Move shipped proposals out of "future-releases".

munificent:
I also fixed any broken references to "future-releases" that I could find within the language repo, including a few already-broken ones.


leafpetersen (approved):
Thanks!

PR #2993, opened 2023-04-11
[Class modifiers] Removed mentions of interface and final mixins

sgrekhov:
Mixins cannot be `interface` or `final`

### DIFF
--- a/accepted/future-releases/class-modifiers/feature-specification.md
+++ b/accepted/future-releases/class-modifiers/feature-specification.md
@@ -680,16 +680,16 @@ It's a compile-time error if:
     class M with S {} // Error, for several reasons.
     ```
 
-*   A class extends or mixes in a declaration marked `interface` or `final`
-    from another library _(with some exceptions for platform libraries)_.
+*   A class extends a declaration marked `interface` or `final` from
+    another library _(with some exceptions for platform libraries)_.
 
     _(You cannot inherit implementation from a class marked `interface`
     or `final` except inside the same library. Unless you are in a
     pre-feature library and you are inheriting from a platform library.)_
 
     More formally:
-    A declaration *C* from library *L* has a declared superclass or mixin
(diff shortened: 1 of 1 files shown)


eernstg (approved):
This looks good to me. @munificent, @lrhn, do you have further comments?

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -680,16 +680,16 @@ It's a compile-time error if:
     class M with S {} // Error, for several reasons.
     ```
 
-*   A class extends or mixes in a declaration marked `interface` or `final`
-    from another library _(with some exceptions for platform libraries)_.
+*   A class extends a declaration marked `interface` or `final` from
+    another library _(with some exceptions for platform libraries)_.
```
lrhn, line 684: Well, it's true that it's a compile-time error, because you cannot mix in a non-mixin class, and if it's marked `interface` or `final`, it's a non-`mixin` class.

So not wrong, just redundant, and actually deliberate redundancy (I did it on purpose!).

Because of pre-feature libraries using platform libraries, it's possible that they *do* have
a `final class` as a mixin declaration.
I like that it's explicitly mentioned here that that is *not* an error.

But if that was the goal, it's not sufficient. Then it should also mention `enum` declarations,
which can have mixin declarations too.

It is (probably) correct that there won't be any more, or fewer, compile-time errors if we remove the mixin-restriction from here, but I'd still like to keep it. 
Just to be sure.

eernstg: OK, I think the Solomonic decision would be to eliminate the change in lines 683-684.

It's just so _nice_ to ignore legacy libraries, but we can't do it a 100%, yet. ;-)

eernstg:
OK, thanks for looking into this anyway!

Issue #2998, opened 2023-04-13
Missing statement?


eernstg:
The changelog entry about version 2.7 in the [class-modifiers feature specification](https://github.com/dart-lang/language/blob/master/accepted/future-releases/class-modifiers/feature-specification.md) says that it should

> State that `enum` declarations count as `final`.

The implication would be that certain declarations are not errors, even though they require certain modifiers on certain subtypes:

```dart
base class B {}
enum E implements B {} // OK, even though there is no `base`, `final`, or `sealed` modifier.
```

This makes sense, because every `enum` declaration is already restricted in ways that are at least as strong as the ones implied by those class modifiers.

I tried to find the location where that statement occurs, but I can't find it. Perhaps it was omitted by accident?

There is no implementation effort, the analyzer and CFE already behave in this manner.

/cc @munificent.

eernstg:
Cf. https://github.com/dart-lang/language/pull/2999.

PR #2999, opened 2023-04-13
Mention that enums are implicitly final

eernstg:
Cf. https://github.com/dart-lang/language/issues/2998: This PR adds a paragraph where it is stated that every enum declaration is considered to be a `final` declaration with respect to the rules where certain class modifiers are required to exist.

The changelog says that the feature spec already says so, but I couldn't find it.

### DIFF
--- a/accepted/future-releases/class-modifiers/feature-specification.md
+++ b/accepted/future-releases/class-modifiers/feature-specification.md
@@ -657,6 +657,13 @@ we can specify the restrictions imposed by modifiers._
 
 ### Basic restrictions
 
+With respect to compile-time errors caused by missing required class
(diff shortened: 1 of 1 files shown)


munificent (approved):

sgrekhov:
Does it makes sense to add that enum is `final` even in a pre-feature library? For example we cannot mix in a `base` class in a pre-feature library (because we need `base` of `final` modifiers to do that but they are unavailable in a pre-feature world). But it is possible with an enum because if is implicitly `final`
```dart
// lib.dart
base mixin class BaseMixinClass {}

// test.dart
// @dart=2.19

enum E with BaseMixinClass {} // no error. `enum` is final and therefore is allowed to have `base` superdeclaration
```

PR #3022, opened 2023-04-27
Allow `extends Object` and `= Object with N` for mixin classes.

lrhn:
@leafpetersen @eernstg 

If that is what we want, this might do it.

### DIFF
--- a/accepted/future-releases/class-modifiers/feature-specification.md
+++ b/accepted/future-releases/class-modifiers/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom, Lasse Nielsen
 
 Status: Accepted
 
-Version 1.7
+Version 1.8
 
 Experiment flag: class-modifiers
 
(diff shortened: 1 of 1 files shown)


munificent:
What's the context on this? The proposal is essentially out the door, so it seems like this should probably be handled as a separate language change, right?

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -813,10 +814,23 @@ ensures that it can be used as both a `class` and a `mixin`.
 It's a compile-time error if a `mixin class` declaration:
 *   has an `interface`, `final` or `sealed` modifier. _This is baked
     into the grammar, but it bears repeating._
-*   has an `extends` clause,
-*   has a `with` clause, or
+*   does not have `Object` from `dart:core` as superclass.
```
leafpetersen, line 817: Maybe say "immediate superclass" just to be 100% clear?

leafpetersen (approved):

leafpetersen:
> What's the context on this? The proposal is essentially out the door, so it seems like this should probably be handled as a separate language change, right?

See dart-lang/sdk#52062 for more context, but TL;DR this brings the spec into alignment with what we shipped.  I'm happy to discuss further, but I think this is probably the right way to move forward.

munificent (approved):

PR #3023, opened 2023-04-27
Specify primary constructors

eernstg:
This PR adds a proposed feature specification for primary constructors, applicable to normal classes as well as inline classes.


eernstg:
A bunch of different declarations can be explored by means of `scripts/show_primary_constructors.dart`. Try

```
> cd scripts
> dart show_primary_constructors.dart --show-normal --show-struct class.json
```

to see a single example. Try `*.json` to see many different variants, etc.

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,437 @@
...
+needed then it is always possible to express the same thing as a
+non-primary constructor (i.e., any constructor which isn't a primary
+constructor).
+
+The formal parameter declarations use the regular syntax, specified in the
+grammar by the non-terminal `<formalParameterList>`.
+
+This implies that there is no way to indicate that the instance variable
+declarations should have the modifiers `covariant`, `late`, or `external`
+(because formal parameters cannot have those modifiers). This omission is
```
lrhn, line 75: Formal parameters *can* be `covariant`.

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,437 @@
...
+
+<inlineMemberDeclaration> ::= <classMemberDeclaration>;
+```
+
+The word `inline` is now used in the grammar, but it is not a reserved word
+or a built-in identifier.
+
+### Static processing
+
+A class declaration (of any kind including `inline`) with a primary
```
lrhn, line 372: Except mixin application class declarations?

They can't declare instance variables anyway, and get forwarded constructors from the mixin application.
But maybe this can be a way to declare a different constructor, which can only have `super.x` parameters,          and which might suppress adding forwarding constructors?

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,437 @@
...
+   | ';';
+
+<inlineClassDeclaration> ::=
+     'final'? 'inline' 'class' <classNamePart> <interfaces>? <inlineClassBody>;
+
+<inlineClassBody> ::=
+     '{' (<metadata> <inlineMemberDeclaration>)* '}'
+   | ';';
+
+<inlineMemberDeclaration> ::= <classMemberDeclaration>;
```
lrhn, line 364: Should it be available on enums? They seem like prime candidates.

working/2365 - primary constructors/feature-specification.md
eernstg: Ah, of course! The constructor parameter error for `covariant` is not a syntax error, it's a regular compile-time error. OK, so I'll just include support for `covariant` for consistency.

working/2365 - primary constructors/feature-specification.md
eernstg: Indeed! Added support for that, and an example. Looks very nice!

working/2365 - primary constructors/feature-specification.md
eernstg: Let's just keep them out for now. That's already the case because the grammar rules don't include them.

eernstg (commented):
Review response.

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,472 @@
...
+instance variable for each formal parameter in said parameter list.
+
+A primary constructor cannot have a body, and it cannot have an initializer
+list (and hence, it cannot have a superinitializer, e.g., `super(...)`, and
+it cannot have assertions). 
+
+The motivation for these restrictions is that a primary constructor is
+intended to be small and easy to read at a glance. If more machinery is
+needed then it is always possible to express the same thing as a
+non-primary constructor (i.e., any constructor which isn't a primary
```
mit-mit, line 67: maybe s/non-primary constructor/regular constructor ?

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,472 @@
...
+constructor).
+
+The formal parameter declarations use the regular syntax, specified in the
```
mit-mit, line 70: This is a very detailed statement given we're still in the Introduction.

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,472 @@
...
+  final int x;
+  final int y;
+  const D.named(this.x, [this.y = 0]);
+}
+
+class const D.named<TypeVariable extends Bound>(int x, [int y = 0])
+    extends A with M implements B, C;
+```
+
+There was a proposal from Bob that the primary constructor should be
```
mit-mit, line 276: This is getting quite discussion based, and long (we're still in the Introduction?)

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,472 @@
...
+```dart
+// Current syntax.
+
+class Point {
+  int x;
+  int y;
+  Point(this.x, this.y);
+}
+
+// Same thing using a primary constructor.
```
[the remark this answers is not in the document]
eernstg: Done. Also rewrote the first example to make a few things more explicit, including the fact that every example shows two class declarations with the same meaning.

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,472 @@
...
+[kotlin primary constructors]: https://kotlinlang.org/docs/classes.html#constructors
+[scala primary constructors]: https://www.scala-lang.org/files/archive/spec/2.11/05-classes-and-objects.html#constructor-definitions
+[scala primary constructor examples]: https://www.geeksforgeeks.org/scala-primary-constructor/
+[primary-constructors label]: https://github.com/dart-lang/language/issues?q=is%3Aissue+is%3Aopen+primary+constructor+label%3Aprimary-constructors
+
+## Introduction
+
+Primary constructors is a conciseness feature. It does not provide any new
+semantics at all. It just allows us to express something which is already
+possible in Dart, using a less verbose notation. Here is a simple example:
```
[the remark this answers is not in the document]
eernstg: Done.

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,472 @@
...
+because they will work exactly as the example using the current syntax.
+
+Note that an empty class body, `{}`, can be replaced by `;`.
+
+The basic idea is that a parameter list that occurs just after the class
+name specifies both a constructor declaration and a declaration of one
+instance variable for each formal parameter in said parameter list.
+
+A primary constructor cannot have a body, and it cannot have an initializer
+list (and hence, it cannot have a superinitializer, e.g., `super(...)`, and
```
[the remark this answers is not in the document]
eernstg: Do you think it would be better to postpone the information about the initializer list as a whole? (The fact that we can't have `super(...)` is just one of the consequences of not having an initializer list.)

The reason why I mentioned this already at this point is that it seems so fundamental that we can't have several syntactic elements (body, initializer list).

But I'll add a few words about the superinitializer when the super parameters are mentioned.

working/2365 - primary constructors/feature-specification.md
eernstg: I did try to use several different terms for this. I'm slightly worried that it is too generic if we say 'regular constructor', because this could also mean "a generative constructor" (that is, any constructor which isn't a `factory`), or "a non-redirecting constructor", or "a non-constant constructor", and perhaps even more if we introduce implicit constructors etc.

Did not change anything so far.

working/2365 - primary constructors/feature-specification.md
eernstg: Done.

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,510 @@
...
+   | <inlineClassDeclaration> // New alternative.
+   | ...;
+
+<classDeclaration> ::= // First alternative modified.
+     (<classModifiers> | <mixinClassModifiers>)
+     'class' <classNamePart> <superclass>? <interfaces>? <classBody>
+   | ...;
+
+<classNamePart> ::= // New rule.
+     'const'? <constructorName> <typeParameters>? <formalParameterList>
```
lrhn, line 299: I think the position of type parameters should be `C<T>.foo(args)`, not `C.foo<T>(args)`.
That's consistent with how the constructor is called.

So 
```ebnf
    'const'? <identifier><typeParameters>?('.' (<identifier>|'new'))? <formalParameterList>
```

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,510 @@
...
+occur as follows:
+
+The current scope of the formal parameter list of the primary constructor
+in _D_ is the current scope of the class/enum declaration *(in other words,
+the default values cannot see declarations in the class body)*.  Every
+default value in the primary constructor of _D_ is replaced by a fresh
+private name `_n`, and a constant variable named `_n` is added to the
+top-level of the current library. *(This means that we can move the
+parameter declarations including the default value without changing its
+meaning.)*
```
lrhn, line 368: This changes the context type of the constant, which will make the following fail:
```dart
class Foo(final double d = 0);
class Bar(final List<int> l = []);
```
The only solution I can see, which doesn't depend on waiting until after type inference, is to copy the type expression of the parameter into the constant variable declaration.

That gives us another problem, when the type depends on type parameters:
```dart
class Foo<T>(final List<T> list = const []);
```
Here we will, eventually, infer a context type of `List<T>` for `const[]`, then infer a type argument of `const <Never>[]`. But that happens so much later that we cannot delay rewriting until we understand what types are. 
(This rewrite probably has to happen before we even validate the type hierarchy, so we don't know what a "type" is yet.)
If we just extract the literal into `const _n = const [];` we get a `List<dynamic>` instead, which is not a valid default value.

We can either do some symbolic processing of the type expression (that's never going to end well), or just admit defeat and say that when we copy the expression, we *include its lexical scope*. Assign a scope to every expression. Or allow some expressions to have lexical scope overrides. (Or whatever we can say which matches what the front-end actually does.)
That assumes the rewriting happens after we have introduced the top-level scope.

But it's not a purely syntactic rewrite. _They never are._

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,510 @@
...
+
+Next, _k_ is a constant constructor iff the keyword `const` occurs just
+before the class name in the header of _D_, or _D_ is an `enum`
+declaration.
```
lrhn, line 372: Which means that _k_ is prefixed with the `const` modifier. Stay syntactic, we're creating new syntax here.
Saying that a constructor "exists" and "is const" at a point where the syntax of the program hasn't stabilized yet, is too soon.

(Unless you really are trying to introduce a *semantic* constructor, like the ones we infer from parsing constructor syntax, which I'd be quite happy about. But that's not what it says. I think. The distinction isn't definitely not made clear.)

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,510 @@
...
+
+Where no processing is mentioned below, _D2_ is identical to _D_. Changes
+occur as follows:
+
+The current scope of the formal parameter list of the primary constructor
+in _D_ is the current scope of the class/enum declaration *(in other words,
+the default values cannot see declarations in the class body)*.  Every
+default value in the primary constructor of _D_ is replaced by a fresh
+private name `_n`, and a constant variable named `_n` is added to the
+top-level of the current library. *(This means that we can move the
```
lrhn, line 366: Here you're rewriting *D*, but not into *D2*. 

When later steps refer to *D*, is it the one before or after this change? (Probably after.)

In other words: Is this intended as an iterative rewrite, with various intermediate steps, or as an atomic rewrite from *D* to *D2*, where *D2* includes all the changes relative to *D*?

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,510 @@
...
+The formal parameter list _L2_ of _k_ is identical to _L_, except that each
+formal parameter is processed as follows.
+
+In particular, the formal parameters in _L_ and _L2_ occur in the same
+order, and mandatory (respectively optional) positional parameters remain
+mandatory (respectively optional, with the same default value, if any), and
+named parameters preserve the name and the property of being `required`.
+
+- An initializing formal parameter *(e.g., `this.x`)* is copied from _L_ to
+  _L2_ without changes.
```
lrhn, line 394: Initializing formal parameters can have default values. If the rewrite of those into constant variables above was considered done on *D*, then it doesn't need mentioning again. If the rewrite is intended to be atomic, it needs to be mentioned here (and probably *done* here).

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,510 @@
...
+_D2_ omits the formal parameter list _L_ that follows the class name and
+possibly `.id` in _D_.
+
+The formal parameter list _L2_ of _k_ is identical to _L_, except that each
+formal parameter is processed as follows.
+
+In particular, the formal parameters in _L_ and _L2_ occur in the same
+order, and mandatory (respectively optional) positional parameters remain
+mandatory (respectively optional, with the same default value, if any), and
+named parameters preserve the name and the property of being `required`.
```
lrhn, line 391: As a syntactic rewrite, I'd say that the parameter list is copied, including its uses of `[...]` and `{..}`, then each formal parameter of it is potentially modified as follows:

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,510 @@
...
+
+There are several reasons why this is not supported. First, primary constructors
+should be small and easy to read. Next, it is not obvious how the
+superconstructor arguments would fit into a mixin application (e.g., when the
+superclass is `A with M1, M2`), or how readable it would be if the
+superconstructor is named (`class B(int a) extends A.name(a);`). For instance,
+would it be obvious to all readers that the superclass is `A` and not `A.name`,
+and that all other constructors than the primary constructor will ignore the
+implied superinitialization `super.name(a)` and do their own thing (which might
+be implicit).
```
lrhn, line 429: We could also allow `extends A.name` with no arguments, just to choose where `super.x` parameters are forwarded.

Parsing-wise we then have to decide whether `id1.id2` is a prefix + class-name or a class-name+constructor-name. (And accept `id1.id2.id3` for the prefix-imported named constructor.)

Probably not an problem, it's only necessary to understand the identifiers at the point where we begin resolving identifiers, and there we can see what the identifiers point to.

working/2365 - primary constructors/feature-specification.md
eernstg: Now using 'body constructor'.

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,510 @@
...
+
+class Point(int x, int y);
+```
+
+These examples will serve as an illustration of the proposed syntax, but
+they will also illustrate the semantics of the primary constructor
+declarations, because those declarations work exactly the same as the
+declarations using the current syntax.
+
+Note that an empty class body, `{}`, can be replaced by `;`.
```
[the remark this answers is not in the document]
lrhn: ACK. I'm considering whether we want to remove `class C = S with M;` from the language, and perhaps let `class C extends S with M;` to be the new syntax for the same thing. 
That obviously only works when we are not introducing members into `C` with a primary constructor.

The distinction between `class C extends Object with M {}` and `class C = Object with M;` is that the latter can be a mixin class, it has `Object` as superclass, with `M` mixed in on top, the former cannot because it has `Object&M` as superclass.

(But we could also just expand the `mixin` declaration to allow defining composite mixins, then renaming a mixin and adding interfaces is just the trivial case of a singleton composite mixin.)

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,510 @@
...
+  final String s;
+  const E(this.s);
+}
+
+enum E(String s) { one('a'), two('b') }
+```
+
+This mechanism follows an existing pattern, where `const` modifiers can be
+omitted in the case where the immediaty syntactic context implies that this
+modifier _must_ be present. For example `const [const C()]` can be written
```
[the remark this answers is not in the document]
eernstg: We had a somewhat similar dilemma with `=>` function declarations (not function literals) with declared return type `void`: Some errors are not reported (e.g., `void f() => 3;`) with `=>`, but they _are_ reported with the otherwise equivalent form using a class body (`void f() { return 3; }`). A rationale which has often been mentioned in order to justify this inconsistency is that the return type is right there with `=>` for the reader, but it may be out of sight when the function has a `{...}` body.

So we could use the same rationale to say that we don't have to put `const` just two tokens after `enum`, because everybody can see the `enum` when they can see the `const`; and yet, `const` has to be stated explicitly on body constructors because the `enum` keyword could be out of sight.

I would also be quite happy about allowing `const` to be omitted from body constructors of an enum declaration, based on the assumption that a reader of an enum declaration is likely to be aware of the fact that it is an enum declaration (e.g., because they differ visually from other declarations, due to the typically prominent list of values).

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,510 @@
...
+const Name` rather than `const class Name`.
+
+The modifier `final` on a parameter in a primary constructor has no meaning
+for the parameter itself, because there is no scope where the parameter can
+be accessed. Hence, this modifier is used to specify that the instance
+variables declared by this primary constructor are `final`.
+
+In the case where the constructor is constant, and in the case where the
+declaration is an `inline` class or an `enum` declaration, the modifier
+`final` on every instance variable is required. Hence, it can be omitted:
```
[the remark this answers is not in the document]
eernstg: I think we should resolve that issue in general. When we know what we _want_ to do with formal parameters using `var`, we may introduce that treatment in general, and primary constructors is just one case among others.

(I do _not_ think it's an attractive idea to use `var` to play a special role for the formal parameters of a primary constructor just because "they are about variables as well as parameters, so we can use `var` to mean [something]". Primary constructors are currently using formal parameters that are syntactically and semantically the same as other formal parameters, and I think it would be nice if we can preserve that property.)

The current state is that a formal parameter declaration in an instance method whose first token is `var` yields a parameter type which is determined by override inference, falling back to `dynamic` if there's no overriding relation. In functions that aren't instance methods, it means `dynamic`. (This treatment is not fully specified, as far as I can see, so I'm just reporting the behaviors that I can see via a few experiments).

One obvious way to go is to say that a formal parameter declaration whose first token is `var` is a compile-time error. We could lint/warn against that now, and then make it an error later on, and in that case it seems perfectly fine to make it an error for primary constructor parameters now.

Note that we have at least many hundreds of examples of `var` used as a "formal parameter type" in existing code, which means that we will need to consider some actual breakage.

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,510 @@
...
+modifier _must_ be present. For example `const [const C()]` can be written
+as `const [C()]`. In the examples above, the parameter-and-variable
+declarations `final int x` and `final int y` are written as `int x` and
+`int y`, and this is allowed because it would be a compile-time error to
+omit `final` in an `inline` class, and in a class with a constant
+constructor. In other words, when we see `inline` on the class or `const`
+on the class name, we know that `final` is implied on all instance
+variables.
+
+A proposal which was mentioned during the discussions about primary
```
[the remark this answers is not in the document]
eernstg: Done.

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,510 @@
...
+     'final'? 'inline' 'class' <classNamePart> <interfaces>? <inlineClassBody>;
+
+<inlineClassBody> ::=
+     '{' (<metadata> <inlineMemberDeclaration>)* '}'
+   | ';';
+
+<inlineMemberDeclaration> ::= <classMemberDeclaration>;
+
+<enumType> ::= // Modified rule.
+     'enum' <enumNamePart> <mixins>? <interfaces>? '{'
```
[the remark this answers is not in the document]
eernstg: OK, done.

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,510 @@
...
+```
+
+The word `inline` is now used in the grammar, but it is not a reserved word
+or a built-in identifier.
+
+### Static processing
+
+A class declaration with a primary constructor *(including `inline`, but
+not `<mixinApplicationClass>`, because the grammar rules do not include
+those)* is desugared to a class declaration without a primary
```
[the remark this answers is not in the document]
eernstg: > I really, really want to avoid "desugaring" if possible.

Of course, we can just outlaw that approach in general. However, I don't think it's helpful in this case.

I actually think we want to be able to promise that a simple rewrite operation will show exactly how a primary constructor works. If we insist on using natural language and no desugaring then we need to start entirely from scratch describing how a primary constructor works, and we're going to create the impression that this is a new primitive that provides support for things that we could not express previously. That's just misleading.

The danger associated with desugaring is that it specifies many properties simultaneously. We usually cut down on this by specifying the static analysis of the pre-desugaring syntax separately, and we must then be careful to ensure that the desugaring step transforms every program with no compile-time errors into a program with no compile-time errors. This makes it possible to rely on the dynamic semantics of the desugared program, and it also supports the property that the static analysis can ignore the desugaring step.

> We'll need to say _when_ the rewrite occurs. There is a program before rewrite, and a program after rewrite.

True. I believe the approach that I've outlined above is similar to the approach that we're taking in similar situations, and it has worked so far. For example, shorting `?.` chains is a much, much more complex mechanism, but we haven't specified that in terms of a direct semantics, we've used a rewriting.

Desugaring is a slippery rope in general, but I think we should use it judiciously rather than strictly avoiding it, and I don't think this feature is among the most tricky ones in that respect.

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,510 @@
...
+desugaring step in the first place.*
+
+A compile-time error occurs if a formal parameter that contains a
+term of the form `this.v`, or `super.v` where `v` is an identifier has the
+modifier `covariant`. *For example, `required covariant int this.v` is an
+error.*
+
+Conversely, it is not an error for the modifier `covariant` to occur on
+other formal parameters of a primary constructor. *This extends the
+existing whitelist of places where `covariant` can occur.*
```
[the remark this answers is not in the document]
eernstg: I'd prefer to be able to perform static analysis on the pre-desugared code. Added an error for this.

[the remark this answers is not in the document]
eernstg: Done.

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,510 @@
...
+
+The desugaring consists of the following steps, where _D_ is the class or
+enum declaration in the program that includes a primary constructor, and
+_D2_ is the result of desugaring. The desugaring step will delete elements
+that amount to the primary constructor, and it will add a new constructor
+_k_.
+
+Where no processing is mentioned below, _D2_ is identical to _D_. Changes
+occur as follows:
```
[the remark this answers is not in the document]
eernstg: Indeed, done.

working/2365 - primary constructors/feature-specification.md
eernstg: Maybe. I tried that, but it doesn't actually read very well:

```dart
// Follow the constructor invocations.
class C<X extends List<Y>, ... /*lots of stuff*/>.name(Y y) {...}

// Allow `C.name` to occur together.
class C.name<X extends List<Y>, ... /*lots of stuff*/>(Y y) {...}
```

It is certainly true that the first choice is more consistent with invocations, but I suspect that it matters more that `C.name` should be searchable, as it would be when the declaration is a body constructor.

So I did not change that for now, that sounds like a big dog fight in an issue. ;-)

working/2365 - primary constructors/feature-specification.md
eernstg: Good catch!

I'm now using a rule where the constant variables have a type which is the least closure with respect to such type parameters.

working/2365 - primary constructors/feature-specification.md
eernstg: Improved now!

working/2365 - primary constructors/feature-specification.md
eernstg: Indeed, improved now.

working/2365 - primary constructors/feature-specification.md
eernstg: > We could also allow `extends A.name`

Yes, certainly. I expect that when this document is landed in 'working', discussions about lots of subtopics can start, as usual, and this would be one relevant area to discuss.

working/2365 - primary constructors/feature-specification.md
eernstg: I improved significantly on this part.

working/2365 - primary constructors/feature-specification.md
eernstg: It is intended to be an atomic rewrite in the sense that the pre-desugared program contains _D_, and _D_ is used to obtain various elements, and they are used to create _D2_, and then _D_ is replaced by _D2_ in one step.

eernstg:
Thanks, @lrhn, very useful feedback! I've closed a bunch of threads that I thought were uncontroversial, and left others open.

It is indeed tricky to specify anything in terms of a desugaring step. Thinking about it. ;-)

working/2365 - primary constructors/feature-specification.md
```diff
@@ -0,0 +1,510 @@
...
+
+*In other words, there is no other semantics than the desugaring. When
+desugared, the resulting class declaration is treated exactly the same as
+it would have been if the developer had written the result of the
+desugaring step in the first place.*
+
+A compile-time error occurs if a formal parameter that contains a
+term of the form `this.v`, or `super.v` where `v` is an identifier has the
+modifier `covariant`. *For example, `required covariant int this.v` is an
+error.*
```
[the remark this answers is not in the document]
lrhn: Og destinationer only determines the dynamic semantics, then we need to specify the complete static semantics of the primary constructor syntax. That's still tricky, because it includes complete type inference, but at least for a fairly limited syntax. We're probably able to reuse the type inference of the parameter list from the analysis of other constructors, but we still have to check that any place where we say that "class declares a constructor" or "class declares an instance variable", it also includes a class with a primary constructor. We cannot assume that a declared variable has a particular declaring syntax any more.

So, basically, we have to say, during static analysis, that a class using the primary constructor syntax _declares_ instance variables and a constructor, and specify their declared types, finality, covariance, and the constructor's function type, which is inferred by type inference like other constructors.

Doable. But gotta keep the i's dotted and t's crossed all the way through.

The sooner we move away from the physical syntax, the sooner we can unify different ways to specify the same thing.

working/2365 - primary constructors/feature-specification.md
lrhn: I'm a cat person, so I'll aim for a cat fight. 😸

working/2365 - primary constructors/feature-specification.md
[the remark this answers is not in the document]
eernstg: > considering whether we want to remove `class C = S with M;` from the language [and replacing it]

Maybe we could simply remove that syntax. The fact that it is no longer possible to create exactly the same entity may not be a problem, and developers could use `class C extends S with M;` as a very-very-very-nearly equivalent replacement.

> composite mixins

That would be great!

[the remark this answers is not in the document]
eernstg: In any case, it sounds like there is no big conflict on letting `class C extends S with M;` mean `class C extends S with M {}`.

working/2365 - primary constructors/feature-specification.md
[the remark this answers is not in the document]
eernstg: I'll assume that we handle this question separately, dealing with _all_ parameters using `var` at the same time, or just the ones in primary constructors, but even that can be handled in a discussion issue separately from the first version of the feature spec proposal.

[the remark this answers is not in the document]
eernstg: So I'll resolve the thread.

working/2365 - primary constructors/feature-specification.md
eernstg: I think this has been clarified.

eernstg:
The language team agreed to land this proposal at this time. It is in the 'working' directory, which implies that it is a proposal, not a design which has received a final acceptance.

PR #3027, opened 2023-04-28
Add an error for `on F` where `F` is final and outside the current library

eernstg:
This PR adds an error to the class modifiers feature spec for the case where an `on` type is `final` and from a different library.

### DIFF
--- a/accepted/future-releases/class-modifiers/feature-specification.md
+++ b/accepted/future-releases/class-modifiers/feature-specification.md
@@ -715,12 +715,12 @@ It's a compile-time error if:
     from another library _(with some exceptions for platform libraries)_.
 
     _(You cannot inherit implementation from a class marked `interface`
-    or `final` except inside the same library. Unless you are in a
-    pre-feature library and you are inheriting from a platform library.)_
+    except inside the same library. Unless you are in a pre-feature
+    library and you are inheriting from a platform library.)_
 
     More formally:
-    A declaration *C* from library *L* has a declared superclass or mixin
-    declaration *S* marked `interface` or `final` from library *K*, and neither
+    A declaration *C* from library *L* has a declared superclass
+    declaration *S* marked `interface` from library *K*, and neither
     * *L* and *K* is the same library, nor
     * *K* is a platform library and *L* is a pre-feature library.
 
@@ -735,8 +735,8 @@ It's a compile-time error if:
     ```
 
 *   A declaration implements another declaration, and the other
-    declaration itself, or any of its super-declarations,
-    are marked `base` and are not from the first declaration's
+    declaration itself, or any of its super-declarations, are marked
+    `base` or `final` and are not from the first declaration's
     library _(with some exceptions for platform libraries)_.
 
     _(You can only implement an interface if *all* `base` or `final`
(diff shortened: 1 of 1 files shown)


accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -687,6 +687,26 @@ It's a compile-time error if:
     class M with S {} // Error, for several reasons.
     ```
 
+*   A mixin has a declared `on` type from another library which is marked
```
lrhn, line 690: If we do need an extra case for `final`, I'd go all in and declare all the errors here.

> *   A declaration has a direct super declaration from another library, 
>      which is marked `final`.
>
>      More formally:
>      A declaration *D* from library *L* has a direct superdeclaration *S* 
>      marked `final` (so necessarily a `class` declaration) in library *K*, and neither
>      * *L* and *K* is the same library, nor
>      * *K* is a platform library and *L* is a pre-feature library.
> 
>      ```dart
>     // a.dart 
>     final class F {}
>
>     // b.dart    
>     import 'a.dart';
>     
>     class C1 extends F {}  // Error.
>     class C2 implements F {}  // Error.
>     mixin class C3 implements F {}  // Error.
>     mixin M1 implements F {}  // Error.
>     enum E1 implements F {}  // Error.
> ```

Then remove `final` from the direct `interface` superdeclaration cases below.

I think it could be cleaner that way.

(The transitive `base`/`final` restriction case is still needed, although even there I think the `final` wouldn't be needed, since any transitive chain to a final class in another library, which doesn't have an otherwise invalid direct step, will also contain at least one `base` declaration in the same library as the `final` library.)

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -687,6 +687,26 @@ It's a compile-time error if:
...
+    ```dart
+    // a.dart
+    final class F {}
+
+    // b.dart
+    import 'a.dart';
+
+    mixin M on F {}  // Error.
+    ```
+
```
lrhn, line 709: In line 719 below: Mixins can't be `interface` or `final` any more, so "or mixin" isn't needed.

accepted/future-releases/class-modifiers/feature-specification.md
eernstg: Yes, that's cleaner. Did that.

accepted/future-releases/class-modifiers/feature-specification.md
eernstg: Done.

eernstg:
Review response uploaded.

accepted/future-releases/class-modifiers/feature-specification.md
```diff
@@ -687,34 +687,56 @@ It's a compile-time error if:
     class M with S {} // Error, for several reasons.
     ```
 
-*   A class extends or mixes in a declaration marked `interface` or `final`
+*   A declaration has a direct super declaration from another library
+    which is marked `final`.
```
lrhn, line 691: Should we add the `_(with some exceptions for platform libraries)_` here too?

lrhn (approved):

accepted/future-releases/class-modifiers/feature-specification.md
eernstg: Of course, done.

PR #3033, opened 2023-05-01
Prohibit variable and identifier patterns from being named `when` or `as`, to simplify parsing.

stereotype441:
This spec change reflects the discussion at https://github.com/dart-lang/sdk/issues/52199#issuecomment-1526297771.

Test cases are here: https://dart-review.googlesource.com/c/sdk/+/300320

### DIFF
--- a/accepted/future-releases/0546-patterns/feature-specification.md
+++ b/accepted/future-releases/0546-patterns/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: Accepted
 
-Version 2.31 (see [CHANGELOG](#CHANGELOG) at end)
+Version 2.32 (see [CHANGELOG](#CHANGELOG) at end)
 
 Note: This proposal is broken into a couple of separate documents. See also
 [records][] and [exhaustiveness][].
@@ -512,6 +512,10 @@ switch (record) {
 }
 ```
 
+To simplify parsing, the identifier in a variable pattern may not be `when` or
+`as`. *This reduces the amount of lookahead needed to correctly parse a guarded
+pattern such as `foo when !bar`.*
+
 *There are some restrictions on when `var` and `final` can and can't be used.
 They are specified later in the "Pattern context" section.*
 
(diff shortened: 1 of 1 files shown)


stereotype441:
@munificent I decided not to change the behaviour of `when` and `as` in constant patterns that are qualified names, or the behaviour of `when` in object patterns. Does that seem reasonable to you?

munificent:
> Implementation difficulty aside, I'm also not convinced it would be doing a service to users. If I'm unlucky enough to be the consumer of an API that contains a getter named `when`, I might have a legitimate reason to write an object pattern like `Foo(when: var x?)`. Or if I'm switching on an enum and one of the enum values is named `when`, I might still want to be able to write `case Foo.when:`. And I might not really be in a good position to rename the code I depend on, so my only choice at that point is, don't use patterns. It doesn't seem like it would benefit anyone.

Good point! OK, I'm sold on the approach in this PR.

> That's a really good question! I don't know how best to handle this.

cc @dart-lang/dart-team 

Maybe the larger team has thoughts.

munificent (approved):

leafpetersen:
> Maybe the larger team has thoughts.

For the scale of change we seem to be settling on, I'd just suggest landing it as a change to the existing spec (i.e. this PR).

leafpetersen:
> On a related note:
> 
> * Do you think this change should be language versioned? (My preference is no, since I think breakages will be minimal)

No.  

> * Do you think we should go through the breaking change process for this? (Again my preference is no, for the same reasons)

No.

> * If I can get this change done quickly, do you think it would be worth trying to cherry-pick it into the stable release?

Yes.

dnfield:
How does this change affect https://pub.dev/documentation/mockito/latest/mockito/when.html?

stereotype441:
> How does this change affect https://pub.dev/documentation/mockito/latest/mockito/when.html?

With the narrower scope we agreed on in https://github.com/dart-lang/language/pull/3033#issuecomment-1533731710, there's no effect on https://pub.dev/documentation/mockito/latest/mockito/when.html. The only things being prohibited here are:
The ability to declare a local variable named `when` through the use of a pattern, e.g.:
```dart
switch (...) {
  case var when:
    ...
}
```
The ability to assign to a local variable named `when` through the use of a pattern, e.g.:
```dart
String who, what, where, when, why, how; // This is ok
(who, what, where, when, why, how) = getQuestions(); // This isn't
```
And the ability to refer to lone (unprefixed) constants named `when` in a pattern or a switch, e.g.:
```dart
const when = 1;
switch (...) {
  case when:
    ...
}
```
None of which should interfere with the use of `when` in mockito.

dnfield:
Thanks for the explanation!

PR #3040, opened 2023-05-04
Specify implicit construction

eernstg:
This PR adds a proposal to the working directory about implicit constructors, provided as members of static extensions.

### DIFF
--- a/working/0107 - implicit-constructors/feature-specification.md
+++ b/working/0107 - implicit-constructors/feature-specification.md
@@ -190,7 +190,7 @@ of `E` is `C`, and the _constructor return type_ of `E` is `C<T1 .. Tk>`.
 In both cases, `E` is an identifer `id` which is optionally followed by a
 term derived from `<typeParameters>`. We say that the identifier `id` is
 the _name_ of the static extension. *Note that `T1 .. Tk` above may contain
-occurrences of those type variables.*
+occurrences of those type parameters.*
 
 ### Static Analysis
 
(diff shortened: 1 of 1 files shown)


eernstg:
The proposal should be complete enough to enable some discussions at this point.

sigmundch:
Thanks so much for putting this together @eernstg. There are many elements of this proposal I really like. I agree it's important to limit the scope of the implicit constructors. I think requiring a `show` may actually be an interesting requirement, or I'd potentially go even further and require that developers state what is the implicit conversion itself (showing an extension may not really hint to someone reading the code what's the on-class or what's the constructor that is in use).

For example:
```dart
import 'foo.dart'; 
   show E implicit constructor Distance Function(int);
...
```

I know my suggestion is very verbose, but just sharing the concept here :smile: 

We could also go further and limit conversions to different scope levels (library, class, or even method):

```dart
@ImplicitConversions([Distance Function(int)]) // conversion allowed in this library
import '...';

...
@ImplicitConversions([Distance Function(double)]) // conversion only allowed in this class
class A {
  ...
  @ImplicitConversions([Distance Function(String)]) // conversion only allowed in this method
  m1() {
  }
}
```

Thoughts?


On a separate note, this last week I came across interesting cases with `==` and `!=` operations. Technically `==` operators are a potential assignment point, but, because `==` is defined as taking an `Object` argument we wouldn't be able to recognize it as a place to use an implicit conversion. Here is a concrete example:

```dart
int height = 0;
if (height == api.height /*returns JSNumber */) {
 ...
}
```

What are your thoughts on possible ways in which `==` could also be considered a trigger for conversions?

working/0107 - implicit-constructors/feature-specification.md
```diff
@@ -0,0 +1,603 @@
...
+can contain constructors.*
+
+A compile-time error occurs if a static extension has an on-clause of the
+form `on C<T1 .. Tk>`, and the actual type parameters passed to `C` do not
+satisfy the bounds declared by `C`.  The static extension may declare one
+or more type parameters `X1 extends B1 .. Xs extends Bs`, and these type
+variables may occur in the types `T1 .. Tk`. During the bounds check, the
+bounds `B1 .. Bs` are assumed for the type variables `X1 .. Xs`.
+
+A compile-time error occurs if a static extension _D_ declares a
```
leafpetersen, line 221: This makes it a breaking change to add a constructor to a class.  Probably worth at least calling this out for consideration?  It might be better to do what we do with extension methods and just say that the class version takes precedence (or the other way around, I suppose).

working/0107 - implicit-constructors/feature-specification.md
```diff
@@ -0,0 +1,603 @@
...
+parameter scope whose enclosing scope is the library scope. The current
+scope for the on-clause of _D_ is the type parameter scope.
+
+A static extension _D_ introduces a body scope, which is the current scope
+for the member declarations. The enclosing scope for the body scope is the
+type parameter scope, if any, and otherwise the library scope.
+
+Static members in a static extension are subject to the same static
+analysis as static members in other declarations. There is one extra rule:
+It is a compile-time error if a static member of a static extension has a
```
leafpetersen, line 275: As above, this makes it a breaking change to add a static member to a class.

eernstg:
@sigmundch wrote:
> require that developers state what is the implicit conversion itself

Interesting idea! We could _allow_ implicit conversions to be enabled explicitly, and then have a lint catching the cases where it isn't used (a la "you're using this-and-that implicit constructor to convert [a specific expression], but that constructor hasn't been explicitly enabled").

```dart

// Library 'client.dart'.
import 'js_interop_static_extensions.dart' enable DoubleExtensions.fromDouble;

JSNumber jsNumber = ...;
double d = jsNumber; // Uses implicit conversion.
```

We could also rely on the function type (here: `double Function(JSNumber)`), but I think it's a safer bet to enable each implicit constructor by name: There could be multiple implicit constructors with the same function type, or there could be subtypes or supertypes of that function type. If the name is mentioned explicitly then it should be rather easy to find the corresponding function type, at least in an IDE, by hovering or by going-to-declaration.

This would again be a mechanism that we can leave out at first, and it could then be introduced later, and developers could make the choice about whether/when/where they'd enable the lint. For instance, there could be a library where certain conversions are used all the time, and `enable Some, constructors` will be a significant convenience ( `// ignore_for_file: implicit_conversion_not_explicitly_enabled`), and there could be a different library where those conversions are the rare exception, and the maintainers prefer to write the constructor invocation explicitly.

> What are your thoughts on possible ways in which `==` could also be considered a trigger for conversions?

Maybe we should think about the underlying question: Is it useful to include the operands of `==` among the assignment positions in code? It sounds like it could be a prolific source of ambiguities. ;-)

eernstg (commented):
Review response (in this particular case I actually did not change the PR, but there's a discussion in https://github.com/dart-lang/language/issues/3095).

working/0107 - implicit-constructors/feature-specification.md
eernstg: It is no more a breaking change.

working/0107 - implicit-constructors/feature-specification.md
eernstg: It is no more a breaking change.

eernstg:
Following the approach used with the [primary constructor proposal](https://github.com/dart-lang/language/pull/3023), I'll land this document in 'working' at this time.

Issue #2325, opened 2022-06-29
Define the semantics of `augment super` for operators


jakemac53:
From @johnniwinther: "What does augment super[0] mean? Is it the call to the operator [] that is currently being augmented (like in augment int operator[](int index) => augment super[index];) or a call to operator [] on the value of the currently augmented getter (like in augment List<int> get property => [augment super[0]];, or both depending on context?"

I would say that inside of an augmented `operator []` method it calls the `operator []` method that is being augmented. In all other cases it would only be valid if augmenting a getter, and it would call `operator []` on the return value of the augmented getter. This does make it context dependent, but also I think makes it read most intuitively.

I also think combining `augment super` into a single keyword like `augmented` would help readability here.

jakemac53:
My 2 cents - it calls the index operator `[]` on the currently augmented declaration.

In other words `augment super` gives a reference to the augmented declaration, and you can then invoke any methods you like on that (including `[]`).

Essentially `augment super` is always wrapped in parenthesis :).

jakemac53:
This might be more obvious if we had a single reserved word which always referred to the augmented declaration - it mostly looks weird because it is two separate words but is really just a reference to a single thing.

lrhn:
I'm very much not sold on the `augment super` syntax. We should find something better, and more ... readable.

(What's better than `super`? Maybe `ultra`?!)

scheglov:
`augmented`

jakemac53:
I do personally like `augmented`, the push back on that has been not wanting to create more reserved words, but this would only be reserved in the scope of an augmentation and might be ok imo.

munificent:
I definitely don't love `augment super`. I'm also not particularly fond of `import augment` or `library augment`. I do think `augment` as a modifier on declarations works fairly well.

This was a part of the proposal that I sort of made up a semi-placeholder syntax for and never really revisited since I was mostly focused on the semantics. I picked `augment super` at the time because the syntax was fairly close to `super` calls. But that's no longer the case as I understand it.

With `super`, you can invoke any member from the superclass from any member:

```dart
class C extends B {
  void foo() {
    super.bar();
    //    ^^^ Not foo().
  }
}
```

But we don't support that for augmentations (for good reason). So there's no `.` or identifier after the `super` in `augment super`. So we're not getting much user value out of reusing `super` for this. It suggests a syntax that doesn't actually work.

[...]

Issue #2324, opened 2022-06-29
Can you tear off `super augment` members?


jakemac53:
For example, is something like `var m = super augment;` allowed? Is it a problem to leak these augmented members?

My sense is that we should allow it, and there isn't an immediate issue.

munificent:
I *suspect* that we won't want to allow these to be torn off. It might work in all circumstances, but I would be surprised if we didn't run into corners where it becomes tricky to materialize the augmented member and allow it to be invoked later.

For example, tearing off a reference to an augmented generative constructor means you now have a function that when invoked runs an initializer list that can assign to final fields. What happens if you call that multiple times? (Now that I think about it, I suppose we have that same question even if you don't tear it off and just immediately call it multiple times.)

I'm not aware of a use case for tearing off augmented member references and they are a pretty weird concept, so my inclination is to not support it for now unless we discover a compelling use case for it.

jakemac53:
I agree I don't have a particular use case in mind, but I also can't think of a reason to not allow it? It seems like it should probably be allowed just to make things complete, there isn't anything special about those functions that I can think of?

munificent:
>  there isn't anything special about those functions that I can think of?

They are definitely kind of special when it comes to generative constructors. Whether other ones are special... probably not? But if it turns out that it makes things easier on the implementation teams to implement `augment super` calls in a special way that would make tear-offs harder, I wouldn't want to force them to implement a more complex feature that we don't know is useful.

scheglov:
From the analyzer POV I don't see any difference between implementing invocation `super augment foo()` and tear-off `super augment foo`.

munificent:
> > They are definitely kind of special when it comes to generative constructors.
> 
> You can take a tearoff of those today right?

You can tear off a generative constructor, yes. When you invoke the resulting closure, you get a brand new instance that goes through the entire construction process each time, so that makes sense.

It's weirder when you tear-off the `augment super` call for a generative constructor because the instance has already been created. Though now that I go back and re-read the proposal I actually wrote... maybe it's not so bad? The constructor initializers have already been run for both constructors and can't be re-run. The only thing the `augment super` call does is run the wrapped constructor's *body*. So even before you call it, the instance is in a well-defined state.

Yeah, I guess it's OK then.

jakemac53:
I believe this also helps to clarify the answer to https://github.com/dart-lang/language/issues/2325

lrhn:
[...]

PR #3045, opened 2023-05-05
Add clarifying section on the semantics of `super augment` in various augmenting contexts

jakemac53:
Closes https://github.com/dart-lang/language/issues/2325
Closes https://github.com/dart-lang/language/issues/2324

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -285,17 +285,23 @@ It is a compile-time error if:
 The exact result of an `augment super` expression depends on what is being
 augmented, but it follows generally the same rules as any normal identifier:
 
-*   **Augmenting getters**: When augmenting a getter, `augment super` invokes
-    the getter and evaluates to the return value.
+*   **Augmenting getters**: Within an augmenting getter `augment super` invokes
+    the getter and evaluates to the return value. If augmenting a field with a
+    getter, this will invoke the implicit getter from the augmented field.
 
-*   **Augmenting setters**: When augmenting a setter, it can only be used to
-    directly invoke that setter, such as `augment super = <expression>;`.
+*   **Augmenting setters**: Within an augmenting setter `augment super` must be
+    followed by an `=` and will directly invoke the augmented setter. If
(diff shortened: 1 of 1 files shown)


working/augmentation-libraries/feature-specification.md
```diff
@@ -278,6 +280,38 @@ It is a compile-time error if:
...
+    the getter and evaluates to the return value. If augmenting a field with a
+    getter, this will invoke the implicit getter from the augmented field.
+
+*   **Augmenting setters**: Within an augmenting setter `augment super` must be
+    followed by an `=` and will directly invoke the augmented setter. If
+    augmenting a field with a setter, this will invoke the implicit setter from
+    the augmented field.
+
+*   **Augmenting fields**: Within an augmenting field, `augment super` can only
+    be used in an initializer expression, and refers to the original fields
```
munificent, line 298: Nit: "fields" -> "field's".

working/augmentation-libraries/feature-specification.md
```diff
@@ -278,6 +280,38 @@ It is a compile-time error if:
...
+    initializer expression, which is immediately evaluated. If there is no
+    initializer expression (or augmenting a getter and/or setter), then
+    `augment super` is not allowed.
```
munificent, line 301: Instead of "is not allowed", I would be more precise:

> It is a compile-time error to use `augment super` in an augmenting field's initializer if the member being augmented is not a field with an initializer.

Also, we might need to be clearer about how this works with chained augmentations. If you:

1. A field declaration.
2. Augmenting that with a getter.
3. Augmenting that with a field.

Can the initializer in 3 use `augment super` to invoke the initializer from 1?

munificent (approved):

working/augmentation-libraries/feature-specification.md
jakemac53: > Can the initializer in 3 use `augment super` to invoke the initializer from 1?

Haha, nothing is ever easy is it?

jakemac53: I added a TODO here, this brings up a whole can of worms issue.

PR #3085, opened 2023-05-18
blast_repo fixes

devoncarew:
dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,3 +7,5 @@ updates:
     directory: /
     schedule:
       interval: monthly
+    labels:
+      - autosubmit


.github/dependabot.yml
```diff
@@ -7,3 +7,5 @@ updates:
     directory: /
     schedule:
       interval: monthly
+    labels:
+      - autosubmit
```
munificent, line 11: What does this do?

devoncarew: It's part of a new service from a flutter engprod github bot; we can opt some PRs into being autolanded (by the `fluttergithubbot`) if they pass all their regular checks and have the `autosubmit` label applied. This will save us from having to click on ~100 PRs per month.

munificent: Makes sense, but I also somewhat question the value of dependabot constantly rolling pinned dependencies. If the plan is to automatically roll them, and also automatically review the changes that roll them, and automatically land those changes... why are the dependencies being pinned in the first place?

munificent (approved):

Issue #3041, opened 2023-05-04
Define the semantics when augmenting external declarations


jakemac53:
Can an augmentation provide an implementation for an `external` declaration, and does that count as a valid implementation? Or do we still require some outside system to fill in the implementation, and then any augmentations are actually augmenting that?

Can you call `augment super` when augmenting an external declaration? Only if something externally is actually providing an implementation?

Are both external sources and augmentations allowed to provide an implementation of an external declaration, and which one wins?

cc @munificent

jakemac53:
Note that we do currently have this error specified:

```
The original function is declared `external` and the augmenting function  uses `augment super()`.
```

But I am not necessarily convinced that is the right answer.

jakemac53:
cc @johnniwinther when it comes to using augmentations for the core libraries, I presume this may be important.

lrhn:
External declarations should work like any other, fully declared, declaration. There is no reason they should work differently wrt. augmentation.
The only thing special about external declarations is that the implementation is in the compiler instead of in the library source code. It's still just as much there. The `external` is like a function body, it really should have been at the end instead of the start (`void foo(int x) external;` would be more consistent with how we place implementation modifiers, like `async`).

The way it's currently specified, you cannot write `augment external void foo();`, and I think you *should* be able to do that.
Say `@JS("something') augment external void foo();`.

If the goal is to allow an `external` declaration with no compiler-supplied implementation to be *implemented* by an `augment` declaration of the same name, then that's fine too. Valid goal, close to how patch files work, but then it should be phrased as an error to have an `augment super` to an `external` declaration *which has no implementation*.
(I think the specification makes calling un-implemented externals into runtime errors. I think they should be compile-time errors in at least some cases.)

jakemac53:
> External declarations should work like any other, fully declared, declaration. There is no reason they should work differently wrt. augmentation.

We do have to specify some stuff about how they are handled wrt. augmentations though. There are a few possible interpretations, and some may slightly change what `external` actually _means_ (while others would not).

> If the goal is to allow an `external` declaration with no compiler-supplied implementation to be _implemented_ by an `augment` declaration of the same name, then that's fine too.

I am not sure if it should be a goal or not - I could see arguments either way :). But if we do allow it, then I do think it makes it a bit complicated, because we need to be able to distinguish between the two as that influences whether or not `augment super` should be allowed.

> The way it's currently specified, you cannot write `augment external void foo();`, and I think you _should_ be able to do that.
[...]

PR #3086, opened 2023-05-18
specify some semantics around augmenting external members

jakemac53:
Closes https://github.com/dart-lang/language/issues/3041

This reverses the previous behavior for `external`, so that `augment super` is always allowed but may throw at runtime.

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -267,10 +267,8 @@ introduction of this new identifier will be language versioned in order to make
 it non-breaking for old code.
 
 It is also allowed for a non-abstract class to have abstract members, if those
-member bodies are filled in by an augmentation. You should not explicitly mark
-those members as `abstract` or `external`, you can simply leave off the body to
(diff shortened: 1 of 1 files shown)


working/augmentation-libraries/feature-specification.md
```diff
@@ -266,6 +266,12 @@ syntactically mark a declaration as an augmentation of an existing one. The
 introduction of this new identifier will be language versioned in order to make
 it non-breaking for old code.
 
+It is also allowed for a non-abstract class to have abstract members, if those
+member bodies are filled in by an augmentation. You should not explicitly mark
+those members as `abstract` or `external`, you can simply leave off the body to
+make them implicitly abstract. This is primarily useful for macros, which may
+be used to provide a body for an abstract member.
```
munificent, line 273: Do we allow abstract field declarations? It was a later syntactic addition to Dart. It looks like:

```dart
abstract class Foo {
  abstract int bar;
}
```

If so, and I think we do, you'll want to reword the paragraph a bit. Maybe just omit the second sentence.

munificent (approved):

Issue #3090, opened 2023-05-20
[inline-classes] Allow an inline class to implement a non-inline class


eernstg:
This issue describes a proposal to allow an inline class to implement a non-inline class, as long as it is a supertype of its representation type.

As it is currently [specified](https://github.com/dart-lang/language/blob/main/accepted/future-releases/inline-classes/feature-specification.md), the `implements` clause of an inline class declaration is used to establish a subtype relationship to other inline classes. This allows the given inline class to "inherit" member implementations from those other inline classes, and it establishes a subtype relationship.

```dart
inline class A(int i) {
  int get next => i + 1;
}

inline class B(int i) implements A {}

var b = B(1).next; // OK.
int i = B(2); // Error: unrelated types.
```

Another subtype relationship which can safely be adopted is that the inline class is a subtype of its representation type, or any supertype of the representation type. We could broaden the applicability of the `implements` clause to establish that kind of subtype relationship as well:

```dart
inline class A(int i) implements int {} // `A` is now a subtype of `int`.

var a = A(2);
int i = a; // OK.
List<int> list = <A>[a]; // OK.
```

(I'm using primary constructors for brevity, assuming something like https://github.com/dart-lang/language/pull/3023)

This is sound because the run-time value of an expression of type `A` is an instance of `int`. It may be useful in the case where there is no desire to protect  an object accessed as an `A` against being accessed as an `int`, and it is even considered to be better that it will readily "become an `int` whenever needed" than not.

The corresponding change to the feature specification could be that the following sentence in the section [Composing inline classes](https://github.com/dart-lang/language/blob/main/accepted/future-releases/inline-classes/feature-specification.md#composing-inline-classes) is adjusted:

> A compile-time error occurs if _V1_ is a type name or a parameterized type which occurs as a superinterface in an inline class declaration _DV_, but _V1_ does not denote an inline type.

It is adjusted to end as follows:

> ... declaration _DV_, unless _V1_ denotes an inline type, or _V1_ is a supertype of the representation type of _DV_.

Moreover:

> A compile-time error occurs if the implements clause of _DV_ contains two or more types that are non-inline types.

We have had discussions about how to handle invocations of members of such non-inline superinterfaces. Here is one of the options that we had on the table:

> If _DV_ has a non-inline superinterface `S` then invocations of members of `S` that are not redeclared in _DV_ are
> performed as if the receiver had had the representation type of _DV_.

[...]

PR #3138, opened 2023-06-08
Feature specification of #3090, allowing an inline class to implement its representation type

eernstg:
This PR contains a proposal for an addendum to the feature specification of inline classes, specifying how we could support the situation where an inline class `V` has an `implements ... R` clause where `R` is the representation type (or any supertype thereof).

The point is that this provides (1) the subtype relationship `V <: R`, and (2) a receiver of type `V` can have members of `R` invoked on it.

In this proposal, `V` is allowed to redeclare members of `R` (which means that `V` "overrides" the implementation of that member, albeit with static dispatch only).

Conflict resolution is specified to be the same as it has always been for superinterfaces: If both `R` and some other superinterface have a member with the same name then there is an error, but `V` can redeclare the member and that will resolve the conflict.

### DIFF
--- a/working/1426-extension-types/feature-specification-implements-rep.md
+++ b/working/1426-extension-types/feature-specification-implements-rep.md
@@ -197,7 +197,7 @@ same type based on alpha equivalence (renaming of type variables).*
 
 A similar rule applies for the case where _DV_ has two superinterfaces,
 direct or indirect, which are function types or record types *(in which
-hase the types _must_ be non-inline types)*: They must also be the same
+case the types _must_ be non-inline types)*: They must also be the same
 type.
 
 Let _DV_ be an inline class declaration named `V` with representation type
@@ -208,11 +208,11 @@ else is an error.*
 Assume that `m` is a member which not declared by _DV_, and none of _DV_'s
 inline superinterfaces have a member named `m`, but one or more of the
 interfaces of `R1 .. Rk` has a member named `m`. A compile-time error
-occurs if there exist `j1` and `j2` in `1 .. k` and a member name `m` such
-that `m` does not have a combined member signature for `R1 .. Rk`.
-Otherwise the member signature of `m` is that combined member signature.
-In this situation we say that this is the member signature of `m` in
-`R1 .. Rk`.
+occurs if there exists a member name `m` such that at least two of `R1 .. Rk`
+have a member with that name, but `m` does not have a combined member
+signature for `R1 .. Rk`.  Otherwise the member signature of `m` is that
+combined member signature.  In this situation we say that this is the
+member signature of `m` in `R1 .. Rk`.
 
 Invocations of members declared by _DV_ or declared by an inline
 superinterface of _DV_ and not declared by any of `Rj`, `j` in `1..k` are
--- a/working/1426-extension-types/feature-specification-implements-rep.md
+++ b/working/1426-extension-types/feature-specification-implements-rep.md
@@ -64,7 +64,7 @@ List<int> list = <A>[a]; // OK.
 ```
 
 This subtype relationship is sound because the run-time value of an
-expression of type `A` is an instance of `int`. 
+expression of type `A` is an instance of `int`.
 
 This subtype relationship may be desirable in the case where there is no
 need to protect an object accessed as an `A` against being accessed as an
@@ -156,8 +156,8 @@ inline class MapEntry<K, V> implements (K, V) {
 }
 ```
 
-It is not a compile-time error for an inline class to have a non-inline
-class `C` as a superinterface based on the class modifiers of `C`.
+It is not a compile-time error for an inline class declaration to have a
+non-inline class `C` as a superinterface based on the class modifiers of `C`.
 
 *For instance, `inline class V implements B ...` does not give rise to a
 compile-time error because `B` is a class declared in a different library
@@ -166,6 +166,53 @@ is that the type `V` is a static device that allows us to work with an
 instance of `B` in a convenient way, it is not a mechanism that allows
 anything like "`V` is a subtype of `B`" to be a fact at run time (because
 `V` doesn't even exist at run time).*
+
+A compile-time error occurs if an inline class declaration _DV_ has two
+direct superinterfaces of the form `V1<T1 .. Tn>` and `V2<S1 .. Sm>` such
+that `V1` and `V2` resolve to the same declaration.
+
+*This rule is similar to a rule for non-inline classes that makes
+`class C implements B, B {}` an error.*
+
+Assume that an inline class declaration _DV_ has two superinterfaces,
+direct or indirect, of the form `V1<T1 .. Ts>` and `V2<S1 .. Ss>`. Assume
+that `V1` and `V2` both denote the same inline class declaration. A
+compile-time error occurs if `Tj` and `Sj` are not the same type, for any
+`j` in `1 .. s`. This rule applies also in the case where one or both of
+the two types is a raw type, and the actual type arguments have been
+computed by instantiation to bounds, or by any other implicit mechanism.
+
+In this rule 'same type' is defined as in the section
+[Runtime type equality operator][] in the null safety specification.
+
+[Runtime type equality operator]: https://github.com/dart-lang/language/blob/main/accepted/2.12/nnbd/feature-specification.md#runtime-type-equality-operator
+
+*This means that `Tj` and `Sj` are the same type if the normalized form of
+both types are structurally identical up to names of type variables. For
+instance, `prefix.C<int>` and `C<int>` are the same type if `prefix.C` and
+`C` resolve to the same class declaration, and similarly for `int`;
+`FutureOr<Object>` is the same type as `Object` because of the
+normalization; and `void Function<X>(X)` and `void Function<Y>(Y)` are the
+same type based on alpha equivalence (renaming of type variables).*
+
+Assume that an inline class declaration _DV_ has two superinterfaces,
+direct or indirect, of the form `V1<T1 .. Ts>` and `V2<S1 .. Ss>`. Assume
+that `V1` and `V2` both denote the same non-inline class declaration. A
+compile-time error occurs unless one of these types is a subtype of the
+other.
+
+*It follows that when _DV_ implements more than two such non-inline types,
+it is an error unless one of them is at least as specific as all the
+others.*
+
+Assume that an inline class declaration _DV_ has two superinterfaces,
+of the form `V1<T1 .. Ts>` and `V2<S1 .. Ss>`, where the former is a direct
+superinterface and the latter is indirect. Assume that `V1` and `V2` both
+denote the same non-inline class declaration. A compile-time error occurs
+if the former is not a subtype of the latter.
+
+*In other words, a more special inline type can implement a more special
+non-inline type, not a less special one or an unrelated one.*
 
 Let _DV_ be an inline class declaration named `V` with representation type
 `R` and assume that the `implements` clause of _DV_ includes the non-inline
@@ -178,6 +225,8 @@ interfaces of `R1 .. Rk` has a member named `m`. A compile-time error
 occurs if there exist `j1` and `j2` in `1 .. k` and a member name `m` such
 that `m` does not have a combined member signature for `R1 .. Rk`.
 Otherwise the member signature of `m` is that combined member signature.
+In this situation we say that this is the member signature of `m` in
+`R1 .. Rk`.
 
 Invocations of members declared by _DV_ or declared by an inline
 superinterface of _DV_ and not declared by any of `Rj`, `j` in `1..k` are
@@ -192,10 +241,9 @@ inline superinterface of _DV_ also declares a member named `m`.
 
 Let `m` be a member name which is not declared by _DV_ and not declared by
 an inline superinterface of _DV_. Assume that the interface of `Rj` has a
-member named `m` with signature `s` *(this is the combined member signature
-that may depend on other types in `R1 .. Rk`)*. An invocation of `m` on a
-receiver of type `V` (or `V<T1 .. Ts>` if _DV_ is generic) is then treated
-as the same invocation, but with signature `s`.
+member named `m` with signature `s` in `R1 .. Rk`. An invocation of `m` on
+a receiver of type `V` (or `V<T1 .. Ts>` if _DV_ is generic) is then
+treated as the same invocation, but with signature `s`.
 
 *It is already specified in the inline class feature specification to be an
 error if two inline superinterfaces `V1, V2` of _DV_ both declare a member
--- a/working/1426-extension-types/feature-specification-implements-rep.md
+++ b/working/1426-extension-types/feature-specification-implements-rep.md
@@ -17,10 +17,11 @@ of type `V`.
 ## Summary
 
 This is a proposal to add a new feature to inline classes: An inline class `V`
-may include a superinterface in its `implements` clause which is a
-supertype `R0` of the ultimate representation type `R` of `V` (this allows
-`implements R` as a special case). This causes `V` to be a subtype of `R0`,
-and it enables invocations of members of `R0` on a receiver of type `V`.
+may include one or more superinterfaces in its `implements` clause which are
+supertypes `R1 .. Rk` of the ultimate representation type `R` of `V` (this
+allows `implements R` as a special case). This causes `V` to be a subtype
+of each of `Rj`, `j` in `1 .. k`, and it enables invocations of members of
+`Rj` on a receiver of type `V`.
 
 As it is currently
 [specified](https://github.com/dart-lang/language/blob/main/accepted/future-releases/inline-classes/feature-specification.md),
@@ -109,7 +110,7 @@ inline class A {
   A(this.n);
 }
 
-inline class B implements num {
+inline class B implements num { // OK.
   final A a;
   A(this.a);
 }
@@ -118,35 +119,40 @@ inline class B implements num {
 *This is allowed because the representation object for an expression of
 type `B` is an object of type `num`, because it is the representation
 object of an expression of type `A`. Note that there is no need for (or any
-problem with) a subtype relationship between `A` and `B`.*
-
-Moreover, a compile-time error occurs if the implements clause of _DV_
-contains two or more types that are non-inline types. A compile-time error
-occurs if the implements clause of _DV_ contains a non-inline type that
-occurs in any other position than the last one.
-
-*This ensures that it is only necessary for a reader of the declaration to
-check the last implemented type in order to see whether or not it is using
-this feature. It is somewhat similar to the rule that `super(...)` can only
-occur as the last element in a constructor initializer list.*
+problem with) a subtype relationship between `A` and `B`. The relationship
+between an inline type and its instantiated representation type and the
+subtype relationship for an inline type are independent concepts.*
 
 Let _DV_ be an inline class declaration named `V` with representation type
 `R` and assume that the `implements` clause of _DV_ includes the non-inline
-type `R0`. *We then have `R <: R0`, because anything else is an error.*
+types `R1 .. Rk`. *We then have `R <: Rj` for each `j`, because anything
+else is an error.*
+
+Assume that `m` is a member which not declared by _DV_, and none of _DV_'s
+inline superinterfaces have a member named `m`, but one or more of the
+interfaces of `R1 .. Rk` has a member named `m`. A compile-time error
+occurs if there exist `j1` and `j2` in `1 .. k` and a member name `m` such
+that `m` does not have a combined member signature for `R1
+.. Rk`. Otherwise the member signature of `m` is that combined member
+signature.
 
 Invocations of members declared by _DV_ or declared by an inline
-superinterface of _DV_ and not declared by `R0` are unaffected by the fact
-that _DV_ implements `R0`.
+superinterface of _DV_ and not declared by any of `Rj`, `j` in `1..k` are
+unaffected by the fact that _DV_ implements `R1 .. Rk`.
+
+*This could be an invocation of a member declared by _DV_, or by any of its
+non-inline superinterfaces, or both, but the rules are unchanged.*
 
 Let `m` be a member name which is not declared by _DV_. Assume that the
-interface of `R0` has a member named `m`. A compile-time error occurs if an
+interface of `Rj` has a member named `m`. A compile-time error occurs if an
 inline superinterface of _DV_ also declares a member named `m`.
 
 Let `m` be a member name which is not declared by _DV_ and not declared by
-an inline superinterface of _DV_. Assume that the interface of `R0`
-has a member named `m`. An invocation of `m` on a receiver of type `V` (or
-`V<T1 .. Ts>` if _DV_ is generic) is then treated as the same invocation,
-but with receiver type `R0`.
+an inline superinterface of _DV_. Assume that the interface of `Rj` has a
+member named `m` with signature `s` *(this is the combined member signature
+that may depend on other types in `R1 .. Rk`)*. An invocation of `m` on a
+receiver of type `V` (or `V<T1 .. Ts>` if _DV_ is generic) is then treated
+as the same invocation, but with signature `s`.
 
 *In other words: Conflicts among superinterfaces are treated the same,
 whether it is an inline or a non-inline superinterface. In both cases, _DV_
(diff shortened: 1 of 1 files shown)


working/1426-extension-types/feature-specification-implements-rep.md
```diff
@@ -0,0 +1,227 @@
...
+
+[inline class feature specification]: https://github.com/dart-lang/language/blob/main/accepted/future-releases/inline-classes/feature-specification.md
+
+## Change Log
+
+## Summary
+
+This is a proposal to add a new feature to inline classes: An inline class `V`
+may include a superinterface in its `implements` clause which is a
+supertype `R0` of the ultimate representation type `R` of `V` (this allows
```
lrhn, line 21: Should "supertype" here be "superinterface"?

That is, can an inline class on a `List<int>` choose to implement `Iterable<num>`?

Can it implement both `List<num>` and `Iterable<int>`, and therefore presumably also `Iterable<num>`, if we assume the shared members end up having a most specific signature?

working/1426-extension-types/feature-specification-implements-rep.md
```diff
@@ -0,0 +1,227 @@
...
+
+> A compile-time error occurs if _V1_ is a type name or a parameterized type which occurs as a superinterface in an inline class declaration _DV_, but _V1_ does not denote an inline type.
+
+It is adjusted to end as follows:
+
+> ... declaration _DV_, unless _V1_ denotes an inline type, or _V1_ is a supertype of the ultimate representation type of _DV_.
+
+
+Moreover, a compile-time error occurs if the implements clause of _DV_
+contains two or more types that are non-inline types. A compile-time error
```
lrhn, line 98: No need for the "two or more" rule.
If you have:
```dart
class _A implements I1, I2 { ... }
inline class A implements I1, I2 {
   final _A _this;
}
```
there is no problem.

working/1426-extension-types/feature-specification-implements-rep.md
eernstg: I don't see why we should prevent an inline class with representation type `List<int>` to have `implements Iterable<num>` and then be a subtype of `Iterable<num>` and have members from `Iterable` accordingly.

I haven't considered allowing any declaration with superinterfaces to have two or more distinct generic instantiations in their superinterface graph. In other words, `Iterable<num>` and `Iterable<int>` in the same superinterface graph is an error.

We could certainly create an expression of type `V<num>` whose representation object is a `List<int>`, which means that there is nothing at run-time which can occur when this is allowed which couldn't already occur. So it would be kind of an empty safety measure, just prohibiting something which can occur anyway.

working/1426-extension-types/feature-specification-implements-rep.md
```diff
@@ -0,0 +1,227 @@
...
+this feature. It is somewhat similar to the rule that `super(...)` can only
+occur as the last element in a constructor initializer list.*
+
+Let _DV_ be an inline class declaration named `V` with representation type
+`R` and assume that the `implements` clause of _DV_ includes the non-inline
+type `R0`. *We then have `R <: R0`, because anything else is an error.*
+
+Invocations of members declared by `V` or declared by an inline
+superinterface of _DV_ are unaffected by the fact that _DV_ implements
+`R0`.
```
lrhn, line 113: That's counter to the next paragraph which makes it an error if an inline superinterface and R0 both declare the same name (which is what I'd want, so great!)

So this paragraph is only true for invocations of members that `R0` does not declare a similarly named member for.

working/1426-extension-types/feature-specification-implements-rep.md
```diff
@@ -0,0 +1,227 @@
...
+
+Let `m` be a member name which is not declared by _DV_. Assume that the
+interface of `R0` has a member named `m`. A compile-time error occurs if an
+inline superinterface of _DV_ also declares a member named `m`.
```
lrhn, line 117: Should we use _basename_ here, to avoid allowing an inline superinterface method and an non-inline class setter?

working/1426-extension-types/feature-specification-implements-rep.md
```diff
@@ -0,0 +1,227 @@
...
+
+Let `m` be a member name which is not declared by _DV_ and not declared by
+an inline superinterface of _DV_. Assume that the interface of `R0`
+has a member named `m`. An invocation of `m` on a receiver of type `V` (or
+`V<T1 .. Ts>` if _DV_ is generic) is then treated as the same invocation,
+but with receiver type `R0`.
```
lrhn, line 123: Here I'd allow multiple non-inline superinterfaces, compute their combined interface as normal, and if that's unconflicted, use that as the signatures of the unshadowed members of _DV_.

working/1426-extension-types/feature-specification-implements-rep.md
```diff
@@ -0,0 +1,227 @@
...
+can resolve the conflict by redeclaring the given member. No override check
+is applied, any signature with the given name will resolve the conflict. If
+there is no conflict then _DV_ will "forward to" the members of `R0`.*
+
+*An implementation may choose to implicitly induce forwarding members into
+_DV_ in order to enable invocation of members of `R0`. However, such
+forwarding members must preserve the semantics of a direct invocation. In
+particular, if an invocation omits some optional parameters then the
+invocation of a member of `R0` must use the default value for that
+parameter, not a statically known value.*
```
lrhn, line 136: -> parameter of the actual instance method being called, ...

working/1426-extension-types/feature-specification-implements-rep.md
```diff
@@ -0,0 +1,227 @@
...
+  B(this.i);
+}
+
+void main() {
+  B b = B(42);
+  b.isEven; // OK.
+}
+```
+
+TODO!!!
```
lrhn, line 189: DO TOO! (But what?)

working/1426-extension-types/feature-specification-implements-rep.md
eernstg: I think that's true. However, I was aiming for a minimal model where we could add features like this later (if the complexity vs. benefit trade-off is OK).

Note also that we could certainly do this:

```dart
class _A implements I1, I2 { ... }
inline class A implements _A {
  final _A _this;
  A(this._this);
}
```

This would contribute the same members to `A` as `I1` and `I2` do, with signature resolution as done in `_A`, if needed, plus possibly some members of `_A`. If we just want to use `implements I1, I2` where `I1` and `I2` are non-inline types that are supertypes of the representation type then we can do it like this:

```dart
class _I1AndI2 implements I1, I2 {...} // Only abstract members, resolving conflicts, if any.
inline class A implements _I1AndI2 {...}
```

This means that `_I1AndI2` provides _exactly_ the members from `I1` and from `I2`, with conflict resolution as needed. 

The remaining issue is that `inline class A implements I1, I2` looks nicer in DartDoc.

Still looks like a future extension that we can return to if needed.

working/1426-extension-types/feature-specification-implements-rep.md
```diff
@@ -0,0 +1,227 @@
...
+
+It is adjusted to end as follows:
+
+> ... declaration _DV_, unless _V1_ denotes an inline type, or _V1_ is a supertype of the ultimate representation type of _DV_.
+
+
+Moreover, a compile-time error occurs if the implements clause of _DV_
+contains two or more types that are non-inline types. A compile-time error
+occurs if the implements clause of _DV_ contains a non-inline type that
+occurs in any other position than the last one.
```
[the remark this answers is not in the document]
eernstg: Completely unnecessary technically, so we could of course remove it. We could have a lint.

In any case, wouldn't it be nice if we had some convention about this? It is easier to require it now and remove the restriction later on..

working/1426-extension-types/feature-specification-implements-rep.md
eernstg: Right. Cleaned up.

working/1426-extension-types/feature-specification-implements-rep.md
```diff
@@ -0,0 +1,227 @@
...
+*This ensures that it is only necessary for a reader of the declaration to
+check the last implemented type in order to see whether or not it is using
+this feature. It is somewhat similar to the rule that `super(...)` can only
+occur as the last element in a constructor initializer list.*
+
+Let _DV_ be an inline class declaration named `V` with representation type
+`R` and assume that the `implements` clause of _DV_ includes the non-inline
+type `R0`. *We then have `R <: R0`, because anything else is an error.*
+
+Invocations of members declared by `V` or declared by an inline
```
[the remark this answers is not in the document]
eernstg: Done.

working/1426-extension-types/feature-specification-implements-rep.md
eernstg: Weird exception: We _do_ require getters and setters to be provided together as extension methods:

```dart
class A {
  set p(int i) {}
}

extension on A {
  int get p => 0;
}

void main() {
  A().p = A().p;
}
```

We get an error for `A().p` as an expression, because we can't find the getter: We look up the basename and find the setter as an instance member of `A`, then decide that we shouldn't even try extension methods, and declare that there is no getter.

This might be considered consistent because getting a member from an extension could be considered as being a completely different thing than getting it from the class of the receiver, so maybe it's fine that we insist on bundling getters and setters such that they must be provided by the class together, or by an extension together, if we somehow have both of them.

But the argument would then be that inline classes have a more symmetric relationship to each other: _Every_ inline class member is similar to an extension method, and the `implements` relationship ensures that we can always detect every conflict and require that it is resolved explicitly. I think this justifies the ability to put together a getter `p` and a setter `p=` from inline classes `I1` respectively `I2` by means of `inline class I implements I1, I2 {...}`. We can do the same thing with regular classes, and `I1`/`I2` are arguably in the "same domain".

So the remaining question is: Should we treat basename conflicts involving non-inline superinterfaces differently?

working/1426-extension-types/feature-specification-implements-rep.md
eernstg: Done!!! ;-)

working/1426-extension-types/feature-specification-implements-rep.md
eernstg: Did it anyway: There is now no restriction on the number of non-inline superinterfaces. They need to have a combined member signature in cases where more than one of them specify a member signature with the same name.

working/1426-extension-types/feature-specification-implements-rep.md
```diff
@@ -0,0 +1,227 @@
...
+
+Moreover, a compile-time error occurs if the implements clause of _DV_
+contains two or more types that are non-inline types. A compile-time error
+occurs if the implements clause of _DV_ contains a non-inline type that
+occurs in any other position than the last one.
+
+*This ensures that it is only necessary for a reader of the declaration to
+check the last implemented type in order to see whether or not it is using
+this feature. It is somewhat similar to the rule that `super(...)` can only
+occur as the last element in a constructor initializer list.*
```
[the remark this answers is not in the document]
eernstg: Removed the restriction on the order, and this commentary.

working/1426-extension-types/feature-specification-implements-rep.md
eernstg: Done.

working/1426-extension-types/feature-specification-implements-rep.md
lrhn: We require getter and setter to be provided together everywhere but in instance member inheritance.

For example we also consider it a conflict if two different imports provide a getter and a setter with the same base name.

The question here then becomes whether this is like instance member inheritance or not.
And whether inline class superinterfaces and non-inline-class superinterfaces (or supertypes) are somehow *different*. 
It should probably just work, no matter where you get the member from.

working/1426-extension-types/feature-specification-implements-rep.md
```diff
@@ -0,0 +1,227 @@
...
+The corresponding change to the feature specification is be that the
+following sentence in the section
+[Composing inline classes](https://github.com/dart-lang/language/blob/main/accepted/future-releases/inline-classes/feature-specification.md#composing-inline-classes)
+is adjusted:
+
+> A compile-time error occurs if _V1_ is a type name or a parameterized type which occurs as a superinterface in an inline class declaration _DV_, but _V1_ does not denote an inline type.
+
+It is adjusted to end as follows:
+
+> ... declaration _DV_, unless _V1_ denotes an inline type, or _V1_ is a supertype of the ultimate representation type of _DV_.
```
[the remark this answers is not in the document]
lrhn: Or maybe not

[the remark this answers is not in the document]
eernstg: OK, I'll keep 'supertype'.

working/1426-extension-types/feature-specification-implements-rep.md
eernstg: Sounds good. This is what we get by never saying 'basename', so we should be OK.

working/1426-extension-types/feature-specification-implements-rep.md
```diff
@@ -0,0 +1,268 @@
...
+of type `V`.
+
+[inline class feature specification]: https://github.com/dart-lang/language/blob/main/accepted/future-releases/inline-classes/feature-specification.md
+
+## Change Log
+
+## Summary
+
+This is a proposal to add a new feature to inline classes: An inline class `V`
+may include one or more superinterfaces in its `implements` clause which are
```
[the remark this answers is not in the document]
eernstg: Cool stuff! ;-)

First, though, I think this choice between 'superinterface' and 'supertype' is irrelevant: The direct superinterfaces of a given declaration that introduces a type (including `inline class` declarations) are the types mentioned in certain clauses (in this case we only have `implements`, but with two different kinds of elements in them). The indirect superinterfaces are just derived from that. So with `implements T` where `T` is a non-inline type, `T` must be a supertype of the representation type, and the _role_ that it plays is that of a direct superinterface.

Next, I would say that non-interface types should be included already among the types that we can specify like this, but I added a few words to emphasize that we actually mean it when we say that it can be any supertype of the representation type, and then I also used the `MapEntry` example. Nice!

I hadn't thought about the syntax issue. I haven't done anything about it yet, and a typedef can certainly give the appropriate syntax for any type. Otherwise, I tend to think (again!) that we should use a separate syntactic construct in order to specify declared subtype relationships to non-inline types. Perhaps `reveals T1, .. Ts`.

In any case, it _does_ make a difference that we wish to enable `int?` and `void Function([String])` as non-inline superinterfaces of an inline class, but non-inline classes can't have any of those.

I think it will work with multiple function types and the redeclared `call` method.

I don't think we need anything special for members of records: They are getters, they will be required to have a combined member signature, and that's well defined already. A subtype of two record types will need to be so based on covariance, because different shapes have unrelated types (or it's a bottom type, but an inline class whose representation type is `Never` isn't very important anyway).

`dynamic`: It would be nice if we could make this mean "invoke the members declared in the inline class statically; invoke anything else dynamically (except that `Object` members are somewhat statically known, as usual)". It would certainly conflict with any other superinterface because `dynamic` will always be able to come up with a member signature which is in conflict with that other superinterface. However, I've made it an error for now.

`void`: As usual, I think this should be an error.

`Null`: Anything special here? The representation type is probably `Null`, too.

`Never`: Similar. Probably OK, because it won't happen, and if someone does it then it's useless, but benign.

`Function`: Probably benign and useless again.

working/1426-extension-types/feature-specification-implements-rep.md
```diff
@@ -0,0 +1,268 @@
...
+void main() {
+  V v = V(D());
+  // v.m(3, 4); // Compile-time error: `m` signature is from `C`.
+  v.m(); // Returns 42, not 0.
+}
+```
+
+It is an implementation specific behavior whether a closurization
+*(also known as a tear-off)* of an inline class instance member which is
+obtained from the interface of `R0` is a tear-off of the member of the
```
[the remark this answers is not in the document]
eernstg: Yes, changed that.

It is not guaranteed that there exists a combined interface for `R1 .. Rk`, but each member which isn't redeclared has a combined member signature (otherwise we have a compile-time error). I'm currently referring to "a member of `R1 .. Rk`" and relying on earlier text where the member signature must be achieved by combination.

working/1426-extension-types/feature-specification-implements-rep.md
```diff
@@ -0,0 +1,268 @@
...
+Let `m` be a member name which is not declared by _DV_. Assume that the
+interface of `Rj` has a member named `m`. A compile-time error occurs if an
+inline superinterface of _DV_ also declares a member named `m`.
+
+Let `m` be a member name which is not declared by _DV_ and not declared by
+an inline superinterface of _DV_. Assume that the interface of `Rj` has a
+member named `m` with signature `s` *(this is the combined member signature
+that may depend on other types in `R1 .. Rk`)*. An invocation of `m` on a
+receiver of type `V` (or `V<T1 .. Ts>` if _DV_ is generic) is then treated
+as the same invocation, but with signature `s`.
```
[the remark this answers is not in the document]
eernstg: Yes, I think that should be ok.

working/1426-extension-types/feature-specification-implements-rep.md
```diff
@@ -0,0 +1,268 @@
...
+```
+
+It is an implementation specific behavior whether a closurization
+*(also known as a tear-off)* of an inline class instance member which is
+obtained from the interface of `R0` is a tear-off of the member of the
+representation object, or it is a tear-off of an implicitly induced
+forwarding method.
+
+*This makes no difference for the behavior of an invocation of the
+tear-off, but it does change the results returned by `==`.*
```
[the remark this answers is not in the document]
eernstg: Good catch, that is true!

I would tend to say that even this difference is allowed to be implementation specific. I think it should be OK for an implementation to avoid the forwarding functions entirely if possible, and it should also be possible to use forwarding functions as tear-offs (such tear-offs could be constant expressions "after desugaring")

In the case where the forwarder needs to preserve a default value which is not statically known I think we can rely on backends to use sentinel values, or on `foo$1` calling `forwardee.foo$1`, `foo$2` calling `forwardee.foo$2`, such that the forwarded call _actually_ passes the same number of actual arguments as the call site. Leaf had the impression that this could be implemented with a reasonable amount of work.

(We might also want to be in a good position to support faithful forwarding in a more general sense.)

(For `implemements X`, each invocation/tear-off of a member from the interface of `X` is known to be such an invocation/tear-off at the call site. If we actually allow the implementation to make the choice and, e.g., perform the invocation/tear-off directly on the forwardee, then the choice to make this implementation specific behavior could actually be quite helpful.)

So the main issue is actually that we could observe the forwarder based on the function object identity, and based on the run-time type of the function object.

I think this is still acceptable, so I changed the text to explain that the run-time type could depend on this choice (tear off the representation object member directly, or tear off a forwarding function).

working/1426-extension-types/feature-specification-implements-rep.md
```diff
@@ -0,0 +1,268 @@
...
+## Change Log
+
+## Summary
+
+This is a proposal to add a new feature to inline classes: An inline class `V`
+may include one or more superinterfaces in its `implements` clause which are
+supertypes `R1 .. Rk` of the ultimate representation type `R` of `V` (this
+allows `implements R` as a special case). This causes `V` to be a subtype
+of each of `Rj`, `j` in `1 .. k`, and it enables invocations of members of
+`Rj` on a receiver of type `V`.
```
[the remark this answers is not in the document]
eernstg: > we need to be more careful about which types are incompatible

I tend to think that there shouldn't be many new things here. In particular, it's only about member signatures, and both function types and nullable types have very few members, namely the ones of `Object`.

`NonNull(InlineType) == InlineType`, so that shouldn't create any new considerations.

If an inline type is a subtype of several nullable types and function types (and what not, insofar as this is even possible) then we'd need to assign it to one of those supertypes before we can use the properties of that specific supertype (for instance, we need to go from `InlineType implements int?` to `int?` before we can use `?.` on it.

> assignable to `int Function(int)` and `int Function(int, int)`, but not to `int Function(int, [int])`

I don't see a problem: A subtype rule about superinterfaces can confirm the connection to each of those function types, which yields the assignability to each of them, and then we're at home again. The inline type itself can't _use_ those superinterfaces to get any members unless all the supertypes agree (they say the same, or nothing).

> doesn't work, since `S` and `T` are not supertypes of `dynamic`

Agreed, that's an error. We really do insist that the representation type is a subtype of each of the non-inline superinterfaces.

> I only found problems with implementing the same interface in more than one way when I wrote something which needed a single answer.

It is possible that we can allow the same inline class to implement several different generic instantiations of the same class. However, I don't think that's going to be any better than implementing the least (most special) of those types, if it exists. There _must_ be some shared subtypes because the representation type is one such type.

Is the following example important? Why wouldn't we simply say that it is a compile-time error, and go on with our lives? ;-)

```dart
inline class A(List<num> it) implements List<num> {}
inline class B(List<int> it) implements A, List<int> {
  ... // Trying really hard, redeclare all 30-50 members that have a conflict.
}
```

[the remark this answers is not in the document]
eernstg: Actually, we wouldn't need to redeclare all those members as long as there is a most specific signature for each of the members (such that they have a combined member signature). This _does_ mean that we'd consider a `B` to have an `add` that accepts a `num` and a `first` that yields an `int`.

Still not sure we want this. ;-)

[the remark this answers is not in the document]
lrhn: ACK

> `NonNull(InlineType) == InlineType`, so that shouldn't create any new considerations.

The inline class type is not nullable because `Null` is not a subtype of it, even if the underlying object might be nullable, and the inline class itself is a subtype of `Object?` (but then, everything is).

working/1426-extension-types/feature-specification-implements-rep.md
```diff
@@ -0,0 +1,268 @@
...
+member named `m` with signature `s` *(this is the combined member signature
+that may depend on other types in `R1 .. Rk`)*. An invocation of `m` on a
+receiver of type `V` (or `V<T1 .. Ts>` if _DV_ is generic) is then treated
+as the same invocation, but with signature `s`.
+
+*In other words: Conflicts among superinterfaces are treated the same,
+whether it is an inline or a non-inline superinterface. In both cases, _DV_
+can resolve the conflict by redeclaring the given member. No override check
+is applied, any signature with the given name will resolve the conflict. If
+there is no conflict then _DV_ will "forward to" the members of `R0`.*
```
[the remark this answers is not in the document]
eernstg: > a synthetic forwarder method to an instance method, which we allow as an implementation, does not count as a static method.

Right! -- added a sentence to this effect.

working/1426-extension-types/feature-specification-implements-rep.md
```diff
@@ -0,0 +1,395 @@
...
+both types are structurally identical up to names of type variables. For
+instance, `prefix.C<int>` and `C<int>` are the same type if `prefix.C` and
+`C` resolve to the same class declaration, and similarly for `int`;
+`FutureOr<Object>` is the same type as `Object` because of the
+normalization; and `void Function<X>(X)` and `void Function<Y>(Y)` are the
+same type based on alpha equivalence (renaming of type variables).*
+
+A similar rule applies for the case where _DV_ has two superinterfaces,
+direct or indirect, which are function types or record types *(in which
+hase the types _must_ be non-inline types)*: They must also be the same
```
scheglov, line 200: `hase` -> `case`

working/1426-extension-types/feature-specification-implements-rep.md
```diff
@@ -0,0 +1,395 @@
...
+
+Let _DV_ be an inline class declaration named `V` with representation type
+`R` and assume that the `implements` clause of _DV_ includes the non-inline
+types `R1 .. Rk`. *We then have `R <: Rj` for each `j`, because anything
+else is an error.*
+
+Assume that `m` is a member which not declared by _DV_, and none of _DV_'s
+inline superinterfaces have a member named `m`, but one or more of the
+interfaces of `R1 .. Rk` has a member named `m`. A compile-time error
+occurs if there exist `j1` and `j2` in `1 .. k` and a member name `m` such
```
scheglov, line 211: We define `j1` and `j2`, and don't use them.
Do I miss something?
I understand what this paragraph saying, and it makes total sense.
Just nit picking.

working/1426-extension-types/feature-specification-implements-rep.md
eernstg: Thx, fixed!

working/1426-extension-types/feature-specification-implements-rep.md
eernstg: True, fixed, thanks again!

eernstg:
Note that the proposal in this PR is included in the proposal about extension types, https://github.com/dart-lang/language/pull/3182.

eernstg:
Closing: This feature has been included in the 'extension types' feature specification, so there's no need to have a separate feature specification (which is what this PR will add).

PR #3146, opened 2023-06-13
Update arguments spec to allow TypeAnnotations instead of Identifiers

jakemac53:
The real intention of allowing Identifiers was to allow types, but we also want to allow type arguments. This removes the ability to have arbitrary Identifier arguments and replaces that with TypeAnnotation arguments.

We might want to add Identifier arguments also, to handle non-type identifiers, but I don't really have a use case.

### DIFF
--- a/working/macros/feature-specification.md
+++ b/working/macros/feature-specification.md
@@ -1125,7 +1125,7 @@ macro), then the caller is responsible for creating the Code object.
 As usual, it is a compile-time error if the type of any argument value (which
 may be a Code object) is not a subtype of the corresponding parameter type.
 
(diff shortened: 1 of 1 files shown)


working/macros/feature-specification.md
```diff
@@ -1124,18 +1126,17 @@ As usual, it is a compile-time error if the type of any argument value (which
 may be a Code object) is not a subtype of the corresponding parameter type.
 
 It is a compile-time error if an macro class constructor invoked by a macro
```
munificent, line 1128: Nit: "if an macro" -> "if a macro".

munificent (approved):

Issue #1929, opened 2021-10-21
Should we allow non-literal const expressions for macro value arguments?


munificent:
Arguments to macro applications can be unevaluated expressions or values of primitive types. For the latter, only literals are supported. Do we want to expand that to allow other const expressions that are safe to evaluate before macros are expanded? If so, what expressions do we allow?

We probably need to at least allow unary `-` since otherwise there's no way to pass negative number as a value argument to a macro. :)

Do we want to allow other arithmetic operators (on literal values)?

One option is to allow any valid const expression as long as it contains no identifiers. That would let users do basic arithmetic and maybe string concatenation, but should avoid the kinds of expressions that can't be evaluated before macros are run.

jakemac53:
In https://github.com/dart-lang/language/pull/3147 I expand the syntax to allow all constant expressions but only containing literals.

I think it significantly simplifies things to keep it limited to literals, and we can always add the ability to reference identifiers later on if necessary.

PR #3147, opened 2023-06-13
allow all constant expressions as long as they contain only literals

jakemac53:
Closes https://github.com/dart-lang/language/issues/1929 by allowing all constant expressions, but they can only contain literals.

### DIFF
--- a/working/macros/feature-specification.md
+++ b/working/macros/feature-specification.md
@@ -1092,17 +1092,13 @@ it specifies through parameter types:
 
 *   If the parameter type is `bool`, `double`, `int`, `Null`, `num`, `String`,
     `List`, `Set`, or `Map`, (or the nullable forms of any of those), then the
-    argument expression must be a boolean, number, null, string, list, set, or
-    map literal.
+    argument expression must be a constant expression containing only boolean,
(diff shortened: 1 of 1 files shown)


working/macros/feature-specification.md
```diff
@@ -1092,21 +1092,13 @@ it specifies through parameter types:
...
-      but all the actual entries must be of one of the supported types.
-
-    **TODO**: Do we want to allow more complex expressions? Could we allow
-    constant expressions whose identifiers can be successfully resolved before
-    macro expansion (#1929)?
+    argument expression must be a constant expression containing only boolean,
+    number, null, string, list, set, or map literals.
+
+    * Any type arguments (inferred or explicit) must be one of the allowed
+      parameter types or `Object`, recursively. Note that `Object` is allowed
```
munificent, line 1099: Probably need to call out that `Object?` is allowed too since it doesn't fall out from the previous bullet point.

munificent (approved):

working/macros/feature-specification.md
jakemac53: I moved this around a bit, added `Object` and `dynamic` explicitly above.

Issue #3148, opened 2023-06-14
Function literal return type inference fails on divergent future


eernstg:
Consider the following program:

```dart
class Divergent<T> implements Future<Divergent<Divergent<T>>> {
  noSuchMethod(Invocation invocation) => super.noSuchMethod(invocation);
}

void main() => () async => Divergent<int>();
```

This program is rejected by the CFE and the analyzer with similar error messages. Analyzer:

```
The return type 'Divergent<int>' isn't a 'Future<Divergent<Divergent<Divergent<int>>>>', as required by the closure's context.
```

CFE:

```
lib/main.dart:6:28:
Error: A value of type 'Divergent<int>' can't be returned from an async function with return type 'Future<Divergent<Divergent<Divergent<int>>>>'.
 - 'Divergent' is from 'package:dartpad_sample/main.dart' ('lib/main.dart').
 - 'Future' is from 'dart:async'.
void main() => () async => Divergent<int>();
                           ^
```

The return type inference for a function literal is specified [here](https://github.com/dart-lang/language/blob/main/resources/type-system/inference.md#function-literal-return-type-inference).

Checking the current definition of `flatten` we get `flatten(Divergent<T>) == Divergent<Divergent<T>>`.

For the inference of the return type of the function literal, we note that the imposed return type schema has no relevant information (and we get the same error with `var _ = /*same function literal*/;`), and also that the returned expression has type `Divergent<int>` as stated, which is then also the type of the returned expression after type inference for any context type because there's no thing for type inference to do.

So the actual returned type is `flatten(Divergent<int>)`, that is, `Divergent<Divergent<int>>`. This implies that the inferred return type `R` for the function literal is `Future<flatten(Divergent<Divergent<int>>)>`, that is, `R == Future<Divergent<Divergent<Divergent<int>>>>`.

This shows that the implementations are computing the specified return type, and the only remaining problem is that the given returned expression does not satisfy the typing constraints associated with that return type.

So is it justified that we're returning a `Divergent<int>` in an `async` function whose return type is `R == Future<Divergent<Divergent<Divergent<int>>>>`? Should we get an error or not?

Let's explore some parts of the supertype graph of `Divergent<int>`, in order to see whether or not we can find a supertype which should be accepted. If this is true then `Divergent<int>` should arguably also be accepted.

[...]

PR #3149, opened 2023-06-15
Adjust function literal return type inference

eernstg:
Issue https://github.com/dart-lang/language/issues/3148 showed that we need to adjust the rules about function literal return type inference. This PR changes the rules accordingly.

The basic idea is that the last step in the inference of an `async` function used to use the inferred return type of `Future<flatten(S)>` where `S` is taken from the context or from the returned expressions, but it should instead be `Future<S>`. Also, `flatten` then needs to be applied to the type obtained from the context, such that we still have the same return type as before when the context type gets to decide the outcome.

### DIFF
--- a/resources/type-system/inference.md
+++ b/resources/type-system/inference.md
@@ -356,10 +356,11 @@ With null safety: if `R` is `void`, or the function literal is marked `async`
 and `R` is `FutureOr<void>`, let `S` be `void` (without null-safety: no special
 treatment is applicable to `void`).
 
-When the function is marked `async`: if `T <: futureValueType(R)` then let `S`
-be `T`; otherwise let `S` be `futureValueType(R)`. When the function is not
-marked `async`: if `T <: R` then let `S` be `T`; otherwise let `S` be `R`. The
-inferred return type of the function literal is then defined as follows:
+If the previous paragraph did not yield a value for `S`: When the function is
+marked `async`: if `T <: futureValueType(R)` then let `S` be `T`; otherwise let
+`S` be `futureValueType(R)`. When the function is not marked `async`: if `T <:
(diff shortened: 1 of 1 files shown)


resources/type-system/inference.md
```diff
@@ -345,15 +349,22 @@ all `return` and `yield` statements in the block body have been considered.
...
 With null safety: if `R` is `void`, or the function literal is marked `async`
 and `R` is `FutureOr<void>`, let `S` be `void` (without null-safety: no special
 treatment is applicable to `void`).
 
-Otherwise, if `T <: R` then let `S` be `T`.  Otherwise, let `S` be `R`.  The
-inferred return type of the function literal is then defined as follows:
+If the previous paragraph did not yield a value for `S`: When the function is
+marked `async`: if `T <: futureValueType(R)` then let `S` be `T`; otherwise let
+`S` be `futureValueType(R)`. When the function is not marked `async`: if `T <:
+R` then let `S` be `T`; otherwise let `S` be `R`.
```
eernstg, line 362: We need to use `futureValueType(R)`, not `flatten(R)`. If we were to use flatten then we'd carry `?` over from the context. So we could have a context type that wants a `Future<num>?`, then we'd get `int?` as the value of `R`, and then we'd get an inferred return type of `Future<int?>` where the context wants a return type of `Future<int>?`, and that's a type error.

```dart
Future<int>? Function() f = () async => 1;
```

In this example `R` is `Future<int>?` and `flatten(R)` is `int?`. But we don't want the function literal to have a return type of `Future<int?>`, it should be `Future<int>`.

eernstg:
Closing: https://github.com/dart-lang/language/pull/3151 contains a PR which simplifies a larger number of locations in this section because it avoids wrapping `K` in `FutureOr` in the first place, and that's an overall improvement.

PR #3151, opened 2023-06-15
Update inference of closures with no context.

lrhn:
Issue https://github.com/dart-lang/language/issues/3148 showed that we need to adjust the rules about function literal return type inference. This PR changes the rules accordingly.

Refactors to add the `FutureOr` to the context type in the individual return statements for `async` functions, rather than add it to the context type scheme for the function. This simplifies the rules later, but requires splitting `async` and plain functions. (Well, doesn't require, but most of the text would be by-case anyway.)

Also make it clear that `return` statements, or reaching the end of the body, does not affect the element types of generators.


resources/type-system/inference.md
```diff
@@ -286,9 +291,9 @@ type schema `S` as follows:
     `Stream<S1>` for some `S1`, then the context type is `S1`.
   - If the function expression is declared `sync*` and `S` is of the form
     `Iterable<S1>` for some `S1`, then the context type is `S1`.
```
eernstg, line 293: If the imposed context type schema is `Stream<int>?` or `Object` (or a number of other types that aren't of the form `Stream<S1>` for any `S1`) then we do not get a binding of `S1` here.

We already have the notion of 'the element type of a generator', so we'd need 'the element type of a type'. This would be `S` for any type of the form `Stream<S>`, `Stream<S>?`, `Iterable<S>`, `Iterable<S>?`, and it would be `Object?` for any other type. A generator function is a compile-time error unless it has a return type which is a supertype of `Iterable<Never>` resp. `Stream<Never>`, but the context type could be anything, and it could also be `dynamic` or `void` or `Stream<int>?` and so on.

This is going to give us a better context type for generators.

resources/type-system/inference.md
```diff
@@ -315,45 +320,60 @@ described below with a typing context as computed above.
 The actual returned type of a function literal with a block body is computed as
```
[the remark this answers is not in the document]
leafpetersen: I think this also needs updating to use FutureOr<K> where K is the typing context from above (when required), since you no longer do this up front.

resources/type-system/inference.md
```diff
@@ -315,45 +320,60 @@ described below with a typing context as computed above.
...
-    context `K`, and update `T` to be `UP(flatten(S), T)` if the enclosing
-    function is `async`, or `UP(S, T)` otherwise.
-  - For each `return;` statement in the block, update `T` to be `UP(Null, T)`.
+  - If the enclosing function is a non-`async` non-generator function,
+    for each `return e;` statement in the block, let `S` be the inferred type
+    of `e`, using the local type inference algorithm described below with typing
+    context `K`, and update `T` to be `UP(S, T)`.
+  - If the enclosing function is marekd `async`, for each `return e;` statement
+    in the block, let `S` be the inferred type of `e`, using the local type
+    inference algorithm described below with typing context `FutureOr<K>`,
```
leafpetersen, line 336: I don't think this is quite right.  You only want to use `FutureOr<K>` when `K` is not empty.  That is, the way that this was specified before, I think `() async { return e; }` would infer `e` using `_` as the context, whereas as written I think this implies that you use `FutureOr<_>` as the typing context.  I believe this can be observable.

resources/type-system/inference.md
```diff
@@ -315,45 +320,60 @@ described below with a typing context as computed above.
...
     context `K`, and update `T` to be `UP(S, T)`.
   - If the enclosing function is marked `sync*`, then for each `yield* e;`
     statement in the block, let `S` be the inferred type of `e`, using the
     local type inference algorithm described below with a typing context of
-    `Iterable<K>`; let `E` be the type such that `Iterable<E>` is a
-    super-interface of `S`; and update `T` to be `UP(E, T)`.
+    `Iterable<K>`. If there exists a type `E` such that `Iterable<E>` is a
+    super-interface of `S`, update `T` to be `UP(E, T)`. Otherwise update
+    `T` to be `UP(S, T)`.
+    _It is a compile-time error if *S* is not a assignable to
```
leafpetersen, line 349: This seems reasonable, the previous behavior was underspecified.  Note that the analyzer and the CFE currently disagree here, so there is an implementation change implied.  Test case to reveal the difference:
```dart
void main() {
  var f = () sync* {
    yield* (throw "hello");
  };
  int x = f();
  var g = () sync* {
    yield* (3 as dynamic);
  };
  int y = g();
}
```

resources/type-system/inference.md
```diff
@@ -315,45 +320,60 @@ described below with a typing context as computed above.
...
+    `Iterable<Object?>`, so either *S* implements `Iterable`, or it is one of
+    `dynamic` or `Never`._
   - If the enclosing function is marked `async*`, then for each `yield* e;`
     statement in the block, let `S` be the inferred type of `e`, using the
     local type inference algorithm described below with a typing context of
-    `Stream<K>`; let `E` be the type such that `Stream<E>` is a super-interface
-    of `S`; and update `T` to be `UP(E, T)`.
+    `Stream<K>`. If there exists a type `E` such that `Stream<E>` is a
```
leafpetersen, line 355: I would validate whether this implies an implementation change, as above.

resources/type-system/inference.md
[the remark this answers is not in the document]
lrhn: ACK.

resources/type-system/inference.md
lrhn: Oh wonders! :)

OK, so if `K` is not `_`, use `FutureOr<K>` as context type schema for `e`, otherwise use `_`.

lrhn:
Updated to account for `=>`-bodies.
Also update to compute a correct and valid "context type" for, e.g., a `sync*` function with context type `FutureOr<Iterable<int>?>`, which should be returning a `Future<int>`.

(That should also be the returned iterable type for a function with that as *declared* return type.)

resources/type-system/inference.md
lrhn: Is it a *problem* to use `FutureOr<_>`, or is it just a *difference*. (We probably don't want a difference either, just want to make sure I'm not missing something problematic.)

resources/type-system/inference.md
```diff
@@ -315,45 +320,60 @@ described below with a typing context as computed above.
...
 Let `R` be the greatest closure of the typing context `K` as computed above.
 
-With null safety: if `R` is `void`, or the function literal is marked `async`
-and `R` is `FutureOr<void>`, let `S` be `void` (without null-safety: no special
-treatment is applicable to `void`).
+With null safety, if `R` is `void`, let `S` be `void`
+_(without null-safety: no special treatment is applicable to `void`)_.
 
-Otherwise, if `T <: R` then let `S` be `T`.  Otherwise, let `S` be `R`.  The
+Otherwise (_without null safety or if `R` is not `void`_),
```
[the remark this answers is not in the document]
lrhn: Then we haven't specified `S` if `R` is void and without null safety.

We have one clause:
If (with null safety && R is void): S = `void`
And an otherwise, which should cover both of:
* with null safety && R is not void
* without null safety
(aka "without null safety or if R is not void")

lrhn:
Bad rebase!

eernstg:
> Bad rebase!

Ah, so that's the reason why it adds `extension-types/feature-specification.md`.

resources/type-system/inference.md
```diff
@@ -299,61 +308,137 @@ The function **futureValueTypeSchema** is defined as follows:
...
-- Otherwise, for all `S`, **futureValueTypeSchema**(`S`) = `Object?`.
-
-_Note that it is a compile-time error unless the return type of an asynchronous
-non-generator function is a supertype of `Future<Never>`, which means that
-the last case will only be applied when `S` is `Object` or a top type._
+- Otherwise, for all other `S`, **futureValueTypeSchema**(`S`) = `Object?`.
+
+The function **streamElementTypeSchema** is defined as follow:
+
+- **streamElementTypeSchema**(`S?`) = **streamElementTypeSchema**(`S`),
```
leafpetersen, line 315: These can't actually return null right?  So why do we allow these?

resources/type-system/inference.md
```diff
@@ -299,61 +308,137 @@ The function **futureValueTypeSchema** is defined as follows:
...
+    for all `S`.
+- **streamElementTypeSchema**(`S*`) = **streamElementTypeSchema**(`S`),
+    for all `S`.
+- **streamElementTypeSchema**(`FutureOr<S>`) = **streamElementTypeSchema**(`S`),
```
leafpetersen, line 319: Similarly, why allow this?

resources/type-system/inference.md
```diff
@@ -299,61 +308,137 @@ The function **futureValueTypeSchema** is defined as follows:
...
+
+Let `K` be the value context schema for the function body as computed above
+from the imposed return type schema.
+_When we refer to the inferred type of an expression with a typing context,
+it is the static type of the expression inferred using the local type inference
+algorithm described below._
+
+The actual value type of a function literal with an expression body, `=> e`,
+_(which cannot be a generator function)_ is computed as follows:
+  - If the enclosing function is marked `async`,
```
leafpetersen, line 361: I think "enclosing function" is ambiguous here - I think you mean "function which encloses `e`", but the text also supports a reading as "function which encloses the function literal".

resources/type-system/inference.md
leafpetersen: I'm confused - I don't see where this happens.  The text seems to uniformly infer `return e` in a context of `FutureOr<K>`, which means when `K` is `_`, it will infer it using `FutureOr<_>` no?

eernstg:
Maybe force-push the state of this PR at a commit which precedes the bad rebase?

I'm assuming that several modified files should simply not be modified by this PR (e.g., accepted/future-releases/extension-types/feature-specification.md, specification/dart.sty, specification/dartLangSpec.tex, working/1426-extension-types/feature-specification-inline-classes.md, working/macros/example/analysis_options.yaml, working/macros/example/benchmark/simple.dart, and several others).

resources/type-system/inference.md
```diff
@@ -299,66 +308,143 @@ The function **futureValueTypeSchema** is defined as follows:
...
+
+  - If the function literal is marked `sync*`, the relevant statements
+    are `yield e;` or `yield* e;` statement.
+    * For a `yield e;` statement, let `V` be the inferred type of `e` with
+      typing context `K`.
+    * For a `yield* e;` statement, let `S` be the inferred type of `e` with
+      typing context `Iterable<K>`.
+      If `S` implements `Iterable<R>` for some `R`, let `V` be `R`.
+      Otherwise let `V` be `S`.
+      _It is a compile-time error if `S` is not a assignable to
```
leafpetersen, line 404: not a assignable -> not assignable

resources/type-system/inference.md
```diff
@@ -299,66 +308,143 @@ The function **futureValueTypeSchema** is defined as follows:
...
+
+  - If the function literal is marked `async*`, the relevant statements are
+    `yield e;` or `yield* e;` statements.
+    * For a `yield e;` statement, let `V` be the inferred type of `e` with
+      typing context `K`.
+    * For a `yield* e;` statement, let `S` be the inferred type of `e` with
+      typing context `Stream<K>`.
+      If `S` implements `Stream<R>` for some `R`, let `V` be `R`.
+      Otherwise let `V` be `S`.
+      _It is a compile-time error if `S` is not a assignable to
```
leafpetersen, line 416: not a assignable -> not assignable

resources/type-system/inference.md
```diff
@@ -299,66 +308,143 @@ The function **futureValueTypeSchema** is defined as follows:
...
+The **actual value type** of the function literal is the value of `T` after
+all relevant statements in the block body have been processed.
+
+Let `T` be the **actual value type** of a function literal as computed above.
 Let `R` be the greatest closure of the typing context `K` as computed above.
 
-With null safety: if `R` is `void`, or the function literal is marked `async`
-and `R` is `FutureOr<void>`, let `S` be `void` (without null-safety: no special
-treatment is applicable to `void`).
+_Then compute the actual value/element type, `S`, to use in the inferred return
```
leafpetersen, line 428: I think this is the first time element type is mentioned in this doc.  If you're going to use it define it, but maybe better to just delete?

resources/type-system/inference.md
```diff
@@ -299,66 +308,143 @@ The function **futureValueTypeSchema** is defined as follows:
...
+type of the function literal, based on the actual value type, but bounded
```
leafpetersen, line 429: This comment is confusing to me.  You say something like "compute X, to use in Y, based on X, but bounded by Z".  I think the two Xs above are meant to be different, so use different terminology?

leafpetersen (approved):
LGTM.  I'm mildly worried about unexpected behavior changes.  Can we aim to have a small set of tests that capture the behavior changes that we expect from this, and then communicate that any other test changes should be punted to us for review?

PR #3176, opened 2023-06-29
Add functional widget macro example

jakemac53:
This adds a new macro `@FunctionalWidget(<identifier>)`. It generates a widget with the given name (from the identifier) from a function. Fields are inferred from the parameters, for example:

```dart
@FunctionalWidget(MyApp)
Widget _buildApp(BuildContext context, ValueNotifier<int> counter,
    {String? appTitle, String? homePageTitle}) {
  return MaterialApp(
      title: appTitle ?? 'Flutter Demo',
      theme: ThemeData(primarySwatch: Colors.blue),
      home: MyHomePage(counter,
          title: homePageTitle ?? 'Flutter Demo Home Page'));
}
```
Would generate this widget class:
```dart
class MyApp extends StatelessWidget {
  final ValueNotifier<int> counter;
  final String? appTitle;
  final String? homePageTitle;
  
  const MyApp(this.counter, {this.appTitle, this.homePageTitle, Key? key})
      : super(key: key);
      
  @override
  Widget build(BuildContext context) => _buildApp(context, counter,
      appTitle: appTitle, homePageTitle: homePageTitle);
}
```

I also applied some fixes to other macros to update them to the latest apis.

### DIFF
--- a/working/macros/example/lib/functional_widget.dart
+++ b/working/macros/example/lib/functional_widget.dart
@@ -45,7 +45,8 @@ class FunctionalWidget implements FunctionTypesMacro {
               in positionalFieldParams.followedBy(function.namedParameters))
             DeclarationCode.fromParts([
               'final ',
-              param.type,
+              param.type.code,
+              ' ',
               param.identifier.name,
               ';',
             ]),
--- a/working/macros/example/lib/functional_widget.dart
+++ b/working/macros/example/lib/functional_widget.dart
@@ -5,7 +5,7 @@
 // There is no public API exposed yet, the in progress api lives here.
 import 'package:_fe_analyzer_shared/src/macros/api.dart';
 
-class FunctionalWidget implements FunctionTypesMacro {
+macro class FunctionalWidget implements FunctionTypesMacro {
   final Identifier? widgetIdentifier;
 
   const FunctionalWidget(
--- a/working/macros/example/lib/functional_widget.dart
+++ b/working/macros/example/lib/functional_widget.dart
@@ -1,0 +1,75 @@
+// Copyright (c) 2023, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// There is no public API exposed yet, the in progress api lives here.
+import 'package:_fe_analyzer_shared/src/macros/api.dart';
+
+class FunctionalWidget implements FunctionTypesMacro {
+  final Identifier? widgetIdentifier;
+
+  const FunctionalWidget(
+      {
+      // Defaults to removing the leading `_` from the function name and calling
+      // `toUpperCase` on the next character.
+      this.widgetIdentifier});
+
+  @override
+  void buildTypesForFunction(
+      FunctionDeclaration function, TypeBuilder builder) {
+    if (!function.identifier.name.startsWith('_')) {
+      throw ArgumentError(
+          'FunctionalWidget should only be used on private declarations');
+    }
+    if (function.positionalParameters.isEmpty ||
+        // TODO: A proper type check here.
+        (function.positionalParameters.first.type as NamedTypeAnnotation)
+                .identifier
+                .name !=
+            'BuildContext') {
+      throw ArgumentError(
+          'FunctionalWidget functions must have a BuildContext argument as the '
+          'first positional argument');
+    }
+
+    var widgetName = widgetIdentifier?.name ??
+        function.identifier.name
+            .replaceRange(0, 2, function.identifier.name[1].toUpperCase());
+    var positionalFieldParams = function.positionalParameters.skip(1);
+    builder.declareType(
+        widgetName,
+        DeclarationCode.fromParts([
+          'class $widgetName extends StatelessWidget {',
+          // Fields
+          for (var param
+              in positionalFieldParams.followedBy(function.namedParameters))
+            DeclarationCode.fromParts([
+              'final ',
+              param.type,
+              param.identifier.name,
+              ';',
+            ]),
+          // Constructor
+          'const $widgetName(',
+          for (var param in positionalFieldParams)
+            'this.${param.identifier.name}, ',
+          '{',
+          for (var param in function.namedParameters)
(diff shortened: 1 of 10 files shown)


jakemac53:
One cool thing to note here is that the argument passed to `FunctionalWidget` is not a string but an actual Identifier, which will end up being a reference to the actual class that is generated by the macro. You should be able to navigate directly to the class in the augmentation library which is super handy.

Also, this macro does not require any introspection and so it runs purely in the first (types) phase, and completes very quickly (well under <1 ms for a very trivial example). As with other macros formatting the code takes longer than running the macro by a lot, so I am increasingly wary of trying to format macro generated code, unless we used a different formatter that does a less good job but runs faster.

rrousselGit:
> One cool thing to note here is that the argument passed to `FunctionalWidget` is not a string but an actual Identifier, which will end up being a reference to the actual class that is generated by the macro. You should be able to navigate directly to the class in the augmentation library which is super handy.

In a real-world example, I doubt this would be the desired API. Instead, folks would likely want the widget name to be picked up based on the annotated function name.


What would instead be more useful would be if we could augment the annotated function to add extra annotations to it. Like:


```dart
@stateless
Widget myApp() => Text('hello world');

// generates:
class MyApp {...} // nothing special here

@internal // Mark "myApp" as internal or deprecated to avoid it being used
augment Widget myApp();
```

jakemac53:
> In a real-world example, I doubt this would be the desired API. Instead, folks would likely want the widget name to be picked up based on the annotated function name.

The macro does support this as well, it's an optional parameter. But it is nice imo to illustrate that you can do this and there are some benefits imo to it in terms of discoverability of the actual class.

> What would instead be more useful would be if we could augment the annotated function to add extra annotations to it. 

I did recently expand the apis to allow adding doc comments - we could similarly allow adding annotations. That should be totally doable.

working/macros/example/lib/functional_widget.dart
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) 2023, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// There is no public API exposed yet, the in progress api lives here.
```
munificent, line 5: "in progress api" -> "in-progress API". :)

working/macros/example/lib/functional_widget.dart
```diff
@@ -0,0 +1,76 @@
...
+          @override
+          Widget build(BuildContext context) => ''',
+          function.identifier,
+          '(context, ',
+          for (var param in positionalFieldParams) '${param.identifier.name}, ',
+          for (var param in function.namedParameters)
+            '${param.identifier.name}: ${param.identifier.name}, ',
+          ');',
+          '}',
+        ]));
```
munificent, line 74: Some kind of tagged string or template syntactic sugar would definitely be nice for building the generated code.

munificent (approved):
Looks great!

munificent:
> As with other macros formatting the code takes longer than running the macro by a lot, so I am increasingly wary of trying to format macro generated code, unless we used a different formatter that does a less good job but runs faster.

Almost all of the runtime code of formatting is line splitting, so a hack you can use to make it faster is to set a much wider line length.

I have ideas for a better internal representation for the formatter that I hope (but can't prove) will improve performance. I'm hoping to try that out this summer.

jakemac53:
> I have ideas for a better internal representation for the formatter that I hope (but can't prove) will improve performance. I'm hoping to try that out this summer.

Fwiw, on a macro like this in the benchmark I have (with all the usual caveats) I am seeing formatting take just under 1ms, while the macro (including instantiation) takes about 0.2 ms. So unless we can get it down by roughly an order of magnitude I think it's going to be problematic.

It generally just doesn't seem to me like having code formatting on the critical path of compilation is a good idea haha.

working/macros/example/lib/functional_widget.dart
jakemac53: Yeah it really would, it would also make it much easier to format the code nicely by  yourself.

rrousselGit:
I assume that the main issue would be error messages.
If there's a compilation error, we'd want line numbers to be correct


So maybe the compiler would do formatting only if there's a missing identifier, for the sake of debugging the problem.
Speed likely doesn't matter as much then

jakemac53:
> I assume that the main issue would be error messages.
> If there's a compilation error, we'd want line numbers to be correct

Yeah this is exactly the problem. In an ideal world I think we would show the code formatted to the user, but not actually format it. We should only need a source map between the original and formatted versions to support that. But I don't ultimately know how feasible it would be.

rrousselGit:
Honestly I think it wouldn't be terrible if the error message in generated files isn't great.
If necessary, I wouldn't mind the error either pointing to unformatted code, or sight up not having line info. 

Heck, that's kind of the current behavior with build_runner. 
If there's a syntax error, in generated code, formatting won't work. So we have non-ideal errors. But I'm personally used to it.

___

Ultimately, if a macro generates invalid code, that's a bug on the macro side.
They should show a proper error message on the annotated element instead


So for the vast majority of the community, they probably don't care at all.

jakemac53:
I am a lot less concerned about the static analysis errors and more concerned about stack traces for runtime errors.

bernaferrari:
> I have ideas for a better internal representation for the formatter that I hope (but can't prove) will improve performance. I'm hoping to try that out this summer.

Would re-writing the formatter in Rust/Zig/C++/etc be something desirable? Specially with GPT/Bard, you could probably translate most of formatter in no time at all. From my experience, Rust formatter is extremely fast.

jakemac53:
Just to clarify here, I am not claiming the Dart formatter is slow. For normal use I would argue it is extremely fast. I am only claiming it's probably not a good idea to put it on the hot path for compilation, because that falls far outside the typical usage pattern (it would be common to be formatting thousands of files, so even at ~1ms/file that adds significant time to a compile).

The only time I notice slow formatting in dart is actually because the Dart language server is single threaded, so in your IDE the formatting task can be slow if some other task is blocking it.

PR #3182, opened 2023-06-30
Extension types feature specification

eernstg:
This PR is a proposal for how we could specify extension types, following up on the discussions it the language team this week.

cc @dart-lang/language-team


eernstg:
It includes the special case of the primary constructor syntax, and it makes this syntax required (it's a syntax error to not have it). It allows non-extension-type superinterfaces as long as they are supertypes of the representation type.

As a thought experiment (which we could take seriously), it does _not_ make the representation name available as a getter. This means that the representation object is treated, as we've mentioned in various discussions, as a name which is in scope in the extension type body but not accessible from anywhere else (just like type parameters).

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1098 @@
...
+constructors (in particular, it can occur by means of `this.id`, or in
+an initializer list, or by an initializing expression in the
+declaration itself, but it is an error if it does not occur at all).*
+
+An extension type `V` used as an expression (*a type literal*) is allowed
+and has static type `Type`.
+
+An extension type declaration _DV_ can have the type modifier
+`final`. In this case it is a compile-time error for another extension type
+declaration _DV2_ to have _DV_ as a superinterface.
```
bwilkerson, line 834: Is this limited to extension types outside the library declaring _DV_, or is there an intentional difference here? (If the latter, consider making the distinction explicit in the text.)

bwilkerson (commented):
FWIW, I think this proposal would be much less confusing to users than the inline classes proposal. In particular I think the semantics of both redeclaration and fields are far less surprising.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1098 @@
...
+  - Initial version, which is a copy of the views proposal, renaming 'view'
+    to 'inline', and adjusting the text accordingly. Also remove support for
+    primary constructors.
+
+
+## Summary
+
+This document specifies a language feature that we call "extension types".
+
+The feature introduces the _extension type_ feature. Is feature introduces a new
```
scheglov, line 36: `Is feature` what?

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1098 @@
...
+disambiguate the extension type declaration with a fixed lookahead.*
+
+Some errors can be detected immediately from the syntax:
+
+A compile-time error occurs if the extension type declaration declares any
+instance variables.
+
+The _name of the representation_ in an extension type declaration is the
+identifier `id` that occurs in the representation declaration, and the
+_type of the representation_ is the declared type of `id`.
```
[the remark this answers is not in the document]
eernstg: Right, that was ambiguous. Fixed.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1098 @@
...
+the implements clause of several extension type declarations
+<code>V<sub>1</sub> .. V<sub>k</sub></code>, thus "inheriting" the members
+of `V` into all of <code>V<sub>1</sub> .. V<sub>k</sub></code> without code
+duplication.*
+
+*The reason why this mechanism uses the keyword `implements` rather
+than `extends` to declare a relation that involves "inheritance" is
+that it has a similar semantics as that of extension members (in that
+they are statically resolved, and each member is applicable to every
+subtype).*
```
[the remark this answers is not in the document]
eernstg: The big difference is that a mixin application is ordered, and members with the same name will have the implementation that is declared in the most specific (latest applied) mixin. With extension method, you have unlimited symmetric combination ("multiple inheritance"), and then it's an error if there are name clashes on equally specific extensions.

With extension types, we insist on resolving name clashes explicitly by a redeclaration in the most specific extension type. We don't have a notion of `super` invocations, so the developer who just wants to select one specific implementation must write a forwarder (something like `void m() => (this as V3).m();`). We could add some notion of `super` later on, and this could allow those forwarders to be much more concise (avoiding the duplication of long parameter lists and such), but we don't have that kind of mechanism now. It should be no problem to start using it later, if we add it.

`implements` is relevant because extension methods are applicable to any receiver whose type is a subtype of the on-type:

```dart
extension type V2(int i) { void m() {}}
extension type V1(int i) implements V2 {}

extension E2 on V2 { void n() {}}
extension E1 on V1 {} // We don't need this one, but it doesn't hurt either.

void main() {
  var v1 = V1(42);
  v1.m(); // `V2.m` is applicable to `v1` because `V1 <: V2`.
  v1.n(); // `E2.n` is applicable to `v1` because `V1 <: V2`.
}
```

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1098 @@
...
+object in a final instance variable. The extension type thus provides an
+interface which is chosen independently of the interface of the wrapped
+object, and it provides implementations of the members of the interface of
+the extension type, and those implementations can use the members of the
+wrapped object as needed.
+
+However, even though an extension type behaves like a wrapping, the wrapper
+object will never exist at run time, and a reference whose type is the
+extension type will actually refer directly to the underlying wrapped
+object.
```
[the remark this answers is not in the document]
eernstg: Added a comment to that effect.

accepted/future-releases/extension-types/feature-specification.md
eernstg: My intention was to make it more strict, but I've changed it back to the approach where the error only occurs if _DV_ and _DV2_ are declared in different libraries. Consistency is a value on its own, in spite of the fact that it can be tempting to break it. ;-)

accepted/future-releases/extension-types/feature-specification.md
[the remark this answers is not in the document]
bwilkerson: Ok, that makes sense, and is a good reason not to use `with`. `implements` isn't great either, but I can't think of a better word at the moment, and I'm convinced that `implements` might be the least-bad option. (Although it might be worth playing around with other words, if you haven't already, such as `subtypes` or `enhances` to try to find something that doesn't already have a semantic different than what we're doing here.)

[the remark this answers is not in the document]
eernstg: We've played around with all these words for years. I'd like to see something landing firmly. But `implements` has been justified by this relationship with extension method resolution for many months now, so I think that explanation sticks.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1131 @@
...
+<code>X extends C<Y>, Y extends X</code> to
+<code>X extends Y, Y extends X</code>,
+which is an error.*
+
+An extension type declaration _DV_ named `Name` may declare one or
+more constructors. A constructor which is declared in an extension type
+declaration is also known as an _extension type constructor_.
+
+*The purpose of having an extension type constructor is that it bundles
+an approach for building an instance of the representation type of an
```
scheglov, line 816: "building an instance" -> "bundling an instance", or something like this?

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1131 @@
...
+<code>Name&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;.name(...)</code>
+is used to invoke these constructors, and the type of such an expression is
+<code>Name&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;</code>.
+
+*In short, extension type constructors appear to be very similar to
+constructors in classes, and they correspond to the situation
+where the enclosing class has a single, non-late, final instance
+variable, which is initialized according to the normal rules for
+constructors (in particular, it can occur by means of `this.id`, or in
+an initializer list, or by an initializing expression in the
```
scheglov, line 855: "initializing expression" - can it?
The grammar says just `'(' <type> <identifier> ')'` - there is no place for the initializing expression.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1131 @@
...
+this case we say that `V1` is a _superinterface_ of _DV_.
+
+If _DV_ does not include an `<interfaces>` clause then _DV_ has 
+`Object?` or `Object` as a direct superinterface, according to the subtype
+relation which was specified earlier.
+
+A compile-time error occurs if `V1` is a type name or a parameterized type
+which occurs as a superinterface in an extension type declaration _DV_, but
+`V1` does not denote an extension type, and `V1` does not denote a
+supertype of the ultimate representation type of _DV_.
```
scheglov, line 903: The `V1` is not a supertype, or the extension type erasure of `V1` is not a supertype of the extension type erasure of `DV`?

scheglov: But this would be what is written below: A compile-time error occurs if `R` is not a subtype of `S`. So, I don't understand the meaning of this paragraph.

scheglov: Hm... Does it mean maybe that we can write this?
```dart
class A {}

class B {}

extension type C(B it) implements A {}
```
But why?

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1131 @@
...
+We will omit 'declared' and 'instantiated' from the phrase when it is clear
+from the context whether we are talking about the extension type
+declaration itself, or we're talking about a particular generic
+instantiation of an extension type. *For non-generic extension type
+declarations, the representation type is the same in either case.*
+
+Let `V` be an extension type of the form
+<code>Name&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;</code>, and let
+`R` be the corresponding instantiated representation type.  If `R` is
+non-nullable then `V` is a proper subtype of `Object`, and `V` is
```
scheglov, line 749: What means "proper subtype"?

accepted/future-releases/extension-types/feature-specification.md
eernstg: Right, receiving an already fully created and initialized representation object will probably be much more common than building it in the constructor. Adjusted the text to reflect this.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Good catch, thanks! So we don't have that initializing expression any more. Adjusted the text.

accepted/future-releases/extension-types/feature-specification.md
eernstg: In the case where `V1` denotes an extension type, there's nothing new compared to the old inline class feature specification.

In the case where `V1` denotes a non-extension type, it is a situation where [this feature](https://github.com/dart-lang/language/pull/3138) is used. That feature was not part of the inline class proposal (when it landed), so that's a new thing.

With this, you can do exactly this kind of thing (note that we can only `implements` a supertype of the representation type):

```dart
class A {}

class B implements A {}

extension type C(B it) implements A {}
```

This allows the extension type to be "open" (to some extent, here we announce `A` even though we could have used `B` to make it even more open).

The point is that this kind of (open) extension type allows the representation type to work much like a "sticky" extension type:

First, we have access to all the members of `A`, so we can treat a `C` as if it had been an `A` (with respect to member invocations). Next, a `C` is assignable to a variable of type `A` (because of the subtype relationship, and similarly for a `List<C>` vs. a variable of type `List<A>`, etc.). This implies that the extension type works very much like a regular `extension` declaration. Another way to describe these behaviors is that an expression of type `C` makes it very easy to drop the extension type (that is, to _leave_ the abstraction). So you'd use it in cases where that is exactly what you want.

However, the extension members (that is, the members declared by `C` or "inherited" by `C` from some other extension type) are available on an expression of type `C` even in the case where the library that contains the member invocation does not directly import the library where `C` is declared. This makes the extension type different (and, in some situations, considerably more convenient) than the existing `extension` declarations.

Finally, again based on the subtype relationship, you can use promotion to _enter_ the abstraction very smoothly (again, you'd use this mechanism exactly when you want this): `List<A> list` can be promoted to `List<C>` (which requires a run-time type of `List<B>`, because the run-time value of `C` is `B`), and this means that each element of that list can be used as a `C` without further type tests or casts.

eernstg:
Just landed a commit that drops the special exception whereby `e.id` is an error when `id` is the representation name and `e` isn't `this`. In other words, we're dropping the lexical-scope-only rule, the representation declaration works just like any other getter now. If the developer wishes to prevent direct access (at least from other libraries), the representation name  should be private.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+as the representation type of an extension type declaration, as the on-type
+of an extension declaration, as the type in the `onPart` of a try/catch
+statement, in a type test `o is V<...>`, in a type cast `o as V<...>`, or
+as the body of a type alias. It is also allowed to create a new instance
+where one or more extension types occur as type arguments (e.g.,
+`List<V>.empty()`).*
+
+A compile-time error occurs if the type
+<code>V&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;</code>
+is not regular-bounded.
```
lrhn, line 612: So we say that it's "allowed" when it's well-formed (correct number of type arguments).
Presumably also as a raw type.

Then later we check if it's regular-bounded, and complain if it's not.

Why not say "compile-time error" first, instead of "allowed":

> Given a term of the form <Code>V\<T<sub>1</sub>, .. T<sub>s</sub>\></code>
> where `V` resolves to an extension type declaration of the form:
> ```dart
> extension type V<X1 extends B1, .. Xs extends Br>(T id) ... {
>   ... // Members.
> }
> ```
> it's a compile-time error if <Code>s &ne; r</code>.

Then we don't get into saying "allowed", without saying what happens if it's *not* allowed. Everything is allowed, if it's not a compile-time error.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+
+*In other words, such types can not be super-bounded. The reason for this
```
lrhn, line 614: can not -> cannot

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+restriction is that it is unsound to execute code in the body of `V` in
+the case where the values of the type variables do not satisfy their
+declared bounds, and those values will be obtained directly from the static
+type of the receiver in each member invocation on `V`.*
+
+A compile-time error occurs if a type parameter of an extension type
+declaration occurs in a non-covariant position in the representation type.
+
+When `s` is zero,
+<code>V&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;</code>
```
lrhn, line 624: (You can use `\<` instead of `&lt;`.)

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+simply stands for `V`, a non-generic extension type.
```
lrhn, line 625: > When `s` is zero,

This is a little unclear, `s` occurs in multiple places. But if *s* is the number of type parameters of the declaration of `V`, then it makes sense.

Maybe just add: `... zero (that is, the declaration of `V` is not generic),`

(I initially thought it was about the following instantiation, and then I wondered if `V` could be generic and `V` was a raw type then. I had to get to the next section to figure out it wasn't that.)

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+When `s` is greater than zero, a raw occurrence `V` is treated like a raw
+type: Instantiation to bound is used to obtain the omitted type arguments.
```
lrhn, line 627: (is it "instantiation to bound" or "instantiations to bounds"? Feels like it should be plural in the general case.)

We can do instantiation to bounds before type-inference, right?
All we need is the type hieararchy, not any of those pesky members.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+*Note that this may yield a super-bounded type, which is then a
+compile-time error.*
+
+We say that the static type of said variable, parameter, etc.
```
lrhn, line 631: Which variable, parameter, etc?

If it is the ones in the example above, they shouldn't be referenced from normative text.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+_is the extension type_
+<code>V&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;</code>,
+and that its static type _is an extension type_.
+
+It is a compile-time error if `await e` occurs, and the static type of
+`e` is an extension type which is not a subtype of `Future<T>` or
+`FutureOr<T>` for any `T`.
+
+A compile-time error occurs if an extension type declares a member whose
+name is declared by `Object` as well.
```
lrhn, line 641: -> is declared as an *instance member* by `Object` as well.

(`Object` has static `hash`-functions too.)

Do we allow same base-name, `set hashCode`? (Shouldn't be a problem, just asking.)

This precludes both static and instance members with those names, which is probably the right choice.

Also need to disallow the representation object field being named like that.
Maybe also constructors (but not actually sure, those were always quirky).

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+method `toString`, not the extension type method. Also, when we make this an
+error for now, we have the option to allow it, perhaps with some
+restrictions, in a future version of Dart.*
+
+A compile-time error occurs if an extension type is used as a
+superinterface of a class or a mixin, or if an extension type is used to
+derive a mixin.
+
+*In other words, an extension type cannot occur as a superinterface in an
+`extends`, `with`, `implements`, or `on` clause of a class or mixin.  On
```
lrhn, line 657: or enum

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+arguments <code>T<sub>1</sub>, ..., T<sub>s</sub></code>, with the actual
+argument part `args`.
+
+Similarly, `e.m` is treated an invocation of the extension type member `m`
+on the receiver `e` according to the extension type declaration `Name` and
+with the actual type arguments <code>T<sub>1</sub>, ...,
+T<sub>s</sub></code> and no actual argument part.
+
+*Setter invocations are treated as invocations of methods with a
+single argument.*
```
lrhn, line 675: And operators are treated like funny-named methods, `e1 + e2` works as if `+` was a method name and was invoked as `e1.+(e2)`.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+*The purpose of having an extension type constructor is that it bundles an
+approach for receiving or building an instance of the representation type
+of an extension type declaration _DV_ with _DV_ itself, and creating an
+expression whose static type is the extension type and whose value at run
+time is said representation object. Extension type constructor bodies can
+also be used to verify that the representation object satisfies the
+requirements for having that extension type.*
+
+The `<representationDeclaration>` works as a constructor. The optional
+`('.' <identifier>)` in the grammar is used to declare this constructor
```
lrhn, line 825: Should this be `('.' (<identifier> | 'new'))?` ?

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+with a name of the form `<identifier> '.' <identifier>` *(at times
+described as a "named constructor")*. It is a constant constructor: If `e`
```
lrhn, line 827: Is it a constant constructor, or do you have to write `const` in front?

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+is a constant expression and `V(e)` is not an error, then `V(e)` is a
+constant expression. Other constructors may be declared `const` or not,
+following the normal rules for constant constructors.
+
+A compile-time error occurs if an extension type constructor includes a
+superinitializer. *That is, a term of the form `super(...)` or
+`super.id(...)` as the last element of the initializer list.*
```
lrhn, line 834: or `super.new(...)`. 

We should have `<constructorNameOpt> ::= '.' <identifier> | '.' 'new' | ;` and just say `'super' <constructorNameOpt> <arguments>'`.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+
+A compile-time error occurs if an extension type constructor declares a
+super parameter. *For instance, `Name(super.x);`.*
+
+*In the body of a generative extension type constructor, the static type of
```
lrhn, line 839: non-redirecting

(but that is implicit, because a redirecting generative constructor has no body.)

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+An instance creation expression of the form
+<code>Name&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;(...)</code>
+or
+<code>Name&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;.name(...)</code>
+is used to invoke these constructors, and the type of such an expression is
+<code>Name&lt;T<sub>1</sub>, .. T<sub>s</sub>&gt;</code>.
+
+*In short, extension type constructors appear to be very similar to
+constructors in classes, and they correspond to the situation where the
+enclosing class has a single, non-late, final instance variable, which is
```
lrhn, line 853: I have seen no specification of this final instance variable yet. It's mentioned only ones in the motivation. Should it be specified before there?

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+initialized according to the normal rules for constructors (in particular,
+it can occur by means of `this.id`, or in an initializer list, but it is an
+error if the initialization does not occur at all).*
+
+An extension type `V` used as an expression (*a type literal*) is allowed
+and has static type `Type`.
```
lrhn, line 859: Mention that its runtime value is the type object of its fully expanded representation type?

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+relation which was specified earlier.
+
+A compile-time error occurs if `V1` is a type name or a parameterized type
+which occurs as a superinterface in an extension type declaration _DV_, but
+`V1` does not denote an extension type, and `V1` does not denote a
+supertype of the ultimate representation type of _DV_.
+
+A compile-time error occurs if any direct or indirect superinterface
+of _DV_ is the type `Name` or a type of the form `Name<...>`. *As
+usual, subtype cycles are not allowed.*
```
lrhn, line 903: How about:
```dart
extension type A<T>(Object? _) {}
extension type C<T>(List<Never> _) implements A<List<C<T>>> {}
```
The last line has an infinite expansion, but only in the superinterface hieararchy.

(I have no idea if it works or not.)

lrhn (commented):
Got to line 915.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
+# Extension Types
+
+Author: eernst@google.com
+
+Status: Accepted.
+
```
mit-mit, line 6: Can we start with a motivating example, similar to https://github.com/dart-lang/language/blob/main/accepted/future-releases/inline-classes/feature-specification.md#motivation ?

lrhn: (I'm slighlty split here. Motivating examples are nice for getting things accepted, but once accepted, I'd like a tight, sharp specification the of what the feature does, or discussions about alternatives. Just the facts.

Could we split it in two files? Motivation/discussion and specification. If the specification doesn't work without the other text, then it's already non self-contained enough.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+there may also be some extension members).
+
+Hence, using a supertype as the static type allows us to see only a
+subset of the members, possibly with a more general member signature. Using
+an extension type allows us to _replace_ the set of members, with
+subsetting as a special case.
+
+This functionality is entirely static. Invocation of an extension type
+member is resolved at compile-time, based on the static type of the
+receiver.
```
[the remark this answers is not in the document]
eernstg: Done.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+object, and it provides implementations of the members of the interface of
+the extension type, and those implementations can use the members of the
+wrapped object as needed.
+
+However, even though an extension type behaves like a wrapping, the wrapper
+object will never exist at run time, and a reference whose type is the
+extension type will actually refer directly to the underlying "wrapped"
+object. This fact also determines the behavior of `as` and `is`: Those
+operations will refer to the run-time type of the representation object,
+and the run-time value of the extension type is the representation type.
```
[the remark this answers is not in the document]
eernstg: The specification uses 'actual value', but that's less guessable. Done.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+```dart
+extension type IdNumber(int i) {
+  // Assume that it makes sense to compare ID numbers
+  // because they are allocated with increasing values,
+  // so "smaller" means "older".
+  operator <(IdNumber other) => i < other.i;
+
+  // Assume that we can verify an ID number relative to
+  // `Some parameters`, filtering out some fake ID numbers.
+  bool verify(Some parameters) => ...;
```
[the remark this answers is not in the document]
eernstg: Done.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+interaction as in a regular class.
+
+A reference to the representation object typed by its run-time type or a
+supertype thereof (that is, typed by a "normal" type for the
+representation) is available as a declared name: The extension type
+declares the name and type of the representation in a way which is a
+special case of the [primary constructor proposal][].
+In the body of the extension type the representation object is in scope,
+with the declared name and type, as if it had been a final instance
+variable in a class.
```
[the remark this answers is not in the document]
eernstg: Right. Done.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+extension type methods would be highly confusing and error prone. For
+example, collections would still call the instance operator `==`, not the
+extension type operator, and string interpolation would call the instance
+method `toString`, not the extension type method. Also, when we make this an
+error for now, we have the option to allow it, perhaps with some
+restrictions, in a future version of Dart.*
+
+A compile-time error occurs if an extension type is used as a
+superinterface of a class or a mixin, or if an extension type is used to
+derive a mixin.
```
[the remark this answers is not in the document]
eernstg: Right. Modified to talk explicitly about roles played in a mixin application.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+or a member named `noSuchMethod` or `toString`. The rationale is that these
+extension type methods would be highly confusing and error prone. For
+example, collections would still call the instance operator `==`, not the
+extension type operator, and string interpolation would call the instance
+method `toString`, not the extension type method. Also, when we make this an
+error for now, we have the option to allow it, perhaps with some
+restrictions, in a future version of Dart.*
+
+A compile-time error occurs if an extension type is used as a
+superinterface of a class or a mixin, or if an extension type is used to
```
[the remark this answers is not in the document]
eernstg: Done.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+instance variables.
+
+The _name of the representation_ in an extension type declaration with a
+representation declaration of the form `(T id)` is the identifier `id`, and
+the _type of the representation_ is `T`.
+
+*There are no special rules for static members in extension types. They can
+be declared and called or torn off as usual, e.g.,
+`AnExtensionType.myStaticMethod(42)`.*
+
```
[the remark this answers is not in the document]
eernstg: Added a rule to that effect.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+
+## Extension Type Member Invocations
+
+This document needs to refer to extension type method invocations including
+each part that determines the static analysis and semantics of this
+invocation, so we will use a standardized phrase and talk about: An
+invocation of the extension type member `m` on the receiver `e` according
+to the extension type declaration `V` and with the actual type arguments
+<code>T<sub>1</sub>, ..., T<sub>s</sub></code>.
```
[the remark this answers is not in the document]
eernstg: Changed '`V` and with the actual type arguments' to '`V` with the actual type arguments'.

The following paragraph explicitly mentions that an invocation of `m` may have an `<argumentPart>` and that argument part may include actual type arguments to `m`.

I think that is enough to disambiguate the text.

This section introduces that long phrase ("an invocation of the extension type member `m` ..."), and the specification of the static analysis of such invocations as well as the dynamic semantics are given in other sections.

So at this point we just need to refer to the receiver in a way that fits the static analysis (where `e` is a good match) and the dynamic semantics (where we use `o` to talk about the actual receiver object, but we obtain by evaluating `e`). I think that works.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+(*so we will not specify that case explicitly*).
+
+*We need to mention all these elements together, because each of them plays
+a role in the static analysis and the dynamic semantics of the invocation.
+There is no syntactic representation in the language for this concept,
+because the same extension type method invocation can have many different
+syntactic forms, and both `V` and
+<code>T<sub>1</sub>, ..., T<sub>s</sub></code> are implicit in the actual
+syntax.*
+
```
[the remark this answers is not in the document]
eernstg: > Could we introduce a meta-notation for referencing members of an extension type, even if it's not one you can write directly

Thas was _exactly_ what I did in some earlier proposals, and reviewers preferred the long phrase in natural language instead.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+superinterface `V` that has a member named `n`. In both cases,
+_the member declaration named `n` that DV has_ is said declaration.
+
+*For a declaration in an extension type, this definition is unambiguous for
+an extension type that has no compile-time errors, because name clashes
+must be resolved by `V`. If the declaration is from a superinterface which
+is not an extension type then it is handled specially, and we do not need
+to have a unique declaration of `n` "that _DV_ has".*
+
+Consider an invocation of the extension type member `m` on the receiver `e`
```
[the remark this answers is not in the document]
eernstg: Done.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+
+*Note that it is known that
+<code>V&lt;T<sub>1</sub>, ..., T<sub>s</sub>&gt;</code>
+has no compile-time errors. In particular, the number of actual type
+arguments is correct, and it is a regular-bounded type,
+and the static type of `e` is a subtype of
+<code>V&lt;T<sub>1</sub>, ..., T<sub>s</sub>&gt;</code>,
+or a subtype of the corresponding instantiated representation type
+(defined below). This is required when we decide that a given
+expression is an extension type member invocation.*
```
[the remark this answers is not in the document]
eernstg: Added a few words to indicate that these properties are indeed enforced earlier. The point here is that those checks have already been performed, not that they must have been performed as the most recent step in static analysis.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+
+If the name of `m` is a name in the interface of `Object` (*that is,
+`toString`, `==`, etc.*), the static analysis of the invocation is
```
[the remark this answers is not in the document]
eernstg: Done.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+treated as an ordinary instance member invocation on a receiver of
+type `Object` and with the same `args`, if any.
```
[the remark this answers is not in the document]
eernstg: Done.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+
+Otherwise, a compile-time error occurs if `V` does not have a member
+named `m`.
+
+If `V` has a member named `m` which is declared in a non-extension type
+superinterface `S` and not redeclared by any extension type superinterfaces
```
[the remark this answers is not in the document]
eernstg: > Here `E` has a member `foo` which is declared in non-extension super-interface `_A`

The intention is that we'd focus on a member named `m` which is in the interface of a superinterface `A` (direct or indirect) and not redeclared by any of the extension types on a superinterface path to `A`. In other words, this member `m` is _purely_ an instance member in the interface of the non-extension types that are superinterfaces.

It's a compile-time error if there is an extension type superinterface that has a member named `m` which is declared by an extension type as well as an `m` which is in the interface of a non-extension type. (So we know that it's not the case where `EB` has a `foo`.) Also, there is no notion of extension type members winning over non-extension type members or vice versa. If there is a conflict like this among superinterfaces then the extension type declaration must redeclare `m`. But it's OK that different non-extension type members have different signatures, as long as we can compute a combined member signature.

We can't say 'immediate' because we need to remember that a given `m` that a superinterface _has_ was originally introduced by a non-extension type or by an extension type. So we need to talk about members that a possibly indirect superinterface has, and say that this member cannot be redeclared by any of the extension type superinterfaces on the path to that non-extension superinterface.

We just need to ensure that (1) it is an error (unconditionally) if the extension type has a member which is declared by an extension type superinterface and a member with the same name which is present in the interface of a non-extension type superinterface, but (2) we can handle a _purely_ non-extension type set of members, and (3) we can handle a purely extension-type member (but we must have exactly one declaration which isn't redeclared).

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+expression is an extension type member invocation.*
+
+If the name of `m` is a name in the interface of `Object` (*that is,
+`toString`, `==`, etc.*), the static analysis of the invocation is
+treated as an ordinary instance member invocation on a receiver of
+type `Object` and with the same `args`, if any.
+
+Otherwise, a compile-time error occurs if `V` does not have a member
+named `m`.
+
```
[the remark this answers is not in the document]
eernstg: No, that's a bunch of cases where we do not have an error, and they are described in line 499++. Made that explicit.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+Otherwise, let _Dm_ be the declaration of `m` that `V` has.
+
+If _Dm_ is a getter declaration with return type `R` then the static
+type of the invocation is
+<code>[T<sub>1</sub>/X<sub>1</sub> .. T<sub>s</sub>/X<sub>s</sub>]R</code>.
+
+If _Dm_ is a method with function type `F`, and `args` is omitted, the
+invocation has static type
+<code>[T<sub>1</sub>/X<sub>1</sub> .. T<sub>s</sub>/X<sub>s</sub>]F</code>.
+*This is an extension type method tear-off.*
```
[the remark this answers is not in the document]
eernstg: Arghh, right!! Done. And woohoo!

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+
+If _Dm_ is a method with function type `F`, and `args` exists, the static
+analysis of the extension type member invocation is the same as that of an
+invocation with argument part `args` of a function with type
+<code>[T<sub>1</sub>/X<sub>1</sub> .. T<sub>s</sub>/X<sub>s</sub>]F</code>.
+*This determines the compile-time errors, if any, and it determines the
+type of the invocation as a whole.*
```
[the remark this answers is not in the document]
eernstg: That's true. But, for now, let us try to get to something which is sufficiently precise to make it useful during implementation.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+<code>X<sub>1</sub>, ..., X<sub>s</sub></code>.
+
+Let _Dm_ be the declaration named `m` that `V` has.
+
+Evaluation of this invocation proceeds by evaluating `e` to an object
+`o`.
+
+Then, if `args` is omitted and _Dm_ is a getter, execute the body of
+said getter in an environment where `this` and the name of the
+representation are bound to `o`, and the type variables of `V` are
```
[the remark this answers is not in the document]
eernstg: Done.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+<code>[T<sub>1</sub>/X<sub>1</sub> .. T<sub>s</sub>/X<sub>s</sub>]F</code>.
+*This determines the compile-time errors, if any, and it determines the
+type of the invocation as a whole.*
+
+
+### Dynamic Semantics of an Extension Type Member Invocation
+
+Consider an invocation of the extension type member `m` on the receiver `e`
+according to the extension type declaration `V` and with actual type
+arguments <code>T<sub>1</sub>, ..., T<sub>s</sub></code>. If the invocation
```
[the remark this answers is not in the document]
eernstg: We have this long natural language phrase that denotes all the parts that make up an extension type member invocation, and that natural language phrase is used elsewhere: This section specifies the dynamic semantics of the long phrase, and those places 'elsewhere' are the 'call sites' for the long phrase.

Look for 'according to' and you'll find the places where those actual type arguments are specified.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+then the invocation completes throwing the same object and stack
+trace.
+
+Otherwise, if `args` is omitted and _Dm_ is a method, the invocation
+evaluates to a closurization of _Dm_ where
+`this` and the name of the representation are bound to `o`, and the
+type variables of `V` are bound to the actual values of
+<code>T<sub>1</sub>, .. T<sub>s</sub></code>.
+The operator `==` of the closurization returns true if and only if the
+operand is the same object.
```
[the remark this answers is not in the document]
eernstg: It follows: If two function objects are the same object then they are equal (we certainly can't have a usable `==` if this is not true). We do not require that the function objects must compare equal if they are, say, tear-offs of the same extension type member with the same representation object.

We cannot require this equality to hold in all cases, because it would be wrong (if they have different type arguments then they certainly shouldn't be `==`), but, just like extension methods, we don't require the equality to hold even in the case where the type arguments are the same.

So the main idea here is: Just like extension methods.

I do not think this prevents optimizations where certain tear-offs are the same ("constant") object even though they are torn off at two different occasions. For example, an extension type with no type arguments and an extension type member that doesn't use `this` could presumably be compiled to a top-level function and reused for all tear-offs of that member.

accepted/future-releases/extension-types/feature-specification.md
eernstg: I don't know where it says "allowed" with that meaning. The only "allowed" I could spot now has to do with the ability to use `C<T1 .. Tk>` as a type, and that doesn't imply anything about that being a correct type. 'Regular-bounded' implies that the number of type arguments is the same as the number of type parameters, and the list of type arguments satisfies the list of bounds (it doesn't make sense to say that each actual type argument satisfies its bound, because some type parameters can be used in some bounds).

accepted/future-releases/extension-types/feature-specification.md
eernstg: Done.

accepted/future-releases/extension-types/feature-specification.md
eernstg: OK, thanks!

accepted/future-releases/extension-types/feature-specification.md
eernstg: Done.

accepted/future-releases/extension-types/feature-specification.md
eernstg: The language specification consistently uses 'instantiation to bound' about a type. 

Yes, we can do that without type inference. The implementations do have a notion of partial instantiation to bound which is used when some type arguments are chosen based on additional information, but that hasn't been specified.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Following extension methods, I made this an error when it is the same basename. Can relax later.

Added commentary about the representation name, repeating that it counts as an instance variable declaration during static analysis.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Done.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Done.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Yes, thanks!

accepted/future-releases/extension-types/feature-specification.md
eernstg: We haven't decided. I've mentioned that choice that seemed to have broad support at some point: We make the constructor constant in every case, because that is possible. We can _allow_ `const` to be specified explicitly later on, if that serves a purpose.

If the language team wants something else then we will just change it.

I added a new subsection in the 'Discussion' section about this, to keep it in mind.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Done now!

accepted/future-releases/extension-types/feature-specification.md
eernstg: I don't think that's a problem, we do not ever iterate over all supertypes, we only traverse superinterface graph paths, and they only have an edge when we have a syntactic relation expressed using a clause like `extends`, `implements`, `on`, or `with`.

One reason why it does not seem to be dangerous is that we can have this, no problems:

```dart
class A<T> {
  final Object? rep;
  A(this.rep);
}

class C<T> implements A<List<C<T>>> {
  final List<Never> rep;
  C(this.rep);
}
```

eernstg:
Great feedback, @lrhn! Changed a bunch of things and uploaded.
Didn't do anything about the introductory example, yet.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -0,0 +1,1212 @@
...
+combined member signatures of all declarations of `m` in the direct
+superinterfaces of `V`. 
+
+*In other words, members "inherited" from non-extension type
+superinterfaces are invoked as normal class instance members, as if we
+could "see through the veil" that is the extension type and call members of
+the representation type which have been unveiled by including `S` as a
+superinterface, directly or indirectly.*
+
+Otherwise, let _Dm_ be the declaration of `m` that `V` has.
```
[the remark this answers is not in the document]
eernstg: Added general language to specify that these substitutions take place as part of the lookup.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Of course! Done.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Ah, reading this one more time, I could see the problem. Revised.

eernstg:
Further revisions have been made now. In particular, a lot of things became simpler by introducing the notion of 'an extension type member' and 'a non-extension type member' maintains the distinction that the declaration of the given member is located in an extension type declaration respectively it is in the interface of a non-extension type superinterface.

So we don't need to re-explain that we need to look up through the superinterface graph in order to see whether any given member came from an extension type or from a non-extension type.

eernstg:
Just updated the grammar to support `const` on the primary constructor, and the rules to say that the constructor is constant if and only if `const` is present in the declaration header.

PR #3218, opened 2023-07-17
Adjust generator element type, cf. dart-lang/language#3148

eernstg:
In issue #3148, it is reported that there is a soundness issue with the computation of the element type of a generator function whose declared return type is `FutureOr<Iterable<int>?>`, and similar issues with other union types.

This PR changes the definition of this element type such that return types are simplified to `Iterable<...>` or `Stream<...>` before the element type is computed, if possible. `Object` and top types still yield the element type `dynamic`.

Moreover, this PR also adjusts the specification of the dynamic type of the returned iterable/stream accordingly.

NB: This PR introduces null-safety related material into the existing language specification. We have done this in other cases as well. In general, we're aiming to integrate all the null-safety related material into the language specification, piece by piece, so there is no violation of current practice in doing that with this PR as well.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -11688,17 +11688,17 @@ as follows, using the first applicable case:
 \item If $T$ is \code{$X$\,\&\,$S$}
   for some type variable $X$ and type $S$ then
   \begin{itemize}
-  \item if $S$ derives future type $U$
+  \item if $S$ derives a future type $U$
     then \DefEquals{\flatten{T}}{\code{\flatten{U}}}.
   \item otherwise,
     \DefEquals{\flatten{T}}{\code{\flatten{X}}}.
   \end{itemize}
 
-\item If $T$ derives future type \code{Future<$S$>}
+\item If $T$ derives a future type \code{Future<$S$>}
   or \code{FutureOr<$S$>}
   then \DefEquals{\flatten{T}}{S}.
 
-\item If $T$ derives future type \code{Future<$S$>?}\ or
+\item If $T$ derives a future type \code{Future<$S$>?}\ or
   \code{FutureOr<$S$>?}\ then \DefEquals{\flatten{T}}{\code{$S$?}}.
 
 \item Otherwise, \DefEquals{\flatten{T}}{T}.
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -54,6 +54,8 @@
 %   been eliminated (it is spelled out each time which modifiers must be
 %   present or not present). The concept is confusing now where a variable can
 %   be late and final, and assignments to it are allowed by the static analysis.
+% - Change the definition of the 'element type of a generator function', due to
+%   soundness issue with the current definition.
 %
 % Mar 2023
 % - Clarify how line breaks are handled in a multi-line string literal. Rename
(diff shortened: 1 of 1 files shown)


eernstg:
Friendly ping. Note that the current specification is unsound (cf. https://github.com/dart-lang/language/issues/3148#issuecomment-1607381741).

specification/dartLangSpec.tex
```diff
@@ -2073,30 +2073,38 @@ \section{Functions}
...
+For example, the union-free type of \code{FutureOr<int?>?} is \code{int}.%
+}
+
+\LMHash{}%
+We define the
+\IndexCustom{element type of a generator function}{%
+    function!generator!element type}
+$f$ as follows:
+%
+Let $S$ be the union-free type of the declared return type of $f$.
```
leafpetersen, line 2092: Can we just reject programs where the declared return type is not either `Iterable` (`Stream` respectively) or a top type?

leafpetersen:
We will need a plan for making sure whatever specification we end up gets tested and implemented.

leafpetersen:
Also, how does this interact with https://github.com/dart-lang/language/pull/3151/files ?

specification/dartLangSpec.tex
```diff
@@ -2073,30 +2073,38 @@ \section{Functions}
 a function marked \code{\SYNC*} or \code{\ASYNC*} is \VOID.
 
 \LMHash{}%
-We define the notion of the
-\IndexCustom{element type of a generator}{function!generator!element type}
-as follows:
+We define the
+\Index{union-free type}
```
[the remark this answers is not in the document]
eernstg: Right; we might then want to change 'the future type of a type' similarly. Did both.

specification/dartLangSpec.tex
eernstg: If think this might reject some valid programs. It's a long shot ("come on, who would write a program like that?"), but the following example could actually occur in some context where the developer doesn't have much of a choice (example from [here](https://github.com/dart-lang/language/issues/3148#issuecomment-1649253527)):

```dart
class A {
  FutureOr<Iterable<int>?> m() => null;
}

class B extends A {
  FutureOr<Iterable<int>?> m() sync* { yield 1; yield 2; }
}

class C extends B {
  FutureOr<Iterable<int>?> m() => Future.value();
}
```

eernstg:
> We will need a plan

There is a co19 task in the pipeline, cf. https://github.com/dart-lang/co19/issues/2142.

sgrekhov:
> There is a co19 task in the pipeline, cf. [dart-lang/co19#2142](https://github.com/dart-lang/co19/issues/2142).

It's a https://github.com/dart-lang/co19/issues/2139

eernstg:
> how does this interact with https://github.com/dart-lang/language/pull/3151/files?

If we do not introduce new compile-time errors (for declarations like `Iterable<int>? f() sync* {...}`) then I believe this PR serves as a bug fix in the definition of 'the element type of a generator function'. In that case there wouldn't be any interactions, except that 'the element type of a generator function' is now less buggy.

specification/dartLangSpec.tex
leafpetersen: Ack.

leafpetersen (approved):

specification/dartLangSpec.tex
```diff
@@ -2073,30 +2075,40 @@ \section{Functions}
...
+We define the
+\Index{union-free type derived from}
+a type $T$ as follows:
+If $T$ is of the form \code{$S$?}\ or the form \code{FutureOr<$S$>}
+then the union-free type derived from $T$ is
+the union-free type derived from $S$.
+Otherwise, the union-free type derived from $T$ is $T$.
+\commentary{%
+For example, the union-free type derived from
+\code{FutureOr<int?>?} is \code{int}.%
```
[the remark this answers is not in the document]
eernstg: > Could it have a name that is more specialized

Yes, I think you're right, it should have a more specialized name.

Issue #1934, opened 2021-10-21
How does the macro API evolve as the language changes?


munificent:
One of the major challenges of exposing a programmatic API for working with Dart syntax is that new language syntax features can force changes to that API. If we aren't careful, those could be breaking changes, which would break macros. How do we handle this?

Levi-Lesches:
I feel like this would be a good place to discuss the pros and cons of using a `Code`-like API and outputing code as a string. From https://github.com/dart-lang/language/issues/1482#issuecomment-896306774: 

> 1. Simply, an API would have to be designed and implemented (beyond the prototype). Outputting strings to a file is simple
> 2. The arguments to a constructor like [`Declaration`](https://jakemac53.github.io/macro_prototype/doc/api/definition/Declaration-class.html) are strings anyway
> 3. The Dart team would have to update the API with new language features, syntax, and grammar before they can be used
> 4. An API cannot be type-safe if types are statically unknown, so they're just as unsafe as strings
> 5. An API has to be expressive enough to cover every single use-case. It essentially has to duplicate Dart
> 6. An API has to be easy for all developers to learn to use. Dart users already knows basic Dart
> 7. An API has to be easy to maintain. Adding one line of code to generate shouldn't mean ten more lines and deeper indentation
> 8. An API has to communicate what code it is generating. By just writing Dart code as a string, WYSIWYG
> 9. An API has to generate all its code as a string in the end anyway. Why not remove a potential layer of problems?
> 10. If generated code can be inspected by devs, it has to look good and have claer names and structure. That might mean more than just outputting functional code and running it through `dart format`.
> 
> Using an API will most definitely lead to cases where the code isn't coming out the way it was intended to, either because of fundamental limitations or user error. And when it would work, the macro source code would be hard to read and follow. Since macros generate separate files, why not have the contents of those files _be_ the source code of the macros?

jakemac53:
Finalizing the approach here https://github.com/dart-lang/language/issues/1934.

PR #3219, opened 2023-07-17
solidify and clarify the API versioning strategy for macros

jakemac53:
Closes https://github.com/dart-lang/language/issues/1934

### DIFF
--- a/working/macros/feature-specification.md
+++ b/working/macros/feature-specification.md
@@ -1328,14 +1328,62 @@ are compiled with exactly the same version of the API that macros are compiled
 with. This ensures the communication protocol between macros and the SDK
 binaries are compatible.
 
-This package will have tight upper bound SDK constraints, and will need to be
-released any time a new minor release is published. It is expected that most of
-these releases would be patch releases, which only increase the max SDK
-constraint.
+This package will have tight upper bound SDK constraints, constrained to the
+minor release versions instead of major release versions.
 
-Any time the API changes (breaking or otherwise), the min SDK constraint of the
-package must be updated to match the version of the SDK where the API changed.
-This must always correspond to a minor SDK version.
+The release/versioning strategy for the package is as follows:
+
+- We will not allow any changes to the macro APIs in patch releases of the SDK,
+  even non-breaking changes. The package restricts itself to minor releases and
+  not patch releases of the SDK, so these changes would silently be visible to
+  users in a way that wasn't versioned through the package.
+
+- When a new version of the Dart SDK is released which **does not have any**
+  changes to the macro API, then we will do a patch release of this package
+  which simply expands the SDK upper bound to include that version (specifically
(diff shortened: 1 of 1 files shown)


working/macros/feature-specification.md
```diff
@@ -1300,28 +1300,68 @@ given library. This will be allowed through the library introspection class,
...
+* Give early and actionable error messages to consumers of macros when a macro
+  they are using does not support a new language feature.
+* Enable as much forwards/backwards compatibility as possible.
+
+#### Solution: Ship macro APIs as a Pub package
+
+The implementation of this package will always come from the SDK, through a
+`dart:_macros` library, which will be exported by this package. The
+`dart:_macros` library will be blocked from being imported or exported by any
+library other than other `dart:` libraries.
```
jakemac53, line 1324:
```suggestion
library other than other `dart:` libraries and this package.
```

jakemac53:
cc @leafpetersen you may want to take a look also

working/macros/feature-specification.md
```diff
@@ -1300,28 +1300,68 @@ given library. This will be allowed through the library introspection class,
...
+
+We use a `dart:` library here to ensure that the binaries shipped with the SDK
+are compiled with exactly the same version of the API that macros are compiled
+with. This ensures the communication protocol between macros and the SDK
+binaries are compatible.
+
+This package will have tight upper bound SDK constraints, and will need to be
+released any time a new minor release is published. It is expected that most of
+these releases would be patch releases, which only increase the max SDK
+constraint.
```
munificent, line 1334: This is a little opaque if you are not steeped in package versioning. Maybe walk through two scenarios:

* When a new version of the Dart SDK is released which does not have a breaking change in the macro API, then we will release a patch version of the macro package which simply expands the SDK upper bound to include that version. Since this is just a patch release, all existing packages that depend on the macro package support it already, and no work is required on macro authors' part to work with the new Dart SDK.

* When a new version of the Dart SDK does include a breaking change to the macro API, then we will release a new version of the macro package with an increased *major* version number. By default, existing packages containing macros will not accept that version of the macro package and thus will not work with the new Dart SDK.

    Authors of packages containing macros will need to test to see if their macro is compatible with the latest macro API. If so, they can ship a new patch version of their package with a constraint on the macro package that includes the new major version as well as the previous major version it's already known to work with. If their package is broken by the macro API change, then, they will fix their macro and ship a new version of their package with a dependency on the macro package that *only* allows the latest major versions.

(Walking through an example with concrete version numbers and package names would be even better, but you get the idea.)

munificent (approved):
LGTM!

working/macros/feature-specification.md
jakemac53: Updated

leafpetersen (approved):

PR #3256, opened 2023-08-04
add InheritedWidget macro example and fix up FunctionalWidget

jakemac53:
This would reduce the "Scoped Data Propagation" example [here](https://www.jetpackcompose.app/compare-declarative-frameworks/JetpackCompose-vs-SwiftUI-vs-Flutter) to the following (when used with FunctionalWidget):

```dart
@InheritedWidget()
class CustomInheritedWidget {
  final String data;
}

@FunctionalWidget()
Parent _parent(BuildContext context, String data) => CustomInheritedWidget(
      data: data,
      child: Intermediate(),
    );

@FunctionalWidget()
Intermediate _intermediate(BuildContext context) => Child();

@FunctionalWidget()
Child _child(BuildContext context) {
  final data = CustomInheritedWidget.of(context).data;
  return Text("Received data: $data");
}

// Usage
Parent(data: "Some data")
```

I also cleaned up FunctinalWidget a bit while I was at it.

cc @goderbauer wdyt about a macro like this?

### DIFF
--- a/working/macros/example/lib/inherited_widget.dart
+++ b/working/macros/example/lib/inherited_widget.dart
@@ -1,0 +1,97 @@
+// Copyright (c) 2023, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// There is no public API exposed yet, the in-progress API lives here.
+import 'package:_fe_analyzer_shared/src/macros/api.dart';
+
+/// A macro that annotates a class and turns it into an inherited widget.
+///
+/// This will fill in any "holes" that do not have custom implementations,
+/// specifically the following items will be added if they don't exist:
+///
+/// - Make the class extend `InheritedWidget`.
+/// - Add a constructor that will initialize any fields that are defined, and
+///   take `key` and `child` parameters which it forwards to the super
+///   constructor.
+/// - Add static `of` and `maybeOf` methods which take a build context and
+///   return an instance of this class using `dependOnIheritedWidgetOfExactType`.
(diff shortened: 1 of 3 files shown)


jakemac53:
also cc @mit-mit

working/macros/example/lib/inherited_widget.dart
```diff
@@ -0,0 +1,97 @@
...
+
+  @override
+  Future<void> buildDeclarationsForClass(IntrospectableClassDeclaration clazz, MemberDeclarationBuilder builder) async {
+    final fields = await builder.fieldsOf(clazz);
+    final methods = await builder.methodsOf(clazz);
+    if (!methods.any((method) => method is ConstructorDeclaration)) {
+      builder.declareInType(DeclarationCode.fromParts([
+            'const ${clazz.identifier.name}(',
+            '{',
+            for (var field in fields)...[
```
munificent, line 39: Nit: Looks like this hasn't been run through dart format?

(I noticed the lack of space before `...`. This is what working on the formatter has done to me. :) )

...Oh, I just realized you can't because of the `macro` keywords. :(

munificent (approved):

working/macros/example/lib/inherited_widget.dart
jakemac53: Yeah because dart format doesn't allow the `macro` keyword so its a pain to use it lol.

jakemac53: Fwiw, we should have a larger discussion, but the current way the formatter works with experiments is causing other problems with deprecating old experiments. We should re-evaluate the whole thing and come up with a way to explicitly enable any experiments imo.

Issue #3261, opened 2023-08-10
[extension-types] Adjust the rule for allowed non-extension type superinterfaces


eernstg:
Thanks to @johnniwinther for raising this issue!

The extension type feature specification has the following [rule](https://github.com/dart-lang/language/blob/main/accepted/future-releases/extension-types/feature-specification.md#:~:text=was%20specified%20earlier.-,A%20compile%2Dtime%20error%20occurs,-if%20V1%20is):

> A compile-time error occurs if `V1` is a type name or a parameterized type which occurs as a superinterface in an extension type declaration _DV_, but `V1` does not denote an extension type, and `V1` does not denote a supertype of the extension type erasure of the representation type of _DV_.

The motivation for including the erasure is that it provides some added flexibility. However, this issue recommends that the rule is adjusted as follows:

> A compile-time error occurs if `V1` is a type name or a parameterized type which occurs as a superinterface in an extension type declaration _DV_, but `V1` does not denote an extension type, and `V1` does not denote a supertype of the representation type of _DV_.

Here is some background, and an example illustrating the considerations about this topic that were raised in this discussion:

```dart
extension type FancyList(List<int> it) implements List<int> {
  ... // Members making the `List<int>` more convenient for some purpose.
}
```

In this example, `FancyList` is positioned as an extended version of `List<int>`: It is assignable to `List<int>` and may occur due to promotion from some supertype of `List<int>`, it's just providing those extra members.

The current specification rule allows for the representation type to use a similar type:

```dart
extension type FancyInt(int it) implements int {
  ... // Convenience members on int.
}

extension type FancyList2(List<FancyInt> it) implements List<int> {
  ... // Like before, but more convenient for working on the elements.
}
```

This allows the body of `FancyList` to use the convenience members of `FancyInt` on the elements of the list. It is allowed because the extension type erasure of `List<FancyInt>` is `List<int>`, which is a subtype of `List<int>`.

However, we couldn't use the following variant with the rule as stated today:

```dart
// `FancyInt` unchanged.

extension type FancyList3(List<FancyInt> it) implements List<FancyInt> { // Error, using the current rule!
  ... // Like before.
}
```

The error occurs because the extension type erasure of `List<FancyInt>` (that is, `List<int>`) isn't a subtype of `List<FancyInt>`, so we can't have the latter in the `implements` clause.

[...]

PR #3262, opened 2023-08-10
Adjust the rule about `implements NonExtensionType`

eernstg:
This PR changes the requirement on a non-extension type `T` that occurs in `implements ... T ...` in an extension type declaration: It must be a supertype of the representation type. Before this change it had to be a supertype of the erasure of the representation type, but that rule has some counter-intuitive and conceptually dubious consequences.

Cf. https://github.com/dart-lang/language/issues/3261 for more details.

### DIFF
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -1072,10 +1072,9 @@ If _DV_ does not include an `<interfaces>` clause then _DV_ has
 relation which was specified earlier.
 
 A compile-time error occurs if `V1` is a type name or a parameterized type
-which occurs as a superinterface in an extension type declaration _DV_, but
-`V1` does not denote an extension type, and `V1` does not denote a
-supertype of the extension type erasure of the representation type of
-_DV_.
+which occurs as a superinterface in an extension type declaration _DV_, and
+`V1` denotes a non-extension type which is not a supertype of the
+representation type of _DV_.
 
 *In particular, in order to have `implements T` where `T` is a
 non-extension type, it must be sound to consider the representation object
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -1080,9 +1080,9 @@ representation type of _DV_.
 non-extension type, it must be sound to consider the representation object
 as having type `T`.*
 
-A compile-time error occurs if any direct or indirect superinterface
-of _DV_ is the type `Name` or a type of the form `Name<...>`. *As
-usual, subtype cycles are not allowed.*
+A compile-time error occurs if any direct or indirect superinterface of
+_DV_ denotes the declaration _DV_. *As usual, subtype cycles are not
+allowed.*
 
 Assume that _DV_ has two direct or indirect superinterfaces of the form
 <code>W\<T<sub>1</sub>, .. T<sub>k</sub>&gt;</code>


eernstg:
@sgrekhov, we may need to revisit some tests if and when this PR is landed.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1072,10 +1072,9 @@ If _DV_ does not include an `<interfaces>` clause then _DV_ has
 relation which was specified earlier.
```
lrhn, line 1072: Here I wouldn't say it has `Object?` as superinterface. It's a supertype, but top types are supertypes of all types anyway. No need to add it, or start saying that `Object?` can be a superinterface.

Also, this mixes types and declarations (or we're not specifying what I thought we'd be specifying).

I expect an extension type to have `Object` as supertype, and to implement `Object`, if its representation type is non-nullable. For a generic extension type, that may differ between different instantiations.

An extension type *declaration* may be given `Object` as immediate superinterface if its representation type is definitely non-nullable, independent of generic instantiation.
```dart
extension type E<T>(T value) {}
```
doesn't have `Object` as an immediate declared superinterface, and the type `E<Null>` doesn't have it as a superinterface at all, but `E<int>` does. (Or that's what I thought we were specifying.)

Also notice that as written, we don't add `Object` as superinterface if there is an `implements` clause, even if that implements clause doesn't necessarily introduce `Object` as transitive superinterface.
```dart
extension type _NotObject(Object? _) {}
extension type Foo(Object _) implements _NotObject {}
```
This creates an extension type with a non-nullable representation type which doesn't implement `Object`.

That's fine with me, but let's be sure we don't think it's a problem. (It may or may not still be a subtype of `Object`, there is a distinction between "implements" and "has as supertype" which matters for **Up** calculations, which only looks at superinterface, not supertypes. Let's be absolutely clear on which property we give extension types, and when.)

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1072,10 +1072,9 @@ If _DV_ does not include an `<interfaces>` clause then _DV_ has
...
 
 A compile-time error occurs if `V1` is a type name or a parameterized type
-which occurs as a superinterface in an extension type declaration _DV_, but
-`V1` does not denote an extension type, and `V1` does not denote a
-supertype of the extension type erasure of the representation type of
-_DV_.
+which occurs as a superinterface in an extension type declaration _DV_, and
+`V1` denotes a non-extension type which is not a supertype of the
+representation type of _DV_.
```
lrhn, line 1077: Change LGTM.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1072,10 +1072,9 @@ If _DV_ does not include an `<interfaces>` clause then _DV_ has
...
 
 *In particular, in order to have `implements T` where `T` is a
 non-extension type, it must be sound to consider the representation object
```
lrhn, line 1080: Below, use "declared direct or indirect superinterface", to distinguish the syntactic clauses of a declaration from the semantic superinterfaces implemented by the semantics of that declaration.
(Also, "is the type `Name`" should say "denotes the declaration _DV_", to better account for type aliasand shadowing. It's a compile-time error for a different reason if a type parameter is also called `Name`, but then that would be what the implements clause denotes.)

lrhn (approved):
I have too many comments! I will take a separate pass at the spec.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Sounds good!

accepted/future-releases/extension-types/feature-specification.md
eernstg: > Here I wouldn't say it has `Object?` as superinterface. It's a supertype, but top types are supertypes of all types anyway.

That's only true to the extent that we say so (subtype relationships exist because we have positively stated that they exist), and we say so for extension types in the section 'Static analysis of extension types'.

At this location we also specify that there is a superinterface relationship. This relationship is needed with `Object` in order to maintain a non-special cased treatment of extension types that are non-nullable during the Dart 1 least upper bound algorithm, when we specify which non-extension type members an extension type has, when we specify whether or not there is a conflict because there is an extension type member and a non-extension type member with the same name, when we compute the subtype relationship between `C<T1...>` and `C<S1...>`, in the definition of "a type implements a type", and probably elsewhere.

I think we need to introduce an explicit "plain" top type, say `Any`, which will allow us to eliminate several anomalies. I created https://github.com/dart-lang/language/issues/3265 in order to handle that.

eernstg: Let's return to this when we have some feedback on #3265.

accepted/future-releases/extension-types/feature-specification.md
eernstg: That's much better! Done.

About the word 'declared': The only 'undeclared' superinterfaces are `Object` which may implicitly be the superclass of a class (and _will_ implicitly be the superclass of a mixin class), and `Object` as an implicit left operand of `with` (which is then implicitly a superinterface of the mixin application). I can't see how it would create a problem for the correctness of the cyclicity rule that `Object` is included in these situations, just like the explicit superinterfaces.

PR #3276, opened 2023-08-15
[extension-types] Add rules about non-extension types

eernstg:
[Edit: This PR is now much simpler. Adjusted the description accordingly.]

This PR adds a covariance rule for subtyping of extension types, specifies that it is a compile-time error for an extension type member to be abstract, and it specifies that the last case in en the type function UP (that is, the Dart 1 algorithm for computation of the 'least upper bound' of two types) will work on a slightly extended superinterface graph.

### DIFF
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -1229,8 +1229,11 @@ rather than from `Object`)*.
 *This change is needed because some extension types are subtypes of
 `Object?` and not subtypes of `Object`, and they need to have a
 well-defined depth. Note that the depth of an extension type can be
-determined by its actual type arguments because type parameters of the
-extension type may occur in its representation type.*
+determined by its actual type arguments, if any, because type parameters of
+the extension type may occur in its representation type. In particular, the
+depth of an extension type is a property of the type itself, and it is not
+always possible to determine the depth from the associated extension type
+declaration.*
 
 
 ## Dynamic Semantics of Extension Types
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -1179,7 +1179,7 @@ all members of `V1, .. Vk` that are not redeclared by a declaration in _DV_
 can be invoked on a receiver of the type introduced by _DV_.*
 
 
-### Changes to subtype rules
+### Changes to type functions and types
 
 The previous sections have introduced some subtype relationships involving
 extension types. This section adds further relationships where extension
@@ -1197,12 +1197,99 @@ introduced which makes `E<S1 .. Sk>` a subtype of `E<T1 .. Tk>` if
 *For example, `E<int>` is a subtype of `E<Object>` because `int` is a
 subtype of `Object`, also in the case where `E` is an extension type.*
 
-
-### Change to type functions
-
 The type function `futureValueType` is updated as follows:
 
+- `futureValueType(S?) = futureValueType(S)`, for any `S`.
+- `futureValueType(Future<S>) = S`, for any `S`.
+- `futureValueType(FutureOr<S>) = S`, for any `S`.
+- `futureValueType(void) = void`.
+- `futureValueType(dynamic) = dynamic`.
+- `futureValueType(E) = S` if `E` is an extension type
+  that implements `Future<S>`.
+- Otherwise, `futureValueType(S) = Object?`, for any `S`.
 
+The Dart 1 algorithm which is used to compute the standard upper bound of
+two distinct interface types will work in the same way as it does today,
+albeit on a different superinterface graph:
+
+We introduce an interface type `Any` which declares `operator ==`,
+`hashCode`, `runtimeType`, `noSuchMethod`, `toString` with the same
+signatures as `Object` does today, and `Object` as well as `Null` have
+`Any` as a direct superinterface.
+
+`Any` is sealed. *That is, we cannot create additional direct subtypes of
+`Any` This is required for the soundness of the rule below about `Object?`
+being a supertype of `Any`.*
+
+*This basically means that we eliminate the anomaly which is created by
+saying that `Object` declares those members, but `Null` "has them too,
+by magic". It also implies that all paths from the top type via `Object`
+are now one step longer, which is significant during the computation of the
+standard upper bound.*
+
+For subtyping, we specify that every type is a subtype of `void`, of
+`dynamic`, and of `Any`.
+
+Moreover, we specify that `Any` is a subtype of `Object?`. *This implies
+that we collapse `Any` and `Object?`, which is basically the same thing as
+promising that `Any` will never have other subtypes than `Null` and
+`Object`. Other types (type variables and extension types) will always be
+populated by objects which are of type `Object` or `Null`, which means that
+it is sound to claim that `Object?` and `Any` are "the same type" (mutual
+subtypes).*
+
+We specify that `NonNull(Any)` is `Object`. *This is sound, and it is
+needed in order to allow `Any` to be used as a more concise way to write
+`Object?`, which is presumably quite valuable.*
+
+
+### Changes to functions
+
+Let `f` be a function with declared return type `R` whose body is `async`.
+
+An existing rule states that it is a compile-time error if `R` is not a
+supertype of `Future<T>` for any `T`.
+
+This is changed as follows: If `R` is an extension type that implements
+`Future<R1>` for some `R1` which is also the representation type of `R`
+then let `S` be `Future<R1>`. Otherwise let `S` be `R`. Next, it is a
+compile-time error if `S` is not a supertype of `Future<T>` for any `T`.
+
+*In short, nothing changes, except that an `async` function can now return
+an extension type if that extension type declares that it "is a future",
+and it has a representation type which is that same future type (not, for
+example, a subtype like `MyFuture`).*
+
+We say that said `Future<T>` is _the future return type_ of `f`. The
+run-time type of the object returned by `f` is a type that implements its
+future return type.
+
+*This is sufficient to ensure that it is a sound return value both in the
+case where `f` returns an extension type and a non-extension type.*
+
+Similarly, let `f` be a function with return type `R` whose body is marked
+`sync*`. 
+
+If `R` is an extension type that implements `Iterable<R1>` for some `R1`
+which is also the representation type of `R` then let `S` be
+`Iterable<R1>`. Otherwise let `S` be `R`. Next, it is a compile-time error
+if `S` is not a supertype of `Iterable<T>` for any `T`.
+
+We say that said `Iterable<T>` is the _iterable return type_ of `f`. The
+run-time type of the object returned by `f` is a type that implements its
+iterable return type.
+
+Finally, let `f` be a function with return type `R` whose body is marked
+`async*`. 
+
+If `R` is an extension type that implements `Stream<R1>` for some `R1`
+which is also the representation type of `R` then let `S` be
+`Stream<R1>`. Otherwise let `S` be `R`. Next, it is a compile-time error
+if `S` is not a supertype of `Stream<T>` for any `T`.
+
+We say that said `Stream<T>` is the _stream return type_ of `f`. The
+run-time type of the object returned by `f` is a type that implements its
+stream return type.
 
 
 ## Dynamic Semantics of Extension Types
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -1228,7 +1228,9 @@ rather than from `Object`)*.
 
 *This change is needed because some extension types are subtypes of
 `Object?` and not subtypes of `Object`, and they need to have a
-well-defined depth.*
+well-defined depth. Note that the depth of an extension type can be
+determined by its actual type arguments because type parameters of the
+extension type may occur in its representation type.*
 
 
 ## Dynamic Semantics of Extension Types
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -1179,6 +1179,32 @@ all members of `V1, .. Vk` that are not redeclared by a declaration in _DV_
 can be invoked on a receiver of the type introduced by _DV_.*
 
 
+### Changes to subtype rules
+
+The previous sections have introduced some subtype relationships involving
+extension types. This section adds further relationships where extension
+types occur more indirectly.
+
+*For instance, earlier sections stated that every extension type is a
+subtype of the top types (like `Object?`), and that `implements` introduces
+a subtype relationship for extension types. Moreover:*
+
+Assume that `E` denotes an extension type that takes `k` type arguments.
+Corresponding to the subtype rule about covariance for classes, a rule is
+introduced which makes `E<S1 .. Sk>` a subtype of `E<T1 .. Tk>` if
+`Sj` is a subtype of `Tj` for each `j` in 1 .. k.
+
+*For example, `E<int>` is a subtype of `E<Object>` because `int` is a
+subtype of `Object`, also in the case where `E` is an extension type.*
+
+
+### Change to type functions
+
+The type function `futureValueType` is updated as follows:
+
+
+
+
 ## Dynamic Semantics of Extension Types
 
 For any given syntactic construct which has been characterized as an
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -17,7 +17,8 @@ information about the process, including in their change logs.
 
 2023.08.17
   - Add covariance subtype rule for extension types. Add rule that it is an
-    error for an extension type member to be abstract.
+    error for an extension type member to be abstract. Mention that it is
+    allowed for extension type members to be external.
 
 2023.07.13
   - Revise many details, in particular the management of ambiguities
@@ -698,6 +699,12 @@ declaration is abstract.
 *Extension type member invocations and tear-offs are always statically
 resolved, and abstract members only make sense in the case where the given
 member is resolved at run time.*
+
+*It is not an error for an extension type member to have the modifier
+`external`. As usual, an implementation can report a compile-time error for
+external declarations, e.g., if they are not bound to an implementation,
+and the method by which they are bound to an implementation is tool
+specific.*
 
 Assume that
 <code>T<sub>1</sub>, .. T<sub>s</sub></code>
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -1190,7 +1190,7 @@ all members of `V1, .. Vk` that are not redeclared by a declaration in _DV_
 can be invoked on a receiver of the type introduced by _DV_.*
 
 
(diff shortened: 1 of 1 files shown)


bwilkerson:
@scheglov

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1179,6 +1179,119 @@ all members of `V1, .. Vk` that are not redeclared by a declaration in _DV_
...
+- Otherwise, `futureValueType(S) = Object?`, for any `S`.
+
+The Dart 1 algorithm which is used to compute the standard upper bound of
+two distinct interface types will work in the same way as it does today,
+albeit on a different superinterface graph:
+
+We introduce an interface type `Any` which declares `operator ==`,
+`hashCode`, `runtimeType`, `noSuchMethod`, `toString` with the same
+signatures as `Object` does today, and `Object` as well as `Null` have
+`Any` as a direct superinterface.
```
lrhn, line 1218: Let's not get ahead of ourselves. I'm not sure adding `Any` is worth the complexity.

I'd prefer to say that the Dart 1 LUB algorithm is extended to work on super-interface-sets of types which include extension types, and which may not alwaus contain `Object`.
That means that it is possible to have two sets of super-interfaces, whose intersection does not contain any level with exactly one element. 
In that case the LUB algorithm does not have a solution, and the UP step which defers to it then makes the result be `Object?`.

Adding a new type _just_ to avoid the LUB algorithm having zero solutions isn't really worth it.
I'm happy to discuss the value of adding `Any` in general, as a separte feature, but it's not worth it just for this.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1179,6 +1179,119 @@ all members of `V1, .. Vk` that are not redeclared by a declaration in _DV_
...
+
+`Any` is sealed. *That is, we cannot create additional direct subtypes of
+`Any` This is required for the soundness of the rule below about `Object?`
+being a supertype of `Any`.*
+
+*This basically means that we eliminate the anomaly which is created by
+saying that `Object` declares those members, but `Null` "has them too,
+by magic". It also implies that all paths from the top type via `Object`
```
lrhn, line 1226: It doesn't actually remove the anomaly entirely, it just fixes it for `Null`.
It doesn't explain why union types have the same members, that's just something we say that they do, or why `dynamic` treats those members specially. All it does is introduce a *nominative type* which declares an interface that everybody must implement. But we could just say that "this is the set of members that every type must implement", without making it any particular type's responsiblity to introduce them.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1179,6 +1179,119 @@ all members of `V1, .. Vk` that are not redeclared by a declaration in _DV_
...
+are now one step longer, which is significant during the computation of the
+standard upper bound.*
+
+For subtyping, we specify that every type is a subtype of `void`, of
+`dynamic`, and of `Any`.
+
+Moreover, we specify that `Any` is a subtype of `Object?`. *This implies
```
lrhn, line 1233: That's just saying that any type is a subtype of `void`, `dynamic`, `Any` and `Object?`.

Maybe we could collapse all those to just:
> * `S` \<: `T` if **TOP**(`T`)

(The current subtyping rule is:
> * **Right Top**: if `T1` is a top type (i.e. `dynamic`, or `void`, or `Object`?) then `T0` \<: `T1`

It should probably use "e.g." instead of "i.e." here, and really use **TOP**(`T1`) as the definition.)

And then we keep making **TOP**(`Object?`) be true, like it is to day, and add **TOP**(`Any`) = true as well.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1179,6 +1179,119 @@ all members of `V1, .. Vk` that are not redeclared by a declaration in _DV_
...
+that we collapse `Any` and `Object?`, which is basically the same thing as
```
lrhn, line 1234: We collapse all top types, with `Any` just being the newest member of the club.
It's true that by making `Object?` a top type, we promise that there can be no values that are not either `null` or an instance of `Object`, since all values can be made to have static type `Object?`, and it wouldn't be sound if that value wasn't `null` or an `Object`.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1179,6 +1179,119 @@ all members of `V1, .. Vk` that are not redeclared by a declaration in _DV_
...
+needed in order to allow `Any` to be used as a more concise way to write
+`Object?`, which is presumably quite valuable.*
+
+
+### Changes to functions
+
+Let `f` be a function with declared return type `R` whose body is `async`.
+
+An existing rule states that it is a compile-time error if `R` is not a
+supertype of `Future<T>` for any `T`.
```
lrhn, line 1251: (Rule says "a supertype of `Future<Never>`".)

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1179,6 +1179,119 @@ all members of `V1, .. Vk` that are not redeclared by a declaration in _DV_
...
+
+This is changed as follows: If `R` is an extension type that implements
+`Future<R1>` for some `R1` which is also the representation type of `R`
+then let `S` be `Future<R1>`. Otherwise let `S` be `R`. Next, it is a
+compile-time error if `S` is not a supertype of `Future<T>` for any `T`.
```
lrhn, line 1256: This should match the rule change above.

Here, we can go with:
> A return type `T` of an `async` function is a valid async function return type iff:
> * `T` is an extension type with representation type `R` and `R` is valid async function return type, or
> * `Future<Never>` \<: `T`
> 
> It's a compile-time error if the return type is not a valid async function return type.


We should perhaps just define `futureValueType` to be able to have no result, and consider that a compile time error. Rather than defining this rule which effectively amounts to "futureValueType must be well-defined for the type" and then define `futureValueType` separately.

Using the partial future value type would be something like:
> The function *futureValueType*: *type* &rarr; *type*<sub>&bot;</sub>
> is a partial function that attempts to return a type `T` for a type `S` such that `Future<T> is S` must be true,
> or which returns &bot; when such a type cannot be derived.
>
> The function is defined by the following cases, in priority order:
> * *futureValueType*(`FutureOr<T>`) = `T`
> * *futureValueType*(`Future<T>`) = `T`
> * *futureValueType*(`T?`) = *futureValueType*(`T`)
> * *futureValueType*(`E`) = *futureValueType*(`R`) if `E` is an extension type with representation type `R`
> * *futureValueType*(`dynamic`) = `dynamic`
> * *futureValueType*(`void`) = `void`
> * *futureValueType*(`T`) = `Object?` if `Future<Never>` \<: `T` (aka if **TOP**(`T`) or **OBJECT**(`T`)).
> * *futureValueType*(`T`) = &bot;

Then we can say:
> Let *V* be *futureValueType*(`R`) where `R` is the declared function type of the async function.
> It's a compile-time error if *V* is &bot;.
> Otherwise ... analyze the body using *V* as target type for returns.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1179,6 +1179,119 @@ all members of `V1, .. Vk` that are not redeclared by a declaration in _DV_
...
+
+*In short, nothing changes, except that an `async` function can now return
+an extension type if that extension type declares that it "is a future",
+and it has a representation type which is that same future type (not, for
+example, a subtype like `MyFuture`).*
```
lrhn, line 1261: It's transitive representation type must be a supertype of `Future<Never>`.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1179,6 +1179,119 @@ all members of `V1, .. Vk` that are not redeclared by a declaration in _DV_
...
+*This is sufficient to ensure that it is a sound return value both in the
+case where `f` returns an extension type and a non-extension type.*
+
+Similarly, let `f` be a function with return type `R` whose body is marked
+`sync*`. 
+
+If `R` is an extension type that implements `Iterable<R1>` for some `R1`
+which is also the representation type of `R` then let `S` be
+`Iterable<R1>`. Otherwise let `S` be `R`. Next, it is a compile-time error
+if `S` is not a supertype of `Iterable<T>` for any `T`.
```
lrhn, line 1276: Same issue. The implemented type is a red herring, it cannot be used unless it precisely matches the representation type, so always use the representation type.

Here:
> * *iterableElementType*(`T?`) = *iterableElementType*(`T`)
> * *iterableElementType*(`FutureOr<T>`) = *iterableElementType*(`T`)
> * *iterableElementType*(`Iterable<R>`) = `R`
> * *iterableElementType*(`E`) = *iterableElementType*(`R`) if `E` is an extension type with representation `R`
> * *iterableElementType*(`dynamic`) = `dynamic`
> * *iterableElementType*(`void`) = `void`
> * *iterableElementType*('T') = `Object?` if `Iterable<Never>` \<: `T` (aka **TOP**(`T`) or **OBJECT**(`T`)).
> * *iterableElementType*('T') = &bot;

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1179,6 +1179,119 @@ all members of `V1, .. Vk` that are not redeclared by a declaration in _DV_
...
+`Future<R1>` for some `R1` which is also the representation type of `R`
+then let `S` be `Future<R1>`. Otherwise let `S` be `R`. Next, it is a
+compile-time error if `S` is not a supertype of `Future<T>` for any `T`.
+
+*In short, nothing changes, except that an `async` function can now return
+an extension type if that extension type declares that it "is a future",
+and it has a representation type which is that same future type (not, for
+example, a subtype like `MyFuture`).*
+
+We say that said `Future<T>` is _the future return type_ of `f`. The
```
lrhn, line 1263: `T` is not defined here, there can be many different `T`s such `Future<T>` \<: `R`. 

It should be `Future<futureValueType(R)>`, which is specified as a single type.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1179,6 +1179,119 @@ all members of `V1, .. Vk` that are not redeclared by a declaration in _DV_
...
+
+Similarly, let `f` be a function with return type `R` whose body is marked
+`sync*`. 
+
+If `R` is an extension type that implements `Iterable<R1>` for some `R1`
+which is also the representation type of `R` then let `S` be
+`Iterable<R1>`. Otherwise let `S` be `R`. Next, it is a compile-time error
+if `S` is not a supertype of `Iterable<T>` for any `T`.
+
+We say that said `Iterable<T>` is the _iterable return type_ of `f`. The
```
lrhn, line 1278: `Iterable<iterableElementType(R)>`

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1179,6 +1179,119 @@ all members of `V1, .. Vk` that are not redeclared by a declaration in _DV_
...
+run-time type of the object returned by `f` is a type that implements its
+iterable return type.
+
+Finally, let `f` be a function with return type `R` whose body is marked
+`async*`. 
+
+If `R` is an extension type that implements `Stream<R1>` for some `R1`
+which is also the representation type of `R` then let `S` be
+`Stream<R1>`. Otherwise let `S` be `R`. Next, it is a compile-time error
+if `S` is not a supertype of `Stream<T>` for any `T`.
```
lrhn, line 1292: And:

> * *streamElementType*(`T?`) = *streamElementType*(`T`)
> * *streamElementType*(`FutureOr<T>`) = *streamElementType*(`T`)
> * *streamElementType*(`Stream<T>`) = `T`
> * *streamElementType*(`E`) = *streamElementType*(`R`) if `E` is an extension type with representation `R`
> * *streamElementType*(`dynamic`) = `dynamic`
> * *streamElementType*(`void`) = `void`
> * *streamElementType*('T') = `Object?` if `Stream<Never>` \<: `T` (aka **TOP**(`T`) or **OBJECT**(`T`)).
> * *streamElementType*('T') = &bot; otherwise.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1179,6 +1179,119 @@ all members of `V1, .. Vk` that are not redeclared by a declaration in _DV_
...
+
+We say that said `Stream<T>` is the _stream return type_ of `f`. The
```
lrhn, line 1294: `Stream<streamElementType(R)>`

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1179,6 +1179,119 @@ all members of `V1, .. Vk` that are not redeclared by a declaration in _DV_
...
+
+The type function `futureValueType` is updated as follows:
+
+- `futureValueType(S?) = futureValueType(S)`, for any `S`.
+- `futureValueType(Future<S>) = S`, for any `S`.
+- `futureValueType(FutureOr<S>) = S`, for any `S`.
+- `futureValueType(void) = void`.
+- `futureValueType(dynamic) = dynamic`.
+- `futureValueType(E) = S` if `E` is an extension type
+  that implements `Future<S>`.
```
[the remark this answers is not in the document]
eernstg: > I'd say:

I'm restricting the admissible extension types such that the representation type and the implemented type are the same type. So it makes no difference.

However, I'd prefer to ensure that the return type of an `async` bodied function is a subtype of `Future<T>` for some `T`, such that invocations can be detected by 'unawaited_futures' and the like.

The underlying conceptual motivation is that we should not drop any futures, so we shouldn't deliver futures using a type that hides that it is a future. (Again, we can relax this restriction later if it turns out to be useful, but for now I think it's a serious code smell).

> the returned future value must be a `Future<int>`, not a `Future<num>`.

This is again something which has already been taken into account.

accepted/future-releases/extension-types/feature-specification.md
eernstg: > I'd prefer to say [something more complex]

I'd prefer the simple and consistent approach, which is to add `Any`. Let's get some feedback from the rest of the language team.

accepted/future-releases/extension-types/feature-specification.md
eernstg: We have always used the shared members as a reasonable hypothetical rule for the interface of a union type. This would simply use that as an actual rule. Granted, this is new, technically, but it shouldn't be an alien idea.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Ack.

accepted/future-releases/extension-types/feature-specification.md
eernstg: This is already guaranteed with the proposed rules.

accepted/future-releases/extension-types/feature-specification.md
eernstg: > The implemented type is a red herring, it cannot be used unless it precisely matches the representation type

Which is indeed guaranteed with the rules as proposed.

lrhn: Missed the "which is also the representation type of `R`".

Saying that one type "is" also another type means we have to determine whether two type terms denote the same type.
If one is `Future<Object?>` and the other `Future<dynamic>`, then presumably this rule would not apply.

Also, i actualy *want* to allow 
```dart
extension type Delay._(Future<void> _delay) {
  Delay(Duration duration) => Delay._(Future.delayed(duration, (){}));
  Delay add(Duration duration) => Delay._(_delay.then((_) => Future.delayed(duration, () {})));
  Future<void> get whenDone => _delay;
  static Delay max(Iterable<Delay> delays) =>
      Delay._(Future.wait(delays as Iterable<Future<void>>).then((_) {})); 
  static Delay consecutive(Iterable<Delay> delays) async {
    for (var delay in delays) await delay._delay;
  }
}
```
which is not a future, but where the `consecutive` function is an `async` function which returns a `Delay`.

So, I really want *just* the representation type to count, because otherwise I have to introduce an indirection like:
```dart
  static Delay consecutive(Iterable<Delay> delays) => Delay._(() async {
    for (var delay in delays) await delay._delay;
  }());
```
and nobody will thank us for making them do that.
(I expect streams to be more likely operations to wrap as something else.)

accepted/future-releases/extension-types/feature-specification.md
lrhn: We don't actually use the "members from both types are available on union" rule, and there are cases where it could apply, other than `Object?`. Fx:
```dart
void foo<Z extends Future<Z>>(Z z) {
  Future<Z> f1 = z;
  f.then; // Does not exist, but both `Z` and `Future<Z>` has the same member, both from `Future<Z>`.
}
```

accepted/future-releases/extension-types/feature-specification.md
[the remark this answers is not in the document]
eernstg: Closing, this text has been deleted.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Closing, this text has been deleted.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Closing, this text has been deleted.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Closing, this text has been deleted.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Closing, this text has been deleted.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Closing, this text has been deleted.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Closing, this text has been deleted.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Closing, this text has been deleted.

eernstg (commented):
The PR has been simplified substantially. It no longer introduces `Any`, and it no longer allows an `async` bodied function to return an extension type, and similarly for `sync*` and `async*`. PTAL.

eernstg:
@lrhn, friendly ping.

lrhn (approved):
LGTM.

We should probably say explicitly that the depth of an extension type is not fixed by its declaration, but may differ depending on whether the representation type of a particular instantiation is non-nullable or not.

eernstg:
Oh, that's a good point! Done.

eernstg:
@scheglov, @johnniwinther, this PR has now been landed. I think several changes have already been implemented (e.g., I think the CFE supports the covariance subtype rule also for extension types), but there might still be some missing elements.

@sgrekhov, do we already have tests for all of these changes?

sgrekhov:
We have Subtyping tests that check covariance and https://github.com/dart-lang/co19/pull/2223
Please also note https://github.com/dart-lang/sdk/issues/53290

eernstg:
Sounds good, including the crash, thanks! ;-)

PR #3278, opened 2023-08-15
Fix how trailing list subpatterns are cached

### LINKED ISSUE #2922: [patterns] Invocation key for list pattern tail element may need to use a different index
Consider [the following part](https://github.com/dart-lang/language/blob/2a3a253ca81d3cf0b78ad88e34aaaf6e185b5345/accepted/future-releases/0546-patterns/feature-specification.md#side-effects-and-exhaustiveness:~:text=Else%20s%20is%20a%20non%2Drest%20element%20after%20the%20rest%20element) of the patterns feature specification:

> Else `s` is a non-rest element after the rest element:
> &nbsp;&nbsp;a. Let `e` be `i : ("tail[]", [index])` where `index` is the zero-based index of this element subpattern.

This invocation key identifies an invocation of operator `[]` on the given matched value (which is a `List` of length `l`) with argument `l - t + tailIndex` where `t` is the number of element patterns after the rest element, and `tailIndex` is the zero based index of this element pattern in the list of tail element patterns.

Presumably this could be encoded as `i : ("tail[]", [t - tailIndex])`, which would allow for the computation of `l - t + tailIndex` and hence access to the relevant element in the matched value. (We could also use the argument list `[t, tailIndex]`, but a single integer is sufficient to enable the required computation at run time.)

However, the current wording mentions the zero-based index of the given element, and that does not sound like it could be a compile-time constant value for any element pattern after a rest element. (It could be the zero-based index in the list of element patterns in the list pattern, including the rest pattern, but then we would presumably need to know `h` in order to be able to perform the denoted computation.)

In particular, the approach where the argument list is `t - tailIndex` will associate the last pattern (that is, `p2`, `p4`, `p7`) with invocation key argument 1 for all the list patterns `[p1, ..., p2]`, `[..., p3, p4]`, `[p5, p6, ..., p7]`, and they would indeed look up the same element in the same scrutinee, but I can't see how the original wording would allow us to recognize that it is the same element (namely `scrutinee[length - 1]`).

Hence, I'd suggest changing this bullet to something like the following:

> Else `s` is a non-rest element after the rest element:
> &nbsp;&nbsp;a. Let `e` be `i : ("tail[]", [t - tailIndex])` where `tailIndex` is the zero-based index of this element subpattern in the list of element subpatterns that occurs after the rest element.

@munificent, WDYT?

munificent:
I did some testing and it looks like the implementations are already doing the right thing and caching trailing list elements across patterns, so we may as well update the proposal to match.

This should help ensure we specify it right when it ends up in the spec.

Fix #2922.

### DIFF
--- a/accepted/3.0/patterns/feature-specification.md
+++ b/accepted/3.0/patterns/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: Accepted
 
-Version 2.32 (see [CHANGELOG](#CHANGELOG) at end)
+Version 2.33 (see [CHANGELOG](#CHANGELOG) at end)
 
 Note: This proposal is broken into a couple of separate documents. See also
 [records][] and [exhaustiveness][].
@@ -3389,8 +3389,9 @@ To bind invocation keys in a pattern `p` using parent invocation `i`:
 
         3.  Else `s` is a non-rest element after the rest element:
 
-            1.  Let `e` be `i : ("tail[]", [index])` where `index` is the
-                zero-based index of this element subpattern.
+            1.  Let `e` be `i : ("tail[]", [tailIndex])` where `tailIndex` is
+                the number of element subpatterns following this element
+                subpattern.
 
                 *Note the "tail" in the invocation key name. This is to
                 distinguish elements after a rest element at some position from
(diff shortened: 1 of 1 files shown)


eernstg (approved):
LGTM!

Issue #3301, opened 2023-08-23
[extension-types] Specify that it is not an error to declare an external variable


eernstg:
The feature specification about extension types currently says that

> A compile-time error occurs if the extension type declaration declares any instance variables.

This should be extended to say that `external` instance variable is not an error, both final and non-final.

eernstg:
Cf. https://github.com/dart-lang/language/pull/3303.

PR #3303, opened 2023-08-23
Clarify that an extension type _can_ have instance variables which are external

eernstg:
The current rule about instance variable declarations in an extension type declaration is that every instance variable declaration is a compile-time error (we only allow the implicitly induced instance variable which is caused by the parameter declaration in the primary constructor which is now syntactically the `<representationDeclaration>`, because we don't yet have primary constructors).

However, it is not an error to declare an external getter and/or setter, and an `external` instance variable declaration is just a convenient notation for exactly that.

This PR just clarifies that an external instance variable is not an error.

Fixes https://github.com/dart-lang/language/issues/3301.

### DIFF
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -429,7 +429,10 @@ disambiguate the extension type declaration with a fixed lookahead.*
 Some errors can be detected immediately from the syntax:
 
 A compile-time error occurs if the extension type declaration declares any
-instance variables.
+instance variables, unless they are `external`.
+
+*An external instance variable is just a convenient notation for an external
+getter and (if not `final`) an external setter. They are allowed.*
 
 The _name of the representation_ in an extension type declaration with a
 representation declaration of the form `(T id)` is the identifier `id`, and


jakemac53 (approved):

eernstg:
Thanks!

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -429,7 +429,10 @@ disambiguate the extension type declaration with a fixed lookahead.*
 Some errors can be detected immediately from the syntax:
 
 A compile-time error occurs if the extension type declaration declares any
-instance variables.
+instance variables, unless they are `external`.
+
+*An external instance variable is just a convenient notation for an external
+getter and (if not `final`) an external setter. They are allowed.*
```
lrhn, line 435: Extremely nit-picky: Consider changing "They" to "Those".
I like the more demonstrative pronoun.

Alternately ", which are both allowed."

lrhn (approved):

accepted/future-releases/extension-types/feature-specification.md
eernstg: Done in https://github.com/dart-lang/language/pull/3309.

PR #3315, opened 2023-08-31
add basic checks extension macro

jakemac53:
This macro generates extensions for a list of types, for use with `package:checks`. It isn't yet runnable as I need to do a bit of upstream work, but should be a good start.

I also updated the spec for extensions, allowing introspection on the `on` type _as well as all generic type parameters_. @munificent I added you to review that part. I also reworded that section just a bit and added a bit more general clarification.

### DIFF
--- a/working/macros/example/bin/checks_main.dart
+++ b/working/macros/example/bin/checks_main.dart
@@ -1,0 +1,23 @@
+@ChecksExtensions([Person])
+library;
+
+import 'package:checks/checks.dart';
+import 'package:test/test.dart';
+
+import 'package:macro_proposal/checks_extensions.dart';
+
+void main() {
+  test('can use generated extensions', () {
+    final draco = Person(name: 'Draco', age: 39);
+    check(draco)
+      ..name.equals('Draco')
+      ..age.equals(39);
+  });
+}
+
+class Person {
+  final String name;
+  final int age;
+
+  Person({required this.name, required this.age});
+}
--- a/working/macros/example/lib/checks_extensions.dart
+++ b/working/macros/example/lib/checks_extensions.dart
@@ -1,0 +1,140 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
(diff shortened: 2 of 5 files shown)


working/macros/example/bin/checks_main.dart
```diff
@@ -0,0 +1,23 @@
...
+void main() {
+  test('can use generated extensions', () {
+    final draco = Person(name: 'Draco', age: 39);
+    check(draco)
+      ..name.equals('Draco')
+      ..age.equals(39);
+  });
+}
+
+class Person {
```
natebosch, line 18: Can we make an example where the class we generate for is imported?

working/macros/example/lib/checks_extensions.dart
```diff
@@ -0,0 +1,140 @@
...
+            'explicit generics');
+      }
+      final name = '${type.identifier.name}Checks';
+      builder.declareType(
+          name,
+          DeclarationCode.fromParts([
+            '@',
+            checksExtension,
+            '()',
+            'extension $name on ',
```
natebosch, line 45: This is interesting - it looks like the macro runs in two steps?

Does that mean that folks who want more customization can do something like

```dart
@ChecksExtension()
extension CustomName on Subject<SomeType> {}
```

natebosch (approved):

working/macros/example/lib/checks_extensions.dart
jakemac53: Yeah, it wouldn't have to work this way but that is imo the best way to write it. It would have to run in two phases regardless.

The macro could be smart enough to skip adding extensions that already exist though, so you could write custom ones and have it fill in the rest.

I am not sure all other macro use cases would want the 2nd macro (`@ChecksExtension()` here) to actually be visible so it is a little weird. It could actually exist in a `src` library, so that might be the solution "hiding" it. But it has to be public.

working/macros/example/bin/checks_main.dart
jakemac53: It doesn't make any difference in this case - and the boilerplate required right now for running these is already a lot, I don't really want to add more files.

For the real macro, once you can run it in a more sane way, you could absolutely do that.

munificent (approved):

PR #3333, opened 2023-09-08
remove mention of concrete variables, simplify rules around variable augmentation

jakemac53:
Per our conversation today, this:

- Removes all mention of concrete variables.
- Augmenting with a variable now only ever means augmenting the initializer of the augmented variable.
- Disallow augmenting external and abstract variables with variables.


scheglov (approved):

jakemac53:
Going to go ahead and merge this, but @munificent if you have any comments I will follow up, or revert, as needed :).

PR #3338, opened 2023-09-11
Clarify that the usual getter/setter type rule exists for extension types as well

eernstg:
We have a rule saying that it is an error to have a getter returning `R` and a corresponding setter whose parameter type is `S` if `R` is not a subtype of `S`.

This PR introduces the same rule for an extension type declaration.


lrhn (approved):

eernstg:
Thanks!

Issue #3350, opened 2023-09-15
Rework the augmentations library structure and ordering section for macros


jakemac53:
There is a core built in assumption of [this section](https://github.com/dart-lang/language/blob/main/working/macros/feature-specification.md#augmentation-library-structure-and-ordering) which is incorrect, so the whole section needs to be reworked.

The assumption was that we can always just append new code to the bottom for subsequent augmentations, but this isn't accurate because new augmentations might add _imports_ at the top. This means offsets of previous declarations _will_ have to be updated not only between phases, but also during the middle of the declarations phase.

Given this fact, we also may want to adjust the other decisions made, such as creating separate type augmentations for each macro application. It is better from a user perspective if these get merged. 

cc @johnniwinther @scheglov

jakemac53:
Note that there is another possible option here, which is to allow directives anywhere in augmentation libraries, but I think that would be bad.

PR #3359, opened 2023-09-18
Change the specified augmentation library structure for macros

jakemac53:
This moves us to a more user friendly format (all augmentations for a single type grouped together), and removes the incorrect statements about source offset stability.

Closes https://github.com/dart-lang/language/issues/3350

### DIFF
--- a/working/macros/feature-specification.md
+++ b/working/macros/feature-specification.md
@@ -449,7 +449,7 @@ rules).
 ```dart
 augment class A {
   void c() {} // Added in phase 2, ran before `AddTopLevelFoo()`.
-  augment c() {} // Added in phase 3, but merged into the previous augmentation.
+  augment b() {} // Added in phase 3, but merged into the previous augmentation.
 }
 
 int foo = 1; // Added in phase 2, after `c` was added to `A`.
--- a/working/macros/feature-specification.md
+++ b/working/macros/feature-specification.md
@@ -307,35 +307,40 @@ run the same macros with the same inputs, and get the same augmentation library.
 This allows debugging and stack traces to work consistently, and be meaningful
 and useful.
 
-It is also important that if a declaration is added in Phase 1, the source
-offsets to that declaration should not change after Phase 2 or 3 run. This means
-that tools don't need to update source offsets after each phase.
+We have several rules based around maintaining the consistency of generated
+output across tools.
 
-Another important consideration is that in Phase 2, the ordering of macros is
-user perceptable, and so augmentation results should be serializable to disk -
-with stable offsets - at multiple points throughout the process.
+#### Rule #1: Nested augmentations on type declarations are merged
 
-We have several rules based around maintaining this stability of source offsets
-and consistency of generated output across tools.
+When there are multiple augmentations of the same type declaration, they are
+merged into a single `augment <type> {}` block. This is easier for end users to
+understand. This includes multiple augmentations of the _same_ declaration, they
+should appear as separate declarations within the same type augmentation.
 
-#### Rule #1: Each macro application appends a new independent augmentation
+This does result in constantly shifting source offsets between phases, and in
+particular throughout Phase 2 of macro expansion, given that some macros can see
+the outputs of other macros within that same phase.
 
-While augmentations on a given type declaration can be grouped together, they
-are not required to be. You can have as many `augment class A {}` declarations
-as you want in a given augmentation library. We take advantage of that fact, and
-require that every macro application creates its own augmentation.
+Note that we previously considered an "append only" approach, with no merging of
+augmentations. The goal was to avoid changing source offsets, but this doesn't
+work since later augmentations may need to add additional imports, which would
+result in shifting offsets anyways. Since we have to deal with the shifting
+offsets either way, we might as well derive user value out of it.
 
-This means each macro application is only appending new top level augmentation
-declarations to the augmentation library. It only grows over time, and previous
-lines are never altered.
+#### Rule #2: Augmentations are sorted by phase, application, then source order
 
-#### Rule #2: Augmentations are added in application order, then source order
+Firstly, augmentations from earlier phases appear before augmentations from later
+phases.
 
 Where an application order is explicitly defined, the augmentations are appended
 in that same order as the primary sort.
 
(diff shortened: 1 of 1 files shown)


working/macros/feature-specification.md
```diff
@@ -352,16 +357,20 @@ separating declarations.
 In the future, we may decide to run `dart format` or some other lighter weight
 formatter on augmentations which would also enforce consistent whitespace.
 
+#### Rule #4: New types are declared separately from their augmentations
+
+If a macro declares a new type and then later augments it, this will result in
+separate type declarations. One normal one followed by an augmentation of that
```
scheglov, line 363: Is this to support augmenting definitions?
```dart
class A {
  void foo() {
    print('foo');
  }
}

augment class A {
  augment void foo() {
    print('before1');
    super augment foo();
    print('after1');
  }
}

augment class A {
  augment void foo() {
    print('before2');
    super augment foo();
    print('after2');
  }
}
```

Or the "last statement" as I referred to it before, is about this?
But we can put all augmentations sequentially like this.
```dart
class A {
  void foo() {
    print('foo');
  }

  augment void foo() {
    print('before1');
    super augment foo();
    print('after1');
  }

  augment void foo() {
    print('before2');
    super augment foo();
    print('after2');
  }
}
```
Ah, I see the issue - we cannot have non-augmentation `class A` and augmentation methods in it.
So, we do them separately.
```dart
class A {
  void foo() {
    print('foo');
  }
}

augment class A {
  augment void foo() {
    print('before1');
    super augment foo();
    print('after1');
  }

  augment void foo() {
    print('before2');
    super augment foo();
    print('after2');
  }
}
```

This seems fine to me - most macros probably will generate new methods, not augment definitions.

But if we generate a new class, and just add new methods to it - currently we also call it "augmentation", right? If we have
```dart
@MakeClassB
class A {}
```
then
```dart
library augment 'a.macro.dart';

@AddMethodFoo
class B {}
```
Do we get this?
```dart
library augment 'a.macro.dart';

@AddMethodFoo
class B {
  void foo() {}
}
```
or this?
```dart
library augment 'a.macro.dart';

@AddMethodFoo
class B {
  void foo() {}
}

augment class B {
  void foo() {}
}
```

working/macros/feature-specification.md
```diff
@@ -307,36 +307,41 @@ run the same macros with the same inputs, and get the same augmentation library.
...
-user perceptable, and so augmentation results should be serializable to disk -
-with stable offsets - at multiple points throughout the process.
+#### Rule #1: Nested augmentations on type declarations are merged
 
-We have several rules based around maintaining this stability of source offsets
-and consistency of generated output across tools.
+When there are multiple augmentations of the same type declaration, they are
+merged into a single `augment <type> {}` block. This is easier for end users to
+understand. This includes multiple augmentations of the _same_ declaration, they
+should appear as separate declarations within the same type augmentation.
```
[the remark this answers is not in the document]
jakemac53: The second example means if you are augmenting `foo` multiple times:

```dart
augment class A {
  augment void foo() {}
  augment void foo() {}
  augment void foo() {}
  augment void foo() {}
}
```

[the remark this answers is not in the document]
jakemac53: Note that the augmentation library proposal might need to be updated to explicitly allow this, I think it might be ambiguous today.

working/macros/feature-specification.md
jakemac53: You get the latter, which is what this paragraph is attempting to describe :).

```dart
library augment 'a.macro.dart';

@AddMethodFoo
class B {
  void foo() {}
}

augment class B {
  void foo() {}
}
```

scheglov: Oh, sorry, my last example was not correct.
I did not mean to have two `foo()` declarations.
So, it should be instead
```dart
library augment 'a.macro.dart';

@AddMethodFoo
class B {}

augment class B {
  void foo() {}
}
```

Which seems unfortunate. I can imagine a macro that is applied to a function for example, to create a class that wraps all parameters, and the created class itself is annotated with `@generateEquals`. It would seem natural to put the generated `operator ==()` into the same class.

jakemac53: > Which seems unfortunate. I can imagine a macro that is applied to a function for example, to create a class that wraps all parameters, and the created class itself is annotated with `@generateEquals`. It would seem natural to put the generated `operator ==()` into the same class.

I agree it is less than ideal from a user perspective. I have it this way because it is the easiest/safest way to generate the library itself without actually parsing the code. I could explicitly call that out though?

scheglov: Sorry, I don't understand. What parsing?

jakemac53: > Sorry, I don't understand. What parsing?

When you declare types in phase one, what we get is mostly just an opaque String with some identifiers in it. If we wanted to merge _new_ declarations from phase 2 into that declaration, we would have to parse it to find out where in the string we could inject new declarations. With the current model we don't have to do that, because we get the individual augmentations for a given type as standalone declarations and synthesize the surrounding `augment <type> {...}` declaration. So we can easily group them together without parsing.

working/macros/feature-specification.md
```diff
@@ -307,36 +307,41 @@ run the same macros with the same inputs, and get the same augmentation library.
...
-It is also important that if a declaration is added in Phase 1, the source
-offsets to that declaration should not change after Phase 2 or 3 run. This means
-that tools don't need to update source offsets after each phase.
+We have several rules based around maintaining the consistency of generated
+output across tools.
 
-Another important consideration is that in Phase 2, the ordering of macros is
-user perceptable, and so augmentation results should be serializable to disk -
-with stable offsets - at multiple points throughout the process.
+#### Rule #1: Nested augmentations on type declarations are merged
```
munificent, line 313: These rules would definitely be easier to understand with some concrete examples like the ones in the comments below.

munificent (approved):

jakemac53:
I pushed several examples, and removed the existing example which was not actually relevant any longer (it was based on the appending strategy).

munificent (approved):

scheglov (approved):

PR #3384, opened 2023-10-06
Add updated inference rules for `num` types.

lrhn:
No changes intended, but the existing rules are phrased based on assumptions that are no longer true with extension types. Adds more restrictions on when the rules apply, to rule out things that previously couldn't happen.


lrhn:
This is a smaller, more focused, subset of my earlier PR.

It makes the new rules precise and explicit. I'm writing tests to match.
(We can obviously discuss the actual rules, but I think they're "safe enough".)

eernstg:
Discussed this IRL yesterday: We will return to this later if needed.

lrhn:
We will, as a minimum, need to say that the rules only apply if invoking an instance method, not an extension type method.
I'll prepare a PR saying just that.

Issue #3383, opened 2023-10-06
Disallow extension typed declarations with bottom as representation type - from implementing interface types.


lrhn:
## Proposal

It's a compile time error of an extension type declaration has a declared representation type which is a bottom type, and the extension type declaration directly or transitively declares a non-extension type supertype.

An extension type decoration directly declares a non-extension type supertype if its declaration has an `implements` clause with an entry which denotes a non-extension type.

It transitively declares a non-extension type supertype if it has an `implements` clause with an entry which denotes an extension type whose declaration directly or transitively declares a non-extension type supertype.

## Background
As currently specified, you can declare:
```dart
extension type Madness(Never _) implements int, double {
  Madness operator -() => this;  // Only a few self-typed nullary members conflict.
  Madness abs() => this;
  Madness get sign => this;
}
```
That's a type which implements `int` *and* `double`. It's not particularly useful, but it's a complication that the specification has to account for. Simply because it *can* exist.
This one, in particular, complicates the special-case type inference for number operators.

Another complicating factor for those rules would be:
```dart
extension type Silly(Never _) implements int, Other {
}
abstract interface class Other {
  Never operator+(Object? a);
}
```
which implements `int` and has a combined interface for `operator+` of `Never Function(Object?)`. (An interface which is trivially satisfied by `Never`.)

There might be more places where such types can confuse the current specification, perhaps a type implementing two different `enum`s.

I propose we either:
* Disallow a representation type _declaration_ which is a bottom type, or
* Disallow a an extension type with a representation type _declaration_ which is a bottom type, from implementing interface types.

It won't prevent the representation type from actually being `Never`. The representation type declaration can be a type variable with a non-bottom bound, which is bound to `Never`, but that won't allow declaring `implements` clauses below the bound.

It just means that an *actual*, compile-time botton representation type either is not allowed (the type is uninstantiable anyway), or it's allowed, but like a representation type of `Null` or `int?`, there are no interfaces that it can implement.

Maybe the rule should be that if the representation type is not an "interface type", you cannot implement interface types, using the same categorization as for which types you can implement. You'll still get `Object` for free if the representation type is non-nullable.

So, mostly out of an abundance of caution, I think we should disallow this, most likely useless, complication, like we disallowed actually implementing `Null`, `Never`, `Function` and `Record` (because the spec may make assumptions about any non-bottom subtype of `Record` actually being a record type.)

eernstg:
[...]

PR #3393, opened 2023-10-11
Make it an error for the representation type of an extension type to be a bottom type

eernstg:
In response to the discussion in https://github.com/dart-lang/language/issues/3383, this PR introduces a new compile-time error to the extension type feature specification: It is an error if the representation type is `Never` (or something with the same meaning).

### DIFF
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -766,8 +766,20 @@ the treatment of expressions like `e as V`.*
 
 A compile-time error occurs if a type parameter of an extension type
 declaration occurs in a non-covariant position in the representation type.
+
 A compile-time error occurs if the representation type of an extension type
 declaration is a bottom type.
+
+*Note that it is still possible for the instantiated representation type 
+of a given extension type to be a bottom type. For example, assuming
+`extension type E<X>(X x) {}`, `E<Never>` would be an extension type whose
+instantiated representation type is `Never`. The reason for this error is that
+we could otherwise have, for example, an extension type that `implements int,
+double`, and the ability to have such subtypes of arbitrary sets of interface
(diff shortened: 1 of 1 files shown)


accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -766,6 +766,8 @@ the treatment of expressions like `e as V`.*
 
 A compile-time error occurs if a type parameter of an extension type
 declaration occurs in a non-covariant position in the representation type.
+A compile-time error occurs if the representation type of an extension type
+declaration is a bottom type.
```
lrhn, line 770: That's more than what I was asking for, but probably not too much. It's not losing any actually useful functionality.

Have to make sure that the phrasing refers to the (type-alias-expansion of) the declared representation type, and not the actual representation type after type parameter instantiation of the extension type.

That is, `extension type E<T>(T v) { ... }` is valid, since `T` is not a bottom type, and `E<Never>` is valid because it's going to be impossible to prevent anyway. 

Which also means that if you really, really want a `Never` represenation type, you can do:
```dart
extension type _E<T>(T it) { ... }
typedef E = _E<Never>;
```
so we're not  *preventing* a bottom representation type, we're just preventing the extension type declaration itself from *knowing* (and using) that the type is a bottom type.

(That was what I was trying to achieve in #3383 by disallowing implementing interface types when the representation type is a bottom type, but for safety, it should probably disallow implementing extension types as well, at which point we might as well disallow the bottom representation type entirely. It's easier to explain.)

lrhn (approved):
LGTM if the phrasing means what it's supposed to mean.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Added commentary to explain this. No changes to normative text was needed.

PR #3396, opened 2023-10-12
State that special-case `num` typing rules do not apply to extension type members

lrhn:
The special-case rules for typing, e.g, `int + int` does not apply if `+` denotes an extension type member,
even if the receiver is a subtype of `int`.

### DIFF
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -15,6 +15,10 @@ information about the process, including in their change logs.
 [1]: https://github.com/dart-lang/language/blob/master/working/1426-extension-types/feature-specification-views.md
 [2]: https://github.com/dart-lang/language/blob/master/working/extension_structs/overview.md
 
+2023.10.12
+  - State that special-case `num` typing only applies to non-extension members
+    invocations.
+
 2023.09.11
   - Add missing rule about getter/setter signature correspondence.
 
@@ -601,6 +605,28 @@ If _Dm_ is a method with function type `F`, and `args` exists, the static
 analysis of the extension type member invocation is the same as that of an
 invocation with argument part `args` of a function with the given type.
 
+### Special-case Typing Rules for `num` Members
+
+The language specification contains special typing rules for invocations
+of `+`, `-`, `*`, `%`, `remainder` and `clamp` methods on 
+subtypes of `num`, `int` and `double`.
+
+Those rules do *not* apply to invocations of extension or extension type
+members.
+The rules are otherwise unchanged when applied to non-extension type
+member invocations, which will necessarily invoke the corresponding members
+of an instance of `int` or `double`.
+
+_It's not currently possible to have a subtype of `num` which has
+an extension member for those member names, but if it ever becomes possible,
+the rules will still not apply to such an invocation._
+
+_It's currently also not possible to introduce a subtype of `num`, `int` or
+`double`, which does not have the same member signatures for the special-cased
+members as the type that it subtypes, so the special-case rules are safe to
+assume that signature. If we introduce the ability to narrow or change the
+signatures of non-extension-type members of extension types, 
+the special-case typing rules will have to be re-visited._
 
 ### Dynamic Semantics of an Extension Type Member Invocation
 


accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -601,6 +605,28 @@ If _Dm_ is a method with function type `F`, and `args` exists, the static
...
+subtypes of `num`, `int` and `double`.
+
+Those rules do *not* apply to invocations of extension or extension type
+members.
+The rules are otherwise unchanged when applied to non-extension type
+member invocations, which will necessarily invoke the corresponding members
+of an instance of `int` or `double`.
+
+_It's not currently possible to have a subtype of `num` which has
+an extension member for those member names, but if it ever becomes possible,
```
eernstg, line 621:
```dart
extension E on int {
  String operator +(String other) => '$other, ';
  String get g => '${this + 'Hello'}world!';
}

void main() {
  print(42.g);
}
```

Seems that it should be possible to have exactly one of those extension members, but it would not be possible to invoke it implicitly except in the body of the extension declaration. It looks like the explicitly resolved extension member invocation `E(42) + 'Hello'` _does_ get the correct treatment, but the current behavior of both the CFE and the analyzer is to report an error for the implicitly resolved case.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -601,6 +605,28 @@ If _Dm_ is a method with function type `F`, and `args` exists, the static
...
+the rules will still not apply to such an invocation._
+
+_It's currently also not possible to introduce a subtype of `num`, `int` or
```
eernstg, line 624:
```suggestion
_It's currently also not possible to introduce a non-extension type which is a subtype of `num`, `int` or
```

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -601,6 +605,28 @@ If _Dm_ is a method with function type `F`, and `args` exists, the static
...
+`double`, which does not have the same member signatures for the special-cased
+members as the type that it subtypes, so the special-case rules are safe to
+assume that signature. If we introduce the ability to narrow or change the
+signatures of non-extension-type members of extension types, 
+the special-case typing rules will have to be re-visited._
```
eernstg, line 629: Perhaps:
```suggestion
the special-case typing rules may have to be re-visited._
```

We could claim that this case has already been covered: An extension type subtype of `int`, say, can certainly declare an extension type operator `+` with a different signature and implement it.

It would be a new thing if we allow, for example, an _abstract_ declaration with the given name in an extension type declaration to modify the static analysis of a member whose implementation is provided by a non-extension superinterface. However, I'd suggest that we treat this case as an implicitly induced forwarding method; for instance, it would be confusing if such a member is torn off and the resulting function object doesn't have the function type of the member as declared.

eernstg (approved):
LGTM, with a few suggestions.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -601,6 +605,28 @@ If _Dm_ is a method with function type `F`, and `args` exists, the static
 analysis of the extension type member invocation is the same as that of an
 invocation with argument part `args` of a function with the given type.
 
+### Special-case Typing Rules for `num` Members
+
+The language specification contains special typing rules for invocations
+of `+`, `-`, `*`, `%`, `remainder` and `clamp` methods on 
+subtypes of `num`, `int` and `double`.
```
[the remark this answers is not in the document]
lrhn: Or just say "on instances of"...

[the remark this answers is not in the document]
eernstg: I think this thread can just be settled by using one of the suggested wordings.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Sorry, that's not true: `this + 'Hello'` does _not_ perform a lexical lookup for operator `+`, it is an invocation with an explicit receiver (`this`, in an extension member body, which means that it will be subject to normal resolution and run the instance operator).

Closing.

eernstg:
This one has had an LGTM for a long time. I just closed a couple of discussion threads, and I don't think there's anything controversial in the remaining ones. So let's get it landed.

PR #3398, opened 2023-10-12
Tweak some wordings in the extension type specification.

lrhn:
Tweak some wordings in the extension type specification to be more precise.

Only actual change is saying that it's an early error to declare an abstract member.


eernstg (approved):
LGTM!

PR #3400, opened 2023-10-13
[extension-types] Add error for duplicate `implements` type

eernstg:
This PR adds an error for `implements ... T ... T ...` which is identical to the corresponding rule about classes, mixins, and enums. It is still an opinion rather than a soundness requirement or any "hard" requirement like that, but we might as well have it everywhere, and then we can change it everywhere in the future, if needed.

### DIFF
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -1127,6 +1127,12 @@ A compile-time error occurs if any direct or indirect superinterface of
 _DV_ denotes the declaration _DV_. *As usual, subtype cycles are not
 allowed.*
 
+It is a compile-time error if two elements in the type list of
+the `implements` clause of an extension type specifies the same type.
+
+*This rule against duplicates is identical to the existing rule about 
+classes, mixins, and enums.*
+
 Assume that _DV_ has two direct or indirect superinterfaces of the form
 <code>W\<T<sub>1</sub>, .. T<sub>k</sub>&gt;</code>
 respectively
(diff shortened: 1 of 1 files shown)


accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1127,6 +1127,12 @@ A compile-time error occurs if any direct or indirect superinterface of
 _DV_ denotes the declaration _DV_. *As usual, subtype cycles are not
 allowed.*
 
+It is a compile-time error if two elements in the type list of
+the `implements` clause of an extension type specifies the same type.
```
lrhn, line 1131: extension type -> extension type declaration

I really want to make it explicit that it's what is about what is written in the declaration, to avoid issues like we've had (has?) with mixin applications and desugaring.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1127,6 +1127,12 @@ A compile-time error occurs if any direct or indirect superinterface of
...
+
+*This rule against duplicates is identical to the existing rule about
```
lrhn, line 1133: It's not *identical* (it contains different words, like "extension type"), but it does serve the same purpose.

I'd change "about" here to "for". Mainly a feeling. The word "about" is vague.

Consider:

> *This rule against duplicate types mirrors the similar rule for class, mixin and enum declarations.*

lrhn (approved):

accepted/future-releases/extension-types/feature-specification.md
eernstg: Of course! Done.

PR #3402, opened 2023-10-16
Correct rule about Object/Object? being a superinterface

eernstg:
The existing rule about `Object` or `Object?` being a superinterface of an extension type was too restrictive. One case which shows that it needs to be broadened is this issue: https://github.com/dart-lang/co19/issues/2313.

This PR changes the rule to include a greater set of cases, thus ensuring that `Object` or `Object?` is actually found in the superinterface graph of an extension type in all cases.

It could be claimed that the broader rule introduced by this PR is surprisingly aggressive (we will add an edge in the superinterface graph from every single extension type to `Object` or to `Object?`, depending on whether the representation type is non-nullable or not). However, we do need to have this edge in some cases, and it is benign to have the edge even in cases where there is already a path to `Object`/`Object?` via other superinterfaces.

### DIFF
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -1114,9 +1114,8 @@ Assume that _DV_ is an extension type declaration named `Name`, and
 `V1` occurs as one of the `<type>`s in the `<interfaces>` of _DV_. In
 this case we say that `V1` is a _superinterface_ of _DV_.
 
-If _DV_ does not include an `<interfaces>` clause then _DV_ has
-`Object?` or `Object` as a direct superinterface, according to the subtype
-relation which was specified earlier.
+_DV_ has `Object?` or `Object` as a direct superinterface, according to the
+subtype relation which was specified earlier.
 
 A compile-time error occurs if `V1` is a type name or a parameterized type
 which occurs as a superinterface in an extension type declaration _DV_, and
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -1114,9 +1114,6 @@ Assume that _DV_ is an extension type declaration named `Name`, and
 `V1` occurs as one of the `<type>`s in the `<interfaces>` of _DV_. In
 this case we say that `V1` is a _superinterface_ of _DV_.
 
-_DV_ has `Object?` or `Object` as a direct superinterface, according to the
-subtype relation which was specified earlier.
-
 A compile-time error occurs if `V1` is a type name or a parameterized type
 which occurs as a superinterface in an extension type declaration _DV_, and
 `V1` denotes a non-extension type which is not a supertype of the
@@ -1129,6 +1126,12 @@ as having type `T`.*
 A compile-time error occurs if any direct or indirect superinterface of
 _DV_ denotes the declaration _DV_. *As usual, subtype cycles are not
 allowed.*
+
+Assume that `V` is an extension type of the form
+<code>Name\<T<sub>1</sub>, .. T<sub>k</sub>&gt;</code>
+where `Name` resolves to _DV_.
+`V` has `Object?` or `Object` as a direct superinterface, according to
+the subtype relation which was specified earlier.
 
 It is a compile-time error if two elements in the type list of the
 `implements` clause of an extension type declaration specifies the


accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1110,9 +1110,8 @@ Assume that _DV_ is an extension type declaration named `Name`, and
 `V1` occurs as one of the `<type>`s in the `<interfaces>` of _DV_. In
 this case we say that `V1` is a _superinterface_ of _DV_.
 
-If _DV_ does not include an `<interfaces>` clause then _DV_ has
-`Object?` or `Object` as a direct superinterface, according to the subtype
-relation which was specified earlier.
+_DV_ has `Object?` or `Object` as a direct superinterface, according to the
+subtype relation which was specified earlier.
```
lrhn, line 1114: I have at least two worries her.

First the practical: It's potentially a mistake to make `Object` be a direct superinterface.
Example:
```dart
class Bad {
  bool operator ==(covariant Bad other) => something;
}
extension type MoreBad(Bad _) implements Bad {
  /// ...
}
void main() {
  MoreBad(Bad()) == 1; // No error
```
By making `Object` a *direct* superinterface of `MoreBad`, the member signature of `operator==` should be the combined member signatures of `Bad` and `Object`, which is `bool operator==(covariant Object)`. 
That's probably not intended.

(I see the behavior above today if I add `, Object` to the implements list. Adding things to implements lists is not a no-op, even if the type already implements that interface.)

If being a direct superinterface does *not* mean that the member signature should be included in the combined member signature, it needs to be stated very, very clearly. And probably should just not use "direct superinterface" at all.

So, maybe don't include `Object` or `Object?` if not necesary. Something like:
> If _DV_ does not include an `<interfaces>` clause, then _DV_ has `Object` as direct superinterface
> if the declared representation representation type is non-nullable _(a subtype of `Object`)_,
> otherwise it has `Object?` as direct superinterface.
> If _DV_ does include an `<interfaces>` clause, then it has `Object` as direct superinterface if
> the declared representation representation type is non-nullable, and non of the declared 
> superinterfaces from the `<interfaces>` clause have `Object` as superinterface.

That is, we add a direct superinterface *only* if it's needed.

lrhn, line 1114: My second worry is more nit-picky.

Does it even make sense to say that _DV_, a *declaration*, has a superinterface *type*, direct or otherwise?

In the class modifiers spec, I introduced "declared interface" to distinguish the declaration from the type.
The problem here is also that `Object?` is not an interface. I'd be *much* less worried if we dropped the notion of `Object?` being treated like a nominative type, which can be a superinterface. It's unnecessary, and it makes things more complicated, not less, because it breaks down distinctions that have so far been clear.
(You can't have a structural type as superinterface. Except, now you can? What does that even mean.)

Even without that, with declarations like:
```dart
extension type E1<X extends int?>(X _) {}
extension type E2<X extends int>(E<X> _) {}
```
will `E1` or `E2` have `Object` as superinterface?
(My point being that it's a meaningless question without knowing `X`.)

We can probably say that `E1` does not have `Object` as direct superinterface,
and `E2` does, because `E1<X>` (a type, not a declaration) does have `Object` as supertype.

But what were we even *answering* when we said that `E1` does not have `Object` as direct superinterface?
When is that information needed?

And because of that, should we stop saying that `Object` and `Object?` are superinterfaces of the _declaration_, and make sure to say it only about the *types*, where it's just a superinterface, not a direct superinterface (I think).

eernstg: I think it's fair to say that this has been resolved.

eernstg:
PTAL!

I corrected the error about "this declaration has that superinterface".

However, I did not change the superinterface graph structure (that is, the text still says that every extension type has `Object` or `Object?` as a direct superinterface, even in the cases where the given superinterface can already be reached via some other path in the superinterface graph).

I do think it's a surprising property of extension types, but I also think the very small and very well-known set of `Object` members will not create much confusion by having this extra superinterface, and if it does affect anyone at all then it is probably more confusing if the existence of this direct superinterface depends on the entire superinterface graph.

Note that the disruption of the member signatures will basically only affect operator `==` and `noSuchMethod`, and only in the case where the parameter is made covariant, and that's very, very unlikely to be useful with `noSuchMethod`.

The getters are not affected, because the combined member signature will be the getter with the most special return type, which is surely also what we'd expect and want.

So operator `==` with a covariant parameter is the only case that could create real confusion, and this is the same kind of confusion that we already have with covariant member signatures (the non-covariant one generally wins).

accepted/future-releases/extension-types/feature-specification.md
eernstg: It does get more complex, but we could add a case like this:

> No warning is emitted for a downcast from an extension type `E1` to another extension type `E2` in the case where the instantiated representation type of `E1` differs from the instantiated representation type of `E2`.

We also considered the possibility that an extension type can declare that it is 'open' by means of a metadata annotation:

```dart
@openExtensionType
extension type Foo(int it) {
  // For some reason, we don't want to export the interface of int,
  // so we don't have `implements int`. But we _do_ want to allow
  // casting into this extension type: "Absolutely any `int` will do".
  ... // Members.
}

void main() {
  var foo = Foo(42);
  int i = foo; // Compile-time error, not a subtype.
  foo = i as Foo; // OK, no lint, `Foo` is open.
  var xs = [1, 2, 3];
  var ys = xs as List<Foo>; // OK, no lint.
}
```

The lint would treat type tests/casts/matches on open extension types like regular type tests/casts/matches on the representation type.

eernstg:
Abandoned: https://github.com/dart-lang/language/issues/3415 solves several tricky issues, and in particular this update is obsolete given that the language team has accepted the proposal in #3415.

Issue #3397, opened 2023-10-12
Allow extension type to implement its representation type.


lrhn:
TL;DR: I don't think I can explain to users why the pattern `extension type E(T _) implements T { moreMembers() }` is not allowed if `T` is an extension type, because there is no reason to prevent it. It's a valid and sound pattern, which today is only allowed if `T` is not an extension type.

## Proposal

Change the rule for when an extension type declaration can implement another extension type to:

> Assume that an extension type declaration _DV_ named `Name` has
> representation type `R`, and that the extension type `V1` with
> declaration _DV1_ is a superinterface of _DV_ (*note that `V1` may
> have some actual type arguments*).  Assume that `S` is the
> instantiated representation type corresponding to `V1`. A compile-time
> error occurs if `R` is not a subtype of `S` and `R` is not a subtype of _DV1_.

The only change is the addition of "and `R` is not a subtype of _DV1_".

## Reasoning:

_The proposed change is the *minimal* needed change, but it would be conceptually cleaner to make the change in the rule that allows extension type to implement a **non-extension-type** supertype of its representation type, because the change is really about removing the "non-extension type" limit from **that** rule._

I expect to see extension type with `implements` clauses mainly using one of two fundamental patterns:
* Additive: Implement/expose the representation type, add more members.
* Parallel: Implement another extension type, whose representation type is a supertype of your representation type.

The pattern for adding more members to a type using extension types is:
```dart
extension type MoreFoo(Foo _) implements Foo {
  moreMembers() {}
}
```
That currently only works if `Foo` is a non-extension type. (See https://github.com/dart-lang/sdk/issues/53739).

The parallel case would be something like:
```dart
extension type FancyWidget(Widget _) {
  // something
}
extension type FancyListWidget(ListWidget _) implements FancyWidget {
 // something
}
```
where the extension types *wrap* and mirror the type hierarchy of the representation types.

I expect users to understand extension types mainly in terms of those two patterns of usage,
and I believe the two patterns should be supported equally and independently, 
meaning that an extension type `E` with representation type `R` may implement a type `T` iff:
* `T` is not one of the types we disallow explicitly (structural types, `Null`, etc.), and 
* `R` \<: `T` (Additive case) or
* `T` is an extension type with representation type `S` and `R` \<: `S` (Parallel case).

[...]

PR #3405, opened 2023-10-17
[extension types] Relax rule on representation/implements type relation

eernstg:
Based on the discussion in https://github.com/dart-lang/language/issues/3397, this PR introduces a relaxation in the rule about the required typing relationship between the representation type `R2` of an extension type `E2` and an implemented extension type `E1` with instantiated representation type `R1`:

The existing rule requires that `R2` is a subtype of `R1`. The new rule includes the old rule, but also allows the situation where `R2` is a subtype of `E1`.

Example, allowed before and now:

```dart
extension type FancyWidget(Widget _) {
  // Something
}

extension type FancyListWidget(ListWidget _) implements FancyWidget {
  // Something
}
```

Example, allowed now (but used to be an error):

```dart
extension type FancyString(String it) implements String {
  // Something
}

extension type EvenFancierString(FancyString it) implements FancyString {
  // Something
}
```

The point is that this allows us to use the same idiom for superinterfaces that are extension types and superinterfaces that are non-extension types: Both `FancyString` and `EvenFancierString` use the following template:

```dart
extension type B(A it) implements A {}
```

### DIFF
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -15,6 +15,11 @@ information about the process, including in their change logs.
 [1]: https://github.com/dart-lang/language/blob/master/working/1426-extension-types/feature-specification-views.md
 [2]: https://github.com/dart-lang/language/blob/master/working/extension_structs/overview.md
 
+2023.10.17
+  - Allow an extension type to have `implements T` where `T` is a
+    supertype of the representation type (the old rule only allows
+    it when the representation type of `T` is a supertype).
+
 2023.10.16
   - Add error for `implements ... T ... T ...`, like the corresponding
     error with classes, mixins, etc.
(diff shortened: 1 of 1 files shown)


accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1158,7 +1163,7 @@ representation type `R`, and that the extension type `V1` with
 declaration _DV1_ is a superinterface of _DV_ (*note that `V1` may
 have some actual type arguments*).  Assume that `S` is the
 instantiated representation type corresponding to `V1`. A compile-time
-error occurs if `R` is not a subtype of `S`.
+error occurs if `R` is not a subtype of `S` nor a subtype of `V1`.
```
lrhn, line 1166: Nit: use "neither" to match "nor".
Otherwise use "and not".

lrhn (approved):

accepted/future-releases/extension-types/feature-specification.md
eernstg: Of course, done!

eernstg:
I put this topic on the agenda for the language team meeting today.

PR #3411, opened 2023-10-18
[extension types] Specify promotion of representation variables

eernstg:
This PR specifies how promotion of the representation variable of an extension type proceeds: It is applicable to any representation variable whose name is private, and it does not interfere with promotion of private class instance variables, or vice versa.

### DIFF
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -15,6 +15,10 @@ information about the process, including in their change logs.
 [1]: https://github.com/dart-lang/language/blob/master/working/1426-extension-types/feature-specification-views.md
 [2]: https://github.com/dart-lang/language/blob/master/working/extension_structs/overview.md
 
+2023.10.18
+  - Specify how to promote the representation variable of an extension
+    type.
+
 2023.10.16
   - Add error for `implements ... T ... T ...`, like the corresponding
     error with classes, mixins, etc.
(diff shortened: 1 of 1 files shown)


accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -701,6 +705,17 @@ Member Conflicts' occur as well in an extension type declaration.
 and has an instance member named `V`, and it is a compile-time error if it
 has a type parameter named `X` and it has an instance member named `X`.*
 
+If the representation name of an extension type is a private name `_n` then
+the representation variable is subject to promotion according to the rules
+about promotion of private instance variables, except that promotion is
+_not_ eliminated by the existence of other declarations (of any kind) named
+`_n` in the same library.
```
lrhn, line 712: If we had promotion of private _non-instance_, aka. statically resolved, variables, we should combine it into that instead of into "instance member access". But we don't. (Right?)

lrhn (approved):

accepted/future-releases/extension-types/feature-specification.md
eernstg: Right, we do not currently promote any library or static variables.

> Do we need concepts to distinguish between _virtual_ instance members and _statically resolved_ instance members?

I usually mention the kind of enclosing declaration: "a class / mixin / mixin class / enum instance member" vs. "an extension / extension type instance member". It might be useful to generalize this, but I don't think it's a big deal.

PR #3412, opened 2023-10-18
Add error for non-covariant type variable in superinterface

eernstg:
This PR adds an error  to extension types that corresponds to a similar rule for classes, mixins, etc: It is a compile-time error for a type variable declared by an extension type to occur in a non-covariant position of a superinterface. For example:

```dart
extension type E1<X>(int i) {}
extension type E2<X>(int i) implements E1<void Function(X)> {} // Error: `X` occurs contravariantly.
```


natebosch (approved):

eernstg:
Thanks!

sgrekhov:
https://github.com/dart-lang/co19/issues/2317

Issue #3258, opened 2023-08-09
[extension-types] Ambiguity in grammar


eernstg:
The implementation of extension types currently allows trailing commas in the `<representationSpecification>`. The `<representationSpecification>` is the grammatical construct which is very likely to be a temporary placeholder for a primary constructor. It seems reasonable to be well prepared for this generalization, that is, to build on parts of the primary constructor syntax as far as possible, and then restrict it using (non-syntax) compile-time errors. One of the likely consequences of this generalization is that trailing commas will be allowed (e.g., `extension type E(int i,) {}`).

However, @lrhn noticed an ambiguity that arises if we allow this trailing comma:

```dart
extension type on (int i,) {}
```

This could be an `extension` named `type` whose on-type is the record type `(int i,)`, but it could also be an `extension type` named `on` with a primary constructor with a trailing comma.

I'd recommend that we consider accepting a (presumably) small amount of breakage and make `type` a built-in identifier.

Con: If we do this then every `extension` declaration named `type` will break. I have not found any.

Pro: This eliminates the ambiguity, and it does not use any special non-CFG mechanisms (it does not introduce a new disambiguation rule, or depend on whitespace, etc).

@dart-lang/language-team, WDYT?

eernstg:
I think this is a small issue, but I do have some reasons to lean toward making `type` a built-in identifier.

> not making `type` a built-in identifier ...
> That may not work directly for your parser generator,

ANTLRv4 doesn't have syntactic predicates, but that's just a tool thing which should not matter at all for the decision about the language. (I do have a work-around, too.)

I'm worried about the overall properties of the Dart syntax: Every single time we introduce a non-CFG mechanism for syntactic disambiguation (or indeed for any purpose), we introduce issues in relation to language processing by anyone other than the parsers that are part of the analyzer and common front end. A mechanism that happens to be easy to implement in those parsers may be a substantial hurdle for, say, the Tree-Sitter parser generator, for any LR (including LALR) parser generator, etc.

So I tend to think about this as a kind of budget that we shouldn't waste (as in "don't gratuitously make Dart even less CFG-parsable than it is already").

Another consideration could be that it would make sense to make `type` a built-in identifier if we wish to use it for other constructs (say, any kind of declaration where `type` occurs as a keyword).

> I don't know if we restrict extensions to not be named using a built-in identifier

We do. But I do think you're right that it isn't strictly necessary (I couldn't easily come up with situations where it causes parsing difficulties).

[...]

PR #3431, opened 2023-10-25
Update the grammar to prevent an extension from having the name `type`

eernstg:
This PR adjusts the Dart grammar such that the name of an `extension` declaration can not be `type`, without turning `type` into a built-in identifier. This eliminates an ambiguity where the same declaration could at the same time be an extension type declaration and an extension declaration. See https://github.com/dart-lang/language/issues/3258 for more details.

### DIFF
--- a/specification/dart.sty
+++ b/specification/dart.sty
@@ -62,6 +62,7 @@
 \def\THROW{\keyword{throw}}
 \def\TRUE{\keyword{true}}
 \def\TRY{\keyword{try}}
+\def\TYPE{\keyword{type}}
 \def\VAR{\keyword{var}}
 \def\VOID{\keyword{void}}
 \def\WHILE{\keyword{while}}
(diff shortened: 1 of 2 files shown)


lrhn:
LGTM.

Is this a breaking change or a language versioned change?
That is, will it affect language versions prior to releasing extension types?
(I'm fine with it being breaking.)

If not breaking, it should probably not go into the spec before the rest of extension types.

lrhn (approved):

eernstg:
It is a breaking change at this time or later, because we could have a declaration of the form `extension type on (T v,) {...}` which is currently an extension named `type`. That declaration would be a syntax error unless and until the extension type feature is enabled, and it would then surely be have some compile-time errors because the declaration would need to use `this` differently when it is parsed as an extension type. Also, every call-site for members declared by this extension would almost inevitably turn into compile-time errors.

Nevertheless, I cannot see any occurrences of `extension type on` anywhere, and it seems likely that there are no such declarations in the first place.

So it's a breaking change no matter what, and we might as well deal with it now.

eernstg:
OK, I initiated the breaking change process: https://github.com/dart-lang/sdk/issues/53883.

Issue #3415, opened 2023-10-18
Make extension type non-nullability not be implicit.


lrhn:
The current extension type specification makes an extension type be non-nullable (subtype of `Object`) whenever its representation type is non-nullable.

That made more sense when it wasn't possible to implement non-extension types.

Now you can, so it's possible to write `implements Object` if being a subtype of `Object` is desired, and allowed by the declared representation type being non-nullable.

Making the "being a subtype of `Object`" implicit has two issues:

* It's not possible to *not* be a subtype of `Object` if the representation type is non-nullable. There is no opt-out.
* There are generic extension type declarations where being a subtype of `Object` depends on the *instantiation*, not the declaration.

The former has the effect that changing representation type may also changes non-nullability. 
While the representation type of an extension type is not strongly encapsulated, it's still something authors could recommend/instruct people to not depend on, so that the representation can be changed as a non-breaking change later. (If someone did depend on it, and got broken, they'd be using implementation details that were explicitly not covered by semver promises, aka, asking for it.)
For example, changing a representation type of `extension type Maybe<T>` from `Option<T>` to `T?` would change whether `Maybe<int>` was assignable to `Object` or not.

The latter is more of an implementation complication. 
For all other types, whether it's assignable to `Object` can be determined by looking at the _kind_ of type it is, every type except `T?`, `FutureOr<T>` and type variables are non-nullable, `T?` is not, `FutureOr<T>` is iff `T` is, and a type variable is if its bound or promotion is, which is just structural recursion (type parmeter bounds being a little less trivial). 
But for an extension type, it's necessary to look at the actual instantiation, push that through into the representation type, and then look at that type again. It's not limited to structural recursion, which is likely why https://github.com/dart-lang/language/issues/3263 happens.

So, **should we change that**?

That is: Don't make any extension type automatically implement `Object`.

You can still add `implements Object` explicitly if you want to, if the declared representation type is non-nullable, just as you can always implement any other interface that is a super-type of the representation type.

What you can't do then, is having
```dart
extension type E<T>(T value) { ... }
```
which is a subtype of `Object` if `T` is, and not if it isn't. An `E<int>(1)` won't be assignable to `Object`.
I'm not sure that's actually a problem in practice. It's also not assignable to `int`, and haven't made any attempt to make it be so. In fact, it's a feature that it *isn't*.

[...]

PR #3434, opened 2023-10-26
Change extension types such that they must declare subtype relations to `Object`

eernstg:
See the proposal in https://github.com/dart-lang/language/issues/3415. This PR changes the extension type feature specification such that the nullability of the representation type does not matter for the determination of the superinterface graph and subtype relationships of the types expressed based on a given extension type declaration:

An extension type is always a proper subtype of `Object?`, and as such it is potentially nullable (unless some further subtype relations exist). An extension type can declare that it `implements T` for various `T`, and this may cause the extension type to be a subtype of `Object` (directly or indirectly). In this case the extension type is non-nullable.

This change is a substantial simplification because the superinterface graph is shared among of all types `E<T1 .. Tk>` where `E` resolves to an extension type declaration (which includes the non-generic case when `k` is `0`). 

(In contrast, with the current feature specification, we could have an extension type `E` such that the superinterface graph of `E<int>` was different from the superinterface graph of `E<int?>`. They will now always have the same superinterface graph, though possibly with different actual type arguments in each node of the graph.)

### DIFF
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -988,9 +988,10 @@ non-nullable, unless it implements `Object` or a subtype thereof
 (like all other expressions). Non-extension types (except bottom types and
 `dynamic`) cannot be assigned to extension types without an explicit cast.
 Similarly, the null object cannot be assigned to an extension type without
-an explicit cast (or a constructor invocation). Since an extension type is
-potentially non-nullable, an instance variable whose type is an extension
-type must be initialized. It will not be implicitly initialized to null.*
+an explicit cast (or if it has a static type which is an extension type,
+e.g., `E(null)`). Since an extension type is potentially non-nullable, an
+instance variable whose type is an extension type must be initialized. It
+will not be implicitly initialized to null.*
 
 In the body of a member of an extension type declaration _DV_ named
 `Name` and declaring the type parameters
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -976,24 +976,18 @@ declaration itself, or we're talking about a particular generic
 instantiation of an extension type. *For non-generic extension type
 declarations, the representation type is the same in either case.*
 
-Let `V` be an extension type of the form
-<code>Name\<T<sub>1</sub>, .. T<sub>s</sub>&gt;</code>, and let
-`R` be the corresponding instantiated representation type.  If `R` is
-non-nullable then `V` is a proper subtype of `Object`, and `V` is
-non-nullable.  Otherwise, `V` is a proper subtype of `Object?`, and
-`V` is potentially nullable.
+An extension type `V` is a proper subtype of `Object?`. It is potentially
+non-nullable, unless it implements `Object` or a subtype thereof
+*(as described in the section about extension types with superinterfaces)*.
 
 *That is, an expression of an extension type can be assigned to a top type
-(like all other expressions), and if the representation type is
-non-nullable then it can also be assigned to `Object`. Non-extension types
-(except bottom types and `dynamic`) cannot be assigned to extension types
-without an explicit cast. Similarly, null cannot be assigned to an
-extension type without an explicit cast, even in the case where the
-representation type is nullable (even better: don't use a cast, call a
-constructor instead). Another consequence of the fact that the extension
-type is potentially non-nullable is that it is an error to have an instance
-variable whose type is an extension type, and then relying on implicit
-initialization to null.*
+(like all other expressions). Non-extension types (except bottom types and
+`dynamic`) cannot be assigned to extension types without an explicit cast.
+Similarly, null cannot be assigned to an extension type without an explicit
+cast (even better: don't use a cast, call a constructor instead). Another
+consequence of the fact that the extension type is potentially non-nullable is
+that it is an error to have an instance variable whose type is an extension
+type, and then relying on implicit initialization to null.*
 
 In the body of a member of an extension type declaration _DV_ named
 `Name` and declaring the type parameters
@@ -1137,8 +1131,7 @@ Assume that _DV_ is an extension type declaration named `Name`, and
 this case we say that `V1` is a _superinterface_ of _DV_.
 
 If _DV_ does not include an `<interfaces>` clause then _DV_ has
-`Object?` or `Object` as a direct superinterface, according to the subtype
-relation which was specified earlier.
+`Object?` as a direct superinterface.
 
 A compile-time error occurs if `V1` is a type name or a parameterized type
 which occurs as a superinterface in an extension type declaration _DV_, and
@@ -1303,12 +1296,7 @@ rather than from `Object`)*.
 
 *This change is needed because some extension types are subtypes of
 `Object?` and not subtypes of `Object`, and they need to have a
-well-defined depth. Note that the depth of an extension type can be
-determined by its actual type arguments, if any, because type parameters of
-the extension type may occur in its representation type. In particular, the
-depth of an extension type is a property of the type itself, and it is not
-always possible to determine the depth from the associated extension type
-declaration.*
+well-defined depth.*
 
 
 ## Dynamic Semantics of Extension Types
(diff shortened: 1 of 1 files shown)


eernstg:
@scheglov, @johnniwinther, @chloestefantsova, the language team accepted this simplification of the rules about the superinterface graph structure of an extension type. 

Sorry about the wasted work which has been done in order to handle the previous ruleset! Nevertheless, I'm sure we will be happy in the future about having these new, simpler rules.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -976,24 +976,18 @@ declaration itself, or we're talking about a particular generic
...
-without an explicit cast. Similarly, null cannot be assigned to an
-extension type without an explicit cast, even in the case where the
-representation type is nullable (even better: don't use a cast, call a
-constructor instead). Another consequence of the fact that the extension
-type is potentially non-nullable is that it is an error to have an instance
-variable whose type is an extension type, and then relying on implicit
-initialization to null.*
+(like all other expressions). Non-extension types (except bottom types and
+`dynamic`) cannot be assigned to extension types without an explicit cast.
+Similarly, null cannot be assigned to an extension type without an explicit
```
lrhn, line 986: Complete nit: I'd code quote the "null":  `` `null` ``

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -976,24 +976,18 @@ declaration itself, or we're talking about a particular generic
...
+cast (even better: don't use a cast, call a constructor instead). Another
```
lrhn, line 987: No need to editorialize. :)

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -976,24 +976,18 @@ declaration itself, or we're talking about a particular generic
...
+consequence of the fact that the extension type is potentially non-nullable is
+that it is an error to have an instance variable whose type is an extension
```
lrhn, line 989: Long sentence.

Consider changing 
> Another consequence of the fact that the extension type is potentially non-nullable is that it ....
to
> Because an extensiont type is potentially nullable, it

Or maybe even continue:

> Because an extension type is potentially non-nullable, 
> an instance variable whose type is an extension type must be initialized.
> It will not be implicitly initialized to `null`.

You can't "rely on implicit initialization to null", when implicit initializtion to null doesn't happen. You can try, but that's not what the code means.
It's better to say what is, than what could have been, had things been different, in some unspecified way (and not another unspecified way).

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1137,8 +1131,7 @@ Assume that _DV_ is an extension type declaration named `Name`, and
 this case we say that `V1` is a _superinterface_ of _DV_.
 
 If _DV_ does not include an `<interfaces>` clause then _DV_ has
-`Object?` or `Object` as a direct superinterface, according to the subtype
-relation which was specified earlier.
+`Object?` as a direct superinterface.
```
lrhn, line 1134: I'd still drop this. 
It will just not have any direct superinterface. And that's OK.

(I'm still not sure what it means, since `Object?` is not an interface.)

lrhn (approved):

accepted/future-releases/extension-types/feature-specification.md
eernstg: Actually, I'm using the word 'null' rather than the source code snippet '`null`' because the topic here is the semantic value associated with that syntax (and with many other code elements with varying syntactic forms, e.g., an expression whose type is of the form `T?`).

So the generality would go down, but the readability could of course go up.

Not sure. Haven't changed anything yet.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Right, removed.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1303,12 +1296,7 @@ rather than from `Object`)*.
 
 *This change is needed because some extension types are subtypes of
 `Object?` and not subtypes of `Object`, and they need to have a
-well-defined depth. Note that the depth of an extension type can be
-determined by its actual type arguments, if any, because type parameters of
-the extension type may occur in its representation type. In particular, the
-depth of an extension type is a property of the type itself, and it is not
-always possible to determine the depth from the associated extension type
-declaration.*
+well-defined depth.*
```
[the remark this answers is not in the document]
eernstg: But this means that the depth of an extension type is defined if it has `Object` as an explicitly declared superinterface (directly or indirectly, via other extension types or via other non-extension types), and it is undefined otherwise.

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -1377,21 +1365,20 @@ used as an expression *(also known as the ultimate representation type)*.
...
-  particular, `o is V0` and `o as V0` have the same dynamic
-  semantics as `o is R` respectively `o as R`, and
+- *At run time, the type `V0` is identical to the extension type erasure
+  `R0`. In particular, `o is V0` and `o as V0` have the same dynamic
+  semantics as `o is R0` respectively `o as R0`, and
   `t1 == t2` evaluates to true if `t1` is a `Type` that reifies
-  `V0` and `t2` reifies `R`, and the equality also holds if
-  `t1` and `t2` reify types where `V0` and `R` occur as subterms
-  (e.g., `List<V0>` is equal to `List<R>`).*
+  `V0` and `t2` reifies `R0`, and the equality also holds if
```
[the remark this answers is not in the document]
eernstg: I changed several things in this item, at was somewhat clunky. However, I left one thing unchanged:

> "`Type` that reifies `V0`" is not a thing

An extension type `E` has a run-time representation and semantics (which is the same as that of the extension type erasure of `E`), which in turn corresponds to a specific set of instances of `Type` (a single canonical instance in some cases).

I don't actually think it's going to help anybody if we insist that we should say "let `t1` be a `Type` that reifies the extension type erasure of `V0`" rather than "let `t1` be a `Type` that reifies `V0`".

Developers will see `return T;` or similar constructs where we usually talk about a `Type` that reifies the type `T`, and this relationship doesn't get destroyed by the fact that we take two steps to get from an extension type to the corresponding `Type`, rather than one. In any case, `V0` is a type in a Dart program, and there is a corresponding set of `Type` instances at run time, and we usually use the word 'reify' to characterize that relationship.

accepted/future-releases/extension-types/feature-specification.md
lrhn: In that case, use "the null value". The word "null" by itself is just too ambiguous by itself (IMO).

accepted/future-releases/extension-types/feature-specification.md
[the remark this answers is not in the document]
lrhn: In which case we can define *depth* for extension types with no non-extension type superinterfaces too, without needing to change changing the super-interface hierarchy.

> Let `T` be a type declared by of a class, mixin, enum or extension type declaration. let *S* be the set of direct > superinterfaces of `T`. Then we define *depth*(`T`) as:
> * *depth*(`T`) = 0 if *S* is empty. `T` has no direct superinterfaces. _This includes `Object`, `Null` and any extension type without an `implements` clause._
> * *depth*(`T`) = max{`I` in *S*| *depth*(`I`)} + 1 if *S* is not empty.

This defines a *depth* function on all the nominative types, without introducing `Object?` as an interface (which it isn't) and without touching the superinterface relation. Minimal change, but enough.
The only other needed change is that LUB can now fail to find a shared superinterface.
And that's OK, in that case **UP** is `Object?`.

Still, *all* changes are entirely local to the LUB algorithm (which I believe is the only thing using *depth*), so there should be no risk of accidental or undesired side-effects elsewhere.

accepted/future-releases/extension-types/feature-specification.md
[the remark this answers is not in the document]
lrhn: We needs some kind of verbage for this, probably in order to write the runtime semantics too.

If we have a type literal expression *e*, the runtime semantics of evaluating it should be something like:
* Let *T* be the static type denoted by the type expression *e*. (This includes type inference of missing type variables).
* Let *R* be the runtime type corresponding to *T*. *R* is created from *T* by:
    * Perform extension type erasure on *T* to get *T'*. 
    * If *T'* contains type variables, those are replaced by the runtime type currently bound to that type variable.
    * The result of that is *R*.
* Let *v* be a constant `Type` object reifying *R*.

The "runtime type corresponding to *T*" is needed for type variables already. Don't know how we've phrased it.

accepted/future-releases/extension-types/feature-specification.md
eernstg: https://en.wikipedia.org/wiki/Deferent_and_epicycle

accepted/future-releases/extension-types/feature-specification.md
eernstg: Yes: We have used 'the null object' to make it obvious that it isn't syntax, I'll do that.

accepted/future-releases/extension-types/feature-specification.md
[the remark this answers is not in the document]
eernstg: We can define a notion of depth, but it doesn't stop there. We need a special exception to handle the choice of a result: It will no longer work to say "take the singleton set with the greatest depth", we need to say "take the singleton set with the greatest depth; if no such set exists then take `Object?`". For example, if we have an extension type with no `<interfaces>` and some other type then the entire set of superinterfaces is empty; we still follow the rule and take `Object?`, but the explanation involving empty sets and a special exceptional treatment of `Object?` is simply not intuitive.

accepted/future-releases/extension-types/feature-specification.md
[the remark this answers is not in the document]
eernstg: > The "runtime type corresponding to T" is needed for type variables already. Don't know how we've phrased it.

Right, we do have that part, it's described as 'the actual type' [here](https://github.com/dart-lang/language/blob/8a4003d2d773e324d1a2ac7c9bbb553260f5cd5b/specification/dartLangSpec.tex#L22769).

accepted/future-releases/extension-types/feature-specification.md
eernstg: OK, dropped it.

accepted/future-releases/extension-types/feature-specification.md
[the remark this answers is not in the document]
eernstg: Note that the 'Summary of Typing Relationships' is commentary, there is no normative text in that section.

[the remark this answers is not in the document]
eernstg: I simplified this considerably, and it doesn't talk about reification any more.

eernstg:
Is this one ready to land? The text still says that for the purpose of computing the Dart 1 least upper bound, we pretend that the superinterface graph has a top node. That's simple to say and simple to understand, but the existence of the top node may be controversial, even though it's clearly labeled as being only for this single purpose.

PR #3435, opened 2023-10-27
Move up the intersection type rules in the definition of `UP`

eernstg:
This PR moves the rules about intersection types up to an earlier position in the definition of `UP`, such that no results are of the form `(X & B)?`. (Types of that form should never exist, so `UP` shouldn't produce them.)

This is a much nicer solution than https://github.com/dart-lang/language/pull/2427, but it may have to be considered as a breaking change.

We should probably try it out and see how much breaks. Otherwise we can adopt #2427.

[Edit: The `X extends B` rules are no longer moved along with the `X & B` rules, that was a more substantial breaking change, and not useful.]

### DIFF
--- a/resources/type-system/upper-lower-bounds.md
+++ b/resources/type-system/upper-lower-bounds.md
@@ -85,11 +85,25 @@ We define the upper bound of two types T1 and T2 to be **UP**(`T1`,`T2`) as foll
 - **UP**(`T1`, `T2`) = `T2` if **BOTTOM**(`T1`)
 - **UP**(`T1`, `T2`) = `T1` if **BOTTOM**(`T2`)
 
+- **UP**(`X1 extends B1`, `T2`) =
+  - `T2` if `X1 <: T2`
+  - otherwise `X1` if `T2 <: X1`
+  - otherwise **UP**(`B1a`, `T2`)
+    where `B1a` is the greatest closure of `B1` with respect to `X1`,
+    as defined in [inference.md].
+
 - **UP**(`X1 & B1`, `T2`) =
   - `T2` if `X1 <: T2`
   - otherwise `X1` if `T2 <: X1`
   - otherwise **UP**(`B1a`, `T2`)
     where `B1a` is the greatest closure of `B1` with respect to `X1`,
+    as defined in [inference.md].
+
+- **UP**(`T1`, `X2 extends B2`) =
+  - `X2` if `T1 <: X2`
+  - otherwise `T1` if `X2 <: T1`
+  - otherwise **UP**(`T1`, `B2a`)
+    where `B2a` is the greatest closure of `B2` with respect to `X2`,
     as defined in [inference.md].
 
 - **UP**(`T1`, `X2 & B2`) =
@@ -136,20 +150,6 @@ We define the upper bound of two types T1 and T2 to be **UP**(`T1`,`T2`) as foll
 - **UP**(`T1?`, `T2?`) = `S?` where `S` is **UP**(`T1`, `T2`)
 - **UP**(`T1?`, `T2`) = `S?` where `S` is **UP**(`T1`, `T2`)
 - **UP**(`T1`, `T2?`) = `S?` where `S` is **UP**(`T1`, `T2`)
-
-- **UP**(`X1 extends B1`, `T2`) =
-  - `T2` if `X1 <: T2`
-  - otherwise `X1` if `T2 <: X1`
-  - otherwise **UP**(`B1a`, `T2`)
-    where `B1a` is the greatest closure of `B1` with respect to `X1`,
-    as defined in [inference.md].
-
-- **UP**(`T1`, `X2 extends B2`) =
-  - `X2` if `T1 <: X2`
-  - otherwise `T1` if `X2 <: T1`
-  - otherwise **UP**(`T1`, `B2a`)
-    where `B2a` is the greatest closure of `B2` with respect to `X2`,
-    as defined in [inference.md].
 
 - **UP**(`T Function<...>(...)`, `Function`) = `Function`
 - **UP**(`Function`, `T Function<...>(...)`) = `Function`
--- a/resources/type-system/upper-lower-bounds.md
+++ b/resources/type-system/upper-lower-bounds.md
@@ -85,25 +85,11 @@ We define the upper bound of two types T1 and T2 to be **UP**(`T1`,`T2`) as foll
 - **UP**(`T1`, `T2`) = `T2` if **BOTTOM**(`T1`)
 - **UP**(`T1`, `T2`) = `T1` if **BOTTOM**(`T2`)
 
-- **UP**(`X1 extends B1`, `T2`) =
-  - `T2` if `X1 <: T2`
-  - otherwise `X1` if `T2 <: X1`
-  - otherwise **UP**(`B1a`, `T2`)
-    where `B1a` is the greatest closure of `B1` with respect to `X1`,
-    as defined in [inference.md].
-
 - **UP**(`X1 & B1`, `T2`) =
   - `T2` if `X1 <: T2`
   - otherwise `X1` if `T2 <: X1`
   - otherwise **UP**(`B1a`, `T2`)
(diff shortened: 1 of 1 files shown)


eernstg:
Here's an example that shows a core difference between the models in play:

```dart
void f<X extends num?>(bool b, X x) {
  if (x is int) { // `x` promoted to `X & int`.
    var y = b ? x : null;
    // Type of `y`:
    //   Current behavior: `(X & int)?`.
    //   #2427: `X?`.
    //   #3435: `int?`.
  }
}
```

eernstg:
Note [this comment](https://github.com/dart-lang/language/pull/2427#issuecomment-1783765948) on #2427 which could be taken to imply that this PR (#3435) does seem to be a breaking change for the treatment of `Null`, but this is in a sense a good kind of breakage because it makes the treatment of `Null` and the treatment of all other types more consistent.

resources/type-system/upper-lower-bounds.md
```diff
@@ -81,6 +85,34 @@ We define the upper bound of two types T1 and T2 to be **UP**(`T1`,`T2`) as foll
 - **UP**(`T1`, `T2`) = `T2` if **BOTTOM**(`T1`)
 - **UP**(`T1`, `T2`) = `T1` if **BOTTOM**(`T2`)
 
+- **UP**(`X1 extends B1`, `T2`) =
```
lrhn, line 88: Do we need to define what `X extends B` means, "a type variable `X` with bound `B`"?

It's not the syntax we use for type variables otherwise.

resources/type-system/upper-lower-bounds.md
```diff
@@ -81,6 +85,34 @@ We define the upper bound of two types T1 and T2 to be **UP**(`T1`,`T2`) as foll
...
+  - `T2` if `X1 <: T2`
+  - otherwise `X1` if `T2 <: X1`
+  - otherwise **UP**(`B1a`, `T2`)
+    where `B1a` is the greatest closure of `B1` with respect to `X1`,
+    as defined in [inference.md].
+
+- **UP**(`X1 & B1`, `T2`) =
+  - `T2` if `X1 <: T2`
+  - otherwise `X1` if `T2 <: X1`
```
lrhn, line 97: We could get `X1 & B1` if `T2 <: B1`.

resources/type-system/upper-lower-bounds.md
```diff
@@ -81,6 +85,34 @@ We define the upper bound of two types T1 and T2 to be **UP**(`T1`,`T2`) as foll
 - **UP**(`T1`, `T2`) = `T2` if **BOTTOM**(`T1`)
 - **UP**(`T1`, `T2`) = `T1` if **BOTTOM**(`T2`)
 
+- **UP**(`X1 extends B1`, `T2`) =
+  - `T2` if `X1 <: T2`
+  - otherwise `X1` if `T2 <: X1`
+  - otherwise **UP**(`B1a`, `T2`)
+    where `B1a` is the greatest closure of `B1` with respect to `X1`,
```
lrhn, line 92: (Why do we need the greatest closure? Is it to avoid bad recursion?)

resources/type-system/upper-lower-bounds.md
```diff
@@ -81,6 +85,34 @@ We define the upper bound of two types T1 and T2 to be **UP**(`T1`,`T2`) as foll
...
+  - otherwise **UP**(`T1`, `B2a`)
+    where `B2a` is the greatest closure of `B2` with respect to `X2`,
+    as defined in [inference.md].
+
+- **UP**(`T1`, `X2 & B2`) =
+  - `X2` if `T1 <: X2`
+  - otherwise `T1` if `X2 <: T1`
+  - otherwise **UP**(`T1`, `B2a`)
+    where `B2a` is the greatest closure of `B2` with respect to `X2`,
+    as defined in [inference.md].
```
lrhn, line 114: Could we have a rule of
- **UP**(`X & B1`, `X & B2`) =
  - `X & B3` if `B3` <: `B0`
  - `X` otherwise
  where `B3` = **UP**(`B1`, `B2`), `B0` is the bound of `X`.

With the rules above, we always end up at the `X` in that case.

lrhn (commented):
SGTM to me, if the change is accepted.

resources/type-system/upper-lower-bounds.md
eernstg: We have this notation in several documents (at least the NNBD feature specification and this document). I haven't used it in the language specification (the update with null safety replaces it by an explicit introduction of the bound), so it will presumably be replaced like that as various feature specifications are integrated into the language specification. Hence, there's no need to change it here.

eernstg: True, it is inconsistent that we write `X extends B` to indicate that a certain type denoted by `X` is a type variable and its declaration specifies `B` as the bound, and then we use `X & B2` to denote a promoted type variable. 

However, I don't think we have to worry about this because it will be reworded to introduce the bound explicitly in natural language ('Assume that `X` is a type variable with bound `B`') and then we'll just use `X` in the rules.

resources/type-system/upper-lower-bounds.md
eernstg: The argument was that we'd need to obtain a type that doesn't depend on `X`, and hence it would be a problem if `B1` is an F-bound that contains `X`.

resources/type-system/upper-lower-bounds.md
eernstg: Right, I think we could have that. It's probably not going to be used very often, but it seems possible, and it would improve on the typing in those situations where it is applicable.

eernstg (commented):
Responded to review comments.

This would presumably await a discussion at the language team meeting, but it would be fine if we can have a version which is ready to land if and when the underlying proposal is accepted by the language team.

resources/type-system/upper-lower-bounds.md
lrhn: ACK.

resources/type-system/upper-lower-bounds.md
lrhn: Maybe, some day.

lrhn (approved):
OK if we accept the change.

stereotype441 (approved):
LGTM! This feels much more principled to me than #2427, and I agree that in your example from https://github.com/dart-lang/language/pull/3435#issuecomment-1783052483, inferring a type of `int?` for `y` seems more useful than `X?`.

I agree that we should try this out and measure the amount of breakage (which I expect to be small).

lrhn:
I already tried it: https://dart-review.googlesource.com/c/sdk/+/333005

Couldn't build the SDK. The error is that `UP(X extends Object?, null)` becomes `Object?`, not `X?`. We may need to special-case an `X` with a nullable bound.

eernstg:
> I already tried it: https://dart-review.googlesource.com/c/sdk/+/333005

Cool! ;-)

We could move just the intersection type rules up, and keep the `X extends B` rules where they used to be. I moved them up together because it seems surprising that those rules are identical in so many ways, but they get a different treatment because of this ordering choice. Nevertheless, if it works then we could do that.

eernstg:
I moved the type variable rules as mentioned; this means that the PR moves the intersection type rules up before the `NULL` rules, and nothing else.

eernstg:
Here is a CL that makes the changes as currently specified in this PR (so we're only moving the intersection type rules, not the rules about type variables without promotion): https://dart-review.googlesource.com/c/sdk/+/333922.

The change caused compile-time errors in 6 locations in the SDK (`{identity,linked,custom}_hash_map.dart`), and they have been corrected. They are all marked `// coerce undefined to null.`, and it seems likely that this kind of typing situation does not arise in "normal" code.

eernstg:
Said CL https://dart-review.googlesource.com/c/sdk/+/333922 breaks a couple of tests specifically about UP, and a couple of front-end tests where a type of the form `X? & Object?` occur (that might just be the format chosen for `(X & Object)?` when kernel code is prettyprinted). If we agree that those types should not exist then it is basically non-breaking in the SDK.

Issue #3348, opened 2023-09-14
Consider changing shadowing rules for macro generated code


jakemac53:
Relevant specification section https://github.com/dart-lang/language/blob/main/working/macros/feature-specification.md#shadowing-declarations.

tldr; it is currently an error for a macro to generate code which changes how some identifier in the same library would have been resolved.

In talking with @scheglov I think I have been convinced that we should not make that an error, but instead just allow it. Effectively, this is option 2 in the linked section "Re-resolve all identifiers after macros are applied, which may change what they resolve to.", except that is a bit misleading since we shouldn't ever have to re-resolve. And in fact, making it an error forces at least 2 resolves to detect the error.

jakemac53:
cc @munificent @leafpetersen thoughts?

jakemac53:
This also makes macro augmentations libraries behave identically to, and have the same capabilities as, user written augmentations libraries.

jakemac53:
There is an issue here related to annotation evaluation. We currently say this:

```
Because macros are not allowed to generate code that shadows an identifier
in the same library, we know that if an annotation class or any arguments to it
could be resolved, then we can assume that resolution is correct.

This allows us to provide an API for macro authors to attempt to evaluate an
annotation in _any phase_. The API may fail (if it requires more macro
expansion to be done), but that is not expected to be a common situation. In
the case where it does fail, users should typically be able to move some of
their code to a separate library (which they import). Then things from that
library can safely be used in annotations in the current library, and evaluated
by macros.
```

Which would no longer be valid.

scheglov:
Other possible approach is to prevent already observed identifiers to change their meaning. The meaning of "observed" can be either explicit, e.g. whether a macro generator actually asked for resolution of an annotation, or implicit, e.g. whether an identifier was used as a type annotation, that we resolved after the types phase, but then shadowed by declaring a top-level function (the naming convention does not like it, but technically possible). So, these observed identifiers should be re-checked, but not `=> x` body that we initially talked about - nobody saw it, or even could see it, so we accept that the users intention was to use whatever it happens at the end. The key is consistency throughout phases.

jakemac53:
[...]

PR #3452, opened 2023-11-07
change shadowing rules to remove need for multiple resolutions

jakemac53:
Closes https://github.com/dart-lang/language/issues/3348

cc @scheglov

### DIFF
--- a/working/macros/feature-specification.md
+++ b/working/macros/feature-specification.md
@@ -950,8 +950,10 @@ later resolved to a different declaration.
 
 In order to resolve these discrepancies we add this rule:
 
-- **It is an error for a macro to add a declaration which shadows any**
-  **previously resolved identifier.**.
+- **It is a compile time error for a macro to add a declaration which shadows**
+  **any previously resolved identifier.**. These errors occur after a macro
+  runs, when the compiler is merging in the macro results, and so it is not
+  catchable or detectable by macros.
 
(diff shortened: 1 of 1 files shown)


scheglov (approved):

working/macros/feature-specification.md
```diff
@@ -907,68 +907,56 @@ has two colliding declarations.
...
 
-This follows from the general principle that macros should not alter the
-meaning of existing code. Adding the getter `x` in the example above shadows the
-top-level `x`, changing the meaning of the original code.
+In order to resolve these discrepancies we add this rule:
 
-Note, that if the getter were written as `int get y => this.x;`, then a macro
-*would* be allowed to introduce the new getter `x`, because `this.x` could not
-previously be resolved.
+- **It is an error for a macro to add a declaration which shadows any**
```
munificent, line 953: We should be clear what kind of error this is. Does this mean that the attempted constant evaluation during the macro's execution will throw a runtime (where "runtime" means "macro execution time") exception? If so, that SGTM, but it's probably worth making that explicit.

munificent (approved):

working/macros/feature-specification.md
jakemac53: I specified that it is a compile time error, and not catchable (since it happens after macro execution)

Issue #3460, opened 2023-11-10
[enhanced enums] Need new error about enum values and factory constructors


eernstg:
Consider the following program:

```dart
enum E {
  e1,
  e2.foo();

  const E();
  const factory E.foo() = ET.foo;
}

extension type const ET(E it) implements E {
  const ET.foo() : this(E.e1);
}

void main() {
  print(identical(E.e1, E.e2));
}
```

This program is currently (DartPad based on Dart SDK 3.3.0-16.0.dev) accepted by the analyzer, and reported as a confusing error by the CFE

```
Error compiling to JavaScript:
lib/main.dart:3:6:
Error: Couldn't find constructor 'E.foo'.
  e2.foo();
     ^^^
Error: Compilation failed.
```

I would say that we need a new error on enum declarations, something like this:

> It is a compile-time error if _entry_ is a term derived from `<enumEntry>` that occurs as one of the entries of an enum declaration, and the constructor invoked in order to obtain that value is not generative.

The point is that a non-generative constructor (that is, a factory constructor) in an enum declaration may (and indeed always will) return an existing object, which is not the appropriate semantics for an `<enumEntry>`.

Obviously, we should briefly consider this case, too:

```dart
enum E {
  e1.foo();

  const E();
  const factory E.foo() = ET.foo;
}

extension type const ET(E it) implements E {
  const ET.foo() : this(E.e1);
}

void main() {
  print(E.e1);
}
```

This is the loop where the object `E.e1` is defined in terms of itself. The analyzer reports this as a cyclic dependency, which is fine, but it would also be prevented by the new error. It would probably be fine to report the new error and not report the cyclic dependency, because we could say that the cyclic dependency never comes into existence when we refuse to even consider calling the factory constructor in order to obtain `E.e1`.

eernstg:
@dart-lang/language-team, WDYT? We should probably move rather quickly on this one in order to avoid having these surprising situations in real life when extension types are released.

eernstg:
Spec update proposal here: https://github.com/dart-lang/language/pull/3461.

lrhn:
Agree. And good catch.

It's a new issue due to extension types, so we should just make sure it's part of the extension types implementation.

It didn't use to be necessary to say, because it wasn't possible due to the other restrictions on enums.

eernstg:
[...]

PR #3461, opened 2023-11-10
Add error about factory constructor generating an enum value

eernstg:
It is possible to use a constant redirecting generative constructor in an extension type together with a constant redirecting factory constructor in an `enum` declaration to create enumerated values of that enum that aren't distinct.

This PR introduces a compile-time error for enumerated value declarations that give rise to an invocation of a factory constructor, such that this situation cannot arise.

See https://github.com/dart-lang/language/issues/3460 for further information.

### DIFF
--- a/accepted/2.17/enhanced-enums/feature-specification.md
+++ b/accepted/2.17/enhanced-enums/feature-specification.md
@@ -64,9 +64,11 @@ It is a **compile-time error** to refer to a declared or default generative cons
 * As the target of a redirecting generative constructor of the same `enum` declaration (`: this(...);`/`: this.targetName(...);`), or
 * Implicitly in the enum value declarations of the same `enum` (`enumValueName(args)`/`enumValueName.targetName(args)`).
 
-_No-one is allowed to invoke a generative constructor and create an instance of the `enum` other than the enumerated enum values. 
-That also means that a redirecting *factory* constructor cannot redirect to a generative constructor of an `enum`,
-and therefore no factory constructor of an `enum` declaration can be `const`, because a `const` factory constructor must redirect to a generative constructor._
+_No-one is allowed to invoke a generative constructor and create an instance of the `enum` other than the enumerated enum values._
+
+It is a **compile-time error** if a declaration of an enumerated value of an `enum` declaration gives rise to an invocation of a factory constructor.
+
+_A redirecting factory constructor can redirect to a generative constructor of an `extension type`, which may in turn return one of the existing values of that `enum`. This must be an error, because the enumerated values of an `enum` declaration must be distinct._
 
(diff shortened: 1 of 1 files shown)


accepted/2.17/enhanced-enums/feature-specification.md
```diff
@@ -64,9 +64,11 @@ It is a **compile-time error** to refer to a declared or default generative cons
 * As the target of a redirecting generative constructor of the same `enum` declaration (`: this(...);`/`: this.targetName(...);`), or
 * Implicitly in the enum value declarations of the same `enum` (`enumValueName(args)`/`enumValueName.targetName(args)`).
 
-_No-one is allowed to invoke a generative constructor and create an instance of the `enum` other than the enumerated enum values. 
-That also means that a redirecting *factory* constructor cannot redirect to a generative constructor of an `enum`,
-and therefore no factory constructor of an `enum` declaration can be `const`, because a `const` factory constructor must redirect to a generative constructor._
+_No-one is allowed to invoke a generative constructor and create an instance of the `enum` other than the enumerated enum values._
+
+It is a **compile-time error** if a declaration of an enumerated value of an `enum` declaration gives rise to an invocation of a factory constructor.
```
lrhn, line 69: "Gives rise to" is vague.

Maybe:
> It is a **compile-time error** if the constructor implicitly invoked by an enum value declaration is a factory constructor.

I think this change can be moved into the extension type specification, because it wasn't an issue before.

It should not be breaking, since it was impossible to have an enum value target a non-`const` constructor, or to have a `const` factory constructor which was not redirecting, and a redirecting constructor had to eventually reach a generative constructor of the class, or a subclass. There are no subclasses, and the factory constructor is not allowed to refer to the generative constructors.

(And now there are subtypes with generative constructors which won't necessarily need to create new objects. Another perfect case of breaking previous implicit invariants.)

(But that also means it's not breaking to put it into this spec, since it changes nothing. We just need to be sure implementors see the change as part of implementing extension types.)

lrhn (approved):

accepted/2.17/enhanced-enums/feature-specification.md
eernstg: > .. is vague.

Indeed, now using the proposed text except that I'm using 'enumerated enum value declaration' rather than just 'enum value declaration', in order to remain consistent with other phrases in the vicinity.

> (But that also means it's not breaking to put it into this spec, since it changes nothing. We just need to be sure implementors see the change as part of implementing extension types.)

I think we can ensure that.

eernstg:
@sgrekhov, could you keep an eye on this PR? If it is landed then we'd need tests, in particular a couple of tests similar to the examples here: https://github.com/dart-lang/language/issues/3460#issue-1987292756.

sgrekhov:
> @sgrekhov, could you keep an eye on this PR? If it is landed then we'd need tests, in particular a couple of tests similar to the examples here: [#3460 (comment)](https://github.com/dart-lang/language/issues/3460#issue-1987292756).

Of course!

eernstg:
@sgrekhov, this one is ready now! ;-)

PR #3480, opened 2023-11-23
Update macro spec to allow more introspection in Phase 2.

### LINKED ISSUE #3442: Introspect members of other types in the declarations phase
`built_value` is one example of when this is needed, as it creates a new class with fields corresponding to, but not identical to, the macro target.

```
@BuiltValue()
class Foo {
  final int x;
}
```

-->

```
class FooBuilder {
  int? x;
}
```

This is problematic because the declaration phase can add as well as introspect declarations. One possible approach would be to make introspecting another type's members wait until the declaration phase of all macros have finished for that type.

davidmorgan:
Fix #3442

### DIFF
--- a/working/macros/feature-specification.md
+++ b/working/macros/feature-specification.md
@@ -273,7 +273,9 @@ whatever the application order. In most cases this achieved by the split into
 phases: within each phase macros can run in any order because the output is not
 visible to other macros until the next phase. As a special case, introspection
 of types in Phase 2 waits as needed for other macro applications to complete,
-failing if there are cycles.
+failing if there is a cycle.
+
+TODO: give an example of a cycle.
 
 ### Augmentation library structure and ordering
 
@@ -590,10 +592,7 @@ returning, meaning it waits for the macro applications on the target type to
 finish.
 
 If a cycle arises in macro applications waiting for other macro applications to
-complete then a special value is returned indicating that introspection is not
-possible due to a cycle. If the macro can still correctly apply then it should
-continue normally. If it cannot succeed without the introspection results then
-it should report an error.
+complete then a `StateError ` is thrown.
 
 Rules might be added in future to decide in some specific cases which macro
 should run with incomplete introspection results to break a cycle. For example,
--- a/working/macros/feature-specification.md
+++ b/working/macros/feature-specification.md
@@ -268,36 +268,12 @@ order of macros:
 
     Here, the macros applied to C are run `First()`, `Second()`, then `Third()`.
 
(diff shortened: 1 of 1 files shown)


working/macros/feature-specification.md
```diff
@@ -606,17 +582,24 @@ of a function or initializer for a variable. It is encouraged to provide a body
...
-TODO: Define the introspection rules for extension types.
+Phase two macros can introspect on all of the members of a type. If the type
+to be introspected is declared in the same library cycle and has one or more
+macros applied to it then this introduces an ordering constraint between the
+macro applications: the introspection call waits for complete results before
+returning, meaning it waits for the macro applications on the target type to
+finish.
+
+If a cycle arises in macro applications waiting for other macro applications to
+complete then a special value is returned indicating that introspection is not
```
jakemac53, line 593: I do think it makes sense to throw an error here, we can't easily model a special value otherwise (I don't want to go down the path of Option types or something).

working/macros/feature-specification.md
```diff
@@ -268,36 +268,12 @@ order of macros:
...
-Aside from these rules, macro introspection is limited so that evaluation order
-is not user visible. For example, if two macros are applied to two methods in
-the same class, there is no way for those macros to interfere with each other
-such that the application order can be detected.
+Aside from these rules, macros are constrained so that the result is the same
+whatever the application order. In most cases this achieved by the split into
+phases: within each phase macros can run in any order because the output is not
+visible to other macros until the next phase. As a special case, introspection
+of types in Phase 2 waits as needed for other macro applications to complete,
+failing if there are cycles.
```
jakemac53, line 276: We may want to provide an example of a cycle here also, maybe add a TODO?

davidmorgan: Done.

working/macros/feature-specification.md
jakemac53: Fwiw I am totally fine with making this an Error instead of an Exception, to indicate that it should not be caught/handled? I don't think many macros would be able to actually do anything meaningful.

I mostly just think the easiest way to surface this error (and terminate the macro), is to just have the API throw. I don't think we really have to worry if some macro author does actually try to handle the error, they can't do anything _harmful_, its just that they probably can't do anything _useful_ (but maybe some could, who knows).

davidmorgan: Sounds good--how about like this? (just throw `StateError`).

Thanks.

jakemac53 (approved):
LGTM

davidmorgan:
Thanks :)

PR #3483, opened 2023-11-24
Specify ordering of dynamic error for p.f(...) where p is a deferred prefix

eernstg:
This PR changes the specified ordering of the dynamic error that occurs if `p` is a deferred prefix of a non-loaded deferred library, and an invocation of the form `p.f(...)` is executed: The error occurs before any actual arguments to `f` are evaluated.

This change is motivated by the fact that the actual implemented behavior is as specified _with_ this PR, whereas the currently specified behavior differs from the actual behavior.

This is not a breaking change because the actual behavior will not change, we are simply adjusting the specification to say what the implementations already do, and have indeed done for several years.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -41,6 +41,11 @@
 % version of the language which will actually be specified by the next stable
 % release of this document.
 %
+% Nov 2023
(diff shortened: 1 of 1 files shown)


lrhn (approved):

eernstg:
Thanks!

Issue #3488, opened 2023-11-29
Soundness issue with cast patterns


eernstg:
Consider the following program:

```dart
sealed class A {
  bool get b => true;
}

class B1 extends A {}
class B2 extends A {}

class Cheat implements B1, B2 {
  bool get b => false;
}

void main() {
  A a = Cheat();
  int i = switch (a) {
    B1(b: true) as B1 => 1,
    B2(b: true) as B2 => 2,
  };
  print(i); // 'null'.
}
```

The program contains a switch expression which is accepted at compile time, that is, its set of cases is classified as exhaustive. However, no case matches an object of type `A` whose `b` is false, such as an instance of `Cheat`.

The problem is that the first case is considered to handle every instance of `B2` (because they are assumed to cause the cast to `B1` to throw) and the second case is considered to handle every instance of `B1` (because they are assumed to cause the cast to `B2` to throw), but an instance of `Cheat` will complete both casts normally, and then be rejected by the requirement that its `b` must be true.

At first, this is an implementation issue.

However, the specification is so similar that we can construct a variant of the above example which shows that there is an issue with the specification as well. The rule about the space obtained for a cast pattern is as follows:

> Formally, the space union `spaces` for a cast pattern with cast type `C` is a union of:
>
> 1.  The lifted space union of the cast's subpattern in context `C`.
> 2.  For each space `E` in the *expanded spaces* (see below) of `M`:
>   a.  If `E` is not a *subset* (see below) of `C` and `C` is not a subset of `M`, then the lifted space union of `E`.

The actual implementation requires that `C` is a subtype of `M` in order to take (2) into account at all (the spec has no such requirement), and then the implementation requires that `E` is not a subtype of `C` and `C` is not a subtype of `E` (at the same point, the spec requires that `C` is not a subtype of `M`).

So we have a discrepancy between the implementation and the specification, and it will not suffice to adjust the specification such that it follows the implementation (because the implementation has a soundness issue, as shown above).

However, it is also not sufficient to change the implementation such that it follows the specification precisely. The following example has a similar soundness issue, and it would be accepted with the specified cast pattern space:

```dart
sealed class A {
  bool get b => true;
}

class C1 {}
class C2 {}

class B1 extends A, implements C1 {}
class B2 extends A, implements C2 {} 

class Cheat implements B1, B2 { bool get b => false; }

void main() {
  A a = Cheat();
  int i = switch (a) {
    B1(b: true) as C1 => 1,
[...]

PR #3490, opened 2023-11-30
[patterns] Adjust cast pattern space definition

eernstg:
This PR changes the definition of the lifted space union of a cast pattern, cf. the discussion in https://github.com/dart-lang/sdk/issues/54125 and the soundness issue https://github.com/dart-lang/language/issues/3488. The specification is intended to be well aligned with the [upcoming implementation](https://dart-review.googlesource.com/c/sdk/+/333922).

### DIFF
--- a/accepted/3.0/patterns/exhaustiveness.md
+++ b/accepted/3.0/patterns/exhaustiveness.md
@@ -429,7 +429,7 @@ The lifted space union for a pattern with matched value type `M` is:
 
     test(A a) => switch (a) { // `a` should be a `B` or a `C`.
       B() => 0,
-      final b as C => 1,
+      _ as C => 1,
     };
     ```
 
--- a/accepted/3.0/patterns/exhaustiveness.md
+++ b/accepted/3.0/patterns/exhaustiveness.md
@@ -417,24 +417,24 @@ The lifted space union for a pattern with matched value type `M` is:
     exhaustiveness algorithm doesn't model "all objects that are not `int`", so
     it can't tell that this is exhaustive.*
 
-    *However, in the case where the matched value type is sealed we can
-    have a situation where a cast pattern is guaranteed to "handle" all
-    objects (that is, it will match or it will throw).*
+    *However, we can have a situation where a cast pattern is used to ensure
+    that a case "handles" all possible matched values because it will match
+    or it will throw.*
 
     ```dart
-    sealed class S {}
-    class A implements S {}
-    class B implements S {}
-    class C implements S {}
+    abstract class A {}
+    class B implements A {}
+    class C implements A {}
+    class D implements A {}
 
-    test(S s) => switch (s) { // `s` should be an `A` or a `B`.
-      A() => 0,
-      final b as B => 1,
+    test(A a) => switch (a) { // `a` should be a `B` or a `C`.
+      B() => 0,
+      final b as C => 1,
     };
     ```
 
-    *In this example, `test` should deal with instances of `A` and instances of
-    `B`, but it should simply throw if `s` has any other type. So we use the
+    *In this example, `test` should deal with instances of `B` and instances of
+    `C`, but it should simply throw if `a` has any other type. So we use the
     cast pattern in the last case to ensure that the switch is exhaustive, which
     is true because all instances of `B` will be handled by the subpattern, and
     all other objects will cause the cast to throw.*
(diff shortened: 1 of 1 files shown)


accepted/3.0/patterns/exhaustiveness.md
```diff
@@ -417,39 +417,37 @@ The lifted space union for a pattern with matched value type `M` is:
...
-    1.  The lifted space union of the cast's subpattern in context `C`.
+    Let `S` be the lifted space union of the cast's subpattern in context `C`.
 
-    2.  For each space `E` in the *expanded spaces* (see below) of `M`:
+    1.  If `C` is a *subset* (see below) of `S` then the result is the lifted
+        space union of `M`.
 
-        1.  If `E` is not a *subset* (see below) of `C` and `C` is not a subset
-            of `M`, then the lifted space union of `E`.
+    2.  Otherwise, the result is `S`.
```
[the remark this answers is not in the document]
eernstg: Oh, that's nice! Using that now.

accepted/3.0/patterns/exhaustiveness.md
```diff
@@ -417,39 +417,49 @@ The lifted space union for a pattern with matched value type `M` is:
...
+    is true because all instances of `B` will be handled by the subpattern, and
+    all other objects will cause the cast to throw.*
 
     Formally, the space union `spaces` for a cast pattern with cast type `C` is
-    a union of:
+    computed as follows:
+
+    Let `S` be the lifted space union of the cast's subpattern in context `C`.
+
+    1.  If `C` is a *subset* (see below) of `S` then the result is the lifted
```
johnniwinther, line 447: **subset** is not defined for a type and a space and the **space subsetting** definition is not adequate for our purpose here. 

A possible definition could be:

---
***A) Type subset of space union***
A type `C` is a **subset** of space union `S` iff

Either
1. `C` is **subset** of each single space `S_i` of `S` (see below), or
2. for each **subtype** `C_i` of `C` there exists a single space `S_i` of `S` such that `C_i` is a subset of `S_i`.

***B) Type subset of single space***
A type `C` is a **subset** of single space `S` iff
1. `C` is a subtype of the type `T` of `S`, 
2. `S` has an open restriction, and
3. For each property `p_j` of `S_i` with space `S_j` and static type `C_j`, `C_j` is a subset of space `S_j`.

***C) Subtypes of type***
The **subtypes** of a type `C` is
1. If `C` is sealed, the expanded subtypes is the enumerated subtypes of `C`.
2. Otherwise the singleton set `{C}`.
---
I'm sure we have a definition of "enumerated subtypes", though, but it is meant to be the subtypes found in the section "The declaration D of T is a sealed type" of [Expanding a space](https://github.com/dart-lang/language/blob/main/accepted/3.0/patterns/exhaustiveness.md#expanding-a-space) algorithm.

The motivation for these rules are:
```
sealed class M {}
class A extends M {}
sealed class B<T> extends M {}
class C<T> extends B<T> {}
class D<T, S> extends B<T> {}

method1(o) => switch (o) {
    (B() || B()) as B => 0, // Rules A.1, B.1 (is subset)
  };
method2(o) => switch (o) {
    (A() || B()) as M => 0, // Rules A.2, B.1, C.1 (is subset)
  };
method3(o) => switch (o) {
    [_] as List, // Rule B.2, (not subset)
    [...] as List, // Rule B.2, (is subset)
  };
method4(o) => switch (o) {
    int(isEven: true) as int, // Rules A.1, B.3, C.2 (not subset)
    int(: var isEven) as int, // Rules A.1, B.3, C.2 (is subset)
}
```
Do to the limitations of the "subtypes" algorithm used in C.1 we don't find this to be exhausting:
```
method5(o) => s
    (A() || C() || D()) as M => 0, // Rules A.2, B.1, C.1 (transitively)
  };
```
but this similar to this, which also don't find exhausting:
```
method6(M o) => s
    A() => 0,
    C() => 0,
    D() => 0,
  };
```

eernstg: @johnniwinther and I have discussed this topic today. I'll wait a bit, and then try to come up with a revised PR when the matter has stabilized.

eernstg: Thanks, I used this in the new commit (but with several changes). PTAL!

johnniwinther (approved):

PR #3494, opened 2023-12-06
Allow doc comments to use formal parameter initializer scope

parlough:
Since initializing formals are not added to the `formal parameter scope`, doc comments can't access them. If the `initializing formal parameter scope` exists, doc comments should be allowed to use them. That way constructors can reference super parameters.

I've never adjusted the spec before, so please review closely :) - For example, with this exception, is it not clear what the scope for other constructors is supposed be? Should the function one expand its text?

\cc @eernstg


specification/dartLangSpec.tex
```diff
@@ -23028,7 +23028,15 @@ \subsubsection{Comments}
...
-the declaration of a function $f$ is the formal parameter scope of $f$.
+the declaration of a function $f$ is the
+\IndexCustom{formal parameter scope}{scope!formal parameter} of $f$.
+
+\LMHash{}%
+The current scope for a documentation comment immediately preceding
+the declaration of a non-redirecting generative constructor $c$
+with initializing formals is the
+\IndexCustom{formal parameter initializer scope}{%
+  scope!formal parameter initializer}of $c$.
```
eernstg, line 23039: I'd recommend not creating an index entry for this page about 'formal parameter scope': This phrase is defined in the section 'Formal Parameters', and index entries in the language specification are created for the defining occurrence of the given phrase, not for usages. (We could have index entries for usages as well, but it would probably give rise to so many index entries that it would be unwieldy. Just search the document for usages using the 'Find' feature of the viewer.)

On the other hand, there are many references to the enclosing section when a phrase is used (that is, 'formal parameter scope' may be followed by a reference to the section 'Formal Parameters').

For some (unknown ;-) reason, constructors are usually called `$k$`.

So we could have something like this:

```suggestion
The current scope for a documentation comment immediately preceding
the declaration of a non-redirecting generative constructor $k$
with initializing formals is the formal parameter initializer scope of $k$
(\ref{generativeConstructors}).

\LMHash{}%
Otherwise, the current scope for a documentation comment immediately preceding
the declaration of a function $f$ is the formal parameter scope of $f$
(\ref{formalParameters}).
```

eernstg (commented):
Thanks, this looks good! However, I've suggested a few changes.

This is mostly in order to follow the pretty-much-unwritten rules of the language specification (like: "which occurrences of a phrase will give rise to an index entry?" or "where do we put line breaks in LaTeX source?"), but also one thing that affects the meaning directly: We should have a link between the generative constructor case and the "otherwise" case. I've suggested putting the narrowest case first and then covering everything else by starting the next paragraph with 'Otherwise'.

parlough:
Thanks for the feedback @eernstg, I applied your suggestions! Starting with the narrowest case makes a lot of sense, I didn't consider that :)

One final question: In terms of the spec, is the otherwise case stating "function" sufficient for also covering constructors not covered in the special case?

eernstg (approved):
LGTM, thanks!

The word 'function' does cover constructors: https://github.com/dart-lang/language/blob/0ecd65a8988520dd5bec097056d1657f7b4b6d54/specification/dartLangSpec.tex#L1944.

parlough:
@eernstg Feel free to land if you're happy with this change. I don't have commit access on this repository. Thanks again!

eernstg:
OK, thanks! Landing.

PR #3496, opened 2023-12-06
Adjust objectPattern to allow `Function()`

eernstg:
See https://github.com/dart-lang/sdk/issues/54246 where @modulovalue raised this issue.

This PR changes the syntax of `<objectPattern>` such that it allows `Function` to be the denotation of a type (for example: `Function(: var toString)`).

The analyzer and the common front end already implement this behavior, hence there's no implementation effort.


eernstg:
We discussed this at the language meeting today and decided to keep the currently implemented behavior (which means that we do adopt the specification change which is performed by this PR).

PR #3503, opened 2023-12-08
Specify that an `initializerExpression` cannot be a function literal

eernstg:
It has been a compile-time error (possibly a syntax error) for many years if an `initializerExpression` is a function literal. For example:

```dart
class A {
  var x;
  A() : x = () {}; // Error.
}
```

The reason for this is that the block, `{}`, could be the  body of the constructor (at least in the case where the formal parameter declarations can be an expression, which includes cases like `()` and `(x)` and `(x, y)`), and we will not know whether the block is the body of a function literal or the body of the constructor until we encounter (or do not encounter) a semicolon at the very end.

This rule has been around for a long time, and it is there for a good reason, but it wasn't specified. This CL adds a paragraph to the language specification that specifies this rule.

This CL also adds two missing cases to the derivation rule for `initializerExpression` (throw expressions and assignments). They need to be added in order to make the specification match the implementations.

No implementation effort is associated with this specification change: The tools already have the behavior which is specified after the change.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -45,7 +45,8 @@
 % - Allow `~/` on operands of type `double` in constant expressions, aligning
 %   the specification with already implemented behavior.
 % - Broaden the grammar rule about `initializerExpression` to match the
-%   implemented behavior.
+%   implemented behavior. Specify that an initializer expression can not be
+%   a function literal.
 %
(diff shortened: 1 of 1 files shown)


jakemac53 (approved):

eernstg:
Thanks!

PR #3515, opened 2023-12-14
Specify null safety subtyping

eernstg:
This PR contains the null-safety updates from #2605 (the overall null-safety PR) that are concerned with subtyping. The intention is that we process and land this PR, and #2605 will then be made smaller when it is rebased over this one, thus allowing us to deal with the null-safety update in smaller portions.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -23458,7 +23458,9 @@ when used to determine static semantic properties,
 like subtyping relationships
 (\ref{subtypes})
 or upper and lower bounds
-(\ref{standardUpperBoundsAndStandardLowerBounds}).
+(\ref{standardUpperBoundsAndStandardLowerBounds}),
+because it is concerned with relationships among program elements
+that occur in different binding environments.
 
 \commentary{%
   The notion of being the same type at run time is different from
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -2690,7 +2690,7 @@ they are considered to be the same type.
   different type parameters, F-bounds, and the types of formal parameters.
   However, we do not wish to distinguish between two function types
   if they have the same structure and only differ in the choice of names.
-  This treatment of names is also known as alpha-equivalence.%
+  This treatment of names is also known as alpha equivalence.%
 }
 
 \LMHash{}%
@@ -22092,24 +22092,18 @@ concrete syntax denoting $T$.
   However, we must eliminate the difficulties associated with
   different syntax denoting the same type,
   and different types denoted by the same syntax.
-  We do this by assuming that every type in the program has been expressed
-  in a manner where those situations never occur,
-  because each type is denoted by the same globally unique syntax everywhere.%
+  We do this by defining in detail what it means to be the same type
+  (\ref{sameStaticType}).%
 }
 
 \LMHash{}%
 In section~\ref{subtypes} and its subsections,
 all designations of types are considered to be the same
-if{}f they have the same canonical syntax
-(\ref{theCanonicalSyntaxOfTypes}).
+type if{}f they satisfy the criteria in
+Section~\ref{sameStaticType}.
 
-\commentary{%
-  Note that the canonical syntax also implies
-  transitive expansion of all type aliases
-  (\ref{typedef}).
-  In other words, subtyping rules do not need to consider type aliases,
-  because all type aliases have been expanded.%
-}
+\LMHash{}%
+The subtype rules assume that all type aliases have been transitively expanded.
 
 \LMHash{}%
 Every \synt{typeName} used in a type mentioned
@@ -22495,19 +22489,13 @@ come in two variants:
   coincide in this case.%
 }
 
-\LMHash{}%
-Two function types are considered equal if consistent renaming of type
-parameters can make them identical.
-
 \commentary{%
-  A common way to say this is that we do not distinguish
-  function types which are alpha-equivalent.
-  For the subtyping rule below this means we can assume that
-  a suitable renaming has already taken place.
-  In cases where this is not possible
-  because the number of type parameters in the two types differ
-  or the bounds are different,
-  no subtype relationship exists.%
+  Two function types are the same type
+  if, roughly, consistent renaming of type
+  parameters can make them identical.
+  This is also known as alpha equivalence.
+  The detailed rules are described in
+  Section~\ref{sameStaticType}.%
 }
 
 \LMHash{}%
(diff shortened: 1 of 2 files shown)


specification/dartLangSpec.tex
```diff
@@ -23130,156 +23146,238 @@ \subsection{Operator Precedence}
...
-if $T$ is a type variable or a function type.
-For several other forms of type, e.g.,
-a promoted type variable,
-\code{Object}, \DYNAMIC, \VOID,
-\code{FutureOr<$T$>} for any $T$, or \FUNCTION,
-it is known that it will never occur as $T$ for rule~\SrnSuperinterface,
-which means that this seemingly expensive step can be confined to some extent.%
-}
+  \commentary{%
+    Note that this rule is admissible, and can be safely elided if desired.%
```
lrhn, line 23247: Define "admissible". It's a term of art not used elsewhere in the specification, so we cannot assume the reader to be familiar with it.

lrhn (commented):
Can't really review this as source.
Do we still have a way to generate a rendering showing the changes?

eernstg:
[dartLangSpec-3515.pdf](https://github.com/user-attachments/files/17414000/dartLangSpec-3515.pdf)

eernstg:
Based on the output from `git diff main...`, here is an overview of the changes:

### Changed locations

#### Front matter

Only comments.

#### @@ -21633,128 +21631,188 @@

Section 'Subtypes': Now mentions intersection types.

Figure 'Subtype rules': Changed a lot.

Subsection 'Meta-Variables': Slightly changes presentation of
metavariables.

Subsection 'Subtype Rules': Introduces the 'canonical syntax' of a
type. Controversial! An alternative could be to say that subtype rules
are _not_ applicable to syntax, they are applicable to semantic
entities that are normalized in the same way as canonical syntax does
it. Extensively rewritten.

Subsection 'Being a Subtype': Extensively rewritten.

Subsubsection 'Informal subtype rule descriptions': Rewritten extensively.

Subsection 'Additional Subtyping Concepts': Extensively rewritten.

Section 'Function Types': Last paragraph modified.

Section 'Type Function': Rewritten.

#### @@ -22482,7 +22498,7 @@

Just the title of section 'Type dynamic' was adjusted, for consistency
(that is: titles do not use non-standard fonts).

Section 'Type FutureOr': Just adding missing `\Error{}` on static
errors, and updating `Object` to `Object?` in an example.

Section 'Type void': Removed commentary that reiterated an obsolete
definition of `Type` equality. Removed the item that enabled `await
expressionOfTypeVoid`, because that rule that enabled this usage has
been removed from the language. Example at the end is updated.

Subsection 'Void Soundness': Example updated, plus body text about the
example.

#### @@ -23324,156 +23324,238 @@

Section 'Appendix: Algorithmic Subtyping': Updated to use the
NNBD-updated presentation of the subtype rules.

eernstg:
I tried to use `latexdiff` to generate a PDF that shows the differences, but it doesn't handle "non-standard" LaTeX (which seems to be difficult to avoid). I can generate a PDF that shows some differences and stops at page 232, but it would take some more experimentation to produce something that shows the differences in a way that can be relied upon.

specification/dartLangSpec.tex
```diff
@@ -21633,128 +21631,188 @@ \subsection{Subtypes}
...
-Similarly, $X \& S$ may be seen as an intersection type,
-which is a subtype of $X$ and also a subtype of $S$.
-Intersection types are \emph{not} supported in general,
-only in this special case.%
-}
-Every other form of type may occur during static analysis
-as well as during execution,
-and the subtype relationship is always determined in the same way.
+and there are several other restrictions on where they can occur
+(\ref{intersectionTypes}).
```
eernstg, line 21852: This reference is dangling. The missing section will be added later.

specification/dartLangSpec.tex
```diff
@@ -21995,28 +22009,14 @@ \subsubsection{Being a subtype}
...
-  with formal type parameters \List{X}{1}{s}.
-  \commentary{%
-    Each formal type parameter $X_j$ may have a bound,
-    but the bounds are never used in this context,
-    so we do not introduce metavariables for them.%
-  }
-\end{itemize}
+In rules~\SrnCovariance{} and~\SrnNominal,
+the first premise is that the given name denotes an interface type
+(\ref{interfaceTypes})
```
eernstg, line 22229: This is a dangling reference, the missing section will be added later.

eernstg:
I added a section about 'being the same type', which means that the thing previously known as 'canonical syntax' (now renamed to 'explicitly resolved syntax') is encapsulated in this section.

eernstg (commented):
Added a couple of comments about section references which will be resolved in a separate PR where those sections are added.

specification/dartLangSpec.tex
```diff
@@ -21839,128 +21839,188 @@ \subsection{Subtypes}
...
-    \RuleRaw{\SrnTop}{Right Top}{T \in \{\code{Object}, \DYNAMIC, \VOID\}}{S}{T}
-    \RuleRaw{\SrnNull}{Left Null}{T \not= \bot}{\code{Null}}{T}
+    % ------------------------------------------------ Right Top
+    \RuleRaw{\SrnRightTop}{%
+      T \in \{\code{Object?}, \DYNAMIC, \VOID\}}{S}{T}
+    % ------------------------------------------------ Right Object
+    % TODO(eernst): Comment out this rule as redundant and
+    % renumber, it is an instance of \SrnRightFutureOrB.
+    \RuleRaw{\SrnRightObjectFour}{%
+      \mbox{$S$ is an interface type or \FUNCTION}}{S}{\code{Object}}
```
chloestefantsova, line 21946: Should it say the similar for `Record`?

specification/dartLangSpec.tex
```diff
@@ -22025,41 +22087,29 @@ \subsubsection{Subtype Rules}
...
+type if{}f they satisfy the criteria in
+Section~\ref{sameType}.
+
+\LMHash{}%
+The subtype rules assume that all type aliases have been transitively expanded.
+
+\LMHash{}%
+Every \synt{typeName} used in a type mentioned
+in section~\ref{subtypes} and its subsections
+is assumed to have no compile-time error,
```
chloestefantsova: Should it say "no compile-time errors" instead?

specification/dartLangSpec.tex
```diff
@@ -21839,128 +21839,188 @@ \subsection{Subtypes}
...
+    \RuleRaw{\SrnRightTop}{%
+      T \in \{\code{Object?}, \DYNAMIC, \VOID\}}{S}{T}
+    % ------------------------------------------------ Right Object
+    % TODO(eernst): Comment out this rule as redundant and
+    % renumber, it is an instance of \SrnRightFutureOrB.
+    \RuleRaw{\SrnRightObjectFour}{%
+      \mbox{$S$ is an interface type or \FUNCTION}}{S}{\code{Object}}
+    % ------------------------------------------------ Left Null 2
+    \Rule{\SrnNullTwo}{\code{Null}}{T}{%
+      \code{Null}}{\code{FutureOr<$T$>}}
```
chloestefantsova: Isn't `\SrnNullTwo` a special case of `\SrnRightFutureOrB`? If so, should it be removed as redundant?

specification/dartLangSpec.tex
```diff
@@ -21839,128 +21839,188 @@ \subsection{Subtypes}
...
-    \Rule{\SrnRightFutureOrA}{Right FutureOr A}{S}{\code{Future<$T$>}}{%
-      S}{\code{FutureOr<$T$>}}
-    \Rule{\SrnLeftPromotedVariable}{Left Promoted Variable B}{S}{T}{X \& S}{T}
-    \RuleRaw{\SrnRightFunction}{Right Function}{T\mbox{ is a function type}}{%
-      T}{\FUNCTION}
+    % ------------------------------------------------ Left Nullable
+    \RuleTwo{\SrnLeftNullable}{S}{T}{\code{Null}}{T}{%
+      \code{$S$?}}{T}
+    % ------------------------------------------------ Left Promoted Variable A
+    \Axiom{\SrnLeftPromotedVariableOne}{X \& S}{X}
```
chloestefantsova: Shouldn't `\SrnLeftPromotedVariableOne` be defined similarly to rule `\SrnLeftPromotedVariableTwo`? The latter is more general, and it seems that the rules for both sides of the intersection should be defined similarly. The updated rule `\SrnLeftPromotedVariableOne` can, for example, look as follows:

```
D |- X <: T
--------------------
D |- X & S <: T
```

specification/dartLangSpec.tex
```diff
@@ -22489,33 +22489,28 @@ \subsection{Function Types}
...
+  Two function types are the same type
+  if, roughly, consistent renaming of type
+  parameters can make them identical.
+  This is also known as alpha equivalence.
+  The detailed rules are described in
+  Section~\ref{sameType}.%
 }
 
 \LMHash{}%
 A function object is always an instance of some class $C$ that implements
```
chloestefantsova: I wonder what the phrase "an instance of some class $C$ that implements" contributes to the paragraph. Neither the front-ends nor the backends (as far as I know) create such an intermediate class that implements a function type. Could the phrase be omitted?

specification/dartLangSpec.tex
```diff
@@ -23529,157 +23668,266 @@ \subsection{Operator Precedence}
...
+    }).
+  \item
+    Otherwise \SubtypeNE{S}{T} is true.
+  \end{itemize}
+\item
+  \textbf{Left Null (\SrnRightFutureOrB*, \SrnNullTwo, \SrnNullOne):}
+  if $S$ is \code{Null} then:
+  \begin{itemize}
+  \item
+    if $T$ is a type variable (promoted or not) the query is false.
```
chloestefantsova, line 23806: Is there a reason besides optimization for this rule? If my understanding is correct, the case of a type variable will be handled in line 23752.

specification/dartLangSpec.tex
```diff
@@ -23529,157 +23668,266 @@ \subsection{Operator Precedence}
...
+    The requirement that \List{Z}{1}{k} are fresh names is as usual required
+    such that type variable names do not get captured.
+    It is valid to choose $Z_i$ to be $X_i$ or $Y_i$,
+    so long as capture is avoided.%
+  }
+
+\item
+  If $S$ is \code{($V_0$, \ldots, $V_n$, \{$V_{n+1} d_{n+1}$, \ldots, $V_m d_m$\})}
+  and $T$ is \code{($S_0$, \ldots, $S_n$, \{$S_{n+1} d_{n+1}$, \ldots, $S_m d_m$\})}
+  then \SubtypeNE{S}{T} if{}f \SubtypeNE{V_i}{S_i} for each $i \in 0 .. m$.
```
chloestefantsova: Rules for records are commented out elsewhere in the PR. Should lines 23917–23920 be commented out as well?

specification/dartLangSpec.tex
chloestefantsova: Specifically, it's unclear how `X & S <: B` can be proven without such a reformulation, in case when `X extends Y` and `Y extends B`.

specification/dartLangSpec.tex
eernstg: Yes. `Record` is not in the language at 2.13, but I added a TODO to handle this. I also removed the TODO which was there previously because it applies to `\SrnNullTwo` rather than `\SrnRightObjectFour`.

specification/dartLangSpec.tex
eernstg: Exactly, done.

specification/dartLangSpec.tex
eernstg: Yes, done.

specification/dartLangSpec.tex
eernstg: Done! No need to have weasel talk that doesn't even match the reality.

specification/dartLangSpec.tex
eernstg: Agreed, we need the more general form because we don't have a general transitivity rule. Done.

specification/dartLangSpec.tex
eernstg: Optimization is definitely one of the main reasons for having this ruleset &mdash; it is intended to show how Dart subtyping can be decided with good performance. The inference rules in figure 4 are intended to specify the intended relation with no concerns about decision speed. I think that's OK.

specification/dartLangSpec.tex

eernstg:
Hi @chloestefantsova, I responded to the review comments. I had to rebase in order to know that the changes fit into the current language specification, I hope that doesn't create too much noise. Could you take another look?

eernstg:
Hello again, @chloestefantsova, could you take another look at this one?

chloestefantsova (approved):
lgtm :)

munificent:
It landed! 🎉 🎉 🎉 🎉 🎉 🎉

Issue #3516, opened 2023-12-15
Can an external constructor have an initializing formal?


eernstg:
Thanks to @sgrekhov for bringing this up. Consider the following library:

```dart
class C {
  num x;
  external C(this.x);
}
```

This declaration is rejected by the common front end, based on the claim that external constructors cannot declare initializing formals. It is accepted by the analyzer. The specification does not provide any information about this specific situation.

In other words, it is not a breaking change to say that it is a compile-time error for an external constructor to have an initializing formal parameter. Also, it is not at all clear (to me) what the dynamic semantics would be.

In any case, there might be a useful and meaningful way to define what it would do, in which case we might want to specify that it is allowed.

@dart-lang/language-team, WDYT? I tend to think we should just specify that it is an error.

lrhn:
It should be an error.
It should also not be allowed to have super parameters

An external declaration is like an abstract declaration in that it doesn't specify implementation, only signature.

Unlike abstract declarations, it's not restricted to instance members, so it's not surprising that constructors is a case that has to be addressed directly, that's the only kind of declaration which doesn't exist as an instance member.

But its parameters should probably be just as restricted as a forwarding factory constructor.

eernstg:
Not even default values?

lrhn:
Not even default values.

Unless we allow those for external methods, then we should be consistent. Which we do. 
(Or we should fix that too.)

So I guess we can allow default values as a legacy feature, but that default value has no *effect*. It's purely documentation.
It doesn't constrain the actual implementation to use the same default value, or any default value (an external natively implemented function can likely be made to detect whether an argument was passed or not, and not need a default value).

It would be more honest to not allow it.

Example, can run in DartPad:
```dart
import "package:js/js.dart";
@JS("console.log")
external int foo([int x = 2]);
void main() {
  foo(42); // Logs 42
  foo(); // Logs nothing, because `undefined` is ignore.
}
```

eernstg:
Closing:

The specification was updated in https://github.com/dart-lang/language/pull/3517. It is now an error for an external constructor to declare a formal parameter which is initializing (that is, a formal parameter declaration derived from `<fieldFormalParameter>`).

An implementation has been requested in https://github.com/dart-lang/sdk/issues/54485.

PR #3517, opened 2023-12-15
Specify precisely where it is an error to declare an initializing formal

eernstg:
The compile-time error which is reported by the common front end for a declaration of a formal parameter as an initializing formal (`this.name`) was not specified precisely until now (it was just mentioned in commentary and never spelled out in normative text).

This PR changes the commentary to normative text.

One question may still give rise to debate: Can external constructors have initializing formals? This would be an enhancement, and it is handled in https://github.com/dart-lang/language/issues/3516. If that enhancement is accepted then we will change the specification later on in a different PR&mdash;this PR just specifies the current situation and makes it an error.

There will be an implementation effort in any case: If this PR is landed and the error reported by the CFE follows the updated specification then the analyzer would need to be changed to report that error. If an enhancement is accepted then all tools would need to implement it.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -41,6 +41,7 @@
 % version of the language which will actually be specified by the next stable
 % release of this document.
 %
+<<<<<<< HEAD
 % Dec 2023
 % - Allow `~/` on operands of type `double` in constant expressions, aligning
 %   the specification with already implemented behavior.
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -4003,12 +4005,12 @@ \subsubsection{Generative Constructors}
...
-Note that it is a compile-time error for an initializing formal
-to occur in any function which is not a non-redirecting generative constructor
-(\ref{requiredFormals}),
-so there is always an enclosing class or enum.%
-}
+\LMHash{}%
+It is a compile-time error for an initializing formal parameter
+to occur in any function which is not a generative constructor.
+Also, it is a compile-time error for an initializing formal parameter
+to occur in a redirecting and in an external constructor.
```
natebosch, line 4012: [optional] 

```suggestion
to occur in a redirecting or external constructor.
```

natebosch (approved):

Issue #3491, opened 2023-12-01
Require that the type in an `on` clause of a mixin must be regular bounded


eernstg:
```dart
class A<X extends num> {}
mixin M on A<Object?> /*a super-bounded type*/ {}
```

It is currently not an error for the type in a mixin `on` clause to be super bounded (as opposed to types in `extends`, `with`, or `implements` clauses: they must all be regular bounded).

However, we should probably just add the mixin `on` clause in this list of places where the type must be regular bounded. Note that it is never useful anyway. Let's try to use the above mixin `M`:

```dart
class B extends A<S> with M {}
```

`S` would have to be `Object?` in order to avoid having the error which is caused by having multiple different instantiations of the same generic class as superinterfaces ("you can't implement `List<int>` and `List<num>` at the same time!"). At the same time, `S` cannot be `Object?`, because it is an error to use a super bounded type in an `extends` clause.

I believe it is impossible to use a mixin whose `on` clause has a super bounded type. With this in mind, I'd recommend that we make it a compile-time error.

lrhn:
Agree. The mixin introduces an interface, that interface has the `on` type as superinterface. If you're not allowed to have a superbounded type on an `implements` clause, you shouldn't be allowed to have it on an `on` clause.

(I would have expected it to be an error with the current specification, because of the definition of combined superinterface, but maybe we dodge that by having only one `on` type.)

eernstg:
As of Dec 29 2023, this has been implemented.

eernstg:
The specification has this rule as well: https://github.com/dart-lang/language/blob/0ecd65a8988520dd5bec097056d1657f7b4b6d54/specification/dartLangSpec.tex#L21371.

PR #3535, opened 2023-12-29
Add rule to make it an error for a mixin `on` type to be super-bounded

eernstg:
See https://github.com/dart-lang/language/issues/3491 where this change was discussed.

This PR makes it a compile-time error for a mixin `on` type to be super-bounded. This rule is similar to several other cases (it is an error for a type in an `extends`, `implements`, or `with` clause to be super-bounded as well). For mixins it is justified by the fact that it is impossible to use a mixin with a super-bounded `on` type.

There is no implementation effort because this is already the implemented behavior.


eernstg:
Hello @lrhn, friendly ping. ;-)

lrhn (approved):

eernstg:
Thanks!

Issue #2710, opened 2022-12-10
Extension types and `await`.


lrhn:
(NOTICE: Written when the feature was called "inline class", is now "extension type", and is not completely the same. New conclusion too, probably, so check comments.)

TL;DR: _flatten_(_N_) of an inline class type _N_ should be _N_.

What happens if you do `await e` and the static type of `e` is an inline class, _N_?

The runtime semantics are pretty clear, so it boils down to defining what _flatten_ returns on an inline class.
(And it's probably going to be trivial, but let's make sure it's documented.)

As currently defined, an inline class can only implement other inline classes, so it cannot implement `Future` (or `FutureOr`).
It would seem valid to make _flatten_(_N_) = _N_ for an inline-class type _N_, like we do for other types which do not imply an expected `Future` type.

At runtime, the inline class type does not exist, it's replaced entirely by the representation type. 
(Basically, at runtime the inline class type can be treated as a type alias for the representation type, instead of a separate type. It needs to have all the same cyclic-reference restrictions as a type alias to avoid an infinite expansion when going from the static type to the runtime type.)

If _N_ has representation type _R_, the static type of `await e` will be _flatten_(_N_), aka _N_, which is the same as _R_ at runtime.
The `await` needs to do an `is Future<R>` check on the value of `e` to see if it should be, and can soundly be, awaited.
If the representation type is actually `Future<int>` itself, we won't await that future, because it's not a `Future<Future<int>>`.

That *should* work, and is probably what we want.

An alternative could be to looks at the representation type immediately, and see if it mentions a future type, and if so assume that we are awaiting that type. 
I don't think it's a good idea. Mainly because it breaks the abstraction of the inline class, but also because it opens up questions about where _R_ being a Future\<_R_\> should make the `await e` switch back to being typed as _N_ or not.
So, let's not do that, the _flatten_(_N_) = _N_ is simpler.

If, at a later time, we allow an inline class type to implement interfaces or other types (which must then also be supertypess of the representation type), then it may become possible for an inline class type to implement `Future` or `FutureOr`. At that point we should look at the future-behavior of the type itself, just as we do for other types. That should be sound wrt. the representation type, since it requires the representation type itself to have those types.

eernstg:
Good question! :smile:

[...]

PR #3560, opened 2024-01-15
Adjust the rule about await e where the type of e is an extension type

eernstg:
See the discussion in https://github.com/dart-lang/language/issues/2710.

This PR introduces the notion of a type which is _incompatible with await_ and uses that to specify the compile-time error which occurs in the case where `await e` is encountered, and the static type of `e` is incompatible with await. The main case is when said static type is an extension type that does not implement `Future` (that's the case that we had already), but this new rule includes more cases, e.g., the case where the static type is `E?` where `E` is an extension type that does not implement `Future`.

### DIFF
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -873,8 +873,10 @@ one of the following criteria holds:
 
 - `T` is an extension type, and `T` does not implement `Future`.
 - `T` is `S?`, and `S` is incompatible with await.
-- `T` is `X & B`, `B` does not derive a future type, and `X` is
-  incompatible with await.
+- `T` is `X & B`, and either:
+  - `B` is incompatible with await, or
+  - `B` does not derive a future type, and `X` is
+    incompatible with await.
 - `T` is `S` bounded, and `S` is incompatible with await.
 
 Consider an expression of the form `await e`. A compile-time error 
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -864,9 +864,16 @@ _is the extension type_
 <code>V\<T<sub>1</sub>, .. T<sub>s</sub>&gt;</code>,
 and that its static type _is an extension type_.
 
-It is a compile-time error if `await e` occurs, and the static type of
-`e` is an extension type which is not a subtype of `Future<T>` for any
-`T`.
+We say that a type `T` is _incompatible with await_ if at least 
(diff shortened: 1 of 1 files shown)


eernstg:
NB: The rule about being incompatible with await is intended to be evolvable over time, covering the case where we want to make more types incompatible with await (that is, if we wish to make all or part of `await_only_futures` a compile-time error). We'd simply add more atomic cases to this rule, similar to the first case (about implementing an extension type that doesn't implement `Future`).

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -864,9 +868,19 @@ _is the extension type_
...
+We say that a type `T` is _incompatible with await_ if at least 
+one of the following criteria holds:
+
+- `T` is an extension type, and `T` does not implement `Future`.
+- `T` is `S?`, and `S` is incompatible with await.
+- `T` is `X & B`, and either:
+  - `B` is incompatible with await, or
+  - `B` does not derive a future type, and `X` is
+    incompatible with await.
+- `T` is `S` bounded, and `S` is incompatible with await.
```
lrhn, line 880: I still prefer to keep every rule one-step, so this one would be
> - `T` is a type variable `X` with bound `S`, and `S` is incompatible with await.

By using "`S` bounded" we include both transitive bounds *and* promoted type variables, which means that "`X & B` where `B` is incompatible with await" is now double-handled, since it's *also* handled by "`S` bounded and `S` is incompatible with await".

It's probably not wrong, but it introduces an extra complexity that makes it harder to reason about what the cases actually do. It's not just the "T is a type variable with bound `S`" or even just the transitive of that, because it also includes promoted type variables.
And when it's the only rule which does that, it stands out and makes the rest of the rules harder to think about too, when you can't use the same approach to every rule.

If we ever change "`S`-bounded" to include something more, we risk this rule being affected, even if we don't 
actually *want* it. It probably won't, but I'd rather be explict.

lrhn (approved):
Still prefer the phrasing without "`T` is `S` bounded".

accepted/future-releases/extension-types/feature-specification.md
eernstg: OK, changed the rule to use your wording.

I was a little bit worried about reducing the coverage of the ruleset as a whole (such that we don't catch situations that we would catch with "bounded"), but I don't see any such cases with the concrete set of examples that we've discussed, and a check on the definition of "bounded" seems to support the claim that nothing is lost.

This is great, the rules are actually rather easy to read now! ;-)

PR #3561, opened 2024-01-15
Add missing word 'bounded' to definition of flatten

eernstg:
This PR adds the word 'bounded' to the `S?` case in the definition of flatten. This change is needed for consistency. Here is an example:

```dart

extension<X> on X {
}

Future<void> test<X extends Future<int>?>(X x) async {
  // ignore: await_only_futures
  var y = await x;
  y.expectStaticType<Exactly<int?>>; // This is what we want.
}

void main() async {
  await test<Future<int>?>(null);
  await test<Future<int>>(Future.value(1));
}
```

If we do not have the word 'bounded' in said rule then `await x` will have the type `X` and the future will not be awaited. The type will be `int?` as shown above if the rule is changed and 'bounded' is included.

There is no breaking change, and indeed no implementation effort associated with this specification change: the implementations behave as if the spec change had already been performed.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -41,6 +41,12 @@
 % version of the language which will actually be specified by the next stable
 % release of this document.
 %
+% Jan 2024
+% - Add the word 'bounded' to a rule in the definition of 'flatten'. This is
+%   already the implemented behavior, and it is the consistent approach. The
+%   main consequence is that `await e` will await the future when the static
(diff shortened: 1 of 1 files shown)


eernstg:
Hi, Nate!

It is indeed true that some cases in the current implementations behave as if we had that word 'bounded' in the specification already&mdash;but it isn't true in all cases. So the situation isn't as simple as it looked at first.

So I'll retract this PR and then we'll have to discuss the topic in more detail in the language team.

Sorry about the bothering you for no good reason!

Issue #3524, opened 2023-12-18
Do extension types follow their representation type in being always-exhaustive?


eernstg:
[Edit: The language team [decided](https://github.com/dart-lang/language/issues/3524#issuecomment-1875649080) that the answer is yes.]

The pattern feature specification defines the notion of an [always-exhaustive](https://github.com/dart-lang/language/blob/main/accepted/3.0/patterns/feature-specification.md#exhaustiveness-and-reachability) type and specifies that a switch statement must be checked for exhaustiveness when the static type of the scrutinee is always-exhaustive.

Exhaustiveness properties of patterns involving extension types is achieved by taking the extension type erasure of all the relevant types, as specified [here](https://github.com/dart-lang/language/blob/main/accepted/future-releases/extension-types/feature-specification.md#:~:text=V%3E.empty()\).-,An%20extension%20type%20V,-(which%20may%20include).

The intention behind that paragraph in the extension type feature specification was that extension type erasure should apply to all aspects of exhaustiveness, but it has not been spelled out that the matched value type should be extension type erased as well.

If we do that then it follows that a switch statement must be checked for exhaustiveness when the static type of the scrutinee is an extension type whose erasure is always-exhaustive.

The conceptual justification would be that extension types should only be used in association with pattern matching in the case where it is OK to adopt the extension type (that is, casting from the representation type or a subtype to the extension type). We can have other devices (say, lints) to check that this is OK in the first place. Assuming that it is OK, it seems reasonable to me to say that the property of being always-exhaustive should also be allowed to propagate from the representation type to the extension type.

@dart-lang/language-team, WDYT?

lrhn:
I'll say "no" as default, "yes" if the extension type implements a sealed class type.

And that's still work caveats.

If the extension type implements an `enum` type (or `bool`), it'l be tested as that.

If it implements a sealed type, then it matters how.
If it writes `implements SealedType` and the representation type is `SealedType`, then yes.

If the representation type is not `SealedType`, then no. Having a subclass of a sealed class as representation type, and implementing the sealed type, is a way to create a wrapped _alternative_ to a sealed type direct subclass. It doesn't represent the entire sealed type hierarchy.

If it's not _all_ of the sealed type, because of generics, then _probably_ still yes.
Say there is an `Option<T>`/`Value<T>`/`None` sealed class hierarchy. If an extension type only applies to `Option<int>`... then that's still a sealed sub-hierarchy. Maybe.

[...]

PR #3568, opened 2024-01-17
Add a rule to determine which extension types are always-exhaustive

eernstg:
See https://github.com/dart-lang/language/issues/3524#issuecomment-1875649080 for background information.

This PR adds a rule to the extension type feature specification, specifying that an extension type is always-exhaustive if and only if its instantiated representation type is always-exhaustive (note that this rule may be applied repeatedly because the representation type could itself be an extension type).

Example:

```dart
extension type SmartBool(bool b) {}

String foo(SmartBool sb) {
  switch (sb) { // This switch must be exhaustive.
    case SmartBool(b: true): return 'True';
    case SmartBool(b: false): return 'False';
  }
}
```

### DIFF
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -18,6 +18,7 @@ information about the process, including in their change logs.
 2024.01.17
   - Specify that a type is 'incompatible with await', and use that to specify
     a compile-time error at `await e;`.
+  - Specify which extension types are always-exhaustive.
 
 2023.11.14
   - Specify that a method declaration will shadow an otherwise "inherited"
@@ -827,6 +828,9 @@ in an object pattern *(e.g., `case V(): ...` where `V` is an extension
 type)*.  Exhaustiveness analysis will treat such patterns as if they had
 been an object pattern matching the extension type erasure of `V` (defined
 below).
+
+An extension type `V` is always-exhaustive if and only if its instantiated 
+representation type is always-exhaustive.
 
 *In other words, we make no attempt to hide the representation type during
 the exhaustiveness analysis. The usage of such patterns is very similar to
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -829,8 +829,16 @@ type)*.  Exhaustiveness analysis will treat such patterns as if they had
 been an object pattern matching the extension type erasure of `V` (defined
 below).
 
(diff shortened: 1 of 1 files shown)


eernstg:
@sgrekhov, I think we already acted on this change when the underlying language design question was decided? The actual text in the feature specification should not be controversial, and it should not change anything, but it would be great if you can double check the situation when this lands.

sgrekhov:
I'll check it

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -828,6 +829,9 @@ type)*.  Exhaustiveness analysis will treat such patterns as if they had
 been an object pattern matching the extension type erasure of `V` (defined
 below).
 
+An extension type `V` is always-exhaustive if and only if its instantiated 
+representation type is always-exhaustive.
```
lrhn, line 833: Could we add examples?

> _For example a type `TriBool` defined as `extension type TriBool(bool? _) {}`, is always-exhaustive because the representation type, `bool?`, is.
> For a declaration like `extension type Pair<S, T>((S, T) _) {}`, the type `Pair<SomeEnum, TriBool>` is always-exhaustive because `(SomeEnum, TriBool)` is, and `Pair<String, bool>` is not always-exhaustive because `(String, bool)` isn't._

accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -828,6 +829,9 @@ type)*.  Exhaustiveness analysis will treat such patterns as if they had
...
+
```
lrhn, line 834: Should we say somewhere, maybe around line 916, that `extension type E(SomeType _) implements SealedType {}` does not count as one of the "immediate subtypes" of the sealed type. Either the representation type *is* the sealed type, in which case the extension type inherits the always-exhaust, the representation type is an immediate subtype of the sealed type, in which case it counts towards exhausting that immediate subtype, or it's a further subtype of one or more immediate subtypes, in which case it probably doesn't count towards exhausting the sealed type (unless the intermediate types are also sealed).

lrhn (approved):

accepted/future-releases/extension-types/feature-specification.md
eernstg: Nice! Done.

accepted/future-releases/extension-types/feature-specification.md
eernstg: Done!

PR #3570, opened 2024-01-17
Update spec to say that reporting warnings is not mandatory.

munificent:
I'm working on a larger doc to explain this but here's a proposed small change to the spec to say that warnings don't have to be reported.

@dart-lang/language-team

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -1267,8 +1267,10 @@ It is then the runtime's decision whether to terminate the isolate or not.%
 are situations that do not preclude execution,
 but which are unlikely to be intended,
 and likely to cause bugs or inconveniences.
-A static warning must be reported by a Dart compiler
-before the associated code is executed.
+A Dart compiler is free to report some, all,
+or none of the specified static warnings before executing the associated code.
+A Dart compiler may also report additional warnings
+not defined by this specification.
 
 \LMHash{}%
 When this specification says that a \Index{dynamic error} occurs,


munificent:
I didn't mean to commit this. Clicked "Squash and Merge" on the wrong tab. 🙃

specification/dartLangSpec.tex
```diff
@@ -1267,8 +1267,10 @@ \section{Errors and Warnings}
 are situations that do not preclude execution,
 but which are unlikely to be intended,
 and likely to cause bugs or inconveniences.
-A static warning must be reported by a Dart compiler
-before the associated code is executed.
+A Dart compiler is free to report some, all,
+or none of the specified static warnings before executing the associated code.
```
lrhn, line 1271: Nit: Compilers don't execute code, so maybe "before the associated code is executed."

(Which would even allow the compiler to insert code that prints a warnings when the compiled program is run.)

We can then wonder why it's not allowed to report the warning *after* the code is executed, but now I think I'm just being pointlessly pedantic. :grin:

specification/dartLangSpec.tex
```diff
@@ -1267,8 +1267,10 @@ \section{Errors and Warnings}
...
+A Dart compiler may also report additional warnings
```
lrhn, line 1272: I'd consider making it "... may also choose to report ...". Just to emphasize that it's a *choice*, and it's not because there is some mandated list of warnings outside of this specification.

specification/dartLangSpec.tex
munificent: Good point. Done: https://github.com/dart-lang/language/pull/3584

specification/dartLangSpec.tex
munificent: +1: https://github.com/dart-lang/language/pull/3584

munificent:
I didn't mean to commit this yesterday, but we did agree to the change in the language meeting today, so this commit is now officially intended to be landed. :)

PR #3574, opened 2024-01-18
Enhance the definition of "derives a future type"

eernstg:
See https://github.com/dart-lang/sdk/issues/54628 for some background information.

This PR changes the definition of 'derives a future type' (which is used to define _flatten_) such that it includes some additional cases (in particular, it will recursively handle `?` on types, which was not the case previously).

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -11721,13 +11721,36 @@ We say that $S$ is the
 $T$ in the following cases, using the first applicable case:
 
 \begin{itemize}
-\item $T$ implements $S$
-  (\ref{interfaceSuperinterfaces}),
-  and there is a $U$ such that $S$ is \code{Future<$U$>}.
-\item $T$ is $S$ bounded
+\item
+  If $T$ implements \code{Future<$U$>}
+  (\ref{interfaceSuperinterfaces})
+  for some $U$ then
+  $T$ derives the future type \code{Future<$U$>}.
+\item
+  Whenever $S$ is of the form
+  \code{FutureOr<$U$>}, \code{Future<$U$>?}, or \code{FutureOr<$U$>?},
+  the following holds:
+  If $T$ is $S$ bounded
   (\ref{bindingActualsToFormals}),
-  and there is a $U$ such that
-  $S$ is \code{FutureOr<$U$>}, \code{Future<$U$>?}, or \code{FutureOr<$U$>?}.
+  then $T$ derives the future type $S$.
+\item
+  If $T$ is \code{$T_1$?} for some $T_1$ and
+  $T_1$ derives the future type $S_1$
+  then $T$ derives the future type \code{$S_1$?}.
+\item
+  If $T$ is a type variable with bound $B$ and
+  $B$ derives the future type $S$
+  then $T$ derives the future type $S$.
+\item
+  If $T$ is of the form \code{$X$\,\,\&\,\,$B$} then the following holds:
+  \begin{itemize}
+  \item
+    if $B$ derives the future type $S$ then $T$ derives $S$.
+  \item
+    if $B$ does not derive a future type,
+    but $X$ derives the future type $S$
+    then $T$ derives $S$.
+  \end{itemize}
 \end{itemize}
 
 \LMHash{}%
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -11765,7 +11765,7 @@ and $F$ is always of the form \code{$G$<...>} or \code{$G$<...>?},
 where $G$ is \code{Future} or \code{FutureOr}.
 Also note that 'derives' in this context refers to the computation
 where a type $T$ is given, the supertypes of $T$ are searched,
-and a type of one of those forms is selected.
+and a type $F$ of one of those forms is selected.
 There is no connection to the notion of a 'derived class' meaning 'subclass'
 that some programming language communities use.%
 }
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -11735,11 +11735,11 @@ $T$ in the following cases, using the first applicable case:
   then $T$ derives the future type $S$.
 \item
   If $T$ is \code{$T_1$?} for some $T_1$, and
-  $T_1$ derives the future type $S_1$
+  $T_1$ derives the future type $S_1$,
   then $T$ derives the future type \code{$S_1$?}.
 \item
   If $T$ is a type variable with bound $B$, and
-  $B$ derives the future type $S$
+  $B$ derives the future type $S$,
   then $T$ derives the future type $S$.
 \item
   If $T$ is of the form \code{$X$\,\,\&\,\,$B$} then the following holds:
@@ -11748,7 +11748,7 @@ $T$ in the following cases, using the first applicable case:
     if $B$ derives the future type $S$ then $T$ derives $S$.
   \item
     if $B$ does not derive a future type,
-    but $X$ derives the future type $S$
+    but $X$ derives the future type $S$,
     then $T$ derives $S$.
   \end{itemize}
 \end{itemize}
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -11716,18 +11716,43 @@ \subsection{Function Expressions}
...
+  (\ref{interfaceSuperinterfaces})
+  of $T$ is \code{Future<$U$>} for some $U$,
+  then $T$ derives the future type \code{Future<$U$>}.
+\item
+  If $T$ is the type \code{FutureOr<$U$>} for some $U$,
+  then $T$ derives the future type \code{FutureOr<$U$>}.
+\item
+  If $T$ is \code{$S$?} for some $S$, and
+  $S$ derives the future type $F$,
+  then $T$ derives the future type \code{$F$?}.
```
lrhn, line 11740: (Would we want to avoid having multiple `?`s, or won't it matter?)

specification/dartLangSpec.tex
```diff
@@ -11716,18 +11716,43 @@ \subsection{Function Expressions}
...
+\item
+  If $T$ is a type variable with bound $B$, and
+  $B$ derives the future type $F$,
+  then $T$ derives the future type $F$.
+\item
+  If $T$ is of the form \code{$X$\,\&\,$S$} then:
```
lrhn, line 11746: We currently only use `"derives a future type" from *flatten* on the promotion type of an `X&B`, which means that `B` cannot itself be a promoted type variable. So this case should be unreachable.

specification/dartLangSpec.tex
```diff
@@ -11716,18 +11716,43 @@ \subsection{Function Expressions}
...
+  \begin{itemize}
+  \item
+    if $S$ derives the future type $F$,
+    then $T$ derives the future type $F$.
+  \item
+    if $B$ does not derive a future type,
+    but $X$ derives the future type $F$,
+    then $T$ derives the future type $F$.
```
lrhn, line 11754: (Not sure about the last case. The way I can see it's reachable are constructions like:
* `(X extends FutureOr<int>?) & int` - the non-future branch of a `FutureOr`
* `(X extends FutureOr<int>?) & Null` - the non-future-deriving branch of a `?`
* `(X extends FutureOr<int>?) & Never` - Never, always fun, never useful.

In all of these cases, I'd actually prefer to say that they type does *not* derive a future type, since we've chosen, by type-checking to get promotion, the *non-future* branch of the type. We've *checked* that it's not a future! So let's grant the user that.)

specification/dartLangSpec.tex
```diff
@@ -11737,7 +11762,12 @@ \subsection{Function Expressions}
 \commentary{%
 Note that if $T$ derives a future type $F$ then \SubtypeNE{T}{F},
 and $F$ is always of the form \code{$G$<...>} or \code{$G$<...>?},
```
lrhn, line 11764: Or more `?`s, if we don't avoid it.

lrhn (approved):

specification/dartLangSpec.tex
```diff
@@ -11716,18 +11716,43 @@ \subsection{Function Expressions}
...
+  $B$ derives the future type $F$,
+  then $T$ derives the future type $F$.
+\item
+  If $T$ is of the form \code{$X$\,\&\,$S$} then:
+  \begin{itemize}
+  \item
+    if $S$ derives the future type $F$,
+    then $T$ derives the future type $F$.
+  \item
+    if $B$ does not derive a future type,
```
chloestefantsova, line 11752: What is `B` in this context?

lrhn: Think it should be `$S$`. Used to use the `X extends B` for type variable with bound and `X&B` for promoted type variable, because at the start they were treated as equivalent. (Which they aren't here, because it matters that the bound is a supertype of the variable and the promotion is a subtype.)

specification/dartLangSpec.tex
lrhn: The more I think about that rule, the less I like it. (Seemed like a great idea at the time, etc.)

If we have a promoted type variable which doesn't derive a future type, and the type variable itself does (so its bound does), then someone has actively type-checked their way _out of_ deriving a future type.

That means either promoting to `Null` or `Never` (unlikely) or to the non-future branch of a `FutureOr`, in which case awaiting that should probably not go back and use the `FutureOr` type anyway.

Say, having an `X extends FutureOr<num>` and doing `x is int` to promote to `int` and then awaiting that will give something of type `num`. 
I think it would be fine to use `int` instead, and not have a fallback when flattening a promoted type variable, where the promotion doesn't derive a future type.
(Yes, you shouldn't be awaiting an `int`, but it is currently a valid way to insert a delay. And it might still happen in an async return.)

eernstg:
Finally got around to this one again. @lrhn, you may wish to take another look because I made some significant changes.

specification/dartLangSpec.tex
```diff
@@ -11716,18 +11720,38 @@ \subsection{Function Expressions}
...
+  then $T$ derives the future type \code{$F$?}.
+\item
+  If $T$ is a type variable with bound $B$, and
+  $B$ derives the future type $F$,
+  then $T$ derives the future type $F$.
+\item
+  \commentary{%
+    There is no rule for the case where $T$ is of the form \code{$X$\,\&\,$S$}
+    because this will never occur
+    (this concept is only used in \flattenName, which is defined below).%
```
chloestefantsova, line 11753: If I understand it correctly, the concept is also used in the definition of types incompatible with `await`, where the right-hand side of an intersection can be tested for deriving a future. If it does, the entire intersection type is deemed not incompatible with `await`. Consider the following example.

```dart
foo<X extends Future<int>?, Y extends Future<int>?>(X x) async {
  if (x is Y) {
    await (x ?? x!);
  }
}
```

The static type of the awaited expression `x ?? x!` is `X & (Y & Future<int>)`. To test if it's incompatible with `await`, we look at the right-hand side of the intersection, which is `Y & Future<int>` and see if it derives a future type. Intuitively it seems that it does. So it looks like the rule for intersection type should be defined.

chloestefantsova: UPDATE: The example works with `await x!` too. The static type of `x!` is the same nested intersection type `X & (Y & Future<int>)`. I used `(x ?? x!)` in the original example because it provides a visible representation of the static type of the awaited expression as shown in the following snippet.

```
library;
import self as self;
import "dart:async" as asy;
import "dart:core" as core;

static method foo<X extends asy::Future<core::int>?, Y extends asy::Future<core::int>?>(self::foo::X% x) → dynamic async /* emittedValueType= dynamic */ {
  if(x is self::foo::Y%) {
    await let final self::foo::X% & self::foo::Y% /* '%' & '%' = '%' */ #t1 = x{self::foo::X% & self::foo::Y% /* '%' & '%' = '%' */} in #t1 == null ?{self::foo::X% & self::foo::Y% & asy::Future<core::int> /* '%' & '!' = '!' */ /* '%' & '!' = '!' */} x{self::foo::X% & self::foo::Y% /* '%' & '%' = '%' */}! : #t1{self::foo::X% & self::foo::Y% & asy::Future<core::int> /* '%' & '!' = '!' */ /* '%' & '!' = '!' */};
  }
}
```

eernstg:
NB: LGTM'd or not, this PR will not land before we've had the next language team meeting. We may need to treat it as a breaking change.

specification/dartLangSpec.tex
```diff
@@ -11716,18 +11720,38 @@ \subsection{Function Expressions}
...
 \begin{itemize}
-\item $T$ implements $S$
-  (\ref{interfaceSuperinterfaces}),
-  and there is a $U$ such that $S$ is \code{Future<$U$>}.
-\item $T$ is $S$ bounded
-  (\ref{bindingActualsToFormals}),
-  and there is a $U$ such that
-  $S$ is \code{FutureOr<$U$>}, \code{Future<$U$>?}, or \code{FutureOr<$U$>?}.
+\item
+  %% TODO(eernst): Come mixin classes and extension types: add them.
```
lrhn, line 11731: They have come. (But maybe not to the spec.)

lrhn (approved):

PR #3578, opened 2024-01-22
[macros] Add standalone program for the example observable macro

jakemac53:
This is runnable today, with `dart --enable-experiment=macros bin/observable_main.dart`


davidmorgan (approved):
Thanks :)

PR #3579, opened 2024-01-23
Revert "Update spec to say that reporting warnings is not mandatory."

munificent:
Reverts dart-lang/language#3570

Clicked "Submit" on the wrong tab!


munificent:
Today, we decided to go ahead with this change, so my preemptive commit can stay.

Issue #3577, opened 2024-01-22
Consider different name for "augmentation libraries"


lrhn:
The augmentation libraries feature specification introduces "augmentation libraries" which are not Dart libraries.
An augmentation library has things in common with libraries (can have its own imports), and things in common with parts (has the same declaration scope as the parent library's), but it's really a neither.

I worry that using the word "library" sets the wrong expectations. Where we currently say "library", we'll have to say "non-augmentation library", unless we want to include it, in which case I'd say "augmentation or non-augmentation library".
The word "library" itself will be lost, because it's ambiguous whether it refers to both the new (different kinds) of libraries, or just the old-style actual library.

If anything, staying in the terminology of augmentation, the file is more of a "library augmentation", which augments the main library, than it is an "augmentation library", with focus on the "library".

If we use that, then:
* A library augmentation is a file that looks similar to a library in many ways, just like a class augmentation looks like a class.
* It adds or modifies the existing library by having declarations or declaration augmentations, just like a class declaration modifes a class.

So, conceptually I think it feels OK.

_The biggest difference is that a library declares its own augmentations, where other declarations do not. A class augmentation just appears (later) in the same library as the class declaration. But since libraries do not exist in a container that can also hold augmentations, it's probably unavoidable that the link to the augmentation is in the library itself.

So, suggestion:
* It's called a "library augmentation", not an "augmentation library".
* It's imported using `augment import "...";` or `augment part "...";`, just like any other augmentation declaration starts with `augment`. (And `augment` will be a built-in identifier, so that should not be a problem.)
* It's marked at the top as `augment of "parent.dart";`, `augment library "parent.dart"`, or `augmented library "parent.dart"`. Not using a `library` declaration with `library` first, because it is not a library. Either treat it as declaring that it augments the parent library, or that it references the parent library using `augmented`, but the latter would be the only use of `augmented` outside of body code, so probably not worth it.

The library augmentation file doesn't have a library name. It cannot be imported directly. It's not a library.

Then we can keep saying "library" to refer to just the plain Dart library,
and "library augmentation" to refer to the library augmentations that a library or library augmentation may include.

jakemac53:
All of this sounds good to me 👍

rrousselGit:
> It's called a "library augmentation", not an "augmentation library".

Why not just "augments"? If we can augment a part, saying "library augmentation" feels wrong to me

bwilkerson:
@MaryaBelanger For the documentation impact.

[...]

Issue #3576, opened 2024-01-22
Augmentations: Allow parts in augmentations, augmentations in main library


lrhn:
The current augmentation feature specification introduces *augmentation libraries* which a main library or augmentation library can include. An augmentation library:
* Can contain augmentation declarations, and
* Cannot contain part files.

An augmentation library can contain multiple augmentations for the same declaration, and it can contain both a declaration and an augmentation for that declaration.

The main library file can contain part files, but cannot contain augmentations.

I suggest removing the restrictions from this:
* Allow augmentation libraries to have part files.
* Allow a non-augmentation library to contain augmentations.

The restrictions do not appear to be *necessary*, and I don't think we should be enforcing a purely stylistic choice on how people choose to organize their code.
It's an unnecessary breach of orthogonality.
(The one argument I can see is that allowing augmentations in part files will make the ordering of part files significant, where today it isn't. But today, no part file contains augmentations, so it won't be breaking. And if augmentation library order is significant, I have no problem with part file order being signficant if they include augmentations.)

The change would require only one new thing: Integrating part files in the augmentation application order.

My proposal is:
* Every (augmentation or non-augmentation) declaration in a library or augmentation library, including in its part files or transitive augmentation libraries, is ordered in _augmentation application order_.
* The ordering of the declarations for a library or augmentation library is:
   * First the declarations in the main library file, in souce order.
   * Then for each `part` declaration of the library, in source order:
       * The declarations of that part file in source order.
   * Then for each augmentation import of the library, in source order:
       * All the declarations of that augmentation library in augmentation application order.

That is the same as the current specification if a main library contains no augmentations and an augmentation library contains no parts. Adding either will insert the augmentation library parts or main library augmentations in a consistent place.

The rule would still be that an augmentation of a declaration must occur after the declaration in augmentation application order, and that augmentations of the same declaration are applied in augmentation applicatation order, each's `augmented` references referring to the previous, possibly augmented, declaration.
(We could loosen that and allow non-augmentation declarations to be collected first, from the entire libray, and then augmentations are applied in order, so an augmentation can occur before the declaration it augments. It's probably not necessary, and it's easier to have only one order.)

Even if we won't allow augmentation libraries to have parts, or even part files to have augmentations, we should still allow main library files to contain agumentations. There is no reason not to, when the main library file can be just `augment import "real_library.dart";`. Basically, the main library should be able to *be* an augmentation library. 
(If nothing else, that. Even if we choose to not allow it to also have parts.)

[...]

PR #3583, opened 2024-01-23
Change "augmentation library" wording to just "augmentation", other updates

jakemac53:
A variety of proposed changes related to https://github.com/dart-lang/language/issues/3577.

- Changes references to 'augmentation library' to 'library augmentation', it isn't really a library on its own
- Changes 'library augment' => 'augment library' (reads better)


working/augmentation-libraries/feature-specification.md
```diff
@@ -142,59 +141,58 @@ It is a compile-time error if:
...
 library.
 
 ### Merge order
 
 A library may apply multiple augmentations to itself. Also, augmentation files
-may themselves contain `import augment` directives. The entire tree of
-augmentations is recursively applied to the main library. The merge order is
-defined as a depth-first pre-order traversal of the `import augment` directives
-starting at the main library.
+may themselves contain `augment` directives. The entire tree of augmentations is
```
lrhn, line 181: "`augment` directives" have something of the same issue as plain "augmentation", because I'm pretty certainly going to use *something like that* that to cover `augment class` declarations too, which an augmentation file can also contain.
If the distinction is that the "augment import" is a *directive* and the other ones are *declarations*, I don't think that's going to be distinct enough. 
At least not unless we make that distinction *very explicit* where we introduce the directive above.

Say, something like:
> A directive of the form `augment "file.dart";` which applies a library augmentation to the library of the directive, is called an _`augment` directive_, which distinguishes it from _`augment` declarations_, which are declarations that an augmentation file can also contain, and whose syntax also starts with `augment`. These `augment` declarations will be introduced later.

If we do that, very up front, I think we *can* use "`augment` directive" unambiguously.
(It's really about defining the term. If we do nothing, "`augment` directive" is just a descriptive term, not a declared concept, and people can interpret it, and interpret it incorrectly. If we *define* the meaning, then the term has only that one meaning in this document. I'm a big fan of giving things defined names.)

working/augmentation-libraries/feature-specification.md
```diff
@@ -142,59 +141,58 @@ It is a compile-time error if:
...
+recursively applied to the main library. The merge order is defined as a
+depth-first pre-order traversal of the `augment` directives starting at the main
```
lrhn, line 183: of the augmentation files, in the source order of their `augment` directives,

working/augmentation-libraries/feature-specification.md
```diff
@@ -142,59 +141,58 @@ It is a compile-time error if:
...
+library
 
-Within a single augmentation library, you may augment the same declaration
-multiple times, whether they are top level or nested declarations. The merge
-order of these is defined as the source order of the augmentations.
+Within a single augmentation file, you may augment the same declaration multiple
+times, whether they are top level or nested declarations. The merge order of
```
lrhn, line 187: Nit: "they" should refer to a plural object, but I think it refers to "same declaration", so "it is" (or better "the declaration is", unqualified "it"s risk being ambiguous.)

working/augmentation-libraries/feature-specification.md
```diff
@@ -142,59 +141,58 @@ It is a compile-time error if:
...
+these is defined as the source order of the augmentations.
```
lrhn, line 188: of the agumentation declarations. 

(Source order imples syntax, declarations are syntax. And the unqualified "augmentations" here is an example of what would be ambiguous if "augmentation" was defined as a library augmentation.)

working/augmentation-libraries/feature-specification.md
```diff
@@ -301,7 +299,7 @@ It is a compile-time error if:
 
 *   An augmenting declaration appears in a library before the library where the
     original declaration occurs, according to merge order. *An augmentation
-    library can both declare a new declaration and augment it in the same file.*
+    can both declare a new declaration and augment it in the same file.*
```
lrhn, line 302: (Can it augment the same member twice in the same `augment class`? ... I know, I should just check. Should it?)

working/augmentation-libraries/feature-specification.md
```diff
@@ -406,7 +404,7 @@ It is a compile-time error if:
     doesn't accomplish anything semantically. But it ensures that anyone reading
     the augmenting type can see the declarations of any type parameters that it
     uses in its body and avoids potential confusion with other top-level
-    variables that might be in scope in the augmentation library.*
+    variables that might be in scope in the augmentation.*
```
lrhn, line 407: augmentation file.

working/augmentation-libraries/feature-specification.md
```diff
@@ -494,14 +492,14 @@ replaces the variable's implicit getter body with the augmenting getter's.
...
+    in the main library. Inside the augmenting body, an `augmented` expression
+    invokes the original getter.
 
-*   **Augmenting with a setter:** A setter in an augmentation library can
-    augment a setter in the main library or the implicit setter defined by a
-    non-final variable in the main library. Inside the augmenting setter, an
+*   **Augmenting with a setter:** A setter in an augmentation can augment a
+    setter in the main library or the implicit setter defined by a non-final
+    variable in the main library. Inside the augmenting setter, an
     `augmented =` expression invokes the original setter.
```
lrhn, line 503: Drive-by: Consider showing the operand: 
> <code>augmented = _expr_</code> expression ...

working/augmentation-libraries/feature-specification.md
```diff
@@ -573,8 +571,8 @@ It is a compile-time error if:
     a variable with an initializer.
 
 *   A final variable is augmented with a setter. (Instead, the augmentation
-    library can declare a *non-augmenting* setter that goes alongside the
-    implicit getter defined by the final variable.)
+    can declare a *non-augmenting* setter that goes alongside the implicit
+    getter defined by the final variable.)
```
lrhn, line 575: (This is the general rule of "you can't augment what doesn't exist", and a final variable doesn't introduce a setter, so an `augment` setter declaration cannot augment what isn't there. Right?)

working/augmentation-libraries/feature-specification.md
```diff
@@ -599,40 +597,41 @@ An augmenting enum value is allowed to invoke a different constructor than
 the original enum value, or provide an argument list where none was present
 before.
 
-New enum values may also be defined in the augmenting library, and they will
-be appended to the original values in source and augmentation traversal order.
+New enum values may also be defined in the augmentation, and they will be
+appended to the original values in source and augmentation traversal order.
```
lrhn, line 601: (I'd try to define it so that there is only one order, which is source order within a file, then recursive traverseal of the augmentation files of `augment` directives in the current file, in their source order. Then this is just "in augmentation traversal order", the source order within a file is included in that. Otherwise pedants might start asking what the priority is between "source" and "augmentation traversal" order. They do that.)

working/augmentation-libraries/feature-specification.md
```diff
@@ -710,13 +709,13 @@ top-level scope where declarations are defined. They also share a single private
 namespace. This means that private declarations in the main library or an
 augmentation are visible to all augmentations.
 
-Unlike part files, an augmentation library has its own import scope surrounding
+Unlike part files, an augmentation has its own import scope surrounding that
```
lrhn, line 712: augmentation file

also: "that \\n that" -> "that"

working/augmentation-libraries/feature-specification.md
```diff
@@ -710,13 +709,13 @@ top-level scope where declarations are defined. They also share a single private
...
 that shared top-level scope. Any libraries the augmentation imports are visible
```
lrhn, line 713: the augmentation file imports
or
the library augmentation imports

(imports can be treated both syntactically or semantically, either works)

working/augmentation-libraries/feature-specification.md
```diff
@@ -752,8 +751,8 @@ augment class C {
 ```
 
 This code is fine. Code in C in the main library can refer to members added in
```
lrhn, line 753: "in C" -> "in the class `C`"

(Code quote source names, and make it clear that there is no C (the language) code here.)

working/augmentation-libraries/feature-specification.md
```diff
@@ -1,9 +1,9 @@
-# Augmentation Libraries
+# Augmentations
 
 Author: rnystrom@google.com, jakemac@google.com
 Version: 1.13 (see [Changelog](#Changelog) at end)
 
-Augmentation libraries allow splitting a Dart library into files. Unlike part
+Augmentations allow splitting a Dart library into files. Unlike part
```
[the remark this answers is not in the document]
munificent: Yeah, I think I'm with Lasse on this. Part of why I called these "augmentation libraries" is to avoid the ambiguity of "augmentation" on its own.

Initially, I called them "augmentation libraries" to emphasize that (unlike part files), they are library-like in that they can have their own imports. This is less true now that they share a private scope with the main library, so removing "library" from the name probably makes sense.

Maybe "library augmentation"? So the file itself isn't a library, but it is a thing that augments a library.

working/augmentation-libraries/feature-specification.md
```diff
@@ -115,7 +114,7 @@ library. Augmentations are distinguished by a special `library` directive with
 an `augment` modifier, like so:
 
 ```dart
-library augment 'main_library.dart';
+augment library 'main_library.dart';
```
munificent, line 117: For what it's worth, my thinking with the original syntax was that you would read `library augment 'foo.dart'` as "a `library` that `augment`s `'foo.dart'`".

But I like the syntax you have here much better. I like that it parallels `augment` as a modifier on other declarations where `augment` precedes the thing it augments.

working/augmentation-libraries/feature-specification.md
```diff
@@ -142,59 +141,58 @@ It is a compile-time error if:
...
+### Applying an augmentation
 
-A main library applies an augmentation to itself using a new import directive
-that looks like this:
+A main library applies an augmentation to itself using a new `augment` directive
+which looks like this:
 
 ```dart
-import augment 'some_augmentation.dart';
+augment 'some_augmentation.dart';
```
munificent, line 152: This reads strangely to me now because in all the other places where we use `augment` it means that the subsequent thing is the thing *being augmented* but here it means *thing doing the augmenting*.

I don't have any great suggestions, but maybe keep `import augment` because that might be a little clearer? Perhaps `with augment`?

munificent (changes requested):
I like changing `library augment` to `augment library`. I'm not as keen on changing `import augment` to just `augment` and I think using "augmentation" as a standalone noun will end up confusing.

But, overall, I think this is a step in the right direction. Thanks for taking my strawman syntax and beating it into shape. :)

working/augmentation-libraries/feature-specification.md
jakemac53: Yes I agree, this is the one part I don't like now, for the same reasons. I suggested "augmentation" above for this but then its yet another special word to introduce.

munificent: It would be nice to not have to introduce every conjugation of "augment" into the language: `augment`, `augments`, `augmentation`, `augmenting`, `augmentable`, `augmenter`, etc.. :)

I think `import augment` works OK.

jakemac53: Fwiw, to me this is the strongest argument for just re-using the "part" directive 🤣. But, we would be breaking the semantics of it, so I think that is a no-go. But, we could just be bold...

munificent: > Fwiw, to me this is the strongest argument for just re-using the "part" directive 🤣.

I would *love* to repurpose part files for this feature. "Augment" works OK as a word, but it's not exactly common and is pretty hard for ESL speakers from what I've gathered.

> But, we would be breaking the semantics of it, so I think that is a no-go. But, we could just be bold...

Yeah, it's probably a bridge too far. :( From a quick scrape of a pub package corpus:

```
-- File (56652 total) --
  45719 ( 80.701%): no part files    ===================================
   7759 ( 13.696%): is part file     ======
   3174 (  5.603%): uses part files  ===
```

So part files are actually pretty common. I suspect almost all of those are generated code?

jakemac53: > So part files are actually pretty common. I suspect almost all of those are generated code?

Could you check how many of these files have multiple extensions, as a proxy for generated files (*.*.dart)?

munificent: From skimming the file paths as they blew by on my terminal, I didn't notice a lot of extensions that suggested they were generated.

lrhn: We could use `part` here, and then whether the part file starts with `part of ....` or `augment library ...` decides whether it can have imports or not. Whether it's a simple part file or a library augmentation part file. (But if we allow `augment` in every file, #3576, we no longer have library augmentations, we just have files which may or may not include member augmentation. Then it's just about whether that file allows imports or it inherits the parent file's imports. At which point the inclusion syntax shouldn't mention `augment`, and I don't know which name would fit so it could be a `*word* part of ...` vs `part of ....`. )

jakemac53:
> @jakemac53 what's the status of these changes? I want to update the textmate grammar (and LSP semantic tokens) but I was waiting for this to ensure any tests I add are using the correct code (definitely not urgent, just trying to clear some things off my list :-))

The status is I am overwhelmed 🤣 . I will try to circle back here soon though, thanks for the ping.

working/augmentation-libraries/feature-specification.md
```diff
@@ -820,15 +819,15 @@ libraryAugmentImport ::= metadata 'import' 'augment' uri ';'
 ```
 
 A library directive may contain `augment` followed by a URI to denote the file
-as an augmentation library:
+as an augmentation:
```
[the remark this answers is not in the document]
jakemac53: Added a TODO, I want to tackle this separately.

working/augmentation-libraries/feature-specification.md
```diff
@@ -903,11 +902,10 @@ To apply an augmentation to the main library:
 1.  Merge the augmentation's declarations into the main library's top-level
     namespace using the following procedure.
 
-1.  For each `import augment` directive in the augmentation library, in
-    syntactic order:
+1.  For each `augment` directive in the augmentation, in syntactic order:
```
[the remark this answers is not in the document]
jakemac53: That reads weird I think, I made the change above though, which helps.

working/augmentation-libraries/feature-specification.md
```diff
@@ -66,20 +66,19 @@ part of macro expansion, it's still useful to have it be in a canonical
...
 into smaller files.
 
-## Augmentation libraries
+## Augmentations
 
-An augmentation library is a separate file that *augments* another *main*
-library. An augmentation is somewhere between a separate library and a part
-file.
+An augmentation is a separate file that *augments* another *main* library. An
+augmentation is somewhere between a separate library and a part file.
```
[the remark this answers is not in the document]
jakemac53: I reworked this entirely

jakemac53:
I pushed some changes based on the feedback here but didn't get to it all yet, will try to finish up tomorrow.

working/augmentation-libraries/feature-specification.md
```diff
@@ -1,12 +1,12 @@
...
-Augmentations allow splitting a Dart library into files. Unlike part
-files, each augmentation has its [own imports][part imports] and top-level
-scope. Augmentations can add new top-level declarations, inject new members into
-classes, and wrap functions and variables in additional code.
+Augmentations allow spreading your implementation across multiple locations,
+both within a single file and across multiple files. They can add new top-level
+declarations, inject new members into classes, and wrap functions and variables
+in additional code.
 
 [part imports]: https://github.com/dart-lang/language/issues/519
```
munificent, line 11: You can remove this link reference if you're removing the bit about part files from the intro.

working/augmentation-libraries/feature-specification.md
```diff
@@ -66,51 +66,51 @@ part of macro expansion, it's still useful to have it be in a canonical
...
-    library, or provide a body if none was present.
+*   Function augmentations, which can replace the body of a function, or provide
+    a body if none was present.
 
-*   A variable in the augmentation may wrap the initializer of a variable in the
-    main library, or provide an initializer if none was present.
-
-*   An enum value in the augmentation may replace the argument list of an enum
-    value in the main library, or provide an argument list if none was present.
+*   Variable augmentations may wrap the initializer of a variable in the
```
munificent, line 105: Because I'm a tech writing nerd now and list bullets lists to have parallel grammar structure:

> Variable augmentations, which can wrap...

working/augmentation-libraries/feature-specification.md
```diff
@@ -66,51 +66,51 @@ part of macro expansion, it's still useful to have it be in a canonical
...
+    augmented library, or provide an initializer if none was present.
 
 These can't be expressed today using only imports, exports, and part files.
 
-### Defining an augmentation
+### Defining an library augmentation
```
munificent, line 110: "an" -> "a".

working/augmentation-libraries/feature-specification.md
```diff
@@ -66,51 +66,51 @@ part of macro expansion, it's still useful to have it be in a canonical
...
 
-An augmentation has almost the same syntax and semantics as a normal Dart
-library. Augmentations are distinguished by a special `library` directive with
+An library augmentation has almost the same syntax and semantics as a normal
```
munificent, line 112: Here too.

working/augmentation-libraries/feature-specification.md
```diff
@@ -122,30 +122,34 @@ syntax][import]?**
 
 [import]: https://github.com/dart-lang/language/blob/master/working/0649%20-%20Import%20shorthand/proposal.md
 
-The URL points to the main library that this augmentation is applied to.
+The URL points to the augmented library that this augmentation is applied to,
+which must be a regular library (and not a library augmentation).
```
munificent, line 126: The parenthetical implies that a hand-authored library augmentation can't contain macro applications. Is that deliberate?

The "Merge Order" section still says that an entire tree of library augmentations is possible. Or is it that every library augmentation must point all the way back to the augmented library and not any of the library augmentations?

working/augmentation-libraries/feature-specification.md
```diff
@@ -122,30 +122,34 @@ syntax][import]?**
...
+*   A library augmentation contains a normal `library` directive. They are not
+    self-contained libraries, only pieces of the augmented library.
+
 ### Applying an augmentation
 
-A main library applies an augmentation to itself using a new `augment` directive
+A library applies an augmentation to itself using a new `augment` directive
 which looks like this:
 
 ```dart
```
munificent, line 155: I'm iffy on this syntax. It reads to me like it's saying that *this* file is augmenting `'some_augmentation.dart'` when it's the exact opposite.

This used to be `import augment`, right? I think I still lean towards that syntax.

working/augmentation-libraries/feature-specification.md
```diff
@@ -239,12 +242,13 @@ augment void trace() {
...
-declarations in those libraries&mdash;new declarations or augmentations&mdash;
-are processed in that order, and source order within that.
+declarations in those library augmentations
+&mdash;new declarations or augmentations&mdash; are processed in that order,
+and source order within that.
 
 This order is user-visible in two ways:
 
-*   A non-augmenting declaration must appear first before it can be augmented.
+*   A regular declaration must appear first before it can be augmented.
```
munificent, line 251: Nit: I do like "non-augmenting" here because it's a little less vague than "regular", which isn't explicitly defined.

Or maybe do:

> A regular (i.e. not augmenting)...

working/augmentation-libraries/feature-specification.md
```diff
@@ -785,45 +789,47 @@ some_lib.dart       | some_augment.dart
...
-3.  Resolve and type-check the main library and all of its augmentations now
-    that all type namespaces are complete.
+3.  Resolve and type-check the augmented library and all of its augmentations
+    now that all type namespaces are complete.
 
 ## Syntax
 
-The grammar changes are fairly simple. A main library can apply an augmentation
-using a new directive:
+The grammar changes are fairly simple. A library can apply an augmentation using
```
munificent, line 814: Heh, I suppose this first sentence as a little too optimistic about simplicity. :)

working/augmentation-libraries/feature-specification.md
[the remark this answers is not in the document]
munificent: I'll help out with the grammar changes.

working/augmentation-libraries/feature-specification.md
```diff
@@ -930,26 +936,26 @@ To merge a set of declarations `D` into a namespace:
...
+            namespace of the augmented type.
 
     1.  Else, if the declaration is a function, getter, setter, or operator:
 
-        1.  Replace the body of the main function with the augmenting function's
-            body. Inside the augmenting body, a `super augment()`, `super
-            augment`, `super augment =`, or `super augment <op>` expression as
-            appropriate calls the original function body.
+        1.  Replace the body of the augmented function with the augmenting
+            function's body. Inside the augmenting body, a `augmented()`,
```
munificent, line 956: Nit: "a `augmented()`" -> "an `augmented()`"

munificent (changes requested):
I'd like to talk more about whether we want `augment` or `import augment` (and I'm sorry if we discussed that before I don't remember all the details now). Aside from that, this LGTM.

working/augmentation-libraries/feature-specification.md
jakemac53: Yeah, I will probably revert this part, as I agree it is a bit weird. I don't like either of them really, but I don't have an alternative proposal either 🤷‍♂️ .

working/augmentation-libraries/feature-specification.md
jakemac53: Yes it can, and we do this in the macro generated augmentations to make them more concise, if multiple macros augment the same member

working/augmentation-libraries/feature-specification.md
jakemac53: Yes, I don't think it hurts to explicitly call it out though? But that is the reasoning for sure.

working/augmentation-libraries/feature-specification.md
jakemac53: I changed to just "in augmentation traversal order", that is what you meant right? :)

jakemac53:
Updated to change `augment <uri>;` back to `import augment <uri>;`, addressed the rest of the feedback, PTAL

munificent (approved):
This looks great!

lrhn:
My only worry with `augment import` is that it's long.
There have been suggestions that it should replace `part` as the way to modularize a library. Then it might count against it that the syntax to include focuses on the augmentation, even if the file it referred to doesn't contain any `augment` declarations at all.

(But I guess this is me coming back to suggesting that we just use `part`, allow augmentation declarations in all files, allow import, export and part declarations in part files, and if a part file declares any import, then it doesn't get the parent file's imports. We don't need a new kind of file, another way to include declarations, we already have `part`.)

jakemac53:
> My only worry with `augment import` is that it's long.

I agree, I do think we might want to come back to this, with some new keyword or by re-using `part`. So this isn't the end of that discussion, I just wasn't happy with a bare `augment` as the thing.

jakemac53:
cc @johnniwinther @scheglov note that this change does presumably require some worth for the CFE/Analyzer and the code that generates augmentation libraries. I am not sure how we want to handle those changes in the best way.

scheglov:
I like this change, it reads consistently with other augmentations.

https://dart-review.googlesource.com/c/sdk/+/358060 probably prepares `analyzer/` and `analysis_server/` for the parser change. I cannot land it, of course. So, it is to be merged with the parse changes.

PR #3595, opened 2024-01-31
Stop pinning deps in working/macros/example/pubspec.yaml.

stereotype441:
Pinning the deps was creating a version skew problem in continuous integration (which is configured in `.github/workflows/dart.yml`). The continuous integration configuration uses `sdk: main` (meaning that the code will be analyzed using the latest bleeding edge build of Dart). But the macros in `working/macros/example` import macro support functionality from `package:_fe_analyzer_shared/src`, so they use whatever version of that package that is configured in the pubspec (which, prior to this change, was pinned to version `3.4.0-56.0.dev`).

The reason this creates a version skew problem is that the macro support functionality in `package:_fe_analyzer_shared/src` includes deserialization logic that decodes data structures created during analysis and compilation. Which means that any time the serialization format changes, that could potentially break the continuous integration bot.

Eventually we plan to move the macro support functionality into a `dart:` library that will ship as part of the SDK, so there will be no more version skew problem. Until then, an easy way to work around the version skew is simply to pin the pubspec to the `main` branch rather than a specific release, so that it matches what's configured in continuous integration.

A side effect of this change is that if the API for writing macros gets changed in a way that breaks the examples in
`working/macros/example`, that will cause the continuous integration bot to immediately start failing. (Previously it wouldn't cause a failure because the API was pinned to `3.4.0-56.0.dev`). Considering that the API for macros is under active development, this is probably a good thing; it will mean that if we make inadvertent API changes that break the examples, we'll be more likely to find out about it.

---

- [ ] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>

</details>

### DIFF
--- a/working/macros/example/pubspec.yaml
+++ b/working/macros/example/pubspec.yaml
@@ -17,84 +17,84 @@ dependency_overrides:
     git:
       url: https://github.com/dart-lang/sdk.git
       path: pkg/_fe_analyzer_shared
-      ref: 3.4.0-56.0.dev
+      ref: main
   _js_interop_checks:
     git:
       url: https://github.com/dart-lang/sdk.git
       path: pkg/_js_interop_checks
-      ref: 3.4.0-56.0.dev
+      ref: main
   analyzer:
     git:
       url: https://github.com/dart-lang/sdk.git
       path: pkg/analyzer
-      ref: 3.4.0-56.0.dev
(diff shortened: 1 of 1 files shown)


jakemac53 (approved):

jakemac53:
Thanks for looking into this!

PR #3598, opened 2024-01-31
Adjust incompatible-with-await to be more strict on intersections

eernstg:
The PR https://github.com/dart-lang/language/pull/3560 introduced the notion of a type which is 'incompatible with await' and used that to introduce an error on `await e` when the static type of `e` is such a type.

This PR adjusts the definition of 'incompatible with await' such that `X & S` is incompatible with await when `S` is incompatible with await. This is a bit more strict than the previous version. It is motivated by the fact that a promotion of a variable of type `X` to a type `X & S` where `S` isn't compatible with await is a strong signal that it should not be awaited.

Of course, `await (e as T)` for some suitable value of `T` can always be used to force the await.

The change is not breaking because 'incompatible with await' has not yet been implemented in any release.

### DIFF
--- a/accepted/future-releases/extension-types/feature-specification.md
+++ b/accepted/future-releases/extension-types/feature-specification.md
@@ -885,10 +885,7 @@ one of the following criteria holds:
 
 - `T` is an extension type that does not implement `Future`.
 - `T` is `S?`, and `S` is incompatible with await.
-- `T` is `X & B`, and either:
-  - `B` is incompatible with await, or
-  - `B` does not derive a future type, and `X` is
-    incompatible with await.
+- `T` is `X & B`, and `B` is incompatible with await.
 - `T` is a type variable with bound `S`, and `S` is incompatible
   with await.
 


accepted/future-releases/extension-types/feature-specification.md
```diff
@@ -885,10 +885,7 @@ one of the following criteria holds:
 
 - `T` is an extension type that does not implement `Future`.
 - `T` is `S?`, and `S` is incompatible with await.
-- `T` is `X & B`, and either:
-  - `B` is incompatible with await, or
-  - `B` does not derive a future type, and `X` is
-    incompatible with await.
+- `T` is `X & B`, and `B` is incompatible with await.
```
lrhn, line 888: Should be sufficient.

If `B` is compatible with await, then either `B`:
* is not a (nullable-bounded-)extension type, which means the bound of `X` is also not an extension type. In which case *flatten* will not yield a (nullable-)extension type.
* or is an extension type which implements `Future`, in which case that's the future which will be awaited as expected.

In either case, the *flatten* doing "if `B` does not derive a future type, use *flatten*(`X`)` will not end up using a non-future-extension-type as a type to flatten, which is precisely what we want: The user never needing to know how to flatten a non-future-extension-type.

So LGTM!

lrhn (approved):

PR #3613, opened 2024-02-13
Fix records type inference rules to match implementations.

stereotype441:
If the context of a record literal is the same shape as the record literal, but the actual type of one or more fields is not a subtype of the corresponding field in the context type, and no coercion is possible, then the implementations do not necessarily issue an error. They only issue an error if the type mismatch leads to an assignability error during subsequent analysis.

An example of a circumstance where the type mismatch doesn't lead to an assignability error is if the context resulted from assignment to a previously promoted local variable. For example:

    f(Object o) {
      if (o is (int,)) {
        o = ('',); // OK; demotes `o` back to `Object`.
      }
    }

This change adjusts the spec to agree with what was implemented.

---

<details>

</details>

### DIFF
--- a/accepted/3.0/records/feature-specification.md
+++ b/accepted/3.0/records/feature-specification.md
@@ -524,7 +524,7 @@ dm : K{n+m}})` then:
       appropriate implicit coercion(s) on `ei`.  Let `Ti` be the type of the
       resulting coerced value (which must be a subtype of `Ri`, possibly
       proper).
-    - Otherwise, it is a static error.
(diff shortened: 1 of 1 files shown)


munificent:
This change looks good to me, but @leafpetersen wrote the inference section for records and he'll understand much better than I would if there are complications here that I'm not seeing.

munificent (approved):

leafpetersen (commented):
LGTM, thanks.

leafpetersen (approved):

PR #3630, opened 2024-02-27
Specify the interaction between record expressions and constant contexts

eernstg:
See https://github.com/dart-lang/sdk/issues/55008#issuecomment-1964229795 for some background information.

The feature specification about records does not specify the meaning of `const` in front of a record expression, but it does allow `const` in that position. This PR adds a paragraph about this question. The idea is that `e` occurs in a constant context if it occurs as `const (..., e, ...)` or if `e` occurs in a record expression `(..., e, ...)` that occurs in a constant context.

This would not be a breaking change: The CFE already behaves in this way. The analyzer accepts `const (...)` where `(...)` is a record expressions where one or more field expressions are non-constant, so there is a certain amount of work to do on the analyzer.

### DIFF
--- a/accepted/3.0/records/feature-specification.md
+++ b/accepted/3.0/records/feature-specification.md
@@ -133,6 +133,10 @@ not captured by the grammar. It is a compile-time error if a record has any of:
     field. *For example: `('pos', $1: 'named')` since the named field "$1"
     collides with the getter for the first positional field.*
 
+Let _R_ be a record expression that occurs in a constant context, or a
+record expression whose first token is `const`. In this case, each record
+field in _R_ occurs in a constant context.
+
 In order to avoid ambiguity with parenthesized expressions, a record with
 only a single positional field must have a trailing comma:
 


lrhn (approved):

eernstg:
Thanks! I'll await input from Bob as well.

munificent (approved):
Haha, oh dear. Allowing `const` in the grammar was a mistake. I put it in there early on in the proposal with a TODO to figure out the semantics around const records. Then, later, Lasse had the brilliant suggestion that we could infer const-ness based on the fields, so `const` before a record isn't required. I incorporated that change but, uh, never removed `const` from the grammar. And it looks like our implementation friends went ahead and implemented it. 🙃 

I *do* like the idea of allowing `const` there and having it just mean "const context", so I think what they implemented and the change you have here makes sense @eernstg. (In fact, I'd consider eventually going farther and allowing `const (<expr>)` to establish a constant context around any expression, like we do in patterns.)

Thanks for the fix!

bwilkerson:
> ... we could infer const-ness based on the fields, so `const` before a record isn't required.

Now that the syntax is allowed and well defined, should we have a lint to catch the unnecessary use of `const` before a record literal? :-/

eernstg:
> should we have a lint to catch the unnecessary use of const before a record literal?

Is it unnecessary? If one or more of the expressions in the record literal is not constant then it is an error, and this is presumably a useful heads-up to a developer who _intended_ all those expressions to be constant. So we probably don't think that's an unnecessary `const`. If it is _not_ an error then it may serve as documentation, and it might ensure that `"a" + "b"` is canonicalized, which again makes it necessary.

```dart
void main() {
  print(identical(const ("a" + "b",), const ("ab",))); // 'true'.
  print(identical(("a" + "b",), ("ab",))); // 'false'.
}
```

So, as far as I can see, the only obvious way we could encounter an unnecessary `const` in front of a record literal is when that record literal is itself in a constant context, as in `const (const (1,),)`. I think it does make sense to report the inner `const` in that case.

lrhn:
Well, `print(identical(("a" + "b",), ("ab",)));` can always print `false`, records do not have consistent identity, not even constant records, not unless the `identical` invocation is itself constant, so that proves nothing.

A similar example that shows the difference could be:
```dart
void main() {
  var v1 = ("a" + "b",);
  var v2 = const ("a" + "b",);
  print(identical("ab", v1.$1)); // False in VM
  print(identical("ab", v2.$1)); // True in VM
}
```

There is nothing you can do with `const (e1, e2)` that you can't with `(const e1, const e2)` (where `const` makes sense on `e1` or `e2`, otherwise remove it).

Whether the `const` can be removed entirely depends on whether any field is a collection literal without a `const`.
That is, whether the subexpression *would* have been constant even if not in a constant context.
Have to get that right, which is not an issue for other containers, because those are all inherently constant or not.
(Records are also the only collection of values that can be *potentially constant*).

bwilkerson:
Ok, thanks for the discussion. It sounds like a lint is unlikely to add significant value, so I think we won't add one unless we get user feedback requesting it.

PR #3638, opened 2024-03-04
[macros] Fix data_class_main.dart macros example

sgrekhov:
`CopyWith().buildDeclarationsForClass(clazz, context)` generates parameters for `copyWith()` constructor as 
```dart
        ParameterCode(
            name: field.identifier.name,
            type: field.type.code.asNullable,
            keywords: const [],
            defaultValue: null)
```
So, non-nullable parameters can't have `null` as default value, therefore code, generated by macros has a compile-time error. This change fixes it

### DIFF
--- a/working/macros/example/bin/data_class_main.dart
+++ b/working/macros/example/bin/data_class_main.dart
@@ -14,9 +14,9 @@ void main() {
 
 @DataClass()
 class User {
-  final int age;
-  final String name;
-  final String username;
+  final int? age;
+  final String? name;
+  final String? username;
 }
 
 @DataClass()


jakemac53:
This change should not be necessary - for the constructor they are all listed as required params and for copyWith we actually pass `age ?? this.age` as the value, so it is always non-null.

jakemac53:
It is likely that you are seeing errors here trying to run the main script as an actual program - it looks like the CFE just doesn't fully support that yet. The analyzer does however and won't complain. So, this is just waiting on full CFE support.

sgrekhov:
> It is likely that you are seeing errors here trying to run the main script as an actual program - it looks like the CFE just doesn't fully support that yet. The analyzer does however and won't complain. So, this is just waiting on full CFE support.

Yes, I'm trying to run the main script. Ok, thank you!

jakemac53:
> `ChecksExtensions` macros expects `List<TypeAnnotation>` argument, but actual type is `List<Type>`. Is this an issue or there should be some implicit conversion from `Type` to `TypeAnnotation`?

Yes there is supposed to be an implicit conversion here.

> If I see more such issues what's the best way to proceed? To fill an issue, add github comment to the appropriate code line, try to make a fix?

I wouldn't file issues at this time - the implementations are pretty far off from being complete, so there are a lot of edge cases here.

Language tests would be the most useful way to help out, that is the best way to basically document that something isn't yet implemented (by approving the failure). I am not sure if a bunch of hodgepodge language tests for known failures is that useful either though, we really need to tackle each part of the feature holistically.

Issue #3656, opened 2024-03-15
Failing analysis of macro test prevents other actions


eernstg:
The Dart CI log at https://github.com/dart-lang/language/actions/runs/8084567444/job/22725574044 shows a failure during the analyzer task `dart analyze --fatal-infos .`, in working/macros/example/benchmark/src/shared.dart:91:71 and elsewhere.

This causes other tasks to remain undone (in this case: no new PDF is created for the language specification when it is updated).

@jakemac53, would it be possible to fix the example, or somehow adjust the processing such that it doesn't block other tasks?

jakemac53:
> @jakemac53, would it be possible to fix the example, or somehow adjust the processing such that it doesn't block other tasks?

Yes we can allow it to fail, I can do that

lrhn:
(@davidmorgan Saw the "S" label. Consider if it can be combined with the ["small-feature"](https://github.com/dart-lang/language/issues?q=is%3Aissue+is%3Aopen+label%3Asmall-feature) label.)

jakemac53:
The goal is to have unified labels for this across repos, and these aren't so much individual language features so I don't know that small-feature is a good fit.

davidmorgan:
Thanks @lrhn :) as Jake says I think it's different, we'd like to have S/M/L/XL for project planning.

PR #3660, opened 2024-03-18
allow macro analysis step to fail without stopping other jobs

jakemac53:
Fixes https://github.com/dart-lang/language/issues/3656

~This should resolve it, I am going to push an intentional failure and then revert to test the behavior, so please wait to land until I have a chance to do that :).~ Tested and it works as expected.

Note that if there is a particular job we always want to run, we could also try adding `if: always()` to that, which might actually be an additional change we want to make.


davidmorgan:
LGTM :)

jakemac53:
Closing in favor of https://github.com/dart-lang/language/pull/3878

PR #3663, opened 2024-03-19
Update grammar rules of augmentation library feature specification

eernstg:
This PR updates various grammar rules of the augmentation library feature specification. In particular, it adds rules to specify augmented extension, extension type, and enumerated declarations, and it makes several adjustments to the `topLevelDecrlaration` and `declaration` derivation rules in order to be up to date with other features.

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -850,7 +850,7 @@ topLevelDeclaration ::= classDeclaration
   | 'augment'? functionSignature functionBody
   | 'augment'? getterSignature functionBody
   | 'augment'? setterSignature functionBody
-  | 'augment'? ('final' | 'const') type? staticFinalDeclarationList ';'
+  | ('final' | 'const') type? staticFinalDeclarationList ';'
   | 'augment'? 'late' 'final' type? initializedIdentifierList ';'
   | 'augment'? 'late'? varOrType initializedIdentifierList ';'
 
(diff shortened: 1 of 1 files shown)


eernstg:
One case may need an extra comment: I did not include `'augment'?` in the rule for

```ebnf
<declaration> ::= ...
  | 'covariant' 'late' 'final' <type>? <identifierList>
  | ...
```

The reason is that `covariant` is associated with the setter (a getter does not have any parameters, so it can't be covariant), and a `late final` variable with an initializing expression does not have a setter whereas the one with no initializing expression does have a setter.

So this rule is needed in order to ensure that we can declare relevant covariant instance variables, but we can't augment them: If the only thing we can do is to add an initializing expression, and this kind of variable can't have those, the `augment` declaration wouldn't be able to do anything. Hence no `augment` on that one.

eernstg:
Perhaps those variable declarations can be augmented anyway, if we can add metadata?

munificent (approved):
This looks great, thank you!

jakemac53:
> Perhaps those variable declarations can be augmented anyway, if we can add metadata?

I suppose so, yes.

eernstg:
> I suppose so, yes.

OK, I just added `'augment'?` to those variable declarations to enable metadata (and whatever else might come up).

jakemac53 (approved):
Thanks a ton for doing this @eernstg! 🚀

eernstg:
Thanks!

PR #3667, opened 2024-03-20
Augmentation grammar: Adjust `libraryName` and remove `methodSignature` rule

eernstg:
Just noted a couple of places where I think the augmentation grammar should be adjusted. We used to have this:

```ebnf
libraryName ::= metadata 'augment'? 'library'
    ( dottedIdentifierList? |  uri ) ';'
```

but that allows for things like `library 'package:whatever/whatever.dart';` which is not useful. So I adjusted the rule to allow the `uri` with an `augment library` and not with a plain `library`.

I removed the rule for `methodSignature` because it only changed by having `'augment'?` in some alternatives, but `classMemberDefinition` already has `'augment'?` in front of `methodSignature`, and we don't need to allow two occurrences of `augment` on the same declaration.

Finally, I swapped the alternatives for `extensionDeclaration` and `extensionTypeDeclaration` in the `topLevelDeclaration` rule, to match the approach taken in `Dart.g` (we may as well parse `extension type ...` as an extension type rather than trying to make it an `extension` named `type`, and then retrying).

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -829,8 +829,10 @@ as an augmentation:
 TODO: Create special augmentation grammar, similar to library/part files?
 
 ```
-libraryName ::= metadata 'augment'? 'library'
-    ( dottedIdentifierList? |  uri ) ';'
+libraryName ::= metadata libraryNameBody ';'
+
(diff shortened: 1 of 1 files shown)


jakemac53 (approved):
Thanks!

eernstg:
Thanks! @munificent, please create a new issue if you have further comments.

Issue #3665, opened 2024-03-19
Spec is ambiguous about class hierarchy of enums; leads to observable differences in front end and analyzer behavior.


stereotype441:
(This came up organically in a project I'm working on with @sobrienpdx. I've reduced it down to a minimal repro.)

The following code is accepted by the analyzer but rejected by the front end:
```dart
abstract class A {
  String get s;
}

abstract class B implements A {}

enum E1 implements B {
  e1;

  @override
  String get s => 'E1';
}

enum E2 implements B {
  e2;

  @override
  String get s => 'E2';
}

final allValues = [...E1.values, ...E2.values];

main() {
  for (var value in allValues) {
    print(value.s);
  }
}
```

The front end issues this error message:
```
Error: The getter 's' isn't defined for the class 'Object'.
 - 'Object' is from 'dart:core'.
Try correcting the name to the name of an existing getter, or defining a getter or field named 's'.
    print(value.s);
                ^
```

After some digging, I figured out that the difference comes down to a subtle interaction between least upper bound and an ambiguity in the enhanced enums spec.

The spec doesn't fully pin down what the class hierarchy of an enum should look like. [It says](https://github.com/dart-lang/language/blob/main/accepted/2.17/enhanced-enums/feature-specification.md#implementation):

> The specification here does not specify *how* the index and name of an enum is associated with the enum instances. In practice it’s possible to desugar an `enum` declaration to a `class` declaration, as long as the desugaring can access private names from other libraries (`dart:core` in particular).
>
> The existing enums are implemented as desugaring into a class extending a private `_Enum` class which holds the `final int index;` declaration and a `final String _name;` declaration (used by the the `EnumName.name` getter), and both fields are initialized by a constructor. 
>
> In practice, the implementation of the enhanced enums will likely be something similar.
>
> Either first declare `Enum` as:
>
> ```dart
> abstract class Enum {
>   Enum._(this.index, this._name);  
>   final int index;
>   final String _name;
>   String _$enumToString();
>   String toString() => _$enumToString();
> }
> ```
>
> *or* retain the current `_Enum` class and make that the actual superclass of `enum` classes. Either works, I’ll use `Enum` as the superclass directly in the following.

What was implemented was:
- In the analyzer, all enums extend `Enum`, which extends `Object`.
- In the front end, all enums extend `_Enum`, which in turn implements `Enum`, which extends `Object`.

This means that, when analyzing the code example above, the analyzer considers the class hierarchy to be:
```mermaid
[...]

PR #3671, opened 2024-03-21
Specify that enums extend `_Enum` from `dart:core`.

stereotype441:
See
https://github.com/dart-lang/language/issues/3665#issuecomment-2010433135 for details.

---

<details>

</details>

### DIFF
--- a/accepted/2.17/enhanced-enums/feature-specification.md
+++ b/accepted/2.17/enhanced-enums/feature-specification.md
@@ -178,25 +178,9 @@ The recommended formatting of an `enum` declaration is to format the header (bef
 
 The specification here does not specify *how* the index and name of an enum is associated with the enum instances. In practice it’s possible to desugar an `enum` declaration to a `class` declaration, as long as the desugaring can access private names from other libraries (`dart:core` in particular).
 
-The existing enums are implemented as desugaring into a class extending a private `_Enum` class which holds the `final int index;` declaration and a `final String _name;` declaration (used by the the `EnumName.name` getter), and both fields are initialized by a constructor. 
+However, since the definition of least upper bound depends on the depth of classes in the class hierarchy, it _is_ necessary to specify precisely where enums fit into the class hierarchy.
 
-In practice, the implementation of the enhanced enums will likely be something similar.
-
-Either first declare `Enum` as:
-
-```dart
-abstract class Enum {
-  Enum._(this.index, this._name);  
-  final int index;
-  final String _name;
-  String _$enumToString();
-  String toString() => _$enumToString();
-}
-```
-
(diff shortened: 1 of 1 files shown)


accepted/2.17/enhanced-enums/feature-specification.md
```diff
@@ -178,25 +178,9 @@ The recommended formatting of an `enum` declaration is to format the header (bef
...
-  final String _name;
-  String _$enumToString();
-  String toString() => _$enumToString();
-}
-```
-
-*or* retain the current `_Enum` class and make that the actual superclass of `enum` classes. Either works, I’ll use `Enum` as the superclass directly in the following.
-
-Then desugar an `enum` declaration to an actual `class` declaration and rewrite every generative constructor of the `enum` declaration to take two extra leading positional arguments. 
+An `enum` declaration is desugared to an actual `class` declaration, which extends a private `_Enum` class in `dart:core`. _This class holds the `final int index;` declaration and a `final String _name;` declaration (used by the the `EnumName.name` getter), and both fields are initialized by a constructor._ Every generative constructor of the `enum` declaration is rewritten to take two extra leading positional arguments.
```
lrhn, line 183: I really wish this wasn't necessary. We're specifying an implementation detail.

Could we at least avoid specifying the *name* of the class? 
Just say that there exists an anonymous class which implements `Enum` and which `enum` classes extend,
so that the *depth* of an `enum` class is at least *depth*(`Enum`) + 2.

That still implies that there is a class at *depth*(`Enum`) + 1 that other classes can conflict with. Can we introduce a third class at depth *depth*(`Enum`) + 1 that `enum` classes also implement, so **UP** can never yield `_Enum`? Like for `class HideEfficientLengthIterable` does for `EfficientLengthIterable`.

lrhn (approved):

accepted/2.17/enhanced-enums/feature-specification.md
srawlins, line 183: typo: "used by the the"

lrhn: Thinking more about this, I'd prefer to just want to make it unspecified behavior.
Or stop extending `_Enum`. It *is* an implementation detail.
Heck, make `_Enum` not implement `Enum`, and it would give it a lower depth. (And make inferring `Enum` impossible, because LUB is moronic.) Or make `_Enum` a superclass of `Enum`.

I really, really do not want to enshrine a completely arbitrary *private* implementation class in the specification. I'd rather make every tool lie about it. That is a more future proof design.

(I do consider if we can make enum classes be allowed to have superclasses, so they only implement `Enum`. Maybe make `_Enum` a mixin then - probably needs more mixin features.)

Issue #3646, opened 2024-03-11
Augmentation libraries can't be main libraries as well?


eernstg:
Thanks to @sgrekhov for bringing up this question. The specification does not seem to mention the case where an augmentation library augments itself:

```
// Library 'main_lib.dart'.
library augment 'main_lib.dart';
import augment 'main_lib.dart';
...
```

Presumably, that would be an error?

It seems reasonable to assume that the `import augment` graph should be a tree, so we'd naturally ask a similar question about this:

```dart
// Library 'a.dart'.
library augment 'b.dart';
import augment 'b.dart';

// Library 'b.dart'.
library augment 'a.dart';
import augment 'a.dart';
```

It would presumably not be hard to create many kinds of cyclic `import augment` graphs, and I think they should all be errors.

Perhaps we need a rule about a regular import:

> It is a compile-time error if a library imports `uri`, and `uri` denotes an augmentation library.

jakemac53:
Yes, we should block cycles like this.

lrhn:
It should be an error. 
If for no other reason than that a library augmentation file is not a library, so you can't import it directly. (It's an error to `import` or `export` a file, if that file is not a library file. A library augmentation file is not a library file.) Also cannot be a `part`-inclusion, that requires a `part` file, which it also isn't.

That means that if your program (entry point library) actually refers transitively to a library augmentation file, then it's through an `augment import` declaration, and then it's immediately a compile-time error if the library augmentation file header doesn't refer back to the parent file. Which it doesn't if it refers to itself.

So, today, it shouldn't be possible to include such a file in a program without getting a compile-time error.
It's really just the analyzer which needs to complain if it starts at a non-library file, and doesn't find a library file by following the parent-file links (`part of` and `augment library` URIs). A cycle would imply that.

We can say the same thing for 
```dart
// foo.dart
part of 'foo.dart';
```
(I don't think we say that this is an error today, but it's an error for a library to include a part file that doesn't point back to the library, which this one never would, so it cannot be part of a valid Dart program.)

eernstg:
Here's a proposal: https://github.com/dart-lang/language/pull/3676.

PR #3676, opened 2024-03-27
Specify errors about wrong `import augment`/`import` structures

eernstg:
Based on the response in https://github.com/dart-lang/language/issues/3646, this PR specifies a compile-time error which occurs if a main library _L_ which is being augmented along with all augmentation libraries _A<sub>1</sub> .. A<sub>k</sub>_ reachable from _L_ via one or more `import augment` links forms a graph that has a cycle.

This PR also makes it an error for a `<uri>` in an `import augment` directive to denote a regular library, it must be an augmentation library. This change is also based on #3646.

Finally, this PR makes it an error for a `<uri>` in a regular `import` directive (not `import augment`) to denote an entity which is not a library (including: it's an error if it is an augmentation library).

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -1038,6 +1038,12 @@ language and our tools.
 
 ## Changelog
 
+## 1.17
+
+*   Introduce compile-time errors about wrong structures in the graph of
+    libraries and augmentation libraries formed by directives like `import`
+    and `import augment`.
+
 ## 1.16
 
 *   Update grammar rules and add support for augmented type declarations of
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -145,9 +145,16 @@ It is a compile-time error if:
 *   A library augmentation contains a normal `library` directive. They are not
     self-contained libraries, only pieces of the augmented library.
 
-*   There is a cycle in the graph whose edges are the `import augment` 
-    directives of an augmented library and any augmentation library which is
-    directly or indirectly reachable from there via said edges.
+*   An `import augment` directive has a `<uri>` that denotes an entity which
+    is not an augmentation library. *For example, it can not be a library.*
+    
+*   An `import` (not `import augment`) directive has a `<uri>` that denotes
+    an entity which is not a library. *For example, it cannot be an 
+    augmentation library.*
+
+*   There is a cycle in the graph whose edges are the `import augment`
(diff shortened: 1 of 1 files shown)


working/augmentation-libraries/feature-specification.md
```diff
@@ -145,6 +145,17 @@ It is a compile-time error if:
 *   A library augmentation contains a normal `library` directive. They are not
     self-contained libraries, only pieces of the augmented library.
 
+*   An `import augment` directive has a `<uri>` that denotes an entity which
+    is not an augmentation library. *For example, it can not be a library.*
```
munificent, line 149: We recently changed the terminology from "augmentation library" to "library augmentation" to make it clearer that augmentations are not libraries, so change that here and throughout.

working/augmentation-libraries/feature-specification.md
```diff
@@ -145,6 +145,17 @@ It is a compile-time error if:
...
+    
+*   An `import` (not `import augment`) directive has a `<uri>` that denotes
+    an entity which is not a library. *For example, it cannot be an 
+    augmentation library.*
```
munificent, line 153: Can we make this simpler and more explicit by just saying:

> An `import` refers to an entity which is a library augmentation or part file and not a library.

Also, we probably want to say "`export` or `import`".

working/augmentation-libraries/feature-specification.md
```diff
@@ -1027,6 +1038,12 @@ language and our tools.
 
 ## Changelog
 
+## 1.17
+
+*   Introduce compile-time errors about wrong structures in the graph of
+    libraries and augmentation libraries formed by directives like `import`
+    and `import augment`.
```
munificent, line 1045: If there are issue numbers that this addresses, can you mention them here?

munificent (approved):
A few minor things but this looks great. Thank you!

working/augmentation-libraries/feature-specification.md
eernstg: That's great, 'library augmentation' seems to nudge the brain into exactly the right perspective! Done.

working/augmentation-libraries/feature-specification.md
eernstg: Done.

working/augmentation-libraries/feature-specification.md
eernstg: Done.

eernstg:
Thanks! I made some adjustments.

jakemac53 (approved):
🚀

PR #3686, opened 2024-04-02
Extension augment can't specify `on` and other clauses

sgrekhov:
It is a compile-time error for an extension augment to specify `on` clause. Extension also cannot specify `extends`, `implements` or `with` clauses, so it's unecessery in this list


eernstg (approved):
LGTM.

@munificent, I created https://github.com/dart-lang/language/issues/3687 which is a proposal for a few changes in the treatment of `extension` declarations and their augmentations, but I do not think there's any overlap or conflict with this one.

PR #3688, opened 2024-04-02
Add an `enumEntry` grammar rule

eernstg:
The feature specification includes examples where an `enum` value has the form `augment <identifier>`, but the grammar doesn't allow for the keyword `augment` in that position. This PR adds the missing grammar rule.


jakemac53 (approved):
LGTM, although we might want to drop this capability entirely. I am a bit torn on which to do.

eernstg:
Thanks! I'll land this and add the `enumEntry` rule for now, for consistency in the feature spec, and then we can remove whatever we want later on, if needed.

PR #3698, opened 2024-04-08
Adjust terminology

eernstg:
Version 1.15 performed a rename such that the augmentations feature specification used 'library augmentation' rather than 'augmentation library'. This PR updates a couple of locations where the old terminology was used such that they also use the new terminology. The PR also fixes a couple of typos.


eernstg:
Thanks!

PR #3699, opened 2024-04-09
remove dependency override on the macros package, regular constraints work now

jakemac53:


kevmoo (approved):
works great for me locally!

Does the SDK min version need to be bumped?

jakemac53:
> Does the SDK min version need to be bumped?

It doesn't, we don't have a good number we can put. There is an effective min which is transitive.

Just to help with clarity we _could_ bump it though, to 3.5.0-0 or something.

Issue #3687, opened 2024-04-02
Augmentation of extensions seems to be difficult


eernstg:
Thanks to @sgrekhov for bringing up this topic. Consider the following example:

```dart
// --- Library 'a.dart'.
import augment 'b.dart';

extension on int {...}
extension on String {...}

// --- Augmentation library 'b.dart'.
augment library 'a.dart';

augment extension {...}
```

First, we have

> It is a compile-time error if: ...
> - An augmenting extension declares an `on` clause.

so we can't have `augment extension on T ...` for any `T`. However, it's a syntax error if we _don't_ have an `on` clause. So, apparently, we just can't have an `augment` of an `extension` at all.

Next, if the extension is anonymous then it is considered to have a fresh, private name. This means that an anonymous extension augmentation will (presumably?) also have a fresh private name, which again implies that it never matches any other declaration (including any anonymous extensions or other extension augmentations).

In response to these considerations, I'd recommend that we make it an error to have an anonymous `augment extension` declaration. Moreover, we should change the grammar to allow an `augment extension` declaration to omit the `on` clause.

(We could also allow the `augment extension` declaration to have an `on` clause which is identical (in some sense) to the one which is given in the corresponding `extension` declaration, but that seems inconsistent with the treatment of other clauses in `augment` declarations.)

@dart-lang/language-team, WDYT?

lrhn:
The grammar of `augment extension` should probably be tweaked to not have the `on` clause:
```ebnf
<extensionDeclaration> ::= 
   ( 'augment' 'extension' <typeIdentifierNotType> 
  | 'extension' <typeIdentifierNotType>? 'on' <type>)
     '{' (<metadata> <classMemberDeclaration>)* '}'
```
Which means only augmenting named extensions and without specifying an `on` clause.

jakemac53:
> In response to these considerations, I'd recommend that we make it an error to have an anonymous `augment extension` declaration.

SGTM, even with `on` we couldn't reasonably match up extension augmentations without an explicit name, there could be multiple anonymous ones in the library with the same on type.

> The grammar of augment extension should probably be tweaked to not have the on clause:
> Which means only augmenting named extensions and without specifying an on clause.

The suggested grammar changes LGTM

eernstg:
The difficulties raised in this issue have all been addressed in the following PR: https://github.com/dart-lang/language/pull/3701.

[...]

PR #3701, opened 2024-04-10
Adjust the extensionDeclaration syntax

eernstg:
In response to https://github.com/dart-lang/language/issues/3687, this PR changes the grammar rule `<extensionDeclaration>` such that an augmenting extension declaration does not have an `on` clause syntactically. The compile-time error caused by having such an `on` clause is removed (it is now a syntax error). The text about extensions is adjusted slightly elsewhere to fit these changes.

This resolves a contradiction: An `augment extension` declaration used to be a syntax error if it did not have an `on` clause, but an `augment extension` declaration _with_ an `on` clause was a compile-time error.

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -1,7 +1,7 @@
 # Augmentations
 
 Author: rnystrom@google.com, jakemac@google.com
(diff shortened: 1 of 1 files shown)


jakemac53 (approved):

eernstg:
@sgrekhov, this change should enable some augmenting extension declarations that we've otherwise postponed.

Issue #3702, opened 2024-04-12
Some questions about augmentation errors


eernstg:
Thanks to @sgrekhov for bringing up this issue!

The library augmentation feature specification does not seem to specify that there must be a compile-time error in situations like the following:

```dart
// Library 'main.dart'.
import augment 'augment.dart';

class C {
  void f(); // Allowed when implemented in augmentation.
}

// Library augmentation 'augment.dart'.
augment library 'main.dart';

augment class C {
  augment void f() {
    augmented(); // Compile-time error (not specified).
  }
}
```

This situation is similar to a superinvocation of a method where the superclass does not have an implementation of said method. The situation is fully known at compile-time so we should get a compile-time error (not an exception at run time).

We do have a suitable compile-time error in the case where a variable/getter/setter is augmented, [here](https://github.com/dart-lang/language/blob/main/working/augmentation-libraries/feature-specification.md#augmenting-variables-getters-and-setters), but the error should presumably be specified (using essentially the same words) in the section about augmenting functions.

Conversely, the section about augmenting variables/getters/setters has the following error:

> An `abstract` or `external` variable are augmented by a variable.

I'd say that the `abstract` case could be allowed: It should arguably be treated like an abstract getter declaration if `final`, and otherwise like an abstract getter and an abstract setter declaration, and they should certainly allow for an augmentation that adds a regular (`get` / `set`) implementation.

But why wouldn't we allow a variable declaration to be considered as a getter declaration with an implementation (plus a setter declaration with an implementation, if non-`final`)?

You could argue that this would bring up some less straightforward situations:

```dart
// Library 'main.dart'.
import augment 'augment.dart';

class C {
  abstract final int v1;
  abstract int v2;
  abstract final int v3;
}

// Library augmentation 'augment.dart'.
augment library 'main.dart';

class C {
  augment final int v1; // OK.
  augment final int v2; // OK, augments the getter.
  // Some other augmentation can augment the setter, should be OK.
  augment int v3; // Augments the getter, OK, but adds the setter (not OK?)
}
```

I'd tend to say that the straightforward cases should be allowed, and the questionable ones (like `augment int v3`) should be errors. But this is still more permissive than the current specification.

@jakemac53, @munificent, WDYT?

lrhn:
About:
> An `abstract` or `external` variable are augmented by a variable.

We really need "declaration" after all those "variable"s.

[...]

PR #3703, opened 2024-04-12
Add compile-time errors about `augmented`

eernstg:
This PR adds a compile-time error to the feature specification about augmentations. See https://github.com/dart-lang/language/issues/3702 for some background.

The newly specified error occurs when a function augmentation body contains `augmented(...)`, and the augmented declaration has no implementation, for example:

```dart
// Library 'main.dart'.
import augment 'augment.dart';

class C {
  void f(); // Allowed when implemented in an augmentation.
}

// Library augmentation 'augment.dart'.
augment library 'main.dart';

augment class C {
  augment void f() {
    augmented(); // Compile-time error (not today, but if this PR is landed).
  }
}
```

I do not think there is any other reasonable response than reporting a compile-time error:

Another option could be to allow this at compile time and throw at run time (just like some superinvocations involving mixins), but that would contradict the trend in several years of evolution of Dart.

A third option would be to allow the invocation, and make it do nothing. However, this is not compatible with the fact that `augmented` could have a return type which is non-trivial (such that we can't "invent" an object with the required type).

Update: I added some extra rules about `augmented`: It should be an error to declare anything (a parameter, a local variable, a local function, etc.) named `augmented` in a scope where the meaning of `augmented` is that it refers to the augmented entity in the context of an augmentation. I also mentioned explicitly that there is nothing special about the word `augmented` for occurrences that are not inside an augmenting declaration.

The approach I've used is to say that `augmented` has the special augmentation-related meaning whenever it occurs inside an augmenting declaration. For instance, `augmented()` is an invocation of the augmented method in an augmenting method declaration (and it is an error inside a non-augmenting method declaration inside an augmenting class declaration). Also no declaration in the scope of an augmenting declaration can have the name `augmented`, not even inside a non-augmenting declaration which is again inside an augmenting declaration.

But for an occurrence of `augmented` which isn't inside any augmenting (directly or indirectly enclosing) declarations, the static analysis and dynamic semantics is standard: This is just an identifier, and it may or may not refer to a declaration which is in scope, and there's nothing special about the treatment of a reference to such a declaration when it has the name `augmented` vs. when it has any other name.

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -1,7 +1,7 @@
 # Augmentations
 
 Author: rnystrom@google.com, jakemac@google.com
-Version: 1.20 (see [Changelog](#Changelog) at end)
+Version: 1.21 (see [Changelog](#Changelog) at end)
 
 Augmentations allow spreading your implementation across multiple locations,
 both within a single file and across multiple files. They can add new top-level
@@ -458,6 +458,11 @@ It is a compile-time error if:
 
 *   The function augmentation specifies any default values. *Default values are
     defined solely by the original function.*
+
+*   An augmenting declaration uses `augmented` when the original declaration has
+    no concrete implementation. Note that all external declarations are assumed
+    to have an implementation provided by another external source, and they will
+    throw a runtime exception when called if not.
 
 **TODO: Should we allow augmenting functions to add parameters? If so, how does
 this interact with type checking calls to the function?**
@@ -1041,6 +1046,11 @@ recommend the few users using them migrate to library augmentations. In Dart
 simplify the language and our tools.
 
 ## Changelog
+
+## 1.21
+
+*   Add a compile-time error for `augmented(...)` whose target is abstract.
+
 
 ## 1.20
 
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -379,6 +379,16 @@ occurs in a location where the outermost enclosing declaration is
 augmenting. *This error is applicable to all such declarations, e.g.,
 local functions, local variables, parameters, and type parameters.*
 
+A compile-time error occurs if `augmented` is used as an identifier
+expression in a non-augmenting declaration of a kind that can be augmenting
+inside an augmenting declaration.
+
+*For example, inside `augment class C` we could have a declaration like
+`void f() {...augmented()...}`. 
+This is an error because the outer `augment` forces the meaning of `augmented`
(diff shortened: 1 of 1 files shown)


working/augmentation-libraries/feature-specification.md
```diff
@@ -356,6 +356,39 @@ augmented, but it follows generally the same rules as any normal identifier:
...
+keyword because it is unable to refer to a declaration in scope when it
+occurs inside an augmenting declaration, it always has the special meaning
+which is associated with augmentations.*
+
+A compile-time error occurs if a declaration with the name `augmented`
+occurs in a location where the outermost enclosing declaration is
+augmenting. *This error is applicable to all such declarations, e.g.,
+local functions, local variables, parameters, and type parameters.*
+
+A compile-time error occurs if `augmented` occurs in a non-augmenting
```
jakemac53, line 382: This one doesn't seem absolutely necessary to me - it only really needs to be an error if it is inside an augmenting declaration which introduces a meaning for `augmented` - in other words using `augmented()` inside an `augment class C` declaration is safe to do.

But, I can see that it might be hard to specify (and confusing to understand). So I am ultimately OK with specifying it this way as a simplification, cc @munificent thoughts?

sgrekhov:
```dart
abstract interface class I {
  int bar();
}

class C implements I {}

augment class C {
  int bar() {
    return augmented(); // Not safe
  }
}
```

jakemac53: (there is no augmented function, since this isn't an augmenting method, so it is just an identifier called `augmented`)

sgrekhov: > Assuming that there is a declaration augmented in the surrounding scope, why can that not just refer to it?
I didn't think about it. Right

jakemac53 (approved):
LGTM

working/augmentation-libraries/feature-specification.md
jakemac53: We talked briefly in person and the intention here is just to simplify the rules so they are easier to understand, which I think is well motivated 👍

eernstg: I suggested that we should make `augmented` refer to the augmentation mechanism rather than any declared name in scope _everywhere_ in an `augment` declaration because I thought that would be the most readable approach.

I do understand that it would be possible for a non-augment declaration inside an augment declaration to, say, call a top level function named `augmented`, but I thought it would be really tricky for a human reader of the code to understand that this is what we're doing. So, pure readability.

Issue #2509, opened 2022-09-22
Named parameters whose name starts with `_`


eernstg:
A declaration of a name that starts with `_` is private to the current library, unless the declaration is

- a local variable (where privacy makes no difference),
- a label (where privacy also makes no difference, and which is in its own namespace anyway), or
- a formal parameter (where the meaning of privacy has never been fully resolved)

Concretely, the tools are currently dealing with required named formal parameters whose name starts with `_` in slightly different ways:

```dart
// ----- Library 'n011lib.dart'.
class A {
  A({_name});
  void m({_name}) {}
}

class B {
  B({required _name});
  void m({required _name}) {}
}

// ----- Library 'n011.dart'.
import 'n011lib.dart';

void main() {
  var a = A(_name: 0);
  a.m(_name: 0);
  var b = B(_name: 0);
  b.m(_name: 0);
}
```

With the analyzer (commit 92638bf5d3a5668920d0a40a4b24a6a97f982b3f), no issues are reported; that is, `_name` is treated like any other name when it occurs as the name of a named parameter, required or not.

But the common front end reports an error for both cases with the _declaration_ of `A` (effectively saying "a named parameter name cannot be private"), and allows class `B` (.. "except if it's required").

We could say that a formal parameter name starting with `_` is simply a name like all others, with no special exceptions. In particular, call sites in different libraries can invoke the enclosing function/method/constructor and pass an actual argument to a named parameter with such a name.

We could also say that a formal parameter name starting with `_` is a private name. So if it's the name of a named parameter then it can't be passed at call sites in other libraries.

The former makes it easy to use named initializing formals to initialize private instance variables (and the name `_name` in the parameter declaration `this._name` serves as documentation for the fact that we're initializing a private instance variable), but it is unfortunate that this usage will make it harder to rename said instance variable. (It would be nicer if the API would only reveal that we're initializing a private instance variable, and it wouldn't mention the name.)

The latter provides support for a very special kind of privacy: Every caller outside the library _must_ use the default value for a non-required named formal parameter with that kind of name, and the whole function/method/constructor can't be called at all outside the library if the parameter is required. This feature is again potentially useful, but rather accidental in nature.

At the conceptual level it may seem inconsistent to claim that `_name` is just another name (not private to any library) when it's used as the name of a named initializing formal `this._name`, because it _does_ refer to the instance variable whose name is definitely private.

[...]

PR #3729, opened 2024-04-26
Add rules about super and extension types to wildcard spec

eernstg:
This PR adds rules to the wildcards feature specification about super parameters of the form `super._`, and about representation declarations in extension types whose representation variable name is `_`.

In particular, `super._` is an error because super parameters that can't be referenced from the initializer list are inherently broken. In contrast, `extension type E(int _) {}` is allowed: The representation variable name will be `_` and the formal parameter will be inaccessible (which makes no difference at this time, but will matter if a more general primary constructors feature is introduced).

### DIFF
--- a/working/wildcards/feature-specification.md
+++ b/working/wildcards/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: In-progress
 
-Version 1.0
+Version 1.1
 
 Pattern matching brings a new way to declare variables. Inside patterns, any
 variable whose name is `_` is considered a "wildcard". It behaves like a
@@ -99,8 +99,8 @@ A *local declaration* is any of:
 *   For loop variable declarations.
 
     ```dart
-    for (_ = 0;;) {}
-    for (_ in list) {}
+    for (int _ = 0;;) {}
+    for (var _ in list) {}
     ```
 
 *   Catch clause parameters.
@@ -117,11 +117,12 @@ A *local declaration* is any of:
 
     ```dart
     class T<_> {}
+    void genericFunction<_>() {}
 
     takeGenericCallback(<_>() => true);
     ```
 
-A local declaration whose name is `_` does not bind anything to that name. This
+A local declaration whose name is `_` does not bind that name to anything. This
 means you can have multiple local declarations named `_` in the same namespace
 without a collision error. The initializer, if there is one, is still executed,
 but the value is not accessible.
@@ -138,12 +139,12 @@ class C {
   var _ = 'bound';
 
   test() {
-    print(_); // Prints "bounnd".
+    print(_); // Prints "bound".
   }
 }
 ```
 
-Likewise with a top-level named `_`:
+Likewise with a top-level declaration named `_`:
 
 ```dart
 var _ = 'ok';
@@ -207,8 +208,8 @@ quite confusing. In practice, we expect reasonable users will not name fields
 
 ### Initializing formals
 
-An initializing formal named `_` does still initialize a field named `_` (and
-you can still have a field with that name):
+A positional initializing formal named `_` does still initialize a field 
+named `_` (and you can still have a field with that name):
 
 ```dart
 class C {
@@ -218,8 +219,18 @@ class C {
 }
 ```
 
+It is a compile-time error if a named initializing formal has the name `_`:
+
+```dart
+class C {
+  var _;
+  
+  C({this._}); // Error.
+}
+```
+
 But no *parameter* with that name is bound, which means `_` can't be accessed
-inside the initializer list. In the body is fine, since that refers to the
+inside the initializer list. In the body it is fine, since that refers to the
 field, not the parameter:
 
 ```dart
@@ -228,7 +239,7 @@ class C {
   var other;
 
   C(this._)
-    : other = _ { // <-- Error. No "_" in scope.
+    : other = _ { // <-- Error, cannot access `this`.
     print(_); // OK. Prints the field.
   }
 }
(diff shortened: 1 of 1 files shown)


working/wildcards/feature-specification.md
```diff
@@ -99,8 +99,8 @@ A *local declaration* is any of:
 *   For loop variable declarations.
 
     ```dart
-    for (_ = 0;;) {}
-    for (_ in list) {}
+    for (int _ = 0;;) {}
+    for (var _ in list) {}
```
eernstg, line 103: I changed this because we want to see the declaration named `_`. With the previous version the name `_` could resolve to a top-level variable and several other things.

working/wildcards/feature-specification.md
```diff
@@ -117,11 +117,12 @@ A *local declaration* is any of:
 
     ```dart
     class T<_> {}
+    void genericFunction<_>() {}
 
     takeGenericCallback(<_>() => true);
     ```
 
-A local declaration whose name is `_` does not bind anything to that name. This
+A local declaration whose name is `_` does not bind that name to anything. This
```
eernstg, line 125: We're otherwise "binding a name to a value", not "binding a value to a name", so we should probably that direction here as well. (Mnemo: In `'name' -> 42` considered as a notation for binding 'name' to the value 42, the arrow is left-to-right because we will start by finding 'name', and then we'll use the arrow to find 42. The alternate notation `'name' <- 42` seems to allow multiple values to be bound to the same name, which is not intended.)

working/wildcards/feature-specification.md
```diff
@@ -228,7 +239,7 @@ class C {
   var other;
 
   C(this._)
-    : other = _ { // <-- Error. No "_" in scope.
+    : other = _ { // <-- Error, cannot access `this`.
```
eernstg, line 242: `_` is actually in scope because it is an instance member, but it is not allowed for an initializing expression in this context to access `this`. Hence the updated comment.

eernstg (commented):
Added a few explanatory remarks.

working/wildcards/feature-specification.md
munificent: Ah, that mnemonic is very helpful, thank you! I've struggled literally for years with trying to understand whether I should say "bind a name to a value" or "bind a value to a name". Your point that a given name can only have a single value is the key insight I needed.

working/wildcards/feature-specification.md
```diff
@@ -218,8 +219,18 @@ class C {
...
 
+It is a compile-time error if a named initializing formal has the name `_`:
+
+```dart
+class C {
+  var _;
+  
+  C({this._}); // Error.
+}
+```
```
munificent, line 230: This is not a new error, correct? This is just the existing error for private named parameters? If so, it might be good to clarify that here.

working/wildcards/feature-specification.md
```diff
@@ -244,6 +255,39 @@ class C {
...
+An occurrence of `super._` as a declaration of a formal parameter in a
+constructor is a compile-time error.
+
+*It is not an error everywhere: In a method body it could be an invocation
+of an inherited getter named `_`.*
+
+*The desugared meaning of a super parameter includes a reference to the
+parameter in the initializer list of the enclosing constructor declaration,
+but such references are not possible when the parameter name is a
+wildcard.*
```
munificent, line 269: This is a funny subtle point, but made sense once I thought about it some more.

munificent (approved):
This looks great, thank you! And thank you for the explanatory commentary too, it was very helpful.

working/wildcards/feature-specification.md
eernstg: Great, thanks!

working/wildcards/feature-specification.md
kallentu: Yeah, to confirm, this is just https://dart.dev/tools/diagnostic-messages#private_optional_parameter right?

working/wildcards/feature-specification.md
```diff
@@ -218,8 +219,18 @@ class C {
...
+class C {
+  var _;
+  
+  C({this._}); // Error.
+}
+```
+
 But no *parameter* with that name is bound, which means `_` can't be accessed
-inside the initializer list. In the body is fine, since that refers to the
+inside the initializer list. In the body it is fine, since that refers to the
```
kallentu, line 233: Can we rewrite "In the body it is fine" to something like "Using `_` in the body is fine," or something similar. I can't really tell what "it" is and reading that phrase is somewhat difficult.

working/wildcards/feature-specification.md
```diff
@@ -244,6 +255,39 @@ class C {
 }
 ```
 
+### Super parameters
+
+An occurrence of `super._` as a declaration of a formal parameter in a
+constructor is a compile-time error.
```
kallentu, line 261: Can you also help me visualize this, and maybe the following below, by adding a little code snippet/example?
The words make sense, I just learn better with examples.

working/wildcards/feature-specification.md
```diff
@@ -244,6 +255,39 @@ class C {
...
+
+*It always declares exactly one mandatory positional parameter, and the
+meaning of this declaration is that it introduces a formal parameter of a
+constructor of the enclosing extension type as well as a final instance
+variable declaration, also known as the representation variable of the
+extension type.*
+
+This parameter can have the declared name `_`. This means that the
+representation variable is named `_`, and no formal parameter name is
+introduced into any scopes.
```
kallentu, line 284: It would also be nice to have an example listed here, for visualization. :D

working/wildcards/feature-specification.md
```diff
@@ -244,6 +255,39 @@ class C {
...
+
+*Currently that parameter is not in scope for any code anyway, but future
```
kallentu, line 286: If I'm understanding correctly, if we have primary constructors, the formal parameter `_` will be inaccessible anyways. Could we write that more explicitly?

kallentu (approved):
This is great! Thank you so much for updating the spec, Erik!

working/wildcards/feature-specification.md
eernstg: Right, good catch! I thought it was a new error because it can be motivated in terms of wildcard specific properties: A named parameter must have a name, but wildcard parameters do not.

However, we have [this](https://github.com/dart-lang/language/blob/bb23c2b8e12ddb54d485d53c8690d2d698c3cab1/specification/dartLangSpec.tex#L2412) in the language specification already:

> It is a compile-time error if the name of a named optional parameter begins with an `_' character.

(It is updated to say 'if the name of a named parameter' with the [null-safety related updates](https://github.com/dart-lang/language/pull/2605), because it applies to `required` named parameters as well as optional ones.)

If we adopt https://github.com/dart-lang/language/issues/2509 then formal parameters of the form `this._...` and `super._...` will presumably be allowed as long as there is a corresponding public name. But this still makes the bare `_` an error (a public name can't be empty).

I changed it to commentary.

working/wildcards/feature-specification.md
eernstg: Good point, done!

working/wildcards/feature-specification.md
eernstg: Done. This brings out a slight inconvenience, by the way. I added a commentary paragraph about it.

working/wildcards/feature-specification.md
eernstg: Done.

working/wildcards/feature-specification.md
eernstg: I gave an example where it does matter, but it involves both primary constructors and the hypothetical extension which allows superconstructor invocations to be specified in the header. So it's a pretty remote issue, and it might work better to just omit it.

Anyway, it's there now. Take a look! ;-)

kallentu:
> @kallentu, we might want to delete the 20 lines just before 'Unused variable warnings', PTAL.

I would be okay with that. It might be more confusing with that example haha, but thank you for writing it up.

working/wildcards/feature-specification.md
eernstg: OK, I deleted those 20 lines.

eernstg:
OK, done!

PR #3779, opened 2024-05-03
Add a proposal about inference of the modifier `required`

eernstg:
This PR adds a feature specification of the feature whereby `required` is added implicitly to parameters of concrete function and method declarations when it would be an error to not have that modifier.

### DIFF
--- a/working/0015-infer-required/feature-specification.md
+++ b/working/0015-infer-required/feature-specification.md
@@ -6,6 +6,8 @@ Status: Draft
 
 Version 1.0 (see the [CHANGELOG](#CHANGELOG))
 
+Experimental flag: `--enable-experiment=infer-required`
+
 ## Summary
 
 This proposal is built on a large number of issues expressing the desire to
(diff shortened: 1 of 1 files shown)


eernstg:
@dart-lang/language-team, WDYT? Here is a proposal for inference of requiredness. It is just a consolidation of a rather large number of discussions and proposals in this area. The specification itself is just a handful of lines. ;-)

jakemac53 (approved):

leafpetersen (approved):

kallentu (approved):

working/0015-infer-required/feature-specification.md
```diff
@@ -0,0 +1,213 @@
...
+  String address,
+  String phoneNumber,
+  String name,
+  String? avatarUrl,
+});
+```
+
+It seems likely that it will be true quite often that the required, named
+parameters are exactly the ones whose type is potentially non-nullable,
+which means that the migration will simply be to delete every occurrence of
```
natebosch, line 140: If we have an automated migration I think we should make give it full fidelity - a `required` which would not be inferred shouldn't be removed.
I don't know of specific examples, but I think it's reasonable for existing code to use `required` to force callers to explicitly pass `null` when that is the intention.

natebosch (approved):

working/0015-infer-required/feature-specification.md
eernstg: Right, a mechanical migration should certainly be semantics preserving. Added a sentence to clarify this point.

eernstg:
Thanks! I'll land this document now, and we would then use additional PRs in order to make changes as needed.

PR #3785, opened 2024-05-07
[wildcard-variables] Add 'no_wildcard_variable_uses' lint to breaking change section.

kallentu:
This PR adds information about the existing [`no_wildcard_variable_uses`](https://dart.dev/tools/linter-rules/no_wildcard_variable_uses) lint which should help the migration of the wildcards feature. 

It's already a core lint and should be enabled in most projects, meaning that the breakage for the wildcards feature should be small.

Anyways, writing is hard -- so let me know if I should add/remove/change something. Thanks 😄

### DIFF
--- a/working/wildcards/feature-specification.md
+++ b/working/wildcards/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: In-progress
 
-Version 1.1
+Version 1.2
 
 Pattern matching brings a new way to declare variables. Inside patterns, any
 variable whose name is `_` is considered a "wildcard". It behaves like a
@@ -395,7 +395,17 @@ However, this *is* a breaking change. If this ships in the same version as
 pattern matching, we can gate it behind a language version and only break code
 when it upgrades to that version.
 
+### Existing Lints
+
+We have an existing [`no_wildcard_variable_uses`](https://dart.dev/tools/linter-rules/no_wildcard_variable_uses) lint, which advises users to avoid using wildcard parameters or variables.
+
+This lint is included in the core lint set which means that the scale of the breaking change should be small since most projects should have this lint enabled.
+
 ## Changelog
+
+### 1.2
+
+- Add information about the [`no_wildcard_variable_uses`](https://dart.dev/tools/linter-rules/no_wildcard_variable_uses) lint.
 
 ### 1.1
 


pq (approved):
Bob or Eric may suggest some word-smithing but content LGTM!

working/wildcards/feature-specification.md
```diff
@@ -395,8 +395,18 @@ However, this *is* a breaking change. If this ships in the same version as
 pattern matching, we can gate it behind a language version and only break code
 when it upgrades to that version.
 
+### Existing Lints
+
+We have an existing [`no_wildcard_variable_uses`](https://dart.dev/tools/linter-rules/no_wildcard_variable_uses) lint, which advises users to avoid using wildcard parameters or variables.
+
+This lint is included in the core lint set which means that the scale of the breaking change should be small since most projects should have this lint enabled.
```
eernstg, line 402: I guess identifiers like `___` are used exclusively as homegrown wildcards today. Otherwise:

This lint will flag usages of names of the form `__`, `___` and so on, in addition to `_`. Does it give rise to unnecessary churn to flag all the names that have more than one underscore, given that they will keep their status as regular (non-wildcard) identifiers? For example, if someone really wants to use local variables with the name `_` then they might switch to `__` just because it's similar, and not overly long, and they might be unhappy if they still get the lint message.

At least, the lint could make the distinction, or we might want a separate lint that only targets the cases where `_` resolves to a declaration which would be wildcarded when that feature is enabled.

kallentu: Users would be avoiding all types of `_` which may be unnecessary churn for the names with more than one underscore, but if they're already using it and the lint is out there in the core set of lints.

I'm not sure if there's enough value to go back and change the lint now that it's already over-reporting. It's also a set of migrations that regardless, the code author has to go in and change once they learn about the wildcard feature.

And it would be good to have a separate lint that tells users that `_` will be wildcarded once the feature is enabled, I'm just not sure how that would interact with the `no_wildcard_variable_uses` lint. Would these two overlap if they were both in the core lint set?

@pq Maybe you'd have a better sense of these two lints and their use cases.

pq: > Does it give rise to unnecessary churn to flag all the names that have more than one underscore, given that they will keep their status as regular (non-wildcard) identifiers?

Great question. I'm really not sure! I'd be curious to know how often this happens in practice... I'm guessing not terrifically often but it'd be interesting to do a little grepping.

> Maybe you'd have a better sense of these two lints and their use cases.

I'd have to think about this a bit more... How about we open an issue on the linter repo and take the conversation there? I think this is safe to land as is but wouldn't want @eernstg's considerations to just get lost.

kallentu: Filed https://github.com/dart-lang/linter/issues/4968.
Let's continue the conversation there so we don't lose it.
I'll submit this.

munificent (commented):
👍 👍 👍

PR #3800, opened 2024-05-14
Split augmentations into improved part files and augmentation declara…

lrhn:
…tions.

Unifies "augmentation libraries" and part files.
Part files can have imports, exports and further part files. Part files can use configurable URIs.
A `part of` directive can no longer use a library name. Part files inherit the imports of their parent file, and can extend or shadow those with their own imports.

Augmentation declarations can occur in any part or library file. An augmentation must occur "below" the declaration it augments: Either later in the same file ("below" when viewing code), or in the file-tree of a part file of the same file ("below" in the part-file tree of the library).

This ensures that the original declaration occurs above all augmentations of it, and that all augmentations of the same original declaration occur on a single path down the part-file tree of the library. What again ensures that reordering `part` directives does not change the order of augmentations.

Changed the lexical scope of augmenting class-like declarations (declarations with a member scope) to only contain the members declared inside the same class-like declaration, not the collection of all members declared by all declarations with the same name.

Rewrote the part about applying augmentations. This is more speculative since it doesn't provide a complete definition, but more of a pattern for extending semantics that assume a name refers to a single declaration, into one where a name denotes a set (stack depending on augmentation application order), of individual syntactic declarations.
The existing semi-syntactic "merging" may not be a viable specification approach, since it requires merging code from different scopes into a single scope. That's not impossible, we also move code around for mixin applications, but it's also easy to get wrong.

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -1,6 +1,6 @@
 # Augmentations
 
-Author: rnystrom@google.com, jakemac@google.com
+Author: rnystrom@google.com, jakemac@google.com, lrn@google.com <br>
 Version: 1.21 (see [Changelog](#Changelog) at end)
 
 Augmentations allow spreading your implementation across multiple locations,
@@ -64,38 +64,48 @@ part of macro expansion, it's still useful to have it be in a canonical
 well-specified form that users can understand.
 
 This proposal defines that format. The idea is that a Dart compiler executes
-macros and then produces one or more library augmentation files that contain all
-of the changes that the macros made to the library where they are applied. The
-language then automatically merges those library augmentations into the
-augmented library.
+macros and then produces one or more new part files that contain all
+of the changes that the macros made to the library where they are applied, as
+new declarations to be added or augmentations that modify existing
+declarations. The
+compiler then adds those part files to the existing libraries.
 
-But library augmentations aren't *only* a serialization format for macros. They
-are a first-class language feature that can be produced by non-macro code
-generators or written by hand by users who simply want to break a giant library
-or class into smaller files.
+But improved part files and augmenting declarations are not *only* a
+serialization format for macros. They are first-class language features that
+can be produced by non-macro code generators or written by hand by users who
+simply want to break a giant library or class into smaller files.
 
-## Library Augmentations
+## Part files with imports
 
-A library augmentation is a separate file that *augments* an existing library,
-referred to as the *augmented* library. A library augmentation is similar to
-but more powerful than a part file, and it is not a library in its own right.
+As part of the meta-programming and augmentation features, we expand the
+capabilities of part files. See [“Parts with Imports”][parts_with_imports.md].
 
-*   Like a library, it may contain all kinds of declarations&mdash;functions,
-    classes, variables, etc.
+With that feature, a part file can now have its own `import` and `export`
+directives, and further nested `part` files, with part files inheriting the
+imports and prefixes of their parent (part or library) file.
 
-*   Like a library, it has its own import scope and may contain its own imports.
+Augmentation declarations interact with part files mainly in restrictions on
+where an augmenting declaration may occur relative to the declaration it
+augments, as describe below.
 
-*   Like a part file, all of the top-level declarations it produces end up in
-    the top-level scope of the augmented library.
+[parts_with_imports.md]: parts_with_imports.md "Parts with Imports Feature Specification"
 
-*   Like a part file, it shares a private scope with the augmented library and
-    the two have mutual access to private declarations in the other file.
+## Augmentation declarations
 
-Library augmentations may also contain declaration augmentations, which augment
-existing declarations from the library. Some examples include:
+Augmentation declarations are declarations marked with the new built-in
+identifier `augment`, which makes the declaration augment an existing
+declaration, which is itself a normal declaration with zero or more prior
+augmentations applied. An augmentation declaration does not  introduce a new
+name into the surrounding scope, it attaches itself to the existing name.
+
+Making `augment`  a built-in identifier is language versioned, to make it
+non-breaking for pre-feature code.
+
+Augmentation declarations include:
 
 *   Type augmentations, which can add new members to types, including adding new
-    values to enums, or even alter the type hierarchy by adding mixins, etc.
+    values to enums, or even alter the type hierarchy by adding mixin
+    applications to a class.
 
 *   Function augmentations, which can replace the body of a function, or provide
     a body if none was present.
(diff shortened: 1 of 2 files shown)


davidmorgan:
Whoops, I thought this was an issue; it's a PR. Opened #3848 so we have an issue for this.

working/augmentation-libraries/parts_with_imports.md
```diff
@@ -0,0 +1,367 @@
...
+    *   _It’s possible to look further up in the import chain *C* for a prefix
+        scope to extend. Here it’s chosen that that importing parent file gets
+        to decide which names the part file has access to. If it wants to make
+        a transitive parent import prefix available, it should just not shadow
+        it._
+
+That is: The combined import scope of a Dart file is a chain of the combined
+import scopes of the file and its parent files, each step adding two scopes:
+The (unnamed, top-level) import scope of the unprefixed imports and the prefix
+scope with prefixed imports, each shadowing names further up in the chain.
```
jakemac53, line 258: I am not sure about this "shadowing names further up the chain" part, presumably this is what you meant earlier when talking about being able to ignore/shadow parent imports.

Basically, your current import scope would implicitly shadow your inherited import scope? So a Foo which is directly imported, would not conflict with a Foo coming from parent imports, it would shadow it?

I don't personally like that, I would rather it just be a conflict, and have only one import scope.

working/augmentation-libraries/feature-specification.md
```diff
@@ -64,259 +64,130 @@ part of macro expansion, it's still useful to have it be in a canonical
...
+declaration. The augmentation application order is defined rather simply,
+because of a further requirement on where augmentations of the same declaration
+can occur relative to each other.
+
+It’s a **compile-time error** if a library contains an augmentation
+declaration, but no non-augmentation declaration with the corresponding name in
+the same scope. _(A mutable variable declaration counts as having both a getter
+and a setter name.)_
+
+For the following, we’ll say that one declaration of a library is *above* another declaration of the same library if and only if:
```
jakemac53, line 154: nit: long line

working/augmentation-libraries/feature-specification.md
```diff
@@ -64,259 +64,130 @@ part of macro expansion, it's still useful to have it be in a canonical
...
+
+* The former declaration is in the same file as the latter declaration, and it
```
jakemac53, line 156: nit: indentation for these bullets is inconsistent

working/augmentation-libraries/feature-specification.md
```diff
@@ -64,259 +64,130 @@ part of macro expansion, it's still useful to have it be in a canonical
...
+   presented), or
+* The former declaration is in a file that is a direct or transitive parent
+  file of the file of the latter declaration (“above” in the file tree
+  hierarchy).
+
+We can similarly define *below* as the inverse of that relation. Both *before*
+and *after* define *strict partial orders* on declarations in a library.
+
+It’s a **compile-time error** if a library contains an augmentation declaration
+and a corresponding non-augmentation base declaration, and the the base
```
jakemac53, line 167:
```suggestion
and a corresponding non-augmentation base declaration, and the base
```

working/augmentation-libraries/feature-specification.md
```diff
@@ -64,259 +64,130 @@ part of macro expansion, it's still useful to have it be in a canonical
...
+declaration is not *above* the augmentation declaration.
+
+It is a **compile-time error** if a library contains two augmentation
+declaration for the same base declaration, and one augmentation is not *above*
+the other.
+
+These requirements ensure that declarations that contribution to the same
```
jakemac53, line 174:
```suggestion
These requirements ensure that declarations that contribute to the same
```

working/augmentation-libraries/feature-specification.md
```diff
@@ -401,58 +272,79 @@ augment class SomeClass {
...
-    declaration show the complete capabilities of the declaration. It also
+    declaration shows the complete capabilities of the declaration. It also
     deliberately prevents an augmentation from introducing a restriction that
     isn't visible to a reader of the main declaration.*
 
+*   The augmenting type declares an `extends` clause for a `class` declaration,
+    but one was already
+    present _(or the `class` was a `mixin class` declaration, which does not
+    allow `extends` clauses)_. We do not allow overwriting an existing
+    `extends`, but one can be filled in if if none had been specified.
```
jakemac53, line 314:
```suggestion
    `extends`, but one can be filled in if none had been specified.
```

working/augmentation-libraries/feature-specification.md
```diff
@@ -464,41 +356,46 @@ augment int slowCalculation(int a, int b) {
...
     no concrete implementation. Note that all external declarations are assumed
     to have an implementation provided by another external source, and they will
     throw a runtime exception when called if not.
 
-**TODO: Should we allow augmenting functions to add parameters? If so, how does
-this interact with type checking calls to the function?**
+**TODO: Should we allow augmenting functions to add optional parameters? If so,
+how does this interact with type checking calls to the function?**
+
+**TODO: Should we allow an augmenting function to repeat default values? Or
```
jakemac53, line 396: Also not allowed and can be removed

working/augmentation-libraries/feature-specification.md
```diff
@@ -464,41 +356,46 @@ augment int slowCalculation(int a, int b) {
...
+change default values? Invoking the `augmented` function should will supply the
```
jakemac53, line 397:
```suggestion
change default values? Invoking the `augmented` function will supply the
```

working/augmentation-libraries/parts_with_imports.md
```diff
@@ -0,0 +1,367 @@
...
+perfectly with that way of thinking.
+
+## Feature
+
+This feature allows a part file to have `import`, `export` and `part`
+directives of its own, where `import` directives only affect the part file
+itself, and its transitive part files. A library is defined by the source code
+of its library file and *all* transitively included part files, which can be an
+arbitrarily deep *tree*. A part file inherits the imports and import prefixes
+of its parent file (the library or part file that included it) into its
```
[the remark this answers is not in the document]
lrhn: It is transitive. 

The import scope of a part file *extends* the import scope of its parent file (which then transtively extends its parent's import scope).

Let's try to find an understandable way to say that.

working/augmentation-libraries/parts_with_imports.md
```diff
@@ -0,0 +1,367 @@
...
+    *   Because of that, it’s always safe to move one or more declarations into
+        a new part file. _(Although augmentation declarations modifies that
+        slightly.)_
+    *   Similarly it’s always possible and safe to combine a part file with no
+        `import`s back into its parent file.
+*   *Library member declarations are library-global*: All top-level
+    declarations in the library file and all transitive part files are equal,
+    and are all in scope in every file. They introduce declarations into the
+    library’s declaration scope, which is the most significant scope in all
+    files of the library. If there is any conflict, top-level declarations win!
```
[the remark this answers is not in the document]
lrhn: > I am actually not sure this is the correct choice.


> Consider that declarations which are not in the current file are not in the lexical scope

Thta's a crazy idea. I just might work!

They are in scope today, but today there is only one scope in a library. It's the same scope that is the top-level scope for every file of a library. 

What I changed here is to have the import scope be per-file, but then I applied the same declaration scope on top of every import scope. There is no change to that.

If we also inherit the declaration scope, and can extend on top of that, ... how would that work?

- A part file's top-level scope extends the *declaration* scope of its parent, not its import scope.
- It extends with its own import scope and prefix scope (as currently specified) and then with its own declarations *and the declarations of all its sub-parts*, but not the declarations of every file in the library.
    - A library file does need to have access to the declarations of its parts, so it needs to include every declaration in the library in the library file's declaration scope. Anything else would be too breaking.
    - A part file inherits its parent's top-level scope (imports,prefixes,all declarations in all subparts, including this one), but can override anything in there with an import, which can then be overridden by the declarations contained in the part file and all its sub-parts.
- If the part file needs to see any declaration from the parent file, which are also the declarations of its sibling parts and any non-shadowed declarations from the rest of the library, it should avoid shadowing with an import or prefix.

(Hmm. Feels familiar. I'm pretty sure I've tried to specify something like that before. Probably in one of the earlier augmentation issues, or maybe in a document I never published.)

That all seems like it could work. 
And it even has some nice properties. It would mean that any declaration that only exists in a parent or sibling part can be ignored.
A declaration in a sibling will no longer conflict with a prefix in this part.

Depending on how augmenting declarations count, a name may be declared in more than one place. 
So about "shadowing a top level member from an import with an augmentation", that would mean declaring an augmentation with a name which would otherwise resolve to an imported declaration from another library. That doesn't completely make sense if the augmentation itself introduces the name. But maybe it should only ever be the base declaration which introduces the name. 

So not sure that situation _needs_ to be an error, but it _can_ be.
* One can see an augmenting declaration as *referencing* the declaration it augments, and therefore the chosen name should *resolve* to that declaration. If it resolves to something that cannot be augmented, that's just a plain error.
* Or one can see the augmenting declaration as simply having the same *name* as the declaration it augments (that's how it's decided what it augments). We don't resolve the *declared* name of a declaration. It *introduces* a name, it doesn't reference one.

In the latter case, there is no need for the augmented declaration to be in scope. The only question is whether the augmenting declaration introduces the name or not. (But if not, that may actually suggest that the former interpretation is the right one.)

The former case, based on that thought, could mean that augmenting declarations do not introduce a name in the surrounding scope (which now matters because the declaration scope is not every declaration in the library any more), and the name of the declaration *is* a reference to the declaration it augments. You can only refer to declaration's name if you can refer to the *base declaration*, and you can only augment it if you can refer to it.

For *augmenting member declarations*, the name should be resolved relative to the type that is being augmented. A static member or constructor should be resolved against the static scope of the class/mixin/etc that it's inside (including all earlier augmentations, even inside the same scope). An instance member is resolved against the interface of the surrounding type (or whatever makes sense for extensions) also including all earlier augmentations.
So, op problem, those are not lexical scope based. If we can refer to the container declaration, we can augment its members

So, TL;DR:
* The top-level scope of the library file is the same one as today (imports, prefixes, all declarations in the library).
* The top-level scope of a part file *extends* the top-level scope of its parent file with its own imports, prefixes and all non-augmenting declarations in its sub-tree.
* Augmenting top-level declarations do not introduce a name in their surrounding scope, instead the name in the augmenting declaration *refers* to the declaration they want to augment, which must be a declaration of the current library, and must be in scope. (So if they work, they don't need to introduce a name, if they don't, they'd be referring to themselves, which makes no sense.)
* Augmenting namespaced declarations (instance, static, constructor) may or may not introduce a name in the lexical scope. It doesn't hurt, and it allows an augmenting declaration body to safely be recursive, without risking a conflict with an imported name.

SGTM. Let's ask others if they can see any issues.

working/augmentation-libraries/parts_with_imports.md
```diff
@@ -0,0 +1,367 @@
...
+*   *The unit of ownership is the library*. It’s quite possible for one part
+    file to introduce a conflict with another part file. It always was, but
+    there are new ways too. If that happens, the library owner, who most likely
+    introduced the problem, is expected to fix it. There is no attempt to hide
+    name conflicts between declarations in separate tree-branches of the
+    library structure.
+*   *Import inheritance is a only suggestion*: Aka. other files’ imports cannot
```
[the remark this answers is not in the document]
lrhn: By "ignore" I just mean not refer to it. 
If the part file needs to re-use the name for something, it can. (But it can't use the anme as an unqualified identifier and hope it resolves to `this.id`, so can't *completely* ignore it.)

So basically: A part file can always *reuse* an inherited name without any issue, so it can ignore *whether* a name is inherited when it chooses to use that name itself.

working/augmentation-libraries/parts_with_imports.md
lrhn: Yes. If a part introduces a name itself, by an import or import prefix, then it can ignore whether a parent had imported the same name.

If a part file only uses names it has imported itself, then it can completely ignore the inherited imports.

The inherited imports cannot cause a *name conflict* in the part file, any name introduced by the part file will always take precedence.

lrhn: Having only one import scope is where we are today.

Do you mean only one import scope per part file, which *collapses* the inherited scope and the new imports into one, making any name conflict an conflicting name?

Or do you mean any import in any part file being visible in the entire library, and any name conflict being a conflicted import in every file of the library?

The latter seems rather harsh. Not even import prefixes would be local, they too would likely be joined with import prefixes of other libraries, and name conflicts inside would be conflited imports.

The former still makes it much harder for a part or parent file, so any file, to introduce new imports. It would only conflict with imports of its parents, not any import anywhere in the library, but a part file that wants to import something, still has to do it with a prefix (which itself may already be defined) if it wants to be certain to not get a conflict.
(Because you can't hide inherited members.)

I rather think that allowing a part file to *choose* which names it defines itself, gives a good flexibility. You *can* inherit everything, and most days you want to, but you can also import your own declarations.

I can see that having a third party import with no prefix and no `show` can risk the third party library introducing a new name that shadows the name that you used to inherit from the parent file. That would become a compile-time error if we didn't allow conflicts, but without that, it could risk running and doing something wrong. (Most likely something won't type any more, but it's harder to track down the cause.)

I can see that as an argument for not allowing shadowing. (Shadowing by something that can add names at any time is dangerous. But we've been living with that for a long time, not sure this is any different.)

working/augmentation-libraries/parts_with_imports.md
```diff
@@ -312,18 +414,93 @@ Also, every file in a library must belong to the same *package*. The Dart
...
+That is, the extra restrictions enforced by Dart tools are:
+
+*   If any file of a library has a `// @dart=` language version marker, then
+    *all* files in that library must have a language version marker with
+    *the same* language version.
+    Or, phrased differently starting at the root:
+    *   If a library file has a language version marker,
+        then it’s a compile-time error for any sub-part file
+        to not have the same language version marker.
+    *   If a library file has no language version marker,
```
jakemac53, line 434: Ok hear me out - what if you have a part which is defined in a different package, with a different default language version.

Should we explicitly block parts from other packages? Generally the spec doesn't talk about packages, but it seems like a reasonable constraint.

If we do allow it, then I think this should be specified not in terms of the existence of a language version marker, but in terms of the actual computed language version, which must be the same across all parts and the root library file.

lrhn: > Ok hear me out - what if you have a part which is defined in a different package, with a different default language version. Then you are required to have a language marker in those parts.

And that's the reason I explicitly specified that as an error. :sunglasses: 

> Should we explicitly block parts from other packages? 

We should, and I did! 
Part files must belong to the same package as their parent file. That ensures they have the same *default* language version.

We always should have said that, precisely because of this issue.
We've said that a library file and its part file must have the same language file marker, or none.
What we really mean is that they must have the same language version, but we also want that to stay true if someone changes the SDK version.

The *best way*<sup>&tm;</sup> to ensure that they have the same version *no matter what else changes* is to require either all having the same language version marker override, or all having the same default version, and the latter is implied by being in the same package.

working/augmentation-libraries/parts_with_imports.md
lrhn: Thinking about this more, it's definitely possible, but I feel like it's pointing in the opposite direction of the "inherit declartion scope of parents, so you can shadow it" idea.

For that one, instead of every library having all declarations (one global scope), declarations from parents and siblings are only inherited (from the parent) so that the part file can have imports that shadow declarations from outside of its own subtree.
This gives the part file *more* power to override inherited names.

Here, as I read it, we're arguing for making it *harder* to shadow imports inherited from the parent, something that is currently specified as always possible.
That gives the part file *less* power to override inherited names.

I can't make a good argument for why shadowing actual declarations from the current library, something the library author has complete control over, is considerd a good thing, 
but shadowing imports of other libraries, something that is notorious for containing unnecessary names, and being able to introduce new names, if not imported with `show` or `prefix`, is a bad thing.

"Shadow everything!" is  consistent and powerful.
"Shadow nothing!" is consistent, but probably too restrictive.
"Shadow imports!" (what's currently specified) considers imports less important than actual library declarations, which is how I see them. Also because you can *reimport* anything you need by adding imports to the part file.
"Shadow declarations, but not imports" ... I just can't find a good argument for that combination. 

The only benefit I see for the last combiation is that an accidental addition to your own imports not shadowing the original you thought you used, but becoming a conflcit instead.
But it allows an accidental addition to your imports to shadow a declaration of the *current library*, which is at least as bad.

I'll file an issue to discuss these combinations some more.

working/augmentation-libraries/parts_with_imports.md
```diff
@@ -0,0 +1,544 @@
...
+### User guidance tooling
+
+The analyzer and analysis server needs to support and understand the new
+feature. No new user-facing features are needed, but some error handling
+and user guidance may be useful. The following are ideas for
+hypothetical features that the tool may choose to add.
+
+##### Annotations applying to sub-tree
+
+Usually an annotation placed on the `library` declaration applies to the entire
```
jakemac53, line 459: This might have some implications for macros... but I think it is fine. We could always look at allowing macros on parts later on if we have a use case.

jakemac53 (approved):

working/augmentation-libraries/parts_with_imports.md
```diff
@@ -0,0 +1,544 @@
...
+### Export directives
+
+Any Dart file can contain an `export` directive. It makes no difference which
+file an `export` is in, its declarations (filtered by any `hide` or `show`
+modifiers) are added to the library’s single export scope, along with those of
+any other  `export`s in the library and the non-private declarations of the
+library itself. Conflicts are handled as usual (as an error if it’s not the
+*same* declaration).
+
+Allowing a part file to have its own export is mainly intended for macro
```
jakemac53, line 292: Fwiw, macro generated parts can't add exports

lrhn: Well then, it's mainly allowed for consistency then. :)

working/augmentation-libraries/feature-specification.md
```diff
@@ -704,31 +636,48 @@ constructor's initializers, but before any original super initializer or
...
-*   The original constructor is `const` and the augmenting constructor is not
+*   The original constructor is `const` and the augmenting constructor is not,
     or vice versa.
 
-*   The original constructor is a factory constructor and the augmenting
-    constructor has an initializer list.
+*   The original constructor is a non-redirecting factory constructor and the
+    augmenting constructor has an initializer list.
+
+    * The augmenting constructor does not need to repeat the `factory`, but
```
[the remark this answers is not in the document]
lrhn: ACK. Required.

working/augmentation-libraries/parts_with_imports.md
```diff
@@ -0,0 +1,631 @@
...
+that accesses private members, without duplicating the entire library. With
+part files having their own imports, adding configurable URIs for `part`
+directives gives a way to avoid that code duplication, possibly even more
+conveniently if also using augmentations.
+
+The configurable URI for a `part` works just as for imports and exports, it
+chooses the URI that the `part` directive refers to, and after that the
+included file works just as any other part file.
+
+It’s a **compile-time error** if a Dart (parent) file with URI *P* has a `part`
```
scheglov, line 194: When I read this paragraph, it seems that there are a few typos.

"does not parse as a `<partDirective>`" - maybe `< partDeclaration >`?

"does not denote the library of *P*" - we can have `partDirective`s not only in libraries, so maybe "the file with URI *P*"?

lrhn: ACK, `<partDeclaration>` for the entire part file, `<partDirective>` for the `part "...";` directive.
And yes, "the file *P*" is correct. (Damn! I hoped I had found all of the places that assumed a parent file was a library.)

Issue #3791, opened 2024-05-09
[Wildcard Variables][spec] Can we reuse `_` as a non-binding name in function types or record types?


srawlins:
Today, it is illegal to repeat `_` as a name in these spots:

```dart
// The name '_' is already defined.
typedef T = void Function(String _, String _);

// The name '_' is already defined.
typedef void F2(_, _);

// The name '_' is already defined.
void f(void g(_, _)) {}
```

I don't believe function _types_ are addressed in [the current spec](https://github.com/dart-lang/language/blob/main/working/wildcards/feature-specification.md). Should the code above be legal under the wildcards feature?

As for records, it is already illegal to name a positional record field `_`, as "record field names can't be private."

```dart
typedef R = (String _, String _);
```

Would this code become legal under the wildcards feature? Such names would maybe not be considered "private" any longer?

@kallentu @munificent

lrhn:
I'd say yes.

These locations are "documentation names" that have no effect, other than being readable in the source.
I would generally allow such names to be replaced by `_`, precisely because the name doesn't matter, and `_` is a way to say that there is no name. 

For new-style function types and records, you can also just omit the name. 
For old-style function types, ... you should switch to new-style, but using `_` would make it clearer that it's an omitted name, not a type.

kallentu:
I would expect those examples to be allowed.
We can update the spec to include function and record types.

kallentu:
Closing. Updated [the spec](https://github.com/dart-lang/language/blob/main/working/wildcards/feature-specification.md) addressing these explicitly.

Issue #3790, opened 2024-05-09
[Wildcard Variables][spec] Can a local function named `_` be referenced?


srawlins:
Is this allowed under the wildcard feature?

```dart
void f() {
  _() {}
  _();
}
```

It is legal today. AFAIK, `_() {}` is not a _variable declaration_, but is a _local function declaration_.

[The spec](https://github.com/dart-lang/language/blob/main/working/wildcards/feature-specification.md) currently defines "A local declaration is..." with 5 cases, which do not include a local variable declaration.

@kallentu @munificent

lrhn:
I would say that it should not be allowed.
There should not be a `_` bound in a local scope. 

We can make the function decoration allowed, but not introduce a name, but then the function is unreachable, so having it is likely a bug. 
That could also use be an analyzer warning (dead code).

Arguably, we can say the same for a local `var _ = expr;`, but it can at least have a side effect. (Still, just write it as `expr;`, if the code depends on having a context type, bit is only used for its side effect, it's really confusing code.)

kallentu:
Closing. Updated [the spec](https://github.com/dart-lang/language/blob/main/working/wildcards/feature-specification.md) addressing these explicitly.

PR #3802, opened 2024-05-14
[wildcard-variables] Add more explicit details on functions and records.

kallentu:
Updates the spec from discussion on the two following issues: https://github.com/dart-lang/language/issues/3791 and https://github.com/dart-lang/language/issues/3790.

I wanted to update the spec to make some parts clearer to future readers.
Can you both take a look and let me know how I can wordsmith this better?

### DIFF
--- a/working/wildcards/feature-specification.md
+++ b/working/wildcards/feature-specification.md
@@ -147,6 +147,8 @@ It is currently an error for a record field name to begin with `_`
 fields whose name begins with `_` followed by at least one other character
 (even if those later character(s) are `_`).
 
+Named fields of record types are unchanged. It is still a compile-time error for a named field name to start with `_`.
+
 ### Local function declarations
 
 ```dart
--- a/working/wildcards/feature-specification.md
+++ b/working/wildcards/feature-specification.md
@@ -142,9 +142,10 @@ typedef R = (String _, String _);
 (int _, int _) record;
 ```
 
-Record type field names will no longer emit an `INVALID_FIELD_NAME` warning.
-This will have no other effect since the names of positional record fields
-don't matter.
+It is currently an error for a record field name to begin with `_`
+(including just a bare `_`). We relax that error to only apply to record
+fields whose name begins with `_` followed by at least one other character
+(even if those later character(s) are `_`).
 
 ### Local function declarations
 
@@ -158,6 +159,7 @@ void f() {
 It's an error to declare a local function declaration named `_`.
 
 ### Other declarations
+
 Top-level variables, top-level function names, type names, member names,
 etc. are unchanged. They can be named `_` as they are today.
 
--- a/working/wildcards/feature-specification.md
+++ b/working/wildcards/feature-specification.md
@@ -4,7 +4,9 @@ Author: Bob Nystrom
 
 Status: In-progress
 
-Version 1.2
+Version 1.3
+
+## Motivation
 
 Pattern matching brings a new way to declare variables. Inside patterns, any
 variable whose name is `_` is considered a "wildcard". It behaves like a
@@ -74,6 +76,8 @@ library privacy comes into play.
 
 ## Proposal
 
+### Local declarations
+
 A *local declaration* is any of:
 
 *   Function parameters. This includes top-level functions, local functions,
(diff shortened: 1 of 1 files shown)


working/wildcards/feature-specification.md
```diff
@@ -85,6 +89,10 @@ A *local declaration* is any of:
     Foo(_, this._, super._, void _(), {_}) {}
 
     list.where((_) => true);
+
+    void f(void g(_, _)) {}
+
+    typedef T = void Function(String _, String _);
```
munificent, line 95: Ah, good catch. I didn't think about parameters in function types since you can't use them. But the language does still report a collision on them if you have a duplicate and this proposal should remove that error.

working/wildcards/feature-specification.md
```diff
@@ -127,8 +135,31 @@ means you can have multiple local declarations named `_` in the same namespace
...
-Other declarations: top-level variables, top-level function names, type names,
-member names, etc. are unchanged. They can be named `_` as they are today.
+### Record type positional fields
+
+```dart
+typedef R = (String _, String _);
+(int _, int _) record;
+```
+
+Record type field names will no longer emit an `INVALID_FIELD_NAME` warning.
```
munificent, line 145: I believe it's [an error](https://github.com/dart-lang/language/blob/main/accepted/3.0/records/feature-specification.md#record-expressions), not a warning, for a record field name to start with `_`.

Also, the language specifies when an error is reported, but not which error. Here, I'd probably say something like:

> It is currently an error for a record field name to begin with `_` (including just a bare `_`). We relax that error to only apply to record fields whose name begins with `_` followed by at least one other character (even if those later character(s) are `_`).

working/wildcards/feature-specification.md
```diff
@@ -127,8 +135,31 @@ means you can have multiple local declarations named `_` in the same namespace
...
+### Local function declarations
+
+```dart
+void f() {
+  _() {} // Error.
+  _(); // Error.
+}
+```
+
+It's an error to declare a local function declaration named `_`.
```
munificent, line 158: Good catch!

working/wildcards/feature-specification.md
```diff
@@ -127,8 +135,31 @@ means you can have multiple local declarations named `_` in the same namespace
...
+
+### Other declarations
```
munificent, line 160: Markdown nit: Can you put a blank line after this header?

It doesn't really matter but it looks nicer to me at least.

munificent (approved):
Couple of minor things then LGTM! Thanks for turning this into a real proposal!

working/wildcards/feature-specification.md
kallentu: Argh, words. Thank you. I'll just steal that :)

> Also, the language specifies when an error is reported, but not which error.

And this is also good to know.

lrhn: And it's still a compile-time error for a *named* field's name to start with `_`, it's only positional field names that can be ignored.

kallentu: Added an extra line on named fields. I think it's good to reiterate that. Thanks.

kallentu:
Merging now, but open to following up if there are any more comments.

working/wildcards/feature-specification.md
```diff
@@ -85,6 +89,10 @@ A *local declaration* is any of:
     Foo(_, this._, super._, void _(), {_}) {}
```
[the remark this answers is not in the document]
kallentu: Yeah, I ran into this making language tests. I'll remove it. Maybe add an extra bit to talk about ti?

working/wildcards/feature-specification.md
```diff
@@ -74,6 +76,8 @@ library privacy comes into play.
 
 ## Proposal
 
+### Local declarations
+
 A *local declaration* is any of:
```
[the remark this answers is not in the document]
kallentu: Done. I also changed the paragraph after.

working/wildcards/feature-specification.md
kallentu: That is what @lrhn suggested https://github.com/dart-lang/language/issues/3790#issuecomment-2104021741. We can do that. I'll make the change.

PR #3812, opened 2024-05-15
Use separate grammar rules for null-aware elements.

munificent:
Stuffing the `?` directly inside `expressionElement` and `mapEntryElement` lets us reuse some existing specification around leaf elements, but is otherwise confusing because it makes the case analysis for the different kinds of elements non-disjoint.

This fixes that by defining entirely separate rules for null-aware elements.

### DIFF
--- a/working/0323-null-aware-elements/feature-specification.md
+++ b/working/0323-null-aware-elements/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: In-progress
 
-Version 0.1
+Version 0.2 (see [CHANGELOG](#CHANGELOG) at end)
 
 Experiment flag: null-aware-elements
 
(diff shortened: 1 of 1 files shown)


working/0323-null-aware-elements/feature-specification.md
```diff
@@ -466,24 +499,24 @@ add two new cases to that procedure:
 
 [unified-dynamic-element]: https://github.com/dart-lang/language/blob/main/accepted/2.3/unified-collections/feature-specification.md#to-evaluate-a-collection-element
 
-*   If `element` is a null-aware `expressionElement` with expression `e`:
+*   If `element` is a `nullAwareExpressionElement` with expression `e`:
 
     *   Evaluate `e` to `v`.
 
     *   If `v` is not `null` then append it to *result*. *Else the `null` is
         discarded.*
```
lrhn, line 508: Do we need to do something special for coercions here, or will the value be coerced before the `null` check?
```dart
void main() {
  // True, but compiler doesn't know that.
  var b = DateTime.now().millisecondsSinceEpoch > 0;
  var l1 = <Function>[?(b ? CallableObject() : null)];
  var l2 = <int Function(int)>[?(b ? id : null)];
}

T id<T>(T value) => value;
class CallableObject {
  void call() {}
}

````
Will these list literals work or fail to compile?

If we coerce before null-checking, then we coerce `CallableObject?` to `Function?` which is *not* supported, so it's a compile-time error. Same for coercing `T Function<T>(T)?` to `int Function(int)?`. Neither of these are allowed:
```dart
  Function? v1 = (b ? CallableObject() : null);
  int Function(int) v2 = (b ? id : null);
```

If we coerce after null-checking, the code should works. And it's probably what the author wanted.
So should we?

That would either mean being *explicit* about where we do coercion, or somehow suggest to inference that the null check is not a coercion-point (a concept we haven't specified other than by the algorithm used.)

WDYT? @stereotype441 

The static semantics with a context element type `E` would be (paraphrased):
> Let `S be the static type of `e` in context `E?`. Let `S'` be **NonNull**(`S`).
> It's a compile-time error if `S'` is not assignable to the greatest closure of `E`.

and runtime semantics of executing `?e` would be:
> Evaluate `e` to a value *v*.
> If *v* is not the value `null`, then
> * if the runtime type of *v* is a subtype of *E*, let *w* be *v*, otherwise coerce *v* from type `S'` to a value *w* of type `E`. _(Which is always possible because `S'` is assignable to `E`._)
> * Then append *w* to *result*.

(Not sure `dynamic` has an issue, downcasting before or after both work.)

working/0323-null-aware-elements/feature-specification.md
```diff
@@ -466,24 +499,24 @@ add two new cases to that procedure:
...
-        Else continue...
+    *   If `element` has a null-aware key and `kv` is `null`, then stop. Else
+        continue...
 
     *   Evaluate `v` to a value `vv`.
 
-    *   If `element` is a value null-aware element and `vv` is `null`, then
-        stop. Else continue...
+    *   If `element` has a null-aware value and `vv` is `null`, then stop. Else
+        continue...
```
lrhn, line 519: Here we probably want to coerce both after the `null` check, although it's probably usually the value where it makes sense (functions are bad keys).

lrhn (approved):

working/0323-null-aware-elements/feature-specification.md
munificent: Good question! I filed a bug so we can discuss this more: https://github.com/dart-lang/language/issues/3821

Issue #3822, opened 2024-05-21
Do static declarations preclude anything?


eernstg:
With extension types, we introduced the notion that one member declaration can preclude another member declaration. It is always a method that precludes a setter which is in the interface of a superinterface, or a setter that precludes a method which is in the interface of a superinterface (other pairs of kinds of declarations are not subject to preclusion).

For example:

```dart
extension type E1(int it) {
  void m() {}
}

extension type E2(int it) implements E1 {
  set m(_) {} // OK.
}

class C1 {
  void m() {}
}

class C2 extends C1 {
  set m(_) {} // Compile-time error.
}
```

If we consider classes (like `C1` and `C2` above) then we get a compile-time error because `C2` "has" a method named `m` and it "declares" a setter named `m=`, and that's an error.

However, when we consider extension types we say that the setter `E2.m=` _precludes_ the method `E1.m`, and there is no error. So we're basically ignoring the method-respectively-setter from the superinterface when it conflicts with a setter-respectively-method in the subtype.

(The motivation for preclusion is that extension types are allowed to _redeclare_ a name more freely than classes, mixins, etc. can _override_ a name, and we did not want to have a compile-time error for some specific situations involving redeclaration.)

However, the wording in the specification does not indicate that the subtype declaration must be an instance member declaration, which means that it should also work for static declarations:

```dart
extension type E1(int it) {
  void set m() {}
}

extension type E2(int it) implements E1 {
  static m(_) {} // OK?
}
```

In this case we would, arguably, again say that `E2.m=` precludes `E1.m`, and hence there is no error.

However, the implemented behavior is to flag this example as an error (which means that a static declaration _does not_ get to preclude anything).

I think we should adjust the specification to say that preclusion only applies when the given member declarations are instance member declarations, and static declarations are never the cause of preclusion. This makes sense conceptually as well, because the motivation for preclusion was based on the similarities and difference between overriding relations and redeclaration relations.

@dart-lang/language-team, WDYT?

eernstg:
PR here: #3823.

lrhn:
Barring #1711, we should probably be consistent and say that you can't introduce a static member with the same base name as an inherited instance member.

The definition of "inherited" may need to be kept up-to-date, but something like "an instance member in the type signature (set of instance member signatures) that the type, which it did not declare itself."

The instance members it declares itself conflict more eagerly with static members, since they go into the same scope.
Inherited members don't necessarily exist in the same scope as the declarations of the type itself.

[...]

PR #3823, opened 2024-05-21
Clarify preclusion

eernstg:
See https://github.com/dart-lang/language/issues/3822 for some background information.

This PR changes the specification of preclusion in extension types such that it is never caused by a static declaration, it can only be caused by an instance member declaration.


eernstg:
@sgrekhov, it's now settled: Static members cannot give rise to preclusion.

PR #3827, opened 2024-05-22
[wildcard-variables] Fix typo in 'imports' example.

sgrekhov:
Unnamed (and private) extensions are non visible after the import. To make the example valid an extension should have a name

### DIFF
--- a/working/wildcards/feature-specification.md
+++ b/working/wildcards/feature-specification.md
@@ -166,7 +166,7 @@ because the function is unreachable.
 
 ```dart
 // a.dart
-extension on String {
+extension MyString on String {
   bool get isBlank => trim().isEmpty;
 }
 ```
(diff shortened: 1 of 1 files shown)


working/wildcards/feature-specification.md
```diff
@@ -181,7 +181,7 @@ main() {
 ```
 
 Import prefixes named `_` are non-binding and will provide access to the
```
lrhn, line 183: I'd say "but" instead of "and" here. Maybe even "but will still"

The part after "and" is how non-`_` prefixes work too, so erasing it should make no difference.
It's there for emphasis, saying that *even though* `_` does not introduce a name for the imports, the extensions still work.

lrhn (approved):

kallentu (approved):
Ah thanks for catching this.

sgrekhov:
Changed "and" -> "but"

PR #3835, opened 2024-05-24
Add feature specification of static extensions

eernstg:
This PR adds a proposed feature specification of the 'static extensions' mechanism. For background, see https://github.com/dart-lang/language/issues/723. The proposal is mainly an excerpt from an earlier proposal about implicit construction ([here](https://github.com/dart-lang/language/commit/ccb70ae2572472fae38a38181c6babbb583d0082)).


eernstg:
Hi @stereotype441, I've extracted a proposal about extensions that are capable of adding static members and factory constructors to existing declarations that are where such members can be declared (classes, extension types, etc.).

I'd like to use an approach which is similar to what @munificent has done a few times for documents in 'working': The review would just be a brief sanity check, and then we'd handle all the discussions about details in separate issues here in the language repo.

There are two versions. One version introduces a new kind of declaration (`static extension`). I tend to prefer this, because the way the declaration uses the on-type in connection with constructors is so different from the way we're using the on-type of a normal `extension` with instance members. It is also an anomaly that we would need to allow the on-type to be an extension, because they can have static members, so we may want to add them.

The other version is a generalization of the `extension` declaration that we have today. This version has some other benefits: A rather large handful of things need not be specified in this variant, because we already have `extension` declarations, and they already have the rules we need.

Anyway, I expect that we'll have some really nice dog fights about this choice.

Could you take a look in order to have that brief sanity check?

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+declared bounds. A compile-time error occurs if no type arguments are
+passed to `E`, and type arguments `U1 .. Uk` are passed to `C`, but no list
+of actual type arguments for the type variables of `E` can be found such
+that the instantiated constructor return type of `E` with said type
+arguments is `C<U1 .. Uk>`.
+
+*Note that we must be able to choose the values of the type parameters
+`X1 .. Xs` such that the instantiated constructor return type is
+exactly `C<U1 .. Uk>`, it is not sufficient that it is a subtype thereof,
+or that it differs in any other way.*
```
lrhn, line 260: Why not? (Or rather: I just don't understand what the previous paragraphs are trying to do.)

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+
+A compile-time error occurs if the invocation passes actual type arguments
+to both `E` and `C`, call them `S1 .. Ss` and `U1 .. Uk`, respectively,
```
lrhn, line 263: So this is a syntax like `E<T>.C<S>.id(...)`. In that case, it's probably not relevant to this variant. (Or there is something that needs to be said much sooner about how constructors are declared, which likely was in the grammar section of the other variant.)

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+Otherwise, the invocation is partially resolved to a set of candidate
+constructors found in extensions. Each of the candidates _kj_ is
+vetted as follows:
+
+Assume that _kj_ is a constructor declared by a extension _D_ named
+`E` with type parameters `X1 extends B1 .. Xs extends Bs` and on-type
+`C<S1 .. Sm>`.  Find actual values `U1 .. Us` for `X1 .. Xs` satisfying the
+bounds `B1 .. Bs`, such that `([U1/X1 .. Us/Xs]C<S1 .. Sm>) == C<T1 .. Tm>`.
+If this fails then remove _kj_ from the set of candidate constructors.
+Otherwise note that _kj_ uses actual type arguments `U1 .. Us`.
```
lrhn, line 295: I'd rather not remove candidates based on the `on` type.
It's great enough to have a constructor which can only be used for some instantiations, I don't want to also provide overloading, where
`List<int>.defaultFilled()` and `List<String>.defaultFilled()` refer to different constructors.

So, `type.id` always refers to precisely *one* declaration. If there is more than one static extension with that name in scope, you'll need to disambiguate. We don't do that based on the type.

(I know we failed thos for extension instance methods, where `List<int>` and `List<String>` can have different implementations. I think static members are trickier, and constructors should probably be treated as statics. That said, it's probably less error-prone to allow overloading of constructors based on context type than for instance members based on receiver type. The context is often more obvious than the current type of a promoted variable.)

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+just mutual subtypes.*
+
+*Note that explicitly resolved invocations of constructors declared in
+extensions are an exception in real code, usable in the case where a
+name clash prevents an implicitly resolved invocation. However, implicitly
+resolved invocations are specified in the rest of this section by reducing
+them to explicitly resolved ones.*
+
+A constructor invocation of the form `C<T1 .. Tm>.name(args)` is partially
+resolved by looking up a constructor named `C.name` in the class `C` and in
```
lrhn, line 277: This looks explicitly for constructors. It probably should look for any static member, and eventually make it an error if there is only one and it's not a constructor.

I know `C<T...>.name` must be a constructor invocation, but (a) I don't want to hardcode that in case we want "class members" at some point, and (b) it differs from `C.name`, even though both can denote a constructor.

Which means that we can't *generally* spearate lookups into member and constructors, we have to be able to handle `C.name(args)`, which can be either,
and I don't want to make it *better* to write `C<int>.name(args)` than `C.name(args)`-which-infers-`<int>`.

No overloading! :grin:

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+Otherwise note that _kj_ uses actual type arguments `U1 .. Us`.
+
+If all candidate constructors have been removed, or more than one candidate
+remains, a compile-time error occurs. Otherwise, the invocation is
+henceforth treated as `E<U1 .. Us>.C<T1 .. Tm>.name(args)` (respectively
+`E<U1 .. Us>.C<T1 .. Tm>(args)`).
+
+A constructor invocation of the form `C.name(args)` (respectively
+`C(args)`) where `C` resolves to a non-generic class is resolved in the
+same manner, with `m == 0`. *In this case, type parameters declared by `E`
```
lrhn, line 304: Can't do that if "the same manner" throws away extensions with non-constructor `name` members. There needs to be a case for `C.name(args)` which can handle methods, getters and constructors. (And similarly for `C.name`, which can tear off constructors too.)

I think the constructors and members cases should use the same *lookup*, and then only special-case after we know that a constructor or non-constructor was found (and if both, it's a conflict, even if the syntax currently only allows one or the other).

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+is treated as `E.C.name(args)` (respectively `E.C(args)`).
+
+Otherwise, when there are two or more candidates from extensions,
+an error occurs. *We do not wish to specify an approach whereby `args` is
+subject to type inference multiple times, and hence we do not support type
+inference for `C.name(args)` in the case where there are multiple distinct
+declarations whose signature could be used during the static analysis of
+that expression.*
+
+### Dynamic Semantics
```
lrhn, line 328: Runtime semantics.

The word "dynamic" is loaded in Dart, don't use it for anything but `dynamic`.

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+
+```dart
+class SortedList<X> {
+  final Comparator<X> _comparator;
+  SortedList(Comparator<X> this._comparator);
+  // ... lots of stuff that doesn't matter here ...
+}
+
+extension<X extends Comparable<X>> on SortedList<X> {
+  SortedList.ofComparable(): super((X a, X b) => a.compareTo(b));
```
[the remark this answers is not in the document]
eernstg: Of course, thanks!

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+
+### Syntax
+
+The grammar remains unchanged.
+
+However, it is no longer an error to declare a factory constructor in an
+extension declaration that has an on-class, be it redirecting or not,
+constant or not.  *Such declarations may of course give rise to errors as
+usual, e.g., if a redirecting factory constructor redirects to a
+constructor that does not exist, or there is a redirection cycle.*
```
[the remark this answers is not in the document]
eernstg: OK, let's include them. (Just like they were a couple of weeks ago; I was worried about the implications, but it does seem to be safe. ;-)

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+
+In an extension declaration of the form `extension E on C {...}` where `C`
+is an identifier or an identifier with an import prefix that resolves to a
+class, mixin, enum, or extension type declaration, we say that the
+_on-class_ of the extension is `C`. If `C` resolves to a non-generic class
+then we say that the _constructor return type_ of the extension is `C`.
+
+*If `C` resolves to a generic class then the extension does not have a
+constructor return type.*
```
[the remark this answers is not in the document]
eernstg: OK, done.

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+
+In a extension of the form `extension E on C<T1 .. Tk> {...}`
+where `C` is an identifier or prefixed identifier that resolves to a class,
+mixin, enum, or extension type declaration, we say that the _on-class_
+of `E` is `C`, and the _constructor return type_ of `E` is `C<T1 .. Tk>`.
+
+In all other cases, an extension declaration does not have an on-class.
```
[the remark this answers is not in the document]
eernstg: Done.

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+A compile-time error occurs if an extension has an on-clause of the
+form `on C` where `C` denotes a generic class and no type arguments are
+passed to `C` *(i.e., it is a raw type)*, and the extension contains
+one or more constructor declarations.
+
+Tools may report diagnostic messages like warnings or lints in certain
+situations. This is not part of the specification, but here is one
+recommended message:
+
+A compile-time message is emitted if a extension _D_ declares a
```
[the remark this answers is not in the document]
eernstg: Done

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+constructor or a static member with the same name as a constructor or a
+static member in the on-class of _D_.
```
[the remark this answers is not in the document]
eernstg: Ack.

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+form `on C` where `C` denotes a generic class and no type arguments are
+passed to `C` *(i.e., it is a raw type)*, and the extension contains
+one or more constructor declarations.
+
+Tools may report diagnostic messages like warnings or lints in certain
+situations. This is not part of the specification, but here is one
+recommended message:
+
+A compile-time message is emitted if a extension _D_ declares a
+constructor or a static member with the same name as a constructor or a
```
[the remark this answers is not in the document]
eernstg: Done.

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+*This can be used to invoke a static member of a specific extension in
+order to manually resolve a name clash. At the same time, in current Dart
+(without the feature which is specified in this document), this is the only 
+way we can invoke a static member of an extension, so it may be useful for
+backward compatibility reasons.*
+
+A static member invocation on a class `C`, of the form `C.m()` (or any
+other member access), is resolved by looking up static members in `C` named
+`m` and looking up static members of every accessible extension with
+on-class `C` and a static member named `m`.
```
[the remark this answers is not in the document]
eernstg: Constructors are handled separately, later.

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+
+If `C` contains such a declaration then the expression is an invocation of
+that static member of `C`, with the same static analysis and dynamic
```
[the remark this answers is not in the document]
eernstg: The language specification uses 'dynamic semantics' or 'behavior'. I changed it to 'dynamic semantics'.

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+extension declaration that has an on-class, be it redirecting or not,
+constant or not.  *Such declarations may of course give rise to errors as
+usual, e.g., if a redirecting factory constructor redirects to a
+constructor that does not exist, or there is a redirection cycle.*
+
+In an extension declaration of the form `extension E on C {...}` where `C`
+is an identifier or an identifier with an import prefix that resolves to a
+class, mixin, enum, or extension type declaration, we say that the
+_on-class_ of the extension is `C`. If `C` resolves to a non-generic class
+then we say that the _constructor return type_ of the extension is `C`.
```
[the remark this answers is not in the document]
eernstg: Good point! Done.

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+above)*.
+
+#### The instantiated constructor return type of an extension
+
+We associate an extension declaration _D_ named `E` with formal type
+parameters `X1 extends B1 .. Xs extends Bs` and an actual type argument
+list `T1 .. Ts` with a type known as the _instantiated constructor return
+type of_ _D_ _with type arguments_ `T1 .. Ts`.
+
+When an extension declaration _D_ named `E` has an on-clause which is
```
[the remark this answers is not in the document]
eernstg: Done.

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+a non-generic class `C`, the instantiated constructor return type is `C`,
+for any list of actual type arguments.
+
+*It is not very useful to declare a type parameter of an extension
+which isn't used in the constructor return type, because it can only be
+passed in an explicitly resolved constructor invocation, e.g.,
+`E<int>.C(42)`. In all other invocations, the value of such type variables
+is determined by instantiation to bound. In any case, the type parameters
+are always ignored by static member declarations, they are only relevant to
+constructors.*
```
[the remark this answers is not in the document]
eernstg: Right, done.

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+`T1 .. Ts`, the instantiated constructor return type of _D_ with type
+arguments `T1 .. Ts` is `[T1/X1 .. Ts/Xs]C<S1 .. Sk>`.
+
+#### Explicit invocation of a constructor in an extension
+
+Explicit constructor invocations are similar to static member invocations,
+but they need more detailed rules because they can use the formal type
+parameters declared by an extension.
+
+An _explicitly resolved invocation_ of a constructor named `C.name` in a
```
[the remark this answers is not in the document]
eernstg: The constructors are named as in the class, so that would be `C.name`. But we have the form `E.C.name` as well which is used to specify a fully explicit invocation.

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+extension declaration _D_ named `E` with `s` type parameters and on-class
+`C` can be expressed as `E<S1 .. Ss>.C.name(args)`,
```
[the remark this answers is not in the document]
eernstg: It is `E<S1 .. Ss>.C.name(args)` because we need to be able to check all parts against each other (arriving at `E<S1 .. Ss>.C<U1 .. Uk>.name(args)` somehow). This is what it takes to specify a fully explicit static-extension constructor invocation. It is redundant but easy to check, all other forms will be transformed into this fully explicit form in order to define the meaning of the less informative versions.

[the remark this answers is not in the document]
lrhn: It might not be new syntax since `<S1...,Sn>` was made a selector, but I think it does need its own "Grammar-interpretation" section to describe how it's interpreted. It's a new syntax combination, previously a constructor invocation could not have the form `id<Types>.id.id(args)`. 
Something else could, it would not be a single constructor invocation.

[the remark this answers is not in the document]
lrhn: > This is what it takes to specify a fully explicit static-extension constructor invocation.

Only if we *want* that syntax. (Or we *need* it, which may be the case.)

Otherwise I'd have gone with `extension ... on C<...> { E.name(params) ... }`, where `E.name` is a constructor which can also be accessed as `C<T1,..,Tn>.name`, with inference going from `T1`...`Tn` to the type parameters of `N` through the `on` type.

I think that can work for invocation, but it breaks down for tear-offs. Maybe it breaks down if you abstract over the types used in the invocation too.

I can see why `C.name` would make sense, making it a constructor that has the same type variables of `C`, but then I'd want to write those type parameters on the constructor myself, not inherit them from somewhere else.
That is:
```dart
extension E<K, V> on Map<K, V> {
  factory Map<K extends Comparable<K>, V>.ordered() => SplayTreeMap<K, V>(compareComparable<K>);
}
```
Which is new syntax.
Without new syntax, it can at most be:
```dart
extension E<K2, V2> on Map<K2, V2> {
  factory Map.ordered() => SplayTreeMap<K, V>(compareComparable<K>);
}
```
where it's very unclear where the type variables available to `Map.ordered` comes from. If anywhere.

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+In a extension of the form `extension E on C<T1 .. Tk> {...}`
+where `C` is an identifier or prefixed identifier that resolves to a class,
+mixin, enum, or extension type declaration, we say that the _on-class_
+of `E` is `C`, and the _constructor return type_ of `E` is `C<T1 .. Tk>`.
+
+In all other cases, an extension declaration does not have an on-class.
+
+*For example, an extension whose on-type is a type variable does not have
+an on-class, and neither does an extension whose on-type is a function
+type.*
```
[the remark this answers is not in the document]
eernstg: Done.

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+
+### Static Analysis
+
+At first, we establish some sanity requirements for an extension declaration
+by specifying several errors.
+
+A compile-time error occurs if an extension has an on-clause of the
+form `on C` where `C` denotes a generic class and no type arguments are
+passed to `C` *(i.e., it is a raw type)*, and the extension contains
+one or more constructor declarations.
```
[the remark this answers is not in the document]
eernstg: Good point!

It is a compile-time error for any extension member invocation if one or more actual type arguments do not satisfy the declared bounds (that is, all such invocations are regular-bounded), but it _is_ possible for an on-type of an extension to be super-bounded (it's just that we can't use the type parameters of the extension to make it so).

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+that static member of `C`, with the same static analysis and dynamic
+behavior as before the introduction of this feature.
+
+Otherwise, an error occurs if fewer than one or more than one declaration
+named `m` was found. *They would necessarily be declared in extensions.*
+
+Otherwise, the invocation is resolved to the given static member
+declaration in an extension named `Ej`, and the invocation is treated
+as `Ej.m()` *(this is an explicitly resolved invocation, which is specified
+above)*.
```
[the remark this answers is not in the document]
eernstg: Right, I added text to handle the situation where `C.m` denotes a constructor.

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+
+#### The instantiated constructor return type of an extension
+
+We associate an extension declaration _D_ named `E` with formal type
```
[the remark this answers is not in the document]
eernstg: Good point! Rewritten extensively.

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+type of_ _D_ _with type arguments_ `T1 .. Ts`.
+
+When an extension declaration _D_ named `E` has an on-clause which is
+a non-generic class `C`, the instantiated constructor return type is `C`,
+for any list of actual type arguments.
+
+*It is not very useful to declare a type parameter of an extension
+which isn't used in the constructor return type, because it can only be
+passed in an explicitly resolved constructor invocation, e.g.,
+`E<int>.C(42)`. In all other invocations, the value of such type variables
```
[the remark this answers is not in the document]
eernstg: Rewritten.

working/0723-static-extensions/feature-specification-variant2.md
```diff
@@ -0,0 +1,347 @@
...
+it has an on-type `C<S1 .. Sk>`, the instantiated constructor return type
+of _D_ is `C<S1 .. Sk>`. *In this case the on-type is a fixed type (also
+known as a ground type), e.g., `List<int>`. This implies that the
+constructor return type of D is the same for every call site.*
+
+Finally we have the general case: Consider an extension declaration
+_D_ named `E` with formal type parameters `X1 extends B1 .. Xs extends Bs`
+and a constructor return type `C<S1 .. Sk>`. With actual type arguments
+`T1 .. Ts`, the instantiated constructor return type of _D_ with type
+arguments `T1 .. Ts` is `[T1/X1 .. Ts/Xs]C<S1 .. Sk>`.
```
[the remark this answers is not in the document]
eernstg: Deleted this paragraph, the information is now given earlier.

working/0723-static-extensions/feature-specification-variant2.md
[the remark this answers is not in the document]
eernstg: No changes made for this, that would be a separate discussion in the language team, presumably in a github issue.

working/0723-static-extensions/feature-specification-variant2.md
eernstg: I have inserted a paragraph where static member invocations are detected, and a reference to the specification of static member invocations is given.

working/0723-static-extensions/feature-specification-variant2.md
eernstg: The phrase 'dynamic semantics' is used to denote the dynamic or run-time semantics in the language specification. There are no occurrences of 'run-time semantics' in normative text, and only one occurrence in commentary.

We could change this, but it makes no sense to say that we should use non-standard terminology in just one feature specification, that should be a global operation.

working/0723-static-extensions/feature-specification-variant2.md
eernstg: I created https://github.com/dart-lang/language/issues/4049 in order to take this discussion separately.

eernstg (commented):
Several changes introduced based on the review.

eernstg:
@lrhn, PTAL. I created an issue https://github.com/dart-lang/language/issues/4049 in order to discuss the main remaining language design question.

This PR does not update working/0723-static-extensions/feature-specification-variant1.md, it only changes variant2.

working/0723-static-extensions/feature-specification-variant1.md
```diff
@@ -0,0 +1,467 @@
...
+In a static extension of the form `static extension E on C {...}` where `C`
+is an identifier or an identifier with an import prefix, we say that the
+on-class of the static extension is `C`.
+
+If `C` denotes a non-generic class, mixin, mixin class, or extension
+type then we say that the _constructor return type_ of the static extension
+is `C`.
+
+If `C` denotes a generic class then `E` is treated as
+`static extensionE on C<T1 .. Tk> {...}`
```
leafpetersen, line 184: `extensionE` -> `extension E`

working/0723-static-extensions/feature-specification-variant1.md
```diff
@@ -0,0 +1,467 @@
...
+
+In a static extension of the form `static extension E on C {...}` where `C`
+is an identifier or an identifier with an import prefix, we say that the
+on-class of the static extension is `C`.
+
+If `C` denotes a non-generic class, mixin, mixin class, or extension
+type then we say that the _constructor return type_ of the static extension
+is `C`.
+
+If `C` denotes a generic class then `E` is treated as
```
leafpetersen, line 183: Above you say "non-generic class, mixin, mixin class, or extension type", here you say "a generic class".   Is the difference deliberate?

working/0723-static-extensions/feature-specification-variant1.md
```diff
@@ -0,0 +1,467 @@
...
+
+An _explicitly resolved invocation_ of a static member of a static
+extension named `E` is an expression of the form `E.m()` (or any other
+member access, *e.g., `E.m`, `E.m = e`, etc*), where `m` is a static member
+declared by `E`.
+
+*This can be used to invoke a static member of a specific static extension
+in order to manually resolve a name clash.*
+
+A static member invocation on a class `C`, of the form `C.m()` (or any
```
leafpetersen, line 308: "a class, mixin class, mixin etc"?

working/0723-static-extensions/feature-specification-variant1.md
```diff
@@ -0,0 +1,467 @@
...
+`T1 .. Ts`, the instantiated constructor return type of _D_ with type
+arguments `T1 .. Ts` is `[T1/X1 .. Ts/Xs]C<S1 .. Sk>`.
+
+#### Explicit invocation of a constructor in a static extension
+
+Explicit constructor invocations are similar to static member invocations,
+but they need more detailed rules because they can use the formal type
+parameters declared by a static extension.
+
+An _explicitly resolved invocation_ of a constructor named `C.name` in a
```
leafpetersen, line 363: I'm pretty lost in here.  Why do we support anything other than `E<S1 .. Ss>.C.name(args)`?  And what do the other forms mean?  Some examples or motivation might help here.

working/0723-static-extensions/feature-specification-variant1.md
```diff
@@ -0,0 +1,467 @@
...
+static extension declaration _D_ named `E` with `s` type parameters and
+on-class `C` can be expressed as `E<S1 .. Ss>.C.name(args)`,
+`E.C<U1 .. Uk>.name(args)`, or `E<S1 .. Ss>.C<U1 .. Uk>.name(args)`
+(and similarly for a constructor named `C` using `E<S1 .. Ss>.C(args)`,
+etc).
+
+A compile-time error occurs if the type arguments passed to `E` violate the
+declared bounds. A compile-time error occurs if no type arguments are
+passed to `E`, and type arguments `U1 .. Uk` are passed to `C`, but no list
+of actual type arguments for the type variables of `E` can be found such
```
leafpetersen, line 373: "can be found" how?  That is, we must specify what mechanism is used to find it (since otherwise "I can't find it, sorry" is always a valid answer).

working/0723-static-extensions/feature-specification-variant1.md
```diff
@@ -0,0 +1,467 @@
...
+normalization occurs. *In other words, the types must be equal, not
+just mutual subtypes.*
+
+*Note that explicitly resolved invocations of constructors declared in
+static extensions are an exception in real code, usable in the case where a
+name clash prevents an implicitly resolved invocation. However, implicitly
+resolved invocations are specified in the rest of this section by reducing
+them to explicitly resolved ones.*
+
+A constructor invocation of the form `C<T1 .. Tm>.name(args)` is partially
```
leafpetersen, line 396: Should this be a new section?  The section heading here is "Explicit invocation..." but this is now moving into implicit invocation.

working/0723-static-extensions/feature-specification-variant1.md
```diff
@@ -0,0 +1,467 @@
...
+resolved by looking up a constructor named `C.name` in the class `C` and in
+every accessible static extension with on-class `C`. A compile-time error
+occurs if no such constructor is found. Similarly, an invocation of the
+form `C<T1 ... Tm>(args)` uses a lookup for constructors named `C`.
```
leafpetersen, line 400: or `C.new`

working/0723-static-extensions/feature-specification-variant1.md
```diff
@@ -0,0 +1,467 @@
...
+static analysis and dynamic semantics apply. *That is, the class always
+wins.*
+
+Otherwise, the invocation is partially resolved to a set of candidate
+constructors found in static extensions. Each of the candidates _kj_ is
+vetted as follows:
+
+Assume that _kj_ is a constructor declared by a static extension _D_ named
+`E` with type parameters `X1 extends B1 .. Xs extends Bs` and on-type
+`C<S1 .. Sm>`.  Find actual values `U1 .. Us` for `X1 .. Xs` satisfying the
```
leafpetersen, line 412: Find by what mechanism?

working/0723-static-extensions/feature-specification-variant1.md
```diff
@@ -0,0 +1,467 @@
...
+We associate a static extension declaration _D_ named `E` with formal type
+parameters `X1 extends B1 .. Xs extends Bs` and an actual type argument
+list `T1 .. Ts` with a type known as the _instantiated constructor return
+type of_ _D_ _with type arguments_ `T1 .. Ts`.
+
+When a static extension declaration _D_ named `E` has an on-clause which
+denotes a non-generic class `C`, the instantiated constructor return type
+is `C`, for any list of actual type arguments.
+
+*It is not very useful to declare a type parameter of a static extension
```
leafpetersen, line 337: I don't think this is true.  The parameters to the constructor (and the body) can use the type parameter.  Something like:
```dart
extension TypeKeyedMap<T> on Map<Type, Object> {
  Map.singleton(T x) => {T : x};
  void addByType<S>(S x) => this[S] = x;
}
```

working/0723-static-extensions/feature-specification-variant1.md
```diff
@@ -0,0 +1,467 @@
...
+Otherwise note that _kj_ uses actual type arguments `U1 .. Us`.
+
+If all candidate constructors have been removed, or more than one candidate
+remains, a compile-time error occurs. Otherwise, the invocation is
+henceforth treated as `E<U1 .. Us>.C<T1 .. Tm>.name(args)` (respectively
+`E<U1 .. Us>.C<T1 .. Tm>(args)`).
+
+A constructor invocation of the form `C.name(args)` (respectively
+`C(args)`) where `C` denotes a non-generic class is resolved in the
+same manner, with `m == 0`. *In this case, type parameters declared by `E`
```
leafpetersen, line 424: I don't think this is necessarily true, because the type parameters could be determined by the arguments, right?

working/0723-static-extensions/feature-specification-variant1.md
```diff
@@ -0,0 +1,467 @@
...
+will be bound to values selected by instantiation to bound.*
+
+Consider a constructor invocation of the form `C.name(args)` (and similarly
+for `C(args)`) where `C` denotes a generic class. As usual, the
+invocation is treated as in the pre-feature language when it denotes a
+constructor declared by the class `C`.
+
+In the case where the context type schema for this invocation fully
```
leafpetersen, line 432: This all feels like it needs some work.  I'm not sure if I have a full model of this in my head, but my rough initial cut is something like the following (by example).  If we have an extension declaration as follows:
```dart
extension E<S, T> on Map<int, T> {
  Map.bar(S, x, T y) => {3 : y};
}
```
we should semantically view this as if we had declared a static method something like the following:
```dart
  static Map<int, T> E_Map_bar<S, T>(S x, T y) => E<S, T>.Map.bar(x);
```
and hence when we encounter something like `Map.bar("hello")` in downwards context `K`, we treat it as an inference problem equivalent to solving `E_Map_bar("hello")` in downwards context `K`.  What do we do if there are explicit type arguments passed though?  That is, what is the interpretation of `Map<T1, T2>.bar("hello")` in downwards context `K`?  I think one possible answer is that we treat this as equivalent to solving `E_Map_bar("hello")` in downwards context `Map<T1, T2>`.  I'm not 100% sure about that though.  WDYT?

leafpetersen: Filed https://github.com/dart-lang/language/issues/4053 to discuss this.

eernstg:
Thanks for the input on working/0723-static-extensions/feature-specification-variant1.md!

I actually hadn't planned to work on that document in this PR (because that document is nearly a copy-paste of https://github.com/dart-lang/language/blob/main/working/0107%20-%20implicit-constructors/feature-specification.md), but I'll go over it now. This means introducing into 'variant1' many of the improvements to 'variant2' which were suggested in reviews on this PR, because they are similar in so many ways. This might preempt some comments on 'variant1', but I'll try to keep track of it such that nothing in the comments is ignored.

working/0723-static-extensions/feature-specification-variant2.md
eernstg: I created https://github.com/dart-lang/language/issues/4049 to discuss this.

eernstg:
NB: At this time, working/0723-static-extensions/feature-specification-variant2.md is in a reasonably good shape whereas working/0723-static-extensions/feature-specification-variant1.md needs to be revised (in ways that are similar to the changes made to 'variant2' recently).

working/0723-static-extensions/feature-specification-variant1.md
lrhn: That's a bad use of the extension's type argument.
If we had generic constructors, it should be `Map.singleton<T>(T x) => {T: x}`.
Using an *extension* type parameter for that is a hack.

Not saying it won't work, or can't be useful, but it's also quite ugly.
Just use a static function. (Also, constructor should be `factory`.)

leafpetersen: > That's a (IMO) bad use of the extension's type argument.

Not sure what you're saying here.  Are you saying this should be an error?  I'm just trying to flesh out what the feature is.

working/0723-static-extensions/feature-specification-variant1.md
eernstg: Right, but that's true for every task which is loosely described as type inference. We need an algorithm (which is known to be terminating, etc.) in order to get an implementation, but at this point I wanted to specify the constraints that such an algorithm must satisfy.

I believe we should use an algorithm which is different from the one which is otherwise used in order to infer types in dart. That is, I don't think we can directly use something like `E_Map_bar` in https://github.com/dart-lang/language/issues/4053. I think we need to use equality constraints rather than subtype constraints, because we can't have something like `new List<num>.qux()` that returns a `List<int>` or a `List<Object>` (we _could_ specify that the static type of `new List<num>.qux()` is `List<Object>` and it could yield a `List<Object>`, but I think we shouldn't).

In other words, the type arguments for the extension will fully determine the type arguments for the class. If we already have the type arguments for the class and not for the extension then I believe we must find type arguments for the extension such that substitution into the on-type yields _exactly_ the type arguments for the class that we already have (say, because a developer wrote them explicitly).

Equality constraints would basically just have the structural steps, so it should be _a lot_ simpler than general type inference.

```dart
static extension E<X, Y> on Map<String, X> {
  factory Map.name(Y y, X Function(Y) fun) => {y.toString(): fun(y)};
}

void main() {
  var map = Map<String, double>.name(14, (i) => i + 0.5);
}
```

In this case we resolve the invocation of `Map.name` as an invocation of `E.Map.name`, say, because no other static extension declares a static member named `name` or a constructor named `Map.name`.

Then we generate constraints: `X == double`. Next, we infer `E<double, _>.C<String, double>.name` like

```dart
Map<String, double> E_Map_name<Y>(Y y, double Function(Y) fun) =>
    throw "Just for inference";

void main() {
  var map = E_Map_name(14, (i) => i + 0.5); // Infers `E_Map_name<int>(...)`.
}
```

This is the case where the actual type arguments to the class have been specified explicitly. A similar case arises when there is a context type that determines the value of type variables of `E`:

```dart
void main() {
  Map<Object, double> map = Map.name(14, (i) => i + 0.5);
}
```

In this case I'm proposing that we should use the context type. This binds `X` to `double`, which means that we have an instantiated constructor return type of `Map<String, double>`, and we get the value `int` for `Y` again.

working/0723-static-extensions/feature-specification-variant1.md
eernstg: Right, I think it makes sense to just remove the word 'explicit' from the section title.

working/0723-static-extensions/feature-specification-variant1.md
eernstg: Yep, added commentary.

working/0723-static-extensions/feature-specification-variant1.md
eernstg: This is the same algorithm that I'm talking about in the comment about in line 373.

working/0723-static-extensions/feature-specification-variant1.md
eernstg: Exactly. Revised the text to acknowledge this.

working/0723-static-extensions/feature-specification-variant1.md
eernstg: Agreed, this needs some work. ;-) I'll use 4053 to improve on it later. Let me try to see if I can respond to the case at this time as well:

```dart
extension E<S, T> on Map<int, T> {
  Map.bar(S x, T y) => {3 : y};
}

void main() {
  K map = Map.bar(1.5, "hello");
}
```

I think we need to resolve at first that `Map.bar` means `E.Map.bar`. There are different ways to do this, but I've aimed to specify that we accept the declaration in the class `Map` if it exists ("class always wins"), otherwise we find all static extensions declaring a constructor `Map.bar` and/or a static member of `Map` named `bar`. If there are both static members and constructors then it's an error. Assume that we have only constructors. If there are two or more static extensions declaring a constructor named `Map.dar` then it's an error (because we have no idea what the actual type arguments would be, and then we don't want to try to filter the extensions).

OK, so it's `E.Map.bar(1.5, "hello")`.

There will be some actual type arguments of `Map`, and the type of the expression will be `Map<...>` with those actual type arguments.

So we use the downward context `K` on the on-type of `E` (not on the entire signature, it is only the constructor return type that matters):

```dart
Map<int, T> E_Map_bar<S, T>(S x, T y) => throw "For inference only";

void main() {
  K map = E_Map_bar(1.5, "Hello");
}
```

This will bind `S` to `double` (there's no way `K` can influence that), and `T` must be some supertype of `String`, and we will allow the downward phase to choose the type argument "if `K` has an opinion" about that (`K == dynamic` doesn't, but `K == Map<Object, Object>` does).

eernstg (commented):
I made some changes to '...variant1.md' in response to review comments.

Note that I still need to go over '...variant1.md' and make a lot of changes that are similar to the ones that I performed on '...variant2.md'.

working/0723-static-extensions/feature-specification-variant1.md
eernstg: This discussion is now handled via https://github.com/dart-lang/language/issues/4053.

eernstg:
Hi @leafpetersen and @lrhn, cc-hi @dart-lang/language-team. I'd like to land this now. I think it's better to take the remaining discussions via Github issues, and I also think the relevant issues have already been created. Then we can change these documents as needed later on, when the issues have been processed. WDYT?

leafpetersen (approved):
I think we're at a point where the most productive thing is to land this as an interim state and move any remaining discussion to issues.  LGTM to land as WIP.

PR #3840, opened 2024-05-27
[wildcard-variables] Fix a typo in the super parameters example.

sgrekhov:
Missing `extends` added

cc @kallentu


kallentu (approved):
We'll need to make a change to update the spec on super anyways. (Which I'll get to soon).
In the meantime this is fine. Thanks for catching it.

PR #3845, opened 2024-05-28
[wildcard-variables] Allow super._.

kallentu:
Changed the section on `super._` to allow it and described a few cases.
Let me know if I got anything wrong or can write something better.

Bob's out, so if either/both of you can take a look, that'd be great!

### DIFF
--- a/working/wildcards/feature-specification.md
+++ b/working/wildcards/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: In-progress
 
-Version 1.4
+Version 1.5
 
 ## Motivation
 
@@ -317,37 +317,54 @@ class C {
 
 ### Super parameters
 
-An occurrence of `super._` as a declaration of a formal parameter in a
-constructor is a compile-time error. This error also occurs in the case
-where the super parameter has an explicitly declared type and/or default
-value.
+The positional formal parameter `super._` is still allowed in generative
+constructors.
 
-*`super._` is not an error everywhere: In a method body it could be an
-invocation of an inherited getter named `_`.*
+`super._` forwards the argument's value to the super constructor invocation.
 
 ```dart
-class B {
-  final _;
-  B(this._);
+class A {
+  final int _;
+  A(this._);
 }
 
-class C extends B {
-  C(super._); // Error.
+class B extends A {
+  B(super._); // OK.
 }
 ```
 
-*The desugared meaning of a super parameter includes a reference to the
-parameter in the initializer list of the enclosing constructor declaration,
-but such references are not possible when the parameter name is a
-wildcard.*
+Similar to `this._`, no parameter named `_` is bound. `_` can't be accessed in
+the initializer list.
 
-*It may seem inconvenient that `super._` "does not work" in the last
-example, but we do not wish to create exceptions about the ability to refer
-to a declaration whose name is a wildcard, just so we can support this
-particular usage. The advice would be to use a different name than `_` for
-the instance variable in `B` in the first place, or using a different name
-in `B` and possibly `// ignore` a lint that may warn about the names being
-different.*
+```dart
+class A {
+  final int _;
+  A(this._);
+}
+
+class B extends A {
+  final int k;
+  B(super._)
+    : k = _ { // Error.
+    print(_); // OK. Prints the inherited field.
+  }
+}
+```
+
+`super._` will forward each argument's value to the super constructor
+invocation, respectively. The wildcard super parameters will not collide.
+
+```dart
+class A {
+  final int x;
+  final int y;
+  A(this.x, this.y);
+}
+
+class B extends A {
+  B(super._, super._); // Ok.
+}
+```
 
 ### Extension types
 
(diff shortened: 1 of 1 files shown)


working/wildcards/feature-specification.md
```diff
@@ -317,37 +317,54 @@ class C {
 
 ### Super parameters
 
-An occurrence of `super._` as a declaration of a formal parameter in a
-constructor is a compile-time error. This error also occurs in the case
-where the super parameter has an explicitly declared type and/or default
-value.
+The positional formal parameter `super._` is still allowed in generative
```
lrhn, line 320: Pedantic: ... in non-redirecting generative constructors.

(Everybody knows what you mean, most people don't even remember that redirecting generative constructors exist.)

working/wildcards/feature-specification.md
```diff
@@ -317,37 +317,54 @@ class C {
...
+constructors.
 
-*`super._` is not an error everywhere: In a method body it could be an
-invocation of an inherited getter named `_`.*
+`super._` forwards the argument's value to the super constructor invocation.
```
lrhn, line 323: I prefer to start sentences with captial letters, even if it just means adding a "The " in front.
Maybe "The `super._` parameter`.

working/wildcards/feature-specification.md
```diff
@@ -317,37 +317,54 @@ class C {
...
+class B extends A {
+  B(super._); // OK.
 }
 ```
 
-*The desugared meaning of a super parameter includes a reference to the
-parameter in the initializer list of the enclosing constructor declaration,
-but such references are not possible when the parameter name is a
-wildcard.*
+Similar to `this._`, no parameter named `_` is bound. `_` can't be accessed in
```
lrhn, line 336: I'd say "no variable" or "no identifier" instead of "no parameter".

And I'd start the second sentence with a word instead of "\`_\`", maybe even:
> The value of the `super._` parameter can't be accessed ...

working/wildcards/feature-specification.md
```diff
@@ -317,37 +317,54 @@ class C {
...
-*`super._` is not an error everywhere: In a method body it could be an
-invocation of an inherited getter named `_`.*
+`super._` forwards the argument's value to the super constructor invocation.
 
 ```dart
-class B {
-  final _;
-  B(this._);
+class A {
+  final int _;
```
lrhn, line 327: Doesn't have to be named `_`, most parameters won't be. The `A(...)` parameter doesn't even have to be an initializing formal.

I'd take the third example an move it up here, and only have two examples:
* One correct example which shows two `super._`s (as many correctly used features as possible)
* One with an error trying to reference `_` in the initializer list.

Maybe (incorporating all comments below):
> A positional formal parameter of the form `super._` is still allowed in non-redirecting
> generative constructors. Such a parameter forwards the argument's value to the
> super constructor invocation.
>
> ```dart
> class A {
>   final int x, y;
>   A(this.x, this.y);
> }
> class B extends A {
>   final int _;
>   B(this._, super._, super._) {
>     print(_ + x + y); // Refers to `B._`, `A.x` and `A.y`.
>   }
> }
> ```
> Similarly to `this._`, and unlike for example `super.x`,
> a `super._` does not introduce any identifier into the scope of the initializer list.
> Because of that, multiple `super._` and `this._` parameters can occur in the same
> constructor without any name collision.
> 
> ```dart
> class A {
>   final int _, v;
>   A(this._, this.v);
> }
> 
> class B extends A {
>   B(super.x, super._)
>       : assert(x > 0), // OK, `x` in scope.
>         assert(_ >= 0) // Error: no `_` in scope.
>   {
>     print(_); // OK, means `this._` and refers to `A._`.
>   }
> }
> ```

(The three examples made this a little longer than I think it needs to be, combining the examples where possible helps with that.)

working/wildcards/feature-specification.md
```diff
@@ -317,37 +317,54 @@ class C {
...
-to a declaration whose name is a wildcard, just so we can support this
-particular usage. The advice would be to use a different name than `_` for
-the instance variable in `B` in the first place, or using a different name
-in `B` and possibly `// ignore` a lint that may warn about the names being
-different.*
+```dart
+class A {
+  final int _;
+  A(this._);
+}
```
lrhn, line 343: (Here the `_` is used, to show the "Prints inherited field". Could consider showing the difference between the parameter and the field more distinctly:
```dart
 class Capacity {
   final int _;
   Capacity(int value, int limit) : assert(value <= limit), _ = limit = value;
 }
  class B extends Capacity {
    B(super._, super.max)
        : assert(max >= 0),  // OK, `max` in scope.
          assert(_ >= 0) {  // Error, no `_` in scope.
      print(_); // prints value of `A._`
      print(max); // Error, no `max` in scope.
    }
  }
```

working/wildcards/feature-specification.md
```diff
@@ -317,37 +317,54 @@ class C {
...
+class A {
+  final int x;
+  final int y;
+  A(this.x, this.y);
+}
+
+class B extends A {
+  B(super._, super._); // Ok.
+}
+```
```
lrhn, line 367: Yep, also a good example, but I'd use as the first example, and only have two:
* One correct example which shows two `super._`s, and a superclass with fields not named `_`.
* One with an error trying to reference `_` in the initializer list.

This examle makes the first example feel redundant.

lrhn (approved):
Only style and phrasing nits, and suggestions on the examples.
(Aka: I basically restructured it all, while keeping the message. The message is fine.)

working/wildcards/feature-specification.md
kallentu: Looks great. I've taken this :)

kallentu:
Thanks for the suggestions! I've taken the bulk of what you suggested.

Issue #2, opened 2018-06-29
Digit separators in number literals.


lrhn:
_This is currently under implementation: [implementation issue](https://github.com/dart-lang/sdk/issues/56188), [feature specification](https://github.com/dart-lang/language/blob/main/accepted/future-releases/digit-separators/feature-specification.md)._

---

Solution to #1.

To make long number literals more readable, allow authors to inject [digit group separators](https://en.wikipedia.org/wiki/Decimal_separator#Digit_grouping) inside numbers.
Examples with different possible separators:
```text
100 000 000 000 000 000 000  // space 
100,000,000,000,000,000,000  // comma
100.000.000.000.000.000.000  // period
100'000'000'000'000'000'000  // apostrophe (C++)
100_000_000_000_000_000_000  // underscore (many programming languages).
```
The syntax must work even with just a single separator, so it can't be anything that can already validly seperate two expressions (excludes all infix operators and comma) and should already be part of a number literal (excludes decimal point).
So, the comma and decimal point are probably never going to work, even if they are already the standard "thousands separator" in text in different parts of the world.  

Space separation is dangerous because it's hard to see whether it's just space, or it's an accidental tab character. If we allow spacing, should we allow arbitrary whitespace, including line terminators? If so, then this suddenly become quite dangerous. Forget a comma at the end of a line in a multiline list, and two adjacent integers are automatically combined (we already have that problem with strings). So, probably not a good choice, even if it is the preferred formatting for print text.

The apostrope is also the string single-quote character. We don't currently allow adjacent numbers and strings, but if we ever do, then this syntax becomes ambiguous. It's still possible (we disambiguate by assuming it's a digit separator). It is currently used by C++ 14 as a digit group separator, so it is definitely possible.

That leaves underscore, which could be the start of an identifier. Currently `100_000` would be tokenized as "integer literal 100" followed by "identifier _000". However, users would never write an identifier adjacent to another token that contains identifier-valid characters (unlike strings, which have clear delimiters that do not occur anywher else), so this is unlikely to happen in practice. Underscore is already used by a large number of programming languages including Java, Swift, and Python.

We also want to allow multiple separators for higher-level grouping, e.g.,:
```dart
100__000_000_000__000_000_000
```
For this purpose, the underscore extends gracefully. So does space, but has the disadvantage that it collapses when inserted into HTML, whereas `''` looks odd.

[...]

PR #3846, opened 2024-05-29
Add digit-separators specification v1.0

srawlins:
Add a specification for the digit-separators feature, as specified by @lrhn at https://github.com/dart-lang/language/issues/2 and at https://github.com/dart-lang/language/issues/2#issuecomment-2122769878. I add myself as an author as I added a few other sections.

---

<details>

</details>


srawlins:
@stereotype441 has mentioned to the analyzer team that he desires to keep better track of "possible lint rules" and "possible quick fixes" that would support a given language feature. To paraphrase, I think Paul said that these are often brought up in language team meetings, when discussing how developers could be kept on the happy path, or how to support a migration to a language feature.

To that end, I've written _possible_ new lint rules and quick fixes in this specification. But if this inspires frowning, if the specification MD file is not considered a good place for such commentary, I can remove those sections, and just open separate issues against analyzer, and not cross-link in this specification.

working/digit-separators/feature-specification.md
```diff
@@ -0,0 +1,192 @@
...
+var four = 4_0000_000;
+```
+
+If a developer uses the Dart formatter to format their code, they cannot try to
+vertically align the numbers with whitespace (extra space characters are
+removed by the formatter). So we could offer a lint rule to only place
+separators every three digits of a decimal number. Also possibly a similar rule
+for hexadecimal numbers. If a developer ever uses digit separators for a
+different purpose (as in separating the digits of a phone number), the rule may
+not prove useful.
```
lrhn, line 146: Another lint option could be "consistent digit separators", which triggers if the digit groups do not have the same size (except the most significant I one, which can be shorter).
If there are any `__` separators, the number of `_` separated groups between them should also be the same, and repeatedly for higher numbers of `_`s.

working/digit-separators/feature-specification.md
```diff
@@ -0,0 +1,192 @@
...
+feature.
+
+#### Unexpected underscores
+
+With the digit-separators feature, separators can be added between _digits_ of
+a number literal, but nowhere else. In most error cases, the unexpected
+underscore can be detected as such, and we can offer quick fixes to remove
+unexpected errors (for example, `100_`, `100_e.2`, `100._00`). In a few cases,
+the intention is not as straightforward, such as `100._100`, where `_100` can
+be a legal name of an extension member.
```
lrhn, line 161: But it's also possible to check whether it's an extension (in the same library, since it's a private name), and when it most likely isn't, the fix can be recommended.

working/digit-separators/feature-specification.md
```diff
@@ -0,0 +1,192 @@
...
+
+The apostrope is also the string single-quote character. We don't currently
+allow adjacent numbers and strings, but if we ever do, then this syntax becomes
+ambiguous. It's still possible (we disambiguate by assuming it's a digit
+separator). It is currently used by C++ 14 as a digit group separator, so it is
+definitely possible.
+
+That leaves underscore, which could be the start of an identifier. Currently
+`100_000` would be tokenized as "integer literal 100" followed by "identifier
+_000". However, users would never write an identifier adjacent to another token
```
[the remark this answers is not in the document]
srawlins: Done

working/digit-separators/feature-specification.md
srawlins: I like it, added!

working/digit-separators/feature-specification.md
srawlins: Added

working/digit-separators/feature-specification.md
```diff
@@ -0,0 +1,199 @@
...
+vertically align the numbers with whitespace (extra space characters are
+removed by the formatter). So we could offer a lint rule to only place
+separators every three digits of a decimal number. Also possibly a similar rule
+for hexadecimal numbers. If a developer ever uses digit separators for a
+different purpose (as in separating the digits of a phone number), the rule may
+not prove useful.
+
+A separate lint rule could encourage _consistent_ digit separators, which
+triggers if the digit groups do not have the same size (except the most
+significant I one, which can be shorter). If there are any `__` separators, the
```
mateusfccp, line 150: Is this `I` intentional?

```suggestion
significant one, which can be shorter). If there are any `__` separators, the
```

working/digit-separators/feature-specification.md
```diff
@@ -0,0 +1,199 @@
...
+unexpected errors (for example, `100_`, `100_e.2`, `100._00`). In a few cases,
+the intention is not as straightforward, such as `100._100`, where `_100` can
+be a legal name of an extension member (though the presense of such a private
+extension member can be detected).
+
+#### Unexpected commas
+
+The only legal digit separator that is introduced with this feature is the
+underscore character. If a developer attempts to use another character, for
+example commas, as a separator, we may be able to detect this, and offer a
```
mateusfccp, line 174: Wouldn't this fall into the same problem that we have when we consider `,` as a separating digit? How can we be sure that the user intended to use `,` as a separating digit instead of two expressions separated by a comma?

srawlins: We can't. Many automated fixes are offered with the understanding that they are not the only fix, and may not be the desired fix.

I personally doubt we will offer this fix. But there might be some cases, like if the user pastes 100 lines of:

```dart
var x1 = 1,000,000;
var x2 = 2,000,000;
var x3 = 3,000,000;
```

into their editor, from another source text, we _could_ offer a fix to convert the commas, because it looks more likely that they meant to write `x1 = 1_000_000`, since  `var x1 = 1,` cannot be followed by an int literal (it can only be followed by an identifier.

working/digit-separators/feature-specification.md
srawlins: Good catch. Done.

srawlins: Fixed, thanks

srawlins:
This is ready for review; dunno how you assign reviewers, @dart-lang/language-team

working/digit-separators/feature-specification.md
```diff
@@ -0,0 +1,199 @@
...
+There are some possible new automated fix ("quick fix") considerations, but
+none of these are considered vital to the usability or general success of the
+feature.
+
+#### Unexpected underscores
+
+With the digit-separators feature, separators can be added between _digits_ of
+a number literal, but nowhere else. In most error cases, the unexpected
+underscore can be detected as such, and we can offer quick fixes to remove
+unexpected errors (for example, `100_`, `100_e.2`, `100._00`). In a few cases,
```
srawlins, line 165: Actually you can't put period after e. This should be `100_e1.2`

lrhn: Not much better. We can fix `100_e1` to `100e1`, but that `.2` is still going to be a `.` + NUMBER token sequence. You can't have a decimal point after the `e` in a flowing point number. (So the `.2` is possibly spurious to the point being made.)

These can be corrected by removing the `_`: `100_e2`, `100_e+2` or `100_e-2`.

For `100._00`, that's only an error if `int` has no `_00` (extension) member. It probably doesn't, but we should check first.

working/digit-separators/feature-specification.md
lrhn: If the same person pasted 100 lines of `var x = foo(1,000,000);` we'd be having a harder time helping them.
(Not hat we couldn't *suggest* `1_000_000`, especially if we can see that `foo` takes only one argument, but it was not grammatically wrong to begin with.)

One thing that worries me is whether we can do the same fixes *after* the formatter has run.
If not, we may have a problem helping people who auto-format often.
If the code doesn't parse, it likely won't format either, but if it parses, and you get to `var x1 = foo(1, 000, 000);`, it would be sad if it's too late to help the user.

leafpetersen (approved):
Great to see this moving forward, thanks!  Should probably get a sign off from Lasse who's thought about this a lot more, but FWIW LGTM!

lrhn (approved):

PR #3856, opened 2024-05-30
[macros] Fix for analyzer presubmit.

davidmorgan:


davidmorgan:
Ah, I see this is not sufficient ... the analysis won't work unless I run a "pub get" in _all_ the packages.

@jakemac53 I could add that to `.github/workflows/dart.yml`? ... it seems like quite a lot to add to that file, though.

I have a `presubmit` script that does all the pub gets and analysis, possibly adding running that as a step would be a way to go?

I'm not sure if we actually want to analyze this code automatically since it's likely to move somewhere else as soon as there is someone other than me hacking on it.

Please let me know what you think, happy to fix any way that works :) thanks!

jakemac53:
It should run a `pub get` in every package, one way or another, yes. We could cache the pub cache dir to speed that up.

davidmorgan:
Okay, significantly less red now :) PTAL.

jakemac53 (approved):
The remaining failures looks pre-existing, feel free to land on red and I can follow up, or you can push a fix for that and then land.

davidmorgan:
End of day for me so I'll merge and leave it to you :) thanks!

Issue #3807, opened 2024-05-15
[Wildcard Variables][spec] Allow unnamed optional parameters to not have default value.


lrhn:
With "wildcard-variables", a parameter can have *no* name.
For example `void foo(int _) { ... }` does not introduce a name in the parameter scope, the parameter value is inaccessible.

Or stated differently: The _parameter_ does not introduce a _variable_.

The purpose of optional parameter default values is to ensure that a parameter variable has a value when it's read, even if no argument value was passed. A parameter which cannot be read or used in any, one where there is no variable at all, way does not *need* a default value to initialize a variable.

Proposal: It is not a compile-time error for a optional non-initializing non-super parameter to not have a default value, even if its type is not nullable.

The existing clause is:
> It is an error if an optional parameter (named or otherwise) with no default
> value has a potentially non-nullable type **except** in the parameter list of an
> abstract method declaration.

(Or of a forwarding factory constructor, they also cannot declare default values. And come augmentations, it will only apply to the fully augmented declaration, not the individual steps.)

For wildcard variables only, this should be changed to:

> It is an error if an optional parameter (named or otherwise) with no default
> value has a potentially non-nullable type **except** in the parameter list of an
> abstract method declaration or a forwarding factory constructor, or if the 
> declared parameter name is `_` and the parameter is not an initializing 
> formal or super parameter.

(If we have a better way of phrasing that a declaration is non-binding, we can use that instead of "the declared ... name is `_`".)

We have to exclude initializing formals, `this._`, and super parameters, `super._` ~~if~~when we allow those, because the value of that *argument* is used for something, even if it cannot be referenced.
(If we make it possible to forward *not having an argument*, then `super._` can safely be forwarded to another optional positional parameter without having a default value, it will just forward the absence of an argument and let the super-constructor use its default value. Maybe we should just do that anyway! Sub-proposal:
> If an optional super-parameter corresponds to an optional parameter of the super-constructor, and no argument is given to
> the super-paramater, then no argument is passed to the super-constructor either. It's a compile-time error if such an
> optional super-parameter with an optional target has no default value, is potentially non-nullable, and the variable it
> introduces in the parameter scope is referenced. Otherwise the variable will use the default value if no argument was
> passed, even if no value is forwarded to the super constructor.

And a `_`-named parameter is never referenced, so that just works. As long as we can only omit trailing positional parameters, and we cannot combine explicit positional arguments to the super-constructor with positional super-parameters, the forwarding will also work correctly.
Or something.)

[...]

PR #3880, opened 2024-06-04
[wildcard-variables] Add section on unnamed optional parameters and their default values.

kallentu:
Added a section for unnamed optional parameters and them not needing default values.
Fixes: https://github.com/dart-lang/language/issues/3807


lrhn (approved):
LGTM!

Issue #3872, opened 2024-06-03
Pick a supported serialization format for host<->macro communication


davidmorgan:
The [dart_model exploration](https://github.com/dart-lang/language/tree/main/working/macros/dart_model) currently uses JSON but there is no particular reason to think that's a good choice :)

Even if we don't follow the `dart_model` direction, it seems likely we want to formalize the serialization format so that we can have stable serialization. Some well-supported system that allows specification of a schema is probably needed so we can actually have guarantees of compatibility.

It looks like macros often do only trivial computation and so they are going to spend most of their run time deserializing then serializing. The best way to optimize this is to do less work, to send less data; but even then it might be beneficial to also have the fastest possible serialization and deserialization.

[Protocol buffers](https://protobuf.dev/) are an obvious possibility, [flat buffers](https://flatbuffers.dev/) might be more interesting since they are explicitly for cases where deserialization/serialization are an important overhead. If we follow the `dart_model` direction then capability to quickly compute deltas and apply deltas will also be important.

Sending this straight to the backlog since the precise format doesn't matter too much yet.

@jakemac53 @scheglov FYI :)

rrousselGit:
Would this be possibly related to macros sending information to the analyzer ; which the analyzer can then use to emit custom diagonistics?

davidmorgan:
Macros are fundamentally built on serialization: when you call async methods on the macro API, the macro communicates with the host via serialized data. The hosts we have today are the analyzer and the CFE.

davidmorgan:
I looked into this a little today; I don't like the flatbuffers schema format; one possibility that comes to mind is to use a JSON schema coupled with some more efficient way of serializing JSON, possibly we can do something like flatbuffers where data is accumulated into a buffer that is immediately ready to send.

jakemac53:
I don't know anything about flat buffers, do you have a tldr? How does it compare with the adhoc format we have today?

jakemac53:
Also, the idea of using FFI to enable multithreading + shared object memory with the host has been brought up previously. I have no idea what that concretely means or what it would look like though. @mraleph I heard maybe you would be a good reference here?

jakemac53:
It sounds to me from some very quick reading that using FFI to write to a shared buffer (where?), and FlatBuffer to actually control those buffer formats, with Dart wrappers reading from those buffers directly, could be interesting.

davidmorgan:
[...]

Issue #3706, opened 2024-04-15
A more intentional and query-like API for macro source introspection


lrhn:
The current API for introspecting the source of the code that the macro is generating augmentations for, is fairly coarse-grained.

The [example](https://github.com/jakemac53/macros_example) requests all methods from a class, then manually goes through them to check if there is a method named `toString`.
That hides the actual data used by the macro from the macro runtime, it cannot see which methods are actually needed, and which properties of the methods are needed.

I suggest instead having an API where you'd specifically ask for "instance methods named `toString`", and get only the zero or one result, and the object returned would only contain an ID, the name, and that it's an instance method. It wouldn't contain parameter or return types, if you don't *ask* for those. Or there might even be a function to just answer whether a match *exists*, returning just a `bool`.

The request would be sent as a _query object_ to the macro runtime.

(That behaves something like an SQL query of `SELECT id, name, static FROM members WHERE containingClassId = classId, name = "toString", kind = "method", static = false `, meaning that you don't necessarily get fully populated objects. You'd have to be able to ask whether a property is available, so you can do `if (method.hasStatic)` to check if the "is this method static" boolean is even available, and it'll throw if you do `mathod.isStatic` and the information isn't there. You're expected to have asked for the information you need.)

The potential benefit of using such an API is that the query objects are reusable and replayable, and the runtime system can see precisely which information the macro code has requested, and presumably depends on.
Also, the query can be much more targeted, and won't provide results that the macro won't be able to understand, because any new kind of declaration would have to be requested. There is no "give me everything" request that can give you results of types that you don't know about.

If the program *changes*, like dynamic reload or just during development, the runtime system can potentially check whether any answer it would give to a macro has also changed. If not, there is no need to re-run the macro, since it would have precisely the same behavior the second time. (That requires some way to keep IDs consistent across changes.)

A possible API could be:
```dart
  ClassDeclaration clazz = ...;
  if (builder.queryAny(clazz.members(MethodFilter(name: "toString", static: false)))) {
    // There is a non-static method named `toString`
    ...
  }
```
The query methods could be something like:
```dart
  T? queryFirst<T extends SourceEntity>(Query<T> query);
  List<T> queryAll<T extends SourceEntity>(Query<T> query);  
  bool queryAny(Query<SourceEntity> query);
```
and the way to query members of a class is:
```dart
Query<T> members<T extends ClassMember>(MemberFilter<T> filter) => filter._query
```
with a 
```dart
[...]

PR #3893, opened 2024-06-10
[macros] Add dart_model proposal.

davidmorgan:
For #3872 #3706

### DIFF
--- a/working/macros/dart-model.md
+++ b/working/macros/dart-model.md
@@ -1,0 +1,131 @@
+# Dart Model
+
+Author: David Morgan
+
+Status: **Work In Progress**
+
+## Introduction
+
+A package for describing Dart code, augmentations to it, and queries about it.
+
+Motivated by its use in the
+[macros feature](https://github.com/dart-lang/language/blob/main/working/macros/feature-specification.md)
+but not tied to it.
+
+## Concepts
+
+```mermaid
+---
+title: package:dart_model
+---
+flowchart TD
+  Model:::model
+  Query:::query
+  Delta:::delta
+  Augmentation:::augmentation
+  Service:::service
+
+  classDef model fill:#ffd6a5,stroke-width:0px
+  classDef query fill:#fdffb6,stroke-width:0px
+  classDef delta fill:#caffbf,stroke-width:0px
+  classDef augmentation fill:#9bf6ff,stroke-width:0px
+  classDef service fill:#bdb2ff,stroke-width:0px
+```
+
+### Data Types
+
+`Model`, `Query`, `Delta` and `Augmentation` are serializable data types, see details [below](#serialization-and-schema).
+
+### Model
+
+A description of a corpus of Dart code with the following properties:
+
+- It is a **partial** model.
+  - It can omit files, parts of files, and attributes of parts of files, depending on what was requested, what was known about the corpus when the request was made, and considerations about what a particular viewer of the code is _allowed_ to see.
+- It is **sufficient input for codegen or macros**.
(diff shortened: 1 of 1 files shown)


davidmorgan:
Hi @jakemac53, an attempt to put an actual proposal in writing re: dart_model, nothing really new but including the idea that we go with JSON schema + JsonBuffer per my experiments. Lots to discuss I'm sure :)

working/macros/dart-model.md
```diff
@@ -0,0 +1,131 @@
...
+
+`package:dart_model` provides a reference `Service` implementation that is backed by a `Model`. More interesting `Service` implementations for generators and macros will be backed by the analyzer or the CFE.
+
+## Serialization and Schema
+
+`Model`, `Query`, `Delta` and `Augmentation` are each described by a JSON schema published in the same package, as a plain text file and also embedded in Dart code for use by tools.
+
+Each schema is available in two versions: a "strict" schema that only matches what the current package produces, and a "loose" schema that allows for future changes.
+
+Schemas are versioned using semver. Schemas are updated in place for compatible changes, and as new files for major version (incompatible) releases.
```
jakemac53, line 121: How exactly does this interact with the strict vs loose formats?

This package will also be used as a part of SDK binaries, which will not take into account the current users package version, how does that affect things?

working/macros/dart-model.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Performance
+
+Data types can be serialized to and from JSON per their schemas.
+
+A binary serialization format is available for performance. When this is used, values are accumulated directly into a binary buffer, so there is no "serialize" step. The received of the value can use values directly from the buffer, so there is no "deserialize" step.
```
jakemac53, line 127: Does this binary format use String keys, a specified ordering, or something else?

jakemac53:
As a high level doc of the concept this LGTM.

Would be great to build on this and discuss how exactly it should be used for macros though, how/when exactly diffs should be pushed, how/if this interacts with phases (or rounds), how references to specific objects should be sent back to the compiler, etc.

jakemac53 (approved):

davidmorgan:
Thanks! Yes, a description of the higher level protocol on top is something we need; as well as next steps for bootstrapping some kind of package:dart_model within the current setup. Let's chat tomorrow :)

working/macros/dart-model.md
davidmorgan: Good questions.

I'm not currently envisaging _using_ the schemas in any online way; so far I've only considered using them for testing--check that we think we comply with the schema--and for codegen, generate classes that match the schema. So this was mostly a guess about what would be useful.

I already found having the schemas embedded in Dart useful in [testing](https://github.com/dart-lang/language/blob/218da8a51f0bad50d2ca67937f4b572bafe9549b/working/macros/dart_model/dart_model_analyzer_service/test/golden_test.dart), so that's why I mention it.

Re: the client package version, I guess we send it on the wire at some point.

working/macros/dart-model.md
davidmorgan: The [POC](https://github.com/dart-lang/language/blob/218da8a51f0bad50d2ca67937f4b572bafe9549b/working/macros/dart_model/testing/json_benchmark/lib/json_buffer.dart) I hacked together is pretty much a byte buffer representation for nested `Map<String, Object?>`, I like this direction, it should mean we can implement something once and use it without changes as the JSON evolves. It's not a design yet though, just "I think we can model as JSON and still do something fast underneath".

PR #3902, opened 2024-06-13
Specify dynamic Object member invocations

eernstg:
This PR modifies the language specification by adding references to section 'Type dynamic' where the static typing of members of `Object` which is used with receivers of type dynamic at the locations where that static typing is used, that is: 'Ordinary Invocation' and 'Getter access and method extraction'.


stereotype441 (approved):
This is super helpful, thank you!

eernstg:
Thanks!

PR #3918, opened 2024-06-17
Consistently use terminology "instance implementing" and "dynamic type".

stereotype441:
In the code review discussion
https://github.com/dart-lang/language/pull/3850#discussion_r1635527576, we agreed to use the terminology "instance satisfying type T" to mean "instance whose runtime type is T or a subtype thereof", and I made the necessary changes to
`resources/type-system/inference.md`. However, after digging through `dartLangSpec.tex`, it appears that the terminology we were already using in the spec for this concept is "instance implementing type T". This commit modifies `resources/type-system/inference.md` to match the spec terminology.

It also fixes two places in the spec where we were incorrectly saying "instance of type T", but we meant "instance implementing type T".

Furthermore, the spec consistently uses the term "dynamic type of `v`" for the concept that we informally refer to as the "runtime type of `v`", so this commit updates `resources/type-system/inference.md` to follow the spec convention.

---

<details>

</details>

### DIFF
--- a/resources/type-system/inference.md
+++ b/resources/type-system/inference.md
@@ -1026,15 +1026,15 @@ inference depends on the context `K`, as well as the flow analysis state._
 
 A property of expression inference, known as _soundness_, is that when an
 elaborated expression is executed, it is guaranteed either to diverge, throw an
-exception, or evaluate to a value that is an _instance satisfying_ its static
-type. _Instance satisfying_ is defined as follows: a value `v` is an instance
-satisfying type `T` iff the runtime type of `v` is a subtype of the _extension
+exception, or evaluate to a value that is an _instance implementing_ its static
+type. _Instance implementing_ is defined as follows: a value `v` is an instance
+implementing type `T` iff the dynamic type of `v` is a subtype of the _extension
 type erasure_ of `T`. _So, for example, every value is considered an instance
-satisfying type `dynamic`, and all values except `null` are considered an
-instance satisfying type `Object`._
+implementing type `dynamic`, and all values except `null` are considered an
+instance implementing type `Object`._
 
 _This stands in contrast to the notion of "instance of"; a value `v` is an
-instance of a type `T` only if the runtime type of `v` is __precisely__ `T`._
+instance of a type `T` only if the dynamic type of `v` is __precisely__ `T`._
 
 _The type inference rules below include informal sketches of a proof that
 soundness holds for each expression type. These are non-normative, so they are
@@ -1050,10 +1050,10 @@ succintly, the syntax of Dart is extended to allow the following forms:
 
   - Evaluate `m_1`, and let `v` denote the result.
 
-  - If `v` is an instance satisfying type `Future<T>`, then
+  - If `v` is an instance implementing type `Future<T>`, then
     `@AWAIT_WITH_TYPE_CHECK<T>(m_1)` evaluates to `v`.
 
-  - Otherwise, let `u` be a future satisfying type `Future<T>` that will
+  - Otherwise, let `u` be a future implementing type `Future<T>` that will
     complete to the value `v` at some later point. Then,
     `@AWAIT_WITH_TYPE_CHECK<T>(m_1)` evaluates to `u`.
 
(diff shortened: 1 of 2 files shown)


lrhn:
> uses the term "dynamic type of `v`" for the concept that we informally refer to as the "runtime type of `v`"

(I really want to use "runtime type of `v`" for that. The word "dynamic" should be reserved for things relating to the type `dynamic` and dynamic member invocations, and "runtime' should be used for things that happen at runtime. But that doesn't have to be done now.)

resources/type-system/inference.md
```diff
@@ -1026,15 +1026,15 @@ inference depends on the context `K`, as well as the flow analysis state._
 
 A property of expression inference, known as _soundness_, is that when an
 elaborated expression is executed, it is guaranteed either to diverge, throw an
-exception, or evaluate to a value that is an _instance satisfying_ its static
-type. _Instance satisfying_ is defined as follows: a value `v` is an instance
-satisfying type `T` iff the runtime type of `v` is a subtype of the _extension
+exception, or evaluate to a value that is an _instance implementing_ its static
+type. _Instance implementing_ is defined as follows: a value `v` is an instance
```
lrhn, line 1030: I think (checking ... still think it!) that the spec uses "implements `S`" to mean "has `S` as super-interface" (or really: "is `S`-bounded", which means having `S` as super-interface or is a type variable with a bound which implements `S` (recurse as needed). Maybe also a promoted type variable whose intersection type implements `S`.)

That's different from "having as supertype" if the type is generic, so `Future<int>.value(1)` is an object with runtime type `_Future<int>` which has `Future<int>` as super-interface. It implements `Future<int>`. It does not implement `Future<Object>`, it just has that as a supertype.

So, `(Future<int>.value(1) as Future<Object>)` does not satisfy that its value's runtime type implements its static type.

For soundness, if an expression evaluates to a value, the runtime type of that value must be a _subtype_ of the (extension-type-erased) static type of the expression. Using "implements" is likely wrong.


... there it was, in the "Superinterfaces" section:
> ```latex
> When we say that a type $S$
> \IndexCustom{implements}{type!implements a type}
> another type $T$,
> this means that $T$ is a superinterface of $S$,
> or $S$ is $S_0$ bounded for some type $S_0$
> (\ref{bindingActualsToFormals}),
> and $T$ is a superinterface of $S_0$.
> ... snip implementing raw types ...
> 
> \commentary{%
> Note that this is not the same as being a subtype.
> For instance, \code{List<int>} implements \code{Iterable<int>},
> but it does not implement \code{Iterable<num>}.
> Similarly, \code{List<int>} implements \code{Iterable}.
> Also, note that when $S$ implements $T$
> where $T$ is not a subtype of \code{Null},
> $S$ cannot be a subtype of \code{Null}.%
> }
```

File: resources/type-system/inference.md [hunk 2]
```diff
@@ -1026,15 +1026,15 @@ inference depends on the context `K`, as well as the flow analysis state._
...
+implementing type `T` iff the dynamic type of `v` is a subtype of the _extension
 type erasure_ of `T`. _So, for example, every value is considered an instance
-satisfying type `dynamic`, and all values except `null` are considered an
-instance satisfying type `Object`._
+implementing type `dynamic`, and all values except `null` are considered an
```
lrhn, line 1033: (Since `dynamic` is not an interface, you can't have it as super-interface, so you can't implement it. But everything is a subtype of `dynamic` because it's a top type.)

specification/dartLangSpec.tex
```diff
@@ -7759,7 +7759,7 @@ \subsection{Super-Bounded Types}
 all types of the form \code{C<$S$>}
 (noting that all types must be regular-bounded
 when we do not have the notion of super-bounded types).
-So if we wish to allow a variable to hold any instance ``of type \code{C}''
+So if we wish to allow a variable to hold any instance ``implementing type \code{C}''
```
lrhn, line 7762: So this should be
"any instance whose dynamic type is a subtype of \\code{C}"

specification/dartLangSpec.tex
```diff
@@ -22495,7 +22495,7 @@ \subsection{Type FutureOr}
 
 \rationale{%
 The \code{FutureOr<$T$>} type represents a case where an object can be
-either an instance of the type $T$
+either an instance implementing the type $T$
```
lrhn, line 22498: "an instance whose dynamic type is a subtype of $T$"

resources/type-system/inference.md
stereotype441: Ok, I see what you mean. Thanks for the explanation. I'm going to go ahead and abandon this PR and stick with the terminology "instance satisfying type T" in `resources/type-system/inference.md`.

PR #3920, opened 2024-06-18
Simplify reasoning about await soundness using a lemma about flatten.

stereotype441:
The lemma is that for all types `T`, `T <: FutureOr<flatten(T)>`. The lemma was claimed (without proof) in `dartLangSpec.tex`, so I've added a (non-normative) proof of it to the spec, right after the claim.

This made it possible to significantly simplify the reasoning around the soundness of `await` in `inference.md`.

Thanks to @lrhn for the core idea behind this.

---

<details>

</details>

### DIFF
--- a/resources/type-system/inference.md
+++ b/resources/type-system/inference.md
@@ -1060,13 +1060,15 @@ succintly, the syntax of Dart is extended to allow the following forms:
     instance satisfying either `Future<T>` or `T`. We consider the two cases
     below._
 
-  - If `v` is an instance satisfying type `Future<T>`, then
-    `@AWAIT_WITH_TYPE_CHECK(m_1)` evaluates to `v`.
+  - If `v` is an instance satisfying type `Future<T>`, then let
+    `@AWAIT_WITH_TYPE_CHECK(m_1)` evaluate to `v`.
 
-  - Otherwise, let `u` be a future satisfying type `Future<T>` that will
-    complete to the value `v` at some later point. Then,
-    `@AWAIT_WITH_TYPE_CHECK(m_1)` evaluates to `u`. _Such a future is guaranteed
-    to (soundly) exist, because in this case `v` is an instance satisfying `T`._
+  - Otherwise, let `@AWAIT_WITH_TYPE_CHECK(m_1)` evaluate to a future satisfying
+    type `Future<T>` that will complete to the value `v` at some later
+    point. _Such a future could, for instance, be created by executing
+    `Future<T>.value(v)` (this is sound because in this case `v` is an instance
+    satisfying `T`). It also could be created using a more efficient
+    implementation-specific mechanism._
 
   - _Note that these two cases in the abstract correspond concretely to a type
     check in the implementation; this where the name `@AWAIT_WITH_TYPE_CHECK`
--- a/resources/type-system/inference.md
+++ b/resources/type-system/inference.md
@@ -1068,6 +1068,10 @@ succintly, the syntax of Dart is extended to allow the following forms:
     `@AWAIT_WITH_TYPE_CHECK(m_1)` evaluates to `u`. _Such a future is guaranteed
     to (soundly) exist, because in this case `v` is an instance satisfying `T`._
 
+  - _Note that these two cases in the abstract correspond concretely to a type
+    check in the implementation; this where the name `@AWAIT_WITH_TYPE_CHECK`
+    comes from._
+
 - `@CONCAT(m_1, m_2, ..., m_n)`, where each `m_i` is an elaborated expression
   whose static type is a subtype of `String`, represents the operation of
   evaluating each `m_i` in sequence and then concatenating the results into a
--- a/resources/type-system/inference.md
+++ b/resources/type-system/inference.md
@@ -1046,22 +1046,27 @@ The elaboration process sometimes introduces new operations that are not easily
 expressible using the syntax of Dart. To allow these operations to be specified
 succintly, the syntax of Dart is extended to allow the following forms:
 
-- `@AWAIT_WITH_TYPE_CHECK<T>(m_1)`, represents the following operation:
+- `@AWAIT_WITH_TYPE_CHECK(m_1)` represents the following operation:
 
-  - Evaluate `m_1`, and let `v` denote the result.
+  - Let `T_1` be the static type of `m_1`.
+
+  - Let `T` be `flatten(T_1)`.
+
+  - Evaluate `m_1`, and let `v` denote the result. _Note that since `m_1` has
+    static type `T_1`, by soundness, `v` must be an instance satisfying
+    `T_1`. Since `T_1 <: FutureOr<flatten(T_1)>` for all `T_1`, it follows that
+    `v` must be an instance satisfying `FutureOr<flatten(T_1)>`, or
+    equivalently, satisfying `FutureOr<T>`. Therefore, `v` must either be an
+    instance satisfying either `Future<T>` or `T`. We consider the two cases
+    below._
 
   - If `v` is an instance satisfying type `Future<T>`, then
-    `@AWAIT_WITH_TYPE_CHECK<T>(m_1)` evaluates to `v`.
+    `@AWAIT_WITH_TYPE_CHECK(m_1)` evaluates to `v`.
 
   - Otherwise, let `u` be a future satisfying type `Future<T>` that will
     complete to the value `v` at some later point. Then,
-    `@AWAIT_WITH_TYPE_CHECK<T>(m_1)` evaluates to `u`.
-
-    - _TODO(paulberry): explain why such a future is guaranteed to (soundly)
-      exist, by stipulating that if `T_1` is the static type of `m_1`, then `T`
-      must be `flatten(T_1)`, and then proving a lemma that `T_1 <:
-      FutureOr<flatten(T_1)>`; therefore `T_1 <: FutureOr<T>`, so in this
-      "otherwise" case, `v` must be an instance satisfying `T`._
+    `@AWAIT_WITH_TYPE_CHECK(m_1)` evaluates to `u`. _Such a future is guaranteed
+    to (soundly) exist, because in this case `v` is an instance satisfying `T`._
 
 - `@CONCAT(m_1, m_2, ..., m_n)`, where each `m_i` is an elaborated expression
   whose static type is a subtype of `String`, represents the operation of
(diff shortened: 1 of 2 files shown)


stereotype441:
CC @leafpetersen and @fishythefish

specification/dartLangSpec.tex
```diff
@@ -11829,7 +11829,40 @@ \subsection{Function Expressions}
 
 \rationale{%
 This definition guarantees that for any type $T$,
-\code{$T <:$ FutureOr<$\flatten{T}$>}.%
+\code{$T <:$ FutureOr<$\flatten{T}$>}. The proof is by induction on `T`:
```
lrhn, line 11832: Maybe say "on the structure of `T`".
Not that there are any other obvious thing that the type is inductive in, but being explicit shouldn't hurt. 
(And type structure is inductive the way we've defined types).

specification/dartLangSpec.tex
```diff
@@ -11829,7 +11829,40 @@ \subsection{Function Expressions}
...
+
+\begin{itemize}
+\item If $T$ is \code{$X$\,\&\,$S$} then
+  \begin{itemize}
+  \item if $S$ derives a future type $U$,
+    then \code{$T <: S$} and \code{$S <: U$}, so \code{$T <: U$}.
```
lrhn, line 11838: Do we need to show that `S <: U` inductively on the definition of "derives a future type"?

Or generally that if `T` derives a future type `U` then `T <: U (if that is true... I think it is.)

specification/dartLangSpec.tex
```diff
@@ -11829,7 +11829,40 @@ \subsection{Function Expressions}
...
+
+\item If $T$ derives a future type \code{Future<$S$>}
+  or \code{FutureOr<$S$>}, then, since \code{Future<$S$> $<:$ FutureOr<$S$>},
+  it follows that \code{$T <:$ FutureOr<$S$>}. Since \code{$\flatten{T} = S$}
+  in this case, it follows that \code{$T <:$ FutureOr<$\flatten{T}$>}.
+
+\item If $T$ derives a future type \code{Future<$S$>?} or
+  \code{FutureOr<$S$>?}, then, since \code{Future<$S$>? $<:$ FutureOr<$S$>?},
+  it follows that \code{$T <:$ FutureOr<$S$>?}, and therefore
+  \code{$T <:$ FutureOr<$S$?>}. Since \code{$\flatten{T} = S$?} in this case,
```
lrhn, line 11858: ... because `FutureOr<T>? <: FutureOr<T?>` for any type `T`, which can be shown using the union type subtype rules and from `Future<T> <: Future<T?>` by covariance.

specification/dartLangSpec.tex
fishythefish: The spec notes this property after the definition of "derives", but the casework is straightforward:
* Case 1: `T` is a type introduced by a class, mixin, or enum declaration, and `T` or some superinterface of `T` is `Future<S>` for some `S`, and `U = Future<S>`. Then `T <: U` is trivial.
* Case 2: `T` is `FutureOr<S>` for some `S` and `U = FutureOr<S>`. `T <: U` is again trivial.
* Case 3: `T` is `S?` for some `S`, `S` derives the future type `F`, and `U = F?`. Then to show `T <: U`, we must show `S? <: F?`. We have `S <: F` by induction, so `S? <: F?` follows by the usual rules for nullable types.
* Case 4: `T` is a type variable with bound `B`, `B` derives the future type `F`, and `U = F`. Then `B <: F` by induction, so `T <: B <: F = U`.

specification/dartLangSpec.tex
stereotype441: Done.

specification/dartLangSpec.tex
stereotype441: @fishythefish agreed! While you were typing this up, I was modifying my PR, adding a similar argument to the spec.

I _think_ our arguments are equivalent. Let me know if you agree.

specification/dartLangSpec.tex
stereotype441: Good point, this was a reasoning gap in my proof. Updated!

resources/type-system/inference.md
```diff
@@ -1046,22 +1046,27 @@ The elaboration process sometimes introduces new operations that are not easily
...
+  - Let `T_1` be the static type of `m_1`.
+
+  - Let `T` be `flatten(T_1)`.
+
+  - Evaluate `m_1`, and let `v` denote the result. _Note that since `m_1` has
+    static type `T_1`, by soundness, `v` must be an instance satisfying
+    `T_1`. Since `T_1 <: FutureOr<flatten(T_1)>` for all `T_1`, it follows that
+    `v` must be an instance satisfying `FutureOr<flatten(T_1)>`, or
+    equivalently, satisfying `FutureOr<T>`. Therefore, `v` must either be an
+    instance satisfying either `Future<T>` or `T`. We consider the two cases
```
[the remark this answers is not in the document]
stereotype441: Good idea. Done.

resources/type-system/inference.md
```diff
@@ -1046,22 +1046,27 @@ The elaboration process sometimes introduces new operations that are not easily
...
   - Otherwise, let `u` be a future satisfying type `Future<T>` that will
     complete to the value `v` at some later point. Then,
-    `@AWAIT_WITH_TYPE_CHECK<T>(m_1)` evaluates to `u`.
-
-    - _TODO(paulberry): explain why such a future is guaranteed to (soundly)
-      exist, by stipulating that if `T_1` is the static type of `m_1`, then `T`
-      must be `flatten(T_1)`, and then proving a lemma that `T_1 <:
-      FutureOr<flatten(T_1)>`; therefore `T_1 <: FutureOr<T>`, so in this
-      "otherwise" case, `v` must be an instance satisfying `T`._
+    `@AWAIT_WITH_TYPE_CHECK(m_1)` evaluates to `u`. _Such a future is guaranteed
```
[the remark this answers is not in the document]
stereotype441: Hmm, I think we may have had a miscommunication here. I'm not trying to claim that the reader can pick a `u` satisfying certain properties, and no matter which one they pick, `@AWAIT_WITH_TYPE_CHECK(m_1)` will evaluate to it.

Rather, I'm trying to _define_ the runtime behavior of `@AWAIT_WITH_TYPE_CHECK(m_1)`, but I'm trying to do it in a way that allows for some freedom in the implementation. What is required in order for the implementation to be conformant to the spec is for `@AWAIT_WITH_TYPE_CHECK(m_1)` to evaluate to a `v` that has these certain properties.

Does that make sense?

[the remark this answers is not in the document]
stereotype441: Ok, I've spent some more time thinking about this idea, and I think the problem I'm hung up on is that I'm trying to describe the operation of `@AWAIT_WITH_TYPE_CHECK` procedurally. I'm saying "do this, then do that, then do that." If I were defining `@AWAIT_WITH_TYPE_CHECK` as a separate function all its own, then the last step in the function would be "then return `u`", and I think it would be clear to everyone that what I mean by "then return `u`" is "then the implementation should continue with its evaluation of whatever expression or statement contained the invocation of `@AWAIT_WITH_TYPE_CHECK`, considering `u` to be the result of that invocation". But I don't want to say "then return `u`" because I'm _not_ defining `@AWAIT_WITH_TYPE_CHECK` as a separate function, so I'm worried that a reader might misinterpret "then return `u`" as meaning "now exit whatever function is currently being compiled, with a return value of `u`", and that's definitely _not_ what I mean.

So the wording I went with was "`@AWAIT_WITH_TYPE_CHECK(m_1)` evaluates to `u`". Since the context of this text is "`@AWAIT_WITH_TYPE_CHECK(m_1)` represents the following operation:", I thought it would be clear enough that what I mean here is "and now we're done with the operation I'm defining, and `u` is the resulting value." I really didn't expect anyone to think I was suggesting that "for all `u` satisfying certain properties, `@AWAIT_WITH_TYPE_CHECK(m_1)` evaluates to that `u`".

I definitely don't think my phrasing is great, and I'm open to discussing alternatives. But I'm not understanding how to fit your suggestion into a procedural description of the operation.

Are you suggesting that I start off with, "Let `u` be the result of evaluating `@AWAIT_WITH_TYPE_CHECK(m_1)`. Then `u` has the following properties...", like this?

> - `@AWAIT_WITH_TYPE_CHECK(m_1)` represents the following operation. Let `u` be the result of evaluating `@AWAIT_WITH_TYPE_CHECK(m_1)`. Then `u` has the following properties:
>
>   - Let `T_1` be the static type of `m_1`.
>
>   - Let `T` be `flatten(T_1)`.
>
>   - Evaluate `m_1`, and let `v` denote the result. _Note that since `m_1` has static type `T_1`, by soundness, `v` must be an instance satisfying `T_1`. Since `T_1 <: FutureOr<flatten(T_1)>` for all `T_1`, it follows that `v` must be an instance satisfying `FutureOr<flatten(T_1)>`, or equivalently, satisfying `FutureOr<T>`. Therefore, `v` must either be an instance satisfying either `Future<T>` or `T`. We consider the two cases below._
>
>   - If `v` is an instance satisfying type `Future<T>`, then `u` is equal to `v`.
>
>   - Otherwise, `u` is a future satisfying type `Future<T>` that will complete to the value `v` at some later point. _Such a future is guaranteed to (soundly) exist, because in this case `v` is an instance satisfying `T`._

This approach seems really awkward to me, because it no longer feels like a procedural description of an operation, and IMHO a procedural description of an operation is really going to be the most straightforward thing to do here.

CC @lrhn @leafpetersen @eernstg in case any of you have other ideas for how to phrase this more clearly.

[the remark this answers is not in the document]
fishythefish: Ah, now I understand what you're getting at. It wasn't clear to me that this was an implementation procedure. In that case, I might say something like "*let* `@AWAIT_WITH_TYPE_CHECK(m_1)` evaluate to `u`." We can also just omit the name:

"Otherwise, `@AWAIT_WITH_TYPE_CHECK(m_1)` evaluates to a future satisfying type `Future<T>` that will complete to the value `v` at some later point."

I think now my biggest issue (and part of the reason I was confused before) is that we merely claim existence of such a future, but an algorithm should probably offer something more constructive? I know we're trying to avoid overspecifying implementation details, so we can't just say it evaluates `Future.value(v)`. But perhaps we should at least mention that as a witness to the existence property?

specification/dartLangSpec.tex
fishythefish: Yup, I think we're in agreement.

resources/type-system/inference.md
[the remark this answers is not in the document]
stereotype441: That makes sense, thanks! I've changed it to:

>   - If `v` is an instance satisfying type `Future<T>`, then let `@AWAIT_WITH_TYPE_CHECK(m_1)` evaluate to `v`.
>
>   - Otherwise, let `@AWAIT_WITH_TYPE_CHECK(m_1)` evaluate to a future satisfying type `Future<T>` that will complete to the value `v` at some later point. _Such a future could, for instance, be created by executing `Future<T>.value(v)` (this is sound because in this case `v` is an instance satisfying `T`). It also could be created using a more efficient implementation-specific mechanism._

fishythefish (approved):
Thanks for all the discussion and iteration!

resources/type-system/inference.md
```diff
@@ -1046,22 +1046,33 @@ The elaboration process sometimes introduces new operations that are not easily
...
 
-  - Otherwise, let `u` be a future satisfying type `Future<T>` that will
-    complete to the value `v` at some later point. Then,
-    `@AWAIT_WITH_TYPE_CHECK<T>(m_1)` evaluates to `u`.
+  - Evaluate `m_1`, and let `v` denote the result. _Note that since `m_1` has
+    static type `T_1`, by soundness, `v` must be an instance satisfying
+    `T_1`. Since `T_1 <: FutureOr<flatten(T_1)>` for all `T_1`, it follows that
+    `v` must be an instance satisfying `FutureOr<flatten(T_1)>`, or
+    equivalently, satisfying `FutureOr<T>`. Therefore, `v` must either be an
+    instance satisfying either `Future<T>` or `T`. We consider the two cases
```
lrhn, line 1060: either ... or both! In which case we use the `Future`-branch.
That preference can be suggested by phrasing it as:

"`v` must either be an instance satisfying `Future<T>`, or if not, it must be an instance satisfying `T`."

resources/type-system/inference.md
```diff
@@ -1046,22 +1046,33 @@ The elaboration process sometimes introduces new operations that are not easily
...
-      must be `flatten(T_1)`, and then proving a lemma that `T_1 <:
-      FutureOr<flatten(T_1)>`; therefore `T_1 <: FutureOr<T>`, so in this
-      "otherwise" case, `v` must be an instance satisfying `T`._
+  - If `v` is an instance satisfying type `Future<T>`, then let
+    `@AWAIT_WITH_TYPE_CHECK(m_1)` evaluate to `v`.
+
+  - Otherwise, let `@AWAIT_WITH_TYPE_CHECK(m_1)` evaluate to a future satisfying
+    type `Future<T>` that will complete to the value `v` at some later
+    point. _Such a future could, for instance, be created by executing
+    `Future<T>.value(v)` (this is sound because in this case `v` is an instance
```
lrhn, line 1069: Nit: A parentheses inside a commentary (italics) feels like nested commentary.
Maybe change `(this is sound ...)` to `which is sound ...`.

resources/type-system/inference.md
```diff
@@ -1046,22 +1046,33 @@ The elaboration process sometimes introduces new operations that are not easily
 expressible using the syntax of Dart. To allow these operations to be specified
 succintly, the syntax of Dart is extended to allow the following forms:
 
-- `@AWAIT_WITH_TYPE_CHECK<T>(m_1)`, represents the following operation:
+- `@AWAIT_WITH_TYPE_CHECK(m_1)` represents the following operation:
 
-  - Evaluate `m_1`, and let `v` denote the result.
+  - Let `T_1` be the static type of `m_1`.
```
lrhn, line 1051: I think the `-` needs to be indented by four spaces relative to the one above, to count as the next level of indentation.

resources/type-system/inference.md
```diff
@@ -1046,22 +1046,33 @@ The elaboration process sometimes introduces new operations that are not easily
...
+    `@AWAIT_WITH_TYPE_CHECK(m_1)` evaluate to `v`.
+
+  - Otherwise, let `@AWAIT_WITH_TYPE_CHECK(m_1)` evaluate to a future satisfying
+    type `Future<T>` that will complete to the value `v` at some later
+    point. _Such a future could, for instance, be created by executing
+    `Future<T>.value(v)` (this is sound because in this case `v` is an instance
+    satisfying `T`). It also could be created using a more efficient
+    implementation-specific mechanism._
+
+  - _Note that these two cases in the abstract correspond concretely to a type
```
lrhn, line 1073: Not sure I can read this paragraph. Feels like it's talking about a specific type check that I should know about.

(Also not sure this should be an item in the itemized list. Maybe unindent it?)

Consider:
>  - ...
>     - ...
>
>     _Note that the two cases here may imply to a runtime
>     implementation doing an acutual `is Future<T>` type check 
>     and branching on the result, or a compiler may be able
>     to optimize away the check in some cases._

lrhn (approved):

resources/type-system/inference.md
stereotype441: Agreed, that's better. Addressed in https://github.com/dart-lang/language/pull/3931.

resources/type-system/inference.md
stereotype441: I don't believe that's necessary, based on Github's markdown preview (and based on the use of bulleted lists elsewhere in the file).

resources/type-system/inference.md
stereotype441: I didn't realize this was an option, thank you! Addressed in https://github.com/dart-lang/language/pull/3931.

Issue #3879, opened 2024-06-04
Consider whether to loosen "must match the original" in augmentation spec


davidmorgan:
I noticed

- Augmenting constructors signature must match the original
- Type parameters must match the original with bounds and names
- Function signature must match the original

--in the context of macro (incremental) build performance these are all somewhat counter-productive: the macro must do work to find out about the signature enough to reproduce it exactly, including adding any needed imports; then the compiler must do work to check that it was reproduced correctly.

Would it maybe make sense to loosen the constructor+function signature requirements to be just the names + structure, like you would write when overriding a method and expecting the compiler to infer the types from super?

By avoiding mentioning types that could be correctly inferred the macro can be cheaper to compile and invalidated less often.

Re: type parameters, I can't think of any nice way to remove the need to mention types for type parameter bounds; but it does seem unfortunate that a macro that wants to do something unrelated to the type `Foo`:

```
@UsefulMacro()
class X<T extends Foo> {}
--- macro output
augment class X<T extends Foo> {
  void usefulMethod();
}
```

now has to rerun for changes to it. @jakemac53

davidmorgan:
Looks like some checking is missing today :) ... https://github.com/dart-lang/sdk/issues/55478

jakemac53:
We could allow inference to fill in types potentially... but omitted types also don't play very nicely with macros, since initially they only see the syntax so they will just get "OmittedTypeAnnotation" instances which are not very useful, and can't be inferred until later on, plus the actual inference of those types when it does happen will involve further dependencies being added.

davidmorgan:
Hmm I'm not sure I follow--isn't a macro more likely to be looking at the original declaration than the augmentation? So the types will be there. Can a macro even introspect an augmentation?

jakemac53:
> Hmm I'm not sure I follow--isn't a macro more likely to be looking at the original declaration than the augmentation? So the types will be there. Can a macro even introspect an augmentation?

It isn't specified, but definitely some things must involve some knowledge of the augmentation, such as the `hasBody` boolean for functions (if any augmentation has filled one in, it should return true).

Consider also a hand written augmentation which only applies a macro:

```
augment class A {
  @SomeMacro()
  foo(x);
}
```

Here at least the expectation might be that it seems exactly the declaration it was applied to and not the "original".

lrhn:
[...]

PR #3940, opened 2024-06-27
specify that types can be omitted from augmenting declarations, and will be inferred from the original

jakemac53:
Closes https://github.com/dart-lang/language/issues/3879

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -487,7 +487,7 @@ The augmenting function does not have to pass the same arguments to
 `augmented()` as were passed to it. It may call it once, more than once, or not
 at all.
 
-It is a compile-time error if:
+It is a compile-time error if, after type inheritance:
 
 *   The signature of the function augmentation does not exactly match the
     original function. This means the return types must be the same; there must
@@ -719,7 +719,7 @@ original redirecting initializer if there is one.
 In the augmenting constructor's body, an `augmented()` call invokes the
 original constructor's body.
 
-It is a compile-time error if:
+It is a compile-time error if, after type inheritance:
 
 *   The signature of the constructor augmentation does not match the original
     constructor. This means the return types must be the same; there must be the
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -318,6 +318,18 @@ It is a compile-time error if:
     augmentation can both declare a new declaration and augment it in the same
     file.*
 
+### Type inference
+
+Type inference for augmenting declarations works similar to overrides. Whenever
+a type is omitted, if the original declaration has a type specified, that type
+is inferred.
+
+This applies to return types, parameter types, and variable types, but not type
+parameters.
+
+It is a compile time error for a manually written type to be different from the
+inferred type in all cases.
+
 ### Augmented Expression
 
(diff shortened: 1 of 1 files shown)


davidmorgan:
Thanks! I suggest to tweak the blocks starting "The signature of the constructor augmentation" and "The signature of the function augmentation" at the same time, as the sense of these has now changed: you are not required to repeat everything, it won't necessarily be possible to see the full signature.

working/augmentation-libraries/feature-specification.md
```diff
@@ -318,6 +318,18 @@ It is a compile-time error if:
...
     file.*
 
+### Type inference
+
+Type inference for augmenting declarations works similar to overrides. Whenever
+a type is omitted, if the original declaration has a type specified, that type
+is inferred.
+
+This applies to return types, parameter types, and variable types, but not type
+parameters.
```
lrhn, line 328: That is, types, not names. 
What about type parameter bounds? Would be nice to inherit those.

lrhn (approved):

jakemac53:
> Thanks! I suggest to tweak the blocks starting "The signature of the constructor augmentation" and "The signature of the function augmentation" at the same time, as the sense of these has now changed: you are not required to repeat everything, it won't necessarily be possible to see the full signature.

Updated

working/augmentation-libraries/feature-specification.md
```diff
@@ -318,6 +318,19 @@ It is a compile-time error if:
     augmentation can both declare a new declaration and augment it in the same
     file.*
 
+### Type inheritance
+
+When a type is omitted in an augmenting declaration, if the first declaration
+that introduced the member has a type specified in the ommitted types position,
+then that type is inherited.
```
lrhn, line 325: > if the first declaration ...  has a type specified

Even if the original declaration didn't have a type specified, *it* would have had one inferred, inherited or defaulted. That's the type that gets used for the omitted type of the augmentetion.

working/augmentation-libraries/feature-specification.md
```diff
@@ -474,7 +487,7 @@ The augmenting function does not have to pass the same arguments to
 `augmented()` as were passed to it. It may call it once, more than once, or not
 at all.
 
-It is a compile-time error if:
+It is a compile-time error if, after type inheritance:
```
lrhn, line 490: (Type inference can't change the type of the augmenting declaration, it inherits its types, but it can only do so after type inference has given types to its augmented declaration, which is why this is "after type inference".)

lrhn (commented):
Pedantic nits. :)

davidmorgan:
@jakemac53 oh, now I know why I couldn't comment properly ... GitHub doesn't let you comment on unchanged regions :(

The two regions with "...accomplish anything semantically. But it ensures that anyone reading the..." should probably be removed, they no longer make sense since you can omit information if you want to.

jakemac53:
> The two regions with "...accomplish anything semantically. But it ensures that anyone reading the..." should probably be removed, they no longer make sense since you can omit information if you want to.

I left these in, while the types are omitted the parameter names are still required.

working/augmentation-libraries/feature-specification.md
jakemac53: Fixes here https://github.com/dart-lang/language/pull/3948

PR #3948, opened 2024-07-01
follow-up to omitting types in augmentations

jakemac53:
Follow up to some comments on https://github.com/dart-lang/language/pull/3940 after it was merged

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -269,9 +269,8 @@ it if they know what they’re doing.
 
 ### Type inheritance
 
-When a type is omitted in an augmenting declaration, if the first declaration
-that introduced the member has a type specified in the ommitted types position,
-then that type is inherited.
+When a type is omitted in an augmenting declaration, it inherits the final type
+of the augmented declaration (which could itself be inferred, inherited, etc).
 
 This applies to return types, parameter types, variable types, and type
 parameter bounds (but not type parameter names or regular parameter names, those
@@ -383,7 +382,7 @@ declaration (a based declaration and zero or more augmentations that are all
 above the current augmenting type declaration) according to the rules in the
 following subsections.
 
-It is a **compile-time error** if:
+It is a **compile-time error** if, after type inference:
 
 *   The augmenting type and corresponding type are not the same kind: class,
     mixin, enum, extension, or extension type. You cannot augment a class with a
@@ -399,10 +398,10 @@ It is a **compile-time error** if:
     isn't visible to a reader of the main declaration.*
 
 *   The augmenting type declares an `extends` clause for a `class` declaration,
-    but one was already
-    present _(or the `class` was a `mixin class` declaration, which does not
-    allow `extends` clauses)_. We do not allow overwriting an existing
-    `extends`, but one can be filled in if none had been specified.
+    but one was already present _(or the `class` was a `mixin class`
+    declaration, which does not allow `extends` clauses)_. We do not allow
+    overwriting an existing `extends`, but one can be filled in if none had been
+    specified.
 
 *   The augmenting type declares an `extends` clause, but one was already
     present. We don't allow overwriting an existing `extends`, but one can be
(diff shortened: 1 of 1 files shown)


working/augmentation-libraries/feature-specification.md
```diff
@@ -269,9 +269,8 @@ it if they know what they’re doing.
 
 ### Type inheritance
 
-When a type is omitted in an augmenting declaration, if the first declaration
-that introduced the member has a type specified in the ommitted types position,
-then that type is inherited.
+When a type is omitted in an augmenting declaration, it inherits the final type
```
lrhn, line 272: "final type" is not a defined concept.
Also, a type can be omitted which is not the type of the entire augmented declaration.

I'd make the type inheritance part of augmentation application. The augmenting declaration does not specify a type for something. Applying an agumenting declaration with types omitted will preserve the existing type of that thing from the augmented declaration.
(And an augmenting declaration which *does* specify a type for something must specify the same type as the augmented declaration has for that thing, which means the application still preserves the type of the augmented declaration. Types can be written in augmentations, but they cannot change anything, they have no effect other than documentation, and being validated for being the correct type.)

Consider:
> When a type is omitted in an augmenting declaration, be it a return type, variable type, parameter type,
> or type parameter bound type, the result of applying the augmentation has the same type at that position
> as the augmented declaration. _An augmenting declaration "inherits" omitted types from the augmented
> declaration._

lrhn (approved):
This is tricky stuff. I have ideas. I should try to collect them. :smile:

working/augmentation-libraries/feature-specification.md
```diff
@@ -399,10 +398,10 @@ It is a **compile-time error** if:
     isn't visible to a reader of the main declaration.*
 
 *   The augmenting type declares an `extends` clause for a `class` declaration,
-    but one was already
-    present _(or the `class` was a `mixin class` declaration, which does not
-    allow `extends` clauses)_. We do not allow overwriting an existing
-    `extends`, but one can be filled in if none had been specified.
+    but one was already present _(or the `class` was a `mixin class`
```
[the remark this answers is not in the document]
jakemac53: I did reword this a bit, but I think it is all covered by other rules in a general fashion. Above, we say:

```
A class, enum, extension type, or mixin augmentation may specify `extends`,
`implements` and `with` clauses (when generally supported). The types
in these clauses are appended to the original declarations’ clauses of the same
kind, and if that clause did not exist previously, then it is added with
the new types. All regular rules apply after this appending process, so you
cannot have multiple `extends` on a class, or an `on` clause on an enum, etc.
```

working/augmentation-libraries/feature-specification.md
```diff
@@ -458,7 +457,7 @@ The augmenting function does not have to pass the same arguments to
 `augmented(…)` as were passed to it. It may invoke `augmented` once, more than
 once, or not at all.
 
-It is a compile-time error if, after type inheritance:
+It is a compile-time error if, after type inference:
 
 *   The function signature of the augmenting function does not exactly match the
     function signature of the augmented function. This means the return types
```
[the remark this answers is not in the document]
jakemac53: Re-worded this such that they types only have to match when provided

working/augmentation-libraries/feature-specification.md
```diff
@@ -377,36 +378,39 @@ including enum values, are added to the instance or static namespace of the
...
+according to the rules in the following subsections.
+
+It's a **compile-time** error if a library contains two top-level declarations
+with the same name, and:
+
+*   Neither is an augmenting declaration, or
+*   one of the declarations is a class-like declarations and the other is not of
+    the same kind, meaning that at either one is a class, mixin, enum, extension
+    or extension type declaration, and the other is not the same kind of
+    declaration.
```
munificent, line 394: Isn't this error the same as the error below?

jakemac53: Yup, just forgot to delete that one when moving it. I will just push a fix directly.

PR #3961, opened 2024-07-04
Add assertions and variable types inferred from default values

eernstg:
Two obvious enhancements from the language team meeting and from your proposal: Allow primary constructors to have assertions, and allow optional parameters to obtain their declared type from the default value.

### DIFF
--- a/working/2364 - primary constructors/feature-specification.md
+++ b/working/2364 - primary constructors/feature-specification.md
@@ -4,7 +4,7 @@ Author: Erik Ernst
 
 Status: Draft
 
-Version: 1.0
+Version: 1.3
 
 Experiment flag: primary-constructors
 
@@ -81,9 +81,9 @@ The basic idea is that a parameter list that occurs just after the class
 name specifies both a constructor declaration and a declaration of one
 instance variable for each formal parameter in said parameter list.
 
-A primary constructor cannot have a body, and it cannot have an initializer
-list (and hence, it cannot have a superinitializer, e.g., `super(...)`, and
-it cannot have assertions).
+A primary constructor cannot have a body, and it cannot have an normal 
+initializer list (and hence, it cannot have a superinitializer, e.g., 
+`super.name(...)`). However, it can have assertions.
 
 The motivation for these restrictions is that a primary constructor is
 intended to be small and easy to read at a glance. If more machinery is
@@ -202,21 +202,29 @@ enum E {
 class const Point(int x, int y);
 
 enum E(String s) { one('a'), two('b') }
-
-extension type I.name(int x); // Must use a primary constructor.
-
 ```
 
-This mechanism follows an existing pattern, where `const` modifiers can be
-omitted in the case where the immediately syntactic context implies that
-this modifier _must_ be present. For example, `const [const C()]` can be
-written as `const [C()]`. In the examples above, the parameter-and-variable
-declarations `final int x` and `final int y` are written as `int x` and
-`int y`, and this is allowed because it would be a compile-time error to
-omit `final` in an `extension type`, and in a class with a constant
-constructor. In other words, when we see an `extension type` declaration, or we
-see `const` on the name of a class, we know that `final` is implied on all
-instance variables.
+Finally, an extension type declaration is specified to use a
+primary constructor (in that case there is no other choice,
+it is in the grammar rules):
+
+```dart
+// Using a primary constructor.
+
+extension type I.name(int x); // Must use a primary constructor.
+```
+
+The primary constructor feature follows an existing pattern, where `const`
+modifiers can be omitted in the case where the immediately syntactic
+context implies that this modifier _must_ be present. 
+
+For example, `const [const C()]` can be written as `const [C()]`. In the
+examples above, the parameter-and-variable declarations `final int x` and
+`final int y` are written as `int x` and `int y`, and this is allowed
+because it would be a compile-time error to omit `final` in an `extension
+type`, and in a class with a constant constructor. In other words, when we
+see an `extension type` declaration, or we see `const` on the name of a
+class, we know that `final` is implied on all instance variables.
 
 Optional parameters can be declared as usual in a primary constructor, with
 default values that must be constant as usual:
@@ -231,6 +239,14 @@ class Point {
 
 // Using a primary constructor.
 class Point(int x, [int y = 0]);
+```
+
+We can omit the type of an optional parameter with a default value,
+in which case the type is inferred from the default value:
+
+```dart
+// Infer the type of `y` from the default value.
+class Point(int x, [y = 0]);
 ```
 
 Similarly for named parameters, required or not:
@@ -265,6 +281,22 @@ class const D<TypeVariable extends Bound>.named(
 ]) extends A with M implements B, C;
 ```
 
+Finally, it is possible to specify assertions on a primary constructor,
+just like the ones that we can specify in the initializer list of a
+rergular (not primary) constructor:
+
+```dart
+// Current syntax.
+class Point {
+  int x;
+  int y;
+  Point(this.x, this.y): assert(0 <= x && x <= y * y);
+}
+
+// Using a primary constructor.
+class Point(int x, int y): assert(0 <= x && x <= y * y);
+```
+
 ## Specification
 
 ### Syntax
@@ -282,7 +314,11 @@ constructors as well.
 <primaryConstructorNoConst> ::= // New rule.
      <typeIdentifier> <typeParameters>?
      ('.' <identifierOrNew>)? <formalParameterList>
-   
+     <assertions>?
+
+<assertions> ::= // New rule.
+     ':' <assertion> (',' <assertion>)*
+
 <classNamePartNoConst> ::= // New rule.
      <primaryConstructorNoConst>
    | <typeWithParameters>;
@@ -361,14 +397,28 @@ constants *(holding parameter default values)*.
 Where no processing is mentioned below, _D2_ is identical to _D_. Changes
 occur as follows:
 
+Assume that `p` is an optional formal parameter in _D_ that does not have a
+declared type, but it does have a default value whose static type in the
+empty context is a type `T` which is not `Null`. In that case `p` is
+considered to have the declared type `T`. When `T` is `Null`, `p` is
+considered to have the declared type `Object?`.
+
 The current scope of the formal parameter list of the primary constructor
 in _D_ is the current scope of the class/enum declaration *(in other words,
 the default values cannot see declarations in the class body)*. Every
 default value in the primary constructor of _D_ is replaced by a fresh
 private name `_n`, and a constant variable named `_n` is added to the
 top-level of the current library, with an initializing expression which is
-said default value. *(This means that we can move the parameter
-declarations including the default value without changing its meaning.)*
+said default value. 
+
+*This means that we can move the parameter declarations including the 
+default value without changing its meaning. Implementations are free to
+use this particular desugaring based technique, or any other technique
+which has the same observable behavior. In particular, it should not be
+possible for such a default value to obtain a new meaning because an
+identifier in the default value resolves to a declaration in the class body
+when it occurs in _k_ after the transformation, but it used to resolve to
+a top-level or imported declaration before the transformation.*
 
 For each of these constant variable declarations, the declared type is the
 formal parameter type of the corresponding formal parameter, except: In the
@@ -424,6 +474,13 @@ default value.
   removed from the parameter in _L2_, and it is added to the instance
   variable declaration named `p`.
 
+If there are any assertions following the formal parameter list _L_ then
+_k_ has an initializer list with the same assertions in the same order. The
+expressions in the assertions are subject to a transformation similar to
+the one which is used with default values, such that the meaning of the
+expression is the same when it occurs in the class header and when it
+occurs in the initializer list of _k_.
+
 Finally, _k_ is added to _D2_, and _D_ is replaced by _D2_.
 
 ### Discussion
@@ -439,15 +496,15 @@ class B extends A { // OK.
 class B(int a) extends A(a); // Could be supported, but isn't!
 ```
 
-There are several reasons why this is not supported. First, primary constructors
-should be small and easy to read. Next, it is not obvious how the
-superconstructor arguments would fit into a mixin application (e.g., when the
-superclass is `A with M1, M2`), or how readable it would be if the
-superconstructor is named (`class B(int a) extends A.name(a);`). For instance,
-would it be obvious to all readers that the superclass is `A` and not `A.name`,
-and that all other constructors than the primary constructor will ignore the
-implied superinitialization `super.name(a)` and do their own thing (which might
-be implicit).
+There are several reasons why this is not supported. First, primary
+constructors should be small and easy to read. Next, it is not obvious how
+the superconstructor arguments would fit into a mixin application (e.g.,
+when the superclass is `A with M1, M2`), or how readable it would be if the
+superconstructor is named (`class B(int a) extends A.name(a);`). For
+instance, would it be obvious to all readers that the superclass is `A` and
+not `A.name`, and that all other constructors than the primary constructor
+will ignore the implied superinitialization `super.name(a)` and do their
+own thing (which might be implicit).
 
 In short, if you need to write a complex superinitialization like
 `super.name(e1, otherName: e2)` then you need to use a body constructor.
@@ -599,6 +656,12 @@ where it is a compile-time error to not have them, but that is a
 [inferred-required]: https://github.com/dart-lang/language/blob/main/working/0015-infer-required/feature-specification.md
 
 ### Changelog
+
+1.3 - July 4, 2024
+
+* Add support for assertions in the primary constructor. Add support for
+  inferring the declared type of an optional parameter based on its default
+  value.
 
 1.2 - May 24, 2024
 
--- a/working/2364 - primary constructors/feature-specification.md
+++ b/working/2364 - primary constructors/feature-specification.md
@@ -81,9 +81,11 @@ The basic idea is that a parameter list that occurs just after the class
 name specifies both a constructor declaration and a declaration of one
 instance variable for each formal parameter in said parameter list.
 
-A primary constructor cannot have a body, and it cannot have an normal 
-initializer list (and hence, it cannot have a superinitializer, e.g., 
-`super.name(...)`). However, it can have assertions.
+A primary constructor cannot have a body, and it cannot have an normal
+initializer list (and hence, it cannot have a superinitializer, e.g.,
+`super.name(...)`). However, it can have assertions, it can have
+initializing formals (`this.p`) and it can have super parameters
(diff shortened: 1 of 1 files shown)


working/2364 - primary constructors/feature-specification.md
```diff
@@ -81,9 +81,9 @@ The basic idea is that a parameter list that occurs just after the class
 name specifies both a constructor declaration and a declaration of one
 instance variable for each formal parameter in said parameter list.
 
-A primary constructor cannot have a body, and it cannot have an initializer
-list (and hence, it cannot have a superinitializer, e.g., `super(...)`, and
-it cannot have assertions).
+A primary constructor cannot have a body, and it cannot have an normal 
+initializer list (and hence, it cannot have a superinitializer, e.g., 
+`super.name(...)`). However, it can have assertions.
```
lrhn, line 86: and field initializers?

working/2364 - primary constructors/feature-specification.md
```diff
@@ -202,21 +202,29 @@ enum E {
 class const Point(int x, int y);
 
 enum E(String s) { one('a'), two('b') }
+```
```
lrhn, line 205: In line 173 above, it says
> The modifier `final` on a parameter in a primary constructor has no meaning
> for the parameter itself, because there is no scope where the parameter can
> be accessed. 

That is no longer true if we allow asserts (or field initializers or super-constructor invocations).

working/2364 - primary constructors/feature-specification.md
```diff
@@ -202,21 +202,29 @@ enum E {
...
+modifiers can be omitted in the case where the immediately syntactic
+context implies that this modifier _must_ be present. 
+
+For example, `const [const C()]` can be written as `const [C()]`. In the
+examples above, the parameter-and-variable declarations `final int x` and
+`final int y` are written as `int x` and `int y`, and this is allowed
+because it would be a compile-time error to omit `final` in an `extension
+type`, and in a class with a constant constructor. In other words, when we
+see an `extension type` declaration, or we see `const` on the name of a
+class, we know that `final` is implied on all instance variables.
```
lrhn, line 227: This paragraph starts with "For example", so it sounds descriptive.
However, the content feels prescriptive. 

It's the first notion that `final` can be omitted, and it doesn't follow as "for example" from the previous paragraph which only mentions `const`.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -265,6 +281,22 @@ class const D<TypeVariable extends Bound>.named(
 ]) extends A with M implements B, C;
 ```
 
+Finally, it is possible to specify assertions on a primary constructor,
+just like the ones that we can specify in the initializer list of a
+rergular (not primary) constructor:
```
lrhn, line 286: rergular -> regular

working/2364 - primary constructors/feature-specification.md
```diff
@@ -361,14 +397,28 @@ constants *(holding parameter default values)*.
 Where no processing is mentioned below, _D2_ is identical to _D_. Changes
 occur as follows:
 
+Assume that `p` is an optional formal parameter in _D_ that does not have a
+declared type, but it does have a default value whose static type in the
```
lrhn, line 401: Should probably not apply to initializing formals or super-parmeters (if we allow those) because they get their type from their target instead.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -361,14 +397,28 @@ constants *(holding parameter default values)*.
...
+empty context is a type `T` which is not `Null`. In that case `p` is
```
lrhn, line 402: (Probably a good call to special-case `Null`. Although writing `var x = null` is unnecessary, you can just write `var x`, it's also good that the two actually mean the same thing.)

working/2364 - primary constructors/feature-specification.md
```diff
@@ -361,14 +397,28 @@ constants *(holding parameter default values)*.
...
+considered to have the declared type `T`. When `T` is `Null`, `p` is
+considered to have the declared type `Object?`.
```
lrhn, line 404: If there is no default value (including not being optional) and there is no declared type, is the type `dynamic` or `Object?`?

working/2364 - primary constructors/feature-specification.md
```diff
@@ -361,14 +397,28 @@ constants *(holding parameter default values)*.
...
+
 The current scope of the formal parameter list of the primary constructor
 in _D_ is the current scope of the class/enum declaration *(in other words,
 the default values cannot see declarations in the class body)*. Every
```
lrhn, line 408: This scope differs from in-body constructors, but I can see how it's *partly* consistent with the position. Might want to check if this is what users expect. Maybe moving an in-body field declaration `String x = _defaultString;` into the primary constructor parameter list should "just work", and not require rewriting as `String x = MyClass._defaultString`.

The expression *should* be inside the parameter scope, because that's where i syntactically is.
That means:
```dart
static const x = 0, T = 0, C = 0;
class C<T>(final int x, [int y = x + T + C]);
```
should reject all of `x`, `T` and `C` because they are in scope for the expression (they're declared before and around the expression, so they should be in the lexical scope for it).
That can still be outside the body scope of the class, but it's not the surrounding scope of the declaration either (which I think "current scope" means).

working/2364 - primary constructors/feature-specification.md
```diff
@@ -361,14 +397,28 @@ constants *(holding parameter default values)*.
...
 default value in the primary constructor of _D_ is replaced by a fresh
 private name `_n`, and a constant variable named `_n` is added to the
 top-level of the current library, with an initializing expression which is
-said default value. *(This means that we can move the parameter
-declarations including the default value without changing its meaning.)*
+said default value.
```
lrhn, line 412: Rewrites! Can we avoid them? 

I assume this rewrite happens because we also desugar the primary constructor into a synthetic syntactic declaration of a normal constructor, and need to preserve the scope of the moved expressions. That means moving expressions between scopes, and I can see why that's bad.
(If the expressions were in the parameter scope, inside the type parameter scope and the class body scope, like they normally would be for a constructor parameter default value, then moving them into that position wouldn't change anything. Maybe that's another reason to do that.)

Also, with assert expressions, the scope of those cannot be hoisted in the same way. They need access to the parameter scope (class name, type parameters, parameters), even if not the class body scope.

A rewrite is a change, which means there is a before and after.
*When* are the default value *expressions* replaced by the fresh identifier?
What happens before and what happens after.
(I'd guess it happens *very early*, right after parsing, because it can.)

working/2364 - primary constructors/feature-specification.md
```diff
@@ -424,6 +474,13 @@ default value.
   removed from the parameter in _L2_, and it is added to the instance
   variable declaration named `p`.
 
+If there are any assertions following the formal parameter list _L_ then
+_k_ has an initializer list with the same assertions in the same order. The
+expressions in the assertions are subject to a transformation similar to
+the one which is used with default values, such that the meaning of the
+expression is the same when it occurs in the class header and when it
+occurs in the initializer list of _k_.
```
lrhn, line 482: Nice hand-wave, but *how* is that done when the expressions can refer to parameters and type parameters.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -439,15 +496,15 @@ class B extends A { // OK.
...
-superconstructor is named (`class B(int a) extends A.name(a);`). For instance,
-would it be obvious to all readers that the superclass is `A` and not `A.name`,
-and that all other constructors than the primary constructor will ignore the
-implied superinitialization `super.name(a)` and do their own thing (which might
-be implicit).
+There are several reasons why this is not supported. First, primary
+constructors should be small and easy to read. Next, it is not obvious how
+the superconstructor arguments would fit into a mixin application (e.g.,
+when the superclass is `A with M1, M2`), or how readable it would be if the
+superconstructor is named (`class B(int a) extends A.name(a);`). For
```
lrhn, line 503: (I think it's fine.
```dart
class BananaSouffle(
   final int spoons,
   final Color sprinkles,
   BananaSort sort
) extends Banana(sort) with Edible;
```
So, yes, I'll argue that. It will make the syntax useful for exactly the kind of simple classes 
that just extend their superclass with extra fields.
```dart
class Person(
  final String name, 
  final int age, {
  super.id
}) extends LegalEntity(kind: Entity.person);
```
)

working/2364 - primary constructors/feature-specification.md
```diff
@@ -439,15 +496,15 @@ class B extends A { // OK.
...
+instance, would it be obvious to all readers that the superclass is `A` and
+not `A.name`, and that all other constructors than the primary constructor
```
lrhn, line 505: (If we have other non-redirecting generative constructors.)
But yes, I think it will be obvious.

working/2364 - primary constructors/feature-specification.md
eernstg: Yes.

working/2364 - primary constructors/feature-specification.md
eernstg: Right, adjusted the description.

working/2364 - primary constructors/feature-specification.md
eernstg: I removed the paragraph: There is a certain similarity between the implicit provision of `final` on these implicitly induced variable declarations and the implicit provision of the modifier `const` in nested levels of a constant expression, but they are also quite different. So it's gone.

working/2364 - primary constructors/feature-specification.md
eernstg: Done.

working/2364 - primary constructors/feature-specification.md
eernstg: Good catch! Done.

working/2364 - primary constructors/feature-specification.md
eernstg: Ack!

working/2364 - primary constructors/feature-specification.md
eernstg: Now using `Object?`, with a remark about the rationale.

working/2364 - primary constructors/feature-specification.md
eernstg: ;-)

working/2364 - primary constructors/feature-specification.md
eernstg: Findes det virkelig, banansoufflé?! ;-)

In any case, we could do it for non-mixin-application superclasses.

working/2364 - primary constructors/feature-specification.md
eernstg: It's quite similar to the situation with mixins: We claim that a mixin application will give rise to a fresh, implicitly induced class, and the instance members of that class are "obtained" from the mixin declaration.

So we might want to specify that there is a way to obtain "the semantics" of a given member declaration such that we can say that we "add that semantics" to a class.

We could also specify that an AST can be resolved, and we can freely move a resolved AST around because none of the identifiers in that AST will be looked up again.

eernstg: I've rewritten this paragraph to be more explicit on the required outcome, rather than specifying a particular desugaring strategy.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -170,15 +172,20 @@ class", but it is consistent with the rest of the language to say that this
...
+not have side effects, so we can basically ignore this. However, much more
+importantly, this modifier is also used to specify that the instance
+variable declaration which is induced by this primary constructor parameter
+is `final`.
 
 In the case where the constructor is constant, and in the case where the
 declaration is an `extension type` or an `enum` declaration, the modifier
 `final` on every instance variable is required. Hence, it can be omitted
-from the formal parameter in the primary constructor:
+from the formal parameter in the primary constructor, because it is implied
```
jakemac53, line 186: Fwiw this does make converting a constructor from const to non-const potentially error prone - you could easily forget to add `final` to all the fields that you want to keep final.

eernstg: True. But how often does it happen that a constant constructor get turned into a non-constant one? After all, it's going to break every constant expression out there where this constructor is invoked.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -233,6 +236,14 @@ class Point {
 class Point(int x, [int y = 0]);
 ```
 
+We can omit the type of an optional parameter with a default value,
+in which case the type is inferred from the default value:
+
+```dart
+// Infer the type of `y` from the default value.
+class Point(int x, [y = 0]);
```
munificent, line 244: Do we support this inference only for primary constructor parameters? This seems like it would be useful for all parameters (and I expect users would intuit that if it works for parameters in one place, it would work for them everywhere).

working/2364 - primary constructors/feature-specification.md
```diff
@@ -439,15 +515,15 @@ class B extends A { // OK.
...
-should be small and easy to read. Next, it is not obvious how the
-superconstructor arguments would fit into a mixin application (e.g., when the
-superclass is `A with M1, M2`), or how readable it would be if the
-superconstructor is named (`class B(int a) extends A.name(a);`). For instance,
-would it be obvious to all readers that the superclass is `A` and not `A.name`,
-and that all other constructors than the primary constructor will ignore the
-implied superinitialization `super.name(a)` and do their own thing (which might
-be implicit).
+There are several reasons why this is not supported. First, primary
+constructors should be small and easy to read. Next, it is not obvious how
```
munificent, line 519: Is "should be small and easy to read" not an equally compelling argument against supporting assertions in primary constructors?

I find it pretty surprising that we'd be OK with supporting asserts but not other initializers. Seems like a fairly arbitrary limitation.

working/2364 - primary constructors/feature-specification.md
jakemac53: I have no idea, I assume it would only happen when a class that used to be const no longer can be.

It begs the question though, is there really much value in allowing omitting `final` versus just requiring the user to write it? Is it more readable with or without the explicit `final`?

working/2364 - primary constructors/feature-specification.md
eernstg: Right, https://github.com/dart-lang/language/issues/3970.

working/2364 - primary constructors/feature-specification.md
eernstg: We have concrete evidence that developers may need assertions, that was an important point for me. The assertions would be able to specify constraints on the parameters using unsurprising scope rules, there's nothing special to remember in order to use this feature.

Supporting a general initializer list is a significantly bigger step than just supporting assertions: Surely we'd then want to support the introduction of additional instance variables by means of initializer list entries, and this would probably mean that `final int x = e` will introduce a final instance variable of type `int` whereas `this.x = e` will not, and `int this.x = e` is not supported, nor is `final this.x = e`. This implies that developers can _not_ just rely on their prior experience with initializer lists, they'll need to get used to a new syntax and a new semantics of initializer lists, just for primary constructors.

Next, we'd probably support `super.someName(...)` as well (if we want to be able to say that "primary constructors can have initializer lists").

Primary constructors is a conciseness feature and I just don't think it's a good idea to keep piling up additional features, and paying for them in terms of more verbose syntax and more complex semantics. If you want something more general then write a regular constructor.

(OK, the following could actually be a useful technique: Write a primary constructor just to get started, then ask the IDE to transform it into a normal constructor, then add a body and other stuff as needed. This would allow developers to avoid _writing_ the names twice, even in the case where it is known up front that it will be a regular constructor.)

munificent: > We have concrete evidence that developers may need assertions, that was an important point for me.

I mean, we also have concrete evidence that developers may need initializer lists and super constructor calls... :)

You can always just not use a primary constructor if you need assertions, right?

> Supporting a general initializer list is a significantly bigger step than just supporting assertions: Surely we'd then want to support the introduction of additional instance variables by means of initializer list entries,

For what it's worth, I definitely would *not* expect that to work. I would only expect the initializer list to, well, initialize fields that are already declared. I don't know if my expectation would be widely shared.

> and this would probably mean that `final int x = e` will introduce a final instance variable of type `int` whereas `this.x = e` will not, and `int this.x = e` is not supported, nor is `final this.x = e`. This implies that developers can _not_ just rely on their prior experience with initializer lists, they'll need to get used to a new syntax and a new semantics of initializer lists, just for primary constructors.

I think we could sidestep all of this by saying that, no, the initializer list only initializes. The way I would think about it is that there are two language features:

1. Function (or method, or constructor) parameters.
2. Field declarations.

Both of those bind new variables. The magic of the primary constructor feature is that it lets you combine those. But the two things it combines are always *variable declarations*. Allowing field declarations inside the initializer list would be a different thing because there you aren't already declaring something.

Of course, allowing an initializer list in a primary constructor *would* imply that users will want non-field primary constructor parameters, which is another design question we've talked about.

> Next, we'd probably support `super.someName(...)` as well (if we want to be able to say that "primary constructors can have initializer lists").

I suspect we'd just put this after the superclass name in the `extends` clause like other languages do, though that has some design challenges.

> Primary constructors is a conciseness feature and I just don't think it's a good idea to keep piling up additional features, and paying for them in terms of more verbose syntax and more complex semantics. If you want something more general then write a regular constructor.

Agreed! So let's not pile on assertions in them. :)

> (OK, the following could actually be a useful technique: Write a primary constructor just to get started, then ask the IDE to transform it into a normal constructor, then add a body and other stuff as needed. This would allow developers to avoid _writing_ the names twice, even in the case where it is known up front that it will be a regular constructor.)

Yeah, I think that would definitely be useful.

working/2364 - primary constructors/feature-specification.md
eernstg: Definitely. Done.

working/2364 - primary constructors/feature-specification.md
eernstg: I'll preserve the ability (it was always an option, not a requirement) to omit `final` when every instance variable in the declaration must be final (that is, the primary constructor is `const` or the declaration is an enum or an extension type). There are two reasons: (1) I'd like to maintain the emphasis on brevity, and _requiring_ `final` or `var` on every formal parameter that introduces an instance variable will make the syntax considerably more verbose. (2) Extension types already use this abbreviation, and it would be nice if we can say that the `(...)` in the header of an extension type declaration is just a primary constructor, with no special exceptions.

If everybody insists that we must have `var` or `final` on every formal parameter of a primary constructor that introduces an instance variable then it shouldn't be hard to create the syntax and rules for that, but right now I want to keep the more concise form in mind.

eernstg:
I'll go ahead and land this.

We haven't resolved the dilemma: Keep the current restrictions and preserve the conciseness vs. generalize the mechanism and accept the more verbose form. I think it's better to take that discussion a few more times in the language team.

Issue #3957, opened 2024-07-03
Allow an augmenting variable declaration without an initializer


eernstg:
The augmentation specification says that

> Since the initializer expression is the only meaningful part of the augmenting declaration, an initializer expression must be provided

in the location where it specifies an augmenting variable declaration that augments a variable declaration. This means that the following is a compile-time error:

```dart
int x = 1;

/// A DartDoc comment.
@someMetadata
augment int x;
```

This seems benign to me, and potentially useful, so I'd recommend that we turn this into a non-error.

jakemac53:
I agree, and I think there might be a few other similar mentions of errors?

When originally written we hadn't considered augmenting only to provide metadata or doc comments, but we do want to allow that.

PR #3968, opened 2024-07-08
Allow augmentations of metadata and doc comments only for all kinds of augmentations

jakemac53:
Closes https://github.com/dart-lang/language/issues/3957.

- Allows all declarations to be augmented with metadata/doc comments explicitly.
- Allows fields to be augmented with fields that contain no initializer. The augmented initializer is left in tact.
- Allow enum values to be augmented with no argument list. The augmented enum value keeps its original argument list in tact.
- Allow all function, getter, setter augmentations to have no body. The augmented body is left in tact.
- Unrelated, but I cleaned up many usages of "original" => "augmented"

I did not introduce an error to have a totally useless augmentation with no initializer/body, doc comments, or metadata, but I didn't see the need to introduce one. We could though if people think we should.

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -467,7 +467,7 @@ The augmenting function does not have to pass the same arguments to
 `augmented(…)` as were passed to it. It may invoke `augmented` once, more than
 once, or not at all.
 
-Augmenting function declarations may also omit the body, in order to only
+An augmenting function declaration may have an empty body (`;`) in order to only
 augment the metadata or doc comments of the function. In this case the body of
 the augmented member is not altered.
 
@@ -549,9 +549,9 @@ More specifically:
     with the body of the augmenting getter. Inside the augmenting getter’s
     body, an `augmented` expression executes the augmented getter’s body.
 
-    Augmenting getter declarations may also omit the body, in order to only
-    augment the metadata or doc comments of the getter. In this case the body of
-    the augmented getter is not altered.
+    An augmenting getter declaration may have an empty body (`;`) in order to
+    only augment the metadata or doc comments of the getter. In this case the
+    body of the augmented getter is not altered.
 
     Synthetic getters cannot be augmented with metadata or doc comments.
 
@@ -562,9 +562,9 @@ More specifically:
     `augmented = <expression>` assignment invokes the augmented setter with the
     value of the expression.
 
-    Augmenting setter declarations may also omit the body, in order to only
-    augment the metadata or doc comments of the setter. In this case the body of
-    the augmented setter is not altered.
+    An augmenting setter declaration may have an empty body (`;`) in order to
+    only augment the metadata or doc comments of the setter. In this case the
+    body of the augmented setter is not altered.
 
     Synthetic setters cannot be augmented with metadata or doc comments.
 
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -176,16 +176,16 @@ members are filled in by an augmentation declaration. _This is primarily useful
 for macros, which may be used to provide a body for an abstract member._
 
 An augmentation that replaces the body of a function, may also want to preserve
-and run the code of the original declaration that it augments (hence the name
-"augmentation"). It may want to run its own code before the original code,
-after it, or both. To support that, we allow a new expression syntax inside the
-“bodies” of augmenting declarations (function bodies, constructor bodies, and
-variable initializers). Inside an expression of a member marked
-`augment`, the identifier `augmented` can be used to refer to the augmented
-function, getter, or setter body, or variable initializer. This is a contextual
-keyword within `augment` members, and has no special meaning outside of that
-context. See the next section for a full specification of what `augmented`
-means, and how it must be used, in the various contexts.
+and run the code of the augmented declaration (hence the name "augmentation").
+It may want to run its own code before the augmented code, after it, or both.
+To support that, we allow a new expression syntax inside the “bodies” of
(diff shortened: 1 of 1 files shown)


working/augmentation-libraries/feature-specification.md
```diff
@@ -176,16 +176,16 @@ members are filled in by an augmentation declaration. _This is primarily useful
...
-context. See the next section for a full specification of what `augmented`
-means, and how it must be used, in the various contexts.
+and run the code of the augmented declaration (hence the name "augmentation").
+It may want to run its own code before the augmented code, after it, or both.
+To support that, we allow a new expression syntax inside the “bodies” of
+augmenting declarations (function bodies, constructor bodies, and variable
+initializers). Inside an expression of a member marked `augment`, the identifier
+`augmented` can be used to refer to the augmented function, getter, or setter
+body, or variable initializer. This is a contextual keyword within `augment`
+members, and has no special meaning outside of that context. See the next
```
munificent, line 186: Nit: "members" -> "declarations".

Since this paragraph isn't the full spec, it doesn't really matter, but using "declarations" here I think helps hint that `augment` is a keyword within an entire augmented class/mixin/etc. even if the member containing `augment` isn't itself an augmentation.

working/augmentation-libraries/feature-specification.md
```diff
@@ -459,14 +459,18 @@ augment int slowCalculation(int a, int b) {
...
+argument list matching the augmented function's parameter list and returns the
 function's return type.
 
 The augmenting function does not have to pass the same arguments to
 `augmented(…)` as were passed to it. It may invoke `augmented` once, more than
 once, or not at all.
 
+Augmenting function declarations may also omit the body, in order to only
+augment the metadata or doc comments of the function. In this case the body of
+the augmented member is not altered.
```
munificent, line 472: We'll probably want to be clearer about what it means to "omit" the body of a function. Maybe:

> An augmenting function declaration may have an empty body (`;`) in order to only augment the metadata...

working/augmentation-libraries/feature-specification.md
```diff
@@ -545,13 +549,25 @@ More specifically:
     with the body of the augmenting getter. Inside the augmenting getter’s
     body, an `augmented` expression executes the augmented getter’s body.
 
+    Augmenting getter declarations may also omit the body, in order to only
+    augment the metadata or doc comments of the getter. In this case the body of
+    the augmented getter is not altered.
+
+    Only concrete getters are allowed to augment with metadata or doc comments.
```
munificent, line 556: I don't understand what this sentence means. Is it saying that an augmenting getter with a `;` body *is* concrete (i.e. not abstract) or something else?

working/augmentation-libraries/feature-specification.md
```diff
@@ -619,23 +634,24 @@ It is a **compile-time error** if:
...
-*   A non-writable variable declaration is augmented with a setter. (
-    Instead, the author can declare a *non-augmenting* setter that goes
-    alongside the implicit getter defined by the final variable.)
-    _Non-writable variable declarations are any that does not introduce a
-    setter, including non-`late`  `final` variables, `late final` variables
-    with an initializer, and `const` variables._
+*   A non-writable variable declaration is augmented with a setter. (Instead,
+    the author can declare a *non-augmenting* setter that goes alongside the
+    implicit getter defined by the final variable.) _Non-writable variable
+    declarations are any that does not introduce a setter, including
```
munificent, line 652: Either "does" -> "do" or: "_A non-writable variable is any that does not...".

I'd lean towards the latter. (I've generally found it easier to read specs when they use singular and "any" to refer to any of a range of possible things.)

munificent (approved):
One bit of confusion but otherwise LGTM!

working/augmentation-libraries/feature-specification.md
jakemac53: I meant to say only non-synthetic getters can be augmented, but concrete is not the correct term as it implies having a body etc.

Updated the wording to say that synthetic getters (and setters below) cannot be augmented with metadata/doc comments.

We could possibly just remove this restriction, but I don't really know what the behavior should be if we did allow it.

munificent: Heh, I also don't know what a "synthetic" getter or setter is. 🙃

jakemac53: The ones introduced by variable declarations.

jakemac53: Maybe it is just an analyzer term, but basically they are any declarations which are not specified by the user explicitly, and are implicitly added based on some other declaration.

Not sure if the default constructor is considered synthetic or not.

munificent: Ah, I see. In that case, I don't see the value of this restriction. You can augment a variable declaration with a getter. It seems weird that this is OK:

```dart
int x;

augment int get x {
  print('wrapped!');
  return augmented;
}
```

But then this becomes an error:

```dart
int x;

/// Comment.
augment int get x {
  print('wrapped!');
  return augmented;
}
```

I can't think of any other places where adding a comment yields an error.

Can we just remove this restriction?

lrhn: Agree. We can say that doc-comments on augmenting declarations for implicit getters/setters will (or may) be ignored by tooling.

Annotations on the other hand, they're harder to ignore and it makes sense to not allow them.
(As long as you can declare the base variable without adding an initializer, just to update the docs.)

jakemac53: Yes the restriction is primarily because of annotations, especially macro annotations. We could allow doc comments though.

Issue #3459, opened 2023-11-09
Specify how macros interact with default constructors


jakemac53:
Currently the spec doesn't talk at all about default constructors, but it should.

My initial thought is that default constructors should only be added immediately following the declarations phase. If after that point a class does not have a constructor, then the default one is implicitly added, and in the definitions phase you will be able to introspect over it.

scheglov:
Fix for the analyzer: https://dart-review.googlesource.com/c/sdk/+/335383

jakemac53:
Another issue worth mentioning here, is if the default constructor is only added at the end of the declarations phase, then macros running in the declarations phase on that class will not ever be able to see the default constructor. This might be perfectly fine though, or at least reasonable.

lrhn:
If the declaration phase only sees *declarations*, then not seeing implicitly added member implementations may be fine.

Would it see an `int x;` declaration, or an `int get x {...}`+`set x(int _){...}` pair? 
The former is the declaration, the latter are the semantic members introduced by that declaration.

scheglov:
FYI, in the analyzer we:
1. Don't create the synthetic default constructor until after the declarations phase. So, if an actual constructor is added, we don't add the default one.
2. We *do* create synthetic getter/setter for `int x;` as soon as we see the variable declaration. Nothing in the running macro can change the fact that these getter/setter exist. However we don't return synthetic getter/setter from `methodsOf()`, so macros don't see them.

davidmorgan:
Does it look like there is anything needed here? Design, implementation? Thanks.

jakemac53:
I think the spec needs to be updated here to say exactly when the default constructor appears in introspection results. I think we are leaning towards it appearing after the declarations phase, as there isn't anything else reasonable that we can really do.

Whether or not we show synthetic members in general is another open question.

rrousselGit:
About this. Would a phase 2 or 3 be able to add a default constructor if none yet exist? In particular, would that added constructor be able to have required parameters and such?

A case would be when default constructors need to have access to the fields/methods of that class first ; which isn't doable in phase 1.   
For example:

```dart
@data
class Example {
  final int a;
}


// Adds:
augment class Example {
  Example(this.a);
}
```

If the constructor was previously introspected as "There's an empty default synthethic constructor", but then another macro changes that, it could get weird.

lrhn:
I'd say that there is no "default constructor" until after all augmentations have been applied.
[...]

PR #3975, opened 2024-07-10
specify that default constructors should only appear after phase 2

jakemac53:
Closes https://github.com/dart-lang/language/issues/3459

### DIFF
--- a/working/macros/feature-specification.md
+++ b/working/macros/feature-specification.md
@@ -45,6 +45,14 @@ Allowing deep introspection like this in cases where a macro needs it while
 ensuring that users can understand the system and tools can implement it
 efficiently is a central challenge of this proposal.
 
+#### Default Constructors
+
+Default constructors are not introduced until after phase 2 (the phase which
+might introduce generative constructors). They should not appear in
+introspection results until phase 3. If, after phase 2 a class still has no
+generative constructor, then the default one should be added, and it should
+be visible for introspection in phase 3.
+
 #### Omitted Type Annotations and Inference
 
 In general, the introspection APIs will only provide exactly what the user has


working/macros/feature-specification.md
```diff
@@ -45,6 +45,14 @@ Allowing deep introspection like this in cases where a macro needs it while
 ensuring that users can understand the system and tools can implement it
 efficiently is a central challenge of this proposal.
 
+#### Default Constructors
+
+Default constructors are not introduced until after phase 2 (the phase which
+might introduce generative constructors). They should not appear in
+introspection results until phase 3. If, after phase 2 a class still has no
+generative constructor, then the default one should be added, and it should
+be visible for introspection in phase 3.
```
munificent, line 54: I wish we had a clearer principle for what introspection yields. Is it what the user wrote in the source code, or is it what abstract semantic model that desugars to? It seems like we keep stumbling into corners of the language that raise that question:

1. Does introspection show the implicit getter/setter introduced by a variable declaration?
2. Do primary constructors show up as just another constructor or not?
3. Do default constructors show up in introspection?

I suspect we will run into this problem more in the future as we add more declaration-level syntactic sugar features like (potentially) data classes, value types, etc.

The principle I lean towards is that introspection should show what the code is written to say and not what it happens to mean after desugaring. That would imply that default constructors don't show up in introspection at all even in phase 3.

I do realize that default constructors are different from the other desugarings because it's conditional: you only get one if no other constructors are defined. Maybe that warrants special treatment where default constructors never show up in introspection even if we decide that synthetic getters, synthetic setters, and primary constructors do.

munificent (approved):
I'd lean towards not having them in introspection at all, but if you feel confident that having them show up in phase 3 is worth it, I'll defer to you.

working/macros/feature-specification.md
jakemac53: > The principle I lean towards is that introspection should show what the code is written to say and not what it happens to mean after desugaring.

I don't think this is what macro authors want. They want to know how a class can be _used_, not how it was written, generally.

> I wish we had a clearer principle for what introspection yields.

I think it should yield the most complete view of the thing being introspected over possible. That is what macro authors want, they don't really care what introduced the members. From a user perspective, it also shouldn't be breaking to move a constructor to a primary constructor, or visa-versa, and if macros can only see one or the other, or have to explicitly handle primary constructors in a special way, I think it makes it less likely that they can successfully handle them interchangeably.

This probably does mean showing implicit getters/setters, although we would need to highlight whether they were derived from a field or not.

Default constructors are I think unique, because they only are added if no generative constructor is added, and that doesn't play well with the phases.

jakemac53:
I am going to go ahead and merge this, because I do think its useful, but there should quite possibly be some follow up to specify some guiding principles.

PR #3981, opened 2024-07-12
Slightly clarify and copy-edit the augmentation ordering.

munificent:
There are no semantic changes here, I just found the prose a little unclear on first read. In particular:

```
*   If the `part` directive including the file that contains *A*
    is syntactically before the `part` directive including the file that
    contains *B* in source order, then *A* is before *B*.
```

Is ambiguous because "includes" is transitive, so there may be multiple files whose `part` directives include A or B and it's not clear what that means when those `part` directives aren't in the same file.

Also made a couple of style and copy tweaks while I was at it.

I like the semantics!

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -85,58 +85,57 @@ imports and prefixes of their parent (part or library) file.
 
 Augmentation declarations interact with part files mainly in restrictions on
 where an augmenting declaration may occur relative to the declaration it
-augments, as describe below.
+augments, as described below.
 
 For this, we define the following relations on *declarations* based on the
 relations between *files* of a library.
 
 We say that a syntactic declaration *occurs in* a Dart file if the
-declaration’s source code occurs in that Dart file.
+declaration's source code occurs in that Dart file.
 
 We then say that a Dart file *contains* a declaration if the declaration occurs
-in the file itself, or if any of the files included by the Dart file contains
-the declaration. _That is, if the declaration occurs in a file in the sub-tree
+in the file itself, or if any of the files included by the Dart file contain
+the declaration. _That is, if the declaration occurs in a file in the subtree
 of that Dart file._
 
-We then define a partial and a complete *ordering* of declarations of a library
-as follows:
+We then define two *orderings* of declarations in a library, one partial and one
+complete, as follows:
 
 We define a partial ordering on syntactic declarations of a library,
 *is above*, such that a syntactic declaration *A* is *above* a syntactic
 declaration *B* if and only if:
 
-*   *A* and *B* occur in the same file, and the start of the *A* declaration is
-syntactically before the start of the *B* declaration, in source order, or
-*   A file included by the file containing *A* contains *B*.
+*   *A* and *B* occur in the same file, and the start of *A* is syntactically
+    before the start of *B*, in source order, or
+*   The file containing *A* includes the file that contains *B*.
 
 We define a *total ordering relation* (transitive, anti-symmetric, irreflexive)
 on declarations of a library, *is before* (and its reverse, *is after*) such
 that for any two syntactic declarations *A*, and *B*:
 
 *   If *A* and *B* occur in the same file, then:
-    *   If the start of *A* is before the start of *B* in source order,
-        then *A* is before *B*.
+    *   If the start of *A* is syntactically before the start of *B* in source
+        order, then *A* is before *B*.
     *   Otherwise *B* is before *A*.
-*   Otherwise *A* and *B* occur in different files.
-*   Let *F* be the least containing file for those two files.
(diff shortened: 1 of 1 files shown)


working/augmentation-libraries/feature-specification.md
```diff
@@ -85,58 +85,57 @@ imports and prefixes of their parent (part or library) file.
...
 We define a partial ordering on syntactic declarations of a library,
 *is above*, such that a syntactic declaration *A* is *above* a syntactic
 declaration *B* if and only if:
 
-*   *A* and *B* occur in the same file, and the start of the *A* declaration is
-syntactically before the start of the *B* declaration, in source order, or
-*   A file included by the file containing *A* contains *B*.
+*   *A* and *B* occur in the same file, and the start of *A* is syntactically
+    before the start of *B*, in source order, or
+*   The file containing *A* includes the file that contains *B*.
```
eernstg, line 110: Just double checking the earlier text in the current version of the feature specification, I can see that 'contains' is transitive (meaning "occurs in a file which is in the subtree"), and 'includes' is the simple (and non-transitive) relationship declared by `part ...;`.

First of all, this means that "the file containing *A*" is ambiguous, and we'd probably want to say something like "the deepest file containing *A*". This is a file _F_ that contains *A* such that there is no file _F1_ which is included by _F_ where _F1_ contains *A* as well.

We could also talk about "contains *A*" meaning "the declaration of *A* is located here" and "covers *A*" meaning "contains *A* or includes a file that contains *A*", but that's a bigger change because we'd then need to change 'contains' to 'covers' almost everywhere.

In any case, we could say something like this: "The deepest file containing *A* includes a file that contains *B*."

(With "covers" that would be: "The file that contains *A* includes a file that covers *B*".)

working/augmentation-libraries/feature-specification.md
```diff
@@ -85,58 +85,57 @@ imports and prefixes of their parent (part or library) file.
...
-    contains *B* in source order, then *A* is before *B*.
-*   Otherwise *B* is before *A*.
 
 Then *B* *is after* *A* if and only if *A* *is before* *B*.
 
-(Here the first five points can be summarized as “If *A* is above *B*, then *A*
-is before *B*, and vice versa” and the remaining case covers when the two are
-contained in sibling part directives, and at least one of those three cases
-must occur.)
+*In short, if *A* is above *B*, then *A* is before *B*, and vice versa.
```
eernstg, line 132: 'vice versa' is dangerous here, we don't want to imply "if *A* is before *B*, then *A* is above *B*".

Of course, we could say that "*above* implies *before*", but this puts the focus on the relations rather than individual elements, which might be considered _less_ readable. ;-)

On the other hand, 'vice versa', in the sense that makes sense, just means "if *B* is above *A*, then *B* is before *A*". But this doesn't bring anything new, because it's just a different assignment of a concrete example to a couple of metavariables.

So I'd suggest that we simply omit 'vice versa'.

eernstg (approved):
Looks good! I added a couple of comments, too.

working/augmentation-libraries/feature-specification.md
lrhn: Or just use the same phrase used to define "contains", "occurs":

> * The file where *A* occurs includes the file where *B* occurs.

working/augmentation-libraries/feature-specification.md
lrhn: ACK. Better without.

working/augmentation-libraries/feature-specification.md
munificent: Good points! I like Lasse's phrasing. I'd like to avoid introducing any new terminology like "covers" if possible.

working/augmentation-libraries/feature-specification.md
munificent: Good call. Done.

jakemac53 (approved):

working/augmentation-libraries/feature-specification.md
```diff
@@ -107,7 +107,7 @@ declaration *B* if and only if:
 
 *   *A* and *B* occur in the same file, and the start of *A* is syntactically
     before the start of *B*, in source order, or
-*   The file containing *A* includes the file that contains *B*.
+*   The file where *A* occurs includes the file where *B* occurs.
```
eernstg, line 110: Another tweak. ;-)

I don't think there's any support for claiming that 'includes' is transitive. 'Includes F' basically means that there is a directive of the form `part of 'F';`. So the safe bet would be something like this:

```suggestion
*   The file where *A* occurs includes a file that contains *B*.
```

This means that *B* could occur at any level which is strictly deeper than *A* in the file tree, not just one level deeper.

eernstg (commented):
I think it would be useful to make one small adjustment. I'll just mention it here and then we can do it in the next PR, whatever that is.

Issue #3998, opened 2024-07-30
Allow augmenting an abstract variable with a concrete getter/setter?


jakemac53:
@sgrekhov had some questions surrounding this recently.

I could imagine this being useful, especially for extension types, consider this:

```dart
extension type User(Map<String, Object?> json) {
  abstract String name;
  
  // allowed? Today we would require them to also be abstract I believe.
  augment String get name => json['name'];
  augment set name(String value) => json['name'] = value; 
}
```

Specifically, an abstract variable is a nice way of defining the setter/getter without also creating a backing store (presumably extension types wouldn't even allow a non-abstract field). But it is only really useful if an augmentation can actually fill in concrete implementations of the getter and setter.

Conversely though, removing the `abstract` nature of the field does feel like it is fundamentally altering the field. And in general we do not allow you to augment an abstract function with a non-abstract function, so it feels inconsistent.

jakemac53:
Alternatively, we could tell people to use `external` in this case. But that does somewhat violate the intention of `external`.

munificent:
Yeah, I think it's totally reasonable to augment the getter and/or setter of an abstract variable declaration with an augmenting getter and/or setter. An abstract variable is purely syntactic sugar for an abstract getter (and setter if not `final`), so the semantics should be the same as if you'd declared an abstract getter or setter.

jakemac53:
> so the semantics should be the same as if you'd declared an abstract getter or setter.

This would mean _not_ allowing any useful augmentations though, because the augmenting getter/setter would have to be abstract as well. That may still be what we want to do in order to retain the desired semantics for augmentations (which don't generally allow converting abstract things into non-abstract things).

leafpetersen:
It definitely feels to me that this should be consistent.  If you can use augmentations to provide concrete implementations for abstract setter/getters, then you should be able to do the same for methods.

jakemac53:
I do think the consistency argument is a strong one, and we can just tell users to use `external` instead if they want a non-abstract getter/setter pair, which they can augment as they see fit.

I will go ahead and work on clarifying this in the spec. It is how I would interpret the spec currently, but it isn't very explicit.

munificent:
> It is slightly unfortunate that the language won't _require_ a body to be provided via augmentation in this case, which will result in a runtime failure when the functions are called if one is not provided.

[...]

PR #4000, opened 2024-07-30
clarify augmentations of abstract and external variables

jakemac53:
Closes https://github.com/dart-lang/language/issues/3998

- Clarifies some error cases for augmenting the getters/setters introduced by abstract and external variables.
- Removes an error blocking augmentation of an abstract/external variable with a variable. This is allowed as long as it retains it abstract/external keyword. It can be used to add comments or annotations.

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -1,7 +1,7 @@
 # Augmentations
 
 Author: rnystrom@google.com, jakemac@google.com, lrn@google.com <br>
-Version: 1.27 (see [Changelog](#Changelog) at end)
+Version: 1.28 (see [Changelog](#Changelog) at end)
 
 Augmentations allow spreading your implementation across multiple locations,
 both within a single file and across multiple files. They can add new top-level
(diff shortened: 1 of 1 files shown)


working/augmentation-libraries/feature-specification.md
```diff
@@ -719,6 +716,12 @@ It is a **compile-time error** if:
     const variable be augmented by another const variable, changing its value,
     or is that too weird?)**
 
+*   An `abstract` variable is augmented with a non-abstract variable, getter,
+    or setter.
```
sgrekhov, line 720: Github doesn't allow to comment lines 679-683.

> *   An augmenting declaration uses `augmented` when the augmented declaration
>     has no concrete implementation. Note that all external declarations are
>     assumed to have an implementation provided by another external source, and
>     otherwise they will throw a runtime error when called. An `abstract`
>     variable introduces no implementation.

Probably the last sentence here can be removed. If augmenting an abstract variable by non-abstract variable, getter or setter is an error itself we don't care about `augmented` expression into these augmentations any longer.

jakemac53: Agreed that seemed unnecessary, removed

munificent (approved):

Issue #3694, opened 2024-04-05
Change more elements of augmenting type declarations to be optional?


eernstg:
Consider the following program:

```dart
// --- Library 'main.dart'.
import augment 'augment.dart';

extension type ET(int i) {}

// --- Augmentation library 'augment.dart'.
augment library 'main.dart';

augment extension type ET(int i) implements int {}
```

This program could be considered well-formed because the augmenting type declaration (concretely, `extension type` declaration, but this kind of consideration would apply to classes, mixins, etc. as well) repeats the "header" of the declaration, and it is then able to add new members or augment existing ones by adding declarations to the body.

However, we could also consider the syntax `(int i)` in the header of the declarations of `ET` to be a syntactic abbreviation of an instance variable declaration and a constructor declaration, following the nature of the [primary constructor proposal](https://github.com/dart-lang/language/blob/main/working/2364%20-%20primary%20constructors/feature-specification.md). If we adopt this perspective then we would have a conflict because the original declaration of `ET` declares a final instance variable named `i`, and so does the augmenting declaration of `ET`, and that's a conflict even in the case where the two declarations are identical. Similarly for the constructor.

We could solve this problem by adding support for `(augment int i)`, which would mean that the declaration of `i` in the augmenting declaration of `ET` is an augmentation of the one in the initial declaration of `ET` (and it has no effect).

Alternatively, we could allow a declaration like `augment extension type ET implements int {}` which would correspond to a declaration of `ET` that does not contribute that final instance variable and constructor declaration, so there's no conflict.

Another example is augmenting `enum` declarations. We might want to allow those to omit the declarations of values, which would be useful in the case where the augmenting declaration is intended to provide additional instance or static members, but no new values are added, and no augmentation is performed on any of the existing members:

```dart
// --- Library 'main2.dart'.
import augment 'augment2.dart';

enum E {one}

// --- Augmentation library 'augment2.dart'.
augment library 'main2.dart';

augment enum E implements int {
  void someInstanceMethod() {}
  static void someStaticMethod() {}
}
```

This is currently a syntax error because the list of `<enumEntry>` declarations is required. We could specify a new syntax for an `augment enum` declaration such that this list is optional.

lrhn:
Allowing omitting values from enums is tricky, because an enum value declaration can look like a function declaration:
```dart
enum Baz with Qux {
  // Banana!
  foo(x);
  final int x;
  const Baz(this.x);
}
const x = 42;
mixin Qux {
[...]

PR #4003, opened 2024-07-31
update extension type augmentations to be more strict

jakemac53:
Closes https://github.com/dart-lang/language/issues/3694, will merge after https://github.com/dart-lang/language/pull/4002 for changelog consistency purposes.

I went for the simplest answer here, since I don't see a motivating use case for anything more complex.

- The "original" extension type declaration _must_ include the representation type.
- Augmentations of extension types _must not_ include the representation type.

There is a TODO for updating the grammar, I think the simplest option is to make the representation type optional for both regular extension type declarations and augmenting ones, and have the error introduced at a later stage. This would have better behavior for tooling so that the file can still be parsed (ie: the formatter can still format in the face of this error). But, that is just a suggestion and I don't feel strongly if we would prefer to make the grammar more complex and have special rules for each.

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -1,7 +1,7 @@
 # Augmentations
 
 Author: rnystrom@google.com, jakemac@google.com, lrn@google.com <br>
-Version: 1.27 (see [Changelog](#Changelog) at end)
+Version: 1.30 (see [Changelog](#Changelog) at end)
 
 Augmentations allow spreading your implementation across multiple locations,
 both within a single file and across multiple files. They can add new top-level
@@ -1372,6 +1372,11 @@ original documentation comments, but instead provide comments that are specific
 to the augmentation.
 
 ## Changelog
+
+### 1.30
+
+*   Simplify extension type augmentations, don't allow them to contain the
+    representation type at all.
 
 ### 1.27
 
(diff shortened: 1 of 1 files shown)


working/augmentation-libraries/feature-specification.md
```diff
@@ -924,13 +924,29 @@ It is a compile-time error if:
 When augmenting an extension type declaration, the parenthesized clause where
 the representation type is specified is treated as a constructor that has a
 single positional parameter, a single initializer from the parameter to the
-representation field, and an empty body.
-
-This means that an augmentation can add a body to an extension type's
-constructor, which isn't otherwise possible. *(But note that there is no
+representation field, and an empty body. This clause must be present on the
+declaration which introduces the extension type, and must be ommitted from all
```
munificent, line 928: Spelling: "ommitted" -> "omitted".

Also, "this clause" is ambiguous since the previous sentence lists a bunch of stuff. How about:

"This clause" -> "The representation field clause".

working/augmentation-libraries/feature-specification.md
```diff
@@ -941,6 +957,11 @@ looks and behaves much like one, and it cannot be augmented as such. It is a
 compile time error to have any augmenting declaration with the same name as the
 representation object.
 
+It is a compile time error if:
+
+*   An extension type augmentation contains a parenthesized clause after the type
+    name.
```
munificent, line 963: Technically, it could be after the extension type's type parameter list too. Maybe:

"contains a representation field clause".

munificent (approved):

Issue #3977, opened 2024-07-11
Allow a variable initializer to be unresolved, following the treatment of abstract members


eernstg:
It is not obvious from the [augmentations feature specification](https://github.com/dart-lang/language/blob/main/working/augmentation-libraries/feature-specification.md) whether or not it's possible for a non-local variable declaration consisting of an introductory and zero or more augmenting declarations can have an _unresolved_ initializing expression during augmentation.

If we do allow an unresolved initializing expression then we would allow the following:

```dart
int i;
augment int i = 42;
```

This augmenting declaration stack leaves the initializing expression of `i` unresolved at the introductory declaration, but it is resolved as `42` at the augmenting declaration. Of course, it would be an error if the augmenting declaration used `augmented` in the initializing expression, because it would be a reference to an initializing expression that doesn't exist.

But the point is that it would _not_ be an error to have a variable declaration with an unresolved initializer and a potentially non-nullable type, as long as the missing parts are added by augmentation.

This differs from the current behavior of the analyzer:

```
  error • n016.dart:1:5 • The non-nullable variable 'i' must be
          initialized. Try adding an initializer expression. •
          not_initialized_non_nullable_variable
```

However, the support for an unresolved initializer is in line with the fact that a concrete class can contain an abstract member declaration (that's a kind of "unresolved member implementation"):

```dart
// No errors, `a` can be abstract at first because it is concrete after augmentation.
class A { int get a; }
augment class A { augment int get a => 0; }
```

For a variable whose type is nullable we can choose to say that the introductory variable declaration with no initializing expression is considered to have the initializer `= null`. That's a different issue (#3725).

But when the type of the variable is potentially non-nullable then we can't invent an initializing expression in the case where it isn't physically present, and that's the reason for talking about "unresolved" initializing expressions.

It looks like the implementation (so far, it isn't anywhere near complete) does not allow a variable to have an unresolved initializer. This seems at least conceptually inconsistent with the specified and implemented treatment of abstract members. 

So I'd recommend that we _do_ allow those unresolved initializers to exist. However, I cannot see anything in the feature specification that confirms or contradicts this statement, so if we do decide to do this then we may need to add a short paragraph about it in the spec.

@dart-lang/language-team, WDYT? Can a variable initializer be abstract? ;-)

lrhn:
Not sure what "unresolved" means here.

As to whether we should allow:
```dart
int i;
augment int i = 42;
```
then yes, we should.

There should be no rule *preventing* it.

[...]

Issue #3725, opened 2024-04-25
Does implicit null count as an initializing expression?


eernstg:
Here's a corner case which is perhaps somewhat ambiguously specified. Thanks to @sgrekhov for bringing up this issue! 

Consider the following program:

```dart
// --- Library augmentation 'augment.dart'.
augment library 'main.dart';

String f(String? s) => s ?? 'Default';

augment String? sq = f(augmented); // OK?

// --- Library 'main.dart'.
import augment 'augment.dart';

String? sq; // Implicit initializer `= null`.
```

The augmentation specification has the following rule:

> It is a compile-time error to use `augmented` in an augmenting field's initializer if the member being augmented is not a field with an initializer.

So the question is whether or not the implicit initialization to null counts as an initializing expression? I'd recommend that it does not, and the line marked `OK?` is a compile-time error, for simplicity.

However, it could also be argued that some macros can be simpler if they are allowed to use `augmented` as shown above both in the case where there is an explicit initializing expression and in the case where the variable has an implicit initializer with the value null. The argument would be "It's useful, so why not?".

[Update: At this time I think "It's useful, so why not?" wins. I'd recommend that we use the semantic criterion "this variable has an initial value" rather than the syntactic "this variable declaration has an initializing expression". So `OK?` above should be `OK!`.]

@dart-lang/language-team, WDYT?

lrhn:
Yes *and* no! (Aka, "Why not both?", or "It depends!")

There is no good reason to disallow the example program here, it works. Which means that the rule "can't call `augmented` on variable with no initializer" is too strict. It should probably be allowed if the variable's type is nullable, in which case `augmented` evaluates to `null`.

That is: 
> If `augmented` refers to a variable declaration (as defined by a declaration and a number of prior augmentations) with no initializer expression, and the variable's type is nullable, `augmented` evaluates to `null`. _If the variable's type is not nullable, then it's a compile-time error._

But what is the type of the variable, if it has no declared type, yet?
```dart
final x = 42;
augment final x = augmented.toRadixString(16) * 1.5;
augment final num x;
```
Is this valid at all? If not, why not?

If valid, what is the static type of `augmented` here?
If it's `int`, then it's because we have inferred a *preliminary* type for `x`, because the type of `augmented` is the type of the declaration. If it has one?
[...]

PR #4005, opened 2024-07-31
clarify some behavior around initializers

jakemac53:
Closes https://github.com/dart-lang/language/issues/3977
Closes https://github.com/dart-lang/language/issues/3725

- Specify that the implicit null initialization of nullable fields does not happen unless no augmentation adds an initializer.
- Specify that variables which require an initializer do not have to define it initially - it may come from an augmentation.

To be merged after https://github.com/dart-lang/language/pull/4004 for changelog consistency

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -575,7 +575,8 @@ declaration, and if not `final`, also an abstract setter declaration. An
 `external` setter.  Unlike abstract declarations, they are considered to
 have a concrete implementation.
 
-Variables which require an initializer expression need not initially be defined
+Variables which require an initializer expression (such as those which have a
+non-nullable type and are not marked `late`) need not initially be defined
 with one, as long as there exists some augmentation which supplies it.
 
 Augmentations on variables, getters, and setters works mostly at the level of
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -1,7 +1,7 @@
 # Augmentations
 
 Author: rnystrom@google.com, jakemac@google.com, lrn@google.com <br>
-Version: 1.27 (see [Changelog](#Changelog) at end)
+Version: 1.32 (see [Changelog](#Changelog) at end)
 
 Augmentations allow spreading your implementation across multiple locations,
 both within a single file and across multiple files. They can add new top-level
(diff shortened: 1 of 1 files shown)


working/augmentation-libraries/feature-specification.md
```diff
@@ -575,6 +575,9 @@ declaration, and if not `final`, also an abstract setter declaration. An
 `external` setter.  Unlike abstract declarations, they are considered to
 have a concrete implementation.
 
+Variables which require an initializer expression need not initially be defined
+with one, as long as there exists some augmentation which supplies it.
```
munificent, line 579: It would be good to specify which variables require an initializer. I think it's basically just potentially-non-nullable non-late top-level variables and static fields?

munificent (approved):

working/augmentation-libraries/feature-specification.md
jakemac53: It doesn't really feel like the right place to try and enumerate the scenarios, but I will add that as an example, so that we aren't obligated to make it exhaustive or keep it up to date with language changes.

jakemac53:
@eernstg I am going to go ahead and merge this but you may want to take a look still, feel free to leave a comment later with any concerns.

working/augmentation-libraries/feature-specification.md
eernstg: I agree, we could just say 'Variables need not initially...'. We could extend that with commentary along the lines of 'Variables need not initially be defined with an initializing expression *(not even non-late variables with a potentially non-nullable type)*, as long...'.

working/augmentation-libraries/feature-specification.md
```diff
@@ -650,10 +654,11 @@ More specifically:
     usual, external and abstract variables cannot augment their
     initializing expression, since it does not exist.*
 
-    Augmenting initializer expressions replace the augmented initializer. The
-    augmenting initializer may use an `augmented` expression which executes the
-    augmented initializer expression when evaluated. If no initializer is
-    provided then the augmented initializer is not altered.
+    Augmenting initializer expressions replace the augmented initializer (or
+    provide one where none existed previously). The augmenting initializer
+    may use an `augmented` expression which executes the augmented initializer
```
eernstg, line 659:
```suggestion
    may use an `augmented` expression which evaluates the augmented initializer
```
The word 'evaluates' seems more informative than 'executes' in the case where the value of the expression is typically the only outcome of this step, and certainly an important outcome in any case.

working/augmentation-libraries/feature-specification.md
```diff
@@ -683,7 +688,9 @@ It is a **compile-time error** if:
 
 *   An augmenting variable’s initializing expression uses `augmented`, and
     the stack of augmented declarations do not include a variable with an
-    initializing expression.
+    explicit initializing expression. For nullable fields, the implicit null
```
eernstg, line 691: I think 'variables' is a better word than 'fields' here. 'Fields' is generally not used in the specification documents, presumably because 'instance variable' is a more orthogonal way to talk about them. This allows us to state rules about instance variables and static variables and library variables using just 'non-local variables' (or even just 'variables', in cases where 'non-local' can be left implicit).

eernstg (commented):
I added a couple of comments.

PR #4016, opened 2024-08-06
Change the syntax of augmenting extension type declarations

eernstg:
This PR changes the grammar such that an augmenting extension type declaration does not specify a primary constructor. This means that there is no `const` modifier (which is optional on the introductory declaration, but it is not allowed at all on an augmenting extension type declaration). Also, there is no "parameter list" (known as the `<representationSpecification>` in the grammar).

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -952,10 +952,10 @@ omitted from all augmentations of the extension type.
 
 This means that an augmentation can add a body to an extension type's implicit
 constructor, which isn't otherwise possible. This is done by augmenting the
-constructor in the body of the extension type *(But note that there is no
+constructor in the body of the extension type. *Note that there is no
 guarantee that any instance of an extension type will have necessarily executed
 that body, since you can get instances of extension types through casts or other
-conversions that sidestep the constructor)*. For example:
+conversions that sidestep the constructor.* For example:
 
 ```dart
 extension type A(int b) {
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -1400,7 +1400,7 @@ to the augmentation.
 ### 1.33
 
 *   Change the grammar to remove the primary constructor parts of an
-    extension type declaration when it is augmenting.
+    augmenting extension type declaration.
 
 ### 1.32
 
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -1134,37 +1134,39 @@ topLevelDeclaration ::= classDeclaration
 
 classDeclaration ::= 'augment'? (classModifiers | mixinClassModifiers)
     'class' typeWithParameters superclass? interfaces?
-    '{' (metadata classMemberDeclaration)* '}'
+    memberedDeclarationBody
   | 'augment'? classModifiers 'mixin'? 'class' mixinApplicationClass
 
 mixinDeclaration ::= 'augment'? 'base'? 'mixin' typeIdentifier
   typeParameters? ('on' typeNotVoidNotFunctionList)? interfaces?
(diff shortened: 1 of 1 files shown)


working/augmentation-libraries/feature-specification.md
```diff
@@ -1155,9 +1150,11 @@ extensionDeclaration ::=
...
 
 extensionTypeDeclaration ::=
-  'augment'? 'extension' 'type' 'const'? typeIdentifier
-  typeParameters? representationDeclaration interfaces?
-  '{' (metadata classMemberDeclaration)* '}'
+    'extension' 'type' 'const'? typeIdentifier
+    typeParameters? representationDeclaration interfaces?
+    '{' (metadata classMemberDeclaration)* '}'
+  | 'augment' 'extension' 'type' typeIdentifier typeParameters? interfaces?
+    '{' (metadata classMemberDeclaration)* '}'
```
munificent, line 1157: There's a little more redundancy here than I would like. In addition to the one in `classDeclaration`, there are now three places where we have:

```
'{' (metadata classMemberDeclaration)* '}'
```

How about:

```
typeDeclarationBody ::= '{' (metadata classMemberDeclaration)* '}'

extensionTypeHeader ::= 
    'extension' 'type' 'const'? typeIdentifier typeParameters?
    representationDeclaration
  | 'augment' 'extension' 'type' typeIdentifier typeParameters?

extensionTypeDeclaration ::= extensionTypeHeader interfaces? typeDeclarationBody
```

And then update `classDeclaration` to use `typeDeclarationBody` too?

---

Alternatively, we could do:

```
extensionTypeDeclaration ::=
    'augment' 'extension' 'type' 'const'? typeIdentifier
    typeParameters? representationDeclaration interfaces?
    '{' (metadata classMemberDeclaration)* '}'
```

And then outside of the grammar say:

* It's a compile-time error if an extension type marked `augment` is marked `const`.
* It's a compile-time error if an extension type marked `augment` has a `representationDeclaration`.
* It's an compile-time error if an extension type not marked `augment` does not have a `representationDeclaration`.

(In practice, I suspect this is how the parser maintainers will implement it, so that they get better error recovery.)

---

Or just keep what you have. It's certainly not intolerable that way either.

munificent (approved):

working/augmentation-libraries/feature-specification.md
```diff
@@ -1155,9 +1150,11 @@ extensionDeclaration ::=
 extensionBody ::= '{' (metadata classMemberDeclaration)* '}'
 
 extensionTypeDeclaration ::=
-  'augment'? 'extension' 'type' 'const'? typeIdentifier
-  typeParameters? representationDeclaration interfaces?
-  '{' (metadata classMemberDeclaration)* '}'
+    'extension' 'type' 'const'? typeIdentifier
+    typeParameters? representationDeclaration interfaces?
+    '{' (metadata classMemberDeclaration)* '}'
+  | 'augment' 'extension' 'type' typeIdentifier typeParameters? interfaces?
```
jakemac53, line 1156: Are we sure we want to separate the grammar here? This means that if you type:

```dart
augment extension type Foo(int x) {}
```

You will get a parse error. In my experience parse errors lead to extremely bad UX, for example the formatter will choke (on the entire file), and it is likely that everything after this error will give erroneous errors. The error you do get will also likely be something like "unexpected `(`". And not a more specific error saying the representation declaration is not allowed on augmentations of extension types (maybe even with a quick fix to remove it!).

eernstg: That shouldn't matter: The tools have always used a grammar that derives a bigger language than the specified grammar, because that yields a better UX (during recovery and for error messages). Surely they will continue to do so.

jakemac53: Ok, if the parser doesn't strictly follow the exact grammar then sgtm

jakemac53 (approved):

Issue #4017, opened 2024-08-06
How can we augment an abstract variable?


eernstg:
_tl;dr_&nbsp;&nbsp;&nbsp;Allow augmenting an abstract variable by a concrete getter and/or setter, or another abstract variable&nbsp;&nbsp;&nbsp;_rd;lt_

The augmentation specification has the following rules:

> An `abstract` variable declaration is equivalent to an abstract getter declaration, and if not final, also an abstract setter declaration.
>
> ... *abstract variables cannot augment their initializing expression, since it does not exist* ...
>
> It is a **compile-time error** if:
> - ...
> - An abstract variable is augmented with a non-abstract variable, getter, or setter.

*Note that there's nothing new about the first rule, this is also the way abstract variables are specified in the [feature specification](https://github.com/dart-lang/language/blob/main/accepted/2.12/abstract-external-fields/feature-specification.md). So it makes sense to expect that an abstract variable should be treated in exactly the same way as the corresponding abstract getter/setter pair, or at least very similarly.*

Does this mean that we cannot augment an abstract variable declaration at all .. other than by another abstract variable, or an abstract getter and/or setter, e.g., in order to add metadata or DartDoc comment?

This seems to be unnecessarily strict: If the abstract variable is indeed equivalent to an abstract getter and possibly an abstract setter then we should be allowed to augment the getter and/or setter just like any other getter and/or setter.
We do have this, in the section about augmentation of getters and setters:

> *The augmenting declarations can themselves be an implicitly induced getter or setter, or an explicitly declared one.*

We may even be allowed to augment the abstract variable with another variable declaration:

> **Augmenting a variable with a variable:** ... [not an error] ...

We should be able to use this to add metadata or DartDoc comments.

This is more than we can do with explicitly specified getters and setters:

> **Augmenting a getter and/or setter with a variable:** This is a compile-time error in all cases. 

It was mentioned (in commentary) that we cannot augment the initializing expression, but this is consistent with the perspective that the abstract variable is (precisely) an abstract getter and possibly an abstract setter, so it makes sense that we cannot augment the storage part of a (non-existing) variable.

In summary, I believe we should change the compile-time error mentioned above to the following:

> It is a **compile-time error* if:
> - ...
> - An abstract variable is augmented with a non-abstract variable.

@dart-lang/language-team, WDYT?

jakemac53:
I recently updated this section, see https://github.com/dart-lang/language/pull/4000 for the changes.

I think the current status is what we want.

[...]

PR #4023, opened 2024-08-07
Remove some errors around abstract and external variables

jakemac53:
Closes https://github.com/dart-lang/language/issues/4017

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -722,11 +722,9 @@ It is a **compile-time error** if:
     const variable be augmented by another const variable, changing its value,
     or is that too weird?)**
 
-*   An `abstract` variable is augmented with a non-abstract variable, getter,
-    or setter.
+*   An `abstract` variable is augmented with a non-abstract variable.
 
-*   An `external` declaration is augmented with an `abstract` declaration. For
-    variables this also applies to the implicit getter and setter.
+*   An `external` variable is augmented with an `abstract` variable.
 
 ### Augmenting enum values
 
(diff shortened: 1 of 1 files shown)


working/augmentation-libraries/feature-specification.md
```diff
@@ -722,11 +722,9 @@ It is a **compile-time error** if:
     const variable be augmented by another const variable, changing its value,
     or is that too weird?)**
 
-*   An `abstract` variable is augmented with a non-abstract variable, getter,
-    or setter.
+*   An `abstract` variable is augmented with a non-abstract variable.
 
-*   An `external` declaration is augmented with an `abstract` declaration. For
-    variables this also applies to the implicit getter and setter.
+*   An `external` variable is augmented with an `abstract` variable.
```
eernstg, line 727: Even that could be allowed and make sense:

```dart
external int x;

@metadata
augment abstract x;
```

Arguably, to avoid redundancy, we need to use an abstract variable because this allows us to associate the metadata with both the getter and the setter of the definition of `x`. Alternatively, we would do this:

```dart
external int x;

@metadata
augment get x;
@metadata
augment set x(_);
```

We could also repeat the `external` declaration, but this seems more confusing to me. "Does this mean that we have two native implementations of the getter and setter, and we're ignoring the first one?":

```dart
external int x;

@metadata
augment external x;
```

We could do that, but it seems cleaner to me if we use `abstract` which is the "smallest" possible declaration of a variable: No getter implementation, no setter implementation, no storage, no initializer, just a way to refer to the declaration with the given name `x`. Used in this manner, an `abstract` variable will work for the purpose of adding metadata (and the like) with a concrete variable, with a property which was originally expressed as a getter and a setter, with an external variable, the works.

eernstg (commented):
LGTM!

I'm suggesting that we could go even one step further.

eernstg (approved):

working/augmentation-libraries/feature-specification.md
jakemac53: > @metadata
> augment abstract x;

I see, I think I probably would prefer that users just repeat `external` in this case? Otherwise it seems like you are changing the field from external to abstract, which are different.

eernstg: Right ... I tried to describe my mental model about these things here: https://github.com/dart-lang/language/issues/4017#issuecomment-2276228341. It would require certain rules to change (mostly: things would be allowed that are currently specified as errors), but I think it's nice to have a simple and consistent way to understand the semantics of any given stack of declarations for a given property.

PR #4058, opened 2024-08-26
Address extension accessibility and specificity

lrhn:
Makes extension imported by part files be accessible inside the part file, not to the entire library. (Existing phrasing didn't need to  distinguish the two.)

Makes closer imports in the parent-file chain be more specific than further away imports.

This ensures that a part file that imports all its own dependencies, including implicitly referenced extensions, can ignore parent imports and the extensions those parent imports make accessible, and not risk a conflict or override by an equally or more specific extension.

### DIFF
--- a/working/augmentation-libraries/parts_with_imports.md
+++ b/working/augmentation-libraries/parts_with_imports.md
@@ -323,7 +323,7 @@ of the [extension feature specification on accessibility][] to:
 >   not mentioning the name.
 >   In that case the extension is accessible with priority 1.
 >   _This includes non-deferred imports with a prefix, and even an unnamed
->   prefix using the the "wildcard variable" unnamed variable notation
+>   prefix using the "wildcard variable" unnamed variable notation
 >   `as _`._
 > * Or otherwise, if the Dart file is a part file and the extension is
 >   accessible by import in the part file's parent file with priority *n*.
@@ -331,7 +331,7 @@ of the [extension feature specification on accessibility][] to:
 >   with priority *n + 1*. _(Lower is better.)_
 
 The "priority *n*" of an extension for a particular Dart file is a measure
-of how far up in the parent chain the import for that extension occurred.
+of how far up in the `part of` chain the import for that extension occurred.
 A value of "1" means the current Dart file, and each step up the parent
 chain increases the value by one. A value of zero means the declaration
 is in the current library, and that is the priority of the extension
--- a/working/augmentation-libraries/parts_with_imports.md
+++ b/working/augmentation-libraries/parts_with_imports.md
@@ -1,7 +1,7 @@
 # Part files with imports
 
 Authors: rnystrom@google.com, jakemac@google.com, lrn@google.com <br>
-Version: 1.0 (See [Changelog](#Changelog) at end)
+Version: 1.1 (See [Changelog](#Changelog) at end)
 
 This is a stand-along definition of _improved part files_, where the title of
 this document/feature is highlighting only the most prominent part of the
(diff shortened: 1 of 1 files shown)


jakemac53 (approved):

working/augmentation-libraries/parts_with_imports.md
```diff
@@ -296,6 +296,118 @@ library, so the library author should fix the conflict by renaming the prefix.
...
+> An extension is accessible by import with priority *n* in a Dart file iff:
+> * The Dart file *imports* the extension, meaning that the Dart file
+>   contains a non-deferred import directive whose imported
+>   library has the extension in its export scope, and where the name of
+>   the extension is not private, and is not hidden by either a `hide`
+>   combinator mentioning the extension name or by a `show` combinator
+>   not mentioning the name.
+>   In that case the extension is accessible with priority 1.
+>   _This includes non-deferred imports with a prefix, and even an unnamed
+>   prefix using the the "wildcard variable" unnamed variable notation
```
munificent, line 326: Nit: "the the" -> "the".

working/augmentation-libraries/parts_with_imports.md
```diff
@@ -296,6 +296,118 @@ library, so the library author should fix the conflict by renaming the prefix.
...
+>   `as _`._
+> * Or otherwise, if the Dart file is a part file and the extension is
+>   accessible by import in the part file's parent file with priority *n*.
+>   In that case the extension is accessible by import to the current file
+>   with priority *n + 1*. _(Lower is better.)_
+
+The "priority *n*" of an extension for a particular Dart file is a measure
+of how far up in the parent chain the import for that extension occurred.
```
munificent, line 334: Maybe: "parent" -> "part of"?

working/augmentation-libraries/parts_with_imports.md
```diff
@@ -296,6 +296,118 @@ library, so the library author should fix the conflict by renaming the prefix.
...
+> Let *i* be a member invocation with target expression `e` and corresponding
+> member name *m*, and let *E1* and *E2* denote different accessible and
+> applicable extensions for *i*.
+> 
+> Whether *E1* is more specific than *E2* wrt. *i* of *e* is determined by
+> the following steps, stopping when an answer is given:
+> * If *E1* is accessible with priority *n1* &gt; 0 and *E2* is accessible
+>   with priority *n2* &gt; 0 _(so both accessible by import)_ then
+>     * if *n1* < *n2* then *E1* is more specific than *E2*, and
+>     * if *n2* < *n1* then *E1* is *not* more specific than *E2*.
```
munificent, line 364: Is there a reason this is phrased as "E1 is not more specific than E2" instead of saying "E2 is more specific than E1"?

It reads to me like you're trying to define a partial order, but I don't quite see why when `n2 < n1` in this clause. I might be missing some context.

munificent (approved):
I have vague misgivings about the complexity of this. In general, it seems like when we allow "shadowing-like" behavior and try to resolve it with some prioritized choice, we often end up with something that feels brittle and confusing.

But the motivation here seems reasonable to me, so it's probably fine.

working/augmentation-libraries/parts_with_imports.md
lrhn: It is a partial order. And even more, this text is *part* of a partial order specification.
There can be distinct extensions where neither is more specific than the other. If they are in the set of applicable extensions, then lookup fails.

Saying that *E2* is more specific than *E1* would not answer whether *E1* is more specific than *E2* or not. 
That would be assuming that the relation is anti-symmetric. (It probably is, but that's a consequence of the definition deliberately avoiding any mutual more-specific relations, and this rule is part of making that so, so it can't also assume it.)

The relation being defined is "*E1* more specific than *E2*" which should give an answer of true or false.
The items are in priority order, and this item gives an answer of "yes" in one case and "no" in another, and otherwise it falls through to the next item (hence "part of a partial order").
Answering any other question doesn't help. (But the rules do imply that if *E1* is not more specific than *E2* *for this reason*, then *E2* is more specific than *E1*, since it's the same *n* values.)

lrhn:
Did not do it this way.

PR #4065, opened 2024-08-29
Move unquoted-imports to accepted and import-shorthand to archive.

munificent:
I left forwarding stubs like we generally do (though at some point, we should probably clean out the ancient ones).


lrhn (approved):
Ship it!

lrhn:
Wouldn't it be nice if a site based on a versioning tool could automatically detect that someone was accessing a file that no longer exist in the current version, and tell them where it was moved to, without needing to add an actual entry? The information is surely available. :wink:

PR #4083, opened 2024-09-05
Update typos and add info in unquoted imports and wildcard variables …

kallentu:
Just a small typo + some additional small changes to these two specs.


munificent (approved):
Thanks!

PR #4095, opened 2024-09-19
Revise shared memory multithreading proposal

mraleph:
This revises shared memory multithreading proposal based on discussions we had internally and current implementation direction. 

The biggest change is complete removal of `Shareable` type marker - instead we accept that most likely destination for us would be _shared **anything** multithreading_.  

Another change is that proposal now describes in more details the intermediate _shared **native** multithreading_: which does not introduce any fundamentally new capabilities (only native memory can truly be shared), but at the same time addresses a lot of ergonomics issues around interoperability and concurrency model mismatch between Dart and native.

fyi @a-siva @johnmccutchan

### DIFF
--- a/working/333 - shared memory multithreading/proposal.md
+++ b/working/333 - shared memory multithreading/proposal.md
@@ -32,29 +32,45 @@ for concrete code examples.
 > in the Dart SDK. For example, we will be able to move `dart:io` implementation
 > from C++ into Dart and later split it into a package.
 
-The core of this proposal are two new concepts:
+This proposal:
 
-- [Shareable Data](#shareable-data) introduces the relaxation of isolate model
-  allowing controlled sharing of mutable data and static state between isolates
-  within an isolate group. This allows developer to write Dart code which can
-  access shared mutable state concurrently.
-- [Shared Isolates](#shared-isolates) introduces a concept of _shared isolate_,
-  an isolate which only has access to a state shared between all isolates of the
-  group. This concept allows to bridge interoperability gap with native code.
-  _Shared isolate_ becomes an answer to the previously unresolved question
-  _"what if native code wants to call back into Dart from an arbitrary thread,
-  which isolate does the Dart code run in?"_.
+- allows developer to selectively break isolation boundary between isolates by
+declaring some static fields to be [_shared_](#shared-fields) between isolates
+within isolate group.
 
-In addition to introducing these new concepts the proposal tries to suggest a
-number of API changes to various core libraries which are essentially to making
-Dart a good multithreaded language. Some of this proposals, like adding
-[atomics](#atomic-operations), are fairly straightforward and non-controversial,
-others, like [coroutines](#coroutines), are included to show the extent of
+- introduces the concept a concept of [_shared isolate_](#shared-isolates),
+an isolate which only has access to a state shared between all isolates of the
+group. This concept allows to bridge interoperability gap with native code.
+_Shared isolate_ becomes an answer to the previously unresolved question
+_"what if native code wants to call back into Dart from an arbitrary thread,
+which isolate does the Dart code run in?"_.
+
+> [!NOTE]
+>
+> Earlier version of this proposal was also introducing the concept of
+> _shareable data_ based on a marker interface (`Shareable`) which
+> required developer to explicitly opt in into shared memory multithreading
+> for their classes. In this model only instances of classes which implement
+> the marker interface could be shared between isolates.
+>
+> Based on the extensive discussions with the language team and implementors
+> I have arrived to the conclusion that this separation does not have clear
+> benefits which are worth the associated implementation complexity.
+> Consequently I remove this concept from the proposal and instead propose
+> that we eventually allow unrestricted _share everything_ multithreading
+> within the isolate group.
+
+Additionally the proposal tries to suggest a number of API changes to various
+core libraries which are essentially to making Dart a good multithreaded
+language. Some of this proposals, like adding [atomics](#atomic-operations),
+are fairly straightforward and non-controversial, others, like
+[coroutines](#coroutines), are included to show the extent of
 possible changes and to provoke thought.
 
-The [Prototyping Roadmap](#prototyping-roadmap) section of the proposal tries to
-suggest a possible way forward with validating some of the possible benefits of
-the proposal without committing to specific major changes in the Dart language.
+The [Implementation Roadmap](#implementation-roadmap) section of the proposal
+tries to suggest a possible way forward with validating some of the possible
+benefits of the proposal without committing to specific major changes in the
+Dart language.
 
 ## What issues are we trying to solve?
 
@@ -262,7 +278,7 @@ all different solutions fall short.
 
 Before we discuss our proposal for Dart it is worth look at what other popular
 and niche languages do around share memory multithreading. If you feel familiar
-with the space feel free to skip to [Shareable Data](#shareable-data) section.
+with the space feel free to skip to [Shared Isolate](#shared-isolate) section.
 
 **C/C++**, **Java**, **Scala**, **Kotlin**, **C#** all have what I would call an
 unrestricted shared memory multithreading:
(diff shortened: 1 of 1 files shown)


working/333 - shared memory multithreading/proposal.md
```diff
@@ -842,9 +594,7 @@ class Isolate {
   /// Shared isolate contains a copy of the
```
lrhn, line 594: Can a shared function access *local* variables? Are they *shared*?
That is, if I have:
```dart
shared void Function() inc = (){};
shared int Function() val = () => 0;
void initInc() {
  int counter = 0;
  inc = () { counter++; };
  val = () => counter;
}
```
then will it compile? Will it run? Will concurrent accesses update the same variable?

(If you *can't* share closures containing local variables, then some things become much harder. Most things.
Allowing *final* local variables might ease that, but you do want shared mutable state, so it's just a question of whether the compiler boxes the variable or the user has to.)

working/333 - shared memory multithreading/proposal.md
```diff
@@ -1096,43 +793,6 @@ All built-in asynchronous primitives will make the following API guarantee: **a
 callback passed to `Future` or `Stream` APIs will be invoked using executor
 which was running the code which registered the callback.**
```
lrhn, line 794: Can you share futures or streams at all? (They're not trivally shareable, so likely not in the initial proposal.)

If you can, then we'll have solved the problem of `Zone.current` being non-shared state, and every operation on `Future` accessing `Zone.current`. That's not going to be trivial.

(Maybe we do want some non-shared global variables that are made accessible, so we can initialize them if necesssary..)

working/333 - shared memory multithreading/proposal.md
```diff
@@ -1195,7 +855,7 @@ to control threads.
 ```dart
 // dart:concurrent
 
-abstract class Thread implements Shareable {
+abstract class Thread {
```
lrhn, line 858: The below:
> /// ... which uses the
>  /// spawned thread as an executor for all callbacks: this
>  /// means the thread will remain alive as long as there is
>  /// a callback referencing it.

I'm not sure what that means. Are the callbacks `Zone`-functions, like `scheduleMicrotask`, or are they the callback arguments to those zone functions?

Are executors part of zones, or separator?
(If zones are shareable, and you can share the value with a function run with a different `Executor.schedule`, then it seems the executor is not bound by the zone or vice-versa.)

working/333 - shared memory multithreading/proposal.md
```diff
@@ -1272,7 +932,7 @@ throw.
 ```dart
```
lrhn, line 932: Why can't an `AtomicRef` contains not-real-ref-types? Seems like forcing boxing should be sufficient to make it work. Also, if you can store the value in a singleton unmodifiable list, and store the list in the `AtomicRef`, it feels like a distinctions without a real difference.
At least if we have Java-like memory model semantics for final fields, then:
```dart
class LockBox<T> {
  final T value;
  LockBox(this.value);
}
shared AtomicRef<LockBox<int>> _box = LockBox(0);
int get atomicInt => _box.load().value;
set atomicInt(int value) { _box.store(LockBox(value)); }
```
should give you a guaranteed store of an integer.

working/333 - shared memory multithreading/proposal.md
```diff
@@ -1583,22 +1242,13 @@ able to implement proposed semantics on top of it.
...
-> [`ShareableBox`](#shareableboxt)-like wrapper, which can be implemented on top
-> of TLS storage and `WeakMap`.
->
-> An alternative could be to tweak semantics of Wasm's `externref` a bit: tag
-> `externref` with their origin and dynamically checking origin match when
-> `externref` is passed back from Wasm to the host environment (see
-> [Dynamic sharedness checks as an escape hatch](https://github.com/WebAssembly/shared-everything-threads/issues/37)
-> issue).
+> would need to resolve this conundrum by employing some sort of thread local
+> wrapper, which can be implemented on top of TLS storage and `WeakMap`.
```
lrhn, line 1251: If we can have a single object representing each `Thread`, then a `ThreadLocal` is just an `Expando<Thread>`. 
That does mean not cloning `Thread` when communicating with other isolates.

working/333 - shared memory multithreading/proposal.md
```diff
@@ -32,29 +32,45 @@ for concrete code examples.
...
+This proposal:
+
+- allows developer to selectively break isolation boundary between isolates by
+declaring some static fields to be [_shared_](#shared-fields) between isolates
+within isolate group.
+
+- introduces the concept a concept of [_shared isolate_](#shared-isolates),
+an isolate which only has access to a state shared between all isolates of the
+group. This concept allows to bridge interoperability gap with native code.
+_Shared isolate_ becomes an answer to the previously unresolved question
```
aam, line 44: I think the original answer to this question is that this arbitrary thread should "enter isolate" if it wants to run dart code on this thread.
I think "shared isolate" is an answer to the question of "what if native code wants to run dart code without access to any isolate state, with access to only isolate group state".

working/333 - shared memory multithreading/proposal.md
```diff
@@ -477,317 +493,53 @@ shared global state.
...
-
-class Y<T extends Shareable> implements Shareable {
-  T v;  // ok.
-}
-```
+> It might seem strange to include mutable types like `TypedData` into trivially
+> shareable, but in reality allowing to share these type does not actually
+> introduce any fundamentally new capabilities. A `TypedData` instance can
+> already be backed by native memory and as such shared between two
+> isolates.
```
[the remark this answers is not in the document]
mraleph: I am not sure I understand what you are saying. One should already expect that it _can_ be accessed concurrently. We just don't mention this consideration anywhere.

working/333 - shared memory multithreading/proposal.md
```diff
@@ -477,317 +493,53 @@ shared global state.
...
-shareable declared type**. Incorporating this property into the type system
-naturally leads to the desire to use _intersection types_ to express the
-property that some value is both a function of a specific type _and_ shareable:
+> Types like `SendPort` are not `final` so strictly speaking we can't make a
+> decision whether an instance of `SendPort` is trivially shareable or not
+> based on the static type alone. Instead we must dynamically check if
+> `SendPort` is an internal implementation or not. Similar tweak should probably
+> be applied to the specification of `@pragma('vm:deeply-immutable')`
+> allowing classes containing `SendPort` fields to be marked `deeply-immutable`
+> at the cost of introducing additional runtime checks.
```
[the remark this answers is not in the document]
mraleph: Will document that checks happen when the object is created.

working/333 - shared memory multithreading/proposal.md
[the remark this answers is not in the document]
mraleph: > Is this correct?

That is correct.  

> If so, still sounds very restrictive. 

It depends from which angle you are looking at this: if you are purely looking from the perspective of writing pure-Dart in its single-threaded incarnation - yes, it is very restrictive. But there is a different perspective here: this approach actually _lifts_ restrictions which make it impossible to cover interoperability use-cases described at the start of this document in pure Dart. So despite being restrictive in terms of what Dart programmer could actually write - it actually enables Dart programmer to write express more things.

working/333 - shared memory multithreading/proposal.md
```diff
@@ -890,17 +640,10 @@ shared void foo() {
 - Within a `shared` function
   - If `f(...)` is a static function invocation then `f` must be a `shared`
     function.
-  - If `o.m(...)` is an instance method invocation, then `o` must be a subtype
-    of `Shareable` and `m` must be `shared` method.
-  - If `C(...)` is a constructor invocation then `C` must be a shareable class.
+  - If `o.m(...)` is an instance method invocation, then `m` must be `shared` method.
+  - If `C(...)` is a constructor invocation then `C` must be a class with a `shared` constructor.
```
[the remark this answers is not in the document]
mraleph: Removed most of this section.

working/333 - shared memory multithreading/proposal.md
```diff
@@ -930,9 +673,6 @@ class NativeCallable<T extends Function> {
   /// When the native code invokes the function [nativeFunction], the corresponding
   /// [callback] will be synchronously executed on the same thread within a
   /// shared isolate corresponding to the current isolate group.
```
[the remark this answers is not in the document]
mraleph: Switched to `the` though it is unclear at this stage whether or not we will have a single shared isolate or multiples of those.

I think it should be more or less unobservable.

working/333 - shared memory multithreading/proposal.md
mraleph: In the _shared **native** memory multithreading_ this code will not compile - because function is not a trivially shareable type. This also means that crossing isolation boundaries will still imply deep copying of objects: e.g. posting a task from a full isolate to a shared isolate will imply copying of the task.

working/333 - shared memory multithreading/proposal.md
mraleph: You can't share them in the SNMmt (_shared **native memory** multithreading_) but you can share them in SEmt (_shared **everything** multithreading_).

working/333 - shared memory multithreading/proposal.md
mraleph: I think what this comment is trying to say is that if you have some sort of branching asynchronous execution (e.g. you do `future.then(cb)`) then `cb` will run on the spawned thread. 

Maybe this is just bad API: if this behavior is desired users should use executors explicitly.

working/333 - shared memory multithreading/proposal.md
mraleph: I wanted to partition types which behave like value types from those that are reference types to make performant implementation simpler. I also want don't boxing and surprising behaviors which can arise from it: consider the semantics of `compareAndSwap` for example which is defined in terms of `identical`.

working/333 - shared memory multithreading/proposal.md
mraleph: > I think the original answer to this question is that this arbitrary thread should "enter isolate" if it wants to run dart code on this thread.

In the current model Dart code can only run in some isolate, so it is given that native must enter some isolate to be able to execute Dart - that's not the question. The question is really _which_ isolate. There are different answers to this question we have explored over the years: 

- One could answer _same isolate that created callback_. But this answer is unsatisfactory because one can't simply enter an isolate if it is busy: you either need to use asynchronous communication or introduces blocking-and-waiting for the isolate to become free.
- One could answer _fresh temporary isolate created for the purpose_. This answer is also unsatisfactory because there is no uniform answer for the story of static state (and event loop) for such temporary isolate.

Shared isolate gives an answer which is simple to explain and has a very clear semantics.

aam: "introduces the concept a concept of" -> "introduces the concept of"

One thing that seems a little unsettling is that shared isolate is not actually an isolate. The dart code that runs on them really runs in the context of containing isolate group, so for example, proposed name for [new native callback constructor](https://github.com/dart-lang/sdk/issues/54530) is `NativeCallable.isolateGroupShared`.

lrhn: The shared isolate is an isolate. It's just an isolate that runs code *slightly different* from the code of the other isolates in the group (in particular, every global variable throws on read or write instead of what the source code says). That also makes it *stateless* so it can be shared by multiple executors, without anybody being able to see it.

So, as I understand it, the isolate provides the global state and the executor provides that stack state and likely an event loop, which is the dynamic, temporary, execution state of running code. All running code has both. Today they are the same thing.

No non-shared isolate has more than one executor at the same time (today, unless we get real concurrency).

A `ReceivePort` belongs to both an isolate and the executor that created it. Anything that can schedule or create events belong to an executor, and cannot be shared between different executors.

I think we'll want *executor-local* "global" state, not just isolate-local state and isolate-group state (shared globals). For example, `Zone._current` should probably be executor-local because it reflects position in the current stack, which code is currently executing. And `Executor.current` should work in all isolates, even shared isolates, so it's "not state"? Or it's executor-local state, initialized to an object representing the current executor.).

lrhn: > Shared isolate gives an answer which is simple to explain and has a very clear semantics.

Maaaaybe.
The semantics are very simple. Any code that tries to read or write a global or static non-`const` variable will throw instead.

Being simple doesn't necessarily make it clear or easy to reason about, because it's impossible to see on a function whether it, or anything it calls recursively, may access a global variable. 

Something like `print`, `Iterable.toString`, `BigInt(1)` or everything in `dart:async` does access global state today. `ByteData` fails if it needs `Endian.host`. You can't call an `async` function, it throws ... at some point. Maybe immediately, maybe at the first `await`, depending on when it tries to create the `Future` it returns, which isn't specified today because creating a `Future` object is not expected to be able to fail.

In almost all of these examples, the global state can either be shared (isolate-group-wide, has the same value in every isolate, and doesn't contain anything non-shareable) or executor local (caches a value that is used in a single threaded way, but it's OK that it's uninitialized in each new executor).

I'm still worried that it will be a breaking change to add an access to a global variable to any existing code.
That's not great. I'd go as far as saying that we should not have *any* non-shared non-executor-local global variables at all. Choose one. If you choose nothing, it's executor local by default. Which just means that every normal global/static variable is executor-local, not isolate local, and you can make them isolate-group-shared, and isolates no longer carry state. At which point an executor *is* an isolate, and we've come full circle.

So maybe we don't need an executor. We just need to use the isolates we have today, and the "shared isolate" is just a fresh isolate.

working/333 - shared memory multithreading/proposal.md
```diff
@@ -1501,29 +1161,29 @@ access the value.
...
 
-- Introduce `Shareable` interface and enforce suggested restrictions using a
-  Kernel transformation instead of incorporating them into CFE.
-- Introduce minimum amount of core library changes (e.g. `ShareableList`).
+These changes do not significantly change the shape of Dart programming
+language, they streamline the interoperability with native code but do not
+introduce any new fundamental capabilities: developers can already share
+native memory between isolates and that simply makes such sharing more
+convenient to use. There is no sharing of mutable Dart objects at this stage
+yet.
```
mkustermann, line 1177: Actually a very common use case in C interop will be to pass a closure to `NativeCallable.shared()` and make that closure close over arbitrary state. We'd want to use the same closure (and not necessarily deeply clone it) thereby possibly sharing mutable objects.

(If we do copy the closure things get complicated, because the callback may be executed on N threads in parallel then we'd need N copies of the closure - i.e. it's not sufficient to deeply clone the closure once at `NativeCallable.shared()` call time but we'd need to do that every time it's invoked)

Would it make sense to have this API under the experimental flag?

working/333 - shared memory multithreading/proposal.md
```diff
@@ -477,317 +493,53 @@ shared global state.
...
->          _values[_token] = v;
->        });
->     }
-> }
+> Shared field reads and writes are _atomic_ for reference types, but other
+> than that there are no implicit synchronization, locking or strong memory
+> barriers associated with shared fields. Possible executions in terms of
+> observed values will be specified by the Dart's [memory model](#memory-models)
+> which I propose to model after JavaScript's and Go's: **program which is
+> free of data races will execute in a sequentially consistent manner**.
```
mkustermann, line 539: Maybe it would make sense to describe here whether we'd allow or disallow initializers (or maybe it's somewhere in this markdown?)

* Do we allow initializer expressions?
* When concurrent mutators access it in parallel would we execute it N times and a undefined one will win?

mkustermann (approved):
Generally this is exactly what I've been advocating for (mainly the removal of the type system changes, i.e. the `Shared` interface).

Amongst my reasons behind supporting the removal of `Shared` (a type system level concept) are:

* It would take long time to migrate the ecosystem packages (possibly many years)
* Many use cases would actually have ownership of an object graph and want to transfer that object graph ownership to another isolate in O(1).  In order to send the object graph one would need to mark it's classes all as `Sharable`
  => A class being `Sharable` **does not imply it's thread safe to use** on concurrent threads
  => If we make type system changes then at least one would expect to get concurrency safety from it, but a `Sharable` will not give any guarantees at all
* A simple use case like having a helper thread to decode json (that sends decoded json back to main thread) wouldn't even be solved by this because json decoder returns mutable `Map`s/`List`s that don't implement `Sharable`
  => I'd want any shared memory multithreading to support decoding json off-thread

working/333 - shared memory multithreading/proposal.md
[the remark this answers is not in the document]
lrhn: I'm looking at it from the perspective of someone who wants to use the feature, and who wants to know which library features I can't safely use.
I can't tell you. If we change nothing in the libraries I can mention a bunch of things that won't work (`Uri.parse`, `List.toString`, `jsonEncode`, `print` and everything in `dart:async`).

It will take serious refactoring to make parts of the platform libraries useful in a state where global variables throw, and I'm not entirely sure what to refactor it to.

(Could we have a period-of-concept version of such an isolate, something you can spawn using an unpublishable `spawnShared` function that works like `Isolate.spawn`? Just to see how much things break.)

working/333 - shared memory multithreading/proposal.md
```diff
@@ -842,9 +594,7 @@ class Isolate {
   /// Shared isolate contains a copy of the
   /// global `shared` state of the current isolate and does not have any
   /// non-`shared` state of its own. An attempt to access non-`shared` static variable throws [IsolationError].
-  ///
-  /// If [task] is not [Shareable] then [ArgumentError] is thrown.
-  static Future<S> runShared<S extends Shared>(S Function() task);
+  static Future<S> runShared<S>(S Function() task);
```
brianquinlan, line 597: Could you explain a bit more about how the ffi interop would actual work in this model? In particular, do callbacks running as "shared" capture their closure context?

I'm thinking about use cases like this:

You want to implement an Objective-C delegate in Dart to limit the number of HTTP redirects in your HTTP library. The delegate method looks like this and can be invoked on any thread:

```objc
- (void)URLSession:(NSURLSession *)session 
              task:(NSURLSessionTask *)task 
willPerformHTTPRedirection:(NSHTTPURLResponse *)response 
        newRequest:(NSURLRequest *)request 
 completionHandler:(void (^)(NSURLRequest *))completionHandler;
```

Your Dart code could look something like:

```dart
Future<String> doGet(Uri uri, int maxRedirects) {
  ...
  var numRedirects = 0;
  delegate.URLSession_task_willPreformHTTPRedirection_... = 
        (session, task, response, request, completionHandler) {
    numRedirects += 1; // Is this accessable?
    if (numRedirects <= maxRedirects) {
      ...
    } else {
      ...
    }
  }
}
```

Would `numRedirects` be available in the callback? If not, how would you make that sort of non-global state available?

working/333 - shared memory multithreading/proposal.md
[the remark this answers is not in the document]
mkustermann: 👍 To start experimenting with this and putting it behind an API that only works if the experiment is on.

working/333 - shared memory multithreading/proposal.md
brianquinlan: It looks like @aam 's prototype of `isolateGroupShared` does do closure capture so that would solve this problem.

working/333 - shared memory multithreading/proposal.md
mraleph: Ack. Note TLS here refers to Wasm's TLS capabilties which are part of the shared-everything-threads proposal linked above.

working/333 - shared memory multithreading/proposal.md
mraleph: > Being simple doesn't necessarily make it clear or easy to reason about, because it's impossible to see on a function whether it, or anything it calls recursively, may access a global variable.

That is fair. I honestly don't know what is the good solution here: it seems costly to try lifting this segregation into the type-system so that you could statically know at the call site whether or not the invocation is valid in the context of shared isolate.

> Something like `print`, `Iterable.toString`, `BigInt(1)` or everything in `dart:async` does access global state today. `ByteData` fails if it needs `Endian.host`. You can't call an `async` function, it throws ... at some point. Maybe immediately, maybe at the first `await`, depending on when it tries to create the `Future` it returns, which isn't specified today because creating a `Future` object is not expected to be able to fail.

As I see it we will over time fix core libraries to allow most of these things to work in the shared isolate.

> We just need to use the isolates we have today, and the "shared isolate" is just a fresh isolate.

Fresh isolates don't fit here because you don't want them to have state (and event loop). When you invoke Dart code on some thread that is a transient action, thread might never call Dart code again. When is this isolate spawned and when does it get destroyed? What happens if multiple callbacks are called concurrently? What happens if they are called sequentially but on the same thread? There are so many different ways to answer these questions when you allow isolates to have _isolated_ state - which this state comes need for proper lifecycle and I really want to avoid that. 

That being said there is nothing preventing us from allowing developer to explicitly make specific choices here, e.g. imagine we have something like this:

```dart
class Isolate {
  // Runs the given function in the given isolate.
  //
  // This method blocks and waits for isolate to become available.
  Isolated<T> runSync<T>(T Function() body);  
}

// This represents a value produced in the isolate, if it is
// accessed from another isolate it is copied if it is not 
// trivially shareable.
class Isolated<T> {
  T get value;
}
```

Then from the shared isolate you can always execute code inside some other isolate which you decided to create and run for these specific purposes. You can also use other means to switch to another isolate: e.g. you can post a task to another isolate.

working/333 - shared memory multithreading/proposal.md
mraleph: Actually now that I think about it disallowing `AtomicRef<int>` does not disallow `AtomicRef<Object> ref; ref.compareAndSwap(10.0, 20.0)` which requires implementation to peak into the value. I think I need to define that behavior of `compareAndSwap` is undefined for primitives like `int` and `double`.

working/333 - shared memory multithreading/proposal.md
mraleph: I have added the following restriction to `NativeCallable.shared`:

> ```dart
>  /// Throws [ArgumentError] if [callback] captures state which can't be
>  /// transferred to shared isolate without copying.
>  external factory NativeCallable.shared(...);
> ```
> ...
> In _shared **everything** multithreading_ world `callback` can be allowed to
> capture arbitrary state, however in _shared **native memory** multithreading_
> this state has to be restricted to trivially shareable types

I would like to avoid creating mutable shared _Dart_ state between initially.

working/333 - shared memory multithreading/proposal.md
mraleph: I think this code is okay because integers are trivially shareable. 

See my answer to @mkustermann above: https://github.com/dart-lang/language/pull/4095#discussion_r1778662388

However if the state is more complex and not trivially shareable then callback would need to be explicit and enter some isolate either by sending a task to some isolate or by using `Isolate.runSync` to enter isolate synchronously.

aam: > will the mutation of numRedirects in the callback we seen by other threads

There is one instance of `numRedirects` per invocation of `doGet`. So isolateGroupShared handler created in that `doGet` will get access to only that one `numRedirects` (via captured context).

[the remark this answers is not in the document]
mraleph: I think I would prefer to have a single shared isolate - it can also become sorta the representation of the _isolate group_, for which we currently don't have any representation.

working/333 - shared memory multithreading/proposal.md
mraleph: Documented that we expected atomic initialization where only one thread will run the initializer.

working/333 - shared memory multithreading/proposal.md
mraleph: > In my example, is `numRedirects` shared between threads i.e. will the mutation of `numRedirects` in the callback we seen by other threads?

Yes, it will be shared. If multiple threads invoke the same callback they will attempt to mutate the same variable. 

(Which also means that if you expect that you should use an atomic variable).

bernaferrari:
I like the proposal and I think it is really really needed, but my biggest frustration with Isolates is that it is incredibly low level and so hard to use that no one uses it (it is incredibly complex to cancel, for example. When do you want to cancel? Always. What if someone closes the screen? Or tap the screen again?). Maybe it would be a different proposal, but at this point all I wish is that Isolates were so easy to use anyone would use (just like async await), like they do with go, kotlin coroutines or swift now. I think this proposal makes sense, but somehow I still hoped for more :(

mkustermann:
> I like the proposal and I think it is really really needed, but my biggest frustration with Isolates is that it is incredibly low level and so hard to use that no one uses it (it is incredibly complex to cancel, for example. When do you want to cancel? Always. What if someone closes the screen? Or tap the screen again?). Maybe it would be a different proposal

Yes, I think being able to trigger and cancel asynchronous/concurrent threads of execution & cancel those is orthogonal to sharing memory among those concurrent threads of execution.

> but at this point all I wish is that Isolates were so easy to use anyone would use (just like async await)

Why do you say "so easy to use ... just like async await"?

Generally one **cannot cancel** an async operation triggered via async/await. Our `Future`s don't have a `cancel()` that would propagate - that may be the fundamental problem: Any asynchronous abstractions are built on top of `Future` and async/await (even `Isolate.run()`) - so if those abstractions don't provide cancellation ability, it doesn't help much if the lowest level supports them.

> like they do with go, kotlin coroutines or swift now. I think this proposal makes sense, but somehow I still hoped for more :(

In Go: It doesn't seem there's a generic cancel mechanism, instead you pass the go routine a channel and the go routine itself has to react to messages on that channel. One could use the same in Dart: Make an isolate listen on a port and if someone sends a message on that port, the isolate shuts itself down - arguably it's not so easy to use in Dart, but one could make a helper package that provides an abstraction for this.

In Kotlin: Coroutines can be cancelled. It's implemented by throwing catchable exceptions in specific places. It's cooperative - so one doesn't have a guarantee it will actually be cancelled / cleaned up.

Right now in Dart we have an `Isolate.kill()` API but that will trigger forced shutdowns - in effect via uncatchable exceptions. If we supported a way to throw catchable cancel exceptions, then there's the issue that an isolate can run multiple concurrent async functions (imagine, `await Future.wait([asyncFoo(), asyncBar()])`) then we'd need to decide where/how to propagate such cancel exceptions. For this to work properly one would probably want to remember the tree of triggered asynchronous computations and then cancel the entire tree (and if any of them e.g. spawned other isolate computations also cancel those, ...) transitively.

@bernaferrari Please consider filing new issues for feature requests.

mraleph:
FWIW the cancellation is actually mentioned in the proposal, in the section covering [structured concurrency](https://github.com/dart-lang/language/blob/main/working/333%20-%20shared%20memory%20multithreading/proposal.md#structured-concurrency). I think we have a pretty reasonable design for cancellable futures, but nobody seems to have time to actually implement it.

Issue #4015, opened 2024-08-06
Is it possible to augment the `values` of an enum?


eernstg:
Thanks to @sgrekhov for bringing up this topic. We may or may not allow developers to augment an implicitly induced member like the `values` list of an enum:

```dart
enum E { one }

const someAnnotation = 1;

augment enum E {
  augment one;

  @someAnnotation
  augment static const values;
}
```

We do already allow the implicitly induced getter of an instance variable to be augmented, which may be seen as prior art. However, there may also be unwanted consequences of allowing augmentation of `values`.

(Note that the example relies on having #4008 such that `augment static const values;` isn't a syntax error.)

As @jakemac53 mentioned below, we could make augmentations of `index`, `hashCode`, and operator `==` an error as well. This may be covered already, otherwise we could add a couple of words in the augmentations feature specification in order to make it explicit.

`name` might have some restrictions as well, but this is an extension method and there is nothing stopping us from declaring a member named `name` in an `enum`, so we probably can't do anything about that.

@dart-lang/language-team, WDYT?

jakemac53:
We probably should not allow augmenting `values` for enums.

jakemac53:
We currently have this specified in the enhanced enums spec which I think probably covers this?

"It's a **compile-time error** if the enum declaration contains a static or instance member declaration with the name `values`, or if the superclass or any superinterface of the enum declaration has an interface member named `values`. _A `values` static constant member will be provided for the class, this restriction ensures that there is no conflict with that declaration._"

An augmenting declaration is still a declaration, so I think it is covered.

lrhn:
We _could_  allow  augmenting the provided `values` declaration, like any other static constant variable declaration. 
(Do we allow augmenting constant variables? I assume we do not allow augmenting their getter, but we should make sure to say that. So it can only be the initializer. Not impossible. Maybe confusing.)
Let's say we do not allow augmenting `values`, then we can say that an augmenting `enum` declaration that adds enum values (which I think we do allow) implicitly "augments" the `values` declaration as if but
```dart
  augment static const List<ThisEnum> values =
      [...augmented, newValue1, newValue2];
```

If we did allow augmenting, we'd need to be careful about the specification of that `values`, so that the initial initializer existing contains all values declared by all later augmentations. Probably not impossible, but slightly harder to specify than an iterative process.

eernstg:
[...]

PR #4103, opened 2024-09-26
Make adjustments following issue #4015

eernstg:
In issue #4015, it is argued that several system provided members of an enumerated declaration should not be augmentable, namely: `index`, `hashCode`, operator `==`, and `values`. This PR makes it a compile-time error to augment them.

It was also suggested that it should be an error to declare or augment a member of an enum named `name`, because this identifier should denote the system provided extension getter `name`. This PR does not introduce any rules about `name`, because it would be a breaking change to prevent `enum` declarations from declaring a member whose name is `name` (instance or static).

Resolves https://github.com/dart-lang/language/pull/4015.

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -726,7 +726,18 @@ It is a **compile-time error** if:
 
 *   An `external` variable is augmented with an `abstract` variable.
 
-### Augmenting enum values
+### Augmenting enum members
+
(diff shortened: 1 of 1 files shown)


eernstg:
Hi, @sgrekhov! If and when this PR is landed, we'd need a couple of tests.

sgrekhov:
Sure!

jakemac53 (approved):

PR #4106, opened 2024-09-26
Enable various variables and factories with no implementation

eernstg:
With support for augmentation, variables can be declared with no initializing expression in a couple of new cases where this was previously a syntax error, namely constant and final top-level variables as well as constant and final static variables. This PR changes the grammar to allow this kind of declaration. The point is that the omitted initializing expression can be provided by augmentation. Similarly for factory constructor declarations, they do not have to have a body when the body can be provided by augmentation. Finally, static getters, setters, and methods are also generalized to allow omitting the body, that is, they are allowed to look like "abstract static members", again because the body can be provided by augmentation.

The set of elements that are required for each definition are unchanged, but parsing will now succeed, and it is a non-syntax compile-time error if an element is omitted, e.g., if a constant top-level variable has no initializing expression.

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -1126,7 +1126,7 @@ topLevelDeclaration ::= classDeclaration
   | 'augment'? functionSignature functionBody
   | 'augment'? getterSignature functionBody
   | 'augment'? setterSignature functionBody
-  | 'augment'? ('final' | 'const') type? staticFinalDeclarationList ';'
+  | 'augment'? ('final' | 'const') type? initializedIdentifierList ';'
   | 'augment'? 'late' 'final' type? initializedIdentifierList ';'
   | 'augment'? 'late'? varOrType initializedIdentifierList ';'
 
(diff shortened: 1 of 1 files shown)


jakemac53 (approved):

eernstg:
Hello @sgrekhov! These changes were expected already a while ago, and I don't think there's a need to change any co19 tests as a consequence of these changes because those tests are already relying on the updated grammar.

However, the spec parser update in https://dart-review.googlesource.com/c/sdk/+/387160 seems to reveal a couple of syntax errors in augmentation related tests (for example, augmenting_declarations_A05_t01.dart has some missing semicolons; `enum E` seems to need a value like `e1;` in augmenting_declarations_A05_t03_lib.dart, and a few others).

So it would probably be a good idea to wait for the spec_parser update and then take a look at the augmentation tests. Of course, it would be great if I could get a heads-up if the updated grammar somehow fails to handle the actual syntax in those tests, such that we can clarify whether it's the test or the grammar that needs to be fixed.

Issue #4113, opened 2024-09-30
Allow an extension instance member to be resolved implicitly even when it clashes with a static member of the on-type


eernstg:
See https://github.com/dart-lang/sdk/issues/56818 for a more detailed analysis.

I'd recommend that we adjust the language specification such that an extension is applicable even in the case where the on-type has a static member with the same member name.

```dart
class A {
  static void foo() {}
  void bar1() => foo(); // Calls the static member `A.foo`.
}

extension E on A {
  void foo() {}
  void bar2() => foo(); // Calls the extension instance member `E.foo`.
}

void main() {
  A.foo(); // OK, calls the static method.
  A().foo(); // OK, calls the extension instance method.
}
```

This can be done by adding one word to the [relevant rule](https://github.com/dart-lang/language/blob/60bbf6e1d22e0ddafe47f43592e0bc3c8e5f44fd/specification/dartLangSpec.tex#L6664):

```tex
The type $S$ does not have an instance member with the basename $m$,
and $S$ is neither \DYNAMIC{} nor \code{Never}.
```

With the current rule, `A().foo()` is an error.

The feature specification which was used as the vehicle to introduce this language mechanism uses the [same wording](https://github.com/dart-lang/language/blob/main/accepted/2.7/static-extension-methods/feature-specification.md#applicability), which means that the rule has been in effect all the time where this feature has been part of the language.

However, it still seems likely that the intended behavior is to accept `A().foo()` in the above example, and the analysis in https://github.com/dart-lang/sdk/issues/56818 doesn't reveal any difficulties in allowing it.

The change has no associated implementation effort because the analyzer and the CFE already have the proposed behavior. This also makes the change non-breaking. It's basically just fixing a mistake in the specification.

@dart-lang/language-team, WDYT, do you see any dangers or reasons not to perform this adjustment?

lrhn:
Ship it. 
It's a spec bug, so we should just fix it.
It's in the original feature specification too, so it's not new, and yet implementations did the right thing, so it's at least not *very* confusing.

The intent has always been that if an instance member access `e.foo...` is invalid because the type interface of the static type of `e` does not have a `foo` member, then an extension member may be applicable.
Static members are never relevant, so it's a mistake that the spec said "does not have a member" where it should have said "does not have an instance member".

[...]

PR #4116, opened 2024-10-01
Clarify extension member resolution

eernstg:
See https://github.com/dart-lang/language/issues/4113 for some background information.

This PR changes the rule about extension applicability such that it is clarified that static members are irrelevant, it is only required that the static type of the receiver does not have an _instance_ member with the same basename as the selector which is potentially an extension member invocation.

There is no implementation effort associated with this specification adjustment, and it is not a breaking change, because the implementations already behavior in the way which is specified with this change.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -41,6 +41,9 @@
 % version of the language which will actually be specified by the next stable
 % release of this document.
 %
+% Sep 2024
+% - Clarify extension member invocation rule.
+%
 % Jun 2024
 % - Add missing references to section 'Type dynamic' at the points where the
 %   static analysis of Object member invocations is specified.
@@ -6661,7 +6664,7 @@ to $e$ if the following conditions are all satisfied:
     because they do not have a type.%
   }
 \item
-  The type $S$ does not have a member with the basename $m$,
+  The type $S$ does not have an instance member with the basename $m$,
   and $S$ is neither \DYNAMIC{} nor \code{Never}.
 
   \commentary{%


specification/dartLangSpec.tex
```diff
@@ -6661,7 +6666,7 @@ \subsubsection{Applicability of an Extension}
     because they do not have a type.%
   }
 \item
-  The type $S$ does not have a member with the basename $m$,
+  The type $S$ does not have an instance member with the basename $m$,
```
lrhn, line 6669: Do we have a name for "the set of instance members of a type"? I sometimes use "type interface" or "type member signatures". 
It's the signatures of the members that a typed member access/invocation can access.
(Because that's the kind of members this text is talking about. Not declarations, not concrete "instance members", just the things that can be accessed on something with that static type.)

lrhn (approved):

specification/dartLangSpec.tex
eernstg: We do use "members of" a library, class, mixin, or enumerated type, "instance members of" a class or extension, "static members of" a class or extension, "a member of" the type `void`, a super-bounded type, the on-type, "members for" a type, "a member of" an object.

Another branch of this terminology may be better in general: We're quite frequently talking about "the interface of" an interface type, which is considered to be a nominally identified set of member signatures.

So we do certainly talk about the members of a class or class-like entity, explicitly mentioning both of the static and instance kinds. We do also talk about the members of a type (not nearly as many times), again both kinds.

But I think it's fair to say that the elements in the interface of an interface type would be instance members, implicitly or explicitly.

I think we're OK using 'static members of a class' and 'instance members of a class', no problem, and we're pretty much OK using 'static members of an interface type' and 'instance members of an interface type', even though those occurrences should perhaps be modified to talk about the underlying class/mixin/whatever that was used to induce said type.

And I think it's fair to say that the interface of an interface type contains only instance members. We could still include the word 'instance' to double down on the disambiguation.

lrhn: Indeed. The issue here is that the `on` type of an extension may not be an interface type, and yet I want to talk about the interface of it, meaning the set of member signatures that it has. (Which includes the "members of `Object`", for function types an instance method names `call`, for record types instance getters for each field, and for extension types, the extension members of the type. 

If we could consistently say "the type signature of at type" to mean that set of _member signatures_, then that could make some things much easier to talk about. (And the type interface/signature of an interface type is the interface type's interface members, unsurprisingly.)

(I've considered "type interface" and "type signature", but I like "signature" better because it matches "member signature", which is itself a useful abstraction over something with a name *and* a way to use it - getter, setter, or method so far.)

eernstg: > If we could consistently say "the type signature of at type" to mean that set of _member signatures_

I've used "the interface of a type" with that intended meaning many times. The specification does say "interface type" in a number of situations, but I don't see a problem with a slightly more forgiving use of this phrase that includes some non-interface types with the `Object` members as the interface. Record types is a good example, they certainly do have getters named in ways like `$1` and `$foo`, and I don't see why we should special case them in the situation where we just want to talk about the available static information about instance members.

PR #4123, opened 2024-10-07
Update augmentation specification.

lrhn:
*   Says that the entire class scope is in the lexical scope for members.
*   Update `augemented` specification, include constructors.
    *   Only reserved inside expressions and bodies of augmenting
        member-like declarations, not type-like declarations.
    *   Try to give grammar. It's really a parameterized grammar
        per kind of declaration. We don't have good parameterized grammars,
        so it's said in prose.
*   Changes specification for non-redirecting generative constructors.
    *   Adds order of initialization of instance variables with initializers.
    *   Says that body cannot use `augmented`.
    *   Therefore no longer needs to reuse variable bindings, so don't.
    *   Each augmenting declaration binds actuals to formals,
        remember any super parameters or super initializers,
        executes initializer list entries,
        and then executes its augmented declaration.
        When that comes back, it executes its body.
    *   (Layering an augmentation on top of an NRG-constructor
        auto-calls augmented at the end of its initializer list.)

    The initializing formals and super parameters need more work.
*   Some small tweaks to disallow invalid combinations.

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -182,13 +182,15 @@ declaration which is smaller than _D_, according to the 'after'
 ordering)*. A compile-time error occurs if no declaration satisfies the
 requirements on _D1_.
 
-We say that _D1_ is the declaration which is _augmented_ by _D_.
+We say that _D1_ is the declaration which is _augmented_ by _D_,
+or just, in the context of _D_, *the augmented declaration*.
 
 *Note that _D1_ can be an augmentation declaration or an introductory
 declaration.*
 
 An augmentation declaration does not introduce a new name into the surrounding
-scope. *We could say that it attaches itself to the existing name.*
+scope. *It attaches itself to the existing name in the scope, which must exist
+for the augmentation to be valid.*
 
 Making `augment` a built-in identifier is language versioned, to make it
 non-breaking for pre-feature code.
@@ -212,8 +214,7 @@ including the ones that are induced by variables.
 *This means that augmentation of a variable declaration can be a getter
 declaration that augments the induced getter, a setter declaration that
 augments the induced setter, or a variable declaration that augments the
-initializer. The augmenting declarations can themselves be an implicitly
-induced getter or setter, or an explicitly declared one.*
+initializer expression.*
 
 *Note that an abstract variable declaration and an external variable
 declaration correspond to a getter declaration and possibly a setter
@@ -222,8 +223,8 @@ also no support for augmenting it.*
 
 These operations cannot be expressed today using only imports, exports, or
 part files. Any Dart file (library file or part file) can contain
-augmentation declarations. *In particular, an augmentation can augment a
-declaration in the same file in which it occurs.*
+augmenting declarations. *In particular, an augmentation can augment a
+declaration in the file in which it occurs.*
 
 A type augmentation can add new members to an existing type, or augment a
 member declaration in the same context *(that is, in the same type
@@ -237,16 +238,16 @@ which may be used to provide a body for an abstract member._
 
 An augmentation that replaces the body of a function may also want to
 preserve and run the code of the augmented declaration (hence the name
-"augmentation").  It may want to run its own code before the augmented
-code, after it, or both.  To support that, we allow a new expression syntax
+"augmentation"). It may want to run its own code before the augmented
+code, after it, or both. To support that, we allow a new expression syntax
 inside the "bodies" of augmenting declarations (function bodies,
 constructor bodies, and variable initializers). Inside an expression in an
 augmenting member declaration, the identifier `augmented` can be used to
 refer to the augmented function, getter, or setter body, or variable
-initializer. This is a contextual reserved word within `augment`
-declarations, and has no special meaning outside of that context. See the
-next section for a full specification of what `augmented` means, and how it
-must be used, in the various contexts.
+initializer. This is a contextual reserved word within `augment` function
+or variable declarations, and has no special meaning outside of that context.
+See the next section for a full specification of what `augmented` means,
+and how it must be used, in the various contexts.
 
 *Note that within an augmenting member declaration, a reference to a member
 by the same name refers to the final version of the member (and not the one
@@ -307,7 +308,7 @@ can `// ignore:` it if they know what they’re doing.
 
 ### Type inheritance
 
-An augmenting declaration may have no type annotations for a return type,
+An augmenting declaration may omit type annotations for a return type,
 variable type, parameter type, or type parameter bound. In the last case,
 that includes omitting the `extends` keyword. For a variable or parameter,
 a `var` keyword may replace the type.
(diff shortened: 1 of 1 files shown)


lrhn:
Needs more work. No updated version/changelog, pretty sure the rules for using initializing formals and super parameters are inconsistent, aka. not updated everywhere, and super-parameters need more formal specification.

Still: WDYT? How far is it from what we want?

working/augmentation-libraries/feature-specification.md
```diff
@@ -321,82 +322,150 @@ doesn't, it must repeat the type from the augmented definition__.
...
-    refers to the augmented function. Tear offs are not allowed, so this
-    function must immediately be invoked.
-
-*   **Augmenting operators**: When augmenting an operator, `augmented`
-    refers to the augmented operator method, which must be immediately
-    invoked using function call syntax. For example, when augmenting
+    in the function body refers to the augmented function, and it must
+    occur only as the function of a function invocation, which means
+    it is immediately followed by one, or no, type argument list
+    and then an argument list.
```
munificent, line 373: This is a little difficult to read. Maybe:

"which means it is immediately followed by a type argument list (which may be omitted) then an argument list (which must be present but may be empty)."

working/augmentation-libraries/feature-specification.md
```diff
@@ -321,82 +322,150 @@ doesn't, it must repeat the type from the augmented definition__.
...
+    _Tear offs are not possible, neither are member accesses like
+    `augmented.call()`._
+    The expression `augmented` must not occur in an optional parameter
+    default value expressions _It is a reserved word in such expressions,
```
munificent, line 377: "default value expressions" -> "default value expression."

working/augmentation-libraries/feature-specification.md
```diff
@@ -321,82 +322,150 @@ doesn't, it must repeat the type from the augmented definition__.
...
+    but is currently cannot be used._
```
munificent, line 378: "is currently" -> "currently".

working/augmentation-libraries/feature-specification.md
```diff
@@ -321,82 +322,150 @@ doesn't, it must repeat the type from the augmented definition__.
...
+    returned value.
+    Type inference of <Code>augmented\<_typeArgs_\>(_args_)<code>
+    is performed as a function invocation with the function signature
+    of the augmented declaration.
+
+*   **Augmenting operators**: Operators are methods with special
+    invocation syntax. When augmenting an operator declaration, `augmented`
+    refers to the augmented operator *method*, and it must occur only
+    as the function of a function invocation. For example, when augmenting
     `operator +` you could use `augmented(1)` to call the augmented
```
munificent, line 393: "could" -> "would"

working/augmentation-libraries/feature-specification.md
```diff
@@ -321,82 +322,150 @@ doesn't, it must repeat the type from the augmented definition__.
...
+    operator, when augmenting `operator []=` you would use the
+    `augmented(key, value)` syntax, and when augmenting `operator -()`,
+    you would use `augmented()`.
+    Parsing, evaluation and type inference is handled like a
+    function augmentation.
 
 *   **Augmenting enum values**: When augmenting an enum value, `augmented`
-    has no meaning and is not allowed.
+    has no meaning and is not allowed to occur in constructor argument
+    expressions.
```
munificent, line 402: "has no meaning" could be inferred to mean "is not reserved". Maybe say:

"`augmented` is treated as a reserved word that can't be used for anything."

working/augmentation-libraries/feature-specification.md
```diff
@@ -321,82 +322,150 @@ doesn't, it must repeat the type from the augmented definition__.
...
-    `augmented` can only be used in an initializer expression, and refers
-    to the augmented variable's initializing expression, which is
-    immediately evaluated.
-
-    It is a compile-time error to use `augmented` in an augmenting
-    variable's initializer if the member being augmented is not a variable
-    declaration with an initializing expression.
+The identifier `augmented` is a reserved word, with associated special syntax,
+when it occurs inside an expression or body of an augmenting variable, getter,
+setter, function, operator or constructor declaration.
```
munificent, line 327: Just to be clear, the "or body" means that `augmented` is *not* a reserved word inside the *signature* of a declaration, right? So you could have, for example:

```dart
void foo(int x) {}

augment void foo(int augmented) {}
```

Do we want to allow parameters or type annotations to use `augmented` in augmenting declarations? I can't see any good use for it.

working/augmentation-libraries/feature-specification.md
```diff
@@ -321,82 +322,150 @@ doesn't, it must repeat the type from the augmented definition__.
...
+    *   **Non-redirecting generative constructor**:
+        The term `augmented` has no meaning, and must not occur in
+        optional parameter default value expressions, initializer list
+        entry expressions (whether initializing or asserts),
+        super-constructor invocation argument expressions, or in the
+        body. _Non-redirecting generative constructors automatically
+        invoke their augmented constructor after executing their
+        initializer list, and will execute their body after that of
+        their augmented constructor._
+    *   **Non-redirecting Factory constructor**: An `augmented` in the
```
munificent, line 414: Nit: "Factory" doesn't need to be capitalized.

working/augmentation-libraries/feature-specification.md
```diff
@@ -321,82 +322,150 @@ doesn't, it must repeat the type from the augmented definition__.
...
+        constructor body refers to the augmented non-redirecting factory
+        constructor. It must occur only as the function of a (non-generic)
+        function invocation.
+        The term <code>'augmented' \<argumentList\></code> clause parses
+        as a `<primary>` expression.
+        When evaluated, it invokes the augmented constructor on the same
+        class and evaluates to the returned value.
```
munificent, line 421: Maybe add "with the given arguments" just to make it clear that there is no implicit argument forwarding (since generative constructors *do* have the arguments implicitly forwarded).

working/augmentation-libraries/feature-specification.md
```diff
@@ -321,82 +322,150 @@ doesn't, it must repeat the type from the augmented definition__.
...
+        The term `augmented` must not occur in optional parameter
+        default value expressions.
+    *   **Redirecting generative constructor**:
+        The term `augmented` has no meaning, and must not occur in
+        optional parameter default value expressions or in
+        target constructor arguments expressions.
+    *   **Redirecting factory constructor**: Contains no expressions.
+
+    The term <code>'augmented' \<argumentPart\></code> words parses
+    as a `<primary>` expression.
```
munificent, line 433: Is this leftover text? If so, delete. If not... what's it about? At the very least, I think you can delete "words".

working/augmentation-libraries/feature-specification.md
```diff
@@ -321,82 +322,150 @@ doesn't, it must repeat the type from the augmented definition__.
...
-identifier `augmented` occurs in _Dm_.
-
-*For example, inside `augment class C` we could have a declaration like
-`void f() {...augmented()...}`. This is an error because the outer
-`augment` forces the meaning of `augmented` to be about augmentation in the
-entire scope, but the method declaration is introductory and hence there is
-no earlier declaration to augment.*
+invoked with the same value for `this`. For a non-redirecting factory
+constructor, it's invoked with the same type parameter bindings for the
+surrounding type.
```
munificent, line 438: It took me a while to realize why the type parameters are implicitly forwarded. I think it's because in the `augmented(...)` call, there isn't an obvious place to put them. Putting them after `augmented` would make it look like they are type parameters on the constructor itself, not the class. That would be a problem if/when we add support for generic constructors.

It does feel like there is a small expressiveness hole here. I can imagine someone wanting to augment a factory constructor and have it create an object with a slightly different (probably tighter) set of type arguments. It's probably fine to not support this.

Maybe add some non-normative text explaining the motivation here? (And if I didn't guess the motivation right, I would definitely like to read it. :) )

working/augmentation-libraries/feature-specification.md
```diff
@@ -321,82 +322,150 @@ doesn't, it must repeat the type from the augmented definition__.
...
+The grammar change for `augmented` adds the following productions:
+```ebnf
+  <augmentedValueExpression> ::= 'augmented'
+  <augmentedSetterExpression> ::= 'augmented' '=' <expression>
+  <augmentedFunctionExpression> ::= 'augmented' <argumentPart>
+  <augmentedConstructorExpression> ::= 'augmented' <argumentList>
+
+  <primary> ::= ...
+     | <augmentedValueExpression>        -- augmenting variable or getter body
+     | <augmentedSetterExpression>       -- augmenting setter body
```
munificent, line 449: Is slotting this into `<primary>` correct? Wouldn't that allow:

```dart
1 + augmented = 2;
```

working/augmentation-libraries/feature-specification.md
```diff
@@ -321,82 +322,150 @@ doesn't, it must repeat the type from the augmented definition__.
...
+augmenting declaration in the list above. The grammar production is only
+allowed inside the scopes mentioned. _This makes occurrences in parameter
+default values and initializer list expressions invalid._
+
+_The word is also reserved, and has its designated meaning, inside the bodies of
+nested functions inside the expressions or bodies of the augmenting declarations.
+It's possible to close over `augmented` inside an augmenting function, to,
+for example, do `list.forEach((x) => augmented(x));` in a function augmentation,
+but doing `List.forEach(augmented)` is not allowed. That would treat the
+augmented function as if it had an identity._
```
munificent, line 463: This is a good thing to point out. It feels weird at first, but then I realized that `super` has the same limitation and it's never bothered me.

working/augmentation-libraries/feature-specification.md
```diff
@@ -910,7 +959,66 @@ A non-redirecting generative constructor marked `augment` may:
         initializer list containing first the assertions and field initializers
         of the augmented constructor, if any, then the assertions and field
         initializers of the augmenting constructor, and finally any
-        super-initializer of either the augmeted or augmenting constructor.
+        super-initializer of either the augmented or augmenting constructor.
+
+Invoking a non-redirecting generative constructor, with an argument list and a
+new object to initialize, two effects.
```
munificent, line 965: "two effects." -> "has two effects:".

working/augmentation-libraries/feature-specification.md
```diff
@@ -910,7 +959,66 @@ A non-redirecting generative constructor marked `augment` may:
...
+        otherwise use the remembered super initializer.
+        Append all remembered super arguments to the argument list
+        of the super initializer, positional ones in the source order of the
+        super parameters that introduced them.
+        Then *invoke that super initializer*.
+    *   In either case, when that execution or invocation completes, if this
+        declaration has a body, execute that body of this declaration in a
+        scope that has the parameter scope of the binding-actuals-to-formals
+        step as parent scope, and with `this` bound to the object being
+        initialized.
```
munificent, line 1019: This whole section is really good.

working/augmentation-libraries/feature-specification.md
```diff
@@ -1082,43 +1192,62 @@ some_lib.dart       :
...
+| isEven() {...}                       |
+|                        _isOdd() {...}|
+'--------------------------------------'
+`---------------'   :  `---------------'
+         ^          :         ^
+         |          :         |
+.-----------------. : .-----------------.
+| C.isEven() body | : | C._isOdd() body |
+'-----------------' : '-----------------'
 ```
```
munificent, line 1220: My ASCII art diagram lives on! :)

working/augmentation-libraries/feature-specification.md
```diff
@@ -1082,43 +1192,62 @@ some_lib.dart       :
...
+_(The individual member signatures may refer to names that are not
+available in all files.)_
+
+The individual class member declarations have the class scope of the
+class *in the current file* as their parent scope in the lexical scope
+chain. 
+
+_It still is a compile time error for both a static and instance member
+of the same name to be defined on the same type, even if they live in
+declarations. You cannot work around this restriction by moving the
```
munificent, line 1242: "live in declarations" -> "live in different declarations for the same class".

munificent (approved):
These changes are all great. Thank you for warming up my half-baked proposal into something edible. :)

working/augmentation-libraries/feature-specification.md
```diff
@@ -321,82 +322,150 @@ doesn't, it must repeat the type from the augmented definition__.
...
+    The static type of `augmented` is the static type of the initializer
+    expression of the augmented declaration.
+
+*   **Augmenting getters**: Within an augmenting getter's body, the `augmented`
+    expression invokes the augmented getter and evaluates to its return value.
+    When augmenting a variable with a getter, this will invoke the implicitly
+    induced getter of the augmented variable declaration.
+    _This applies even to `late` variable declarations, where the implicit
+    getter does more than just access the underlying storage._
+    The reserved words parses as a `<primary>` expression.
```
jakemac53, line 347:
```suggestion
    The reserved word parses as a `<primary>` expression.
```

working/augmentation-libraries/feature-specification.md
```diff
@@ -321,82 +322,150 @@ doesn't, it must repeat the type from the augmented definition__.
...
+    Evaluating `augmented` invokes the augmented declaration's getter
+    function, and evaluates to the returned value.
+    The static type of `augmented` is the declared return type of the augmented
```
jakemac53, line 350: Should we specify this as the declared return type or something a bit more general, like how we use the "static type" for initializer expressions?

For instance, does "declared return type" cover the override case correctly?

working/augmentation-libraries/feature-specification.md
jakemac53, line 378:
```suggestion
    but it currently cannot be used._
```

working/augmentation-libraries/feature-specification.md
```diff
@@ -321,82 +322,150 @@ doesn't, it must repeat the type from the augmented definition__.
...
-*   **Augmenting operators**: When augmenting an operator, `augmented`
-    refers to the augmented operator method, which must be immediately
-    invoked using function call syntax. For example, when augmenting
+    in the function body refers to the augmented function, and it must
+    occur only as the function of a function invocation, which means
+    it is immediately followed by one, or no, type argument list
+    and then an argument list.
+    _Tear offs are not possible, neither are member accesses like
+    `augmented.call()`._
+    The expression `augmented` must not occur in an optional parameter
```
jakemac53, line 376:
```suggestion
    The expression `augmented` must not occur in optional parameter
```

working/augmentation-libraries/feature-specification.md
```diff
@@ -321,82 +322,150 @@ doesn't, it must repeat the type from the augmented definition__.
...
+    you would use `augmented()`.
+    Parsing, evaluation and type inference is handled like a
+    function augmentation.
 
 *   **Augmenting enum values**: When augmenting an enum value, `augmented`
-    has no meaning and is not allowed.
+    has no meaning and is not allowed to occur in constructor argument
+    expressions.
+
+*   **Augmenting constructors**:
```
jakemac53, line 404: I only got to here so far, will continue reviewing tomorrow (and I know this section will require more brain power than I currently have available 🤣 )

jakemac53 (commented):
Looking good so far, will continue my review tomorrow.

working/augmentation-libraries/feature-specification.md
lrhn: Yes, that is the goal. The `augmented` word is only reserved *inside the body*, and in any expressions outside of the body (default value expression, initializer list/redirection expressions for generative constructors), but not in the signature of the declaration.

I don't necessarily see a use for `augmented` in the parameter list, but I also don't see a use for reserving it outside of where it can be used.
The words `await` and `yield` are only reserved inside the asynchronous body, they're not reserved in the parameter list. 
They are also reseved in the signatures of nested functions, but not those function's body. That might be a mistake, but it avoids you declaring a `void await(Object? _){}` function inside an `async` function.
It also avoids you declaring a `void foo({bool await = false}) {...}` function, even if you need one, you need to do `var foo = (() => ({bool await = false}) {...})();` instead, which you can.
(I'd like to allow nested function declarations to not treat the `augmented`, `yield` and `await` words as reserved in their signatures, even if they can't use it as a name, so that it's possible to declare a function with `augmented`/`await` as a named paramete name. Since you can declare them elsewhere, you might *need* to create a function like that, and the only workaround is to have a function declared outside of the reserved word scope that renames the parameter for you. But that's a different and bigger issue, and for `augmented` it could mean that you can *shadow* the `augmented` of the outer function with a parameter. Introducing `augmented` as a reserved word in entire expressions and bodies is probably the simplest, and therefore currently best, choice.)

I am reserveing `augmented` in initializer list and default value expressions because it would be possible to allow augmenting those and referring to the original expression. Again it's maybe not useful, but I want to keep the door open, and "reserved in any runtime code" (things that get executed or evaluated) is easier to remember than reserving it in initialier list initializer expressions, but not assert expressions.

So "body and expressions" covers that and is fairly straight line for users to remember.

(We _could_ choose to allow conditionally reserved words as parameter names, even in scopes where they are reserved, so you could call a function `foo(await: true, yield: 42)` even inside an `async` function. Any name you *can* use as a parameter name anywhere can also be used as a named argument name anywhere. Or even as a member name, `object.await(true).yield(42)` inside an async function, the prior `.` or following `:` making it be treated as a member-name, not a reserved word. Globally reserved words cannot be used that way. Or maybe they can. If we allow *escaping* reserved words into identifiers, like `this.\for(\await: true, \yield: 42, \if: isTuesday)`, so we  can use it for interop with languages that have different reserved words, then allowing *unescaped* named parameters and member access with reserved words makes the interop experience nicer. Shucks to be a syntax highlighter, but if it can latch onto the `.word` and `word:` patterns, it just might work!)

working/augmentation-libraries/feature-specification.md
```diff
@@ -321,82 +322,150 @@ doesn't, it must repeat the type from the augmented definition__.
...
-    variable's initializer if the member being augmented is not a variable
-    declaration with an initializing expression.
+The identifier `augmented` is a reserved word, with associated special syntax,
+when it occurs inside an expression or body of an augmenting variable, getter,
+setter, function, operator or constructor declaration.
+
+The exact grammar and meaning of the `augmented` keyword depends on what is
+being augmented.
+
+*   **Augmenting variables**: Within an augmenting variable declaration's
```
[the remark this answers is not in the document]
lrhn: It is repetetive, and unnecessarily so.
I think one of the parts is a leftover that should have been removed.

working/augmentation-libraries/feature-specification.md
lrhn: Good point. We want the return type, not the declared return type. Or maybe we want the inferred return type. If there is a difference, it's the type that the declaration is determined to have after (top-level?) type inference has been performed.

We just need to define what the return type of a getter declaration is.

The getter return type of a non-augmenting getter declaration is the type denoted by its declared return type, if any.
If there is no declared return type, the getter is an instance getter of a `class`, `mixin` or `enum` declaration, and any superinterface of the surrounding type has a getter with the same name, then the return type is the return type of the getter with that name in the combined superinterface of the current declaration. _It's a compile-time error if the combined superinterface does not have a combined type signature for that getter._
Otherwise the return type of the getter is `dynamic`. 
(We don't infer the type of bodies in top-level inference, right? It's only for variables.)

The getter return type of a variable declaration is the type denoted by the declared type of the variable, if any. Otherwise, if the variable has no initializer and it's an instance variable, it's the return type of the getter or parameter type of the setter in the combined super-interface, if any. Otherwise it's the static type of its initializer expression.

The getter return type of an augmenting getter declaration is the getter return type of the augmented declaration. (Can't change the type. If the augmenting declaration writes a type, it must be that type.)

It gets complicated, because it's basically repeating all of type inference. I hope there is a way to reuse type inference instead.

working/augmentation-libraries/feature-specification.md
lrhn: I went through a lot of phrasings here, including "zero or one type parameter lists", but the plural annoyed me.
Maybe just:
> it is immediately followed by an argument list, or a type parameter list and then an argument list.

working/augmentation-libraries/feature-specification.md
lrhn: Going with singular "expression" rather than removing "an".

working/augmentation-libraries/feature-specification.md
lrhn: it currently, actually.

working/augmentation-libraries/feature-specification.md
lrhn: I think it's deliberate, and it's really saying that `augmented(args)` is a `<primary>` inside an augmenting function, not a `<primary><selector>` parsed as a `<postfixExpression>`.

It's tricky. It's really a way of saying that we have a parameterized grammar that is unique to each kind of allowed `augmented` syntax. We do not allow parsing `augmented` as a `<primary>` inside an augmenting function or setter, only inside an augmenting variable or getter. At most one of the `augmented` structures is *enabled* at a time.
Our grammar doesn't really allow that (and I don't want five copies of all the rules from `<expression>` down to `<primary>`, so I'm trying to say it with words.
It's like a disambiguation rule: Inside this kind of augmenting declaration, use this production (only).

It's clearly not clear :)

working/augmentation-libraries/feature-specification.md
lrhn: Correct. Just as we are invoking instance members on the same `this`, we invoke the constructor on the same *class* (instantiated type).
That's definitely necessary for generative constructors, they are invariant in the type parameters of the object they initialize, if not by necessity then to avoid runtime checks in initialiers.

It also enforces that the `augmented` that is invoked is a part of *the same constructor*. It's not a different constructor you can call on a different type, it's a part of *this* constructor which is already being called on *this* type. It's a part that is *included* into this body by the invocation. (And inlining is a valid implementation approach.)

The same for instance members, the `augmented` has the same `this` because that is the `this` of this invocation, and the `augmented` is *part* of this invocation.
(We could allow `other.augmented(...)` if we really wanted to, as long as we know `other` extends this class - which is usually something we don't assume, so only if the classes are unimplementable, but then it looks more like overriding than augmenting.)

working/augmentation-libraries/feature-specification.md
lrhn: Good point, it would.
So it should be moved to where assignment expressions live, close to `expression`.

I'm used to thinking of `primary.selector.selector = expr` as a "selector chain", because of how `?.` extends through it, but it's not. It's an assignment expression that has a selector chain as LHS. (And that's also why `++foo?.bar` is allowed. Let's not do that again.)

working/augmentation-libraries/feature-specification.md
jakemac53: Right, but I think all we need to state here is that it uses the inferred return type? We don't need to explain how the inference works.

working/augmentation-libraries/feature-specification.md
```diff
@@ -321,82 +322,150 @@ doesn't, it must repeat the type from the augmented definition__.
...
+        optional parameter default value expressions, initializer list
+        entry expressions (whether initializing or asserts),
+        super-constructor invocation argument expressions, or in the
+        body. _Non-redirecting generative constructors automatically
+        invoke their augmented constructor after executing their
+        initializer list, and will execute their body after that of
+        their augmented constructor._
+    *   **Non-redirecting Factory constructor**: An `augmented` in the
+        constructor body refers to the augmented non-redirecting factory
+        constructor. It must occur only as the function of a (non-generic)
```
jakemac53, line 416: Interesting restriction regarding it being a non-generic invocation. That might be a good limitation, but I will note that this program works today.

```dart
void main() {
  print(A<num>.foo());
}

class A<T> {
  A();

  factory A.foo() {
    return A<int>() as A<T>;
  }
}
```

We might want to consider also blocking this for regular function augmentations if we are going to block it for factory constructors?

working/augmentation-libraries/feature-specification.md
jakemac53: I had a similar comment above... I hadn't considered that the type arguments are on the class and not the function... that does make sense. And in any case it is a very weird thing to return a class with a tighter generic type than was given, it probably shouldn't even be allowed.

working/augmentation-libraries/feature-specification.md
```diff
@@ -705,10 +774,10 @@ It is a **compile-time error** if:
...
         declaration only affects the initializer expression, not setters.
 
-*   A variable is augmented with another variable, and one is `late` and
-    the other is not. *(Augmentation cannot change `late`-ness, and since being
-    `late` does affect the initializer expression, the augmenting variable is
-    required to repeat the `late`.)*
+*   A variable is augmented by another variable declaration, and one is `late`
+    and the other is not. *(Augmentation cannot change `late`-ness, and since
+    being `late` does affect the initializer expression, the augmenting
+    variable is required to repeat the `late`.)*
```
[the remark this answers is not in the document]
jakemac53: I could see expanding this later on, but I do think that being `late` is an important property to be aware of, and hiding that behind an augmentation seems wrong. So, I would rather not do it right now, even if it could be technically correct to allow it in one use case.

working/augmentation-libraries/feature-specification.md
```diff
@@ -724,11 +793,11 @@ It is a **compile-time error** if:
 
 *   An `abstract` variable is augmented with a non-abstract variable.
 
-*   An `external` variable is augmented with an `abstract` variable.
+*   An `external` variable is augmented with a variable declaration.
```
jakemac53, line 796: I am not sure if we need this restriction? It is true that the `external` keyword would be unnecessary though in this case, but I feel like we have gone back and forth on this a bit so and I am curious what the motivation for not allowing this would be?

Do we need to make similar changes for external functions? The CFE today actually requires us to mark certain things external (but this is a bug).

working/augmentation-libraries/feature-specification.md
```diff
@@ -910,7 +959,66 @@ A non-redirecting generative constructor marked `augment` may:
...
         initializers of the augmenting constructor, and finally any
-        super-initializer of either the augmeted or augmenting constructor.
+        super-initializer of either the augmented or augmenting constructor.
+
+Invoking a non-redirecting generative constructor, with an argument list and a
+new object to initialize, two effects.
+
+*   First it initializes any instance variables in the class which have
+    initializing expressions:
+    *   For each instance variable declared by the class, in source order
```
jakemac53, line 969: Do we need to define "source order" given that fields might be defined in different parts?

working/augmentation-libraries/feature-specification.md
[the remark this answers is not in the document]
lrhn: ACK.

working/augmentation-libraries/feature-specification.md
lrhn: I really hope we get generic constructors at some point, so I want to reserve `augmented<int>()` to mean that, and `<int>augmented()` doesn't look great. Or parse at all.

working/augmentation-libraries/feature-specification.md
lrhn: I think it's defined somewhere as the "before" order from parts-with-imports. If not, it should be.

working/augmentation-libraries/feature-specification.md
jakemac53: > I think it was mainly that an external "variable declaration" doesn't introduce a variable to augment, it only introduces a getter and a setter. The only augmentations you should be allowed to make are to getter and setter.

That is a compelling argument for me. I think we can leave things as you have defined here. Possibly expand to allow adding metadata/comments.

> ```dart
> @newData
> augment external int foo; // Reasonable, but looks like the augmentation itself is external.
> ```

This seems ok to me, it just adds the annotation.

> ```dart
> @newData
> augment int foo; // Possible, looks like it's augmenting a variable.
> ```

I don't like this, because it does seem like it is actually making it non-external. And, we don't allow augmenting getters/setters with variables.

> ```dart
> @newData
> augment abstract int foo; // Already not allowed (`abstract` and `external` are usually mutually exclusive).
> ```

lgtm

> ```dart
> augment int foo = 1; // Completely unacceptable.
> ```

agreed (now)

> ```dart
> @newData
> augment foo; // Nice-to-have, but not currently possible.
> ```

I think this looks a bit weird so I wouldn't try to allow it

working/augmentation-libraries/feature-specification.md
jakemac53: SGTM

PR #4126, opened 2024-10-09
A little copy-editing on the enum shorthand proposal.

munificent:
I was reading the latest changes and noticed a duplicated sentence. And once I fixed that, I started noticing other things...

There are no semantic changes. This is just making it hopefully easier to read.

### DIFF
--- a/working/3616 - enum value shorthand/proposal-simple-lrhn.md
+++ b/working/3616 - enum value shorthand/proposal-simple-lrhn.md
@@ -7,11 +7,11 @@ are fairly simple and easy to explain.
 
 ### Elevator pitch
 
-An expression starting with `.` is an implicit static namespaces access on the
+An expression starting with `.` is an implicit static namespace access on the
 *apparent context type*.
 
-The type that the context expects is known, and the expression avoids repeating
-the type, and starts by doing a static access on that type.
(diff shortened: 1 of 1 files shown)


working/3616 - enum value shorthand/proposal-simple-lrhn.md
```diff
@@ -245,8 +250,8 @@ second operand.
 Examples of allowed comparisons:
 
 ```dart
-if (Endian.host == .big) ok!;
-if (Endian.host case == .big) ok!
+if (Endian.host == .big);       // ok!
+if (Endian.host case == .big);  // ok!
```
lrhn, line 254: This was deliberate. The `ok!` is a valid statement (or would be if it had the trailing `;` in the second line). It looks weird to have an `if` wih not body. Valid, but weird.
If the `ok!` doesn't look enough like a statement, then maybe change it to `isOK();`.

lrhn (approved):
Super edit, thanks! Much better!

working/3616 - enum value shorthand/proposal-simple-lrhn.md
munificent: Oh, haha, I didn't think about it being a null-assert of a variable named `ok`. :)

PR #4140, opened 2024-10-24
Add the initial draft for inference-using-bounds

chloestefantsova:
This PR adds the initial draft for the inference-using-bounds language feature specification.


accepted/future-releases/3009-inference-using-bounds/feature-specification.md
```diff
@@ -0,0 +1,319 @@
+# Inference using bounds
+
+Author: Chloe Stefantsova
+
+Status: Accepted
+
+Version 0.1 (see [CHANGELOG](#CHANGELOG) at end)
```
stereotype441, line 7: Personally I would start at 1.0. To me, "0.1" means it's a draft, and considering that we've already pretty much committed to doing this feature, it seems like we're well out of "draft" territory 😃

stereotype441 (approved):
lgtm, but since I'm relatively inexperienced in writing specs, I'm going to tag in @eernstg as another reviewer

kevmoo:
Cool to see the inline mermaid!

accepted/future-releases/3009-inference-using-bounds/feature-specification.md
```diff
@@ -0,0 +1,319 @@
...
+
+  boundsDownwards --> csstvDownwards
+  boundsUpwards --> csstvUpwards
+```
+
+## Proposed changes
+
+We propose generating more constraints from existing constraints and the bound. The original example is solved by this proposal and a few more beneficial examples are discussed afterwards.
+
+In the motivating example, adding the best-effort bound approximation didn't affect the result since the set of constraints for `X` already contained a lower-bound constraint for `X` (`C <# X`), and that constraint had the priority over the upper-bound "best-effort" approximation constraint, since type `C` is fully known (i.e. doesn't contain `_`). In the current proposal ***we aim to derive a lower-bound constraint by combining the already existing lower-bound constraint with the actual bound***.
```
eernstg, line 224: These sentences could be made a bit shorter, which would probably help the reader:
```suggestion
In the motivating example, adding the best-effort bound approximation didn't affect the result. The set of constraints for `X` already contained a lower-bound constraint for `X` (`C <# X`), and that constraint has priority over the upper-bound "best-effort" approximation constraint, since type `C` is fully known (i.e. doesn't contain `_`).

With this feature, we derive a lower-bound constraint by combining the already existing lower-bound constraint with the actual bound.
```

I split the paragraph in two because this allows the last sentence to be more prominent, without the unusual ***doubly embellished font*** usage.

accepted/future-releases/3009-inference-using-bounds/feature-specification.md
```diff
@@ -0,0 +1,319 @@
...
+
+In general, if we have two type constraints of the form `E <: Y` and `Y <: F`, where `Y` is a type variable and `E` and `F` are type schemas, it follows that `E <: F` holds if the `E <: Y` and `Y <: F` hold. So, what if instead of adding the constraint `X <: B'` in step 5 of CSSTV, where `B'` is the best-effort bound approximation, we generate a new set of constraints, running the subtype constraint generation algorithm described in [Subtype constraint generation][subtype-constraint-generation] for `P = C`, `Q = B`, `L = {..., X, ...}`.
+
+`C` doesn't contain type variables from `L` by construction, and `B` may contain type variables from `L` (confidently, `X` in case X is F-bounded, but also any type variables from the same declaration, regardless of F-boundness of `X`), which makes `C <# B` a valid input for [Subtype constraint generation][subtype-constraint-generation] and a good source of additional type information.
```
eernstg, line 228:
```suggestion
By construction, `C` doesn't contain type variables from `L`. `B` may contain type variables from `L` (surely, it contains `X` in case where `X` is F-bounded, but it may also contain other type variables from the same declaration, whether or not `X` is F-bounded), which makes `C <# B` a valid input for [Subtype constraint generation][subtype-constraint-generation] and a good source of additional type information.
```

accepted/future-releases/3009-inference-using-bounds/feature-specification.md
```diff
@@ -0,0 +1,319 @@
...
+
+The new constraints, generated by [Subtype constraint generation][subtype-constraint-generation] from `C <# B` are then added to the overall set of constraints for the current phase of type inference. Then, the solution for type variable `X` is computed from the new constraint set, at the end of the updated step 5 of CSSTV. 
+Note that when `X` isn't F-bounded and its bound doesn't refer to other type variables from the same declaration, the suggested changes for step 5 of CSSTV work exactly as the unmodified step 5 of the same algorithm.
+
+## Re-evaluating the motivating example
+Let's see how type inference, updated as proposed, works in the motivating example.
```
eernstg, line 234: I think an empty line would be expected here:
```suggestion

Let's see how type inference, updated as proposed, works in the motivating example.
```

accepted/future-releases/3009-inference-using-bounds/feature-specification.md
```diff
@@ -0,0 +1,319 @@
...
+  constraintGeneration --> newConstraint
+  boundsUpwards --> constraintGeneration
+
+  boundsDownwards --> csstvDownwards
+  boundsUpwards --> csstvUpwards
+```
+
+## More examples enabled by the current proposal
+
+In addition to the motivating examples, both the original and the more complex, being unbroken by the current proposal, some cases of non F-bounded type variables are enabled by it as well.
```
eernstg, line 293:
```suggestion
In addition to the motivating examples, both the original and the more complex, being handled by this feature, some cases of type variables that aren't F-bounded are handled as well.
```
Unbroken is a nice word, especially if it exists, but "being handled" is probably safer. ;-)

eernstg (commented):
Looks good!

I think the actual feature specification is the modification of 'inference.md', and the other document is all about background and motivation.

So I'm suggesting that `feature-specification.md` is renamed (perhaps to `design-document.md`).

With this rename in place, it's perfectly fine that the document doesn't specify the precise rules, and that it is completely example-driven.

I do suggest, though, that the formal update (that is, the new text in `inference.md`) is mentioned up front in the design document, such that (1) readers will still know what the diff is without needing to track down the git update where `inference.md` was updated, and (2) the design document is explicit about the precise thing that it's motivating and explaining. This would just be a short section.

accepted/future-releases/3009-inference-using-bounds/feature-specification.md
```diff
@@ -0,0 +1,319 @@
...
+
+Status: Accepted
+
+Version 0.1 (see [CHANGELOG](#CHANGELOG) at end)
+
+Experiment flag: inference-using-bounds
+
+##  Motivating example
+
+The motivating example is discovered and discussed in [#3009][] (Type inference does not solve some constraints involving F-bounds). Consider the following code:
```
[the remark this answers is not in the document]
chloestefantsova: Thanks for the suggestion, Erik! I wasn't sure which style to follow and defaulted to no line breaks. I've updated the .md document as you suggested.

accepted/future-releases/3009-inference-using-bounds/feature-specification.md
chloestefantsova: Thanks for the suggestion! I've updated the file accordingly.

accepted/future-releases/3009-inference-using-bounds/feature-specification.md
chloestefantsova: Agreed! Thanks for the suggestion. I've updated the file accordingly.

accepted/future-releases/3009-inference-using-bounds/feature-specification.md
```diff
@@ -0,0 +1,319 @@
...
+
+Note that adding the best-effort approximation of the bound and solving the constraint set once again didn't change the outcome. ***In general, lower bounds are preferred by the type inference, and the best-effort approximation is added as an upper-bound constraint.***
+
+It is notable that even though the example program contains all the type information needed to infer `B` as the type parameter, the front ends following the specification infers `C`, resulting in a compile-time error during the type checks of the type arguments against the bounds.
```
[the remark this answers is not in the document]
chloestefantsova: Thanks! Fixed.

accepted/future-releases/3009-inference-using-bounds/feature-specification.md
chloestefantsova: Thanks for the suggestion, Erik! I've updated the text accordingly.

accepted/future-releases/3009-inference-using-bounds/feature-specification.md
chloestefantsova: Thanks, Erik! I've updated the text as suggested.

eernstg (approved):
Let me just do LGTM, to make sure I am not blocking anything.

sgrekhov:
Github doesn't allow to comment that line, so I'm writing it here. File `inference.md`, informal section on lines L745-746.

> _The bound of a type variable is only included as a constraint when the
>  choice of type for that type variable is about to be frozen._

Seems that it should be updated as well.

accepted/future-releases/3009-inference-using-bounds/feature-specification.md
chloestefantsova: Thanks for the suggestion. I've updated the file accordingly.

accepted/future-releases/3009-inference-using-bounds/feature-specification.md
```diff
@@ -0,0 +1,319 @@
...
+  csstvDownwards -- "{X = _}" --> csstvUpwards
+  csstvUpwards -- "{X = C}" --> overallOutput
+
+  constraintGeneration --> newConstraint
+  boundsUpwards --> constraintGeneration
+
+  boundsDownwards --> csstvDownwards
+  boundsUpwards --> csstvUpwards
+```
+
```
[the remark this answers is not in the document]
chloestefantsova: Good idea, Erik! I've added a clarifying paragraph and some labels to the graph, to make the intention clear. Hopefully, it reads better now.

accepted/future-releases/3009-inference-using-bounds/feature-specification.md
```diff
@@ -0,0 +1,319 @@
+# Inference using bounds
```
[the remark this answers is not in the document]
chloestefantsova: Thanks for all the comments and for pointing that the update should also be made to the grounded solution for a set of type variables. I've updated the text accordingly.

chloestefantsova:
> Github doesn't allow to comment that line, so I'm writing it here. File `inference.md`, informal section on lines L745-746.
> 
> > _The bound of a type variable is only included as a constraint when the
> > choice of type for that type variable is about to be frozen._
> 
> Seems that it should be updated as well.

Thanks for spotting that! The informal section should be updated, tool.

accepted/future-releases/3009-inference-using-bounds/design-document.md
```diff
@@ -0,0 +1,620 @@
...
+
+In addition to the motivating examples, both the original and the more
+complex, being handled by the current proposal, some cases of non
+F-bounded type variables are enabled by it as well.
+
+The following example demonstrates a tempting assumption that doesn't
+hold today, but will hold with the proposed update. The developers
+might think that they will get the extracted type value for `Y` when
+they specify a type value for `X`. We remember similar cases occurring
+in practice.
```
eernstg, line 596: This might also be a little bit confusing... here's a somewhat reorganized way to say the same thing:
```suggestion
With a class like `A` below, and with this feature, the constructor
invocation in `test` is inferred as `A<List<num>, num>(<num>[])`.
Without this feature, it is inferred as `A<List<num>, dynamic>(<num>[])`.

Intuitively, this happens because the type inference algorithm without this
feature is unable to detect that there is a relationship between `X` and `Y`.
With this feature one could say that the value of `Y` is "extracted" from
the value of `X`. We have seen cases in practice demonstrating that the
ability to transfer this information during type inference is useful.
```

eernstg (approved):
LGTM!

accepted/future-releases/3009-inference-using-bounds/design-document.md
```diff
@@ -0,0 +1,620 @@
...
+
+fun main() {
+  f<B, B>(C1(), C2());
+  f(C1(), C2());
+  print("Done!");
+}
+```
+
+Let's take a look at the source of the described type inference
+behaviour in Dart.
```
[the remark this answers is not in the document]
chloestefantsova: Thanks for the suggestion! Done.

accepted/future-releases/3009-inference-using-bounds/design-document.md
```diff
@@ -0,0 +1,620 @@
...
+  constraintsUpwards -- "(1) Existing constraints\nfrom the\nupwards phase" --> constraintUpwards2
+  constraintsUpwards --> csstvUpwards
+  boundsUpwards --> csstvUpwards
+  boundsUpwards -- "(2b) Contribute A&lt;X&gt; to C <# A&lt;X&gt;\n" --> constraintGeneration
+  constraintsUpwardsUpdated -- "(3) Proceed with CSSTV using\nthe updated constraint set" --> csstvUpwards
+
+  csstvUpwards -- "{X = B}" --> overallOutput
+```
+
+## More examples enabled by the current proposal
```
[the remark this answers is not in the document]
chloestefantsova: Thanks! Done.

accepted/future-releases/3009-inference-using-bounds/design-document.md
```diff
@@ -0,0 +1,620 @@
...
+is limited in two ways:
+
+- it is invoked in-between the phases of type inference and doesn't
+  collect new type constraints from the context, and
+- instead of the actual bound, it uses its best-effort approximation,
+  eliminating the recursive properties of F-bounded type variables.
+
+A convenient way to graphically represent CSSTV is by a rectangle,
+where three sides of it take the three inputs relevant to this
+proposal, and the fourth side produces the output. The relevant inputs
```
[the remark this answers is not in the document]
chloestefantsova: Thanks! Done.

accepted/future-releases/3009-inference-using-bounds/design-document.md
```diff
@@ -0,0 +1,620 @@
...
+  constraintsUpwardsUpdated -- "(3) Proceed with CSSTV using\nthe updated constraint set" --> csstvUpwards
+
+  csstvUpwards -- "{X = B}" --> overallOutput
+```
+
+## More examples enabled by the current proposal
+
+In addition to the motivating examples, both the original and the more
+complex, being handled by the current proposal, some cases of non
+F-bounded type variables are enabled by it as well.
```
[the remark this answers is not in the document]
chloestefantsova: Thanks for the suggestion! Done.

accepted/future-releases/3009-inference-using-bounds/design-document.md
chloestefantsova: Thanks for the suggested rewrite! I like it much more than the previous version. Done.

chloestefantsova:
Thanks for the review and the comments, everybody!

PR #4157, opened 2024-11-07
Add an error and adjust the terminology slightly about augmented redirecting constructors

eernstg:
This PR introduces an error for the case where a redirection (`this(...)`) is augmented by another redirection, and similarly for a redirecting factory.

This follows the treatment of super-initializers, for which it is also an error to include one in an augmenting declaration if the augmented declaration already has an explicitly declared super-initializer.

Other than that, it clarifies the terminology of the augmentation feature specification slightly, such that we keep `this(...)` and `this.x = 42` clearly separated.

'redirecting initializer' is changed to 'redirection' when talking about generative redirecting constructors. The term is derived from `<redirection>`, and the word redirection is used in other specification documents about this kind of term. This also helps avoiding the confusion about `this(...)` possibly being an initializer list element (like `x = 42` or `assert(b)`).

I also added a few words of commentary right after the phrase 'factory redirection' because this phrase is new (the grammar rule actually doesn't have a separate name for the token sequence covered by this phrase). Finally, I deleted a TODO comment which is obsolete today (there are no special grammar rules about `augmented`).

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -947,7 +947,7 @@ constructor.
 
 It is a compile-time error if:
 
-*   The augmented factory constructor has a body, or it is already redirecting.
+*   The augmented factory constructor has a body, or it is redirecting.
 
 #### Extension types
 
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -932,7 +932,8 @@ potentially non-redirecting property of the constructor.
 
 It is a compile-time error if:
 
-*   The augmented constructor has any initializers or a body.
+*   The augmented constructor has an initializer list or a body, or it has a
+    redirection.
 
 #### Redirecting factory constructors
 
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -855,8 +855,8 @@ It is a compile-time error if:
     replace a declared super constructor._ **(TODO: Why not? We allow
     "replacing implementation", and this is *something* like that.)**
 
-*   The resulting constructor is not valid (has a redirecting initializer and
-    other initializers, multiple `super` initializers, etc).
+*   The resulting constructor is not valid (has a redirection as well as
+    some initializer list elements, multiple `super` initializers, etc).
 
 *   A non-redirecting constructor augments a constructor which is not
     potentially non-redirecting.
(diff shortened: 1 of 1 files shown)


working/augmentation-libraries/feature-specification.md
```diff
@@ -923,20 +924,21 @@ present), and it may invoke the augmented body by calling
...
+to the augmented constructor.
 
 This converts it into a redirecting generative constructor, removing the
 potentially non-redirecting property of the constructor.
 
 It is a compile-time error if:
 
-*   The augmented constructor has any initializers or a body.
+*   The augmented constructor has an initializer list or a body, or it has a
+    redirection.
```
eernstg, line 936: I assume this error was intended: "adds its redirection" sounds like the redirection position isn't currently occupied, otherwise we would say "adds or replaces its redirection". Also, the implied statement that the augmented constructor is potentially non-redirecting would not be true if it is already a redirecting generative constructor.

working/augmentation-libraries/feature-specification.md
```diff
@@ -945,7 +947,7 @@ constructor.
 
 It is a compile-time error if:
 
-*   The augmented constructor has a body.
+*   The augmented factory constructor has a body, or it is already redirecting.
```
eernstg: Similar to the generative redirecting case.

working/augmentation-libraries/feature-specification.md
```diff
@@ -945,7 +947,7 @@ constructor.
 
 It is a compile-time error if:
 
-*   The augmented constructor has a body.
+*   The augmented factory constructor has a body, or it is redirecting.
```
eernstg, line 950: Similar to the generative redirecting case.

eernstg (commented):
Commented on two changes.

working/augmentation-libraries/feature-specification.md
munificent: I think @jakemac53 originally wrote this section but that logic sounds right to me.

munificent (approved):
Thanks for clarifying these weird corners!

jakemac53 (approved):

working/augmentation-libraries/feature-specification.md
eernstg: Sounds good, thanks!

eernstg:
Thanks!

eernstg:
@sgrekhov, this should unblock a couple of PRs.

PR #4169, opened 2024-11-22
Revise "declaring constructors".

munificent:
-   Apply review feedback from Lasse.
-   Add section for inferring public parameter names from private ones.
-   Update grammar to allow `var` followed by a type.
-   Add lint for using `final` on parameters.

### DIFF
--- a/working/declaring-constructors/feature-specification.md
+++ b/working/declaring-constructors/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom (based on ideas by Lasse, Nate, et al)
 
 Status: In-progress
 
-Version 0.1 (see [CHANGELOG](#CHANGELOG) at end)
+Version 0.2 (see [CHANGELOG](#CHANGELOG) at end)
 
 Experiment flag: declaring-constructors
 
@@ -253,11 +253,79 @@ with factory constructors, redirecting constructors, initializing formals, and
 super parameters, the amount of constructor-related syntactic sugar is getting a
 little silly.
 
+### Private field parameters and initializing formals
+
+While we're touching constructors, we have the opportunity to fix a
+long-standing annoyance. Initializing formals are common and well-loved, but
+they fail when you want to initialize a *private* field using a *named*
+parameter. The field's name starts with `_`, which isn't allowed for a named
+parameter. Instead, you are forced to write explicit initializers like:
+
+```dart
+class House {
+  int? _windows;
+  int? _bedrooms;
+  int? _swimmingPools;
+
+  House({
+    int? windows,
+    int? bedrooms,
+    int? swimmingPools,
+  })  : _windows = windows,
+        _bedrooms = bedrooms,
+        _swimmingPools = swimmingPools;
+}
+```
+
+Note also that the author is forced to type annotate the parameters as well
+since they are no longer inferred from the initialized field.
+
+When I last [analyzed a corpus][corpus private], 17% of all field initializers
+in initializer lists were doing nothing but shaving off a `_`. There is an
+obvious intended semantics here: simply remove the `_` from the named parameter
+but keep it for the initialized field. Likewise, for declaring constructors,
+the induced field keeps the `_` while the parameter name loses it. That turns
+the above example into:
+
+[corpus private]: https://github.com/dart-lang/language/blob/db9f63185707c4c89a69118e842e4cc6e0e59cc3/resources/instance-initialization-analysis.md
+
+```dart
+class House {
+  int? _windows;
+  int? _bedrooms;
+  int? _swimmingPools;
+
+  House({this._windows, this._bedrooms, this._swimmingPools});
+}
+```
+
+And when combined with a declaring constructor:
+
+```dart
+class House {
+  this({
+    var int? _windows,
+    var int? _bedrooms,
+    var int? _swimmingPools,
+  });
+}
+```
+
+While this is a tiny sprinkle of syntactic sugar, it has a deeper value.
+Initializing formals and declaring constructors are so much more concise that
+users will want to use them whenever they can. But if they don't support private
+fields and named parameters, then we are incentivizing users to make instance
+fields public that they might otherwise prefer to keep private.
+
+It's a well-established software engineering principle to minimize public state,
+so we don't want the language to discourage users from encapsulating fields.
+
 ## Syntax
 
 The syntax changes are small—just using `this` instead of the class name in a
-constructor. But weaving it into the grammar is a little complicated because
-some kinds of constructors can't be declaring:
+constructor and allowing `var` on a parameter along with a type. But weaving it
+into the grammar is a little complicated because some kinds of constructors
+can't be declaring:
 
 ```
 classMemberDeclaration ::=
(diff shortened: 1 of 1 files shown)


working/declaring-constructors/feature-specification.md
```diff
@@ -278,11 +352,44 @@ the parameter. A declaring constructor can be `const` or not. It can't be
...
+
+*We could allow `late` on field parameters and have that apply to the instance
+field (but not the parameter since parameters are always initialized). That
+could be useful in theory if there are other generative constructors that don't
+initialize the field. But to avoid confusion, we simply don't allow it. If a
+user wants a `late` instance field, they can always declare it outside of the
+declaring constructor.*
+
+It is a compile-time error for a `simpleFormalParameter` to have both `var`
+and a type outside of a declaring constructor.
```
lrhn, line 379: Or not. Like, who cares? Unless we want to reserve the syntax for something else.
...

working/declaring-constructors/feature-specification.md
```diff
@@ -278,11 +352,44 @@ the parameter. A declaring constructor can be `const` or not. It can't be
...
+
+*This keeps the normal parameter list grammar consistent with other variable
+declarations. We could allow `var int x` as a parameter outside of a declaring
+constructor, but doing so would be confusing because it looks like a field
+parameter but isn't. Ideally, we would also disallow `final` on parameters
+outside of declaring constructors, but doing so is a breaking change.*
```
lrhn, line 385: Disallowing `final` would probably have to be combined with `final`-as-default, otherwise *some* people will be complaining (because we'd have to choose being mutable as the behavior for all parameters, lest we break a lot of code).
(Or they won't notice and won't complaine, because the same people who insist on `final` on all other local variables also consistently forget to put it on parameters. Maybe it's just because it's not linted.)

working/declaring-constructors/feature-specification.md
```diff
@@ -329,12 +439,79 @@ It is a compile-time error if:
...
+    left alone and also used for the initialized or induced field. *If the
+    parameter is named, this is a compile-time error.*
+
+*For example:*
+
+```dart
+class Id {
+  late final int _region = 0;
+
+  this({this._region, final int _value});
```
lrhn, line 491: Can all add a use of the initializer-list variable, like:
```dart
  this({this._region, final int _value}) : assert(_region > 0 && _value > 0);
```
to show that the local variable is *p*.

working/declaring-constructors/feature-specification.md
```diff
@@ -363,10 +540,19 @@ a *very* small number of authors use.)
...
 problem in practice.
 
+### Private field parameters and initializing formals
+
+Any existing initializing formals with private names must be positional since
+it's a compile-time error to have a private named parameter. Since those
+arguments are passed positionally, the change to give the parameter a public
+name has no effect on any callsites.
+
+Generated documentation may change, but that should be harmless.
```
lrhn, line 550: True. It won't break.
If we have `C(this._foo)` today, documentation saying `[_foo]` will now refer to the field, not to the parameter. It won't become an error, and it does point to *something* relevant, but it might be worth telling people that they should migrate to `[foo]` to refer to the parameter.

(Or DartDoc can be lenient and let both `[_foo]` and `[foo]` refer to the parameter, and show them both as `foo` in the output. That could be the starting point, then we can start warning about `[_foo]` and quick-fixing it when `[foo]` starts working. Maybe make `[_foo]` go back to referring to the field in Dart 4.0.)

working/declaring-constructors/feature-specification.md
```diff
@@ -409,8 +595,31 @@ become a declaring one:
...
+itself non-assignable.
+
+If declaring constructors become popular, then users will likely start to see
+`final` before a parameter and read it as a field parameter. They will then be
+confused if the parameter isn't actually in a declaring constructor and isn't
+a field parameter.
+
+Given that non-assignable parameters aren't actually that *useful*, it may be
+worth discouraging users from marking a parameter with `final` unless it is a
+field parameter. This seems like a good candidate for a lint.
```
lrhn, line 612: If nobody writes `final` in front of parameters today, there is no problem.
If they do, it's because they want the parameter to be a final variable (because they like final variables), and then they'll probably get annoyed if they're told not to.

That measn the lint wont' be a recommended lint. It's too divisive. 
Will _anybody_ then use such a lint?
(And will there be another lint to always put `final` on parameters, unless they are mutated inside the body, because people who like `final` got reminded that they forgot that parameters are local variables?)

Not sure there is an issue at all. I'd definitely wait and see if anyone *asks* for a lint, and if so, which one.

working/declaring-constructors/feature-specification.md
```diff
@@ -329,12 +439,79 @@ It is a compile-time error if:
...
+underscore does not leave something which is is a valid identifier *(as in `_`
+or `_2x`)* or leaves another private name *(as in `__x`)*, then the private name
+has no corresponding public name.
+
+The private declared name, *p*, of an initializing formal or field parameter in
+constructor C has a corresponding *non-conflicting public name* if it has a
+corresponding public name, *n*, and no other parameter of the same constructor
+declaration has either of the names *p* or *n* as declared name. *In other
+words, if removing the `_` leads to a collision with another parameter, then
+there is a conflict.*
```
lrhn, line 460: (And also if not removing it leads to a conflict, `foo(this._foo, int _foo);` is a conflict because the local `_foo` names clash in the initializer list scope. It's always a clash if two parameters have the same name. Now it's also a "clash" if the corresponding public name of a privately named parameter is the same as the name of another parameter. It won't necessarily cause an error, it may just make the corresponding public name not be used: `Foo(this._foo, int foo):` is valid today, and will stay valid with a privately named first parameter.)

lrhn (approved):
LGTM, ship it!
(I'll file issues for further discussions about the few details that I could see being slighlty different.)

working/declaring-constructors/feature-specification.md
munificent: Basically, I just didn't want to change the grammar outside of what this feature adds. Since you can't do `var int x` in a parameter today, this sentence preserves that (outside of field parameters). Let's discuss it more on the issue below.

working/declaring-constructors/feature-specification.md
```diff
@@ -278,11 +352,44 @@ the parameter. A declaring constructor can be `const` or not. It can't be
...
+user wants a `late` instance field, they can always declare it outside of the
+declaring constructor.*
+
+It is a compile-time error for a `simpleFormalParameter` to have both `var`
+and a type outside of a declaring constructor.
+
+*This keeps the normal parameter list grammar consistent with other variable
+declarations. We could allow `var int x` as a parameter outside of a declaring
+constructor, but doing so would be confusing because it looks like a field
+parameter but isn't. Ideally, we would also disallow `final` on parameters
```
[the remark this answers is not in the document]
munificent: Discussing this on the issue SGTM.

working/declaring-constructors/feature-specification.md
```diff
@@ -296,11 +403,14 @@ Given a `declaringConstructor` D in class C:
...
-            the implicitly declared field overrides an inherited getter.*
+        *   Any doc comments and metadata annotations on P are also applied to
+            F. *For example, a user could mark the parameter `@override` if the
+            the implicitly declared field overrides an inherited getter. If a
+            user wants to document the instance field induced by a field
+            parameter, they can do so by putting a doc comment on the
+            parameter.*
 
-        *   P comes an initializing formal that initializes F.
+        *   P now behaves like an initializing formal that initializes F.
```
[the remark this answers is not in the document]
munificent: > Or rather, I'll want to abstract out the "initialize a variable to a value" operation, and say that this.x and final int x both initialize the instance variable x with the argument value (or the default value, if optional and has no argument). No "desugaring" one to the other, but define both in terms of the same defined operation. (Prefer composition to inheritance, also in specification 👍)

SGTM!

> The open questions include: In which ways does it behave like an initializing formal? Which initializing formal?

Spelled it out more specifically.

working/declaring-constructors/feature-specification.md
```diff
@@ -329,12 +439,79 @@ It is a compile-time error if:
...
+        a private named parameter.*
+
+    *   The local variable in the initializer list scope of C is *p*. *Inside
+        the body of the constructor, uses of *p* refer to the field, not the
+        parameter.*
+
+    *   If the parameter is an initializing formal, then it initializes a
+        corresponding field with name *p*.
+
+    *   Else the field parameter induces an instance field with name *n*.
```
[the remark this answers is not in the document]
munificent: Oops! Good catch.

PR #4171, opened 2024-11-25
Add commentary to clarify the rule about `.` at the beginning of expression statements

eernstg:
This PR adds a sentence of commentary to clarify the static access shorthand rule "We also add `.` to the tokens that an expression statement cannot start with". The point is that `.` as a token does not occur in numbers like `.123`, because the number is a single token.


eernstg:
Closing: Done in https://github.com/dart-lang/language/commit/02c4ada9efe6265788571cc062c4ef8889e39fb5.

PR #4182, opened 2024-11-28
Update example in Static access shorthand proposal

sgrekhov:
`Endian.little` is not an enum value but a static constant. Fixed comment and added a real enum value example.


sgrekhov:
Replaced by https://github.com/dart-lang/language/pull/4191

PR #4210, opened 2024-12-17
Adjust function literal return type inference to avoid spurious null

eernstg:
See https://github.com/dart-lang/sdk/issues/59669 where this topic came up.

This PR changes one item in the list of actions taken during function literal return type inference: A `return;` statement _only_ adds `Null` to the return type in cases where the given function literal is a non-generator.

With the current version, `Null` is added to the return type also in cases like `() sync* { yield 1; return; }` such that this function literal gets the inferred return type `Iterable<int?>`. This is an unnecessary loss of typing precision because null is never actually added to the returned iterable. With this update, the inferred return type is `Iterable<int>`.

### DIFF
--- a/resources/type-system/inference.md
+++ b/resources/type-system/inference.md
@@ -6,6 +6,11 @@ Status: Draft
 
 ## CHANGELOG
 
+2024.12.17
+  - Change the function literal return type inference rules to ignore
+    `return;` statements in generators (it doesn't actually cause null to be
+    returned).
+
 2022.05.12
   - Define the notions of "constraint solution for a set of type variables" and
     "Grounded constraint solution for a set of type variables".  These
@@ -324,7 +329,8 @@ schema.
     `e`, using the local type inference algorithm described below with typing
     context `K`, and update `T` to be `UP(flatten(S), T)` if the enclosing
     function is `async`, or `UP(S, T)` otherwise.
-  - For each `return;` statement in the block, update `T` to be `UP(Null, T)`.
+  - If the enclosing function is not marked `sync*` or `async*`: For each
+    `return;` statement in the block, update `T` to be `UP(Null, T)`.
   - For each `yield e;` statement in the block, let `S` be the inferred type of
     `e`, using the local type inference algorithm described below with typing
     context `K`, and update `T` to be `UP(S, T)`.


resources/type-system/inference.md
```diff
@@ -324,7 +329,8 @@ schema.
     `e`, using the local type inference algorithm described below with typing
     context `K`, and update `T` to be `UP(flatten(S), T)` if the enclosing
     function is `async`, or `UP(S, T)` otherwise.
-  - For each `return;` statement in the block, update `T` to be `UP(Null, T)`.
+  - If the enclosing function is not marked `sync*` or `async*`: For each
+    `return;` statement in the block, update `T` to be `UP(Null, T)`.
```
lrhn, line 333: I really think it should change the return type to `void`, but I guess that's a bigger change.

resources/type-system/inference.md
```diff
@@ -324,7 +329,8 @@ schema.
     `e`, using the local type inference algorithm described below with typing
     context `K`, and update `T` to be `UP(flatten(S), T)` if the enclosing
     function is `async`, or `UP(S, T)` otherwise.
-  - For each `return;` statement in the block, update `T` to be `UP(Null, T)`.
+  - If the enclosing function is not marked `sync*` or `async*`: For each
```
lrhn, line 332: The "for each" seems redundant. How about "if the body contains any `return;` statement ..."?

resources/type-system/inference.md
lrhn, line 333: Not: "in the block" is imprecise. It can be read to include the bodies of nested functions, and it can be read as only applying to immediate sub-statements of the block. 
Maybe "`return;` statements returning from this function" or something?

lrhn (approved):

lrhn:
Definitely a spec bug, never intended to make any difference to the return type that a generator contains a return.

resources/type-system/inference.md
eernstg: Perhaps that's already covered pretty well?

For function literal return type inference, the type `void` is given special treatment: If the context type has a return type which is `void` then the return type of the function literal is also taken to be `void`.

```dart
void main() {
  void Function() f = () { return; }; // Return type of function literal is `void`.
  f = () { return 3; }; // Error, can't return an `int` from a void function.
}
```

When the return type from the context type is any other type, the inferred return type of the function literal will not be void unless it actually returns an expression of that type.

```dart
void main() {
  Object? Function() f = () { return; };
  print(f.runtimeType); // Something that means `Null Function()`.
  f = () { return print('Hah!'); };
  print(f.runtimeType); // Something that means `void Function()`.
}
```

This means that the function will always be treated as a void function when this is what the context expects, and it will almost always be treated as a non-void function when this is what the context expects.

In the case where a `return;` statement occurs in a function literal whose inferred return type is not void, we already have other diagnostics:

```dart
void main() {
  int? Function(bool b) f = (bool b) {
    if (b) {
      return; // Error: The return value is missing after `return`.
    } else {
      return 1;
    }
  };
}
```

I think this would already deal with the potential code smells that you were targeting.

resources/type-system/inference.md
eernstg: All four items in this list use a similar quantifier (actually, they all use 'for each'). How would it be redundant? We're running an algorithm, and at this point it needs to iterate over the function body to find all occurrences of `return ...`. Do you think it would improve the overall text to use a natural language version of recursion, rather than the iteration-ish "for each"?

resources/type-system/inference.md
eernstg: Right, done!

eernstg:
Updated, PTAL.

leafpetersen:
What do the two implementations do?  I think from the linked issue, at least the analyzer follows the old spec?  So this is a breaking change?  No objection to the change in principle, but we need to be sure that it gets tracked as a breaking change, and scheduled for implementation as needed.

eernstg:
> What do the two implementations do?  ... the analyzer follows the old spec

Right, the analyzer includes `?` in the inferred element type, the CFE does not. This would imply that there is no change in the dynamic semantics, and also that all existing programs where the CFE has been used (that's all programs, I guess?) will pass without errors if the analyzer is updated as proposed here.

Double checking, the following program (based on the current specification) is accepted by the analyzer, but the CFE reports errors as shown:

```dart
// ignore_for_file: unused_element

void main() {
  //==================== sync* ===============================

  withOutReturnSync() sync* {
    yield 1;
  }
  withOutReturnSync.expectStaticType<Exactly<Iterable<int> Function()>>;

  withReturnSync() sync* {
    yield 1;
    return;
  }
  withReturnSync.expectStaticType<Exactly<Iterable<int?> Function()>>; // Analyzer OK, CFE error.

  //===================== async* =================================

  withoutReturnASync() async* {
    yield 1;
  }
  withoutReturnASync.expectStaticType<Exactly<Stream<int> Function()>>;

  withReturnASync() async* {
    yield 1;
    return;
  }
  withReturnASync.expectStaticType<Exactly<Stream<int?> Function()>>; // Analyzer OK, CFE error.
}

extension<X> on X {
}
```

eernstg:
Based on my previous comment, it seems likely to me that nothing will break. So we could run the breaking change process with an attached "but, most likely, nothing will actually break" plus an explanation. Or we could test internally and conclude that it isn't actually breaking if nothing breaks there.

eernstg:
I think we should be able to proceed with this one.

Checking again, the following program expects the behavior which is being specified if this PR is landed and it is compiled and executed without errors by the CFE; however, the analyzer behaves according to the current specification (and would need to be adjusted if and when this is landed).

```dart
// ignore_for_file: unused_element

void main() {
  withoutReturnSync() sync* {
    yield 1;
  }
  withoutReturnSync().expectStaticType<Exactly<Iterable<int>>>();

  withReturnSync() sync* {
    yield 1;
    return;
  }
  withReturnSync().expectStaticType<Exactly<Iterable<int>>>(); // The analyzer fails here.

  Iterable<int> withReturnSync2() sync* {
    yield 1;
    return;
  }
  withReturnSync2().expectStaticType<Exactly<Iterable<int>>>();

  withoutReturnAsync() async* {
    yield 1;
  }
  withoutReturnAsync().expectStaticType<Exactly<Stream<int>>>();

  withReturnAsync() async* {
    yield 1;
    return;
  }
  withReturnAsync().expectStaticType<Exactly<Stream<int>>>(); // The analyzer fails here.

  Stream<int> withReturnAsync2() async* {
    yield 1;
    return;
  }
  withReturnAsync2().expectStaticType<Exactly<Stream<int>>>();
}

extension<X> on X {
}
```

Given that the CFE already follows the proposed specification (at compile time and run time), no programs should break if we land this and adjust the analyzer.

eernstg:
@dart-lang/language-team, WDYT, can we proceed to land the spec, test the breakage on internal code (no breakage is expected), and adjust the analyzer (if it is true that no breakage is observed)?

natebosch:
If the CFE already implements the expected behavior I agree that it should not cause problems to roll this out, and using internal code as a double check sounds good to me.

eernstg:
@lrhn, do you have further comments on this one?

eernstg:
@leafpetersen, I'd like to move on with this change. We could initiate a breaking change process, or we could rely on the fact that the CFE already has the behavior which is specified if the PR is landed. WDYT?

eernstg:
@leafpetersen, I'd like to push this one forward. WDYT, do we need a breaking change process?

leafpetersen (approved):
LGTM, thanks!  cc @scheglov @johnniwinther @chloestefantsova for any concerns.

scheglov:
https://dart-review.googlesource.com/c/sdk/+/446183

PR #4212, opened 2024-12-19
Update the static extensions proposal to rely on generic constructors

eernstg:
This PR updates the variant of the static extensions proposal that generalizes the `extension` declaration (not `static extension`).

Note that it is not yet ready for review: About half of the document has been updated to use these new ideas, the other half will still be modified substantially.


eernstg:
I'll land this update now (following the approach used with other 'working' documents).

PR #4226, opened 2025-01-08
Remove sentence about `e!` being constant

eernstg:
The feature specification about static access shorthands mentions constant expressions of the form `e!`. They do not exist so there's no need to discuss them. This PR removes that sentence.


lrhn:
Damn, I thought `!` would be constant. Means I have to do `as int` instead of `!` to go from `int?` to `int` in potentially constant expressions.

lrhn (approved):

PR #4236, opened 2025-01-24
Fixed one typo in inference.md

AxeelPy:
* Corrected "iff" to "if"


albertms10:
Isn't _iff_ meant to mean _if and only if_? 🤔

mateusfccp:
Related: #968.

AxeelPy:
Oooh I see, thanks for the clarification

PR #4259, opened 2025-02-10
Add tip for adding feature onto tracking board.

kallentu:
Adding an extra tip here so we don't forget. I keep missing this step and coming back to it.

@itsjustkevin may add this part automatically through the project generation script in Aviary, but we can leave this here until that happens. It's not too hard to do manually anyways.


itsjustkevin (approved):
Thank you!

PR #4262, opened 2025-02-12
With "Unquoted Imports" disallow unquoted part of library names.

### LINKED ISSUE #4038: URI shorthands, not for parts
TL;DR: Don't allow `part` and `part of` to use shorthand URIs. They should always be relative URI references anyway. Added benefit is discontinuing the old syntax without immediately allowing the same syntax with a new meaning, making migration easier.

The [URI shorthand spec](https://github.com/dart-lang/language/blob/main/working/unquoted-imports/feature-specification.md) allows the shorthand to be used for `import`, `export`, `part` and `part of` directives.

An unquoted URI shorthand can be used as shorthand for a `dart:` or `package:` URI. It cannot represent relative URI references.

The spec mentions parsing of `part of` being tricky.
It removes support for `part of library.name;` that refers to the name of a library introduced by its library declaration, and instead interprets `part of library.name;` as `part of "package:library.name/name.dart";`.
It also needs look-ahead because `part of ...` could mean `part "package:of/of.dart" ...`, so parsing will have to disambiguate by looking at what follows. 
If followed immediately by `;` it means `part "package:of/of.dart";`.
If it's followed by one identifier/reserved word only, before a `;`, then that identifier is a shorthand URI package name for that parent file of a `part of ` directive.
If it's followed by an identifier/reserved word and then `.` or `/`, it's definitely a URI shorthand for the parent file.
If it's followed by a `if (` then it's a `part` directive.

This is a lot of complication for something that will likely *never* be used.

With "parts-with-imports" it's made explicit that a part must be in the same package as its parent file. (It's a requirement made by the language versioning feature, not a single language version, a language by itself doesn't care, and the requirement has always been assumed, but not explicitly stated.) Most likely, the part and parent files must both be inside or outside of `lib/` too.

That means that a `part` directive, and a `part of` directive, always can and _should_ use a relative URI reference, which means it won't be using a shorthand. (Not unless/until we introduce a relative shorthand like `./foo`, at which point I'd allow `part`/`part of` to use *relative* shorthands only.)

I'd still disallow `part of library.name;` because it looks ambiguous, and because we'll be removing it in "parts with imports" anyway, so we might as well remove it here, if this feature launches first.

The benefit would be:
* Less parsing complexity. The `part of` is unambiguously starting a `part of` directive.
* An existing `part of pkg.foo;` would become invalid, not *change meaning*. That makes migration easier.
* Writing `part of;` by accident (perhaps removing the now invalid `pkg.foo`) would not become valid syntax for using `package:of/of.dart` as a part. (It probably won't succeed, causing some other downstream compilation error, so tools can still recognize and report it specially if they want to.)
* Pushes people away from using `part foo;`,
    * which they might mistakenly assume means `part "foo.dart";`.
    * or if they do mean `package:foo/foo.dart`, force them to write that out, which would encourage using the relative URI.
  
The drawback is lack of orthogonality. In some places where you can use a URI, you can use a shorthand.
In other places you cannot.
With "parts-with-imports" parts can have conditional URIs, removing an existing lack of orthogonality, so adding another one here feels counter-productive.

However, of `import`, `export` and `part`, one is not like the other two. Both `import` and `export` refer to *libraries*, potentially from other packages (even likely, otherwise you can use a relative import anyway, and don't need the shorthand so much), where `part` refers to *files* inside the current package. Shorthand URIs are for full `package:` URIs, and `part`/`part of` never need those, and shouldn't use them. Not allowing you to do something that you shouldn't be doing anyway, isn't much of a loss. At best, nobody ever notices, because they didn't try. Or we catch someone writing something unquoted that they think is relative, when it really isn't.

We can explicitly use the narrative that unquoted URIs are for *libraries*, not for files. We give people a new syntax for referring to *libraries* in other packages. We don't give them a syntax to refer to *files* in other packages, because that's not something you can do. The syntax can be used for libraries in the current package, but also not for arbitrary files.

We *can* allow `part of unquotedUri;` later if we want to. 
Disallowing it now, at the point where we discontinue `part of library.name;`, makes that a clean cut, and that alone should be reason enough to not give the same syntax new meaning immediately. At least wait until actively maintained code has changed all their `part of` directives to URIs.

@dart-lang/language-team

munificent:
In #4038, Lasse suggests that we shouldn't allow unquoted `part of` directives but change what they mean to be package references since that will likely just break silently and confuse users. Unquoted imports (with the new proposed semantics) in part directives aren't really useful anyway since parts should always be in the same package and the quoted syntax is shorter for that.

So this PR changes the proposal to not apply the new syntax to `part` and `part of` directives. It also makes a language versioned breaking change to remove support for unquoted library names in `part of` directives entirely since the syntax is long deprecated, problematic, and would become very confusing in the context of other unquoted imports.

@dart-lang/language-team, PTAL since this proposal is already accepted so a change should be run by the whole team.

Fix #4038.

### DIFF
--- a/accepted/future-releases/unquoted-imports/feature-specification.md
+++ b/accepted/future-releases/unquoted-imports/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: Accepted
 
-Version 0.4 (see [CHANGELOG](#CHANGELOG) at end)
+Version 0.5 (see [CHANGELOG](#CHANGELOG) at end)
 
 Experiment flag: unquoted-imports
 
@@ -341,8 +341,8 @@ after the directives outside of the path. These are all valid:
 
 ```dart
 import /* Weird but OK. */ some/path;
-export some/path; // Hi there.
-part some/path // Before the semicolon? Really?
+import some/path; // Hi there.
+export some/path // Before the semicolon? Really?
     ;
 ```
 
@@ -353,11 +353,11 @@ parse while looking like a single opaque "unquoted string" to users and tools.
 
 The normative stuff starts now. Here is the proposal:
 
-We add a new rule and hang it off the existing `uri` rule already used by import
-and export directives:
+We add a new rule and hang it off the existing `configurableUri` rule already
+used by import and export directives:
 
 ```
-uri               ::= stringLiteral | packagePath
+configurableUri   ::= ( stringLiteral | packagePath ) configurationUri*
 packagePath       ::= pathSegment ( '/' pathSegment )*
 pathSegment       ::= segmentComponent ( '.' segmentComponent )*
 segmentComponent  ::= IDENTIFIER
@@ -371,22 +371,10 @@ between any of the `segmentComponent`, `/`, or `.` tokens in a `packagePath`.
 *In other words, there can be nothing except the terminals themselves from the
 first `segmentComponent` in the `packagePath` to the last.*
 
-*An import, export, or part directive can continue to use a `stringLiteral` for
-the quoted form (which is what they will do for relative references). But they
-can also use a `packagePath`, which is a slash-separated series of segments,
-each of which is a series of dot-separated components.*
-
-### Part directive lookahead
-
-*There are two directives for working with part files, `part` and `part of`.
-This means that when the parser sees `part of`, it doesn't immediately know if
-it is looking at a `part` directive followed by an unquoted identifier like
-`part of;` or `part of.some/other.thing;` versus a `part of` directive like
-`part of thing;` or `part of 'uri.dart';` It must lookahead past the `of`
-identifier to see if the next token is `;`, `.`, `/`, or another identifier.*
-
-*This may add some complexity to parsing, but should be minor. Dart's grammar
-has other places that require much more (sometimes unbounded) lookahead.*
+*An import or export directive can continue to use a `stringLiteral` for the
+quoted form (which is what they will do for relative references). But they can
+also use a `packagePath`, which is a slash-separated series of segments, each of
+which is a series of dot-separated components.*
 
 ## Static semantics
 
@@ -461,8 +449,7 @@ There are no runtime semantics for this feature.
 
 ## Compatibility
 
-This feature is fully backwards compatible for `import`, `export`, and `part`
-directives.
+This feature is fully backwards compatible for `import` and `export` directives.
 
 For all directives, we still allow quoted "dart:" and "package:" imports. Users
 may be compelled to use the existing syntax in uncommon corner cases where the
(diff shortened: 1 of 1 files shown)


accepted/future-releases/unquoted-imports/feature-specification.md
```diff
@@ -353,11 +353,11 @@ parse while looking like a single opaque "unquoted string" to users and tools.
...
 The normative stuff starts now. Here is the proposal:
 
-We add a new rule and hang it off the existing `uri` rule already used by import
-and export directives:
+We add a new rule and hang it off the existing `configurableUri` rule already
+used by import and export directives:
 
 ```
-uri               ::= stringLiteral | packagePath
+configurableUri   ::= ( stringLiteral | packagePath ) configurationUri*
```
lrhn, line 360: This will only fix it for a brief time since "enhanced parts" allows `part` directives to be configurable.
(If enhanced parts won't lands first.)

Maybe just use the existing grammar, and say that "It's a compile-time error if a `part` or `part of` directive contains a `uri` which is a `packagePath`."

Then the parser can assume that `part`/`part of` must be followed by a string literal, and it's all error recovery if it isn't. (If anyone uses `part mypackagename/src/foo_part;` we can probably tell then what is wrong.)

accepted/future-releases/unquoted-imports/feature-specification.md
```diff
@@ -514,33 +500,33 @@ recommended URI syntax and would not be affected by this change:
...
 
-Given that, I propose that we make a **breaking change** and remove support for
-the long-deprecated library name syntax from `part of` directives. An unquoted
-series of identifiers after `part of` then gets unambiguously interpreted as
-this proposal's semantics. In other words, `part of foo.bar;` is part of the
-library at `package:foo/bar.dart`, not part of the library with name `foo.bar`.
+To avoid potential user confusion and breakage, the new syntax in this proposal
+isn't allowed in `part` and `part of` directives. Further, we make a
+**language-versioned breaking change** and remove support for unquoted library
+identifiers in `part of` directives completely. That way, a user never sees
```
lrhn, line 510: Huzzah! :partying_face: 
(And good riddance!)

accepted/future-releases/unquoted-imports/feature-specification.md
lrhn, line 360: Would also need to update `configurationUri` to allow both quoted URIs and package-paths.
Current grammar is:
```ebnf
<uri> ::= <stringLiteral>
<configurableUri> ::= <uri> <configurationUri>*
<configurationUri> ::= \IF{} `(' <uriTest> `)' <uri>  ;; <-- second occurrence of `uri`.
```

(We really should have a parameterized grammar:
```dart
<configurableRef>[ref] ::=  ref <configurationRef>[ref]*
<configurationRef>[ref] ::= 'if' '(' <uriTest> ')' ref
```
so you could write:
```dart
<importDirective> ::= <metadata> 'import' <configurableRef>[<stringLiteral>|<packagePath>] ...
<partDirective> ::= <metadata> 'part' <configurableRef>[<stringLiteral>] ';'
```
Would avoid a lot of parallel duplication in the grammar.)

accepted/future-releases/unquoted-imports/feature-specification.md
```diff
@@ -514,33 +500,33 @@ recommended URI syntax and would not be affected by this change:
...
+syntax in a `part of` file that *looks* like a package path but means something
+else.
 
-Users affected by the breakage can and should update their `part of` directive
-to point to the URI of the library that the file is a part of, using either the
-quoted or unquoted syntax.
+Since part files should always be part of the same package that contains the
+library that owns them, the new unquoted syntax is of limited use for part files
+anyway. Quoted relative URIs are shorter and more idiomatic.
```
lrhn, line 516: (It's not actually specified anywhere, yet, that part files *must* be in the same library as the library they're part of. I added that rule to the enhanced parts spec. It'd be fine to add it as a normative rule here too, so whoever lands first gets the rule enforced. The sooner the better.)

lrhn (approved):
Fix does not fix all. Otherwise awesome :)

accepted/future-releases/unquoted-imports/feature-specification.md
munificent: I like capturing as much of the syntax in the grammar as reasonably possible, so I went ahead and grammarified it. But I added a note about the connection to parts-with-imports.

munificent: Good catch. I took a slightly different approach and put `packagePath` inside `uri` so that `configurableUri` and `configurationUri` can be left alone (with a note about how parts-with-imports affects it).

And, yes, parametric grammar rules would be *so* nice.

accepted/future-releases/unquoted-imports/feature-specification.md
munificent: I meant "should" in the sense of "good practice" and not normatively in the language spec. I agree the language should enforce that, but given that it has almost no notion of a package anyway, I'm not sure how I'd specify that. Reworded to make it clear I wasn't being spec-normative.

Issue #4213, opened 2024-12-20
Generic and conditional constructors


eernstg:
This is a proposal about a reinterpretation of constructors in Dart that allows them to have higher expressive power, similar to the additional power of constructors in the proposal about [conditional instance members and constructors](https://github.com/dart-lang/language/issues/2313).

First note that the proposal has no effect on constructors of non-generic classes (and mixin classes, enums, extension types, etc, but I'll just say 'class' here for brevity).

In contrast, the interpretation of constructors of generic classes changes. The core idea is that a constructor does not have access to the type parameters of the enclosing class. Instead, the current notation is just an abbreviated form. When the underlying (non-abbreviated, full) form is used, every constructor must declare all the type parameters they need.

```dart
class A<X1 extends B1 .. Xk extends Bk> ... {
  // Current notation, using an arbitrary example signature.
  A(X1 x, Map<X2, X3> map) {
    .../*May use X1 .. Xk*/...
  }

  // Full notation.
  A<Y1 .. Yk>.new<Y1 extends BB1 .. Yk extends BBk>(Y1 x, Map<Y2, Y3> map) {
    .../*May use Y1 .. Yk*/...
  }
}
```

In the full notation, the name of the constructor must be of the form `<typeIdentifier> '.' <identifierOrNew>`. We can use this kind of name without changing the meaning of the declaration by using `.new` when the original constructor name was of the form `<typeIdentifier>`.

In the full notation, in this case where it corresponds to the current form, the constructor declares the same type parameters with the same bounds as the class, up to alpha equivalence (that is, we can freely rename the type parameters as long as it is a consistent renaming). I've used new names in the example above in order to emphasize that the type parameters `X1 .. Xk` are not available in the constructor which uses the full notation, but the new type parameters play exactly the same role as the type parameters of the class do in the current notation. When it is result of this kind of desugaring, the full notation has the same semantics as the current notation.

However, the full notation allows for more expressive power, because it is possible to specify that a given constructor can only create _some_ generic instantiations of the given class. For example, we might have a very strictly confined constructor that only supports a fixed type:

```dart
class A1<X> {
  final X x;
  A1(this.x);
  A1<int>.name([int this.x = 0]);
}

void main() {
  A1<num> a = A1(10); // This will be an `A1<num>`.
  a = A1.name(); // This will be an `A1<int>`.
  A1<String> aa = A1.name(); // Compile-time error.
}
```

[...]

PR #4265, opened 2025-02-13
Add feature specification of generic constructors

eernstg:
Here is a first version of said feature specification, based on the proposal in https://github.com/dart-lang/language/issues/4213.


eernstg:
I'll land this because it's in the 'working' directory. Discussions about details can be handled in further PRs.

Issue #3100, opened 2023-05-22
Change flow analysis to assume sound null safety once unsound null safety support is fully removed


stereotype441:
Currently, flow analysis conservatively assumes that a variable with a non-nullable type might still be `null` because the program might be running with unsound null safety, and we don't want null-related unsoundness to be able to be escalated into full unsoundness (see https://github.com/dart-lang/language/issues/1143 for details).

Although we don't officially support running with unsound null safety anymore, support has not yet been fully removed. Once we remove it, we should remove this conservative assumption from flow analysis.

lrhn:
I'm actually worried that we assume some syntactically allowed control flow cannot happen because of type analysis.
We went back on that because of unsound null safety, but I also think it was the right idea in general.

If code contains a branch, I think we should assume that *all branches can be taken*, if at all possible.
We currently don't in some cases, predicting that a particular branch will never get executed.
That matters for things like definite assignment, where a variable assigned on every *reachable* branch is considered definitely assigned after the branches has been joined. 

The problem with that is that an unreachable branch is dead code. Dead code is *likely a mistake*, or at least a temporary thing. At some point that code will either be deleted, or the branch will be enabled.
If type inference and definite assignment changes when you remove a temporary `&& false` added to an `if` condition, then you might have to go back and do a larger rewrite.

If type analysis and definite assignment has always considered the branch to be executable, then twiddling with the conditions won't suddenly make your code have type errors.

The one issue I see with such branches, is that it's not always clear which types promotable variables have on entry, since the reason the branch is unreachable might be a type test that rules out everything.

Say (probably not a correct example, but maybe there are valid examples):
```dart
// ignore_for_file: prefer_void_to_null
import "dart:async";

void main() {
  FutureOr<Null> x = null as dynamic;
  int y;
  int? z = 0;
  if (x is Future<Null>) {
    // x promoted to `Future<Nul>`
    y = 0;
  } else {
    // x promoted to `Null`
    if (x != null) {
      // x promoted to `Never` ? (Actually not, it's still `Null`)
      z = null;
    } else {
      y = 2;
    }
  }
  print(y);
  z.toRadixString(10); // Allowed because `z=null` doesn't demote.
}
```

Here `y` is considered definitely assigned, and `z` is not demoted to nullable,
because the type inference/definite assignment decides that the `x != null` cannot be true.

We do analyze the "unreacable" branch (can be seen by introducing type errors there), we just don't merge it back into the result afterwards

[...]

PR #4284, opened 2025-03-04
[flow analysis] Update the spec to match implementation of `==`.

stereotype441:
In fully sound null safety mode, flow analysis should know that a test like `expr == null` is guaranteed to evaluate to `false` if the static type of `expr` is non-nullable. But in unsound null safety mode, no such guarantee can be made.

Since support for unsound null safety was only recently removed from the CFE (see
https://github.com/dart-lang/sdk/commit/0060b0f665ac5b6865fdb9d91e8735a20ee3b160), flow analysis still conservatively assumes that an expression with a non-nullable static type might, nonetheless, still be `null`.

This change updates the spec to match the implementation in this regard, and adds a reference to
https://github.com/dart-lang/language/issues/3100, where we are discussing the possibility of changing the behavior.

---

<details>

</details>


eernstg (approved):
LGTM.

It'll be nice when we can go back to the old rule in the spec _and_ the implementation, though. ;-)

PR #4301, opened 2025-03-24
Add a feature specification for the metaobject proposal

eernstg:
This PR adds a feature specification for the metaobject proposal (formerly known as 'more capable Type objects'). For simplicity, it does not include the part about `Type.reify` (that will be specified separately).


working/4200-metaobjects/feature-specification.md
```diff
@@ -0,0 +1,519 @@
...
+actual type arguments as the ones which are part of `T`.
+
+*For example, if we are evaluating a type parameter `X` as an expression,
+and the value of `X` is a type `C<S1, S2>` that has a metaobject class
+`MetaC` then the result will be an instance of `MetaC<S1, S2>`.*
+
+The metaobject has standard semantics, everything follows the normal rules
+of Dart based on the metaobject class declaration.
+
+### Changelog
```
mateusfccp, line 515: Maybe include the changes of the version 1.1?

mateusfccp: Why is the version in the header marked as 1.1, then? I don't know if I understand.

eernstg: Oops, that's a typo. Fixed.

working/4200-metaobjects/feature-specification.md
```diff
@@ -0,0 +1,519 @@
...
+class MetaobjectForB implements Type, Fooable<B> {
+  const MetaobjectForB();
+  String foo(B b) => B.foo(b);
+  bool operator ==(Object other) => ...;
+  int get hashCode => ...;
+}
+```
+
+These classes are implicitly generated by the compiler and analyzer, so we
+can't refer to them in user code (the real name is probably something like
```
lrhn, line 110: The "so" part doesn't follow from the former part.
Also mixing abstraction levels. The classes are *introduced* by the class declaration that has a `static implements` clause. They're not generated by a compiler, they exist *in the program as such* by the languge semantics.

What compilers do is an implementation detail that nobody other than compiler writers should need to know.

working/4200-metaobjects/feature-specification.md
```diff
@@ -0,0 +1,519 @@
...
+`$xBz8_`, and it's certainly a fresh name such that it doesn't coincide
```
lrhn, line 111: I can refer to `$xBz8_`!
(Just say they are always given fresh names, ones that the program does not refer to. *That's* why you can' refer to them in user code. They may have a name, but it's never the one you have written.)

working/4200-metaobjects/feature-specification.md
```diff
@@ -0,0 +1,519 @@
...
+the exact class (mixin, etc.) declaration. This basically means that we
+have made the static members _virtual_, because we're calling them via
+normal instance members (that is, virtual members, with all the usual OO
+semantics) of the metaobject.
+
+The next example illustrates how we can use metaobjects to call
+constructors in a similar way (yielding 'virtual constructors'):
+
+```dart
+abstract class Creation<X> {
```
lrhn, line 148: Maybe we should be allowed to declare a `static class ...` which cannot be instantiated (it has a freshly-named zero-argument `const` constructor) and must be constant.
It can only be used for `static extends`, maybe `static implements`.

working/4200-metaobjects/feature-specification.md
```diff
@@ -0,0 +1,519 @@
...
+}
+
+class C<Y> static implements Creation<C<Y>> {
+  C();
+  C.named(int _): this();
+}
+
+class D static implements Creation<D> {
+  factory D() = _DImpl;
+  D.named(int _);
```
lrhn, line 161: This could also do:
```dart
  D call() => _DImpl();
```
instead of the factory constructor, right?

working/4200-metaobjects/feature-specification.md
```diff
@@ -0,0 +1,519 @@
...
+
+When a class, mixin, mixin class, or enum declaration has a
+`static implements I` clause, the metaobject class for said type will have
+an `implements I` clause, and it is implemented by generating forwarding
+instance members for each of the members of the interface of `I`.
+
+It is also possible to use a `static extends T` clause, in which case the
+metaobject class will have an `extends T` clause. This implies that the
+metaobject class can inherit behaviors from `T` (and possibly implement
+others as forwarders, with members which are not implemented otherwise).
```
lrhn, line 202: No `static with`? Could be put into the `extends`, as `static extends Foo with Mixin`.

working/4200-metaobjects/feature-specification.md
```diff
@@ -0,0 +1,519 @@
...
+  const _EStaticHelper();
+  int get numberOfTypeArguments => 2;
+  R callWithTypeArgument<R>(int number, R callback<Z>()) {
+    return switch (number) {
+      1 => callback<X>(),
+      2 => callback<Y>(),
+      _ => throw ArgumentError("Expected 1 or 2, got $number."),
+    };
+  }
+}
```
lrhn, line 223: (Maybe this is just a little too abstract. Not sure it carries its weight as an example. Doesn't help selling the feature to me.)

tatumizer:
Not all is bad about this write-up. I found the plot rather engaging at the beginning, but then, without any warning, it turns into something impenetrable and boring at CallWithTypeArguments. If it was a movie, I'd just switch the channel. Which I actually did.
If the goal was to replace the abstract categorical nonsense with something more exhilarating, then the write-up achieves the exact opposite.
There's no story in this story. Not clear who the main character is and how it all ends.
Work-in progress, obviously. Please incorporate the ideas from the thread, they are much more exciting!

mraleph:
> Not all is bad about this write-up. 

@tatumizer please be warned: one more comment in this style and I will ban you from the repo. Please respect language team's time and don't say anything if you have nothing useful to say. Thanks. 

cc @mit-mit @leafpetersen for visibility.

tatumizer:
No worries, I will ban myself. This thing has a considerable opportunity cost for me.

working/4200-metaobjects/feature-specification.md
```diff
@@ -0,0 +1,519 @@
...
+`typeArgs` is a comma separated list of the type variables declared in
+`typeParms` in declaration order, and `args` is a comma separated list of
+identifiers declared as positional parameters in `parms`, followed by
+actual arguments of the form `id: id` for the named parameters in `parms`.
+
+In the case where a default value resolves to a declaration in `C` and does
+not have the prefix `C.`, this prefix is added when the corresponding
+default value is specified in the metaobject class. *For example, 
+`int i = myDefault` in `C` will be `int i = C.myDefault` in the derived
+forwarding declaration when `myDefault` is declared in `C`.*
```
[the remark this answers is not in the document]
eernstg: This is handled in a more abstract way now.

working/4200-metaobjects/feature-specification.md
```diff
@@ -0,0 +1,519 @@
...
+
+If `C(parms) ...` or `factory C(parms) ...` is a constructor declared by `C`
+then the corresponding forwarding declaration is
+`C<X1 .. Xk> call(parms1) => C(args);`, here `args` are derived from `parms1`
```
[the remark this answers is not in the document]
eernstg: Yes, that's a name clash. Note that an `static R call(...) ...` method doesn't make anything callable, which means that it is likely to be a rare exception to have such a declaration.

The forwarding invocation is described in a more abstract way now, which skips over the parameter name clash.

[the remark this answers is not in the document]
eernstg: More abstract now.

working/4200-metaobjects/feature-specification.md
```diff
@@ -0,0 +1,519 @@
...
+in the same way as with the static method above, and `parms1` is derived
+from `parms` by erasing `this.` and `super.` and inserting the inferred
+types and default values that are implied by `this.` and `super.` in the
+cases where those declared types and default values have been omitted.
```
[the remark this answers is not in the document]
eernstg: More abstract now.

working/4200-metaobjects/feature-specification.md
```diff
@@ -0,0 +1,519 @@
...
+
+If `C.name(parms) ...` or `factory C.name(parms) ...` is a constructor
+declared by `C` then the corresponding forwarding declaration is
+`C<X1 .. Xk> name(parms1) => C.name(args);`, here `args` are derived from
```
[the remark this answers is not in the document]
eernstg: More abstract now.

working/4200-metaobjects/feature-specification.md
```diff
@@ -0,0 +1,519 @@
...
+`parms1` in the same way as with the static method above, and `parms1` is
+derived from `parms` in the same way as in the previous case.
+
+### Dynamic Semantics
+
+It is allowed, but not required, for every metaobject to be constant if
+possible. *Hence, programs should not depend on the identity of
+metaobjects.*
```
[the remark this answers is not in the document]
eernstg: Adjusted.

In particular, a type literal is a constant expression if it is a constant type expression, and the value of such expressions are canonicalized.

working/4200-metaobjects/feature-specification.md
```diff
@@ -0,0 +1,519 @@
...
+
+When invoked on an object whose run-time type is `C<T1 .. Tk>`, in the case
+where `C` has a `static implements` or `static extends` clause, the
+implementation of the getter `runtimeType` in `Object` returns an instance
```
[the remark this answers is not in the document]
eernstg: Let's take that discussion in a separate issue.

working/4200-metaobjects/feature-specification.md
```diff
@@ -0,0 +1,519 @@
...
+of the metaobject class for `C` with the same type arguments `T1 .. Tk`.
+
+*For example, `C<int, String>().runtimeType` returns `MetaC<int, String>()`
```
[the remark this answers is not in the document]
eernstg: Reworded.

working/4200-metaobjects/feature-specification.md
```diff
@@ -0,0 +1,519 @@
...
+type is evaluated as an expression yields the corresponding metaobject,
+that is, an instance of the metaobject class for `T`, passing the same
+actual type arguments as the ones which are part of `T`.
+
+*For example, if we are evaluating a type parameter `X` as an expression,
+and the value of `X` is a type `C<S1, S2>` that has a metaobject class
+`MetaC` then the result will be an instance of `MetaC<S1, S2>`.*
+
+The metaobject has standard semantics, everything follows the normal rules
+of Dart based on the metaobject class declaration.
```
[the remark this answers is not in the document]
eernstg: True.

eernstg (commented):
Lots of improvements, PTAL!

The change to a more abstract approach does get a little bit silly when we have to reinvent `covariant` along the way, but it does also make some other things simpler. I think it's a slight improvement on average.

However, it's a major point in the design of this feature that it does _not_ introduce a lot of completely new from-scratch primitive semantics, it is simply a way to transform a type `C<T1 .. Tk>` to an object `MetaC<T1 .. Tk>`, and all the expressive power follows from that, based an well-established mechanisms like superinterfaces, overriding declarations, etcetc.

working/4200-metaobjects/feature-specification.md
```diff
@@ -0,0 +1,675 @@
...
+of static members and constructors are bound to a specific call target at
+compile time, which means that the call site depends on that declaration.
+
+The next example illustrates how we can use metaobjects to call
+constructors in a similar way (yielding 'late-bound constructors'):
+
+```dart
+abstract class SimpleCreation<X> {
+  X call();
+  X named(int _);
```
Levi-Lesches, line 172: I haven't read through the rest of the doc nor the conversation, so sorry if this was already suggested, but

Can there be a special case to allow `new()` here? `call()` is pretty niche and rarely used by common Dart code. I think it would be easier to say "this creates a new object" rather than "this lets you use `()` after the identifier, which would be the default constructor if it was a type". It would also make it clear that if you can do

```dart
final x = X();
```

Then you can also do 
```dart
final x = func(X.new);
```

working/4200-metaobjects/feature-specification.md
```diff
@@ -0,0 +1,675 @@
...
+  final String name;
+  A(this.name);
+  static String prettyPrint(A a) => a.name;
+}
+
+// Ditto.
+class B static implements PrettyPrintable<B> {
+  final int size;
+  B(this.size);
+  static String foo(B b) => "B of size ${b.size}";
```
sgrekhov, line 62: May be `static String prettyPrint(B b) => "B of size ${b.size}";`?

working/4200-metaobjects/feature-specification.md
```diff
@@ -0,0 +1,675 @@
...
+  const E();
+  void foo(X x, Y y) {}
+}
+
+void main() {
+  final E<Object, Object> e = E<String, int>();
+
+  // When we don't know the precise type arguments we can't call
+  // `e.foo` safely. But `CallWithTypeParameters` can help!
+  final Object? eType = e.runtimeType;
```
sgrekhov, line 317: May be `final Type eType = e.runtimeType;`? `eType` is treated as non-nullable type below.

eernstg: Ah, I forgot the promotion. We need to use a supertype of `CallWithTypeParameters` as the declared type in order to be able to promote to that type. I added the `if` statement that provides the promotion. Also, I changed the type to `Object` (both `Object?` and `Object` will work, but the latter avoids the detour around null which seems more economical and straightforward ;-).

working/4200-metaobjects/feature-specification.md
eernstg: Summary: I don't know. `new` looks like a good idea in the case `X.new`, but it doesn't generalize very well.

Levi-Lesches: Interesting edge cases. I don't have anything to add to that as you were very thorough, but thanks for taking a look

eernstg: I added this proposal to the document, just for type variables (so `X.new` means `X.call` and `X.new(42)` means `X.call(42)`). I think that makes sense.

eernstg:
I'll land this now, and we can then take detailed discussions in separate language issues, as usual.

Issue #4304, opened 2025-03-26
Flow analysis around `x is Never`


sgrekhov:
Currently, the following code produces an error in either CFE or the analyzer.
```dart
test(x) {
  late int i;
  if (1 > 2) {
    if (x is Never) {
      i = 42;
    }
  }
  i; // Error: Late variable 'i' without initializer is definitely unassigned.
}
```
That's okay, but I don't see which part of the [flow analysis specification](https://github.com/dart-lang/language/blob/main/resources/type-system/flow-analysis.md) allows this. As far as I understand, according to the current specification, flow analysis detects only `if (true)...`, `if (false)...`, `if (x == null)...`, and `if (x != null)...` branches. Additionally, `x`, when promoted to `Never`, is not read. So it cannot make `true` branch dead.

Please correct me if I'm wrong or update the flow analysis specification to align with the implementation.

cc @stereotype441

stereotype441:
@sgrekhov:
You are right that this is not covered by the current flow analysis specification. In the [expressions](https://github.com/dart-lang/language/blob/main/resources/type-system/flow-analysis.md#expressions) section, the spec says:

> - **instance check** If `N` is an expression of the form `E1 is S` where the static type of `E1` is `T` then:
>   - Let `before(E1) = before(N)`
>   - Let `true(N) = promote(E1, S, after(E1))`
>   - Let `false(N) = promote(E1, factor(T, S), after(E1))`

The behavior that was actually implemented is:

> - **instance check** If `N` is an expression of the form `E1 is S` where the static type of `E1` is `T` then:
>   - Let `before(E1) = before(N)`
>   - If `T` is the type `Never`, then:
>     - Let `true(N) = unreachable(after(E1))`.
>     - Let `false(N) = after(E1)`.
>   - Otherwise:
>     - Let `true(N) = promote(E1, S, after(E1))`
>     - Let `false(N) = promote(E1, factor(T, S), after(E1))`

I will fix this.

Note, however: it's not the case that flow analysis detects only `if (true)...`, `if (false)...`, `if (x == null)...`, and `if (x != null)...` branches. The kinds of expressions that flow analysis considers relevant for branches are those expressions for which it assigns a special `true()` or `false()` model. If you search for `true(` and `false(` in the [expressions](https://github.com/dart-lang/language/blob/main/resources/type-system/flow-analysis.md#expressions) section, you'll see that flow analysis assigns special `true()` and `false()` models for the following kinds of expressions:
- `true`
[...]

PR #4307, opened 2025-03-28
[flow analysis expressions] Fix instance check and assignment to match implementation.

stereotype441:
This change fix three spec inaccuracies:

1. The spec claimed that when flow analysis encounters an expression `N` of the form `V = E` (a local variable assignment), it would set `true(N)` and `false(N)` based on `true(E)` and `false(E)`. Note that we *could* have implemented flow analysis this way; if we had, then a test like `if (v1 = v2 is T)` would have promoted `v2` to `T`. But we didn't implement that functionality.

2. The spec failed to document the special behavior of expressions of the form `E is Never`. Flow analysis knows that expressions of this form cannot evaluate to `true`.

3. The spec failed to document the special behavior of expressions of the form `E is! T`. `E is! T` is understood by flow analysis to behave the same as `!(E is T)`.

Fixes https://github.com/dart-lang/language/issues/4304.

---

<details>

</details>

### DIFF
--- a/resources/type-system/flow-analysis.md
+++ b/resources/type-system/flow-analysis.md
@@ -546,8 +546,6 @@ then they are all assigned the same value as `after(N)`.
   where `x` is a local variable, then:
   - Let `before(E1) = before(N)`.
   - Let `after(N) = assign(x, E1, after(E1))`.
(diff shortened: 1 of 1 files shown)


eernstg (approved):
LGTM!

Issue #4311, opened 2025-04-01
Should constant expressions be potentially constant?


eernstg:
The language specification [implies in commentary](https://github.com/dart-lang/language/blob/4ca7482e127958a9aa54788dbf0bb5672cbc2c4d/specification/dartLangSpec.tex#L8546) that every constant expression is also a potentially constant expression:

> The constant expressions is a subset of the potentially constant expressions

However, this is not expressed in the case of [conditional expressions](https://github.com/dart-lang/language/blob/4ca7482e127958a9aa54788dbf0bb5672cbc2c4d/specification/dartLangSpec.tex#L8813) (`e1 ? e2 : e3`):

```tex
\item An expression of the form \code{$e_1$\,?\,$e_2$\,:\,$e_3$}
  is potentially constant if $e_1$, $e_2$, and $e_3$
  are all potentially constant expressions.
  It is constant if $e_1$ is a constant expression and either
  \begin{enumerate}
  \item $e_1$ evaluates to \TRUE{} and $e_2$ is a constant expression, or
  \item $e_1$ evaluates to \FALSE{} and $e_3$ is a constant expression.
  \end{enumerate}
```

For example, `true ? 1 : v` is a constant expression which is not potentially constant in the following example:

```dart
void main() {
  var v = 0;
  const c = true ? 1 : v; // OK.
}
```

We could add the word 'further' to make it 'It is further constant', like many other cases in the same itemized list, which would cause `true ? 1 : v` to be non-constant because it isn't potentially constant (because `v` isn't a constant variable, and it isn't a formal parameter of a constant constructor).

Alternatively, we could decide that it is not a problem that some constant expressions aren't potentially constant, and we could then adjust the commentary in the language specification.

I'd recommend the former. @dart-lang/language-team, WDYT?

Note that this would be a mildly breaking change. For instance, the CFE follows the specification and does not report the error on line 48 and 52 of [LanguageFeatures/Static-access-shorthand/constant_expression_A10_t17.dart](https://dart-current-results.web.app/#/filter=co19/LanguageFeatures/Static-access-shorthand/constant_expression_A10_t17), but the analyzer does report these errors.

lrhn:
What would a real solution to this problem be. (And while we're there, maybe add more obvious static errors to constant expressions).

### Modes and contexts
[...]

PR #4313, opened 2025-04-02
Adjust the constant expression rules for lists, maps, and conditional expressions

eernstg:
See https://github.com/dart-lang/language/issues/4311#issuecomment-2769391705 for background information.

This PR changes the rule about constant expressions of the form `b ? e1 : e2` such that it is required that every constant expression of this form is also a potentially constant expression. The point is that this maintains the subset relationship "every constant expression is also a potentially constant expression", which is otherwise maintained in all cases. Also, this implies that the checks that are applied to conditional expressions are slightly stronger than they have been so far.

The PR also changes the rule about constant map literals. Before this PR, these constant expressions are underspecified. In particular, there are no constraints at all on the key and value expressions in order to make the map literal as a whole a potentially constant expression, and the expression is constant if it evaluates to an object. In other words, a map literal can contain code with arbitrary side effecting constructs in a key and/or a value expression, and the whole map literal can still be a constant expression. So we definitely need to fix this.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -8692,7 +8692,10 @@ are the following:
   \code{\CONST{} <$K$, $V$>\{$k_1$: $v_1$, \ldots, $k_n$: $v_n$\}}, or
   \code{<$K$, $V$>\{$k_1$: $v_1$, \ldots, $k_n$: $v_n$\}}
   that occurs in a constant context,
-  is a potentially constant expression.
+  is a potentially constant expression
+  if $K$ and $V$ are constant type expressions,
+  and $k_1$, \ldots{} , $k_n$ as well as $v_1$, \ldots{} , $v_n$
+  are constant expressions.
   It is further a constant expression
   if the map literal evaluates to an object.
 
@@ -8813,7 +8816,7 @@ are the following:
 \item An expression of the form \code{$e_1$\,?\,$e_2$\,:\,$e_3$}
   is potentially constant if $e_1$, $e_2$, and $e_3$
   are all potentially constant expressions.
-  It is constant if $e_1$ is a constant expression and either
+  It is further constant if $e_1$ is a constant expression and either
   \begin{enumerate}
   \item $e_1$ evaluates to \TRUE{} and $e_2$ is a constant expression, or
   \item $e_1$ evaluates to \FALSE{} and $e_3$ is a constant expression.


specification/dartLangSpec.tex
```diff
@@ -8692,7 +8692,10 @@ \subsection{Constants}
   \code{\CONST{} <$K$, $V$>\{$k_1$: $v_1$, \ldots, $k_n$: $v_n$\}}, or
   \code{<$K$, $V$>\{$k_1$: $v_1$, \ldots, $k_n$: $v_n$\}}
   that occurs in a constant context,
-  is a potentially constant expression.
+  is a potentially constant expression
+  if $K$ and $V$ are constant type expressions,
+  and $k_1$, \ldots{} , $k_n$ as well as $v_1$, \ldots{} , $v_n$
+  are constant expressions.
```
lrhn, line 8698: This copies the text above, which should is good becuase it looks like a mistake it's missing.
But that text is probably flawed to begin with.
```dart
void main() {
  const v1 = bool.fromEnvironment("maybe") ? "A string" : 42;
  const v2 = v1 is String ? <int>[v1.length] : <int>[v1 as int];
  print(v2.first); // 42
}
```
Assume `v1` has the value 42.
The `v2` initializer is a constant expression if `v1 is String` is a constant expression (it is, and it's `false`) and if `<int>[v1.length]` is a potentially constant expression, and `<int>[0]` is a constant expression.
The `<int>[v1.length]` in a constant context is a potentially constant expression if `int` is constant (&checkmark;) and `v1.length` is a constant expression.
`v1.length` is a constant expression if `v1` is and it evaluates to a `String` value, ... which it doesn't.
Whoops.
So this code is not a valid constant when `v1` is 42, because the expression `v1.length` is not constant, even though it's guarded behind a requirement that `v is String`.

(If we negate the `v1` test to make it `"A string"`, `v2`'s initializer is also not constant because `v1 as int` fails to be constant.)


We cannot make *any* requirements that the non-taken branch (or any part of it) is *constant*, because that means evaluating it. We only required the non-taken branch to be potentially constant, because that's a safe static property.
Except that these collection rules link an expression being *potentially constant* back to a subexpression being *constant*, so we have reintroduced evaluation of the non-taken branch.

And then we can't guard against invalid code, since we're evaluating it even when the guard is `false`. That's not great.

The problem goes deeper than just this text missing. :scream:

specification/dartLangSpec.tex
```diff
@@ -8813,7 +8816,7 @@ \subsection{Constants}
 \item An expression of the form \code{$e_1$\,?\,$e_2$\,:\,$e_3$}
   is potentially constant if $e_1$, $e_2$, and $e_3$
   are all potentially constant expressions.
-  It is constant if $e_1$ is a constant expression and either
+  It is further constant if $e_1$ is a constant expression and either
```
lrhn, line 8819: This change is what runs us into problems with the existing list/set rules, because it makes a requirement of the non-taken branch to *be* potentially constant.
(But we can probably end up requiring that without this change too, just by having the expression in a potentially constant context.)

We really do need a notion of "statically valid (unevaluated) constant expression", because that's what the untaken branch must be.

eernstg: Any such problem would arise with `&&`, `||` and `??` where that same 'further' is already specified. I'd suggest that we make the rules consistent by adding 'further' to the conditional expression.

We may then consider additional changes separately, knowing that all 4 cases would either be able to stay the same (and changes may occur elsewhere, if needed), or they'd presumably need to be changed in essentially the same way.

specification/dartLangSpec.tex
eernstg: > But that text is probably flawed to begin with.

I do agree that we have some opportunities for making the static checks on constant expressions more complete. For instance, we have many situations where a constant constructor declaration is accepted without diagnostics, but no constant invocations of said constructor can be written because every one of them is an error; it would have been nice if we could inform the developer who wrote that constructor that it is unusable.

However, I suspect that there is no simple and clean enhancement which will deal with this issue once and for all, it's more likely to be a tarpit (the constant expression sublanguage shouldn't be Turing complete, so it wouldn't be _exactly_ the same as a Turing tarpit ;-). That is, we would be able to fiddle with this again and again, until we decide that it's enough.

One thing I _would_ like to capture is the fact that a formal parameter of the enclosing constructor declaration should never occur as a subexpression of an expression which is required to be constant (and the case we've discussed here is where this parameter `p` occurs in an unused branch of a conditional expression).

```dart
void main() {
  const v1 = bool.fromEnvironment("maybe") ? "A string" : 42;
  const v2 = v1 is String ? <int>[v1.length] : <int>[v1 as int];
  print(v2.first); // 42
}
```

> this code is not a valid constant when `v1` is `42`

Which is not a problem. The very fact that we support `assert` elements in a constant constructor initializer list shows that it is possible to raise a compile-time error with arbitrary values of parameters that we care to prohibit.

In the special case where we consider a constant variable declaration it is equally possible to prohibit specific values of specific (other) variables: `const myVariable = 1 ~/ (42 - v1);`.

If it's a problem that the declaration of `v2` is an error because `v1.length` in the unused branch _should not disrupt_ the usefulness of the declaration as a whole (in particular, because we aren't using that branch), then we just need to stop using that inconsistent requirement with constant collection literals:

```latex
  A constant list literal (\ref{lists}),
  \code{\CONST{} <$T$>[$e_1$, \ldots, $e_n$]}, or
  \code{<$T$>[$e_1$, \ldots, $e_n$]}
  that occurs in a constant context,
  is a potentially constant expression if $T$ is a constant type expression,
  and $e_1$, \ldots{} , $e_n$ are potentially constant expressions.
  It is further a constant expression
  if $e_1$, \ldots{} , $e_n$ are constant expressions
  and the list literal evaluates to an object.
```

Similarly changes would be applied to the two other kinds of collection literals.

eernstg:
I made the change I suggested, such that collection literals follow the same approach as all other constructs.

specification/dartLangSpec.tex
lrhn: ACK. This change is for consistency. It's not because it's what we want, it's just that it's consistently doing the thing we were already doing, by making the minority cases match the majority (map what set/list does, conditional expression what `??`, `||` and `&&` does).

And as commented in the issue, we do indeed already have the problem, also with list/set literals if they are in a potentially constant context, because then they do need to be potentially constant whether `?`/`:` requires it or not. 
```dart
const dynamic v1 = bool.fromEnvironment("maybe") ? "A string" : 42; // is 42.
class C {
  final List<int> l;
  C() : l = v1 is String ? const <int>[v1.length] : const <int>[v1];
}
```
This is a compile-time error whether `bool.fromEnviornment("maybe")`  is true or false, with the currently specified semantics of `?`/`:` and `const[]`.

(Allthough the rule we add to `?`/`:` here is restrictive. It prevents things that would have worked before. Same for the `const{...}` change. So, technically, this is a breaking change. It goes from more permissive than we really want, to more restrictive than we really want. If people won't complain, then that puts us in a better position to hit the sweet spot later.)

lrhn (approved):
Add entry in changelog saying *why* this change was made.

eernstg:
Just discussed this with @leafpetersen who noticed that the rules are redundant in certain ways: Constant object expressions and constant collection literals already have a set of rules in the sections specific to those constructs, and everything in the 'Constants' section was redundant. So they now just say, essentially, "a constant object expression is a potentially constant and constant expression" and similarly for the other forms.

This preserves the property that the set of constant expressions is a subset of the set of potentially constant expressions.

We can (and probably should) add a rule that if `e` is an expression that contains `v` is a subexpression and `v` denotes a formal parameter of an enclosing constant constructor declaration then `e` is not constant, but this PR still doesn't do that (because that's clearly breaking).

eernstg:
@lrhn, PTAL, the latest changes are non-trivial.

specification/dartLangSpec.tex
```diff
@@ -8653,48 +8659,20 @@ \subsection{Constants}
...
-  a constant context,
-  is a potentially constant expression.
-  It is further a constant expression if the invocation evaluates to an object.
-  % \ref{const} requires each actual argument to be a constant expression,
-  % but here we also catch errors during evaluation, e.g., `C(1, 0)` where
-  % `C(double x, double y): z = x / y;`.
-  It is a compile-time error if a constant object expression is
-  not a constant expression (\ref{const}).
+  A constant object expression (\ref{const})
+  is a potentially constant and constant expression.
```
lrhn, line 8663: This doesn't remove the requirement that subexpressions and type arguments are constant expressions/types, because that's written in the "constant constructor invocation".

Should we say this here, as commentary?

lrhn (approved):
LGTM, ship it!

eernstg:
The updated specification of what it takes for a conditional expression to be constant is the already implemented behavior.

An example: The current implementation of the analyzer as well as the CFE (Dart SDK 3.8.0-250.0.dev and Flutter SDK 3.31.0-1.0.pre.363) reject the following as an error:

```dart
var v = 'World';

void main() {
  const c = 2 > 1 ? 'Hello' : v; // Error!
}
```

This clearly shows that the current wording of the rule about constant conditional expressions is not implemented, the actual implementation is more strict (it requires the unused branch to be a constant expression, and the spec does not).

Checking [visitConditionalExpression](https://github.com/dart-lang/sdk/blob/678c4ac77f204da8e1c43c4d31a3072d953bef20/pkg/analyzer/lib/src/dart/constant/evaluation.dart#L777) of the relevant visitor, we can see that the implementation in the analyzer checks exactly the condition which is added by this PR (that is, the new behavior is already the implemented behavior in the analyzer).

@chloestefantsova found the reason why the same error is reported by the CFE, and it turns out to be implemented in a very early phase of the compilation (in the BodyBuilder) where it is checked that every expression in a constant context is potentially constant.

In both cases, this shows that the tools implement the approach which is specified in this PR (namely: a conditional expression is not constant if it is not potentially constant), and hence there is no implementation effort.

PR #4318, opened 2025-04-03
Add exception about extension members in error about nullable receivers

eernstg:

```dart
extension on int? {
  bool get isEvenOrNull {
    var self = this; // To get promotion.
    return self == null || self.isEven;
  }
}

void main() {
  int? iq = 2 > 1 ? null : 42;
  bool b = iq.isEvenOrNull; // OK.
}
```

As implemented, the analyzer and CFE do not report any errors for `iq.isEvenOrNull`. However, the NNBD feature specification says that

> > It is an error to call a method, setter, getter or operator on an expression whose type is potentially nullable and not `dynamic`, except for the methods, setters, getters, and operators on `Object`.

Arguably, this makes extension member invocations an error when the receiver is potentially nullable, even in the case where the given extension is applicable to a receiver whose type is nullable (as in the example).

I believe the implemented behavior is the desired behavior, and the specification should be adjusted.

This change is non-breaking because the implementations already accept this kind of invocation.

### DIFF
--- a/accepted/2.12/nnbd/feature-specification.md
+++ b/accepted/2.12/nnbd/feature-specification.md
@@ -6,6 +6,10 @@ Status: Draft
 
 ## CHANGELOG
 
+2025.04.03
+  - Add an exception for extension members in the error about nullable
+    receivers.
+
 2021.07.28
   - Allow a constant factory constructor in a class with a late final instance
     variable.
@@ -531,7 +535,7 @@ location outside of the statement.
 
 It is an error to call a method, setter, getter or operator on an expression
 whose type is potentially nullable and not `dynamic`, except for the methods,
-setters, getters, and operators on `Object`.
+setters, getters, and operators on `Object`, and except for extension members.
 
 It is an error to read a field or tear off a method from an expression whose
 type is potentially nullable and not `dynamic`, except for the methods and


accepted/2.12/nnbd/feature-specification.md
```diff
@@ -531,7 +535,7 @@ location outside of the statement.
 
 It is an error to call a method, setter, getter or operator on an expression
 whose type is potentially nullable and not `dynamic`, except for the methods,
-setters, getters, and operators on `Object`.
+setters, getters, and operators on `Object`, and except for extension members.
```
lrhn, line 538: (Does "extension members" include extension type members, because you are allowed to call those on an extension type which doesn't implement `Object`.)

Probably correct and good fix, but still indirect.

What would happen if we remove this paragraph entirely?

I feel like it *should* fall through to some rule that says that you can't call somethin which isn't there, and where there is also no extension member. That is, a member invocation would do a lookup on the static type to see if there is a member. If so, see if it's compatible. If not, check for applicable extension functions. If not one such, it's an error.

If not having this paragraph would err out missing something which says which members `T?` has, then that's what we should write:
> A type of the form `T?` has only the same members as `Object`, with the same signatures.

Then it's just an error to call a member that a type doesn't have, as usual.

lrhn (approved):

accepted/2.12/nnbd/feature-specification.md
eernstg: > A type of the form `T?` has only the same members as `Object`, with the same signatures.

This is exactly the approach I'm using in the proposed null safety update to the language specification (#2605), but it would amount to a bigger change in this document, so I chose this other modification here, because it's minimal.

The extension types are not mentioned in this document because they were only added to the language in version 3.3. Nevertheless, I mentioned them (it will be needed at some point anyway, and this helps remembering it).

PR #4323, opened 2025-04-10
Move language spec upload from Github actions to cloud build: Add cloudbuild.yaml

eernstg:
This PR adds a 'cloudbuild.yaml' specification file to the language repository such that the provision of a PDF of the language specification can be achieved via a Cloud Build.

### DIFF
--- a/.cloud_build/specification/cloudbuild.yaml
+++ b/.cloud_build/specification/cloudbuild.yaml
@@ -1,2 +1,3 @@
 steps:
-- name:  'gcr.io/cloud-builders/docker'
+- name: 'bash'
+  args: ['specification/scripts/build_pdf']
--- a/.cloud_build/specification/cloudbuild.yaml
+++ b/.cloud_build/specification/cloudbuild.yaml
@@ -1,0 +1,2 @@
+steps:
+- name:  'gcr.io/cloud-builders/docker'
--- a/specification/scripts/build_pdf
+++ b/specification/scripts/build_pdf
@@ -1,0 +1,12 @@
+#!/bin/bash --norc
+
+sudo apt-get update -qq
+sudo apt-get install \
+    texlive-latex-base \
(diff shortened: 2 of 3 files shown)


parlough:
I'm not sure if it makes sense, but could the deploy from this repo also use the image created in https://github.com/dart-lang/site-shared/tree/main/cloud_build/firebase-ghcli? We could add the various latex tools needed to that.

eernstg:
> use the image created in https://github.com/dart-lang/site-shared/tree/main/cloud_build/firebase-ghcli?

Sounds good! The image could presumably be extended at any time (like ASAP ;-) and then the script used in this PR could rely on having the LaTeX software available, right?

athomas (approved):

.cloud_build/specification/cloudbuild.yaml
```diff
@@ -7,4 +7,4 @@ options:
 artifacts:
   objects:
     location: 'gs://dart-specification/'
-    paths: ['artifacts/**']
+    paths: ['specification/artifacts/**']
```
eernstg, line 10: Wow, that looks easy!

specification/scripts/build_pdf
```diff
@@ -0,0 +1,17 @@
...
+    texlive-latex-extra \
+    texlive-fonts-recommended \
+    lmodern
+cd specification
+make
+mkdir -p artifacts
+if [ "$_BASE_BRANCH" == "main" ]; then
+  cp dartLangSpec.pdf artifacts/DartLangSpecDraft.pdf
+fi
+mkdir -p artifacts/pr/$_PR_NUMBER
```
athomas, line 16: This should be an else if `$_PR_NUMBER` is set

eernstg:
Perhaps the `-m` option is not accepted, and also not necessary?

PR #4335, opened 2025-04-24
Normalize the language specification whitespace

eernstg:
This PR makes no semantic changes to the language specification, but it changes many details about the whitespace and a couple of other formatting properties.

`\commentary{}` and `\rationale{}` blocks are indented consistently (that is, they indent the body by 2 spaces, which was previously only the case in some locations).

An empty line occurs before `\begin{itemize}` and before `\begin{enumerate}`, consistently.

In parenthesized commentary, the parentheses are consistently shown in commentary font (dimmed). This means that `(\commentary{...})` was changed to `\commentary{(...)}` in a number of locations.

Added/removed a missing/extra space (`...}or ...` --> `...} or ...` and `which  is` --> `which is`, some double spaces after a period or colon).

A non-trivial number of locations may seem to have been modified because a line break was moved (typically because some lines were too long after indentation).


eernstg:
Note that reviewing can be done with a "Hide whitespace" setting which will reduce the number of lines in the diff very substantially.

PR #4339, opened 2025-04-26
Reorganize the sections in the augmentations proposal.

munificent:
The proposal grew organically and most of the semantics were stuffed up in the motivation section.

This tries to lay things out in a more reasonable top-down order. There are no behavioral changes to the proposal and all of the prose is left alone. This is just moving things around.

cc @jakemac53


eernstg (approved):
That was a large number of move operations! LGTM.

PR #4345, opened 2025-05-01
Fix enum code block

mnordine:
Use dart as the code block language, remove extra `}`.


munificent:
I'm sorry I didn't catch this PR earlier, but #4357 removed this example entirely. I really appreciate the fix, though. :)

Issue #4356, opened 2025-05-06
How do you augment an enum without adding values?


munificent:
We allow an augmentation of an enum type to add new values and/or members ("enhanced enums"). The grammar works fine if the augmentation adds both:

```dart
enum E { a }

augment enum E {
  b;

  method() {}
}
```

But the `enumType` grammar currently requires at least one value declaration. What if you want to augment an enum to add members but no values? What should that look like?

The obvious answer is:

```dart
augment enum E {
  method() {}
}
```

But I'm not 100% certain that's unambiguous. If it is ambiguous, we could probably require the `;`:

```dart
augment enum E {
  ;
  method() {}
}
```

But it would be nice to avoid that if possible. Filing this issue so we don't forget to figure this out.

munificent:
> It is ambiguous.
> 
> ```dart
> augment enum Foo {
>   name();
> }
> ```

Ah, good catch!

> However, that leaves us with the same problem for the original non-augmenting declaration. If you can add elements in an augmentation, you can also want to start out without any elements in the first declaration.

Yes, definitely. I suspect that this is how most enums that have augmented values will be declared.

> I think requiring a leading `;` whenever you declare any non-element members is fine. That is already the rule, you can omit `;` if there are no members after it, so it's more of a "members start here" separator than an elements list terminator.

SGTM.

This does raise the natural question of whether it's allowed to declare an enum with zero values that *isn't* later augmented by having more values added. Probably not *useful* but I guess it's a thing we could technically allow. Should we?

> (Consider dropping the `augment` keyword, there should be no real difference between the original declaration and an augmenting declaration anyway, especially if the latter declares no augmenting members.)

I do like having the `augment` modifier. If a user accidentally has two top level declarations with the same name, I'd rather report an error than silently treat one of them as an augmentation which could lead to other very confusing errors if the two declarations aren't the same kind, or have colliding members or any other kind of weird overlap.

eernstg:
Closing: There will be a `;` even in the case where no new values are declared.

Issue #4080, opened 2024-09-04
Dart Implicit Static Access


lrhn:
An unqualified identifier is looked up in the lexical scope. If that fails to find a binding, the identifier <code>*id*</code> is then treat as an instance member access on `this`, as if it was <code>this.*id*</code>. 

This only applies to instance member access. If the surrounding class has a static member with the same name, then it is always in the lexical scope. Until augmentations.

With augmentations, one `augment class C { … }` may want to reference a static member declared by *another* `augment class C { … }`.  As currently specified, only the declarations inside the current augmenting class declarations are in its *lexical* scope, so the former class would have to prefix a static access with `C.`.  Which is annoying.

This is a proposal to do a second level of scope lookup after the initial lexical scope lookup has failed to find anything. The lexical scope is based on the individual declarations, and the second scope tried is the combined scope of the *fully augmented* surrounding definitions instead.

This is intended to be introduced with, or just be part of, the augmentations specification.

## Proposal

To resolve an identifier *id*, do the following:

* Look up *id* in the surrounding lexical scope. If anything is found, then that's either an error or what the identifier denotes. _(An error can occur for example if a getter-lookup finding only a setter, or similar.)_
* Otherwise if the identifier occurs **inside** a `class`, `mixin`, `enum`, `extension` or `extension type` declaration _(the declarations that can have instance and static members)_, and the *fully augmented* definition for that declaration contains a static or instance declaration (not constructor) with base name *id*, then the identifier occurrence denotes a static or instance access to that member, just as if it had been found in the surrounding declaration in the lexical scope. _So equivalent to either <code>NameOfClass.*id*</code> or <code>this.*id*</code>, with possible down-stream errors if the only member found with that base name was a setter, or if it was an instance member but *id* occurred where `this` was not available._
* Otherwise the identifier denotes an instance member lookup on the `this` object at its static type. *(There may not be any member with that name in the interface of the static type of `this`, in which case the name may be resolved to an extension member. Or it may just be an error because there is no `this` in the current scope.)*

The "Inside" of such a name-space declaration is defined as:

* Anything inside the *body* of the declaration:

   * Anywhere in a member declaration of the namespace declaration.
   * Anywhere in an enum element declaration.

  If it's inside the `{…}`, it counts.

[...]

PR #4357, opened 2025-05-06
Simplify augmentations.

munificent:
- Remove references to macros.
- Don't allow augmentations to wrap or replace code. Remove support for `augmented` expressions. Disallow an augmentation from providing a body to a declaration that already has one.
- Remove support for augmenting variables.
- Simplify constructor augmentations: no concatenating initializers or merging initializers from one augmentation and a body from another.
- Remove support for augmenting typedefs.
- Remove support for augmenting redirecting constructors.
- Allow a function augmentation to have an `external` body.

There are still more changes I'd like to make. In particular, the grammar for members seems pretty hairy and I suspect could be refactored to be simpler, but I'm not sure if that will cause problems for other people who rely on the existing grammar rules.


working/augmentation-libraries/feature-specification.md
```diff
@@ -622,311 +626,97 @@ It is a compile-time error if:
...
-  first,
-  second.custom(1);
+Variable declarations themselves can't be directly augmented, with the exception
+of abstract fields.
 
-  final int b;
+*A variable declaration implicitly has code for the synthesized getter and
+setter that access and modify the underlying backing storage. Since we don't
+allow augmentations to replace code, that implies that augmentations can't
+change variables. So we don't allow them to be augmented.*
```
lrhn, line 656: That makes sense, except for the lack of ability to provided an initializer expression. But then just make it an abstract variable, and have the augmentation add an actual variable and getter/settter forwarding.

And the `augment abstract ... var` is really just a way to augment with abstract setters and getters.

working/augmentation-libraries/feature-specification.md
```diff
@@ -622,311 +626,97 @@ It is a compile-time error if:
...
-part of 'a.dart';
+  // Augment a getter with an abstract variable:
+  int get c;
 
-augment enum A {
-  fourth;
+  @someMetadata
+  augment abstract final int c; // (Not very useful, but valid.)
 }
+```
```
lrhn, line 687: Might need to say how type inheritance work if the getter and setter have different types, and if the declaration has more than one variable:
```dart
int get x;
set x(num _);
abstract String y;
@markOfShame
augment abstract x, y;
```
(Maybe it's not a problem, we use the same inheritance rules as we would for an override in a subclass ... which don't handle this either;
```dart
abstract class C {
  int get x;
  set x(num _);
  abstract String y;
}

abstract class D implements C {
  abstract var x, y;
}
```
makes `x` an error bacause it cannot inherit a combined signature.)

working/augmentation-libraries/feature-specification.md
```diff
@@ -622,311 +626,97 @@ It is a compile-time error if:
...
-  // invocation.
-  augment third.custom(3);
-}
-```
+An augmentation of an enum type can add new members to the enum, including new
+enum values. Enum values are appended in augmentation application order.
 
-Then `A.values` is `[A.first, A.second, A.third, A.fourth, A.fifth]`.
+Enum values themselves can't be augmented since they are essentially constant
+variables and variables can't be augmented.
```
lrhn, line 695: _(But the static getters can, even if only to add metadata.)_

working/augmentation-libraries/feature-specification.md
```diff
@@ -622,311 +626,97 @@ It is a compile-time error if:
...
+It's a **compile-time error** if:
 
-*   An augmenting getter is defined for an enum value. _An enum value
-    counts as a constant variable._
-*   An enum value augmentation provides an explicit constructor invocation.
+*   A declaration inside an augmenting enum declaration has the name `values`,
+    `index`, `hashCode`, or `==`. *It has always been an error for an enum
+    declaration to declare a member named `index`, `hashCode`, `==`, or
+    `values`, and this rule just clarifies that this error is applicable for
+    augmenting declarations as well.*
```
lrhn, line 703: (We should be lifting that rule up to the level of "augmentation chains" or "complete declarations" and say that the *collected* declarations must not contain these members.)

working/augmentation-libraries/feature-specification.md
```diff
@@ -622,311 +626,97 @@ It is a compile-time error if:
...
-**Non-redirecting factory constructors**: These are much like static methods,
-and might return a subtype of the current type. They also may not create a new
-instance, but return an existing one. They must have a body.
-**Redirecting constructors**: Both generative and factory constructors can be
-redirecting, although the syntax looks slightly different for each.
+*We could support augmenting redirecting constructors, but the use cases seem
+rare. Supporting this would require a way to specify that a constructor is
+redirecting without actually providing the redirection since the augmentation
+needs to fill that in and can't replace an existing redirection in the
+introductory declaration.*
```
lrhn, line 717: A redirecting constructor without a rediction is just, like, a constructor.

I think it should be possible to specify a *signature* of any augmentable declaration, such that the *implementation* can be added by an augmentation.

Whether a generative constructor is initializing or redirecting is an implementation choice.
So, you should be able to declare:
```dart
const Point(int x, int y, {Color? color = Color.black});
```
and have it count as a generative constructor, but not yet a redirecting or initializing one.
Then an augmentation can choose either implementation strategy.

For factory constructors, you should also be able to specify a factory constructor signature:
```dart
factory Point(int x, int y, {Color color = Color.black}); 
```
and then decide whether to implmenet that is forwarding ( `= ...`) or computing (`{ ... }`) in an augmentation.
(Here we run into an issue, because augmentations cannot specify default values, but forwarding constructors also cannot specify default values. That's a problem we made for ourselves, so I believe we can fix it.)

working/augmentation-libraries/feature-specification.md
```diff
@@ -938,79 +728,14 @@ It is a compile-time error if:
...
-constructor with the same target constructor designation as the augmenting
-constructor. This removes the potentially non-redirecting property of the
-constructor.
-
-It is a compile-time error if:
-
-*   The augmented factory constructor has a body, or it is redirecting.
+    constructor is not, or vice versa. *An augmentation can't change whether or
+    not a constructor is generative because that affects whether users are
+    allowed to call the constructor in a subclass's initializer list.*
```
lrhn, line 738: (That only argues for why *adding* `factory` is bad, not why removing it is.)

working/augmentation-libraries/feature-specification.md
```diff
@@ -1040,136 +765,28 @@ extension type A(int b) {
...
+[primary constructors]:
+https://github.com/dart-lang/language/blob/main/working/2364%20-%20primary%20constructors/feature-specification.md
+
+The extension type's representation object is *not* a variable, even though it
+looks and behaves much like one, and it cannot be augmented as such.
+
+It's a **compile-time error** if:
 
-It is a compile time error if:
+*   An augmenting declaration has the same name as the representation object.
```
lrhn, line 776: representation object -> representation variable

(The representation object is the value of the representation variable, only the latter has a *name*.)

working/augmentation-libraries/feature-specification.md
```diff
@@ -1040,136 +765,28 @@ extension type A(int b) {
...
-augmenting type declaration _Dt_. A compile-time error occurs if the
-identifier `augmented` occurs in _Dm_.
-
-*For example, inside `augment class C` we could have a declaration like
-`void f() {...augmented()...}`. This is an error because the outer
-`augment` forces the meaning of `augmented` to be about augmentation in the
-entire scope, but the method declaration is introductory and hence there is
-no earlier declaration to augment.*
+If an augmentation has metadata attached to it, these are appended to the
+metadata of the declaration being augmented.
```
lrhn, line 783: I'd say:
> An augmenting declaration can have metadata attached to it.
> Interpretation of metadata is not part of the language,
> and is left to the tools which define the annotations and which
> inspect the source to find them.
> _Some metadata may apply to the combined declaration,
> others to the individual augmenting declarations.
> Only the tool which defined the annotation knows what it means._

Then we can say (somewhere non-normative) that `dart:mirrors` will only work in the combined declarations, and should see the combined annotations, if it works at all.

lrhn (approved):
More shorter, more better!
I have a gazillion nitpicks and some questions (which might suggest that something could be said to answer that question, or which might suggest that I only read the changed parts and therefore lacked some context.).
And some comments that are better left for later.

Good clean-up!

working/augmentation-libraries/feature-specification.md
```diff
@@ -1040,136 +765,28 @@ extension type A(int b) {
...
-runtime if no implementation was in fact provided.
-
-**NOTE**: Macros should _not_ be able to statically tell if an external body has
-been filled in by a compiler, because it could lead to a different result on
-different platforms or tools.
-
-**TODO: Should we add a syntax to let the augmentation dynamically detect
-whether there is an external implementation to call?**
-
-### Augmenting with metadata annotations and doc comments
```
rrousselGit, line 1070: Why is this removed? This is fairly important

working/augmentation-libraries/feature-specification.md
```diff
@@ -206,129 +164,132 @@ memberedDeclarationBody ::= '{' memberDeclarations '}'
...
-  | constructorSignature (redirection | initializers)?
+declaration ::= 'external'? factoryConstructorSignature ';'
+  | 'augment'? 'external' constantConstructorSignature ';'
+  | 'augment'? 'external' constructorSignature ';'
+  | 'augment'? 'external'? 'static'? getterSignature ';'
+  | 'augment'? 'external'? 'static'? setterSignature ';'
+  | 'augment'? 'external'? 'static'? functionSignature ';'
+  | 'external' ('static'? finalVarOrType | 'covariant' varOrType) identifierList ';'
+  | 'augment'? 'external'? operatorSignature ';'
+  | 'augment'? 'abstract' (finalVarOrType | 'covariant' varOrType) identifierList ';'
```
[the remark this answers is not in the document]
munificent: > So no augmenting with a concrete variable declaration?

Yeah, my pitch is that we just don't support augmenting variable declarations or augmenting with variable declarations at all. If you have an abstract getter and you want to augment it with a backing store, you can always augment with a getter that returns the value of a differently named field.

> In which case, what's the point of augmenting a variable declaration at all, other than to add metadata. I guess that is a reason.

It's my intent to disallow it completely. The only variable-ish clause that still allows `augment` is abstract variables, which aren't really variables at all.

> (I'm also not completely sure what `declaration` covers or where it's used. It doesn't seem to include functions.)

The grammar around this seems really messy to me. I *think* all it does is avoid some repeated `;` in clauses but that doesn't seem that useful to me.

I'm really tempted to refactor this entire grammar around members and declarations but I don't know if that will cause any problems because of other specs/tools/tests/whatever depending on specific grammar rules.

working/augmentation-libraries/feature-specification.md
```diff
@@ -206,129 +164,132 @@ memberedDeclarationBody ::= '{' memberDeclarations '}'
...
-    remove for functions, so change to "A variable declaration". A macro should
-    be able to implement a method as an external with a `@JS()` annotation.)**
-
 ## Static semantics
 
 ### Declaration ordering relations
 
 As part of the meta-programming and augmentation features, we expand the
-capabilities of part files. See ["Parts with Imports"][parts_with_imports.md].
+[capabilities of part files][parts with imports]. With that feature, a part file
```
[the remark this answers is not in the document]
munificent: The file is still named parts_with_imports.md. The title in the doc is "Parts with imports" but then it says it defines "enhanced part files". I'll just leave this as-is until I sort that out in a later change. :)

working/augmentation-libraries/feature-specification.md
```diff
@@ -442,56 +403,120 @@ variable type, parameter type, or type parameter bound. In the last case,
...
-type as the type that the augmented definition has at the corresponding
-position. __An augmenting declaration can omit type annotations, but if it
-doesn't, it must repeat the type from the augmented definition__.
+If the type annotation or type parameter bound is omitted in the augmenting
+declaration, it is inferred to be the same as the corresponding type annotation
+or type parameter bound in the declaration being augmented.
+
+If the type annotation or type parameter bound is *not* omitted, then it's a
+**compile-time error** if the type denoted by the augmenting declaration is not
+the same type as the type in the corresponding declaration being augmented.
```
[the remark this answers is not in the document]
munificent: This is a good question, and definitely a question for @eernstg. :)

I lifted this verbiage straight out of the proposal and moved it here so I think it was written this way deliberately.

working/augmentation-libraries/feature-specification.md
```diff
@@ -442,56 +403,120 @@ variable type, parameter type, or type parameter bound. In the last case,
...
+More precisely, a function or constructor declaration (introductory or
+augmenting) is *incomplete* if all of:
+
+*   The body syntax is `;`.
+
+*   The function is not marked `external`. *An `external` function is considered
+    to have a body, just not one that is visible as Dart code.*
+
+*   There is no initializer list. *Obviously, this only applies to constructor
+    declarations.*
```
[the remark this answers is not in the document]
munificent: Ooh, good catch!

working/augmentation-libraries/feature-specification.md
```diff
@@ -504,59 +529,46 @@ augment class SomeClass {
...
-It is a **compile-time error** if:
+*   A library contains two top-level declarations with the same name, and one of
+    the declarations is a class-like declaration and the other is not of the
+    same kind, meaning that either one is a class, mixin, enum, extension or
+    extension type declaration, and the other is not the same kind of
+    declaration.
 
 *   The augmenting declaration and augmented declaration do not have all the
     same modifiers: `abstract`, `base`, `final`, `interface`, `sealed` and
     `mixin` for `class` declarations, and `base` for `mixin` declarations.
```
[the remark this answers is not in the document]
munificent: Right. We could theoretically allow it, but it seems very unlikely to be useful and almost certainly confusing for readers.

working/augmentation-libraries/feature-specification.md
```diff
@@ -206,129 +164,132 @@ memberedDeclarationBody ::= '{' memberDeclarations '}'
...
+  | 'static' 'late'? varOrType initializedIdentifierList ';'
+  | 'covariant' 'late' 'final' type? identifierList ';'
+  | 'covariant' 'late'? varOrType initializedIdentifierList ';'
+  | 'late'? 'final' type? initializedIdentifierList ';'
+  | 'late'? varOrType initializedIdentifierList ';'
+  | redirectingFactoryConstructorSignature ';'
+  | constantConstructorSignature (redirection | initializers)? ';'
+  | 'augment' constantConstructorSignature initializers? ';'
+  | constructorSignature (redirection | initializers)? ';'
+  | 'augment' constructorSignature initializers? ';'
```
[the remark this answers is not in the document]
munificent: I'm going to leave this alone for now (unless the grammar in the proposal is actually incorrect), but I would like to try to do some grammar refactoring in a later change if doing so won't be too disruptive.

working/augmentation-libraries/feature-specification.md
```diff
@@ -504,59 +529,46 @@ augment class SomeClass {
...
-with the same name, and:
+It's a **compile-time** error if:
 
-*   Neither is an augmenting declaration, or
-*   one of the declarations is a class-like declarations and the other is
-    not of the same kind, meaning that at either one is a class, mixin,
-    enum, extension or extension type declaration, and the other is not the
-    same kind of declaration.
+*   The resulting clauses after being appended would be erroneous if declared
+    directly. This means you can't end up with multiple `extends` clauses on a
```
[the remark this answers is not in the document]
munificent: > * It's a compile-time error if an augmenting class declaration has an extends clause, and any prior declaration for the same class also has an extends clause.

I like this. Done.

working/augmentation-libraries/feature-specification.md
```diff
@@ -504,59 +529,46 @@ augment class SomeClass {
...
-    declaration shows the complete capabilities of the declaration. It also
-    deliberately prevents an augmentation from introducing a restriction that
-    isn't visible to a reader of the main declaration.*
-
-*   The augmenting declaration declares an `extends` clause for a `class`
-    declaration, but one was already present _(or the `class` was a `mixin
-    class` declaration, which does not allow `extends` clauses)_.
+    *This is not a technical requirement, but follows our design principle that
+    what is known from reading the introductory declaration will still be true
+    after augmentation.*
```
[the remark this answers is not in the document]
munificent: > Adding `mixin` to a `class` with no superclass an no or only trivial constructors, retains all capabilities that were known before.

Fair, but I still think it's a simpler model to just say you keep all the modifiers. We could loosen this somewhat if compelling use cases arise, but it's hard to imagine someone wanting an augmentation to turn a `class` into a `mixin class`.

munificent:
cc @johnniwinther @scheglov in case you have thoughts on how this would affect the implementation.

working/augmentation-libraries/feature-specification.md
```diff
@@ -206,129 +164,132 @@ memberedDeclarationBody ::= '{' memberDeclarations '}'
...
+  | 'static' 'final' type? initializedIdentifierList ';'
+  | 'static' 'late' 'final' type? initializedIdentifierList ';'
+  | 'static' 'late'? varOrType initializedIdentifierList ';'
+  | 'covariant' 'late' 'final' type? identifierList ';'
+  | 'covariant' 'late'? varOrType initializedIdentifierList ';'
+  | 'late'? 'final' type? initializedIdentifierList ';'
+  | 'late'? varOrType initializedIdentifierList ';'
+  | redirectingFactoryConstructorSignature ';'
+  | constantConstructorSignature (redirection | initializers)? ';'
+  | 'augment' constantConstructorSignature initializers? ';'
```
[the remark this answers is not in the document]
munificent: That was a deliberate choice, yes. By disallowing augmenting redirecting constructors, we avoid all of the complexity around "potentially redirecting" and "potentially non-redirection" which felt really sketchy to me in the proposal.

We could allow augmenting redirecting constructors, but if we do, I'd want some sort of syntax in the introductory declaration that says "this is definitely a redirecting constructor even though the redirection isn't here yet".

working/augmentation-libraries/feature-specification.md
```diff
@@ -504,59 +529,46 @@ augment class SomeClass {
...
-
-*   The augmenting declaration declares an `extends` clause for a `class`
-    declaration, but one was already present _(or the `class` was a `mixin
-    class` declaration, which does not allow `extends` clauses)_.
+    *This is not a technical requirement, but follows our design principle that
+    what is known from reading the introductory declaration will still be true
+    after augmentation.*
 
 *   An augmenting extension declares an `on` clause *(this is a syntax
     error)*. We also do not allow adding further restrictions to a `mixin`
```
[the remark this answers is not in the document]
munificent: Yeah, this whole error is a little weird. We don't allow an augmenting extension type declaration to contain a representation field, and it's specified by having the grammar directly prohibit it. We don't allow an augmenting mixin declaration to contain an `on` clause and it's *not* specified in the grammar and is instead just made a compile error here. Also, this compile error redundantly points out the thing about extension types.

I'm not sure why these two restrictions were specified differently. I suspect it's because the `on` clause is already optional in a mixin declaration but the representation field is *mandatory* in an introductory extension type declaration and forbidden in an augmenting one.

I went back and specified both in terms of the grammar and then added a note pointing out the syntax change. Then I removed this error because it's now entirely covered by the grammar.

working/augmentation-libraries/feature-specification.md
```diff
@@ -504,59 +529,46 @@ augment class SomeClass {
 }
 ```
 
-This means that instead of creating a new declaration, the augmentation
-modifies a corresponding declaration in the library (which is 'before' this
-one).
-
 A class, enum, extension type, mixin, or mixin class augmentation may
 specify `extends`, `implements` and `with` clauses (when generally
 supported). The types in these clauses are appended to the introductory
```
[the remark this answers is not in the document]
munificent: Changed the grammar to forbid it.

working/augmentation-libraries/feature-specification.md
```diff
@@ -622,311 +626,97 @@ It is a compile-time error if:
...
-
-Enum values can _only_ be augmented by enum values, and the implicit getter
-introduced by them is not augmentable. The only thing you are allowed to do
-when augmenting an enum value is add metadata annotations or doc comments.
+### Augmenting variables
 
-When augmenting an enum value, no constructor invocation should be provided.
-The original value is always used, and the explicit constructor invocation (if
-present) should not be copied.
+A class-like augmentation can add *new* variables (i.e. instance or static
```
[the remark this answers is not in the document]
munificent: I like "that is".

working/augmentation-libraries/feature-specification.md
```diff
@@ -622,311 +626,97 @@ It is a compile-time error if:
...
+fields) to the type being augmented:
```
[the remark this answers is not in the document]
munificent: I like the formal part being not the parenthetical. The parenthetical is basically explaining "wait, what do you mean by 'variables' inside a class?".

working/augmentation-libraries/feature-specification.md
```diff
@@ -622,311 +626,97 @@ It is a compile-time error if:
...
 ```
-// main.dart
-part 'a.dart';
-part 'c.dart';
 
-enum A {
-  first,
-  second.custom(1);
+Variable declarations themselves can't be directly augmented, with the exception
+of abstract fields.
```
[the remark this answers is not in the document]
munificent: Personally, I'm not a fan of conflating an *abstract* declaration (to be filled in by a polymorphic override or implementation) and an *incomplete* declaration (to be filled in by a later augmentation).

The proposal still does that because it seems like `;` is the most natural syntax for an incomplete declaration, but I'd really hesitate to double down on that and use `abstract static` for incomplete static fields that are not abstract at all.

(Though if we go down the path of metaclasses and do something like [`class variables` in Swift](https://www.hackingwithswift.com/example-code/language/whats-the-difference-between-a-static-variable-and-a-class-variable) then maybe `abstract static` starts to make more sense.)

working/augmentation-libraries/feature-specification.md
munificent: > But then just make it an abstract variable, and have the augmentation add an actual variable and getter/settter forwarding.

Right. We don't *need* to support augmenting a getter/setter with a variable because you can accomplish the same thing with a little indirection. So I figured I'd keep it simple and get rid of augmenting variables completely.

working/augmentation-libraries/feature-specification.md
munificent: Good catch. Added a section under type annotation inheritance to talk about this.

working/augmentation-libraries/feature-specification.md
munificent: It was redundant. The proposal says elsewhere that this works.

working/augmentation-libraries/feature-specification.md
munificent: In this PR, the pitch is that variables can't be augmented *at all*, even by an abstract getter that only adds metadata to the variable. We could allow this, but then it needs to be restricted such that the augmenting getter or setter has to be abstract (incomplete). If you augment a variable with a setter, does the metadata get copied to the variable's implicit getter too?

We could work this out, but I suspect it starts to get special-case-heavy and I don't think it's likely to be very valuable since all you could do is add metadata. So I'm inclined to just leave it out entirely. Variables are *overridable* in the sense of polymorphism and uniform access and all that. But when it comes to augmentations, they're fixed. We could always loosen this in the future if we get a use case.

working/augmentation-libraries/feature-specification.md
munificent: Does the distinction matter? For the collected declarations to contain one of these members, either the introductory declaration declares one (already an error) or one of the augmentations does (an error here).

Also, in general we state that if the result of applying augmentations leads to a combined declaration with errors then those errors are reported. (For example, two augmentations adding members with the same name and conflicting signatures.)

working/augmentation-libraries/feature-specification.md
munificent: Oh good point. I thought there was something weird about redirections, but I guess not. A redirecting generative constructive is still generative in all the ways that count, so it's harmless to allow completing a generative constructor with a redirection. Likewise for factories.

Done.

working/augmentation-libraries/feature-specification.md
munificent: An augmentation can't really *remove* anything. If we did allow:

```dart
class C {
  factory C();
}

augment class C {
  augment C() { return ...; }
}
```

Then users would probably reasonably assume that the resulting class took the union of the modifiers like it does with `implements`, `with`, etc.

working/augmentation-libraries/feature-specification.md
munificent: Good catch. I don't particularly care about mirrors and suspect most users don't either. But I wrote a bit of non-normative stuff about the analyzer package.

working/augmentation-libraries/feature-specification.md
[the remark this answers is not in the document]
munificent: Based on your later comment, I went ahead and added support for filling in an incomplete constructor (generative or factory) with a redirection.

working/augmentation-libraries/feature-specification.md
```diff
@@ -442,56 +414,142 @@ variable type, parameter type, or type parameter bound. In the last case,
...
+first one must not be marked `augment` and all the subsequent ones must be.*
 
 An augmentation declaration does not introduce a new name into the surrounding
 scope. *We could say that it attaches itself to the existing name.*
 
-Making `augment` a built-in identifier is language versioned, to make it
-non-breaking for pre-feature code.
+### Complete and incomplete declarations
+
+Augmentations aren't allowed to *replace* code, so they mostly add entirely new
```
johnniwinther, line 475: In terms of implementation complexity, it is fairly easy to support this. Actually the ability to add to the extends/with/implements clauses of a class is at a similar if not higher complexity.

working/augmentation-libraries/feature-specification.md
```diff
@@ -442,56 +414,142 @@ variable type, parameter type, or type parameter bound. In the last case,
...
+}
+
+augment class C {
+  augment C() { body; } // Error. C() is already complete.
+}
+```
 
 ### Application order
 
 The same declaration can be augmented multiple times by separate augmentation
```
johnniwinther, line 535: The ability to do this seems somewhat useless if we can't augment complete methods with complete augmentations (and with that invoke the augmented method). In terms of language complexity we seem to spend a look specification work here for no benefit to the user.

working/augmentation-libraries/feature-specification.md
```diff
@@ -344,13 +316,13 @@ is not in the lexical scope of a class declaration._
...
-available. _This approach is always potentially dangerous, since any
+type, it can use `this.name` for instance members and `TypeName.name` for
+static members to be explicit. Or it can rely on the default if
+`name` is not in the lexical scope at all, in which case it's interpreted
+as `this.name` if it occurs inside a scope where a `this` is available. _This
+approach is always potentially dangerous, since any
 third-party import adding a declaration with the same name would break the
-code. In practice that’s almost never a problem, because instance members
+code. In practice that's almost never a problem, because instance members
 and top-level declarations usually use different naming strategies._
```
eernstg, line 326: Didn't we decide at a language meeting that a membered declaration (class, mixin, mixin class, enum, extension type, extension) with the name `N` should have a flat semantic body scope containing all declarations (introductory plus the augmenting) named `N` in the same declaration context?

This would be consistent with the treatment of the top-level scope (also, and perhaps in particular, in the case where the prefix scope is separated from the library scope): It's always a flat namespace, global to the entire Dart file tree.

One thing that falls out is that there is no need to separately explain why we can't declare two unrelated members named `foo` when they are in two different enclosing declarations (we can't rely on the general "two declarations in the same scope with the same name is an error" rule if it isn't one scope).

working/augmentation-libraries/feature-specification.md
```diff
@@ -442,56 +414,142 @@ variable type, parameter type, or type parameter bound. In the last case,
...
+context, order them according to the "after" relation, and each augments the
+result of all the prior augmentations applied to the original declaration . The
+first one must not be marked `augment` and all the subsequent ones must be.*
 
 An augmentation declaration does not introduce a new name into the surrounding
 scope. *We could say that it attaches itself to the existing name.*
 
-Making `augment` a built-in identifier is language versioned, to make it
-non-breaking for pre-feature code.
+### Complete and incomplete declarations
```
eernstg, line 473: This simplification feels really, really good!

working/augmentation-libraries/feature-specification.md
```diff
@@ -504,59 +562,49 @@ augment class SomeClass {
...
+previously, then it is added with the new types.
+
+*Example:*
+
+```dart
+class C implements I1 with M1 {}
+augment class C implements I2 with M2 {}
+
+// Is equivalent to:
+class C implements I1, I2 with M1, M2 {}
```
eernstg, line 578:
```suggestion
class C with M1, M2 implements I1, I2 {}
```

working/augmentation-libraries/feature-specification.md
```diff
@@ -622,311 +662,89 @@ It is a compile-time error if:
...
 
-  final int b;
-
-  const A() : b = 0;
-
-  const A.custom(this.b);
-}
-}
+Variable declarations themselves can't be augmented (by variables, getters, or
+setters), with the exception of abstract instance variable declarations.
```
eernstg, line 687: I guess an abstract `augment` getter could be acceptable, and it could be used to add metadata and DartDoc?

working/augmentation-libraries/feature-specification.md
```diff
@@ -622,311 +662,89 @@ It is a compile-time error if:
...
 
-augment enum A {
-  third;
+#### Abstract instance variables
 
-  /// Some doc comment
-  augment first; // This is still `first` in values.
+Dart supports `abstract` field declarations. They are syntactic sugar for
+declaring an abstract getter with an optional abstract setter if the variable is
+non-final. They don't actually declare a variable with any backing storage.
```
eernstg, line 698: The word 'optional' is confusing here. Perhaps:
```suggestion
Dart supports `abstract` field declarations. They are syntactic sugar for
declaring an abstract getter and, if not `final`, an abstract setter.
They don't actually declare a variable with any backing storage.
```

working/augmentation-libraries/feature-specification.md
```diff
@@ -938,79 +756,31 @@ It is a compile-time error if:
     *Default values are defined solely by the introductory constructor.*
```
eernstg, line 756: I can't comment on line 751, but 'same type,' should probably be 'same type (when provided),'.

working/augmentation-libraries/feature-specification.md
```diff
@@ -1040,136 +810,40 @@ extension type A(int b) {
...
 
 ## Dynamic semantics
 
 The application of augmentation declarations to an augmented declaration
 produces something that looks and behaves like a single declaration: It has
-a single name, a single type or function signature, and it’s what all
+a single name, a single type or function signature, and it's what all
 references to the *name* refers to inside and outside of the library.
 
 Unlike before, that single *semantic declaration* now consists of multiple
```
eernstg, line 849: I can't comment on line 896, but I think it would be a good idea to replace non-augmenting by introductory, for consistency.

eernstg: Line 928 ends in "the *C* doe". That sentence needs to be revised and completed. Also, the occurrence of `abstract` (in code font) in a description of an "abstract method" is confusing: The method is abstract, but it can't contain the keyword `abstract`.

working/augmentation-libraries/feature-specification.md
```diff
@@ -1286,13 +960,13 @@ For example, we define the *augmented parameter list* of a non-empty stack,
...
-the parameter list of the introductory declaration, but it’s more
+the parameter list of the introductory declaration, but it's more
 direct and clearly correct to use the actual parameter list of the declaration
 when creating the parameter scope that its body will run in._
 
 Similarly we define the _augmented function type_ of the declaration stack.
 Because of the restrictions we place on augmentations, they will all have the
-same function type as the introductory declaration, but again it’s
+same function type as the introductory declaration, but again it's
 simpler to assign a function type to every declaration.
```
eernstg, line 970: Presumably, this step should gather metadata as well?

working/augmentation-libraries/feature-specification.md
```diff
@@ -1286,13 +960,13 @@ For example, we define the *augmented parameter list* of a non-empty stack,
...
 
 #### Invocation
```
eernstg, line 972: All references to 'augmented' in this section should be removed, and the rules simplified accordingly.

eernstg (approved):
LGTM, this is great! In particular, I appreciate the amount of regularization and simplification which is achieved by omitting some well-chosen elements.

Added a bunch of comments, of course. In particular, I think we should split out the discussions about the scoping (and possibly handle them during updates of the 'Part files with imports' feature spec).

working/augmentation-libraries/feature-specification.md
```diff
@@ -442,56 +414,142 @@ variable type, parameter type, or type parameter bound. In the last case,
...
+Augmentations aren't allowed to *replace* code, so they mostly add entirely new
+declarations to the surrounding type. However, function and constructor
+augmentations can fill in a body for an augmented declaration that is lacks one.
+
+More precisely, a function or constructor declaration (introductory or
+augmenting) is *incomplete* if all of:
+
+*   The body syntax is `;`.
+
+*   The function is not marked `external`. *An `external` function is considered
```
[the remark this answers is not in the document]
munificent: Good point. Filed https://github.com/dart-lang/language/issues/4363 to discuss this further. For now in this PR, I'll leave the behavior as it is which is that `;` is overloaded to mean both abstract and incomplete.

working/augmentation-libraries/feature-specification.md
munificent: It's definitely not *high* benefit. But there are at least two places where augmentation order is user visible:

* Adding mixins (`with` clause) to a class.
* Adding values to an enum.

In most cases, users won't care about or be able to witness augmentation application order, but in a couple of cases they can, so it's gotta be pinned down. Sort of like how instance field initializer order has to be specified because you can see it if the instance fields have initializers with side effects. (Though certainly the specification of that is much simpler than what we have here...)

We could probably say that an augmentation can't add values to an enum unless it has no values (which we also need to add support for #4356). But I don't think we can get away with saying that an augmentation can only add a `with` clause if the type doesn't have one. It would be *really* annoying to not be able to use a mixin on your class just because some code generator happens to be adding an unrelated mixin to your type as well.

working/augmentation-libraries/feature-specification.md
```diff
@@ -206,129 +167,140 @@ memberedDeclarationBody ::= '{' memberDeclarations '}'
...
-*Some declarations do not match any of these cases (e.g., a local variable
-declaration in a method body), but this does not matter: We never need to
-talk about the context of a local variable.*
+Prior to this proposal, an entity like a class or function is introduced by a
+single syntactic declaration. With augmentations, an entity may be composed out
+of multiple declarations, the introductory one and any number of augmentations.
+We define a notion of a *context* to help us talk about the location where we
+need to look to collect all of the declarations that define some entity.
+
+* The context of a top-level declaration is the library and its associated tree
```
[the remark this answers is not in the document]
munificent: Yeah, I didn't like that term either but it was already in the proposal. I like "augmentation context". (And, having gone through the proposal to make that change, I did find a couple of other unrelated uses of "context", so you're right about it being too general.)

working/augmentation-libraries/feature-specification.md
```diff
@@ -206,129 +167,140 @@ memberedDeclarationBody ::= '{' memberDeclarations '}'
...
+  of part files.
+
+* The context of a member declaration in a type declaration named *N* is the set
```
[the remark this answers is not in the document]
munificent: I don't know if "membered" really clarifies much and would then have to be defined. I changed it to "inside a declaration". I think "member" itself should be pretty unambiguous about what it applies to. Let me know if you think that doesn't work.

working/augmentation-libraries/feature-specification.md
```diff
@@ -206,129 +167,140 @@ memberedDeclarationBody ::= '{' memberDeclarations '}'
...
+  of type declarations (introductory and augmenting) named *N* in the enclosing
+  set of Dart files.
+
+*Note that context is only defined for the kinds of declarations that can be
+augmented. We don't define a context for, say, local variable declarations,
+because those aren't subject to augmentation.*
 
 ### Scoping
```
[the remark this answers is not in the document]
munificent: I don't think we can make this section fully non-normative because "parts with imports" (correctly) doesn't talk about how identifiers inside type or extension declarations with augmentations are resolved.

This section is honestly kind of a mess. I rewrote it (largely by copying from the original version of this section, ironically).

working/augmentation-libraries/feature-specification.md
munificent: > Didn't we decide at a language meeting that a membered declaration (class, mixin, mixin class, enum, extension type, extension) with the name `N` should have a flat semantic body scope containing all declarations (introductory plus the augmenting) named `N` in the same declaration context?

You know I honestly don't remember, but digging through the notes, I found a reference to #4080. It seems that, yes, once identifier resolution reaches a membered declaration body, the scope it sees there is the combined scope with all augmentations applied. That's definitely what I want.

Rewrote the "Scoping" section to specify this.

working/augmentation-libraries/feature-specification.md
munificent: Heh, oops! Changed locally.

working/augmentation-libraries/feature-specification.md
munificent: I considered that, but it gets a little weird. If you do:

```dart
int x;

@someMeta
augment int get x;
```

What does the metadata get added *to*? Metadata is a syntactic construct. It doesn't have semantics at all, so there isn't really an "implicit getter" that it makes sense to add it to. Do we add it to the variable itself? Is that even useful?

I couldn't think of a concrete use case for this so I figured I'd just say that variables are entirely closed to augmentation. We can always loosen this in the future if a compelling use case arises, but that seems unlikely.

working/augmentation-libraries/feature-specification.md
munificent: Ooh, much better. Applied locally.

working/augmentation-libraries/feature-specification.md
munificent: Good catch. Done.

working/augmentation-libraries/feature-specification.md
munificent: I'll be honest, I'm not even really sure what this "Dynamic Semantics" section is all about. As far as I'm concerned, all we need to do is specified how augmentations are applied and then the resulting declaration's dynamic semantics are exactly as if you had written everything in one declaration (modulo scoping across part files, of course).

This looks like @lrhn's writing style. (@kallentu and I were coincidentally discussing today how we can tell who wrote each part of a spec 😄 ). @lrhn, do you know what this is supposed to be completed with? Do we need this section at all?

working/augmentation-libraries/feature-specification.md
munificent: @lrhn pointed out in his review that metadata doesn't actually have semantics in the language. So I added a section basically saying it's up to tools to decide how metadata is handled but suggesting that it's probably most useful to append it.

working/augmentation-libraries/feature-specification.md
munificent: Done, though again I'm still not really sure what this section is for. :)

working/augmentation-libraries/feature-specification.md
```diff
@@ -206,129 +167,140 @@ memberedDeclarationBody ::= '{' memberDeclarations '}'
...
+  | 'late'? 'final' type? initializedIdentifierList ';'
+  | 'late'? varOrType initializedIdentifierList ';'
+  | 'augment'? redirectingFactoryConstructorSignature ';'
+  | 'augment'? constantConstructorSignature (redirection | initializers)? ';'
+  | 'augment'? constructorSignature (redirection | initializers)? ';'
 ```
 
-It is a compile-time error if:
+*Note that the grammar for putting `augment` before an extension type
+declaration doesn't allow also specifying a representation field. This is by
```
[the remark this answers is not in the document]
munificent: Haha, I had to manually diff the grammar in here versus the spec to see what actually changes it had because I didn't notice it either.

working/augmentation-libraries/feature-specification.md
munificent: That's good to know. I'm still somewhat interested in supporting `augmented()`, but given that we have other more pressing issues, my plan is to leave it out, get this proposal cleaned up and locked down, and move on. We can always revisit augmentations and make them more powerful in the future when we see how users are using them.

munificent:
I'm going to go ahead and land this, but I plan to keep making tweaks, so feel free to comment here or file issues if any of you see other things. Thanks for the excellent reviews!

Issue #4367, opened 2025-05-09
flow-analysis.md treats the order of `promotedTypes` inconsistently.


stereotype441:
Reading through https://github.com/dart-lang/language/blob/main/resources/type-system/flow-analysis.md this morning, I realized that it treats the order of the `promotedTypes` stack inconsistently:

In the [Models](https://github.com/dart-lang/language/blob/main/resources/type-system/flow-analysis.md#models) section, `promotedTypes` is defined as an "ordered set" of types, with "the _final_ entry in the ordered set being the current promoted type of the variable" (emphasis added), and says that "each entry in the ordered set must be a subtype of all previous entries, and of the declared type". This implies that the entries in the list will be ordered from least-promoted to most-promoted.

In the [Promotion](https://github.com/dart-lang/language/blob/main/resources/type-system/flow-analysis.md#promotion) section:
- The "current type" of a variable is defined to be the _first_ element of the `promotedTypes` list (or the declared type if `promotedTypes` is empty).
- The `assign` and `promote` operations are defined so that promotion to type `T` changes `promotedTypes` to `T::promotedTypes`, and so that demotion changes `promotedTypes` to a suffix of `promotedTypes`; this implies that the entries in the list will be ordered from most-promoted to least-promoted.

I propose to change the spec to uniformly order `promotedTypes` from least-promoted to most-promoted, since this matches the order used by the implementation.

PR #4369, opened 2025-05-09
[flow analysis] Clean up promoted types ordering and use of `::`

stereotype441:
The `promotedTypes` list in a `VariableModel` is now consistently ordered from least promoted type to most promoted type, matching the list used in the implementation.

Also, the use of `top::rest` to refer to a stack is changed to `[...rest, top]`. This is consistent with the way some stacks are represented in the flow analysis implementation (as a list where pushing and popping happens at the end of the list).

Fixes https://github.com/dart-lang/language/issues/4367.

### DIFF
--- a/resources/type-system/flow-analysis.md
+++ b/resources/type-system/flow-analysis.md
@@ -108,8 +108,8 @@ that assignment).
 - Lists
   - We use the notation `[a, b]` to denote a list containing elements `a` and
     `b`.
-  - We use the notation `l::a` where `l` is a list to denote a list beginning
-    with all the elements of `l` and followed by `a`.
+  - We use the notation `[...l, a]` where `l` is a list to denote a list
+    beginning with all the elements of `l` and followed by `a`.
 
 - Stacks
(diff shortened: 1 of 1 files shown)


stereotype441:
> LGTM, considering that the most important thing is to be consistent.
> 
> I would mention, though, that `h::t` almost inevitably means a list where `h` is an element which is at the head (first) in the list, and `t` is the tail (the rest) of the list, and `t::h` is going to look really weird to a functional programmer.
> 
> Perhaps use another notation? One notation that would look normal for functional programmers could be `t ++ [h]`, where `++` is a list concatenation operator. That would allow us to connect to actual Dart code like `tail + [head]`, and it would also handle `tail ++ [n1, n0]` in a way which is reasonably concise.

That's a good point. I've switched to a notation that resembles Dart list syntax: `[...tail, head]` and `[...tail, n1, n0]`.

I'm going to go ahead and land the change in this form based on your previous lgtm. If you have concerns, let me know and I can always do a follow up PR.

PR #4374, opened 2025-05-14
Create Static Enough Metaprogramming proposal

mraleph:
This moves content from https://github.com/dart-lang/language/issues/4271 into a markdown file in the repository to make discussion and revisions easier.

I have incorporated some of the feedback from discussions on the issue - but I continue to maintain focus on this as a toolchain feature. I have added some remarks that analyzer can't constant fold everything anyway because it does not have access to the compilation environment.

I would like to collect a few rounds of feedback and then rejuvenate the prototype implementation to get something experimental working across all platforms in the SDK so that we can have an idea of how well this could work in a real world.

### DIFF
--- a/working/4271 - static enough metaprogramming/proposal.md
+++ b/working/4271 - static enough metaprogramming/proposal.md
@@ -1,15 +1,23 @@
-tldr: I propose we follow the lead of **D**, **Zig** and **C++26** when it comes
-to metaprogramming. We introduce an optional (toolchain) feature to force
-compile time execution of certain constructs. We add library functions to
-introspect program structure which are _required_ to execute in compile time
-toolchain supports that. These two together should give enough expressive power
-to solve a wide range of problems where metaprogramming is currently wanted. cc
-@dart-lang/language-team
+# Static Enough Metaprogramming
 
-# History of `dart:mirrors`
+**Author**: Slava Egorov (@mraleph, vegorov@google.com)
+
+**Status**: Draft
+
+**Version**: 1.0
+
+**Summary**: I propose we follow the lead of **D**, **Zig** and **C++26** when it
+comes to metaprogramming. We introduce an optional (toolchain) feature to force
+compile-time execution of certain constructs. We add library functions to
+introspect program structure which are _required_ to execute at compile time
+if the toolchain supports that. These two together should give enough expressive
+power to solve a wide range of problems where metaprogramming is currently
+wanted.
+
+## History of `dart:mirrors`
 
 In the first days of 2017 I have written a blog post ["The fear of
-`dart:mirrors`"][the-fear-of-dartmirrors] which contained started with the
+`dart:mirrors`"][the-fear-of-dartmirrors] which started with the
 following paragraph:
 
 > [`dart:mirrors`][dart-mirrors-1-21-1] might be the most misunderstood,
@@ -18,12 +26,12 @@ following paragraph:
 > fog of uncertainty and marked as _Status: Unstable_ in the documentation -
 > even though APIs have not changed for a very long time.
 
-In 2017 type system was still optional, AOT was a glorified
+In 2017 the type system was still optional, AOT was a glorified
 _"ahead-off-time-JIT"_, and the team maintained at least 3 different Dart
-front-ends (VM, dart2js and analyzer). Things really started shifting with Dart
-2 release: it had replaced optional types with a static type system and
-introduced _common front-end (CFE)_ infrastructure to be shared by all backends.
-Dart 3 introduced null-safety by default (NNBD).
+front-ends (VM, dart2js and analyzer). Things really started shifting with the
+Dart 2 release: it had replaced optional types with a static type system and
+introduced a _common front-end (CFE)_ infrastructure to be shared by all
+backends. Dart 3 introduced null-safety by default (NNBD).
 
 And so 8 years and many stable releases later Dart language and its toolchains
 have changed in major ways, but [`dart:mirrors`][dart-mirrors] remained in the
@@ -33,21 +41,21 @@ on Dart, only in JIT mode and only outside of Flutter.
 How did this happen?
 
 The root of the answer lies in the conflict between _Dart 1 design philosophy_
-and necessity to use _AOT-compilation_ for deployment.
+and the necessity to use _AOT-compilation_ for deployment.
 
 Dart 1 was all in on dynamic typing. You don't know what a variable contains -
-but you can do anything with it. Can pass it anywhere. Can all any methods on
+but you can do anything with it. Can pass it anywhere. Can call any methods on
 it. Any class can override a catch-all `noSuchMethod` and intercept invocations
 of methods it does not define. Dart's reflection system `dart:mirrors` is
 similarly unrestricted: you can [reflect][dart-mirrors-reflect] on any value,
-then ask information about its [type][dart-mirrors-type], ask type about its
+then ask information about its [type][dart-mirrors-type], ask a type about its
 [declarations][dart-mirrors-declarations], ask declared members about their
 [parameters][dart-mirrors-parameters] and so on. Having an appropriate _mirror_
 you can invoke methods, read and write fields, instantiate new objects.
 
 This ability to _indirectly_ act on the state of the program creates a problem
-for static analysis of the program and that in turn affects ability of the AOT
-compiler to produce a small and fast binary.
+for the static analysis of the program and that in turn affects the ability
+of the AOT compiler to produce a small and fast binary.
 
 To put this complexity in simple terms consider two pieces of code:
 
@@ -61,34 +69,34 @@ InstanceMirror m; List args; Symbol name;
 m.invoke(name, args);
 ```
 
-When compiler sees the first piece of code it can easily figure out which
+When a compiler sees the first piece of code it can easily figure out which
 `method` implementation this call can reach and what kind of parameters are
-passed through. With the second piece of code analysis complexity skyrockets -
+passed through. With the second piece of code, analysis complexity skyrockets -
 none of the information is directly available in the source code: to know
-anything about the invocation compiler needs to know a lot about contents of
+anything about the invocation a compiler needs to know a lot about contents of
 `m`, `args` and `name`.
 
 While it is not impossible to built static analysis which is capable to see
-through the reflective access - in practice such analyses are complicated, slow
+through reflective access - in practice such analyses are complicated, slow
 and suffer from precision issues on real world code.
 
-AOT compilation and reflection is pulling into opposite directions: AOT compiler
-wants to know which parts of the program are accessed and how, while reflection
-obscures this information and provides developer with indirect access to the
-whole program. When trying to resolve this conflict you can choose between three
-options:
+AOT compilation and reflection is pulling into opposite directions: the AOT
+compiler wants to know which parts of the program are accessed and how, while
+reflection obscures this information and provides developer with indirect
+access to the whole program. When trying to resolve this conflict you can choose
+between three options:
 
 - The first option is to **make reflection system _just work_ even after AOT
   compilation**. This means retaining all information (and code) which can be
   indirectly accessed via reflection. In practice this means retaining _most_ of
   the original program - because most uses of reflection are notoriously hard to
   analyze statically.
-- The second option is to **allow AOT-compiler to ignore reflection uses it
+- The second option is to **allow an AOT-compiler to ignore reflection uses it
   can't analyze** and providing developer with a way to feed additional
   information about reflective uses into the compiler. If developer forgets (or
   feeds incomplete or incorrect information) reflective code _might_ break after
-  compilation if compiler decides to remove part of the program which it deems
-  unreachable.
+  compilation if the compiler decides to remove part of the program which it
+  deems unreachable.
 - The third option is to capitulate and **disable reflection APIs in AOT
   compiled code**.
 
@@ -96,7 +104,7 @@ Facing this choice is not unique to Dart: Java faces exactly the same challenge.
 On one hand, the package [`java.lang.reflect`][java-reflect] provides indirect
 APIs for accessing and modifying the state and structure of the running program.
 On the other hand, developers want to obfuscate and shrink their apps before
-deployment. Java ecosystem went with the second option: shrinking tools
+deployment. The Java ecosystem went with the second option: shrinking tools
 more-or-less ignore reflection and developers have to [manually inform
 toolchain][android-shrink-code] about the program elements which are accessed
 reflectively.
@@ -109,11 +117,11 @@ reflectively.
 >
 > - [Reflection Analysis for Java][paper-java-suif-reflection]
 > - [Understanding and Analyzing Java Reflection][paper-java-reflection-2019]
-> - [Challenges for Static Analysis of Java Reflection – Literature Review and Empirical Study](paper-java-reflection-challenges)
+> - [Challenges for Static Analysis of Java Reflection – Literature Review and Empirical Study][paper-java-reflection-challenges]
 >
 > Graal VM Native Image (AOT compiler for Java) attempts to fold away as much of
 > reflection uses as it can, but otherwise just like ProGuard and similar tools
-> [relies][java-native-image-reflection] on the developer to inform compiler
+> [relies][java-native-image-reflection] on a developer to inform the compiler
 > about reflection uses it could not resolve statically.
 >
 > R8 (Android bytecode shrinker) has a special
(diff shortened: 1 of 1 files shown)


working/4271 - static enough metaprogramming/proposal.md
```diff
@@ -0,0 +1,1222 @@
+tldr: I propose we follow the lead of **D**, **Zig** and **C++26** when it comes
+to metaprogramming. We introduce an optional (toolchain) feature to force
+compile time execution of certain constructs. We add library functions to
+introspect program structure which are _required_ to execute in compile time
+toolchain supports that. These two together should give enough expressive power
+to solve a wide range of problems where metaprogramming is currently wanted. cc
+@dart-lang/language-team
```
lrhn, line 7: No cc needed here.

working/4271 - static enough metaprogramming/proposal.md
```diff
@@ -0,0 +1,1222 @@
+tldr: I propose we follow the lead of **D**, **Zig** and **C++26** when it comes
```
lrhn, line 1: I'd give this file a header.

- A title line at the top.
- An author line: `Author: @mralpeh`.
- Consider a version line: `version: 1.0` and a version section/changelog at the end, to track changes.

lrhn (approved):

eernstg (approved):
LGTM. (Very interesting indeed!)

I could suggest the addition of 'a' or 'the' in many locations, but decided that this wouldn't be the top priority at this time.

working/4271 - static enough metaprogramming/proposal.md
```diff
@@ -0,0 +1,1222 @@
...
+of relying on the toolchain to provide unrestricted reflection, have developer
+opt-in into specific reflective capabilities for specific parts of the program
+then generate a pile of auxiliary Dart code implementing these capabilities.
+
+> [!NOTE]
+>
+> Another exploration similar in nature was
+> (go/const-tree-shakeable-reflection-objects)[http://go/const-tree-shakeable-reflection-objects].
+
+Fundamentally both of these approaches were dead ends and Web applications
```
jakemac53, line 153: Fwiw, I wouldn't say that the const reflection proposal was a dead end at all, we just simply didn't end up doing it but there wasn't any sort of blocker or anything that I remember encountering.

working/4271 - static enough metaprogramming/proposal.md
```diff
@@ -0,0 +1,1222 @@
...
+> This was one of the showstoppers for enhanced constant proposal, which did not
+> introduce a syntactic marker to delineate code potentially needed for
+> `const`-evaluation from the rest of the program, which would mean that
+> outlines had to contain bodies for most methods - making them impractically
+> large and erasing their benefits.
+>
+> This proposal does not suffer the same issue, even if we decide to support
+> `@konst` evaluation in these tools: as `@konst` actually provides a syntactic
+> marker which partitions the program. Only bodies of methods with `@konst`
+> parameters need to be included into the outline.
```
jakemac53, line 957: Can these functions only invoke other `@konst` functions?

working/4271 - static enough metaprogramming/proposal.md
```diff
@@ -0,0 +1,1222 @@
...
+
+### Non-Goals
+
+This proposal does _not_ intend to cover all cases which are supportable via
+Dart source code generation or which are supported by macro systems in other
+programming languages. For example the following capabilities are out of scope:
+
+- injecting new declarations into the program;
```
jakemac53, line 965: This definitely does drastically simplify the proposal - but it also makes it a lot less useful. Ultimately trying to accomplish this was the downfall of the original macros proposal though.

But, I think it is important to look at the original motivations for macros, and which of those use cases are covered by this.

Ultimately, I think you can end up with some fairly decent solutions for automatic encoding/decoding, equality, toString, etc. But not copyWith or constructors due to the signatures being dependent on the shape of the class (you could generate the bodies of these but that's only half the boilerplate).

working/4271 - static enough metaprogramming/proposal.md
```diff
@@ -0,0 +1,1222 @@
...
+> It is however possible to define method like this:
+>
+> ```dart
+> // Universal function to apply updates specified by the record.
+> T copyWith<@konst T, @konst R extends Record>(T obj, R updates) {
+>   // ...
+> }
+> ```
+>
+> Though usability of this method will be questionable as there will be good
```
jakemac53, line 984:
```suggestion
> Though usability of this method will be questionable as there will be no good
```

working/4271 - static enough metaprogramming/proposal.md
```diff
@@ -0,0 +1,1222 @@
...
+To get the feeling of expressive power, implementation complexity and costs I
+have thrown together a very rough prototype implementation which can be found
+[here][konst-prototype-branch]. When comparing manual toJSON implementation with
+a similar (but not equivalent!) one based on `@konst` reflection I got the
+following numbers:
+
+- Average code size overhead per class: 270 bytes
+- Average JIT kernel generation overhead per-class: 0.2ms (cost of producing
+  specialized functions using Kernel-to-Kernel AST transformation)
+- Average AOT compilation overhead per-class: 2.6ms
```
jakemac53, line 998: What is the baseline?

working/4271 - static enough metaprogramming/proposal.md
```diff
@@ -0,0 +1,1222 @@
...
+
+### Example: Defining `hashCode` and `==`
+
+We could also instruct compiler to handle `mixin`'s (and possibly all generic
+classes) with `@konst` type parameters in a special way: _clone_ their
+declarations with known type arguments. This would allow to write the following
+code:
+
+```dart
+mixin DataClass<@konst T> {
```
jakemac53, line 1134: This would be 🔥

working/4271 - static enough metaprogramming/proposal.md
```diff
@@ -0,0 +1,1222 @@
...
+  @konst external bool get isStatic;
+
+  @konst external TypeInfo<FieldType> get type;
+
+  /// Get the value of this field from the given object.
+  @konst external FieldType getFrom(HostType value);
+}
+```
+
+Note that all methods are annotated with `@konst` so if compiler supports
```
kevmoo, line 865:
```suggestion
Note that all methods are annotated with `@konst` so if the compiler supports
```

working/4271 - static enough metaprogramming/proposal.md
mraleph: Yeah, I can see this happening. I also realized that what I am suggesting does not actually work: consider a simple class like this:

```dart
class Foo {
  final int v;
  const Foo(this.v);

  Foo inc(int u) => Foo(this.v + u);
}
```

There is nothing preventing us from evaluating something like `Foo(1).inc(2)` as constant - but we can't really mark `inc` as requiring constant inputs, because it does not require that. Not everything that can be constant evaluted should always be constant evaluted... Hence you can't use `@konst` as a marker.

Revised the text incorporating your suggestion that we probably should partition imports instead.

working/4271 - static enough metaprogramming/proposal.md
mraleph: I don't have numbers right now but I can try to revive the prototype and update this later.

working/4271 - static enough metaprogramming/proposal.md
mraleph: Yeah, it would be useful in many places. Also it would solve a performance problem we occasionally encounter where inheritance or generics cause polymorphism and severely reduce performance.

working/4271 - static enough metaprogramming/proposal.md
mraleph: I have added a section trying to cover some of the use cases mentioned peaked from an old spreadsheet I managed to find. 

I could not exactly figure out what some of those entail, could you help me a bit (look for `TODO: unclear what this means.` and if you have context please add it in the comment). I am interested about _auto listenable_ and _render accessors_. I think _union types (ala freezed)_ means more concise way of declaring sealed class hierarchies - which I think should just be its own feature (paired with primary constructors).

I think a number of use cases require _property wrappers_ and/or ability to redirect methods - this should probably be its own feature. If property wrappers are available then `@konst` can be used to handle the boilerplate.

Finally some cases (e.g. proxies, functional widgets and reduced widget classes boilerplate) require ability to create class declarations. I think it is actually an okay feature to give via `@konst` reflection. As long as classes are anonymous and are otherwise not visible.

mraleph:
I have made some updates based on comments. PTAL.

munificent (approved):
I still have the same reservations as before, but landing this in the repo so we can discuss more sounds great. It's a very solid well-thought out proposal and the background information it provides is really helpful for any metaprogramming-related features we might do.

working/4271 - static enough metaprogramming/proposal.md
```diff
@@ -0,0 +1,1430 @@
...
+following paragraph:
+
+> [`dart:mirrors`][dart-mirrors-1-21-1] might be the most misunderstood,
+> mistreated, neglected component of the Dart's core libraries. It has been a
+> part of Dart language since the very beginning and is still surrounded by the
+> fog of uncertainty and marked as _Status: Unstable_ in the documentation -
+> even though APIs have not changed for a very long time.
+
+In 2017 the type system was still optional, AOT was a glorified
+_"ahead-off-time-JIT"_, and the team maintained at least 3 different Dart
```
osa1: Is this a typo for "ahead-of-time-JIT"? (off -> of)

working/4271 - static enough metaprogramming/proposal.md
```diff
@@ -0,0 +1,1430 @@
...
+> [!NOTE]
+>
+> Another exploration similar in nature was
+> [go/const-tree-shakeable-reflection-objects](http://go/const-tree-shakeable-reflection-objects).
+
+Fundamentally both `smoke` and `reflectable` were dead ends and Web applications
+written in Dart solved their code size and performance issues by moving away
+from reflection to _code generation_, effectively abandoning runtime
+metaprogramming in favor of build time metaprogramming. Code generators are
+usually written on top of Dart's [analyzer][pkg-analyzer] package: they inspect
```
osa1, line 165: Reading this I was curious how the analyzer can be used for this. It would be helpful to have a link to a code generator written using `analyzer`.

working/4271 - static enough metaprogramming/proposal.md
```diff
@@ -0,0 +1,1430 @@
...
+> code by cloning and specializing functions with `@konst`-parameters _or_ it
+> could choose to retain reflective metadata and forego cloning at the cost of
+> runtime performance. This reduces the size of deployed applications but
+> decreases peak performance.
+
+There are multiple reasons for going this route. First of all, I believe that
+this allows to considerably simplify the design and implementation of this
+feature. But there is another reason for this: _you can't actually perform
+compile time evaluation without knowing compile time environment_. Consider for
+example the following code:
```
osa1, line 918: > I believe that this allows to considerably simplify the design and implementation of this feature

I think this is true for this language feature, but is it also true for the language and tooling in general? For example, AFAIU `invoke` would be `Function.apply` in runtime. This will require that some (maybe all?) of the dynamically typed programming features like `Function.apply` will have to stay, and they'll be an essential part of the language rather than old features we want to get rid of. (because they are slow and require metadata that adds to binary sizes)

working/4271 - static enough metaprogramming/proposal.md
```diff
@@ -0,0 +1,1430 @@
...
+  @konst external TypeInfo<Base>? instantiationOf<@konst Base>();
+
+  /// Return type-arguments of `T` if any.
+  @konst external List<TypeInfo> get typeArguments;
+
+  /// Return `T` default constructor.
+  @konst external Function defaultConstructor;
+
+  /// Return the list of fields in `T`.
+  @konst external List<FieldInfo<T, Object?>> get fields;
```
osa1, line 862: Shouldn't the `Object?` here be `TypeInfo<Object?>` (or `TypeInfo`, `TypeInfo<dynamic>`)? Can the field types be anything else other than a `TypeInfo`?

osa1 (commented):
I'm confused about how `@konst` applies to statements and functions. Given this:

> When `@konst` is applied to loop iteration variables it instructs the compiler to expand the loop at compile time by first computing the sequence of values for that iteration variable, then cloning the body for each value in order and substituting iteration variable with the corresponding constant.

(in particular, "expand the loop at compile time" suggests code generation based on constant values)

I would expect this function

```
void bar<@konst T>(@konst T v) {

}
```

to be monomorphised based on `T`s passed. (and generate one more copy for non-const `T`)

Otherwise when I do things like this code in the JSON examples:

```
Map<String, Object?> toJson<@konst T>(T value) => {
    for (@konst final field in TypeInfo.of<T>().fields)
      if (!field.isStatic) field.name: field.getFrom(value),
  };
```

The loop cannot be unfold at compile time, because the unfolded loop needs to be different for each `T`.

But the text doesn't mention monomorphisation at all. Am I confused?

If this allows monomorphisation (and it has to, if I get it right), then that's a pretty big capability for the langauge that's also worth mentioning.

osa1:
I also find it confusing to see what function calls (or other syntax) is part of the generated code vs. compile-time evaluation. For example, if I change the `toJson` example from the proposal to this:

```
Map<String, Object?> toJson<@konst T>(T value) => {
    for (@konst final field in TypeInfo.of<T>().fields)
      if (!field.isStatic) processFieldName(field.name): processFieldValue(field.getFrom(value)),
  };
```

(`processFieldName` and `processFieldValue` are new)

Just by looking at this code (without seeing `processFieldName` and `processFieldValue` definitions) I can't tell whether these function calls are evaluated in compile time or runtime.

I wonder if we could make stages explicit. There are many languages that do this that we could look for inspiration. For example Common Lisp's [backquote and comma][1], [MetaOCaml][2]'s (`.< expr >.`) and escape (`.~expr`) make the stages explicit.

[1]: https://www.lispworks.com/documentation/HyperSpec/Body/02_df.htm
[2]: https://okmij.org/ftp/ML/MetaOCaml.html

working/4271 - static enough metaprogramming/proposal.md
sigurdm: It will only need to exist on those platforms that do not implement compile-time konst folding.

You'll live in one of two worlds:

* jit (dev), retains full source, uses something like Function.apply to evaluate konst expressions on-demand
* aot (release), evaluates konst at compile time. No further reflection data is retained

osa1:
Something else that I just realized is that if you evaluate some code in runtime in some cases and compile time in others, you can get runtime error in some cases and compile-time error in others.

For example, consider this silly code: (modified from the `operator ==` example in the proposal)

```dart
mixin DataClass<@konst T> {
  @override
  blah(Object? other) {
    final typeInfo = TypeInfo.of<T>();
    for (@konst final field in typeInfo.fields) {
      final value2 = field.getFrom(other as T);
      ...
    }
    return true;
  }
}
```

Here the line `other as T` will fail then `other`'s type is not right, but when the failure happens will depend on when you evaluate `other as T`.

Does the proposal address this? (sorry if I can't find it..)

working/4271 - static enough metaprogramming/proposal.md
mraleph: Fixed. Thanks.

working/4271 - static enough metaprogramming/proposal.md
mraleph: Added a link to https://pub.dev/packages/json_serializable

working/4271 - static enough metaprogramming/proposal.md
mraleph: You mean `FieldInfo<T, TypeInfo<Object?>>`? That would not work for the intended purpose, we would like use these type-arguments to ensure proper typing of access operations.

```dart
final class FieldInfo<HostType, FieldType> {
  @konst external FieldType getFrom(HostType o);
  @konst external void setOn(HostType o, FieldType fieldValue);
}
```

If you have field `String x` in `class C` its `FieldInfo` will have reified type `FieldInfo<C, String>`, which guarantees type safety of `fi.getFrom(...)` and `fi.setOn(...)`.

working/4271 - static enough metaprogramming/proposal.md
mraleph: Yep, what @sigurdm says: the idea is to permit these highly dynamic features only in environments that can tolerate them (e.g. JIT), while in environments that can't (AOT based ones) we only allow these features if they are specialized/monomorphised by the compiler.

mraleph:
@osa1 

> But the text doesn't mention monomorphisation at all. Am I confused?
> 
> If this allows monomorphisation (and it has to, if I get it right), then that's a pretty big capability for the langauge that's also worth mentioning.

There is specifically this paragraph which talks about it though it does not explicitly call it _monomorphisation_. 

> When `@konst` is applied to parameters (including type parameters) it turns
> functions into _templates_: compiler will require that annotated parameter is a
> constant known at compile time and clone the function for a specific combination
> of parameters. The original function is removed from the program: it is
> impossible to invoke it dynamically or tear it off. To annotate `this` as
> `@konst` developer will need to place `@konst` on the declaration of the
> function itself.

> I also find it confusing to see what function calls (or other syntax) is part of the generated code vs. compile-time evaluation.

That's fair, though its somewhat intentional: 

1. to avoid any syntactic changes altogether;
2. to keep code runnable as-is in environments which choose to not implement this feature;

I am open for suggestions on how to make this better if you have any ideas. I think it would need some syntactic designator, but I don't immediately see anything that would be very readable. 

> I wonder if we could make stages explicit.

I think MetaOCaml's or other similar systems approach (Template Haskell, Scala macros, etc) are somewhat different because they operate on pieces of code into which you splice other pieces of code (quotes and splices are two building blocks), so syntax is naturally very clear around what is code and what is a hole in that code into which you can splice more code.

This proposal describes (optional) compile time specializaton - so I think the fact that syntax looks the same is kinda by design. 

> Something else that I just realized is that if you evaluate some code in runtime in some cases and compile time in others, you can get runtime error in some cases and compile-time error in others.

Yep, I think this is by design as well. Though maybe not written as explicitly.

PR #4376, opened 2025-05-14
Fix a handful of augmentations issues.

### LINKED ISSUE #3849: Augmentations in Part files with imports, "augmentation path property"
TL;DR: The path requirement on augmentations disallows augmentations in sibling parts. We should consider making that a lint, not a language requirement.

The specification of "Parts with imports" (https://github.com/dart-lang/language/pull/3800) which combines augmentation libraries and part files into on concept, adds a restriction that requires augmenting declarations to be "below" the prior base or augmenting declaration that they are applied after. Here "below" means *later in the same file* or *in the part-file subtree of a part file of the same file*.

Because every augmenting declaration has to satisfy that, relative to the previously applied on, the transitive effect is that the files that contain augmentations of the same declaration must be on a single downwards path in the part-file tree of the library, starting at the file of the base declaration.
Two "sibling" part files cannot contain augmentations of the same declartion.

That requirement was added beause satisfying it made part directive ordering inside a file unrelated to augmentation application order, because all augmentations would be in a single sub-part. Which is nice.

It has the side-effect that you can't augment a declaration in a sub-part file *and* apply a macro to the base declaration which also augments the declaration, because the macro generated augmentation is put into a separate sub-part file which will be a sibling of the manual augmentation.

That's probably not a *big* problem, most people won't be augmenting their own declarations. They can just write the result of the augmentation directly (except for a few potentially useful cases, where you'd write the augmentation right beside the thing it augments, in the same file, and then there is also no problem).

It's still a little worrisome, and it means that macro generated augmentations *must* be collected into a single part file. If not for this, they could have been in individual files, one per macro per phase, and the only practical problem with that would be showing them when debugging. But with the path-requirement, it's not even allowed.

## Proposal
I suggest that we remove the "path requirement", so that the only requirement is that an augmenting declaration must be below the *base* declaration it augments. 
_(Even that isn't technically necessary, but it makes both reasoning about the code and specifying augmentation easier.)_

Then we add a *lint* which warns if an augmentation is not below the previously applied augmentation. It's just a suggestion, it can be ignored, and the lint ignores augmentations generated by macros (generally shouldn't lint code the user didn't write). If you ignore it, you can't *necessarily* reorder part directives of a file, but then you probably know what you're doing.

The augmentation specification has to specify an augmentation application order, which is now (a little) more complicated because we can't assume that the augmentations are ordered one *below* another.

Definition: A declaration, *A*, in a library is *before* another declaration *B* (equivalently *B* is *after* *A*) in the same library if and only if:
* *B* is declared *below* *A* (later in same file as *A*, or in a sub-part-tree of that file)
* There exists a file in the library such that *A* and *B* are in different sub-part-trees of that file, and the 
   `part` directive for the sub-part-tree containing *B* is syntactically later in the file than 
   the `part` directive for the sub-part-tree containing *A*.

This defines a *total* ordering on declarations of a library. (Proof sketch: Take any two declarations. Induction on tree structure, starting at the library file whose sub-tree must contain both. Continue downwards as long as neither is in the current file, and not in different part sub-trees, because then they must be in the same part.)

The augmentation application ordering of augmentations of the same base declaration is to apply an augmenting declaration after all augmenting declaration that are *before* it and before all that are *after* it.

### LINKED ISSUE #4022: Parameter name of implicit setter.
The augmentation rules say that an augmenting declaration must have the same parameter names as the declaration it augments. 
>  an augmenting declaration with a parameter list which must have the same parameters (names, positions, optionality and types) as its augmented declaration

You are allowed to augment the implicit setter of a variable declaration like:
```dart
int x;
augment set x(...) { ...}
```
The problem is that you *cannot* declare a parameter with the same name as the augmented implicit setter, since it has no name. (Or if it has a name, nobody ever specified what that name is, so you still can't specify it.)

I suggest that we build on top of "wildcard names" (really *unnamed* declarations), and say that
* Positional parameters may have *no* name. A declaration of `int foo(int _) {}` has a positional parameter with *no name*.
* An augmenting declaration's parameter must have the same name as the augmented definition's parameter *if both have a name*. 
    * If the augmented definition's parameter has no name, but the augmenting declarations' parameter does, then that is the name of the parameter in the definition that results from applying the augmenting declaration.
    * If the augmented definition's parameter has a name, but the augmenting declarations' parameter does not, then that is the name of the parameter in the definition that results from applying the augmenting declaration, but the name is not in scope in a body of the augmenting declaration.
    * If neither has a name, the the parameter in the resulting definition also has no name, and later augmentations are still free to introduce a name.
    * (And if both have names, it must be the same name.)

We can *maybe* do the same for type parameters. It's trickier since we might then not be able to express an inherited type, fx:
```dart
T id<T>(T value) => value;
augment id<_>(v) { log("id!"); return augmented(v); }
```
The *result* of that augmentation application is effectively `T id<T>(T v) { ... }` because the types are inherited, and the *type parameter name* is inherited, the name is just not in scope in the augmenting declaration, but since it doesn't use the name, everything is inferred anyway, that doesn't matter.
And then you can write `void foo<_>();` and let an augmentation make that inot `void foo<T>() { ... using T ...}`. Tricky, but consistent.

(Have to be very careful when specifying this, to make sure an inherited type or bound has the correct type variable name. Or maybe it's easy, because if an inherited type *refers* to any name, then that will also be the name of the type parameter in the result.)

### LINKED ISSUE #4047: For extension type constructor augmentation, relax the rule about initializing formals?
Thanks to @sgrekhov for bringing up this topic! Consider an augmenting declaration in an extension type declaration:

```dart
extension type A(int b) {
  augment A(int b) {
    assert(b > 0);
  }
}
```

This example is used in the augmentation feature specification, and it is implied that it is correct. However, the existing proposal about [primary constructors](https://github.com/dart-lang/language/blob/main/working/2364%20-%20primary%20constructors/feature-specification.md) is used in the reasoning about this example, and it doesn't quite match.

Primary constructors are not yet part of Dart, but it is expected that they will be added, and also that the `<representationDeclaration>` of an extension type declaration will become part of a primary constructor declaration (that is, `(int b)` in the example above will become a primary constructor, and in other cases where will be some extra elements, e.g., `const` before the name, hence 'part of').

As a primary constructor, the declaration above has the following meaning:

```dart
extension type A {
  final int b;
  A(this.b);
  augment A(int b) {
    assert(b > 0);
  }
}
```

This conflicts with the general rule about augmentations that a formal parameter of a constructor must consistently be an initializing formal in the introductory declaration and in each augmenting declaration, or it must consistently not be an initializing formal. (That is, we can't go back and forth, we have to choose either `int b` or `this.b`, and stick with that.)

The example could just be changed to use the form which isn't an error according to this reasoning:

```dart
extension type A(int b) {
  augment A(this.b) {
    assert(b > 0);
  }
}

// ... which means:
extension type A {
  final int b;
  A(this.b);
  augment A(this.b) {
    assert(b > 0);
  }
}
```

However, this could also be a good opportunity to discuss the rules about constructor parameters (in general, not just for extension types): We could allow a stack of augmenting declarations of a constructor to use an initializing formal (like `this.b` or `int this.b` or `{required this.b = 0}`, etc) or a regular formal parameter (e.g., `int b` or `{required int b = 0}`) interchangeably.

This would allow us to use the first version of the example, which is arguably more readable.

```dart
// Then we'd allow the following:
extension type A(int b) {
  augment A(int b) {
    assert(b > 0);
  }
}

// ... which means:
extension type A {
  final int b;
  A(this.b);
  augment A(int b) { // OK, just say `this.b` once in an augmentation stack.
    assert(b > 0);
  }
}
```

This would immediately generalize to classes, enums, etc: **The fact that a given formal parameter is an initializing formal must be stated at least once in an augmentation chain, the remaining declarations need not be initializing.**

The status of being an initializing formal in the _definition_ (the result after merging) of the given constructor would then be determined by the existence of an initializing formal: If there is one or more than one initializing formals then it is initializing, otherwise it is a regular formal parameter.

An extension type with a `<representationDeclaration>` would always implicitly introduce a constructor with an initializing formal, which means that we would also have this semantics for extension types (which is necessary because we _must_ initialize the representation variable of an extension type).

In any case, the status of being an initializing formal is a piece of implementation, not an API property, and this seems to justify that a stack of augmenting declarations can be allowed to specify or specify that particular implementation property.

### LINKED ISSUE #4060: Allow "abstract" mixin application classes with missing parts in augmentations?
Thanks to @sgrekhov for bringing up this topic. This issue is similar to https://github.com/dart-lang/language/issues/4025 and https://github.com/dart-lang/language/issues/4008 in that it is concerned with incomplete declarations whose semantics can be well-defined because the missing elements are provided by augmentations.

In this issue it's about mixin application classes (the class declarations with an `=`):

```dart
class A {}
mixin M1 {}
mixin M2 {}

class B = A with M1;
augment class B1 = A with M2; // Allowed? Means `B1 = A with M1, M2`.
augment class B2 = with M2; // Allowed? Means `B2 = A with M1, M2`.
```

I'd recommend that we keep this in mind, but don't do anything at this time. It should not be a problem to add support for this kind of augmentation later on, and it will not be a breaking change.

### LINKED ISSUE #4356: How do you augment an enum without adding values?

munificent:
*   Remove recommend path ordering lint. Commit to making `part` directive
    order meaningful and acceptable to rely on.
*   Allow enum declarations without values.
*   Specify signature matching for implicit setters from abstract variables.
*   Clarify that you can't augment an extension type constructor and add a body.
*   Don't allow augmenting mixin application classes.

Fix #3849.
Fix #4356.
Fix #4022.
Fix #4047.
Fix #4060.

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -106,10 +106,11 @@ from changing the constructor to a factory.
 
 ## Syntax
 
-The grammar changes are fairly simple. The grammar is modified to allow an
-`augment` modifier before various declarations. Also, functions, getters, and
-setters are allowed to have `;` bodies even when not instance members (except
-for local functions, which must still have a real body).
+There are a handful of grammar changes mostly around allowing `augment` before
+the various declarations that can be augmented allowing declarations to be less
+"complete" than they were required to be before this proposal since an
+augmentation can fill them in. For example, non-instance functions without
+bodies and enums without values.
 
 ```
 topLevelDeclaration ::= classDeclaration
@@ -133,7 +134,7 @@ classDeclaration ::=
     'augment'? (classModifiers | mixinClassModifiers)
     'class' typeWithParameters superclass? interfaces?
     memberedDeclarationBody
-  | 'augment'? classModifiers 'mixin'? 'class' mixinApplicationClass
+  | classModifiers 'mixin'? 'class' mixinApplicationClass
 
 mixinDeclaration ::=
     'base'? 'mixin' typeIdentifier typeParameters?
@@ -155,11 +156,6 @@ extensionTypeDeclaration ::=
     typeParameters? interfaces?
     memberedDeclarationBody
 
-enumType ::= 'augment'? 'enum' typeIdentifier
-  typeParameters? mixins? interfaces?
-  '{' enumEntry (',' enumEntry)* (',')?
-  (';' memberDeclarations)? '}'
-
 typeAlias ::= 'augment'? 'typedef' typeIdentifier typeParameters? '=' type ';'
   | 'augment'? 'typedef' functionTypeAlias
 
@@ -169,10 +165,6 @@ memberDeclarations ::= (metadata memberDeclaration)*
 
 memberDeclaration ::= declaration
   | 'augment'? methodSignature functionBody
-
-enumEntry ::= metadata 'augment'? identifier argumentPart?
-  | metadata 'augment'? identifier typeArguments?
-    '.' identifierOrNew arguments
 
 declaration ::=
     'augment'? 'external'? factoryConstructorSignature ';'
@@ -206,81 +198,28 @@ of the introductory declaration and can't specify it.*
 specifying an `on` clause. Only the introductory declaration permits that. We
 could relax this restriction if compelling use cases arise.*
 
+### Enums
+
+For enum declarations, in addition to the `augment` modifier, we allow declaring
+an enum (or augmentation of one) with no values. This is useful if the
+introductory declaration wants to let the augmentation fill in all values, or if
+the augmentation wants to add members but no values.
+
+When there are no values, the enum still requires a leading `;` before the first
+member to avoid ambiguity.
+
+```
+enumType ::= 'augment'? 'enum' typeIdentifier typeParameters?
+    mixins? interfaces? '{' enumBody? '}'
+
+enumBody ::= enumEntry (',' enumEntry)* (',')? (';' memberDeclarations)?
+    | ';' memberDeclarations
+```
+
+*Note that an enum can also have neither values nor members and both `{}` and
+`{;}` are valid.*
+
 ## Static semantics
-
-### Declaration ordering relations
-
-As part of the meta-programming and augmentation features, we expand the
-[capabilities of part files][parts with imports]. With that feature, a part file
-can now have its own `import` and `export` directives, and further nested `part`
-files, with part files inheriting the imports and prefixes of their parent (part
-or library) file.
-
-[parts with imports]: parts_with_imports.md
-
-Augmentation declarations interact with part files in restrictions on where an
-augmenting declaration may occur relative to the declaration it augments. We
-define the following relations on *declarations* based on the relations between
-*files* of a library.
-
-We say that a syntactic declaration *occurs in* a Dart file if the
-declaration's source code occurs in that Dart file.
-
-We say that a Dart file *contains* a declaration if the declaration occurs in
-the file itself, or if any of the part files transitively included by the Dart
-file contain the declaration. *That is, if the declaration occurs in a file in
-the part subtree of that Dart file.*
-
-We then define two orderings of declarations in a library, one partial and one
-complete.
-
-#### "Is above"
-
-A syntactic declaration *A* *is above* a syntactic declaration *B* if and only
-if:
-
-*   *A* and *B* occur in the same file, and *A*'s declared name is syntactically
-    before *B*'s declared name, in source order, or
-
-*   The file where *A* occurs includes the file where *B* occurs. *In other
-    words, if there is a `part` chain from the file where A is declared to the
-    file where B is declared, then A is above B.*
-
-This is a partial order. If A and B occur in sibling part files where neither
-declaration's file contains the other, then there is no is above relation
-between the declarations.
-
-#### "Is before" and "is after"
-
-For any two syntactic declarations *A*, and *B*:
-
-*   If *A* is above *B* then *A* is before *B*.
-
-*   If *B* is above *A* then *B* is before *A*.
-
-*   Otherwise, *A* and *B* are in sibling branches of the part tree:
-
-    *   Let *F* be the least containing file for those two files. *Find the
-        nearest root file in the part subtree that contains both A and B.
-        Neither A nor B will occur directly in F because if it did, then A or B
-        would be above the other and the previous clauses would have handled
-        it.*
-
-    *   If the `part` directive in *F* including the file that contains *A* is
-        syntactically before the `part` directive in *F* including the file that
-        contains *B* in source order, then *A* is before *B*.
-
-    *   Otherwise *B* is before *A*.
-
-Then *B* *is after* *A* if and only if *A* *is before* *B*.
-
-*In short, we complete the partial "is above" order by taking the order of
-`part` directives themselves into account when declarations are in sibling
-branches of the part tree.*
-
-This order is total. It effectively orders declarations by a pre-order
-depth-first traversal of the file tree, visiting declarations of a file in
-source order, and then recursing on `part` directives in source order.
 
 ### Augmentation context
 
@@ -459,7 +398,7 @@ We say that *I* is the declaration which is *augmented by* *D*.
 *In other words, take all of the declarations with the same name in some
 augmentation context, order them according to the "after" relation, and each
 augments the result of all the prior augmentations applied to the original
-declaration . The first one must not be marked `augment` and all the subsequent
+declaration. The first one must not be marked `augment` and all the subsequent
 ones must be.*
 
 An augmentation declaration does not introduce a new name into the surrounding
@@ -525,6 +464,11 @@ augment class C {
 }
 ```
 
+An extension type declaration induces an implicit unnamed constructor that is
+complete. *This is because it implicitly initializes the representation field.
+That implies that you can't augment an extension type declaration's constructor
+to give it a body.*
+
 ### Application order
 
 The same declaration can be augmented multiple times by separate augmentation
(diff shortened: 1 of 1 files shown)


working/augmentation-libraries/feature-specification.md
```diff
@@ -459,7 +398,7 @@ We say that *I* is the declaration which is *augmented by* *D*.
 *In other words, take all of the declarations with the same name in some
 augmentation context, order them according to the "after" relation, and each
 augments the result of all the prior augmentations applied to the original
-declaration . The first one must not be marked `augment` and all the subsequent
+declaration. The first one must not be marked `augment` and all the subsequent
```
lrhn, line 401: (Still think we should make `augment` optional, just like `override`. You can use it to document intent, but you don't have to. It'll try to augment either way, and fail in the same cases, but you also get an error if you write `augment` and it doesn't augment anything.)

working/augmentation-libraries/feature-specification.md
```diff
@@ -525,6 +464,11 @@ augment class C {
 }
 ```
 
+An extension type declaration induces an implicit unnamed constructor that is
```
lrhn, line 467: Not necessarily unnamed. `extension type const Foo._(int _) {}`

working/augmentation-libraries/feature-specification.md
```diff
@@ -525,6 +464,11 @@ augment class C {
...
+complete. *This is because it implicitly initializes the representation field.
+That implies that you can't augment an extension type declaration's constructor
+to give it a body.*
+
```
lrhn, line 471: An extension type declaration also introduces a complete getter for the representation variable.

working/augmentation-libraries/feature-specification.md
```diff
@@ -616,6 +626,59 @@ It's a **compile-time** error if:
...
+*   It has the same set of named parameter names as the augmented declaration.
+
+*   For all corresponding pairs of parameters:
+
+    *   They have the same type (after type annotation inheritance).
+
+    *   They have the same `required` and `covariant` modifiers.
+
+    *   If augmenting a constructor, they must both be initializing formals,
+        super parameters, or neither.
```
lrhn, line 658: No longer relevant. If a constructor has a super parameter or initializing parameter, it's complete and can't be augmented with implementation. (If it us agumented, it should be by an incomplete/abstract signature-only declaration that only adds metadata. That declaration should not have to repeat initializing parameters. Rather, it *must not*.)

working/augmentation-libraries/feature-specification.md
```diff
@@ -616,6 +626,59 @@ It's a **compile-time** error if:
...
+
+    *   The augmenting function's parameter name is `_`, or
+
+    *   The augmenting function's parameter name is the same as the name of the
+        corresponding positional parameter in every preceding declaration that
+        doesn't have `_` as its name.
+
+    *In other words, a declaration can ignore a positional parameter's name by
+    using `_`, but all declarations in the chain that specify a name have to
+    agree on it.*
```
lrhn, line 670: (Can an implementing augmentation omit a parameter name, and if it does so, can it refer to the parameter?
```dart
void foo(int x);
augment void foo(int _) {
  print(x); // Works or not?
}
```
(I'd say "works not!', the name of the implementing declaration is the one that matters to the implementation.)

working/augmentation-libraries/feature-specification.md
```diff
@@ -719,7 +767,24 @@ class C {
 }
 ```
 
-### Augmenting enum members
+For purposes of [signature matching][], the implicit setter induced by a
+non-final abstract variable has a positional parameter named `_`. *This means
```
lrhn, line 771: Also for non-abstract variables.
```dart
int x = 42;

@important
augment set x(value);
```
This should be valid. Otherwise we need to say what the name is, so that someone might be able to match it, or we force all augmentations to use `_`.

lrhn (approved):

working/augmentation-libraries/feature-specification.md
```diff
@@ -206,81 +198,28 @@ of the introductory declaration and can't specify it.*
...
-
-    *   Otherwise *B* is before *A*.
+```
+enumType ::= 'augment'? 'enum' typeIdentifier typeParameters?
+    mixins? interfaces? '{' enumBody? '}'
 
-Then *B* *is after* *A* if and only if *A* *is before* *B*.
+enumBody ::= enumEntry (',' enumEntry)* (',')? (';' memberDeclarations)?
+    | ';' memberDeclarations
+```
```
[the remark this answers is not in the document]
munificent: I'm hoping to eventually split the rest of the grammar changes out into more sections too. I find the giant wall of syntax really hard to read. It's hard to see what's actually *different* and what's the same.

working/augmentation-libraries/feature-specification.md
```diff
@@ -133,7 +134,7 @@ classDeclaration ::=
     'augment'? (classModifiers | mixinClassModifiers)
     'class' typeWithParameters superclass? interfaces?
     memberedDeclarationBody
-  | 'augment'? classModifiers 'mixin'? 'class' mixinApplicationClass
+  | classModifiers 'mixin'? 'class' mixinApplicationClass
```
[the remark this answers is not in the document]
munificent: > Does it say somewhere that you can't augment a mixin application class declaration with, fx, `augment class C with M2;`?

I mean, it doesn't say that you *can*... :)

Added a sentence to explicitly say they can't be augmented.

I'm on the fence about allowing classes to have `;` bodies. It might be nice once we have primary constructors. But it might be really not nice if we ever want to make semicolons optional. Either way, I don't want to make that change as part of this feature.

working/augmentation-libraries/feature-specification.md
munificent: Filed an issue to discuss: https://github.com/dart-lang/language/issues/4377

working/augmentation-libraries/feature-specification.md
munificent: Oh, good catch. Also, I just realized this is completely inconsistent with the section on augmenting extension types which says that you *can* augment the implicit constructor. I moved this paragraph down to that section and removed the inconsistent stuff.

working/augmentation-libraries/feature-specification.md
```diff
@@ -533,18 +477,84 @@ declaration has an augmented declaration which is itself an augmentation
...
+files.
+
+[parts with imports]: parts_with_imports.md
+
+We say that a syntactic declaration *occurs in* a Dart file if the declaration's
+source code occurs in that file.
+
+We say that a Dart file *contains* a declaration if the declaration occurs in
+the file itself, or if any of the part files transitively included by the Dart
+file contain the declaration. *That is, if the declaration occurs in a file in
```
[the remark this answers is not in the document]
munificent: Good point. Reworded. I went with the latter because I like the transitivity being explicit instead of relying on the reader to note the recursive definition.

working/augmentation-libraries/feature-specification.md
```diff
@@ -533,18 +477,84 @@ declaration has an augmented declaration which is itself an augmentation
...
+the part subtree of that Dart file.*
+
+For any two syntactic declarations *A*, and *B*:
+
+*   If *A* and *B* occur in the same file:
+
+    *   If *A*'s declared name is syntactically before *B*'s declared name,
+        in source order, then *A* is before *B.*
```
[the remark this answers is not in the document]
munificent: The grammar prohibits augmenting unnamed extensions. I went ahead and added some verbiage in the grammar section to emphasize that. Then here I just note that the relation isn't meaningful for unnamed declarations. I don't think we need to specify an ordering for things that are never ordered anyway.

working/augmentation-libraries/feature-specification.md
```diff
@@ -533,18 +477,84 @@ declaration has an augmented declaration which is itself an augmentation
...
+
+    *   Otherwise *B* is before *A*.
+
+*   Else if the file where *A* occurs includes the file where *B* occurs then
```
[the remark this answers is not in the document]
munificent: Oh, I wondered about that being not formally defined. Went ahead and added a definition here.

working/augmentation-libraries/feature-specification.md
```diff
@@ -533,18 +477,84 @@ declaration has an augmented declaration which is itself an augmentation
...
+    *   Let *F* be the least containing file for those two files. *Find the
+        nearest root file in the part subtree that contains both A and B.
+        Neither A nor B will occur directly in F because if it did, then the
+        previous clauses would have handled it.*
+
+    *   If the `part` directive in *F* including the file that contains *A* is
+        syntactically before the `part` directive in *F* including the file that
+        contains *B* in source order, then *A* is before *B*.
+
+    *   Otherwise *B* is before *A*.
```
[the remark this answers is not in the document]
munificent: 💯

working/augmentation-libraries/feature-specification.md
munificent: Oh, right! Added a paragraph explaining this.

working/augmentation-libraries/feature-specification.md
munificent: > Can an implementing augmentation omit a parameter name

Yes.

> and if it does so, can it refer to the parameter?

No! Added a note clarifying.

working/augmentation-libraries/feature-specification.md
munificent: It's not allowed. When I simplified the proposal, I said:

> Variable declarations themselves can't be augmented (by variables, getters, or setters, even if the getter or setter is incomplete), with the exception of abstract instance variable declarations.

We could loosen this and allowing augmenting variable declarations with incomplete getters and setters, but I don't think it's particularly useful and probably opens the door to more annoying edge cases.

working/augmentation-libraries/feature-specification.md
```diff
@@ -719,7 +767,24 @@ class C {
 }
 ```
 
-### Augmenting enum members
+For purposes of [signature matching][], the implicit setter induced by a
+non-final abstract variable has a positional parameter named `_`. *This means
+you can augment an abstract variable with a setter that uses whatever positional
+parameter name you want:*
```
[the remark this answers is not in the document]
munificent: Yeah, I read the comments on that issue several times. I think you're recommendation was a really good one.

munificent:
> I don't think the lone `;` at the beginning of an `augment enum` declaration that does not add any new enumerated values is pretty, but it does make sense and I don't have a better idea. ;-)

I don't love it either, but I think it will be rare in practice.

PR #4385, opened 2025-05-20
Add `s?.length` as a constant expression

eernstg:
Following the decision at the language team meeting on May 14 2025, this PR changes the language specification such that `s?.length` is a constant expression when `s` satisfies certain requirements.


eernstg:
Thanks!

Issue #4172, opened 2024-11-25
Revisit unimplemented factory constructors and default values


eernstg:
We need to introduce an extra rule about factory constructors that are subject to augmentation: The can omit default values, even for

With augmentations, a factory constructor can be unimplemented. It is possible to make a late choice about the nature of this constructor (it could be redirecting or it could be non-redirecting):

```dart
class A {
  A();
  factory A.foo(); // Unimplemented factory constructor.
}

augment class A {
  augment factory A.foo() = A;
}

class B {
  B();
  factory B.foo(); // Unimplemented factory constructor.
}

augment class B {
  augment factory B.foo() {
    return A();
  }
}
```

We may need to introduce an extra rule about these unimplemented factory constructors which says that **it is allowed for an optional parameter to omit the default value clause, even in the case where the parameter type is potentially non-nullable**.

For example:

```dart
class A {
  A([int i = 0]);
  factory A.foo([int i]); // Currently an error; this issue proposes to allow it.
}

augment class A {
  augment factory A.foo([int i]) = A; // Default value obtained from redirectee.
}
```

The general rules about augmentations state that no default value can be specified by an augmentation, it must always be specified by the introductory declaration (if there is to be a default value for that parameter at all).

This implies that the declaration above would not allow `A.foo` to be augmented with a class body (such that it is a non-redirecting factory constructor): The parameter type cannot be modified by an augmentation, and an augmentation cannot provide a default value. In other words, the fact that there is no default value has eliminated the expressive power to make the decision about the factory being redirecting or non-redirecting, it's always forced to be redirecting.

We could allow the unimplemented factory constructor to _include_ the default value and then ignore it (because it would be an error to have it) when the constructor is augmented to be redirecting:

```dart
class A {
  A([int i = 0]);
  factory A.foo([int i = 1]); // Currently allowed.
}

augment class A {
  augment factory A.foo([int i]) = A; // Currently an error; this issue suggests we might allow it.
}
```

If we do not allow this then the choice to have a default value will force the constructor to be non-redirecting.

In other words, the ability to "decide late" whether a given factory is redirecting or not will disappear as soon as there is one or more optional parameters whose type is potentially non-nullable. Of course, we can then make the declarations inconsistent: If we have two such parameters and provide a default value for exactly one of them then no augmenting declarations can be written, because they will give rise to a compile-time error no matter what.

[...]

Issue #3690, opened 2024-04-02
Augmentations should specify allowed partial declarations.


lrhn:
Augmentations can add missing pieces to existing declarations. If those missing pieces are *required*, then the initial declaration is not valid by itself.
The current specification would make that declaration a compile-time error.

Augmentations makes *some* such declarations still be compile-time errors that apply before augmentations are even applied, and other declarations be only partially declared, and filled in by an augmentation, in which case the property does not apply to the individual *declarations*, but to the entire declaration+augmentations stack, the "effective declaration".

*If* that allows a declaration to omit something that would otherwise be required by the *grammar*, we need to specify the allowed partial syntax too.

I'm not absolutely sure there are any cases where augmentations allow omitting something that is otherwise grammatically required, except maybe bodies of functions (which are grammatically allowed to be omitted, because of abstract declarations, but just not in every position - top-level functions cannot grammatically end in `;`).

It still means we need to go thrugh every "it's a compile-time error if declaration D ...." and decide whether that applies to the initial declaration syntax, or to the post-augmentation-application effective declaration. (And the latter case would imply that it's a semantic restriction, not a syntactic restriction.)

Some prose-introduced syntax errors are irrecoverable, which probably means being so invalid that applying an augmentation wouldn't make sense anyway. Other errors are not, and should be changed to only apply to the result of augmentation.

Examples:

* It's definitely and eagerly a compile-time error if a static method has a `covariant` parameter or a `this.x` or `super.x` parameter. We can't augment that, because it has no meaning. Anything that would add a meaning to the declaration that cannot possibly apply to that kind of declaration, should be an eager error.

* It's possibly not a compile-time error if a static method has a `;` body ("no body"), if a later applied augmentation adds a body.

* It's possibly not a compile-time error if a non-abstract class has an abstract member declaration, if a later augmentation adds a body (but that was never an eager error, because the class could inherit the implementation). Generally, anything related to satisfying an interface is going to be checked late.

* It's a compile-time error to have a `super.foo` invocation in an instance function, if the superclass doesn't have a `foo` member. It's probably allowed if an augmentation adds a superclass that does have a `foo`. (Which means the check should be written in a way that doesn't point to the surrounding class declaration, but to the semantic superclass of the surrounding class. It probably already is written that way, but we need to check. Everywhere, at least where we say "it's a compile-time error".)

davidmorgan:
[...]

PR #4393, opened 2025-05-28
Add rules to allow omission of certain default values

eernstg:
This PR changes the rules about augmentation of constructors such that an incomplete factory is able to defer the decision about implementing the constructor by a redirection or by a body. See https://github.com/dart-lang/language/issues/4172 for details.

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -733,7 +733,8 @@ It's a **compile-time** error if:
     matching] the signature of the augmented function.
 
 *   The augmenting function specifies any default values. *Default values are
-    defined solely by the introductory function.*
+    defined solely by the introductory function. Note that constructors are
+    treated differently from other functions in this respect.*
 
 *   A function is not complete after all augmentations are applied, unless it
     is in a context where it can be abstract. *Every function declaration
@@ -840,13 +841,21 @@ Augmenting constructors works similar to augmenting a function, with some extra
 rules to handle features unique to constructors like redirections and
 initializer lists.
 
+It is **not** a compile-time error for an incomplete factory constructor to
+omit default values. *That is, they are treated similarly to abstract
+instance methods in this respect. This allows the augmenting declaration to
+implement the constructor by adding a redirection or a body.*
+
 It's a **compile-time error** if:
 
 *   The signature of the augmenting function does not [match][signature
     matching] the signature of the augmented function.
 
-*   The augmenting constructor parameters specify any default values.
-    *Default values are defined solely by the introductory constructor.*
+*   The augmenting constructor parameters specify any default values,
+    and the constructor is a non-redirecting factory. *Default values are
+    defined by the introductory constructor, except when this precludes the
+    augmentation from choosing whether or not the constructor should be
+    redirecting.*
 
 *   The introductory constructor is `const` and the augmenting constructor
     is not or vice versa. *An augmentation can't change whether or not a


working/augmentation-libraries/feature-specification.md
```diff
@@ -840,13 +841,21 @@ Augmenting constructors works similar to augmenting a function, with some extra
 rules to handle features unique to constructors like redirections and
 initializer lists.
 
+It is **not** a compile-time error for an incomplete factory constructor to
+omit default values. *That is, they are treated similarly to abstract
```
munificent: This touches on #3690. With augmentations, it's also not a compile error for a static method declaration to have a `;` body, for a constructor declaration to not have initializers are all final instance fields, etc. As long as the final declaration ("definition"?) once all augmentations have been applied is valid, it's fine.

I'm not sure to what degree we want to spell this all out in the proposal but I'm cautious about having too much normative text that is potentially redundant or implied by other parts of the proposal. I don't want implementers to have to read between the lines to figure out what the proposal is saying, but I also don't want them to have to compare paragraphs and try to tell if the two paragraphs are defining *different* things or just restating a point. Finding the right balance here is hard.

working/augmentation-libraries/feature-specification.md
```diff
@@ -840,13 +841,21 @@ Augmenting constructors works similar to augmenting a function, with some extra
...
+
 It's a **compile-time error** if:
 
 *   The signature of the augmenting function does not [match][signature
     matching] the signature of the augmented function.
 
-*   The augmenting constructor parameters specify any default values.
-    *Default values are defined solely by the introductory constructor.*
+*   The augmenting constructor parameters specify any default values, and
+    the constructor is not a non-redirecting factory. *Default values are
```
munificent: If we're going to support allowing augmentations to specify default values at all, I don't see why we wouldn't generalize that and allow augmentations of all functions to fill them in. Why restrict this to non-redirecting factory constructors? We have to take the complexity hit either way.

munificent:
I like the general direction here, but if we're going to loosen the proposal so that augmentations can fill in default values, I think we should go all the way and let them do it for all functions.

eernstg:
I changed the rules to allow all kinds of augmentable functions (incl. constructors) to specify a default value for an optional parameter in the introductory declaration or in an augmentation, but at most once, PTAL!

eernstg:
@munificent, WDYT?

working/augmentation-libraries/feature-specification.md
```diff
@@ -768,13 +768,45 @@ augment class Person {
...
+
 It's a **compile-time** error if:
 
 *   The signature of the augmenting function does not [match][signature
     matching] the signature of the augmented function.
 
-*   The augmenting function specifies any default values. *Default values are
-    defined solely by the introductory function.*
+*   The augmentation chain has two or more specifications of a default value
+    for the same optional parameter. This is an error even in the case where
```
munificent: I think this might be a little more readable as: "More than one declaration in the augmentation chain specifies a default value for some optional parameter."

working/augmentation-libraries/feature-specification.md
```diff
@@ -768,13 +768,45 @@ augment class Person {
...
+    all of them are identical. *Default values are defined by the introductory
+    function or an augmentation, but at most once.*
+
+*   The augmentation chain has no specifications of a default value for an
```
munificent: Likewise here: "No declaration in the augmentation chain specifies a default value for an..."

working/augmentation-libraries/feature-specification.md
```diff
@@ -845,13 +877,27 @@ Augmenting constructors works similar to augmenting a function, with some extra
...
 
-*   The augmenting constructor parameters specify any default values.
-    *Default values are defined solely by the introductory constructor.*
+*   The augmentation chain has two or more specifications of a default value
+    for the same optional parameter. This is an error even in the case where
+    all of them are identical. *Default values are defined by the introductory
+    declaration or an augmentation, but at most once.*
+
+*   The augmentation chain has exactly one specification of a default value
+    for an optional parameter, and the constructor is a redirecting factory.
```
munificent: Is this error needed? This means that if the introductory declaration has a default value, the augmentation is prohibited from filling in a redirectory factory.

How about we just treat default values in redirecting factory constructors the same way we do abstract methods: they are meaningless documentation. You can write it but it won't do anything.

working/augmentation-libraries/feature-specification.md
```diff
@@ -863,6 +909,13 @@ It's a **compile-time error** if:
     not a constructor is generative because that affects whether users are
     allowed to call the constructor in a subclass's initializer list.*
 
+*   More than one constructor declaration in an augmentation chain specifies
+    a default value for the same parameter.
+
+*   A constructor declaration in an augmentation chain is a redirecting
+    factory, and some declaration in the chain specifies a default value
+    for one or more parameters.
```
munificent: See my previous comment. :)

working/augmentation-libraries/feature-specification.md
eernstg: This is actually about the final declaration and not about the ordering of diagnostics. The point is that this allows a factory constructor to be declared without default values of some parameters whose type is potentially non-nullable, and also without a body. This makes it a non-error to turn it into a regular factory (adding the default values in an augmenting declaration), and it also makes it a non-error to make it a redirecting factory (whose parameters can not have any default values).

In other words, it allows the direct/redirecting choice to be made by the augmenter. It seems reasonable to me that we treat this choice as an implementation detail, not an API property.

If we don't allow this approach (and insist that the incomplete factory must have all the required default values) then it is not possible to make such constructors redirecting. I think that's a loss of expressive power that we can just avoid, so why not?

working/augmentation-libraries/feature-specification.md
eernstg: Done!

working/augmentation-libraries/feature-specification.md
eernstg: Thanks, got it!

working/augmentation-libraries/feature-specification.md
eernstg: Got that, too!

working/augmentation-libraries/feature-specification.md
eernstg: It has always been an error to specify a default value for a parameter of a redirecting factory. I think we should handle this change as a general change (not just for augmentation chains). Also, I think it would make sense to have the ability to declare those default values in general on redirecting factory constructors, I honestly don't see why we don't support that feature. It may make invocations of redirecting factories slightly more expensive (because some compile-time inlining operations may now be actual invocations of that factory), but I don't think that's a good reason to have this weird non-support.

eernstg: I created https://github.com/dart-lang/language/issues/4499 to discuss the handling of redirecting factories and default values. So I'll close this thread.

eernstg:
@munificent, this one was rebased around the file renaming operations, PTAL.

working/augmentation-libraries/feature-specification.md
eernstg: #4499 would treat this situation in a way which is yet more flexible: It would be allowed to specify default values also in the introductory (or any other incomplete) declaration, and this still wouldn't prevent the constructor from being implemented using redirection (of course, a regular, non-redirecting factory would still be possible as well).

working/augmentations/feature-specification.md
```diff
@@ -1128,6 +1174,10 @@ and assume the third point is always true.
 
 ## Changelog
 
+### 1.38
+
+*   Generalize the treatment of default values of optional parameters.
```
munificent: Can you add `(#4172)` here so that anyone reading this knows the issue this change fixes?

munificent (approved):
One small thing then still LGTM. Thanks!

working/augmentations/feature-specification.md
eernstg: Done.

PR #4395, opened 2025-06-04
Allow variables to be augmenting and augmented.

### LINKED ISSUE #4387: Can an abstract getter/setter pair be implemented by an augmenting declaration of a variable?
Without augmentations we can have the following:

```dart
abstract class A {
  abstract int i; // Or `int get i; set i(int _);`.
}

class B implements A {
  int i;
  B(this.i);
}
```

This illustrates that it is a completely normal and useful situation to implement an abstract getter (and possibly setter) by declaring a variable. I'd expect a similar situation to arise with augmentations:

```dart
class A1 {
  abstract int i; // Or `int get i; set i(int _);`.
}

augment class A1 {
  augment var i; // Adds an implementation of said getter and setter.
  A(this.i);
}

class A2 {
  abstract final int i; // Or `int get i;`
}

augment class A2 {
  augment final i;
  A(this.i);
}
```

However, I cannot see any rules in the proposed [feature specification](https://github.com/dart-lang/language/blob/main/working/augmentation-libraries/feature-specification.md) to allow or disallow this.

One thing to consider is the initialization of these variables: I do not think there are any rules preventing the constructor from being added by augmentation (in particular, the class will not acquire a default constructor based on the fact that the introductory declaration doesn't declare any constructors, that decision is taken after all augmentations have been taken into account).

This means that if a class acquires several variables by augmentation (like `A1` or `A2` above) then the initialization must occur in an initializing expression in the variable declaration itself, or it must be done in a single constructor declaration. In other words, we can't write one constructor declaration initializing some variables, and then another (augmenting) constructor declaration that adds initialization of some additional instance variables.

I don't think this is a problem, though, because it isn't more strict than the treatment of many other aspects of augmentation: We can add an implementation of many things, but it must be done in one step.

@dart-lang/language-team, WDYT? It seems likely to me that this would be useful, and I do not think it breaks the underlying rules (in particular: implementations can be added, not overridden or combined).

munificent:
Also allow the abstract variable syntax at the top level and for static fields.

Fix #4387.

### DIFF
--- a/working/augmentation-libraries/feature-specification.md
+++ b/working/augmentation-libraries/feature-specification.md
@@ -106,11 +106,25 @@ from changing the constructor to a factory.
 
 ## Syntax
 
-There are a handful of grammar changes mostly around allowing `augment` before
-the various declarations that can be augmented allowing declarations to be less
-"complete" than they were required to be before this proposal since an
-augmentation can fill them in. For example, non-instance functions without
-bodies and enums without values.
+The syntax changes are simple but fairly extensive and touch several parts of
+the grammar so are broken out into separate sections.
+
+### Top-level augmentations and incomplete top-level members
+
+We allow an `augment` modifier before most top-level declarations.
+
+Also, we allow incomplete declarations at the top level. This reuses the same
+syntax used inside a class to declare abstract variables, methods, getters,
+setters, and operators. For callable members, that means the body is `;`. For
+variable declarations, that means using `abstract`. *Example:*
+
+```dart
+abstract int x;   // Incomplete top-level variable.
+int get y;        // Incomplete top-level getter.
+set z(int value); // Incomplete top-level setter.
+```
+
+The new top-level grammar is:
 
 ```
 topLevelDeclaration ::= classDeclaration
@@ -123,13 +137,21 @@ topLevelDeclaration ::= classDeclaration
   | 'augment'? 'external' getterSignature ';'
   | 'augment'? 'external' setterSignature ';'
   | 'augment'? 'external' finalVarOrType identifierList ';'
-  | 'augment'? functionSignature (functionBody | ';')
+  | 'augment'? 'abstract' finalVarOrType identifierList ';'
   | 'augment'? getterSignature (functionBody | ';')
   | 'augment'? setterSignature (functionBody | ';')
+  | 'augment'? functionSignature (functionBody | ';')
   | 'augment'? ('final' | 'const') type? initializedIdentifierList ';'
   | 'augment'? 'late' 'final' type? initializedIdentifierList ';'
   | 'augment'? 'late'? varOrType initializedIdentifierList ';'
+```
 
+### Class-like declarations
+
+We allow `augment` before class, enum, and extension type declarations. *(Enums
+are handled in the next section.)*
+
+```
 classDeclaration ::=
     'augment'? (classModifiers | mixinClassModifiers)
     'class' typeWithParameters superclass? interfaces?
@@ -164,8 +186,8 @@ declaration ::=
   | 'augment'? 'external'? 'static'? getterSignature ';'
   | 'augment'? 'external'? 'static'? setterSignature ';'
   | 'augment'? 'external'? 'static'? functionSignature ';'
-  | 'external' ('static'? finalVarOrType | 'covariant' varOrType) identifierList ';'
   | 'augment'? 'external'? operatorSignature ';'
+  | 'external' ('static'? finalVarOrType | 'covariant' varOrType) identifierList ';'
   | 'augment'? 'abstract' (finalVarOrType | 'covariant' varOrType) identifierList ';'
   | 'static' 'const' type? initializedIdentifierList ';'
   | 'static' 'final' type? initializedIdentifierList ';'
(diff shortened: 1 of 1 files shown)


working/augmentation-libraries/feature-specification.md
```diff
@@ -123,13 +137,21 @@ topLevelDeclaration ::= classDeclaration
...
   | 'augment'? setterSignature (functionBody | ';')
+  | 'augment'? functionSignature (functionBody | ';')
   | 'augment'? ('final' | 'const') type? initializedIdentifierList ';'
   | 'augment'? 'late' 'final' type? initializedIdentifierList ';'
   | 'augment'? 'late'? varOrType initializedIdentifierList ';'
+```
+
+### Class-like declarations
+
+We allow `augment` before class, enum, and extension type declarations. *(Enums
```
lrhn: and also `mixin` and `extension` declarations?

working/augmentation-libraries/feature-specification.md
```diff
@@ -180,14 +202,26 @@ declaration ::=
...
-declaration doesn't allow also specifying a representation field. This is by
-design. An extension type augmentation always inherits the representation field
-of the introductory declaration and can't specify it.*
+As with top-level declarations, we also reuse the abstract member syntax with a
+`static` modifier to allow declaring incomplete static fields, methods, getters,
+setters, and operators. *Example:*
+
+```dart
+class C {
+  static abstract int x;   // Incomplete static variable.
```
lrhn: Incomplete static getter and setter, really.

It's a variable declaration, syntactically, but semantically it's a getter and setter signature declaration.
(I guess it's a question of which abstraction level we're talking at. As declarations, the comments are correct, but the phrasing is "declaring incomplete static fields, methods ...", and a declaration *is* syntax, but it *declares* sementics, and there is no semantic abstract variable.)

working/augmentation-libraries/feature-specification.md
```diff
@@ -738,85 +782,52 @@ It's a **compile-time** error if:
...
-
-augment class C {
-  int x = 3;
-
-  static int y = 4;
-}
-```
+For purposes of augmentation, a variable declaration is treated as implicitly
+defining a getter whose return type is the type of the variable. If the variable
+is not `final` or is `late`, then the variable declaration also implicitly
```
lrhn: ", or is `late` and does not have an initializer expression"

working/augmentation-libraries/feature-specification.md
```diff
@@ -738,85 +782,52 @@ It's a **compile-time** error if:
...
-}
-```
+*   The signature of the augmenting getter or setter does not [match][signature
+    matching] the signature of the augmented getter or setter.
 
-For purposes of [signature matching][], the implicit setter induced by a
-non-final abstract variable has a positional parameter named `_`. *This means
-you can augment an abstract variable with a setter that uses whatever positional
-parameter name you want:*
+*   An augmenting setter specifies any default values. *Default values are
```
lrhn: Setters cannot have optional parameters, so that would be an error in any case.

working/augmentation-libraries/feature-specification.md
```diff
@@ -738,85 +782,52 @@ It's a **compile-time** error if:
...
+    defined solely by the introductory declaration.*
 
-```dart
-class C {
-  abstract int x;
-}
+*   A `const` variable declaration is augmented or augmenting.
```
lrhn, line 812: (We could allow `@Annotation() abstract const int foo;` to add annotations to `const` variable declarations. Constant variables have signatures too, and a constant getter. We could have an incomplete constant variable declaration, and allow a generated implementation to augment it.)

working/augmentation-libraries/feature-specification.md
```diff
@@ -738,85 +782,52 @@ It's a **compile-time** error if:
...
 
-augment class C {
-  augment set x(int anyNameIWant) { // OK.
-    // ...
-  }
-}
-```
+*   A getter or setter (including one implicitly induced by a variable
+    declaration) is not complete after all augmentations are applied, unless it
+    is in a context where it can be abstract. *Every getter or setter
```
lrhn: Maybe: "unless it's an instance member, and the surrounding class is abstract, or it's a generative constructor"

An instance member of an abstract class can remain abstract, and we can give an abstract generative constructor a trivial default implementation.

I can't think of any other place where a member can validly be abstract or implicitly implemented.

lrhn (approved):
Looks good, only pedantries in comments.

working/augmentation-libraries/feature-specification.md
munificent: Well, we are in the "Syntax" section... :)

Added a parenthetical.

working/augmentation-libraries/feature-specification.md
munificent: Ha, I knew I was going to get an edge case wrong here.

working/augmentation-libraries/feature-specification.md
munificent: Oh, of course. Removed.

working/augmentation-libraries/feature-specification.md
munificent: I can't imagine it's worth it, so let's just leave it out.

munificent:
I'm going to go ahead and land this, but I'm still open to any feedback you have, @eernstg. :)

PR #4410, opened 2025-06-12
Add proposal for private named parameters.

munificent:
To divide the labor up some on primary constructors, I figured it makes sense to handle private named parameters as a separate feature with its own independent proposal.

I do hope we ship them both as a batch, but this way it's easier for us and the implementation teams to work on them in parallel or independently.

### DIFF
--- a/working/2509-private-named-parameters/feature-specification.md
+++ b/working/2509-private-named-parameters/feature-specification.md
@@ -1,0 +1,382 @@
+# Private Named Parameters
+
+Author: Bob Nystrom
+
+Status: In-progress
+
+Version 0.1 (see [CHANGELOG](#CHANGELOG) at end)
+
+Experiment flag: private-named-parameters
+
+This proposal makes it easier to initialize and declare private instance fields
+using named constructors parameters. It addresses [#2509][] and turns code like
+this:
+
+[#2509]: https://github.com/dart-lang/language/issues/2509
+
+```dart
+class House {
+  int? _windows;
+  int? _bedrooms;
+  int? _swimmingPools;
+
+  House({
+    int? windows,
+    int? bedrooms,
+    int? swimmingPools,
+  })  : _windows = windows,
+        _bedrooms = bedrooms,
+        _swimmingPools = swimmingPools;
+}
+```
+
+Into this:
+
+```dart
+class House {
+  int? _windows;
+  int? _bedrooms;
+  int? _swimmingPools;
+
+  House({this._windows, this._bedrooms, this._swimmingPools});
+}
+```
+
+## Motivation
+
+Dart uses a leading underscore in an identifier to make a declaration private to
+its library. Privacy is only meaningful for declarations that could be accessed
+from outside of the library: top-level declarations and members on types.
+
+Local variables and parameters aren't in scope outside of the library where they
+are defined, so privacy doesn't come into play. Except, that is, for named
+parameters. A *named* parameter has one foot on each side of the function
+boundary. The parameter defines a local variable that is accessible inside the
+function, but it also specifies the name used at the callsite to pass an
+argument for that parameter:
+
+```dart
+test({String? _hmm}) {
+  print(_hmm);
+}
+
+main() {
+  test(_hmm: 'ok?');
+}
+```
+
+A public function containing a named parameter whose name is private raises
+difficult questions. Is there any way to pass an argument to the function from
+outside of the library? If the parameter is required, does that mean the
+function is effectively uncallable? Or do we not treat the identifier as private
+even though it starts with an underscore if it happens to be a parameter name?
+
+The language currently resolves these questions by routing around them: it is a
+compile-time error to have a named parameter with a private name. Users must use
+a public name instead. For most named parameters, this restriction is harmless.
+The parameter is only used within the body of the function and its idiomatic for
+local variables to not have private names anyway.
+
+### Initializing formals
+
+However, initializing formals (the `this.` before a constructor parameter)
+complicate that story. When a named parameter is also an initializing formal,
+then the name affects *three* places in the program:
+
+1.  The name of the parameter variable inside the body of the constructor.
+
+2.  The name used to pass an argument at the callsite.
+
+3.  The name of the corresponding instance field to initialize with that
+    parameter.
+
+For example:
+
+```dart
+class House {
+  int? bedrooms; // 3. The corresponding field.
+  House({this.bedrooms}) {
+    print(bedrooms); // 1. The parameter variable.
+  }
+}
+
+main() {
+  House(bedrooms: 2); // 2. The argument name.
+}
+```
+
+This creates a tension. A user may want:
+
+*   **To make a field private.** We [actively encourage them to do so][effective
+    private] because encapsulation is a key software engineering principle.
+
+*   **To make a constructor parameter named.** This is idiomatic in the Flutter
+    framework and Flutter applications and can be important for readability if
+    a constructor takes multiple parameters of the same type.
+
+*   **To use an initializing formal to initialize a field from a constructor
+    parameter.** It's the most concise way to initialize a field, avoids the
+    unusual initializer syntax, and makes it clear to a reader that the field
+    is directly initialized from that parameter.
+
+They want encapsulation, readability at callsites, and brevity in the class
+definition, but because of the compile-error on private named parameters, they
+can only [pick two][].
+
+[effective private]: https://dart.dev/effective-dart/design#prefer-making-declarations-private
+
+[pick two]: https://en.wikipedia.org/wiki/Project_management_triangle
+
+### Workaround
+
+When users run into this restriction, they may deal with it by sacrificing one
+of the three features:
+
+*   They can make the field public and just expect users to not use it. (Often,
+    the class is in application code where it's not imported anyway so it
+    doesn't matter much.)
+
+*   They can make the parameter positional instead.
+
+*   They can commit to the API they want by making the field public and the
+    parameter named, but use a public name for the parameter. Then instead of
+    using an initializing formal, they manually initialize the field from the
+    parameter, as in:
+
+    ```dart
+    class House {
+      int? _windows;
+      int? _bedrooms;
+      int? _swimmingPools;
+
+      House({
+        int? windows,
+        int? bedrooms,
+        int? swimmingPools,
+      })  : _windows = windows,
+            _bedrooms = bedrooms,
+            _swimmingPools = swimmingPools;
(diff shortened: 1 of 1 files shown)


working/2509-private-named-parameters/feature-specification.md
```diff
@@ -0,0 +1,382 @@
...
+Author: Bob Nystrom
+
+Status: In-progress
+
+Version 0.1 (see [CHANGELOG](#CHANGELOG) at end)
+
+Experiment flag: private-named-parameters
+
+This proposal makes it easier to initialize and declare private instance fields
+using named constructors parameters. It addresses [#2509][] and turns code like
```
eernstg: Typo:
```suggestion
using named constructor parameters. It addresses [#2509][] and turns code like
```

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -0,0 +1,382 @@
...
+When users run into this restriction, they may deal with it by sacrificing one
+of the three features:
+
+*   They can make the field public and just expect users to not use it. (Often,
+    the class is in application code where it's not imported anyway so it
+    doesn't matter much.)
+
+*   They can make the parameter positional instead.
+
+*   They can commit to the API they want by making the field public and the
```
eernstg: Typo:
```suggestion
*   They can commit to the API they want by making the field private and the
```

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -0,0 +1,382 @@
...
+With that feature, users will encounter the limitation around private named
+parameters even more often. In-header primary constructors exacerbate this issue
+because an in-header primary constructor *can't* have an initializer list. Users
+will be forced to either make the field public, the parameter positional, or not
+use a primary constructor at all.
+
+We don't want to put users in a position where the code that's most pleasant to
+write requires them to sacrifice semantic properties they want like
+encapsulation.
+
```
eernstg, line 224: At this point it would make sense to add a short paragraph to mention that this proposal contains a couple of rules that are concerned with primary constructors (just so the reader doesn't get confused when this proposal talks about a 'field parameter').

eernstg: OK, reading it again I can see that line 267 says '(for a primary constructor)', but I didn't notice that on the first reading.

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -0,0 +1,382 @@
...
+*   Else the field parameter induces an instance field with name *p*.
+
+*For example:*
+
+```dart
+class Id {
+  late final int _region = 0;
+
+  this({this._region, final int _value}) : assert(_region > 0 && _value > 0);
```
eernstg, line 315: Even here, I think it would be helpful to add a comment saying something like `// If primary constructors are supported.`

eernstg (approved):
LGTM! Just commented on a couple of locations that look like a typo.

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -0,0 +1,382 @@
...
+3.  The name of the corresponding instance field to initialize with that
+    parameter.
+
+For example:
+
+```dart
+class House {
+  int? bedrooms; // 3. The corresponding field.
+  House({this.bedrooms}) {
+    print(bedrooms); // 1. The parameter variable.
```
lrhn: Actually references the instance variable. The `this.bedrooms` parameter only introduce a variable in the initializer list.

So, an example would be:
```dart
class House {
  int? bedrooms; // 3. The corresponding field.
  House({this.bedrooms}) 
      : assert(bedrooms == null || bedrooms >= 0); // 1. The parameter variable
}
```

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -0,0 +1,382 @@
...
+public name, then that is the private name's corresponding public name. *For
+example, the corresponding public name of `_foo` is `foo`.* If removing the
+underscore does not leave something which is is a valid identifier *(as in `_`
+or `_2x`)* or leaves another private name *(as in `__x`)*, then the private name
+has no corresponding public name.
+
+Given an initializing formal or field parameter (for a primary constructor) with
+private name *p* in constructor C:
+
+*   If *p* has no corresponding public name *n*, then compile-time error. *You
```
lrhn, line 287: Only if it's a named parameter?
Or do we not allow privately named positional parameters. (It's consistent, but more breaking, you can no longer do `foo(int _id, {String? id})` which you technically could before.

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -0,0 +1,382 @@
...
+
+*   If any other parameter in C has declared name *p* or *n*, then
+    compile-time error. *If removing the `_` leads to a collision with
+    another parameter, then there is a conflict.*
+
+*   Otherwise, the name of the parameter in C is *n*. *If the parameter is
+    named, this then avoids the compile-time error that would otherwise be
+    reported for a private named parameter.*
+
+*   The local variable in the initializer list scope of C is *p*. *In the
```
lrhn: The name of the ...

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -0,0 +1,382 @@
...
+
+main() {
+  print(Id(region: 1, value: 2)); // Prints "Id(1, 2)".
+}
+```
+
+*Note that the proposal only applies to initializing formals and field
+parameters. A named parameter can only have a private name in a context where it
+is _useful_ to do so because it corresponds to a private instance field. For all
+other named parameters it is still a compile-time error to have a private name.*
```
lrhn: This feels like it conflates the orthogonal concerns of being initializing/formal and being positional/named.
The first sentence is about the former, without referring to the latter.
The second sentence introduces named parameters as the *reason* for the restriction to initializing formals and field parameters, and sounds like it talks only about named parameters. 

That leaves me wondering what the reasoning is for positional parameters, and if they're even allowed.

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -0,0 +1,382 @@
...
+
+## Runtime semantics
+
+There are no runtime semantics for this feature. It's purely a compile-time
+renaming.
```
lrhn, line 335: Nit-pick: The semantics in the "binding actuals to formals" algorithm needs to know about this, and know to bind the argument named *n* to the variable named *p*.

I know that names are (or should be) compiled away, but the language semantics doesn't do that, so there are runtime *semantics*, even if there might be no runtime *implementation*.

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -0,0 +1,382 @@
...
+
+## Compatibility
+
+This proposal takes code that it is currently a compile-time error (a private
+named parameter) and makes it valid in some circumstances (when the named
+parameter is an initializing formal or field parameter). Since it simply expands
```
lrhn, line 341: So it doesn't work for positional parameters?

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -0,0 +1,382 @@
...
+They may wish to implement all, some, or none of this, and will likely have
+further ideas for additional warnings, lints, and quick fixes.
+
+### API documentation generation
+
+Docs generated using [`dart doc`][dartdoc] or other documentation generators
+should document any private named parameters using their public name. The fact
+that the parameter initializes or declares a private field is an implementation
+detail of the class. What a user of the class cares about is the corresponding
+public name for the constructor parameter.
```
lrhn: More importantly: *Authors* should document the public name. You write `// The [foo] is ...` to document `this._foo`.

The generated documentation should follow suit.
The tools may support `[_foo]` as well, and display it as <u>`foo`</u>, but the important thing is what authors should write in the doc source.

lrhn (approved):

working/2509-private-named-parameters/feature-specification.md
munificent: Fixed locally, thanks!

working/2509-private-named-parameters/feature-specification.md
munificent: Heh, oops!

working/2509-private-named-parameters/feature-specification.md
munificent: Yeah, it mentions it, but I agree it would be good to clarify for readers who don't know that proposal. Added a little bit at the beginning.

working/2509-private-named-parameters/feature-specification.md
munificent: Oh, good catch. Done.

working/2509-private-named-parameters/feature-specification.md
munificent: Oops, this was supposed to be restricted to only affect the behavior of named parameters. Positional parameters should behave as they currently do where they can be private or not. Fixed.

working/2509-private-named-parameters/feature-specification.md
munificent: Reworded. The intent was that this only applies to named initializing formals and field parameters.

working/2509-private-named-parameters/feature-specification.md
munificent: Uh... fair, but I'll just leave this as-is. :)

working/2509-private-named-parameters/feature-specification.md
munificent: Positional parameters can already have private names and are unaffected by the proposal. I accidentally dropped that part and made it sound like it applies to all initializing formals and field parameters when I was doing some wordsmithing, but it's fixed now.

working/2509-private-named-parameters/feature-specification.md
lrhn: I'd actually *prefer* if positional parameters got the same treatment, just to make it explicit that
```dart
String value = "Something else";
class Box {
   final int _value;
   /// Creates a box with the given [value].
   C(this._value);
}
```
means that [value] definitely denotes the parameter.

We can make hacks in DartDoc to show public identifiers instead of private ones, so you write `the given [_value]` and it's shown as "The given <ul>value</ul>", but it's bound to cause headaches. Like here, where `value` actually denotes something else in the language.

working/2509-private-named-parameters/feature-specification.md
lrhn: (And that is why I want this for positional parameters too.)

PR #4424, opened 2025-06-26
Add missing dart code block

kevmoo:
Makes reading more pretty – not sure if it's worth a version bump 🤷


munificent:
Thanks! Definitely not worth a version bump. (Version bumps in language features are mainly to get the attention of implementers who need to know what the spec has changed under them. But this feature is in-progress and not even accepted yet, so it's still basically mutable.)

Issue #4382, opened 2025-05-16
Flow analysis layers try/finally backwards in some circumstances


stereotype441:
A tricky part of the implementation of flow analysis is the handling of try/finally statements. Although promotions are tracked separately in the `try` and `finally` blocks, promotions from both blocks need to be merged together at the conclusion of the finally block. This creates an ambiguity, because each type in a promotion chain is required to be a subtype of the previous, and hence multiple promotions of the same  variable are inherently ordered. The ambiguity is: when the promotions from the `try` and `finally` block are merged, which promotions should be applied first?

It seems pretty clear to me that the promotions from the `try` block should be applied first, because that matches the order of code execution.

But that isn't what happens. For example:
```dart
f(bool b, Object x, Object y) {
  if (b) {
    try {
      x as num;
      y as int;
    } finally {
      // Neither `x` nor `y` is promoted at this point, because in principle
      // an exception could have occurred an any point in the `try` block.
      x as int;
      y as num;
    }
  } else {
    x as num;
    y as num;
  }
  x.abs();
  y.abs();
}
```

I would expect that after the try/finally statement, the promotions on `x` should be applied in the order `num`, then `int`, resulting in the promotion chain `[num, int]`. Therefore, after the `if` statement, the promotion chains `[num, int]` and `[num]` should be joined, resulting in the promotion chain `[num]`. So `x` should remain promoted to `num`, and thus `x.abs()` should be accepted.

As for `y`, I would expect that after the try/finally statement, _its_ promotions should be applied in the order `int`, then `num`, resulting in the promotion chain `[int]` (the `num` promotion should have no effect, since `num` is not a subtype of `int`). Therefore, after the `if` statement, the promotion chains `[int]` and `[num]` should be joined, resulting in an empty promotion chain (since the two chains have no types in common). So `y` should be fully demoted at this point, and thus `y.abs()` should be an error.

But in fact the promotions are applied in the opposite order, so `y.abs()` is accepted and `x.abs()` is an error.

A similar thing happens for field promotions:
```dart
class C {
  final Object _f;
  C(this._f);
}

f(bool b, C x, C y) {
  if (b) {
    try {
      x._f as num;
      y._f as int;
    } finally {
      x._f as int;
      y._f as num;
    }
  } else {
    x._f as num;
    y._f as num;
  }
  x._f.abs();
  y._f.abs();
}
```

[...]

PR #4427, opened 2025-06-26
[flow analysis] Rework control flow statements and expressions.

stereotype441:
This change adjusts the specification of flow analysis for the following kinds of statements and expressions, to match what was actually implemented:

- conditional expressions
- if-null expressions
- logical and/or expressions
- null check operators
- break statements
- continue statements
- return statements
- if statements
- while statements
- for statements
- do while statements
- for each statements
- switch statements
- try statements

Also, the paragraphs describing "Local variable conditional assignment" and "Conditional assignment to a non local-variable" are dropped. These paragraphs were quite inaccurate, and the necessary infrastructure isn't in place yet to replace them with more accurate descriptions.

Finally, there's a minor fix: the definition of a promotion chain was backwards (types are listed in least-promoted to most-promoted order).

---

<details>

</details>

### DIFF
--- a/resources/type-system/flow-analysis.md
+++ b/resources/type-system/flow-analysis.md
@@ -288,8 +288,8 @@ We also make use of the following auxiliary functions:
       `VI2` to `joinV(VI1(v), VI2(v))`.  Note that any variable which is in
       domain of only one of the two is dropped, since it is no longer in scope.
 
-  The `join` and `joinV` combinators are commutative and associative by
-  construction.
+  _We expect the `join` and `joinV` combinators to be commutative and
+  associative, but we don't rely on this for the soundness of the algorithm._
 
   For brevity, we will sometimes extend `join` to more than two arguments in the
   obvious way.  For example, `join(M1, M2, M3)` represents `join(join(M1, M2),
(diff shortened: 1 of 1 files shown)


eernstg:
Oh, I didn't get around to look at this one. I should be able to make progress on it on Monday.

resources/type-system/flow-analysis.md
```diff
@@ -270,95 +262,105 @@ We also make use of the following auxiliary functions:
...
+    - `M2 = FlowModel(r2, VI2)`
     - `pop(r1) = pop(r2) = r0` for some `r0`
     - `r3` is `push(r0, top(r1) || top(r2))`
     - `VI3` is the map which maps each variable `v` in the domain of `VI1` and
       `VI2` to `joinV(VI1(v), VI2(v))`.  Note that any variable which is in
       domain of only one of the two is dropped, since it is no longer in scope.
 
-  The `merge`, `join` and `joinV` combinators are commutative and associative by
+  The `join` and `joinV` combinators are commutative and associative by
   construction.
```
eernstg: Perhaps this could be:
```suggestion
  The `join` and `joinV` combinators are commutative and associative.
```
I think 'by construction' is commonly used to indicate that the given properties are satisfied because it is essentially written into the definition (that is, we look at the definition and then it's obvious that each of these properties holds). I think we have these properties for `join` and `joinV` because of a slightly more involved chain of reasoning. We basically have to look at a handful of terms (like `top(r1) || top(r2)`), and then we can mentally outline a proof that, for example, commutativity holds. I think it makes sense to describe this situation in those more neutral words, expecting a reader to re-do this small proof as needed.

resources/type-system/flow-analysis.md
```diff
@@ -270,95 +262,105 @@ We also make use of the following auxiliary functions:
...
-    - `MF = FlowModel(rf, VIF))`
-    - `pop(rb) = pop(rf) = r0` for some `r0`
-    - `r3` is `push(r0, top(rb) && top(rf))`
-    - `b` is true if `v` is in `N` and otherwise false
-    - `VI3` is the map which maps each variable `v` in the domain of `VIB` and
-      `VIF` to `restrictV(VIB(v), VIF(v), b)`.
+  For brevity, we will sometimes extend `join` to more than two arguments in the
+  obvious way.  For example, `join(M1, M2, M3)` represents `join(join(M1, M2),
+  M3)`, and `join(S)`, where S is a set of models, denotes the result of folding
+  all models in S together using `join`.
```
eernstg, line 297: OK, so why would it be obvious that `join(M1, join(M2, M3))` is equal to `join(join(M1, M2), M3)`? I think it is true, but I also think that it would take a bit of proof tinkering to make it explicit in detail why it is true. ;-)

Do we actually need those properties? Are they just "nice to have"?

resources/type-system/flow-analysis.md
```diff
@@ -270,95 +262,105 @@ We also make use of the following auxiliary functions:
...
+- `rebasePromotedTypes(p1, p2)`, where `p1` and `p2` are promotion chains,
+  computes a promotion chain containing promotions from both `p1` and `p2`,
+  discarding promotions if necessary to avoid violating the promotion chain
+  requirement. It is defined as `p3`, where:
+  - If `p2` is empty, then `p3 = p1`.
+  - Otherwise:
+    - Let `T2` be the last type in `p2`.
+    - Let `p1'` be a promotion chain obtained by deleting any elements `T` from
+      `p1` that do not satisfy `T <: T2`.
+    - Let `p3 = [...p2, ...p1']`.
```
eernstg: It looks like the purpose of this operation is to compute the longest possible type-decreasing list that begins with `p2` and is followed by everything in `p1` "that fits".

This is clearly asymmetric. Is that an inherent property of the places where this operation is used, or would it have been equally valid to, say, use `[...p2', ...p1]` where we're filtering `p2` to prevent the "decreasing type" requirement from being violated?

resources/type-system/flow-analysis.md
```diff
@@ -270,95 +262,105 @@ We also make use of the following auxiliary functions:
...
+
+- `attachFinallyV(VM1, VM2, VM3)`, where `VM1`, `VM2`, and `VM3` are variable
+  models, represents the state of a variable model after a `try/finally`
+  statement, where `VM1` is the state after the `try` block, `VM2` is the state
+  before the `finally` block, and `VM3` is the state after the `finally`
```
eernstg: And the state before the `finally` block is the same as the state before the entire `try/finally` statement (because the `try` block can throw at any point)?

Also, the state after the `finally` block, does that mean at the end of the finally block? Presumably the whole point of this operation is that it computes the stat after the `try/finally` statement as a whole, which is also right after the `finally` block?

Perhaps add a bit of commentary text here, to help readers.

resources/type-system/flow-analysis.md
```diff
@@ -270,95 +262,105 @@ We also make use of the following auxiliary functions:
...
+  - Let `p4` be determined as follows:
+    - If the variable's value might have been changed by the `finally` block
+      (_TODO(paulberry): specify precisely how this is determined_), then `p4 =
+      p3`. _Promotions from the `try` block aren't necessarily valid, so only
+      promotions from the `finally` block are kept._
+    - Otherwise, `p4 = rebasePromotedTypes(p3, p1)`.
+  - Let `t4 = t3`. _The `finally` block inherited all tests from the `try`
+    block, so `t3` contains all relevant tested types._
+  - Let `a4 = a1 || a3`. _The variable is definitely assigned if it was
+    definitely assigned in either the `try` or the `finally` block._
```
eernstg: Is that 'definitely assigned at the end of the `try` block'? (So we're using the fact that the `try` block did not throw.)

eernstg (approved):
Looks great! I added a couple of questions here and there. For the last part, I couldn't come up with the rules from scratch, but I'm relying on the assumption that this basically specifies the implemented behavior which is already known to handle a huge set of cases correctly.

resources/type-system/flow-analysis.md
stereotype441: Hmm, good question. I've tried to be explicit everywhere about the order and associativity in which `join` is applied, and in each case I've tried to verify that the spec matches the implementation. So I guess they're just "nice to have".

I'll make a note to talk at our meeting in a few minutes about whether to make further changes to the spec text to account for this.

resources/type-system/flow-analysis.md
stereotype441: It's an inherent property of the places where this operation is used, and I don't think there's any way around that.

Imagine, for example, that `p1` is `[num?, int]` and `p2` is `[Object, num]`. Under my definition, `p3 = [...p2, ...p1']`, `p3` is `[Object, num, int]`. But if we instead defined `p3 = [...p2', ...p1]` (where `T1` is the first type in `p1` and `p2'` is the promotion chain obtained by deleting any elements `T` from `p2` that do not satisfy `T1 <: T`), then `p3` would be `[num?, int]`.

In all the places that I've used `rebasePromotedTypes` I've tried to be careful about the order of the arguments so that the user-visible behavior is sensible. See, for instance, https://github.com/dart-lang/language/issues/4382, where I fixed a place where the order was backwards compared to what I think would be a reasonable user expectation.

resources/type-system/flow-analysis.md
stereotype441: > And the state before the `finally` block is the same as the state before the entire `try/finally` statement (because the `try` block can throw at any point)?

Almost but not quite:
- If there is a local variable that's promoted prior to the `try/finally` statement, and it's assigned in the `try` block, then in the state before the `finally` block, the promotion is removed.
- If there is a local variable that's in the "definitely unassigned" state prior to the `try/finally` statement, and it's assigned in the `try` block, then in the state before the `finally` block, it's in the "neither definitely assigned nor definitely unassigned" state.
- If there is a local variable that's not write captured prior to the `try/finally` statement, and it's write captured somewhere within the `try` block, then in the state before the `finally` block, the variable is considered write captured.

All of these are handled by the use of `conservativeJoin` in this text from the **try finally** bullet in the "Statements" section:

- Let `before(B2) = join(after(B1), conservativeJoin(before(N), assignedIn(B1), capturedIn(B1)))`.

> Also, the state after the `finally` block, does that mean at the end of the finally block? Presumably the whole point of this operation is that it computes the stat after the `try/finally` statement as a whole, which is also right after the `finally` block?
> 
> Perhaps add a bit of commentary text here, to help readers.

Ok, I will add commentary text, in line with what we just discussed in our meeting.

resources/type-system/flow-analysis.md
stereotype441: In our meeting (way back on June 30), we decided to make this text non-normative, and clarify that these definitions are intended to be commutative and associatev, but that these properties aren't relied upon for soundness. I've made that change.

resources/type-system/flow-analysis.md
stereotype441: Upon further reflection, I think a lot of the confusion here was because the `attachFinallyV` function was described first, whereas the explanation of the meaning of the various arguments was in the definition of `attachFinally`, which followed. I've reversed the order of the two to make things clearer. Also, as we discussed in our meeting, I've renamed the arguments to `attachFinally` and `attachFinallyV` so that their meaning is clear from their names.

resources/type-system/flow-analysis.md
stereotype441: Correct.

stereotype441:
Returning to this arc of work after about a month of working on other tasks. When I met with Erik on June 30, we agreed on a set of changes, and decided that I should just make those changes and land the PR. I've made the changes, so I'm landing it now.

resources/type-system/flow-analysis.md
eernstg: Thanks!

PR #4428, opened 2025-06-27
Refresh the primary constructor proposal to use new syntax

eernstg:
This PR adjusts the primary constructor proposal to use the approach where the in-header form is considered to be syntactic sugar for an in-body form ("moving the parameter list to the header"). In particular, this allows even the in-header constructor to have all features (initializer list, superinitializer, body), declared in the body.

It does not introduce abbreviations of non-primary constructor declarations (like `const();` as a short form for `const MyClass();`). That's a separate feature which can be handled elsewhere.

The description of the parameter list as "being moved to the header" implies that the parameter list will have the same semantics when declared in-body and in-header. In particular, the in-header form _does_ require an explicit token for a parameter to be declaring (`var` or `final`), and it _does_ support non-declaring parameters even in the header.

The `const` modifier can remain in the body, such that primary constructors are as similar to other constructors as they can be. It can also be specified in the header, such that tiny data classes can have a constant constructor without needing a body.

This PR only updates the introduction and the grammar, the remaining parts will be updated in a separate PR.

### DIFF
--- a/working/2364 - primary constructors/feature-specification.md
+++ b/working/2364 - primary constructors/feature-specification.md
@@ -503,7 +503,7 @@ constructors as well.
    | 'this' ('.' <identifierOrNew>);
 
 <constantConstructorSignature> ::= // Modified rule.
-     'const' <constructorName> <declaringParameterList>;
+     'const' <constructorSignature>;
 
 <constructorName> ::= // Modified rule.
      (<typeIdentifier> | 'this') ('.' <identifierOrNew>)?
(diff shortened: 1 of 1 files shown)


eernstg:
Hello @munificent! I took the new syntactic ideas and put them into the primary constructor proposal. Could you take a look? I haven't changed anything after the grammar section because I wanted to discuss the syntax first.

eernstg:
There is an updated version of the spec parser here: https://dart-review.googlesource.com/c/sdk/+/437480.

PR #4438, opened 2025-07-04
Update the primary constructors proposal to give the body access to header parameters

eernstg:
This PR updates the primary constructors proposal to include support for access to header parameters in non-late instance variable initializers, and in general updates the static analysis and semantics to the new syntax introduced by #4428.

The approach relies on a very simple change to the scoping structure: The primary header parameters are added to a new scope which is the enclosing scope of the class body scope. This is completely standard according to the syntactic structure, just like the scoping around a function body and its formal parameter declarations.

This means that header parameters with a fresh name (that is, not clashing with the names declared in the class body) are in scope in the class body. It is an error to refer to them in any other location than an initializing expression of a non-late instance variable.

For the case where a name `n` in an initializing expression of a non-late instance variable is resolved to an instance variable declared by the class which has a corresponding primary header parameter (which is necessarily a declaring parameter with name `n`, or an initializing formal parameter with name `n`), it evaluates to the value of the parameter.

The difference between this approach and an approach where the primary header parameters are simply added to the scope of said initializing expressions is that we avoid allowing the given name to refer to two completely unrelated entities:

```dart
class A(final int x, int y) {
  String y;
  int z = x + y;
}
```

If we simply specify that the parameter `int y` is in scope for the initializing expression `x + y` then it all compiles and "works", but it is highly confusing that `y` does not refer to the instance variable named `y`, it refers to the parameter, and there is no relationship whatsoever between the parameter `y` and the instance variable `y`, they're just separate entities.

With this proposal it is a compile-time error for the initializing expression of `z` to refer to `y`: It can't refer to the instance variable because there is no access to `this`, and it can't use the "backup rule" that turns the evaluation of the instance variable into a parameter read (because that parameter has no relation to that instance variable). The point is that this error is useful because the situation is error prone.

In summary, the approach taken in this PR is to only allow an identifier expression like `y` to refer to a parameter if it resolves to an instance variable (according to the normal scope rules), and there is a _corresponding_ primary header parameter. If there is no parameter with the given name, or there is a parameter but it doesn't correspond to the given instance variable, it's an error.

### DIFF
--- a/working/2364 - primary constructors/feature-specification.md
+++ b/working/2364 - primary constructors/feature-specification.md
@@ -1,17 +1,17 @@
-# Primary Constructors
+# Declaring and Primary Constructors
 
 Author: Erik Ernst
 
 Status: Draft
 
-Version: 1.6
+Version: 1.8
 
-Experiment flag: primary-constructors
+Experiment flag: declaring-constructors
 
-This document specifies _primary constructors_. This is a feature that
+This document specifies _declaring constructors_. This is a feature that
 allows one constructor and a set of instance variables to be specified in a
 concise form in the header of the declaration, or in the body. In the case
-where the constructor is specified in the header, some elements are still
+where the constructor is specified in the header, some elements can still be
 specified in the class body, if present: The in-header constructor can have
 an initializer list in the body, including assertions, instance variable
 initializers, and/or a superinitializer. The in-header constructor can also
@@ -27,7 +27,8 @@ feature have taken place in github issues marked with the
 
 Recently, [Bob proposed][] that primary body constructors should use the syntax
 `this.name(...)` rather than `primary C.name(...)`. This proposal includes
-that choice.
+that choice. Several recent updates to this proposal are based on ideas from
+that proposal.
 
 [struct proposal]: https://github.com/dart-lang/language/blob/master/working/extension_structs/overview.md
 [kotlin primary constructors]: https://kotlinlang.org/docs/classes.html#constructors
@@ -38,7 +39,7 @@ that choice.
 
 ## Introduction
 
-Primary constructors is a conciseness feature. It does not provide any new
+Declaring constructors is a conciseness feature. It does not provide any new
 semantics at all. It just allows us to express something which is already
 possible in Dart, using a less verbose notation. Consider this sample class
 with two fields and a constructor:
@@ -52,34 +53,38 @@ class Point {
 }
 ```
 
-A primary constructor in the header allows us to define the same class much
+A declaring constructor in the header allows us to define the same class much
 more concisely:
 
 ```dart
-// A declaration with the same meaning, using a primary header constructor.
+// A declaration with the same meaning, using a declaring header constructor.
 class Point(var int x, var int y);
 ```
 
-A class that has a primary header constructor can not have any other
+A class that has a declaring header constructor can not have any other
 generative non-redirecting constructors. This requirement must be upheld
-because it must be guaranteed that the primary header constructor is
+because it must be guaranteed that the declaring header constructor is
 actually executed on every newly created instance of this class. This rule
 is further motivated below.
 
-A primary body constructor is slightly less concise, but it allows the
+A declaring header constructor is also known as a _primary constructor_, 
+because all other generative constructors must invoke the primary one
+(directly or indirectly).
+
+A declaring body constructor is slightly less concise, but it allows the
 class header to remain simpler and more readable when there are many
-parameters. The previous example would look as follows using a primary body
-constructor:
+parameters. The previous example would look as follows using a declaring
+body constructor:
 
 ```dart
-// A declaration with the same meaning, using a primary body constructor.
+// A declaration with the same meaning, using a declaring body constructor.
 class Point {
   this(var int x, var int y);
 }
 ```
 
 In the examples below we show the current syntax directly followed by a
-declaration using a primary constructor. The meaning of the two (or more)
+declaration using a declaring constructor. The meaning of the two (or more)
 class declarations with the same name is always the same. Of course, we
 would have a name clash if we actually put those two declarations into the
 same library, so we should read the examples as "you can write this _or_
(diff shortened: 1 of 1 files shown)


eernstg:
@mraleph, how do you like this one? ;-)

working/2364 - primary constructors/feature-specification.md
```diff
@@ -60,6 +60,12 @@ more concisely:
 class Point(var int x, var int y);
 ```
 
+A class that has a primary header constructor can not have any other
```
munificent: Somewhat idle thought: With an in-body primary constructor, there may be other generative constructors that don't redirect to it, so it's not really "primary" in the same sense that it is in Scala and Kotlin. That suggests that "primary" isn't a good name for these.

With an in-header primary constructor... there are no other kinds of constructors you can put in the header. That suggests that either or "in-header" or "primary" is redundant. In this case, the constructor *is* primary: all other generative constructors must redirect to it.

I wonder if we should rename things to "declaring constructor" for an in-body primary and "primary constructor" for an in-header primary?

Alternatively, we could say that if a class has an in-body primary constructor then all other generative constructors must redirect to it. That restriction isn't strictly necessary for correctness, but might make it easier for users to understand how instances of the class get initialized. In that case, the current terminology makes sense.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -108,22 +114,25 @@ just after the class name specifies both a constructor declaration and a
...
+style whereby formal parameters are never modified will have a 
+[lint][parameter_assignments] to flag all such mutations.
+
+[parameter_assignments]: https://dart.dev/tools/linter-rules/parameter_assignmentshttps://dart.dev/tools/linter-rules/parameter_assignments
 
-Similarly, with this feature a formal parameter can not be declared with
-the syntax `var name`, it must have a type (`T name`) or the type must be
-omitted (`name`).
+Similarly, with this feature a regular (non-declaring) formal parameter can
+not be declared with the syntax `var name`, it must have a type (`T name`)
```
munificent: Nit: "be declared with" -> "use".

It's a little confusing to have "declaring" and "declared" in the same sentence while referring to different concepts.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -382,6 +388,46 @@ avoiding the duplication of instance variable names and types) even in the
 case where some parameters should not introduce any instance variables (so
 they are just "normal" parameters).
 
+With a primary header constructor, the formal parameters in the header are
+introduced into a new scope. This means that the parameters whose name is
+not introduced by any nested scope (e.g., the class body scope) is in scope
```
munificent, line 398: "is" -> "are" in both places. Or "the parameters" -> "a parameter".

working/2364 - primary constructors/feature-specification.md
```diff
@@ -609,17 +680,59 @@ Object {}`, and all three of them have `Object` as their direct superclass.*
...
+
+*This means that initializing expressions can, apparently, use the value of
+instance variables declared by the same class (not, e.g., inherited ones).
+They will actually get the value of the primary header parameter, but this
+value is also guaranteed to be the initial value of the corresponding
+instance variable.*
+
+*Note that it only applies to identifier expressions. In particular, this
+does not allow initializing expressions to assign to other instance
+variables.*
```
munificent, line 767: It would be good to spell out how this behaves with respect to capture:

```dart
class C(var int x) {
  var closure = () => x;
}

main() {
  var c = C(1);
  c.x = 2;
  print(c.closure()); // Prints 1.
}
```

working/2364 - primary constructors/feature-specification.md
```diff
@@ -675,7 +790,8 @@ of the fact that the primary constructor is actually placed outside the
 braces that delimit the class body.*
 
 Next, _k_ has the modifier `const` iff the keyword `const` occurs just
-before the name of _D_ or before `this`, or _D_ is an `enum` declaration.
+before the name of _D_ or before `this`, or if _D_ is an `enum`
+declaration.
```
munificent, line 830: So we infer `const` for header constructors on enums, but not body constructors? That seems like an irregularity.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -731,15 +846,81 @@ means that they preserve their semantics when moved into the body.*
...
 Finally, _k_ is added to _D2_, and _D_ is replaced by _D2_.
 
+### Warnings
+
+The language does not specify warnings, but the following is recommended:
+
+A warning is emitted in the case where an identifier expression is resolved
+to yield the value of a declaring or initializing formal parameter in a
+primary header constructor, and this identifier occurs in a function
+literal, and the corresponding instance variable is non-final.
```
munificent: An example would help clarify here. :)

working/2364 - primary constructors/feature-specification.md
```diff
@@ -731,15 +846,81 @@ means that they preserve their semantics when moved into the body.*
 
 Finally, _k_ is added to _D2_, and _D_ is replaced by _D2_.
 
+### Warnings
+
+The language does not specify warnings, but the following is recommended:
+
+A warning is emitted in the case where an identifier expression is resolved
```
munificent: I think we should just make this an error. I can't imagine it every doing anything *useful* and will certainly just confuse users.

munificent (approved):
This looks great!

working/2364 - primary constructors/feature-specification.md
eernstg: 'Declaring' makes more sense, overall. Renamed!

I kept the terminology 'primary constructor' as a additional phrase describing a declaring constructor in the header, and motivated the need for a special phrase by the fact that it must be invoked by all other generative constructors, and it changes the scoping of the entire class declaration.

working/2364 - primary constructors/feature-specification.md
eernstg: Done.

working/2364 - primary constructors/feature-specification.md
eernstg: Done; I put the example into the section where it is recommended that this situation causes a warning.

working/2364 - primary constructors/feature-specification.md
eernstg: Hmmm ... that makes sense. I changed the compile-time error in the 'Syntax' section such that the header part and the body part must agree, if we have a primary constructor that includes both parts.

working/2364 - primary constructors/feature-specification.md
eernstg: Done.

working/2364 - primary constructors/feature-specification.md
eernstg: Done!

working/2364 - primary constructors/feature-specification.md
eernstg: Done.

PR #4454, opened 2025-07-08
Clarify the `void` allowlist wrt `[]` and `[]=`

eernstg:
This PR adds a clarification to the `void` allowlist, specifying explicitly that `e1` can have type `void` in the case where `e0[e1]` is encountered and the statically known operator `[]` of `e1` has a parameter whose type is `void`, and similarly for `e0[e1] = e2` where both `e1` and `e2` can have type `void` when the corresponding formal parameters have type `void`.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -41,6 +41,10 @@
 % version of the language which will actually be specified by the next stable
 % release of this document.
 %
+% Jul 2025
+% - Clarify that operator `[]` and `[]=` are included in the `void` allowlist
+%   rule about formal parameters of type `void`.
+%
 % May 2025
 % - Add `s?.length` as a constant expression in the case where `s` satisfies
 %   some requirements.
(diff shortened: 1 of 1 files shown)


specification/dartLangSpec.tex
```diff
@@ -22882,6 +22886,13 @@ \subsection{Type Void}
...
+  This rule also applies to the index operators \code{[]} and \code{[]=}.
+  \commentary{%
+    For example, $e_1$ and $e_2$ may have type \VOID{}
+    in an expression of the form \code{$e_0$[$e_1$] = $e_2$}
+    when the parameters of the statically known operator \code{[]=}
+    both have type \VOID{}.%
+  }
 \item
   In an expression of the form \code{$e_1$\,=\,$e_2$}
   where $e_1$ is an \synt{assignableExpression}
```
lrhn: Does the "or formal parameter" here cover setters?

It seems redundant to say `variable or formal parameter` when a formal parameter is a variable.
But if it means to denote the formal parameter of a setter being invoked, then it's ambiguous with the redundant interpretation, `foo([void x]) { x = print("ok"); }` also has `x` denoting a formal parameter.

(Or should the section above also cover, and mention that it covers, setters, like it mentions index operators.)

eernstg: I changed the previous item to deal with setter invocations as well as other function invocations. I changed this item to be slightly more explicit on the fact that it is only about assignments to local variables. I think it makes sense that local variable assignments are singled out because it is really different from all the other cases where we invoke a setter or any other function.

PR #4468, opened 2025-07-18
Add a "Concerns" section to "Private Named Parameters".

munificent:
Talk a bit about how the feature might be confusing at first and suggest ways we can mitigate that.

cc @bwilkerson for thoughts on error messages with this feature.

### DIFF
--- a/working/2509-private-named-parameters/feature-specification.md
+++ b/working/2509-private-named-parameters/feature-specification.md
@@ -321,7 +321,7 @@ start with underscore, only a named parameter that declares or initializes a
 private instance field.
 
 A private named parameter *looks weird* since privacy makes little sense for an
-argument name and makes even less since for the local parameter variable. (We
+argument name and makes even less sense for the local parameter variable. (We
 already have a lint that warns on using `_` for local variable names since it
 accomplishes nothing.)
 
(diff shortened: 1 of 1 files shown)


working/2509-private-named-parameters/feature-specification.md
```diff
@@ -268,6 +275,79 @@ class House({
...
+
+We mitigate confusion here in a couple of ways:
+
+#### Only allow the syntax where it's meaningful
+
+At the language level, this proposal only allows `_` in a named parameter when
+doing so is *useful and meaningful*. It doesn't allow *any* named parameter to
+start with underscore, only a named parameter that declares or initializes a
+private instance field.
+
```
leafpetersen, line 322: less since->less sense

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -268,6 +275,79 @@ class House({
...
+argument name and makes even less since for the local parameter variable. (We
+already have a lint that warns on using `_` for local variable names since it
+accomplishes nothing.)
+
+If a user sees `_` on a parameter and is trying to figure out what's going on,
+they will reliably be in a context where that parameter is also referring to a
+field. If we're lucky, that may lead them to intuit that the privacy is for the
+field, not the parameter.
+
+#### Provide a teaching errors if they use `_` for other named parameters
```
leafpetersen: "a teaching error"?  Or "teaching errors"?

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -268,6 +275,79 @@ class House({
...
+
+If a user tries to put `_` before a named parameter that *isn't* an initializing
```
leafpetersen, line 335: FWIW, I don't expect the extra effort to implement this "teaching error" to carry its weight.  What is the user trying to accomplish by doing this?

leafpetersen (approved):
LGTM, thanks.

working/2509-private-named-parameters/feature-specification.md
munificent: We have to report *some* error here because the proposal disallows it. I don't think it's any significant extra effort for that error to clarify that you can use `_` for constructor parameters that are initializing or declaring but not otherwise.

I agree it would be weird for users to try to do this, but my impression from reading lots of pub code over the years is that at scale, all possible user behavior will be observed no matter how irrational. :)

PR #4469, opened 2025-07-21
Move "Parts with Imports" to a separate proposal.

### LINKED ISSUE #3939: Strategy for launching augmentations
A small pile of questions to resolve here, that are important in considering macros timelines.

1) There is a possibility that we launch a single macro, e.g. the JSON macro, without opening up the macro feature for general use. If we do that, what state does the augmentation feature need to be in? Is it okay to ship like that if e.g. augmentation feature spec is still in progress? Implementation? Tests? Do we need to split out a distinct feature?

2) There is a possibility that we launch a less powerful macro feature, such as the "phase 3 macros" recently under discussion. If we do that, what state does the augmentation feature need to be in? Do we need to split out a distinct feature?

3) There is an argument for launching augmentations _before_ macros to reduce the size of the macros launch. Do we want that? Should we anyway do the work to separate out augmentations as a distinct feature--currently it's the "macros" experiment flag.

Discussing this probably fits naturally into macro timeline discussions, I'm filing mostly so it shows up as a blocker for having phase 3 macros ready for launch.

munificent:
In #3939, there seemed to be consensus that "parts with imports" should be its own proposal with its own experiment flag. This PR does that. While I was at it, I renamed "Augmentation libraries" to just "Augmentations" (since it's not a new kind of library anymore) and gave that an experiment flag too.

There are no changes to the text of either proposal. (Well, I did get rid of smart quotes in the parts with imports one...)

Fix #3939.

cc @dart-lang/language-team in case anyone disagrees with this move.


munificent:
Friendly ping. :)

natebosch (approved):

PR #4475, opened 2025-07-21
Revise "Private Named Parameters".

### LINKED ISSUE #4471: Private named parameters: feature spec motivation issues
I don't understand the intro to the motivation section: I think it is accurately describing a problem which exists **when you have this feature** but it is describing it as a problem with the current language, which I don't understand.  

> Local variables and parameters aren't in scope outside of the library where they are defined, so privacy doesn't come into > play. Except, that is, for named parameters. A named parameter has one foot on each side of the function boundary.

There's nothing unique to named parameters about this.  This is simply what it means to be public API.  All class members (and classes for that matter) have "one foot on each side of the boundary" in this manner (that is, they are both an external reference, and an internal reference).  

> A public function containing a named parameter whose name is private raises difficult questions.

No more so than a public class containing a field whose name is private.  It is true that you must decide on what the meaning of that is: we have done so.  Now we propose to change it.

In general, I don't really disagree with anything being said in this section, but the framing is really confusing, because it frames this as something odd and different about named parameters in existing Dart, which I don't see at all and which I think distracts from the core direction of the discussion.

I would suggest rephrasing this away from the "named parameters present a puzzle" framing, to one of just "this is how things work now", e.g:

---

Dart uses a leading underscore in an identifier to make a declaration private to
its library. Privacy is only meaningful for declarations that could be accessed
from outside of the library: top-level declarations and members on types.

Local variables and parameters aren't in scope outside of the library where they
are defined, but named parameters can be referenced from outside of the library.  This raises 
the question of how to interpret a private named parameter:

```dart
test({String? _hmm}) {
  print(_hmm);
}

main() {
  test(_hmm: 'ok?');
}
```

Should this be allowed?  And if so, do we treat this as a named parameter which 
can only be called from within the defining library?

The language currently resolves this by saying that it is a
compile-time error to have a named parameter with a private name. Users must use
a public name instead. For most named parameters, this restriction is harmless.
The parameter is only used within the body of the function and it is idiomatic for
local variables to not have private names anyway.

munificent:
I really like the text you proposed on the issue, so here it is. :)

Fix #4471.

### DIFF
--- a/working/2509-private-named-parameters/feature-specification.md
+++ b/working/2509-private-named-parameters/feature-specification.md
@@ -4,7 +4,7 @@ Author: Bob Nystrom
 
 Status: In-progress
 
-Version 0.1 (see [CHANGELOG](#CHANGELOG) at end)
+Version 0.2 (see [CHANGELOG](#CHANGELOG) at end)
 
 Experiment flag: private-named-parameters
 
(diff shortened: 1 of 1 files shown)


working/2509-private-named-parameters/feature-specification.md
```diff
@@ -349,6 +343,34 @@ corresponding field private and that the argument should be the public name. The
...
 teach them the feature.
 
+### Super parameters
+
+We allow private named parameters for initializing formals and (assuming primary
+constructors exist), declaring parameters. What about the other special kind of
+constructor parameter, super parameters (the `super.` syntax)?
+
+Those are unaffected by this proposal. A super parameter generates an implicit
+argument that forwards to a superclass constructor. The argument name is always
```
leafpetersen: It's a bit hard to keep track of which of the two different sets of arguments/parameters are referenced here, so maybe replace "The argument name" with "The super constructor argument name is always public, ..."?

leafpetersen (approved):
LGTM, thanks!

PR #4478, opened 2025-07-23
Clarify some vague text in "Private Named Parameters".

### LINKED ISSUE #4472: Private named parameters: text suggests they're allowed everywhere
The private named parameters proposal says:

```
The basic idea is simple. We let users use a private name in a named parameter. 
```

This more than suggests that the proposal is to allow arbitrary named parameters to have private names, which it is my understanding is explicitly not the proposal.  Maybe:

```
The basic idea is simple. We let users use a private name in a named initializing formal parameter. 
```

?

### LINKED ISSUE #4473: Private named parameters: clarify "name of the parameter"
The private named parameters proposal, in the static semantics section, says:

```
Otherwise, the name of the parameter in C is n. If the parameter is named, this then avoids the compile-time error that would otherwise be reported for a private named parameter.

The name of the local variable in the initializer list scope of C is p. In the initializer list, the private name is used. Inside the body of the constructor, uses of p refer to the field, not the parameter.
```

This first sentence is confusing, since it would normally taken to imply that references to the parameter in question within the scope of the parameter should use the name `n`, which is contradicted by the subsequent bullet.  The intent here (I think) is that the *external name* of the parameter should be taken to `n` for the purposes of parameter passing (including dynamic calls and calls via `super.` parameters), computing the static type of a tear off of the constructor, etc.  Consider making this more explicit?

munificent:
Fix #4472.
Fix #4473.

### DIFF
--- a/working/2509-private-named-parameters/feature-specification.md
+++ b/working/2509-private-named-parameters/feature-specification.md
@@ -105,7 +105,7 @@ then the name affects *three* places in the program:
 1.  The name of the parameter variable inside the constructor initializer list.
     (Inside the constructor *body*, it's the instance field that is in scope.)
 
-2.  The name used to pass an argument at the callsite.
+2.  The name used to pass an argument at the call site.
 
 3.  The name of the corresponding instance field to initialize with that
     parameter.
@@ -138,7 +138,7 @@ This creates a tension. A user may want:
     unusual initializer syntax, and makes it clear to a reader that the field
     is directly initialized from that parameter.
 
-They want encapsulation, readability at callsites, and brevity in the class
+They want encapsulation, readability at call sites, and brevity in the class
 definition, but because of the compile-error on private named parameters, they
 can only [pick two][].
 
@@ -286,7 +286,7 @@ class C {
 ```
 
 While verbose, this code has the advantage of being very clear what's going on.
-A reader can see that the argument name they must use at the callsite is
+A reader can see that the argument name they must use at the call site is
 `variable`, the field is named `_variable`, and the latter is initialized from
 the former.
 
@@ -390,24 +390,26 @@ with private name *p* in constructor C:
 
 *   If *p* has no corresponding public name *n*, then compile-time error. *You
     can't use a private name for a named parameter unless there is a valid
-    public name that could be used at the callsite.*
+    public name that could be used at the call site.*
 
 *   If any other parameter in C has declared name *p* or *n*, then
     compile-time error. *If removing the `_` leads to a collision with
     another parameter, then there is a conflict.*
 
-*   Otherwise, the name of the parameter in the signature of C is *n*. By "in
-    the signature", this means uses of the constructor: invocations, dynamic
-    invocations, tear-offs, chained calls from a subclass's constructor, etc.
-    *If the parameter is named, this then avoids the compile-time error that
-    would otherwise be reported for a private named parameter.*
+If there is no error then:
 
-*   The name of the local variable in the initializer list scope of C is *p*.
-    *In the initializer list, the private name is used. Inside the body of the
-    constructor, uses of *p* refer to the field, not the parameter.*
+*   The parameter name of the parameter in the constructor is the public name
+    *n*. This means that the parameter has a public name in the constructor's
+    function signature, and arguments for this parameter are given using the
+    public name. All uses of the constructor, outside of its own code, see only
+    the public name.
 
-*   If the parameter is an initializing formal, then it initializes a
-    corresponding field with name *p*.
+*   The local variable introduced by the parameter, accessible only in the
+    initializer list, still has the private name *p*. *Inside the body of the
+    constructor, uses of _p_ refer to the instance variable, not the parameter.*
+
+*   The instance variable initialized by the parameter (and declared by it, if
+    the parameter is a field parameter), has the private name *p*.
 
 *   Else the field parameter induces an instance field with name *p*.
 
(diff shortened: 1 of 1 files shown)


working/2509-private-named-parameters/feature-specification.md
```diff
@@ -395,9 +396,11 @@ with private name *p* in constructor C:
     compile-time error. *If removing the `_` leads to a collision with
     another parameter, then there is a conflict.*
 
-*   Otherwise, the name of the parameter in C is *n*. *If the parameter is
-    named, this then avoids the compile-time error that would otherwise be
-    reported for a private named parameter.*
+*   Otherwise, the name of the parameter in the signature of C is *n*. By "in
```
lrhn: The text above says
> given a named initializing formal or declaring parameter …

I want this feature to work for positional parameters too.
It's not a compile-time error to write `C(this_foo)`, but I still rewrite it as `C(int foo) : _foo = foo;` to avoid having a private name in the public API and docs.

I don't want to have to make my parameters named to get this feature.

lrhn: On line 315, it says
> It doesn't allow _any_ parameter to start with underscore …

It continues to say that it's only some, but this first part of the sentence by itself can be read to say "allows no parameter to ...", which is confusing, and makes the reader have to reconsider what they just read. And then reconsider that first part of the sentence when they read the rest. 

Maybe "Not all named parameters …"?

lrhn (approved):

working/2509-private-named-parameters/feature-specification.md
munificent: I don't think it makes sense for positional parameters to be part of this feature because "the parameter name at the callsite" isn't meaningful to the language when it comes to positional parameters.

The language already allows a positional parameter to have a private name and this proposal doesn't change that. I'll make at note in the non-normative dartdoc section to suggest they handle positional parameters with private names in the same way.

munificent: You're right, that sentence is confusing. Rewrote. English is weird.

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -458,14 +461,21 @@ further ideas for additional warnings, lints, and quick fixes.
...
+Likewise, doc generators like [`dart doc`][dartdoc] should document the
 constructor's parameter with its public name. The fact that the parameter
 initializes or declares a private field is an implementation detail of the
 class. What a user of the class cares about is the corresponding public name for
 the constructor parameter.
 
 [dartdoc]: https://dart.dev/tools/dart-doc
 
+The language already allows a *positional* parameter to have a private name
+since doing so has no effect on callsites. Doc generators are encouraged to
```
lrhn: "call sites" in two words?

lrhn: "enocuraged" -> "required"

I want to be able to *depend* on this.
And it's not just DOC-generators, I want *everything* to behave as if the parameter had been given the corresponding public name. If I can't get that, I'll keep writing `C(int foo) : _foo = foo;`, and I don't want to do that.

Which means that I'd prefer if this feature was extended to positional initializing formals too,
making it explicit that their externally visible parameter name is the corresponding public name.
It's not just something we show, and might forget to show, the name *really is* the public name
in every visible way. Including in `toString`.

That does mean that the feature becomes potentially breaking, if someone has `Foo(this._foo, int foo) : _bar = foo` today, because then the `_foo` and `foo` would now conflict.

* You only get a public named parameter if there is no conflict. For a privately named initializing formal, that would still cause an error, not because of the conflict, but because now its parameter name is still private. For a positional parameter, `Foo(this._foo, int.foo)` would keep working, and the parameter name of `this._foo` would still be `_foo`, and visibly so in documentation.

working/2509-private-named-parameters/feature-specification.md
lrhn: So this is the name of the parameter in the *function signature*, or *function type*, of the constructor, which is what you see from the outside.

Maybe:

* The *parameter name* of the parameter in the constructor is the public name *n*.
   This means that the parameter has a public name in the constructor's function
   signature, and arguments for this parameter are given using the public name.
   All uses of the constructor, outside of its own code, sees only the public name.
* The local variable introduced by the parameter, accessible only in the initializer list,
   still has the private name *p*.
* The instance variable initialized by the parameter (and declared by it, if the parameter
   is a field parameter), has the private name *p*. 
   _(This variable is also what the name *p* refers to inside the constructor's body.)_

working/2509-private-named-parameters/feature-specification.md
lrhn: I filed a separate issue: https://github.com/dart-lang/language/issues/4479

munificent: Let's discuss it there and then I'll leave this PR as-is for now.

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -226,10 +226,11 @@ encapsulation.
 
 ## Proposal
 
-The basic idea is simple. We let users use a private name in a named parameter.
-The compiler removes the `_` from the argument name but keeps it for the
-corresponding initialized or declared field. In other words, we do exactly what
-users are doing by hand when they write an initializer like:
+The basic idea is simple. We let users use a private name in a named parameter
+when the parameter also initializes or declares a field. The compiler removes
+the `_` from the argument name but keeps it for the corresponding field. In
```
lrhn, line 231: argument name -> parameter name

leafpetersen: Maybe `argument name` -> `argument label`?  I think calling it the "parameter name" is deeply misleading.   Parameters are caller side things, arguments are callee side things.  The parameter name is the, as the name says, the name of the parameter, and *we do not remove the _ from the name of the parameter*.  That is, you refer to the parameter using the private name.  It is only on the *argument* that you use the public name.

munificent: I think the current text is probably clear enough.

PR #4480, opened 2025-07-31
Revised the declaring/primary constructor proposal according to the l…

eernstg:
This PR changes the declaring/primary constructor proposal to use a different scoping structure, to clarify the placement of the modifier `const` with a declaring constructor, to mention the principle that the class header should contain all parts of the signature of a declaring constructor if it contains any of them (e.g., `class const C();` makes the primary constructor constant, but we can't do `class C() { const this; }`). Moreover, this PR includes several other minor clarifications and corrections.

### DIFF
--- a/working/2364 - primary constructors/feature-specification.md
+++ b/working/2364 - primary constructors/feature-specification.md
@@ -4,7 +4,7 @@ Author: Erik Ernst
 
 Status: Draft
 
-Version: 1.8
+Version: 1.9
 
 Experiment flag: declaring-constructors
 
@@ -67,7 +67,7 @@ because it must be guaranteed that the declaring header constructor is
 actually executed on every newly created instance of this class. This rule
 is further motivated below.
 
-A declaring header constructor is also known as a _primary constructor_, 
+A declaring header constructor is also known as a _primary constructor_,
 because all other generative constructors must invoke the primary one
 (directly or indirectly).
 
@@ -122,7 +122,7 @@ parameter list that has the _declaring_ modifier `var` or `final`.
 With this feature, all other declarations of formal parameters as `final`
 will be a compile-time error. This ensures that `final int x` is
 unambiguously a declaring parameter. Developers who wish to maintain a
-style whereby formal parameters are never modified will have a 
+style whereby formal parameters are never modified will have a
 [lint][parameter_assignments] to flag all such mutations.
 
 [parameter_assignments]: https://dart.dev/tools/linter-rules/parameter_assignmentshttps://dart.dev/tools/linter-rules/parameter_assignments
@@ -380,8 +380,8 @@ class const A.someName(final int x);
 class B extends A {
   final String s1;
   const this(int x, int y, {required final String s2})
-      : s1 = y.toString(), 
-        assert(s2.isNotEmpty), 
+      : s1 = y.toString(),
+        assert(s2.isNotEmpty),
         super.someName(x + 1);
 }
 ```
@@ -393,12 +393,26 @@ avoiding the duplication of instance variable names and types) even in the
 case where some parameters should not introduce any instance variables (so
 they are just "normal" parameters).
 
-With a declaring header constructor, the formal parameters in the header are
-introduced into a new scope. This means that each parameter whose name is
-not introduced by any nested scope (e.g., the class body scope) is in scope
-in the class body. It is a compile-time error to refer to such a parameter
-anywhere except in an initializing expression of a non-late instance
-variable declaration.
+With a declaring header constructor (aka a primary constructor), the formal
+parameters in the header are introduced into a new scope, known as the
+_primary initializer scope_.  This scope is inserted into the scope chain in
+several locations. In particular, it is _not_ the enclosing scope for the
+body scope of the class, even though it is located syntactically in the
+class header. It is actually the other way around, namely, the class body
+scope is the enclosing scope for the primary initializer scope.
+
+The primary initializer scope is the current scope for the initializing
+expression of each non-late instance variable declaration in the class
+body, if any. Similarly, the primary initializer scope is the current scope
+for the initializer list in the body part of the primary constructor, if
+any.
+
+In other words, when a class has a primary constructor, each of the
+initializing expressions of a non-late instance variable has the same
+access to the primary constructor parameters as the initializer list would
+have if it had been a regular (non-declaring) constructor in the body. This
+is convenient, and it makes refactorings from one to another kind of
+constructor simpler and safer.
 
 ```dart
 // Current syntax.
@@ -419,15 +433,6 @@ are evaluated during the execution of the declaring header constructor,
 such that the value of a variable like `delta` is only used at a point in
 time where it exists.
 
-Similarly, if an identifier expression in an initializing expression of a
-non-late instance variable declaration resolves to an instance variable
-(this is currently an error, there is no access to `this`), and there is a
-declaring header constructor parameter that corresponds to this instance
-variable (that is, a declaring parameter or an initializing formal with the
-same name), it will evaluate to the value of the parameter.  For example,
-`x` is used in the initializer for `y` in the example above, which is
-possible because of this mechanism.
-
 This can only work if the declaring header constructor is guaranteed to be
 executed. Hence the rule, mentioned above, that there cannot be any other
 non-redirecting generative constructors in a class that has a declaring
@@ -436,7 +441,7 @@ header constructor.
 This further motivates the special terminology where a declaring header
 constructor is known as a primary constructor as well: The _primary_
 constructor is more powerful than other declaring constructors because it
-changes the scoping for the entire class declaration.
+changes the scoping for specific locations in the entire class body.
 
 Finally, here is an example that illustrates how much verbosity this
 feature tends to eliminate:
@@ -528,6 +533,14 @@ class E extends A {
 }
 ```
 
+Note that the version with a primary constructor can initialize `w` in the
+declaration itself, whereas the two other versions need to use an element
+in the initializer list of the constructor to initialize `w`. This is
+necessary because there cannot be other non-redirecting generative
+constructors when there is a primary constructor, but in the two other
+versions we could add another non-redirecting generative constructor which
+would initialize `w` with some other value.
+
 Moreover, we may get rid of all those occurrences of `required` in the
 situation where it is a compile-time error to not have them, but that is a
 separate proposal, [here][inferred-required] or [here][simpler-parameters].
@@ -580,14 +593,11 @@ constructors as well.
      '}';
 
 <constructorSignature> ::= // Modified rule.
-     <constructorName> <declaringParameterList>
-   | 'this' ('.' <identifierOrNew>);
+     <constructorName> <formalParameterList>
+   | 'this' ('.' <identifierOrNew>) <declaringParameterList>?;
 
 <constantConstructorSignature> ::= // Modified rule.
      'const' <constructorSignature>;
-
-<constructorName> ::= // Modified rule.
-     (<typeIdentifier> | 'this') ('.' <identifierOrNew>)?
 
 <identifierOrNew> ::=
      <identifier>
@@ -602,7 +612,8 @@ constructors as well.
 <declaringParameterList> ::= // New rule.
      '(' ')'
    | '(' <declaringFormalParameters> ','? ')'
-   | '(' <declaringFormalParameters> ',' <optionalOrNamedDeclaringFormalParameters> ')'
+   | '(' <declaringFormalParameters> ','
+         <optionalOrNamedDeclaringFormalParameters> ')'
    | '(' <optionalOrNamedDeclaringFormalParameters> ')';
 
 <declaringFormalParameters> ::= // New rule.
@@ -618,7 +629,7 @@ constructors as well.
    | <superFormalParameter>;
 
 <declaringFunctionFormalParameter> ::= // New rule.
-     'covariant'? ('var' | 'final')? <type>? 
+     'covariant'? ('var' | 'final')? <type>?
      <identifier> <formalParameterPart> '?'?;
 
 <declaringSimpleFormalParameter> ::= // New rule.
@@ -640,7 +651,7 @@ constructors as well.
      (',' <defaultDeclaringNamedParameter>)* ','? '}';
 
 <defaultDeclaringNamedParameter> ::= // New rule.
-     <metadata> 'required'? <declaringFormalParameterNoMetadata> 
+     <metadata> 'required'? <declaringFormalParameterNoMetadata>
      ('=' <expression>)?;
 ```
 
@@ -654,8 +665,8 @@ contains a `<primaryConstructorNoConst>`, and the body of _D_ contains a
 `<constructorSignature>` beginning with `this` that contains a
 `<declaringParameterList>`.
 
-*That is, it is an error to have a declaring parameter list of a primary
-constructor both in the header and in the body.*
+*It is an error to have a declaring parameter list both in the header and
+in the body.*
 
 A compile-time error occurs if _D_ includes a `<classNamePart>` that
 does not contain a `<primaryConstructorNoConst>`, and the body of _D_
@@ -663,21 +674,16 @@ contains a `<constructorSignature>` beginning with `this` that does not
 contain a `<declaringParameterList>`.
 
 *It is an error to have a declaring constructor in the class body, but
-no declaring parameter list, neither in the header nor in the body. Note
-that constant constructors are included because a
-`<constantConstructorSignature>` contains a `<constructorSignature>`.*
+no declaring parameter list, neither in the header nor in the body.*
 
-A compile-time error occurs if _D_ includes a `<classNamePart>` beginning
-with `const`, and the body of _D_ contains a `<constructorSignature>`
-beginning with `this` which is not part of a
-`<constantConstructorSignature>`, and vice versa.
+A compile-time error occurs if a `<constantConstructorSignature>` starts
+with `const this` and no `<declaringParameterList>` is present.
 
-*That is, it is an error for the header to contain `const` if there is a
-declaring constructor in the body as well, and it does not contain `const`;
-and it is an error for the header to not contain `const` if there is a
-declaring constructor in the body as well, and it does contain `const`.
-They must agree. Finally, it is of course allowed to omit `const` in both
-locations. In this case the constructor is not constant.*
+*In this case the class has a declaring header constructor. The keyword
+`const` can then be specified in the class header, but only there. The
+rationale is that when the class header contains any parts of a declaring
+constructor, the class header must be the location where all parts of the
+signature of that constructor is specified.*
 
 *The meaning of a declaring constructor is defined in terms of rewriting it
 to a body constructor (a regular one, not declaring) and zero or more
@@ -689,40 +695,17 @@ class body just like `class C(int i);` and just like `class C extends
 Object {}`, and all three of them have `Object` as their direct
 superclass.*
 
-Finally, a compile-time error occurs in the case where an identifier
-expression is resolved to yield the value of a declaring or initializing
-formal parameter in a declaring header constructor, and this identifier
-occurs in a function literal, and the corresponding instance variable is
-non-final.
-
-*The point is that it is highly confusing if such a parameter reference is
-considered to be "the same thing" as the variable with the same name which
-is in scope, but the parameter has the initial value of that instance
-variable, and the instance variable has been modified in the meantime.*
-
-*For example:*
-
-```dart
-class C(var int x) {
-  var closure = () => x; // Warning!
-}
-
-main() {
-  var c = C(1);
-  c.x = 2;
-  print(c.closure()); // Prints 1.
-}
-```
-
 ### Static processing
 
 Consider a class, enum, or extension type declaration _D_ with a declaring
-header constructor *(note that it cannot be a `<mixinApplicationClass>`,
-because that kind of declaration does not support declaring constructors,
-that is a syntax error)*. This declaration is treated as a class, enum,
-respectively extension type declaration without a declaring header
-constructor which is obtained as described in the following. This
-determines the dynamic semantics of a declaring constructor.
+header constructor, also known as a primary constructor *(note that it
+cannot be a `<mixinApplicationClass>`, because that kind of declaration
+does not support declaring constructors, that is a syntax error)*. This
+declaration is treated as a class, enum, respectively extension type
+declaration without a declaring header constructor which is obtained as
+described in the following. This determines the dynamic semantics of a
+declaring header constructor, and simiarly for a declaring body
+constructor.
 
 A compile-time error occurs if the body of _D_ contains a non-redirecting
 generative constructor. *This ensures that every constructor invocation for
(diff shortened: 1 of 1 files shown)


working/2364 - primary constructors/feature-specification.md
```diff
@@ -654,30 +665,25 @@ contains a `<primaryConstructorNoConst>`, and the body of _D_ contains a
...
-*That is, it is an error for the header to contain `const` if there is a
-declaring constructor in the body as well, and it does not contain `const`;
-and it is an error for the header to not contain `const` if there is a
-declaring constructor in the body as well, and it does contain `const`.
-They must agree. Finally, it is of course allowed to omit `const` in both
-locations. In this case the constructor is not constant.*
+no declaring parameter list, neither in the header nor in the body.*
+
+A compile-time error occurs if a `<constantConstructorSignature>` starts
+with `const this` and no `<declaringParameterList>` is present.
```
munificent: I'm not usually one to argue for encoding more syntax rules in the grammar, but this seems like a fairly easy one to handle, right? I think it's:

```
<constantConstructorSignature> ::= // Modified rule.
     'const' <constructorName> <formalParameterList>
   | 'const' 'this' ('.' <identifierOrNew>) <declaringParameterList>;
```

working/2364 - primary constructors/feature-specification.md
```diff
@@ -732,47 +715,50 @@ order to allow initializers with no access to `this` to use the
...
+*This scoping structure is highly unusual because the declaring parameter
+list of a primary constructor is outside the class body, and yet it is
+treated as if it were nested inside the body, and occurring in multiple
+locations! However, this ensures that the non-late variable initializers
+are treated the same as the initializer elements of an ordinary
+non-declaring constructor. Note that this only occurs when the class has a
+primary constructor, it does not occur when the class has an in-body
+declaring constructor, or when it does not have any declaring constructors
+at all. There is no access to any constructor parameters in the
+initializing expression of a non-late instance variable in those cases.*
```
munificent: A clarifying example would help here, like:

```dart
String x = 'top level';
class C(String x) {
  String instance = x;
  late String lateInstance = x;
}

main() {
  var c = C('parameter');
  print(c.instance); // Prints "parameter".
  print(c.lateInstance); // Prints "top level".
}
```

working/2364 - primary constructors/feature-specification.md
```diff
@@ -732,47 +715,50 @@ order to allow initializers with no access to `this` to use the
...
+
+The primary parameter scope is the current scope for the body of the body
+part of the declaring header constructor, if any.
+
+*The point is that the constructor body should have access to the "regular"
+parameters, but it should have access to the instance variables rather than
+the declaring or initializing parameters with the same names, also in the
+case of a declaring header constructor with a body in the class body. With
+an in-body declaring constructor, these rules just repeat what is already
+specified for the scoping of other constructors.*
```
munificent: How about clarifying example, like:

```dart
class C(var String x) {
  void Function() captureAtDeclaration = () => print(x);
  void Function() captureInInitializer = () => print(x);
  void Function() captureInBody = () => 'temp';

  this : captureInInitializer = (() => print(x)) {
     captureInBody = () => print(x);
  }
}

main() {
  var c = C('parameter');
  c.x = 'instance';
  c.captureAtDeclaration(); // Prints "parameter".
  c.captureInInitializer(); // Prints "parameter".
  c.captureInBody(); // Prints "instance".
}
```

munificent (approved):
Very nice! I agree the scoping is unusual. But it's consistent with initializer lists and those seem to have done the right thing in practice for years without user complaint.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -67,7 +67,7 @@ because it must be guaranteed that the declaring header constructor is
 actually executed on every newly created instance of this class. This rule
```
lrhn, line 67: Three lines up: "can not" -> "cannot"

working/2364 - primary constructors/feature-specification.md
```diff
@@ -67,7 +67,7 @@ because it must be guaranteed that the declaring header constructor is
...
 is further motivated below.
 
-A declaring header constructor is also known as a _primary constructor_, 
+A declaring header constructor is also known as a _primary constructor_,
```
lrhn, line 70: Why introduce "declaring header constructor" as a concept then?

working/2364 - primary constructors/feature-specification.md
```diff
@@ -393,12 +393,26 @@ avoiding the duplication of instance variable names and types) even in the
 case where some parameters should not introduce any instance variables (so
 they are just "normal" parameters).
 
-With a declaring header constructor, the formal parameters in the header are
-introduced into a new scope. This means that each parameter whose name is
-not introduced by any nested scope (e.g., the class body scope) is in scope
-in the class body. It is a compile-time error to refer to such a parameter
-anywhere except in an initializing expression of a non-late instance
-variable declaration.
+With a declaring header constructor (aka a primary constructor), the formal
```
lrhn, line 396: Just pick one, and I suggest "primary constructor".

working/2364 - primary constructors/feature-specification.md
```diff
@@ -393,12 +393,26 @@ avoiding the duplication of instance variable names and types) even in the
...
+parameters in the header are introduced into a new scope, known as the
+_primary initializer scope_.  This scope is inserted into the scope chain in
```
lrhn: This _primary initializer scope_ contains exactly the same entries as the _formal parameter initializer scope_ and the same parent scope, right? 

Maybe just call it by the same name.

lrhn: "inserted into the scope chain" is not something we otherwise say. Sounds like we take a scope chain and insert something in the middle.
I don't think we do that, we just use it as the scope of some code.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -393,12 +393,26 @@ avoiding the duplication of instance variable names and types) even in the
...
+The primary initializer scope is the current scope for the initializing
+expression of each non-late instance variable declaration in the class
+body, if any. Similarly, the primary initializer scope is the current scope
+for the initializer list in the body part of the primary constructor, if
+any.
+
+In other words, when a class has a primary constructor, each of the
+initializing expressions of a non-late instance variable has the same
+access to the primary constructor parameters as the initializer list would
+have if it had been a regular (non-declaring) constructor in the body with
```
lrhn: "same access to ...", because it's the same scope. So again, just give it the same name.

lrhn: That is:
A primary constructor declaration defines a declaring, and therefore non-redirecting generative, constructor.
The parameter list in the class header introduces the same _formal parameter scope_ and _formal parameter initializer scope_ as it would for a declaring constructor declared in the body.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -618,7 +629,7 @@ constructors as well.
    | <superFormalParameter>;
 
 <declaringFunctionFormalParameter> ::= // New rule.
-     'covariant'? ('var' | 'final')? <type>? 
+     'covariant'? ('var' | 'final')? <type>?
      <identifier> <formalParameterPart> '?'?;
```
lrhn, line 656: We *could* choose to not allow this one. We don't have function-style instance variable declarations, and a declaring formal looks like the variable declaration.

Allowing `covariant var int foo()?` is rather confusing. I can't read that as an instance variable declaration *or* as a parameter declaration.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -640,7 +651,7 @@ constructors as well.
      (',' <defaultDeclaringNamedParameter>)* ','? '}';
 
 <defaultDeclaringNamedParameter> ::= // New rule.
-     <metadata> 'required'? <declaringFormalParameterNoMetadata> 
+     <metadata> 'required'? <declaringFormalParameterNoMetadata>
      ('=' <expression>)?;
```
lrhn, line 678: (Where we can never have both `required` and `= expression` on the same production, so I guess we'll have prose to prohibit that.)

working/2364 - primary constructors/feature-specification.md
```diff
@@ -618,7 +629,7 @@ constructors as well.
    | <superFormalParameter>;
 
 <declaringFunctionFormalParameter> ::= // New rule.
-     'covariant'? ('var' | 'final')? <type>? 
+     'covariant'? ('var' | 'final')? <type>?
      <identifier> <formalParameterPart> '?'?;
 
 <declaringSimpleFormalParameter> ::= // New rule.
```
lrhn, line 658: So this is either a simple parameter or a declaring/field parameter, depending on whether it has a `var`/`final` or not. (And it can only have `covariant` if it has `var`, so it could be `('covariant'? 'var'|'final')?` ...`.)

working/2364 - primary constructors/feature-specification.md
```diff
@@ -602,7 +612,8 @@ constructors as well.
 <declaringParameterList> ::= // New rule.
```
lrhn, line 635: So the `fieldFormalParameter` is the initializing formal, not the field parameter (in Bob's terminology).
That's a little confusing.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -654,30 +665,25 @@ contains a `<primaryConstructorNoConst>`, and the body of _D_ contains a
 `<constructorSignature>` beginning with `this` that contains a
 `<declaringParameterList>`.
 
-*That is, it is an error to have a declaring parameter list of a primary
-constructor both in the header and in the body.*
+*It is an error to have a declaring parameter list both in the header and
+in the body.*
```
lrhn, line 700: Does the grammar allow that?
(Assuming that it's already an error to have any other non-redirecting generative constructor declaration if you have a primary constructor.)

If you have a primary constructor, it defines a constructor with a name.

If you have a `this: ...` block in the body, it does not have a parameter list.

If you have `ClassName(...)` in the body, it defines a separate constructor, which is an error if it has the same name, or if it's non-redirecting and generative.
We don't ever get to worrying about whether it has declaring parameters.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -654,30 +665,25 @@ contains a `<primaryConstructorNoConst>`, and the body of _D_ contains a
...
-beginning with `this` which is not part of a
-`<constantConstructorSignature>`, and vice versa.
-
-*That is, it is an error for the header to contain `const` if there is a
-declaring constructor in the body as well, and it does not contain `const`;
-and it is an error for the header to not contain `const` if there is a
-declaring constructor in the body as well, and it does contain `const`.
-They must agree. Finally, it is of course allowed to omit `const` in both
-locations. In this case the constructor is not constant.*
+no declaring parameter list, neither in the header nor in the body.*
```
lrhn, line 708: What is a "declaring constructor"?

There is no *definition* of the term.

It says somewhere that a declaring constructor has a different grammar for its parameter than other constructors.
If so, how can there be a declaring constructor and no declaring parameter list.

(I'm guessing that the idea is that:
* A declaring constructor is a single declaration, where some syntax components can be placed either in the body or in the header.
* So you can have `Foo(args)` in the header, or `this(args)` in the body, and they both define the parameters of the same (single) declaring constructor.
* Then it's an error to have arguments in both or neither places.

Is this correct?

working/2364 - primary constructors/feature-specification.md
lrhn: Agree. The header form is one of
```ebnf
    <typeName><typeParameters>?
  | 'const'? <typeName><typeParameters>? ('.' <newOrIdentifier>)? <declaringParameterList>
```
and the body form is one of:
```ebnf
  'this' |
  'const'? 'this' ('.' newOrIdentifier)? <declaringParameterList>
```
(followed by `(':' <initializerList>)? (<block> | ';')` or something like that.)

Then it's an error to have more than one in-body declaring constructor (we'll worry about augmentations later), and an error to have more or less than one declaring parameter list.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -689,40 +695,17 @@ class body just like `class C(int i);` and just like `class C extends
...
+cannot be a `<mixinApplicationClass>`, because that kind of declaration
+does not support declaring constructors, that is a syntax error)*. This
+declaration is treated as a class, enum, respectively extension type
+declaration without a declaring header constructor which is obtained as
+described in the following. This determines the dynamic semantics of a
+declaring header constructor, and simiarly for a declaring body
+constructor.
 
 A compile-time error occurs if the body of _D_ contains a non-redirecting
 generative constructor. *This ensures that every constructor invocation for
```
lrhn: ... every generative constructor invocation ...

working/2364 - primary constructors/feature-specification.md
```diff
@@ -732,47 +715,50 @@ order to allow initializers with no access to `this` to use the
...
-*This means that initializing expressions can, apparently, use the value of
-instance variables declared by the same class (not, e.g., inherited ones).
-They will actually get the value of the declaring header parameter, but
-this value is also guaranteed to be the initial value of the corresponding
-instance variable.*
-
-*Note that it only applies to identifier expressions. In particular, this
-does not allow initializing expressions to assign to other instance
-variables.*
+a new scope, the _primary initializer scope_, whose enclosing scope is the
```
lrhn, line 788: I think this is just the formal parameter initializer scope again, now also with entries for declaring parameters. Which means that it's exactly the same as the scope of the same declaring parameter list of an in-body declaring constructor. (Which is good, that's what we want.) So no need for a "new" scope or new name.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -732,47 +715,50 @@ order to allow initializers with no access to `this` to use the
...
+body scope of _D_. Every primary parameter is entered into this scope.
```
lrhn, line 789: "entered into this scope" is vague. But we already defined that for the formal parameter initializer scope, so no need to do it again.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -732,47 +715,50 @@ order to allow initializers with no access to `this` to use the
...
+
+The same parameter list also introduces the _primary parameter scope_,
+whose enclosing scope is also the body scope of the class. Every primary
+parameter which is not declaring and not initializing is entered into this
+scope.
```
lrhn: And that's the _formal parameter scope_ again.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -732,47 +715,50 @@ order to allow initializers with no access to `this` to use the
...
 
 A compile-time error occurs if _p_ contains a term of the form `this.v`, or
 `super.v` where `v` is an identifier, and _p_ has the modifier
-`covariant`. *For example, `required covariant int this.v` is an error.*
+`covariant`. *For example, `required covariant int this.v` is an error. The
+reason for this error is that the modifier `covariant` must be specified on
+the declaration of `v` which is known to exist, not on the parameter.*
 
 A compile-time error occurs if _p_ has both of the modifiers `covariant`
 and `final`. *A final instance variable cannot be covariant, because being
```
lrhn, line 877: And the next paragraph says error if `covariant` and neither `var` nor `final`. Can remove `final` from that, because of this paragraph.

So basically: It's a compile-time error if a parameter has the modifier `covariant` and it's not a declaring parameter with the modifier `var`.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -786,16 +772,16 @@ Conversely, it is not an error for the modifier `covariant` to occur on a
...
 where _D_ is the class, extension type, or enum declaration in the program
 that includes a declaring constructor, and _D2_ is the result of the
 derivation of the semantics of _D_. The derivation step will delete
 elements that amount to the declaring constructor; it will add a new
-constructor _k_; it will add zero or more instance variable declarations.
+constructor _k_; and it will add zero or more instance variable
+declarations.
 
 Where no processing is mentioned below, _D2_ is identical to _D_. Changes
 occur as follows:
```
lrhn, line 900: Below here: The type of the field of an untyped declaring parameter is derived from the default value.
It should *first* check if the field can inherit a type from a superinterface.
If any superinterface has a getter and/or setter, inherit types normally. If that doesn't give anything, then try to infer from the default value.

That matches the behavior of
```dart
class C extends S {
  var x = e;
}
```
If `S` has an `x`, the type is inherited. if not, the type is inferred.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -870,15 +856,11 @@ positional or named parameter remains optional; if it has a default value
   in _L2_, and it is added to the instance variable declaration named `p`.
```
lrhn, line 977: You could also not modify anything, and just give the semantics of declaring parameters, which happens to be exactly the same as that of an initializing formal of the same form, for the variable introduced by the declaring parameter.

I'd rather not copy or modify *anything*. Just give semantics to the syntax.

A declaring constructor declaration defines a constructor with the given name, parameter list, initializer list and body, *and* a number of instance variables with the given name, type and finality.

The static semantics of a declaring constructor is the same as that of another non-redirecting generative constructor, other than the declaring parameters, which have their declared type and (eventually public) parameter name, and also bind their name in the formal parameter initialier scope.

The runtime semantics of a declaring constructor is the same as that of another non-redirecting generative constructor, other than  for the declaring parameters, which during binding actuals to formals will initialize their corresponding instance variable to their argument value (which may be their default value if optional and no argument provided), and bind that value to the parameter variable of the formal parameter initialzier scope. _(Aka. the same as an initializing formal for that instance variable.)

A primary constructor also introduces a declaring with the given name, parameter list, initializer list and body, and the instance variables of its declaring parameters.
When executed, it binds actuals to formals the same way as an in-body declaring constructor.
Then, before executing the initializer list, it executes the non-late instance variable initializer expressions of the class, to initialize those variables. Then it continues with the initializer list. When reaching the end, it might invoke the super-constructor defined on the `extends` clause instead of one in the initializer list.

working/2364 - primary constructors/feature-specification.md
```diff
@@ -887,49 +869,37 @@ Finally, _k_ is added to _D2_, and _D_ is replaced by _D2_.
...
+but it also allows this keyword to be omitted. The specified constructor
+will be constant in both cases. This differs from the treatment of regular
+(non-declaring) constructors in an `enum` declaration: They _must_ have the
+modifier `const`, it is never inferred. This discrepancy was included
+because the syntax `enum const E(String s) {...}` seems redundant because
+`enum` implies that every constructor must be constant. This is not the
+case in the body where a constructor declaration may be physically pretty
+far removed from any syntactic hint that the constructor must be constant
+(if we can't see the word `enum` then we may not know that it is that kind
+of declaration, and the constructor might be non-const).
```
lrhn, line 1016: The text above also says it applies to extension types, and that the extension type constructor is a primary constructor.

That's breaking, because it prevents having other non-redirecting generative constructors.
Since extension types cannot have any instance variable declarations, it has no benefit from having the parameters in their scope, and therefore no risk from having other constructors.

So, extension types should be allowed to have a primary constructor *and* other non-redirecting generative constructors (because it cannot have any other "instance variables" than the representation variable, which is not declared in the body).

working/2364 - primary constructors/feature-specification.md
```diff
@@ -654,30 +665,25 @@ contains a `<primaryConstructorNoConst>`, and the body of _D_ contains a
 `<constructorSignature>` beginning with `this` that contains a
```
lrhn: "beginning with `this`" ... maybe just give the second production of `<constructorSignature>` a name, and refer to it by that. Then I don't have to worry about whether `const this...` counts or not. (It does, a `'const' <constructorSignature>` also contains a `<constructorSignature>`. It'll also be an error because of the `const`, but this should make sense either way.)

working/2364 - primary constructors/feature-specification.md
eernstg: Good points, done!

working/2364 - primary constructors/feature-specification.md
eernstg: Thanks, done!

working/2364 - primary constructors/feature-specification.md
eernstg: Great, thanks! I removed the initialization in the declaration of `captureInInitializer` to avoid an error and improve readability.

working/2364 - primary constructors/feature-specification.md
eernstg: Done.

working/2364 - primary constructors/feature-specification.md
eernstg: I think it makes sense to be able to talk about the situation where there is a declaring constructor, be it in the header or in the body. As a refinement thereof, we can talk about a 'declaring header constructor' or a 'declaring body constructor' or about the 'in-header form' or 'in-body form' of declaring constructors. Finally, 'primary constructor' is an effective and short form which will probably be used more frequently, but this doesn't mean that there is no need for this tiny amount of abstraction.

working/2364 - primary constructors/feature-specification.md
eernstg: I'll keep this one unchanged, based on the argument given above: It is not useless to have both forms, and this wording helps establishing that 'primary constructor' is a concise way to say 'declaring header constructor'.

working/2364 - primary constructors/feature-specification.md
eernstg: Revised.

eernstg: I don't think it's a reader-friendly choice to use the same name to denote two different scopes created from a set of declarations that occur in completely different locations in the syntax.

I'll make sure it is emphasized in commentary that the primary initializer scope and the formal parameter initializer scope of a constructor correspond to each other.

working/2364 - primary constructors/feature-specification.md
eernstg: I don't think we can say that it is 'same formal parameter scope and formal parameter initializer scope' because they are used in such a different way.

working/2364 - primary constructors/feature-specification.md
eernstg: You can't read `var int foo` as a parameter declaration, either, unless you're reading it as a declaring parameter, and if you're thinking about a declaring parameter then it will by definition include these forms.

So we will have to get used to new syntax in order to express these new kinds of entities.

If we want to change this syntax then we should do it in lockstep with `<functionFormalParameter>`, of which this is a variant.

working/2364 - primary constructors/feature-specification.md
eernstg: This is again written as a variant of `<normalFormalParameterNoMetadata>`, so if we change any one of them we should also change the other.

eernstg: About the sanity rule: The NNBD feature spec has "It is an error if a required named parameter has a default value". Added a similar rule here for the declaring variant.

working/2364 - primary constructors/feature-specification.md
eernstg: It will be a `<declaringFunctionFormalParameter>` whether or not there is a `var` or a `final` token if it occurs in a syntactic context where a `<declaringFormalParameter>` is expected, but it will be (semantically) non-declaring if there is no `var`-or-`final` token.

It could indeed be `('covariant'? ('var'|'final'))?`, but `covariant` is already allowed in a lot of locations where we have non-grammar rules to make it an error. I think that's a good trade-off for this kind of keyword.

working/2364 - primary constructors/feature-specification.md
eernstg: True, `<fieldFormalParameter>` is a weird name, we don't use 'field' otherwise in the language specification.

working/2364 - primary constructors/feature-specification.md
eernstg: The grammar does allow this:

```dart
class A(final int i) {
  this(final int j); // or `this.name(final int j);`.
}
```

So we need to have the non-grammar error to prevent it.

working/2364 - primary constructors/feature-specification.md
eernstg: Good point! I added a definition, and it follows the idea you described. The error that arises if we have a declaring parameter list in both locations is specified elsewhere.

working/2364 - primary constructors/feature-specification.md
eernstg: Yep, done!

working/2364 - primary constructors/feature-specification.md
eernstg: See earlier comments about the same topic.

working/2364 - primary constructors/feature-specification.md
eernstg: We always use that phrase.

working/2364 - primary constructors/feature-specification.md
eernstg: Again, see earlier comments on the same topic.

working/2364 - primary constructors/feature-specification.md
eernstg:
```dart
class A(
  covariant var   T name1, // OK, `name1=` parameter is covariant.
  covariant final T name2, // Error, there is no `name2=`.
  covariant       T name3, // Error, no instance variable, no setter.
) ...
```

You're right that only the `covariant var` case is allow. However, tools need to emit very different error messages for the two error cases, so I'll keep them separate even though it's a couple of lines longer.

working/2364 - primary constructors/feature-specification.md
eernstg: Great point! Done!

eernstg:
Thanks for great reviews! More changes are coming tomorrow.

working/2364 - primary constructors/feature-specification.md
eernstg: I created https://github.com/dart-lang/language/issues/4488 to clarify this.

working/2364 - primary constructors/feature-specification.md
eernstg: Good point, done!

working/2364 - primary constructors/feature-specification.md
eernstg: Good point! Done.

working/2364 - primary constructors/feature-specification.md
eernstg: I'll create an issue for this one.

PR #4486, opened 2025-08-06
Extend private named parameters to apply to positional parameters too.

### LINKED ISSUE #4479: Private named parameter should apply to positional parameters too.
The current specification for private named parameters only applies to named parameters.

Only named parameters need a feature to make `Foo({this._foo});` not be an error, but a positional parameter like `Foo(this._foo);` would also benefit from *consistently* treating the parameter name as `foo`.

The specification encourages doc-tools to display `this._foo` as `foo` in documentation, but that is only an encouragement, and even if followed, it's *not enough*.

I want to be able to write `Foo(this._foo)` and *trust* that the parameter is treated like it's named `foo` in *every place* the user may see that name, other than the actual source code of the constructor.
If I can't trust that, I'll likely keep writing `Foo(int foo) : _foo = foo;`. And I don't want to do that.

The places where that name is visible includes:
* Documentation. I want to write `[foo]` to document the parameter in the constructor doc.
* Runtime `toString()`s of:
    * the runtime function,
    * its runtime type, and
    * its static type, including where that gets propagated by inference.

_I really do not want to expose any implementation detail in the public API._

So I propose that the renaming of the parameter *also* applies to positional initializing formals and field parameters. If they have a private declared name, which is the name of a private instance variable, the *parameter name* is the corresponding public name, and it should be enforced deep in our tools to ensure that every tool sees that name as the parameter name,

That does mean that the feature becoms potentially breaking, if someone has `Foo(this._foo, int foo) : _bar = foo` today, because then the `_foo` and `foo` would now conflict.

* You only get a public named parameter if there is no conflict. For a privately named initializing formal, that would still cause an error, because now its parameter name is still private. For a positional parameter, `Foo(this._foo, int.foo)` would keep working, and the parameter name of `this._foo` would still be `_foo`.

I still think it's worth it.

munificent:
To avoid breakage (which doesn't seem worth it to me), it's more conservative with how collisions in private positional parameters are handled in order to not be a breaking change. (I don't think the benefits to doc comments are enough to warrant a breaking change.)

Fix #4479.

### DIFF
--- a/working/2509-private-named-parameters/feature-specification.md
+++ b/working/2509-private-named-parameters/feature-specification.md
@@ -374,6 +374,12 @@ void pricyHammer() => Hammer(price: 200);
 
 ## Static semantics
 
+Prior to this proposal, a parameter only had one name which was used everywhere
+the parameter can be referred to. With this proposal, "parameter name" can be
+ambiguous, so we introduce some terminology.
+
+### Public and private names
+
 An identifier is a **private name** if it starts with an underscore (`_`),
 otherwise it's a **public name**.
 
@@ -385,6 +391,40 @@ underscore does not leave something which is is a valid identifier *(as in `_`
 or `_2x`)* or leaves another private name *(as in `__x`)*, then the private name
 has no corresponding public name.
 
+### Declared and access names
+
+*   A parameter's **declared name** is the original identifier used to introduce
+    the formal parameter. It's the name of the local variable used to access the
+    argument value bound to a parameter inside the body of a function or
+    constructor initializer list.
+
+    If the parameter initializes or declares an instance variable, the declared
+    name determines the name of the corresponding instance variable.
+
+*   A parameter's **access name** is the name that allows users of surrounding
+    declaration to work with the parameter. It's the name written at a callsite
+    to pass a named argument to the function or constructor.
+
+    It's the name that should be shown for the parameter in generated
+    documentation from tools like [dartdoc][]. When other documentation wants to
+    refer to the parameter using the `[squareBracket]` identifier syntax, that
+    doc comment should refer to it by its access name.
+
+    If a Dart implementation refers to the parameter in any automatically
+    generated code or strings, those should use its access name. For example,
+    implicit implementations of `toString()`, runtime error messages, stack
+    traces, etc.
+
+    In short, for all *users* of some code, it should appear as if the parameter
+    only has its access name.
+
+[dartdoc]: https://dart.dev/tools/dart-doc
+
+Unless otherwise specified, the access name of a parameter is its declared name.
+The exceptions are:
+
+### Private named parameters
+
 Given a named initializing formal or field parameter (for a primary constructor)
 with private name *p* in constructor C:
 
@@ -392,26 +432,19 @@ with private name *p* in constructor C:
     can't use a private name for a named parameter unless there is a valid
     public name that could be used at the call site.*
 
-*   If any other parameter in C has declared name *p* or *n*, then
-    compile-time error. *If removing the `_` leads to a collision with
-    another parameter, then there is a conflict.*
+*   If any other parameter in C has declared name *n*, then compile-time error.
+    *If removing the `_` leads to a collision with another parameter, then there
+    is a conflict. It is of course already an error if any other parameter has
+    declared name _p_. That's a normal parameter name collision error.*
 
-If there is no error then:
+    ```dart
+    class C {
+      C(int _a, {this._a}); // Already an error for parameter names to collide.
+      C(int a, {this._a}); // New error because the public name collides.
+    }
+    ```
 
-*   The parameter name of the parameter in the constructor is the public name
-    *n*. This means that the parameter has a public name in the constructor's
-    function signature, and arguments for this parameter are given using the
-    public name. All uses of the constructor, outside of its own code, see only
-    the public name.
-
-*   The local variable introduced by the parameter, accessible only in the
-    initializer list, still has the private name *p*. *Inside the body of the
-    constructor, uses of _p_ refer to the instance variable, not the parameter.*
-
-*   The instance variable initialized by the parameter (and declared by it, if
-    the parameter is a field parameter), has the private name *p*.
-
-*   Else the field parameter induces an instance field with name *p*.
+*   Otherwise, the access name of the parameter is *n*.
 
 *For example:*
 
@@ -431,11 +464,65 @@ main() {
 }
 ```
 
-*Note that the proposal only applies named parameters and only to ones which are
+*Note that this section only applies to named parameters that are also
 initializing formals or field parameters. A named parameter can only have a
 private name in a context where it is _useful_ to do so because it corresponds
 to a private instance field. For all other named parameters it is still a
 compile-time error to have a private name.*
+
+### Private positional parameters
+
+Dart already allows a positional parameter to have a private name. This is
+useful for a positional constructor parameter that is an initializing formal or
+a declaring parameter for a private instance variable. For other parameters,
+it's non-idiomatic but harmless to give them private names. *(Dart also allows
+local variables to have private names, which is unusual but harmless. A lint
+suggests that users don't do this.)*
+
+However, the user experience of the language is more than just the semantics.
+Generated documentation, code navigation in doc comment references, and runtime
+error messages all reflect a function or constructor's parameters back to the
+user.
+
+In all of those contexts, we want it to be an implementation detail that a
+positional parameter happens to have a private name. To that end:
+
+Given a positional parameter with private name *p* in formal parameter list L:
+
+*   If *p* has corresponding public name *n* and no other parameter in L has
+    declared name *n*, then the access name of the parameter is *n*.
+
+Since a positional parameter can't be passed using a named argument, this
+doesn't affect the language semantics. It does mean that generated docs, error
+messages, inferred super parameter names, etc. should all use the public name
+of the parameter. For example:*
+
+```dart
+const int foo = 0;
+
+class C {
+  int _foo;
+
+  /// Uses [foo] to set [_foo].
+  C(this._foo);
+}
+```
+
+Here, the generated documentation for C should show the constructor signature
+like `C(int foo)`. If a user clicks the `[foo]` reference in the doc comment
+and navigates to its definition, their IDE should take them to the constructor
+parameter `_foo` and not the unrelated top level constant `foo`.
+
+In short, a class author should never feel that can't use a private name for
+a positional initializing formal or declaring parameter because doing so might
+degrade the user experience of anyone working with the class or reveal the
+implementation detail that the parameter happens to initialize a private field.
+
+*Unlike with named parameters, this section applies to all positional
+parameters, regardless of whether they are initializing formals, declaring
+parameters, or even constructor parameters at all. This is because the language
+already allows using private names in all positional parameters, and we want a
+consistent user experience for all of them.*
 
 ## Runtime semantics
 
@@ -444,12 +531,18 @@ renaming.
 
 ## Compatibility
 
-This proposal takes code that it is currently a compile-time error (a private
-named parameter) and makes it valid in some circumstances (when the named
-parameter is an initializing formal or field parameter). Since it simply expands
-the set of valid programs, it is backwards compatible. Even so, it should be
-language versioned so that users don't inadvertently use this feature while
-their program allows being run on older pre-feature SDKs.
+For named parameters, this proposal takes code that it is currently a
+compile-time error (a private name) and makes it valid in some circumstances
+(when the named parameter is an initializing formal or field parameter). Since
+it simply expands the set of valid programs, it is backwards compatible.
+
+For positional parameters with private names, it falls back to continuing to use
+the private name as the access name in the rare case that there is a collision.
+That avoids breaking existing uses of private names in parameter lists.
+
+Even though non-breaking, this feature should be language versioned so that
+users don't inadvertently use this feature while their program allows being run
+on older pre-feature SDKs.
 
 ## Tooling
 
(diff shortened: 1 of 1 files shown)


munificent:
cc @srawlins and @kallentu if you have thoughts about this PR in relation to my comment here: https://github.com/dart-lang/language/issues/4479#issuecomment-3161958509.

munificent:
If we take this change, is it worth renaming the feature and experiment flag to just "private parameters" or "private parameter names"?

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -385,33 +391,60 @@ underscore does not leave something which is is a valid identifier *(as in `_`
...
+
+*   A parameter's **declared name** is the original identifier used to introduce
+    the formal parameter. It's the name of the local variable used to access the
+    argument value bound to a parameter inside the body of a function or
+    constructor initializer list.
+
+    If the parameter initializes or declares an instance variable, the declared
+    name determines the name of the corresponding instance variable.
+
+*   A parameter's **access name** is the name that allows users of surrounding
```
lrhn, line 404: Not sure "access name" works for me. It feels like the name you use to read it, not to set (pass) it.

I'd just use "parameter name", the name of the parameter in the function type of the function/constructor.

---

A *parameter declaration* has a *declared name*, which is the syntactic identifier used to introduce the formal parameter. _(Or it might not, if it's declared identifier is `_`.)_
_Examples of parameters where `name` is the declared name: `name`, `int name`,  `this.name`, `final int name`, `super.name`, and (for completeness, not encouragement) `int name()`._

The declaration introduces a local _parameter variable_ into the *parameter scope* (normal parameters only) and the *initializer list scope* (all kinds of parameter), bound to the argument value passed to that parameter.
The parameter variable's name is the same as the declared name.
_Example where `x` is in the initializer list scope and `deltaY` is in the parameter
scope:_
```dart
class Point {
  final int x, y;
  Point(this.x, this.y);
  Point._offSetFromDiagonal(this.x, int deltaY) : y = x + deltaY {
    if (deltaY == 0) log("Another diagonal point!: ($x, $x)");
  }
}
```

The parameter declaration also introduces a *function parameter* to the constructors function signature. This name occurs in the function signature, but not in any scope, and it is the name used to refer to the parameter from outside of the constructor declaration. For named parameters, it is the name used to pass an argument at call sites. For positional parameters, the name is never used in code to denote the parameter outside of the function. Both can be referenced indirectly by tools that refer to source names, like DartDoc, in error messages, or while debugging.
_Example with references to the names of `value` and `negated`:_
```dart
extension type const N._(int value) {
  /// Creates [N] with [value] as [N.value], negated if [negated] is true.
  const N(int value, {bool negated = false}) 
    : assert(!(value == 0 && negated), 
          "The 'value' must not be zero when 'negated' is true"),
      value = negated ? -value : value;
}
```
_For non-constructor parameters, there is also a lint to ensure that an overriding method has the same positional parameter names as the 
function it overrides._

... and the parameter name of field parameters and initializing formals
is now not always the declared name ...

----

(Now I waxed didacticly again. I really need to learn to *stop* writing what I'm thinking. Or while I'm thinking. And I'm doing it again!)

Point is: "access name" feels like the opposite of "name used to *pass* an argument. 

I like "parameter name" because that's what we already call the name of the parameter of a function from the outside, which is what this name is.

Use what you want of the above, or ignore it. I tend towards over-explaining things.

Today we use "parameter" and occasionally "parameter name" for internal local variables today, because we haven't made a distinction. Now we should make that distinction, so I suggest "parameter" and "parameter variable", and "parameter name" and "parameter variable name". And for field parameters also a "field name" of the instance variable declared by the field parameter.

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -385,33 +391,60 @@ underscore does not leave something which is is a valid identifier *(as in `_`
...
+    refer to the parameter using the `[squareBracket]` identifier syntax, that
+    doc comment should refer to it by its access name.
+
+    If a Dart implementation refers to the parameter in any automatically
+    generated code or strings, those should use its access name. For example,
+    implicit implementations of `toString()`, runtime error messages, stack
+    traces, etc.
+
+    In short, for all *users* of some code, it should appear as if the parameter
+    only has its access name.
```
lrhn, line 419: A little confusing to say "only ... its access name" when there no alternative mentioned so far. What else *could* it appear as? (Read on to find out! :wink: )

Maybe it would help to say that "With this feature, the parameter name is not always the same as the declared name." Maybe followed by: "Everything outside of the function's own implementation sees only the parameter by its parameter name, and only the function declaration itself refer to the parameter by its the declared name."

Another reason I like to use just "parameter name" for the concept, is that then nothing changes outside of the function. That was what it was already called.
We just introduce a new syntax for declaring a parameter with a public name - and a private parameter variable name, but that's internal to the constructor only.
(We will still have to go through all code and check that it uses the parameter's name, not the parameter declartion's declared name, where there is a difference, but it will be consistent.)

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -385,33 +391,60 @@ underscore does not leave something which is is a valid identifier *(as in `_`
...
+
+[dartdoc]: https://dart.dev/tools/dart-doc
+
+Unless otherwise specified, the access name of a parameter is its declared name.
+The exceptions are:
+
+### Private named parameters
+
 Given a named initializing formal or field parameter (for a primary constructor)
 with private name *p* in constructor C:
```
lrhn, line 429: C -> \*C\*

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -385,33 +391,60 @@ underscore does not leave something which is is a valid identifier *(as in `_`
...
-
-*   The parameter name of the parameter in the constructor is the public name
-    *n*. This means that the parameter has a public name in the constructor's
-    function signature, and arguments for this parameter are given using the
-    public name. All uses of the constructor, outside of its own code, see only
-    the public name.
+*   If any other parameter in C has declared name *n*, then compile-time error.
+    *If removing the `_` leads to a collision with another parameter, then there
+    is a conflict. It is of course already an error if any other parameter has
+    declared name _p_. That's a normal parameter name collision error.*
```
lrhn, line 438: > It is of course already ...

I wouldn't take that as a given. If we change the name of the parameter from *p* to *n*, then whether it is of course already an error depends on where in the semantic sequence the clash is checked. If it's after the renaming, then there is no error for 
`C(int _foo, {this._foo});`, because there is no conflict after the rename. 

Fx, if we say that it's an error if two parameters have the same declared name, we're good. If we say it's an error if they have the same parameter name, then it compares the public name to other parameter names.

(I checked. It *is* an error because both `_foo`s introduce a binding in the initializer list scope, so twice in one scope. But I still don't think it warrants saying "of course" for that. *I* had to look up the details to be sure.)

So let's be thorough:
* It's (still) a compile-time error if two parameters have the same declared name.
* If a named initializing formal or field parameter pf constructor *C* has private name *p*
    * It's a compile-time error if *p* has no corresponding public name.
    * Otherwise let *n* be the corresponding public name of *p*.
    * It's a compile-time error if any other parameter of *C* has declared name *n*.
    * Otherwise the parameter name of the parameter is *n*.

(Plus the elaborations on why, those are great.)

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -431,25 +464,85 @@ main() {
...
 private name in a context where it is _useful_ to do so because it corresponds
 to a private instance field. For all other named parameters it is still a
 compile-time error to have a private name.*
 
+### Private positional parameters
+
+Dart already allows a positional parameter to have a private name. This is
+useful for a positional constructor parameter that is an initializing formal or
+a declaring parameter for a private instance variable. For other parameters,
+it's non-idiomatic but harmless to give them private names. *(Dart also allows
```
lrhn, line 478: but harmless -> , but harmless,

(Says my non-native _always comma before "but"_ rote learning.)

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -431,25 +464,85 @@ main() {
...
+
+However, the user experience of the language is more than just the semantics.
+Generated documentation, code navigation in doc comment references, and runtime
+error messages all reflect a function or constructor's parameters back to the
+user.
+
+In all of those contexts, we want it to be an implementation detail that a
+positional parameter happens to have a private name. To that end:
+
+Given a positional parameter with private name *p* in formal parameter list L:
```
lrhn, line 490: I'd still only do this for initializing formals and field parameters, where there is a reason to have a private name. If you ask for it otherwise, you're on your own.

lrhn, line 490: Maybe mention that `_` is not a private name. (Because it's not a name at all, the parameter is unnamed.)

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -431,25 +464,85 @@ main() {
...
+
+*   If *p* has corresponding public name *n* and no other parameter in L has
```
lrhn, line 492: And maybe also make it explicit that it's a compile-time error if any other parameter has declared name *p*.

(But it is true that every named parameter introduces a binding into the initializer list scope, for its private name, so we will get a "same name twice in one scope" error.)

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -431,25 +464,85 @@ main() {
...
+    declared name *n*, then the access name of the parameter is *n*.
+
+Since a positional parameter can't be passed using a named argument, this
+doesn't affect the language semantics. It does mean that generated docs, error
+messages, inferred super parameter names, etc. should all use the public name
```
lrhn, line 497: The "inferred super parameter names" are the ones of anonymous mixin applications?
Maybe say that: "of mixin application classes". They're probably not specified as super-parameters, just as "forwarding constructors", so pedantically: "Names of parameters of forwarding constructors of mixin application classes".

Or maybe drop it, it's not an exhaustive list, and it's hard to explain.

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -431,25 +464,85 @@ main() {
...
+In short, a class author should never feel that can't use a private name for
+a positional initializing formal or declaring parameter because doing so might
+degrade the user experience of anyone working with the class or reveal the
+implementation detail that the parameter happens to initialize a private field.
+
+*Unlike with named parameters, this section applies to all positional
+parameters, regardless of whether they are initializing formals, declaring
+parameters, or even constructor parameters at all. This is because the language
+already allows using private names in all positional parameters, and we want a
+consistent user experience for all of them.*
```
lrhn, line 525: Maybe it's a good idea. Maybe it's inconsistent that it doesn't work for other kinds of named parameters too. I could see myself wanting to write `bar({int _foo = 0}) { ... }` to not shadow a surrounding `foo`, while still getting `foo` as parameter name.

I think it's an easier story to tell if it only works for initializing formals and field parameters, parameters whose name *also* refers to something else, and that other thing might want to be private. For everything else, if you choose a private name, it's because you wanted one, and you could just use a public name.

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -431,25 +464,85 @@ main() {
...
+error messages all reflect a function or constructor's parameters back to the
+user.
+
+In all of those contexts, we want it to be an implementation detail that a
+positional parameter happens to have a private name. To that end:
+
+Given a positional parameter with private name *p* in formal parameter list L:
+
+*   If *p* has corresponding public name *n* and no other parameter in L has
+    declared name *n*, then the access name of the parameter is *n*.
```
lrhn, line 493: ... , otherwise it's *p*.

And for all other parameter declarations, the parameter name is the declared name.

_Can't fall back on defaults without an "otherwise", because we don't want to hit the default definition at all, which would says that the access name is the declared name. If we can hit that by simple fallthrough, then we'll also hit it from the clause that says the access name is *n*, and then it has two access names. Exhaust **all** the cases!_


So, in total:
* If a named parameter is an initializing formal or field parameter with a private declared name *p* which has a corresponding public name *n*, then:
    * It's a compile time error if any other parameter of the same constructor has declared name *p*.
    * The access name of the parameter is *n*.
* If a positional parameter has a private declared name *p* with corresponding public name *n*, and no other parameter of the same function or constructor has declared name *p* or *n*,
    * then the access name of the parameter is *n*.
* Otherwise the access name of the parameter is *p*.
* It's a compile-time error if two parameters of the same function or constructor have the same access name.

Is that about right?

working/2509-private-named-parameters/feature-specification.md
```diff
@@ -459,18 +552,33 @@ normative*, but is merely suggestions and ideas for the implementation teams.
...
+### Error messages
+
+Compile errors and runtime exceptions (from things like invalid dynamic calls)
+may sometimes show a function's signature to the user or otherwise refer to a
+formal parameter. When tools generate these error strings, they should use the
+access name of each parameter if the error relates to a *use* of the parameter
+list. *For example, an error related to calling a function with an incorrect
+parameter type.*
+
+Errors that refer to accessing the parameter from within the function or
```
lrhn, line 564: ... accessing the parameter variable ...

(to make a distinction between the parameter and its introduced variable.)

working/2509-private-named-parameters/feature-specification.md
munificent: A comma appears before "but" when used as a conjunction to separate two independent clauses, but not necessarily for other uses. It's exactly the same as with and:

* The frog is small and brightly-colored. -> The frog is small but brightly-colored.
* The frog is small, but its croak is loud. -> The frog is small, and its croak is loud.

"But" is not used outside of joining independent clauses as often as "and" but it can be sometimes when you want to describe a thing with two properties that are both true but the second is surprising given the first. 
In all contexts, "but" essentially means "and except unexpectedly so".

working/2509-private-named-parameters/feature-specification.md
munificent: > Not sure "access name" works for me. It feels like the name you use to read it, not to set (pass) it.

Me either, but it was the best I could think of. I thought about it some more and came up with "visible name". I think that works well. It's not exactly the same as "public" (so we can distinguish a public *identifier* from a visible *parameter*), but it conveys pretty much the same thing.

> (Now I waxed didacticly again. I really need to learn to stop writing what I'm thinking. Or while I'm thinking. And I'm doing it again!)

<3 :D

working/2509-private-named-parameters/feature-specification.md
munificent: > Maybe it would help to say that "With this feature, the parameter name is not always the same as the declared name."

Good point. Done.

>  "Everything outside of the function's own implementation sees only the parameter by its parameter name, and only the function declaration itself refer to the parameter by its the declared name."

Also, good. Added that further down.

working/2509-private-named-parameters/feature-specification.md
leafpetersen: Why make things egregiously inconsistent?  If we're going to do this (I'm not convinced) we can at least be consistent.

leafpetersen (changes requested):
I left my main comment on the issue [here](https://github.com/dart-lang/language/issues/4479#issuecomment-3168976035), but for the record here, I am opposed to this change.

munificent:
Closing this based on discussion here and on #4479. (See also #4490.)

PR #4490, opened 2025-08-08
Revise the section on documenting private parameters.

### LINKED ISSUE #4479: Private named parameter should apply to positional parameters too.

munificent:
Fix #4479.

This is a much more narrowly scoped solution to #4479. It doesn't change any scoping rules and there's nothing normative. But it aims to guide doc generators to hide the private implementation details of APIs.

If that's not a strong enough guarantee for the core libraries to rely on, then it may make sense for those libraries to not use private initializing formals in public APIs.

But for most users, I think this is plenty robust enough, and it avoids adding unnecessary complexity to the language.

If this one looks OK, then I'll close #4486.

### DIFF
--- a/working/2509-private-named-parameters/feature-specification.md
+++ b/working/2509-private-named-parameters/feature-specification.md
@@ -461,22 +461,48 @@ further ideas for additional warnings, lints, and quick fixes.
 
 ### API documentation generation
 
-Authors documenting an API that uses this feature should refer to the
-constructor parameter by its public name since that's what users will pass.
-Likewise, doc generators like [`dart doc`][dartdoc] should document the
-constructor's parameter with its public name. The fact that the parameter
-initializes or declares a private field is an implementation detail of the
-class. What a user of the class cares about is the corresponding public name for
-the constructor parameter.
+The reason a parameter has a private name is only a convenience for the
+maintainer of that constructor so that they can use an initializing formal or
+declaring parameter. A *user* of that constructor doesn't care about that
+implementation detail.
+
+Therefore, generated documentation from tools like [`dart doc`][] and in-IDE
+contextual help should show the public names for parameters. For named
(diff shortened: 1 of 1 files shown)


leafpetersen (approved):
LGTM

Issue #3097, opened 2023-05-22
Can't rethrow in switch expression


rubenferreira97:
Shouldn't be possible to `rethrow` in a switch expression since we can `throw e`?
Random example:
```dart
void main() {
  try {
    // some code that might throw
  } catch (e) {
    return switch (e) {
        Exception() => print('Return what this function asks for (void).'),
        Error() => throw e, //Works, but warning "Use 'rethrow' to rethrow a caught exception."
        _ => rethrow, // Error "Undefined name 'rethrow'"
    };
  }
}
```

munificent:
I've removed the "bug" tag since this is working as expected: `rethrow` is a statement and you can't use statements in switch expression bodies. In your example, the switch expression you have there doesn't do anything useful at all. It's in the body of a `return` but only the first case will ever complete normally, and `print()` evaluates to `void`. You're better off just writing:

```dart
void main() {
  try {
    // some code that might throw
  } catch (e) {
    switch (e) {
        case Exception(): return print('Return what this function asks for (void).');
        case Error(): throw e;
        case _: rethrow;
    }
  }
}
```

Switch expressions are not "the new better switch", they're switches you can use in expression position. In a statement position, like you have here, a switch statement is still the idiomatic approach.

Having said that, I can certainly imagine other places where you do have a switch in a catch body that returns values on most branches and where a switch expression would be useful. In those cases, it would be nice if we treated `rethrow` as an expression so you could use it here.

srawlins:
See also (and maybe upvote) https://github.com/dart-lang/language/issues/2025

rubenferreira97:
> Having said that, I can certainly imagine other places where you do have a switch in a catch body that returns values on most branches and where a switch expression would be useful. In those cases, it would be nice if we treated `rethrow` as an expression so you could use it here.

@munificent yes that code is not my use case. Something like this:

```dart
Handler middleware(Handler handler) {
  return (context) async {
    try {
      return await handler(context);
    } catch (e) {
      context.read<Logger>.e(e); // log request exception
      return switch (e) {
        ServiceException(): mapServiceExceptionToResponse(e),
        _: rethrow, // rethrow unknown exceptions, errors, etc...
      };
    }
  };
}
```

I guess I will rewrite it like:

```dart
Handler middleware(Handler handler) {
  return (context) async {
    try {
      return await handler(context);
    } catch (e) {
      context.read<Logger>.e(e); // log request exception
      switch (e) {
[...]

PR #4502, opened 2025-08-27
Document rethrow limitation in switch expressions (Issue #3097)

mes71:
### Motivation

This PR adds a short design note (`working/rethrow_in_switch_expression.md`) related to [[issue #3097](https://github.com/dart-lang/language/issues/3097)](https://github.com/dart-lang/language/issues/3097).

### Content

* Demonstrates why `rethrow` cannot be used in a `switch expression`.
* Shows a code example and the current compiler/analyzer messages.
* Outlines possible directions: (1) allow `rethrow` in expression contexts, (2) improve analyzer messages, or (3) document workarounds.

### Purpose

This is a documentation-only change that helps capture the problem and serves as a starting point for further design discussion.

### DIFF
--- a/working/3097 - rethrow_in_switch_expression/rethrow_in_switch_expression.md
+++ b/working/3097 - rethrow_in_switch_expression/rethrow_in_switch_expression.md
@@ -1,0 +1,51 @@
+# Rethrow in switch expressions
+
+**Related issue:** [#3097](https://github.com/dart-lang/language/issues/3097)
+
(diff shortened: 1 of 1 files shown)


munificent:
Thanks for the PR! We use `working/` only for complete language proposals, so this doesn't belong there. We could put it in `resources/` but I suspect it would end up forgotten. The best place for this kind of context is probably a comment on issue #3097.

PR #4511, opened 2025-09-04
Remove rules about DartDoc comments in the language specification

eernstg:
Update: This PR removes everything in the language specification that is concerned with documentation comments. The treatment of this topic area is now fully determined by each tool team.

---

Previous description:

The language specification specifies the syntax that makes a comment a DartDoc comment, which is unchanged. This PR adds a paragraph of 'rationale' text explaining that everything else is tool specific behavior, and it removes the rules about scoping in DartDoc comments.

There were no rules about the locations where a DartDoc comment can occur, which means that it is still "any location where a comment can occur".  Moreover, each tool can freely choose that a DartDoc comment is processed in a specific way when it occurs in a specific kind of location (including that it is ignored, or a diagnostic message is emitted).

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -42,9 +42,8 @@
 % release of this document.
 %
 % Sep 2025
-% - Remove the rules about scoping with DartDoc comments. This topic is now
-%   handled by each of the tools that process those comments in any way, and
-%   this document only specifies the syntax of DartDoc comments.
+% - Remove the rules about DartDoc comments. This topic is now handled by each
+%   of the tools that process those comments in any way.
 %
 % Jul 2025
 % - Clarify that operator `[]` and `[]=` are included in the `void` allowlist
@@ -23411,28 +23410,14 @@ are sections of program text that are used for documentation.
 Dart supports both single-line and multi-line comments.
 A \Index{single line comment} begins with the token \code{//}.
 Everything between \code{//} and the end of line
-must be ignored by the Dart compiler
-unless the comment is a documentation comment.
+must be ignored by the Dart compiler.
 
 \LMHash{}%
 A \Index{multi-line comment} begins with the token \code{/*}
 and ends with the token \code{*/}.
-Everything between \code{/}* and \code{*}/
-must be ignored by the Dart compiler
-unless the comment is a documentation comment.
+Everything between \code{/*} and \code{*/}
+must be ignored by the Dart compiler.
 Comments may nest.
-
-\LMHash{}%
-\IndexCustom{Documentation comments}{documentation comments}
-are comments that begin with the tokens \code{///} or \code{/**}.
-
-\rationale{%
-  Documentation comments are intended to be processed by
-  a tool that produces human readable documentation.
-  However, this document does not place any constraints
-  on the interpretation or use of documentation comments,
-  that is determined by each of the relevant tools.%
-}
 
 
 \subsection{Operator Precedence}
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -41,6 +41,11 @@
 % version of the language which will actually be specified by the next stable
 % release of this document.
 %
+% Sep 2025
+% - Remove the rules about scoping with DartDoc comments. This topic is now
+%   handled by each of the tools that process those comments in any way, and
+%   this document only specifies the syntax of DartDoc comments.
+%
 % Jul 2025
 % - Clarify that operator `[]` and `[]=` are included in the `void` allowlist
 %   rule about formal parameters of type `void`, and so are setters.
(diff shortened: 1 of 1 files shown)


sigurdm:
I wonder if it would be useful to specify which element a given documentation comment associates with.

I get the language team prefers to not decide everything here. But we need something for the teams to work with.
I propose that we somehow piggy-back on the specification on annotations: any place that can have an annotation can be preceded by a dartdoc comment that will then associate with that element.

eernstg:
> specify which element a given documentation comment associates with

We could say that any given DartDoc comment is associated with a syntactic element of kind `<someNonTerminal>` if it immediately precedes the leftmost token in the derivation from that non-terminal that yields the given program (ignoring whitespace, other comments, and metadata).

This might very well associate any given DartDoc comment with several distinct syntactic elements (because they all start with the same token), but there is no ambiguity unless the DartDoc comment is meaningful for more than one of those syntactic elements (all the others will just "ignore" that DartDoc comment).

However, all these decisions could also be made by the relevant tool team(s), which was the motivation for the change as proposed in this PR. Is it important to standardize explicitly on these things?

specification/dartLangSpec.tex
```diff
@@ -41,6 +41,11 @@
 % version of the language which will actually be specified by the next stable
 % release of this document.
 %
+% Sep 2025
+% - Remove the rules about scoping with DartDoc comments. This topic is now
+%   handled by each of the tools that process those comments in any way, and
+%   this document only specifies the syntax of DartDoc comments.
```
lrhn: Why specify the syntax of DartDoc comments?

To the language, they're just comments.

specification/dartLangSpec.tex
```diff
@@ -23420,25 +23425,14 @@ \subsubsection{Comments}
 \LMHash{}%
 \IndexCustom{Documentation comments}{documentation comments}
 are comments that begin with the tokens \code{///} or \code{/**}.
```
lrhn: So, just remove this too. 
A documentation comments is a concept introduced by, and interpreted by, the DartDoc tool.
The language doesn't know, or need to know, that it exists.

specification/dartLangSpec.tex
```diff
@@ -23420,25 +23425,14 @@ \subsubsection{Comments}
...
-Otherwise, the current scope for a documentation comment immediately preceding
-the declaration of a function $f$ is the formal parameter scope of $f$
-(\ref{formalParameters}).
+\rationale{%
+  Documentation comments are intended to be processed by
+  a tool that produces human readable documentation.
+  However, this document does not place any constraints
+  on the interpretation or use of documentation comments,
+  that is determined by each of the relevant tools.%
+}
```
lrhn: And this paragraph can be dropped too.

Saying nothing is better than saying something inconsequential or unnecessary.

lrhn (approved):

sigurdm:
> However, all these decisions could also be made by the relevant tool team(s), which was the motivation for the change as proposed in this PR. Is it important to standardize explicitly on these things?

If we expect users to use a feature, and multiple tools need to work with it, it needs to be specified/standardized. It doesn't have to be by the language spec per se, but I wonder where is a better place to spec this out...

eernstg:
The language team decided that we should have the syntax of the DartDoc comment in the language specification, but nothing about where it can occur and how it is determined which syntactic construct it is targeting. Now I hear requests for keeping more stuff in the language specification, and less. For now, I'll just keep the text as decided at that meeting. ;-)

bwilkerson:
> If we expect users to use a feature, and multiple tools need to work with it, it needs to be specified/standardized. It doesn't have to be by the language spec per se, but I wonder where is a better place to spec this out...

I agree, all of our tools need to support doc comments in the same way. The two tools that I know of that interpret doc comments are the analysis server and `dart doc`. All of the support for the analysis server is in the `analyzer` package, and most of the support for `dart doc` is too. There is a longstanding desire to move the remaining `dart doc` support into the `analyzer` package, because sharing code is the best way I know of to maintain consistency.

I have no objection to having a written specification of the behavior somewhere, and we minimally need to have user-facing documentation so that users know how to use the feature effectively.

I would recommend that we move this conversation off to a separate thread and make sure that the conversation involves all of the right participants (minimally the pub team, since they now own `dart doc`, the dart model team, who owns the `analyzer` package, and the developer experience team, who owns the analysis server).

eernstg:
@bwilkerson, I created https://github.com/dart-lang/language/issues/4513 in order to discuss the desired level of specification. This PR will lay dormant until https://github.com/dart-lang/language/issues/4513 yields a decision.

eernstg:
The newest commit removed everything in the language specification about documentation comments, which seems to be the approach that has general support.

specification/dartLangSpec.tex
eernstg: Removed.

specification/dartLangSpec.tex
eernstg: Removed.

specification/dartLangSpec.tex
eernstg: Removed.

specification/dartLangSpec.tex
```diff
@@ -23406,40 +23410,15 @@ \subsubsection{Comments}
...
 
 \LMHash{}%
 A \Index{multi-line comment} begins with the token \code{/*}
 and ends with the token \code{*/}.
-Everything between \code{/}* and \code{*}/
-must be ignored by the Dart compiler
-unless the comment is a documentation comment.
+Everything between \code{/*} and \code{*/}
+must be ignored by the Dart compiler.
 Comments may nest.
```
lrhn, line 23420: (That somewhat contradicts the "everything between `/*` and `*/` must be ignored". But not something to fix in this CL.)

eernstg: I had the same thought (both of them), but I don't think it's bad enough to look for a more precise wording. ;-)

PR #4514, opened 2025-09-10
Update spec for yield*.

lrhn:
Was very under-specified.

This change changes the responsiveness of a `yield*` to be that of `StreamController.addStream`, meaning that it *immediately* reacts to pause and cancel and forwards it to the inner stream. Unlike `await for`, a `yield*` is at the yield for the entire duration of the inner stream, and can cancel at any time, not only when control reaches a `yield` inside the loop.

Experience has taught us that with `async*` functions, `await for` and `yield*`, back-pressure using `cancel` or `pause` *must* be acted on immediately, to avoid a stream computation continuing when it has nothing meaningful to do, and the caller knows that and tries to stop it.

For example, if you do `stream.first`, it should cancel after the first event, before starting on the computation of the next event, because then it's too late to cancel. Canonical example:
```dart
// Stream with one event and no done event.
Stream<int> oneValue => Stream<int>.multi((c) => c.add(1));
Stream<int> clone1(Stream<int> stream) async* {
  await for (var event in stream) yield event;
}
Stream<int> clone2(Stream<int> stream) async* {
  yield* stream;
}
void main() async {
  print(await oneValue.first); // 1
  print(await clone1(oneValue).first);
  print(await clone2(oneValue).first);
  print(await clone2(clone1(oneValue)).first);
}
```
If the `first` code's cancel doesn't reach the `clone1` before it goes back to the `await for` loop, the cancel will never be processed because control never reaches a `yield` again.

---

- [ ] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>

</details>

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -19790,7 +19790,6 @@ The current call to \code{moveNext()} returns \TRUE.
 
 \LMHash{}%
 If $m$ is marked \code{\ASYNC*} (\ref{functions}), then:
-
 \begin{itemize}
 \item
   % This error can occur due to implicit casts.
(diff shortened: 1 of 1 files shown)


lrhn:
@eernstg WDYT?

specification/dartLangSpec.tex
```diff
@@ -19801,34 +19800,74 @@ \subsection{Yield-Each}
...
+  If the stream subscription $u$ associated with this execution of $m$
+  has been paused, suspend execution of $m$ until $u$ has been resumed or
+  cancelled.
+\item
+  If $u$ has been cancelled, execution of $s$ returns without a value.
+\item
+  The $o$ stream is listened to by calling its \code{listen} method,
+  creating a subscription $r$.
+\item
+\begin{itemize}
```
eernstg: This looks like an editing glitch. Perhaps:
```suggestion
  \item
```
and then `\begin{itemize}` might fit just before line 19815.

specification/dartLangSpec.tex
```diff
@@ -19801,34 +19800,74 @@ \subsection{Yield-Each}
...
+  Execution of $m$ is suspended. Until execution of $s$ completes in one of
+  the ways specified below, execution of $m$ occurs in the following cases:
+  \item If $u$ is cancelled, whether while waiting for an event from $r$,
```
eernstg, line 19815:
```suggestion
  \begin{itemize}
  \item If $u$ is cancelled, whether while waiting for an event from $r$,
```

specification/dartLangSpec.tex
```diff
@@ -19801,34 +19800,74 @@ \subsection{Yield-Each}
...
+    execution of $s$ throws the error \metavar{err} and stack trace \metavar{st}.
+    Otherwise execution of $s$ completes by returning without a value.
+  \item If $u$ becomes paused and $r$ is not already paused,
+    then pause $r$ by invoking its \code{pause} method with no arguments.
+    Then suspend execution of $m$ again.
     \commentary{%
-      Note that a dynamic error occurs if $x$ is added
-      and the dynamic type of $x$ is not a subtype of
-      the element type of said stream.%
+      No stream must emit events while it's paused, so no event from $r$
```
eernstg: Does that mean 'It is never required for a stream to emit events ...'? It could also mean 'It is prohibited for every stream to emit events ...'.

specification/dartLangSpec.tex
```diff
@@ -19801,34 +19800,74 @@ \subsection{Yield-Each}
...
+\begin{itemize}
+  Execution of $m$ is suspended. Until execution of $s$ completes in one of
+  the ways specified below, execution of $m$ occurs in the following cases:
+  \item If $u$ is cancelled, whether while waiting for an event from $r$,
+    to deliver an event to $u$ or to be resumed from pause:
+    Cancel $r$ by invoking its \code{cancel} method with no arguments,
+    returning a future $d$.
+    \commentary{%
+      No stream can be paused or resumed after being cancelled,
+      and no stream subscription must emit events after its \code{cancel}
```
eernstg: Confusing 'must', similar to line 19832. Would this work?:
```suggestion
      and no stream subscription can emit events after its \code{cancel}
```
.. or is it really "none of them is required to emit events..."?

specification/dartLangSpec.tex
```diff
@@ -19801,34 +19800,74 @@ \subsection{Yield-Each}
...
+      event asynchronously.
     }
-    The function $m$ may suspend.
+  \item If $u$ resumes from being paused, and $r$ is not currently paused
+    while asynchronously delivering an event to $u$, then resume $r$
+    by invoking its \code{resume} method with no arguments.
+    Then suspend execution of $m$ again.
+  \item If $r$ emits a value event with value $v$, then $u$ emits
+    a value event with value $v$,
+    and if $r$ empits an error event with error \metavar{err}
```
eernstg: Typo:
```suggestion
    and if $r$ emits an error event with error \metavar{err}
```

specification/dartLangSpec.tex
```diff
@@ -19801,34 +19800,74 @@ \subsection{Yield-Each}
...
-  If the stream $o$ is done, execution of $s$ completes normally.
 \end{itemize}
 
+\commentary{%
+The semantics here propagates pause and cancel requests directly
+to the nested stream subscription of the \code{\YIELD*}.
+That ensures that a pause or cancel request is responded to as soon as possible,
+to avoid the inner stream doing a larger computation to create a value that
+the outer stream already knows it doesn't need, or if paused,
+that it may not need.
```
eernstg: At this time the language specification has been adjusted to indent commentary and rationale, it would be nice to preserve that property also here:
```suggestion
  The semantics here propagates pause and cancel requests directly
  to the nested stream subscription of the \code{\YIELD*}.
  That ensures that a pause or cancel request is responded to as soon as possible,
  to avoid the inner stream doing a larger computation to create a value that
  the outer stream already knows it doesn't need, or if paused,
  that it may not need.
```

specification/dartLangSpec.tex
```diff
@@ -19790,7 +19790,6 @@ \subsection{Yield-Each}
 
 \LMHash{}%
 If $m$ is marked \code{\ASYNC*} (\ref{functions}), then:
-
```
eernstg, line 19793: Can't comment on line 19751, so I'll put it here: What happens if the evaluation of `e` in line 19751 does not complete normally? The interesting case would be `async*`.

specification/dartLangSpec.tex
lrhn: It is prohibited. "A stream must not" might be better. Changed to that.

specification/dartLangSpec.tex
lrhn: It's "must". They *can*, because we can't prevent them from calling a callback that we have given to them once, but if they do, they are seriously broken, and we do not say what happens.

Changed to "a stream must not" here too.

specification/dartLangSpec.tex
lrhn: The general rule is that if any step of the execution of a statement completes non-normally, unless something else is explicitly stated (like it is for `try`/`catch`), the execution of the statement completes in the same non-normal way.

So when `await for (var x in (throw "Banana")) { ... }` evaluates `throw "Banana"`, the `(throw "Banana")` statement and then the enitire `await for` statement completes throwing "Banana" and a stack trace.

PR #4515, opened 2025-09-12
Add shared native memory proposal

mraleph:
This extracts relevant parts of the bigger proposal into a separate proposal. 

The following things have changed:

* We now include formal memory model: our formalization pretty much follows JavaScripts with some minor tweaks.
* All atomic operations are consolidates in a single `AtomicInt` class - which can be constructed as a view on `Pointer`s or `TypedData` objects. This is a departure from the previous "zoo of extension methods" design. We should discuss if it is actually better. 
* I introduced `ScopedThreadLocal` class which should address some known problematic usages of static state in core libraries.
* There are few design points (marked with **TODO** in the text) which I don't have concrete answers to. We should discuss these. 

@lrhn @mkustermann @aam can you take a look? 

@aam could you give me the current state of core-library APIs (which already work from IG bound code, which you are fixing and which we will not support). I want to align on what we think is MVP here and include this into the proposal.

### DIFF
--- a/working/333 - shared memory multithreading/shared_native_memory.md
+++ b/working/333 - shared memory multithreading/shared_native_memory.md
@@ -143,54 +143,75 @@ We propose to reduce the distance between native code and Dart by:
   which utilizes **shared _native_ memory** without breaking existing
   Dart-level isolation boundaries.
 
-### Trivially shareable objects
+### Deeply immutable instances
 
-_Trivially shareable objects_ are objects which can be shared across isolate
-boundaries (within an isolate group) without breaking isolation of mutable
-Dart state.
+VM recognizes certain instances as [deeply immutable][] and supports sending
+these instances over isolate boundaries without copying because sharing
+such instances does not break the isolation guarantees. These instances are
+by construction guaranteed to neither contain nor reference (directly or
+indirectly) any mutable Dart state.
 
-- Instances of `String`, `int`, `double`, `bool` and `Null` are trivially shareable.
-- Instances of [deeply immutable][] types.
-- Instances of internal implementation of `SendPort`.
-- Functions which only capture variables annotated with `@pragma('vm:shared')`
-  (this includes function which do not capture any variables e.g. tear-offs of
-  static methods).
-- Compile time constants (instances produced by evaluation of `const` expressions).
-- Instances of `TypedData`.
-- Instances of `Pointer` and `Struct`.
+However current definition of deeply immutable types is not general enough to
+support all interesting patterns, so we need to expand it to cover additional
+cases.
 
-> [!WARNING]
->
-> Not all _trivially shareable objects_ can be purely distinguished by their
-> static type - `SendPort` is an example which requires runtime checking.
+A _deeply immutable type_ is a type for which all instances of this type
+are _deeply immutable_.
+
+The following builtin types are deeply immutable: `bool`, `double`, `int`,
+`Null`, `String`, `Float32x4`, `Float64x2`, `Int32x4`, `Pointer`.
+
+Classes providing builtin implementations of the following public core library
+classes are also deeply immutable: `SendPort`, `Capability`, `RegExp`,
+`StackTrace`, `Type`.
+
+All compile time constants are deeply immutable instances.
+
+`TypedData` and `Struct` instances are deeply immutable when backed by native
+(external) memory.
+
+Unmodifiable lists (`List.unmodifiable`) which contain deeply immutable
+instances are deeply immutable.
+
+Closures which capture only `final` variables containing deeply immutable
+instances are deeply immutable.
+
+Finally, instances of classes annotated with `@pragma('vm:deeply-immutable')`
+are deeply immutable. It is a compile error if classes annotated with this
+pragma contain non-`final` fields. It is an compile time error if static
+type of field within annotated class excludes deeply immutable instances.
+If the static type of a field in a deeply immutable class is not
+deeply immutable type - then compiler must insert checks in the constructor to
+guarantee that this field is initialized to a deeply immutable value.
 
 > [!IMPORTANT]
 >
-> **TODO** should we allow sharing of all `TypedData` objects? This seems very
-> convenient. There is a consideration here that not all `TypedData` instances
-> can be shared on the Web, where there is separation between `ArrayBuffer`
-> and `SharedArrayBuffer` exists.
->
-> **TODO** should we allow sharing instances of immutable `List`s which only
-> contain trivially shareable objects? This in general case requires `O(1)`
-> check, but might be very convenient as well.
+> **TODO** should we allow sharing of all `TypedData` (and by extension
+> `Struct`) objects? This seems very convenient. There is a consideration
+> here that not all `TypedData` instances can be shared on the Web, where
+> there is separation between `ArrayBuffer` and `SharedArrayBuffer` exists.
 
 [deeply immutable]: https://github.com/dart-lang/sdk/blob/bb59b5c72c52369e1b0d21940008c4be7e6d43b3/runtime/docs/deeply_immutable.md
 
 ### Shared fields and variables (`@pragma('vm:shared')`).
 
 Static fields and global variables annotated with `@pragma('vm:shared')` are
-shared across all isolates in the isolate group - updating a field from one
-isolate.
+shared across all isolates in the isolate group.
 
 A field or variable annotated with `@pragma('vm:shared')` can only contain
-values which are trivially shareable objects.
+values which are deeply immutable objects.
 
 * It is a compile time error to annotate a field or variable the static type of
-  which excludes trivially shareable objects;
-* If static type of a field is a super-type for both trivially shareable and
-  non-trivially shareable objects then compiler will insert a runtime check
-  which ensures that values assigned to such field is trivially shareable.
+  which excludes deeply immutable objects;
+* If static type of a field is a super-type for both deeply immutable and
+  non-deeply immutable objects then compiler will insert a runtime check
+  which ensures that values assigned to such field are deeply immutable.
+* A field or variable annotated with `@pragma('vm:shared')` must be `final`.
+
+> [!NOTE]
+>
+> Restrictions imposed above are the same as ones imposed on field in deeply
+> immutable classes.
 
 Shared fields must guarantee atomic initialization: if multiple threads
 access the same uninitialized field then only one thread will invoke the
@@ -227,7 +248,7 @@ will cause `FieldAccessError` to be thrown.
   /// the [callback].
   ///
   /// [callback] and [exceptionalReturn] must be
-  /// _trivially shareable_.
+  /// _deeply immutable_.
   ///
   /// This callback must be [close]d when it is no longer
   /// needed. An [Isolate] that created the callback will
@@ -287,9 +308,9 @@ class Isolate {
   /// Throws [StateError] if target isolate is owned by another thread and
   /// thus can't be entered from a different thread.
   ///
-  /// Throws [ArgumentError] if [f] is not trivially shareable.
+  /// Throws [ArgumentError] if [f] is not deeply immutable.
   ///
-  /// Throws [StateError] if result returned by [f] is not trivially shareable.
+  /// Throws [StateError] if result returned by [f] is not deeply immutable.
   R runSync<R>(R Function() f, {Duration? timeout});
 }
 ```
@@ -307,7 +328,7 @@ into one update.
 final class ScopedThreadLocal<T> {
   /// Creates scoped thread local value with the given [initializer] function.
   ///
-  /// [initializer] must be trivially shareable.
+  /// [initializer] must be deeply immutable.
   external factory ScopedThreadLocal([T Function()? initializer]);
 
   /// Execute [f] binding this [ScopedThreadLocal] to the given [value] for the duration of the execution.
@@ -769,7 +790,7 @@ observe object in partially initialized state. Implementations can choose to
 guarantee this property by inserting appropriate barriers when creating objects,
 however that would be a waste for objects that are mostly used in an
 isolate-local manner. Instead, given current restriction that only
-trivially-shareable (deeply immutable objects) can be placed into shared-fields
+deeply immutable objects can be placed into shared-fields
 implementations can instead choose to implement shared fields using
 _store-release_ and _load-acquire_ atomic operations. This would guarantee
 happens-before ordering for initializing stores. We however do not _require_


mraleph:
cc @leafpetersen

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -0,0 +1,775 @@
...
+class Isolate {
+  // Execute the given function in the context of the given isolate.
+  // 
+  // [f] must be trivially shareable. Result
+  // returned by [f] must be trivially shareable.
+  R runSync<R>(R Function() f);
+}
+```
+
+Note that youou can only enter `Isolate` when it is not used by another thread.
```
aam: "you"

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -0,0 +1,775 @@
...
+
+// Open the panel and return. When user selects a file
+// the passed block will be invoked.
+[panel beginWithCompletionHandler: ^(NSInteger result){
+   // Handle the result.
+}];
+```
+
+Trying to port this code to Dart hits the following issue: you can only use this
+API on the UI thread and Dart's main isolate is not running on the UI thread.
```
aam: Is this outdated now given that ui and platform threads are merged on flutter https://github.com/flutter/flutter/issues/150525 ?

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -0,0 +1,775 @@
...
+our own thread pool and provide equivalents of `Dart_SetMessageNotifyCallback`
+and `Dart_HandleMessage`. Though maybe we should not bundle this all together
+into one update.
+
+### Scoped thread local values
+
+```dart
+@pragma('vm:deeply-immutable')
+final class ScopedThreadLocal<T> {
+  /// Creates scoped thread local value with the given [initializer] function.
```
aam, line 329: Nice, but outside of isolate-group-bound callbacks, in the context of standard isolates, does `ScopedThreadLocal` instance behave like a `static` field? Isolates in the end can migrate from one thread to another.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: Deleted this section.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: Note that isolates can't move between threads in *synchronous* calls and this API is fully synchronous.

aam: Will there be something that would prevent this infrastructure from being used in asynchronous context?
```
 static Future<String> iterableToShortString(...) {
  return toStringVisiting.use((toStringVisitingValue) async {
    await Future.delayed(Duration(seconds: 10));
    ...
```

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -0,0 +1,746 @@
...
+  ///
+  /// [initializer] must be trivially shareable.
+  external factory ScopedThreadLocal([T Function()? initializer]);
```
aam, line 332: In the sample use below at line 344 `List<Object>` is not trivially shareable, right?

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: You _can_ use it. But it is not going to do anything useful because the value of the scoped thread local will reset once the synchronous execution completes.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: There is no value of type `List<Object>` - the closure (`initializer`) is trivially shareable because it does not capture anything.

working/333 - shared memory multithreading/shared_native_memory.md
aam: So `toStringVisiting` can become unbound after `await Future.delayed(Duration(seconds: 10))` (in unlikely event) of isolate switching to a different thread? Realistically we are keeping isolate/(vm)thread affinity, so switching should not happen.

mraleph: It is not related to moving between threads at all. `ScopedThreadLocal` are valid within the scope of synchronous execution (hence the name). So if you do `stl.use(f)` or `stl.with(value, f)` once synchronous execution of this expression completes `stl` (e.g. `use` or `with` completes) is unbound.

aam: I was curious about the case when `f` is asynchronous. Would that result in `ScopedThreadLocal` values bound only for initial synchronous part of `f` execution? In the example above - before `await Future.delayed(Duration(seconds: 10));`?

mraleph:
cc @alexmarkov for memory model feedback

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -0,0 +1,778 @@
...
+  // Execute the given function in the context of the given isolate.
+  // 
+  // [f] must be trivially shareable. Result
+  // returned by [f] must be trivially shareable.
+  R runSync<R>(R Function() f);
+}
+```
+
+Note that `runSync` can only enter an `Isolate` when it is not used by
+another thread.
```
lrhn: How will the caller be able to know that?

The other isolate must be back at the event loop, and probably not just the microtask loop.
I guess the isolate can set a flag when its event loop is empty and it's waiting for new events.

You could potentially have a way to check that flag, but that won't guarantee that the isolate didn't get and started processing an event immediately after you checked.

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -0,0 +1,778 @@
...
+our own thread pool and provide equivalents of `Dart_SetMessageNotifyCallback`
+and `Dart_HandleMessage`. Though maybe we should not bundle this all together
+into one update.
+
+**TODO**: Should we facilitate execution of asynchronous code inside a target
+isolate e.g. something like `T runAsync<T>(FutureOr<T> Function() f)`. When
+you invoke this function `f` will be executed in the context of the given
+isolate and then if it produced a `Future` we exit the isolate and block
+the current thread, while allowing the event loop for that isolate to run
+normally until returned future produces result.
```
lrhn: That can be implemented using a synchronous/blocking receive port.

It's possible that there is a use for this, but it's rarely a good idea to block your own isolate.

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -0,0 +1,778 @@
...
+
+  /// Execute [f] binding this [ScopedThreadLocal] to the given [value] for the duration of the execution.
+  external R with<R>(T value, R Function(T) f);
+  
+  /// Execute [f] initializing this [ScopedThreadLocal] using default initializer if needed.
+  /// Throws [NotBoundError] if this [ScopedThreadLocal] does not have an initializer.
+  external void withInitialized<R>(R Function(T) f);
+
+	/// Returns the value specified by the closest enclosing invocation of [with] or
+	/// throws [NotBoundError] if this [ScopedThreadLocal] is not bound to a value.
```
lrhn: Can just use `StateError`.

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -0,0 +1,778 @@
...
+  external factory ScopedThreadLocal([T Function()? initializer]);
+
+  /// Execute [f] binding this [ScopedThreadLocal] to the given [value] for the duration of the execution.
+  external R with<R>(T value, R Function(T) f);
+  
+  /// Execute [f] initializing this [ScopedThreadLocal] using default initializer if needed.
+  /// Throws [NotBoundError] if this [ScopedThreadLocal] does not have an initializer.
+  external void withInitialized<R>(R Function(T) f);
+
+	/// Returns the value specified by the closest enclosing invocation of [with] or
```
lrhn, line 345: I expect I'll use the pattern: 
```dart
String toString() {
  if (!_toStringVisiting.isBound) {
    return _toStringVisiting.runWith([], (_) => toString());
  }
  var visiting = _toStringVisiting.value;
  //...
}
```
in every function that depends on the value. 

It can't all be abstracted into a helper method like it's it's today, the initialization must be on the main trunk of the execution tree.

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -0,0 +1,778 @@
...
+  ///
+  /// [initializer] must be trivially shareable.
+  external factory ScopedThreadLocal([T Function()? initializer]);
+
+  /// Execute [f] binding this [ScopedThreadLocal] to the given [value] for the duration of the execution.
+  external R with<R>(T value, R Function(T) f);
+  
+  /// Execute [f] initializing this [ScopedThreadLocal] using default initializer if needed.
+  /// Throws [NotBoundError] if this [ScopedThreadLocal] does not have an initializer.
+  external void withInitialized<R>(R Function(T) f);
```
lrhn: I think `with` may be a reserved word. I think `runWith` would be the traditional name.

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -0,0 +1,778 @@
...
+>
+> - We need to reimplement locks using some low-level primitives because
+> unlocking mutex on a different thread is often not supported by C level APIs.
+>
+> - It expands the duration of lock in a complicated unpredictable way
+> (if we take lock before dispatching the future completion event). However
+> using locks without asynchronous versions might be hard in a language like
+> Dart - it would require completely blocking event loop to acquire the lock.
+
+### Atomics
```
lrhn, line 466: Is this only for FFI?
If not, when should one use them?
Only if you've shared the typed data with another isolate? Always then?

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -0,0 +1,778 @@
...
+  
+  /// Execute [f] initializing this [ScopedThreadLocal] using default initializer if needed.
+  /// Throws [NotBoundError] if this [ScopedThreadLocal] does not have an initializer.
+  external void withInitialized<R>(R Function(T) f);
+
+	/// Returns the value specified by the closest enclosing invocation of [with] or
+	/// throws [NotBoundError] if this [ScopedThreadLocal] is not bound to a value. 
+  external T get value;
+  
+  /// Returns `true` if this [ScopedThreadLocal] is bound to a value.
```
lrhn: "Whether" 😁

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -0,0 +1,778 @@
...
+}
+```
+
+This makes `Iterable.toString` unusable outside of an isolate (i.e. within
+`isolateGroupBound` callables). However this code can be rewritten like so:
+
+```dart
+/// A collection used to identify cyclic lists during `toString` calls.
+@pragma('vm:shared')
+final ScopedThreadLocal<List<Object>> toStringVisiting = ScopedThreadLocal<List<Object>>(() => <List<Object>>[]);
```
lrhn: `… => <Object>[];`

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -0,0 +1,778 @@
...
+      // ... 
+    } finally {
+      toStringVisitingValue.removeLast();
+    }
+    // ...
+  });
+}
+```
+
+### Synchronization Primitives
```
lrhn, line 425: What are use-cases?

Communicating between isolates?

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -0,0 +1,778 @@
...
+\begin{align*}
+Loc_r(r) &\triangleq [\mathtt{Addr}(r), \mathtt{Addr}(r) + |\mathtt{Data_r}(r)|) \\
+Loc_w(r) &\triangleq [\mathtt{Addr}(r), \mathtt{Addr}(r) + |\mathtt{Data_w}(r)|) \\
+Loc(e) &\triangleq Loc_r(e) \cup Loc_w(e) \\
+Overlap(a, b) &\triangleq Loc(a)\cap Loc(b) \neq \emptyset \\
+SameLoc_{wr}(w, r) &\triangleq Loc_w(w) = Loc_r(r) \\
+SameLoc_{ww}(w, w') &\triangleq Loc_w(w) = Loc_w(w') \\
+Write(e) & \triangleq Loc_w(e) \neq \emptyset \\
+Unordered(e) & \triangleq \neg(\mathtt{Atomic(e)} \vee \mathtt{Init}(e))
+\end{align*}
```
lrhn, line 589: Can't review this, because I can't read how it looks. 😔
OK, I copied it into a markdown renderer.

Now I don't know what "&triangleq;" means. Is this defining functions or relations?
Says "functions and predicates", so probably functions (some to Bool).

The point of markdown is that the source is readable. LaTeX math isn't, so it feels like it's missing the point.

I was going to suggest using Unicode instead, but `&vee;` isn't much better then `\vee`.

Consider just using ASCII and words:
> * *LocR*(*r*) := [**Addr**(*r*), **Addr**(*r*),  + |**DataR**(_r_)|]
> * *LocW*(*r*) := [**Addr**(*r*), **Addr**(*r*),  + |**DataW**(_r_)|]
> * *Loc*(_e_) := **Union**(LocR(_e_), LocW(_e_))
> * *Overlap*(_e_) := **Intersection**(LocR(_e_), LocW(_e_))
> * *SameLocWR*(_w_, _r_) := *LocW*(_w_) = *LocR*(_r_)
> * *SameLocWW*(_w_, _w_’) := *LocW*(_w_) = *LocW*(_w_’)
> * *Write(e)* := *LocW*(_e_) is not empty
> * *Unordered(e)* := not **Atomic**(_e_) and not **Init**(_e_)

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -0,0 +1,778 @@
...
+```dart
+@pragma('vm:deeply-immutable')
+final class ScopedThreadLocal<T> {
+  /// Creates scoped thread local value with the given [initializer] function.
+  ///
+  /// [initializer] must be trivially shareable.
+  external factory ScopedThreadLocal([T Function()? initializer]);
+
+  /// Execute [f] binding this [ScopedThreadLocal] to the given [value] for the duration of the execution.
+  external R with<R>(T value, R Function(T) f);
```
aam: perhaps `withValue` since `with` is a keyword?

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -0,0 +1,778 @@
...
+  }
+  return false;
+}
+
+static String iterableToShortString(
+  Iterable iterable, [
+  String leftDelimiter = '(',
+  String rightDelimiter = ')',
+]) {
+  return toStringVisiting.use((toStringVisitingValue) {
```
aam: `  return toStringVisiting.withInitialized((toStringVisitingValue) {` ?

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -0,0 +1,778 @@
...
+programs behave in sequentially consistent way.
+
+#### Language and Library Semantics
+
+#### Object Construction
+
+When object is constructed all field initializers (i.e. those provided in the
+body of the class or in the initializer list) result in $\mathtt{Init}$
+write events. The same applies to initialization of individual elements of
+`TypedData` objects with `0`.
```
lrhn: And platform lists? (Probably want all object creation to act like initialization, also for collection literals.)

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -0,0 +1,778 @@
...
+
+#### Isolates
+
+Every `SendPort.send` generates a numbered event $\mathtt{Send}(p, i)$.
```
lrhn: Is the number arbitrary or monotonically increasing?
(The "numbered" makes it sound like the numbers are 1, 2, ..., but it doesn't actually say that.)

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -0,0 +1,778 @@
...
+
+Receiving a message (e.g. invoking callback attached to  `ReceivePort`)
+generates a corresponding $\mathtt{Recv}(p, i)$ event.
+
+Sends are synchronized with receives:
+
+$$
+\forall p\,i . \mathtt{Send}(p, i) \leq_\mathtt{asw}\mathtt{Recv}(p, i)
+$$
```
lrhn, line 761: Do we also want:
> &forall; _p_, _i_, _j_, Send(_p_, _i_) &le;<sub>asw</sub> Send(_p_, _j_) &Rarr; Recv(_p_, _i_) &le;<sub>asw</sub> Recv(_p_, _j_)

? That is, sends *on the same port* are not reordered, but that's proably only true if the sends are in the same isolate.

working/333 - shared memory multithreading/shared_native_memory.md
aam: Can't you use `withInitialized` instead?

Given 
```dart
@pragma('vm:shared')
final toStringVisiting = ScopedThreadLocal<List<Object>>(() => <Object>[]);
```
you can write
```dart
String toString() {
  return toStringVisiting.withInitialized((toStringVisitingValue) {...})
}
```

which will only invoke initializer if (isolate-group) `toStringVisiting` instance of `ScopedThreadLocal`  is not bound on this thread yet.

lrhn: I don't want to double the number of stack frames and generate extra closures if I can avoid it.
I don't expect the call to be a tail-call.

mraleph:
Pushed updates based on @aam and @lrhn comments. 

PTAL

mraleph:
@leafpetersen @lrhn I want to discuss something, which kind of goes against our original agreement on how to scope this feature. 

The current state of the proposal is the following: we have _trivially shareable objects_ which are _mostly_ deeply immutable Dart objects and we have _shared static fields and captured variables_ which can only contain trivially shareable objects and are shared between concurrent threads of execution. 

This setup obviously contains shared mutable state in a very round-about way: _closures_. We all know very well that closures and objects are isomorphic, in a sense that you can build an object system out of closures using captured variables to store object state. 

As such I think we are standing on a fork in the road.

We can choose to close the loophole: require that only `final` variables can be annotated as shared. That means neither shared statics nor shared captured variables can be mutated. Proceed with the original plan.

We can choose to open the floodgate: forgo the trivial shareability restriction altogether and allow _shared mutable Dart state_.  

Our current plan is very defensive: gradually remove isolation guarantee, see how people use that and then possibly remove the isolate entirely. Maybe intermediate step does not make much sense anyway - because it does not solve sufficiently enough user challenges (it is solely focused on native interop and requires users to move all of their data into the native heap if they want to use multicore capabilities). Maybe we should take the whole step instead of taking half step.

WDYT?

aam (approved):

mraleph:
Okay, so here is where we are:

@aam has done some initial measurements for the costs of going "share-everything" on platforms with weak memory model. Initial data shows regressions which will require non-trivial work to investigate and address, so I think we are not going to put all our focus on doing that. Instead we are going to continue with the current plan of shipping _specifically_ subset of features defined in this proposal and then return to "share-everything". 

I have updated the proposal to remove the concept of _trivially shareable_ entirely, instead it now says that all deeply immutable objects can be shared - and that's it. The definition of deeply immutable is expanded to cover closures and allow for runtime checking - for the cases where we can't guarantee deep immutability purely based on the static type. Deeply immutable closures are required to only capture final variables (containing deeply immutable objects). I however removed requirement to annotate these variables as `vm:shared` as this seems unnecessary now (given that they are immutable).

I think that concludes necessary updates to the proposal and brings to its final state. 

@aam @lrhn I would appreciate if you take a look at the last changes and voice any concerns if you have them.

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -0,0 +1,816 @@
...
+If the static type of a field in a deeply immutable class is not
+deeply immutable type - then compiler must insert checks in the constructor to
+guarantee that this field is initialized to a deeply immutable value.
+
+> [!IMPORTANT]
+>
+> **TODO** should we allow sharing of all `TypedData` (and by extension
+> `Struct`) objects? This seems very convenient. There is a consideration
+> here that not all `TypedData` instances can be shared on the Web, where
+> there is separation between `ArrayBuffer` and `SharedArrayBuffer` exists.
```
aam, line 192: "there exists a separation between ..." ?

aam (approved):

mraleph:
friendly ping @lrhn

Issue #4082, opened 2024-09-05
Simple parts with imports


eernstg:
I'm worried about the complexity of the scoping structure with the current proposal for enhanced parts (aka [parts with imports](https://github.com/dart-lang/language/blob/main/working/augmentation-libraries/parts_with_imports.md)).

## Currently specified scoping structure

For example, assume the following Dart files (based on an example from @lrhn):

```dart
// --- Library 'main.dart'.
import 'l1.dart' show v1, v2, v3;
import 'q1.dart' as q show q1;
import 'p1a.dart' as p show p1, p2;

part 'part.dart';

void foo() {}
```

```dart
// --- Part file 'part.dart'.
part of 'main.dart';

import 'l2.dart' show v1;
import 'p1b.dart' as p show p1, p3;
import 'p1c.dart' as p show p4;

void bar() {}

const v3 = "a";
```

The imported libraries are not shown, but every name in their exported namespaces are shown using `show` clauses, so this is enough to know which names imported where.

```mermaid
graph BT;
  MainLibrary["'main.dart' top-level scope<br>foo, bar, v3"]
  MainPrefix["'main.dart' prefix scope<br>p, q"]
  MainImport["'main.dart' import scope<br>v1, v2, v3"]
  MainLibrary --> MainPrefix
  MainPrefix --> MainImport
  PartLibrary["'part.dart' top-level scope<br>foo, bar, v3"]
  PartPrefix["'part.dart' prefix scope<br>p"]
  PartImport["'part.dart' import scope<br>v1"]
  PartLibrary --> PartPrefix
  PartPrefix --> PartImport
  PartImport --> MainPrefix
  MainPrefixQ["'main.dart' prefix scope 'q'<br>q1"]
  MainPrefix -->|q| MainPrefixQ
  MainPrefixP["'main.dart' prefix scope 'p'<br>p1, p2"]
  MainPrefix -->|p| MainPrefixP
  PartPrefixP["'part.dart' prefix scope 'p'<br>p1, p3, p4"]
  PartPrefix -->|p| PartPrefixP
  PartPrefixP --> MainPrefixP

  style MainLibrary fill:lightgreen;
  style PartLibrary fill:lightgreen;
  style MainPrefixP fill:yellow;
  style MainPrefixQ fill:yellow;
  style PartPrefixP fill:yellow;
  style MainImport fill:lightblue;
  style PartImport fill:lightblue;
```

[...]

PR #4520, opened 2025-09-26
Update enhanced-parts/feature-specification.md

lrhn:
Remove configurable part directives.

General tweaks and phrasing fixes, and some typos fixed.

Mention extension conflicts.
(You can break sub-parts by importing an extension that causes a conflict.
You can protect a part against that by using explicit extension applications.
Most people won't ever notice, extension conflicts are rare.)

### DIFF
--- a/working/parts-with-imports/feature-specification.md
+++ b/working/parts-with-imports/feature-specification.md
@@ -2,9 +2,9 @@
 
 Authors: rnystrom@google.com, jakemac@google.com, lrn@google.com
 
-Version: 1.2 (See [Changelog](#Changelog) at end)
+Version: 1.3 (See [Changelog](#Changelog) at end)
 
-Experiment flag: parts-with-imports
+Experiment flag: enhanced-parts
 
 This is a standalone definition of _enhanced part files_, where the title of
 this document/feature is highlighting only the most prominent part of the
@@ -20,7 +20,7 @@ meta-programming and augmentations. It's defined as a stand-alone feature, but
 design choices were made based on the augmentations and macro features,
 combined with being backwards compatible.
 
-[Augmentations]: ../augmentations.md "Augmentations feature specification"
+[Augmentations]: ../augmentations/feature-specification.md "Augmentations feature specification"
 
 ## Motivation
 
@@ -115,7 +115,7 @@ The design goals and principles are:
         `import`s back into its parent file.
 
     _(Augmentations modify both of these properties slightly, because order of
-    declarations also matter.)_
+    declarations may also matter.)_
 
 *   *Library member declarations are library-global*: All top-level declarations
     in the library file and all transitive part files are equal, and are all in
@@ -131,21 +131,32 @@ The design goals and principles are:
     name conflicts between declarations in separate tree-branches of the
     library structure.
 
-*   *Import inheritance is a only suggestion*: Aka. other files' imports cannot
-    break your code (at least if you're not depending on them). A part file is
-    never restricted by the imports it inherits from its parent file. It can
-    ignore and override all of them with imports of its own. That allows a
-    file, like a macro generated file, to import all its own dependencies and
-    be completely self-contained when it comes to imports. _It still needs to
-    fit into the library and not conflict with existing top-level names. That's
-    why a macro should document any non-fresh names it introduces, so a library
-    using the macro can rename any declarations that would conflict._
+*   *Import inheritance is a only suggestion*: Aka. ancestor files' imports
+    shouldn't break your code (if you're not depending on them).
+    A part file is never _restricted_ by the imports inherited from its parent file.
+    It can ignore and override all of them with imports of its own, and by making
+    explicit references to its own imported names.
+    That allows such a file, for example code-generated file, to import all its
+    own dependencies and be completely self-contained when it comes to imports.
+    _It still needs to fit into the library and not conflict with existing
+    top-level names. That's why a code generator should document any non-fresh
+    names it introduces, so a library using the macro can rename any declarations
+    that would conflict._
+    Because adding more extension declarations can affect the resolution of
+    implicit extension applications, a part file which aims to be completely
+    independent of its parent imports, should import its own extensions
+    *and* use only explicit extension applications of those.
+    _That is something code generation may want to do anyway. It doesn't care
+    about the verbosity, and it's safe against any later changes._
 
-    *   Because of that, it's possible to convert an existing library into a
-        part file of another library. Since a library is self-contained and
-        imports all external names that it refers to, making it a part file will
-        not cause any conflict due to inherited imports. _(Obviously still need
-        to avoid conflicts with top-level declarations.)_
+    *   The ability to be indpendent of parent file imports should make it
+        possible to convert an existing library into a part file of another library.
+        Since a library is self-contained and imports all external names that it
+        refers to, making it a part file will not cause any conflict due to
+        inherited imports. _(Obviously still need to avoid conflicts with
+        top-level declarations, and it may need to make extension applications
+        explicit in the rare case when the library already contains a different
+        applicable extension that is equally or more specific.)_
     *   And similarly, if a part file *is* self-contained, it can be converted
         into a separate library and imported back into the original library, or
         it can be moved to another position in the part tree hierarchy. _(Again
@@ -155,13 +166,9 @@ The design goals and principles are:
 ### Grammar
 
 We extend the grammar of part files to allow `import`, `export` and `part` file
-directives. We allow `part` files directives to use a configurable URI like the
-other two. We restrict the `part of` directive to only allow the string version.
+directives. We restrict the `part of` directive to only allow the string version.
 
 ```ebnf
--- Changed "<uri>" to "<configurableUri>".
-<partDirective> ::= <metadata> `part' <configurableUri> `;'
-
 -- Removed "<dottedIdentifier>" as option, retaining only "<uri>".
 <partHeader> ::= <metadata> `part' `of' <uri> `;'
 
@@ -174,7 +181,7 @@ other two. We restrict the `part of` directive to only allow the string version.
 The grammar change is small, mainly adding `import`, `export` and `part`
 directives to part files.
 
-The change to `part of` directives to not allow a dotted name was made because
+The change to `part of` directives, to not allow a dotted name, was made because
 we want a part file of a part file to refer back to its parent part file, but a
 dotted library name can only refer to a library. _That doesn't mean that
 part-of-dotted-name cannot be supported for part files that are part files of a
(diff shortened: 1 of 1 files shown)


eernstg:
Maybe "assigned" should be "reviewer"?

lrhn:
They should! And now they are.

eernstg:
Thanks! I won't get around to it today, but tomorrow should work.

working/parts-with-imports/feature-specification.md
```diff
@@ -131,21 +131,32 @@ The design goals and principles are:
...
+    shouldn't break your code (if you're not depending on them).
+    A part file is never _restricted_ by the imports inherited from its parent file.
+    It can ignore and override all of them with imports of its own, and by making
+    explicit references to its own imported names.
+    That allows such a file, for example code-generated file, to import all its
+    own dependencies and be completely self-contained when it comes to imports.
+    _It still needs to fit into the library and not conflict with existing
+    top-level names. That's why a code generator should document any non-fresh
+    names it introduces, so a library using the macro can rename any declarations
+    that would conflict._
```
eernstg: I think the scenario would be that a developer is writing or updating the library (and any manually written parts). The developer would then know which non-fresh names are being introduced by code generation, and the developer would avoid name clashes by using other names in the manually maintained code. So perhaps:

... non-fresh names it introduces, such that the maintainers of the manually written code in that library can avoid name clashes.

By the way, we shouldn't say 'macro' anywhere.

working/parts-with-imports/feature-specification.md
```diff
@@ -131,21 +131,32 @@ The design goals and principles are:
...
+    Because adding more extension declarations can affect the resolution of
+    implicit extension applications, a part file which aims to be completely
+    independent of its parent imports, should import its own extensions
```
eernstg: There shouldn't be a comma after a restrictive clause:
```suggestion
    independent of its parent imports should import its own extensions
```

working/parts-with-imports/feature-specification.md
```diff
@@ -131,21 +131,32 @@ The design goals and principles are:
...
+    *and* use only explicit extension applications of those.
+    _That is something code generation may want to do anyway. It doesn't care
+    about the verbosity, and it's safe against any later changes._
+
+    *   The ability to be indpendent of parent file imports should make it
```
eernstg: Typo:

```suggestion
    *   The ability to be independent of parent file imports should make it
```

working/parts-with-imports/feature-specification.md
```diff
@@ -131,21 +131,32 @@ The design goals and principles are:
...
+        possible to convert an existing library into a part file of another library.
+        Since a library is self-contained and imports all external names that it
+        refers to, making it a part file will not cause any conflict due to
```
eernstg, line 156: Great point!

working/parts-with-imports/feature-specification.md
```diff
@@ -131,21 +131,32 @@ The design goals and principles are:
...
+        inherited imports. _(Obviously still need to avoid conflicts with
```
eernstg: Wouldn't "We obviously" be more natural than the plain "Obviously"?

working/parts-with-imports/feature-specification.md
```diff
@@ -131,21 +131,32 @@ The design goals and principles are:
...
+        top-level declarations, and it may need to make extension applications
```
eernstg: "it" won't do anything on its own, but "we" could do it.

working/parts-with-imports/feature-specification.md
```diff
@@ -183,107 +190,121 @@ been linting against its use for quite a while already. Dotted names in part-of
...
-declaration.
+Imports, and import prefixes, of a parent file can now be shadowed by
+part files. To support that, each file gets its own import scope,
+import prefix scope and top-level scope.
+
+A _pre-feature library_ has only an _import scope_ with the namespace of
+all names imported by non-prefixed imports, which is the parent scope of the 
+*top-level scope* with a namespace containing all top-level declarations of
+the library, from both library file and part files, _and_ all prefix names
+of prefixed imports in the library file.
```
eernstg: That's not exactly how it is specified.

The imported unprefixed namespace except the ones with names that are declared at the top level of the library as well as a binding for each import prefix (which may be used in one or more non-deferred import directives) are included in the library scope, along with all the top-level declarations of the library.

This is similar to having an enclosing scope, especially because we're filtering out the imported names that are also names declared by the library itself.

However, if these import related bindings had instead been an enclosing scope then we'd have no error for a prefix with the same name as a top-level declaration, it would just be inaccessible. We could add a special exception and say that "any name which is shadowed in this particular scope is an error", but this just confirms that it isn't simply yet another enclosing scope.

Also, we need to special case import prefixes because there is no error for multiple non-deferred import directives with the same prefix, but there is an error when one or more of them is deferred. Similarly, we need to special case imported names (prefixed or not), because name clashes are handled by introducing an error entry in the bindings and reporting an error when that name is _used_, and silently accepting that the name clash exists in the first place. None of the scopes of Dart as specified have this kind of behavior.

I believe we do have to treat the import related bindings in a specialized manner, and we do need to be careful that we don't make changes that we don't actually need or want if we re-state the meaning of the current rules, and when we specify new rules.

eernstg: Perhaps:

> A _pre-feature library_ has a library scope as the outermost scope. Import prefix names and bindings imported without a prefix are added to this scope, based on special rules about prefix names used by multiple imports, deferred or not, and clashing imported names with or without a prefix.
> 
> These rules are different from the rules about scopes in Dart, but the resulting behavior is similar to a situation where the import prefixes are considered to denote declarations in the library, and imported names are available from a scope which is the enclosing scope of the library scope.

working/parts-with-imports/feature-specification.md
```diff
@@ -183,107 +190,121 @@ been linting against its use for quite a while already. Dotted names in part-of
...
+The import prefixes are added to the same scope as library declarations,
```
eernstg: This contradicts the statement above that the prefixes are located in the import scope.

working/parts-with-imports/feature-specification.md
```diff
@@ -183,107 +190,121 @@ been linting against its use for quite a while already. Dotted names in part-of
...
+all names imported by non-prefixed imports, which is the parent scope of the 
+*top-level scope* with a namespace containing all top-level declarations of
+the library, from both library file and part files, _and_ all prefix names
+of prefixed imports in the library file.
+The import prefixes are added to the same scope as library declarations, 
+and it is a normal scope-name-conflict if a top-level declaration
+has the same base name as an import prefix.
+The same scope is used by all files of the library.
+
+A _post-feature library_ splits the top-level declaration scope from the
```
eernstg: Let's just keep the normal terminology and call it the library scope. (I agree on the rename using 'top-level' below, but right here we're still talking about current Dart.)

working/parts-with-imports/feature-specification.md
```diff
@@ -183,107 +190,121 @@ been linting against its use for quite a while already. Dotted names in part-of
...
+import prefix scope to allow a part file to inherit and override the
+import prefixes of its parent file separately from the shared top-level
+declarations.
```
eernstg: So we have the library scope, enclosed by the import prefix scope, enclosed by the import scope, plus a special rule which says that it's an error for a binding in the import prefix scope to be shadowed by a declaration in the library.

working/parts-with-imports/feature-specification.md
```diff
@@ -183,107 +190,121 @@ been linting against its use for quite a while already. Dotted names in part-of
...
 
 Each Dart file (library file or part file) defines a _combined import scope_
-which combines the combined import scope of its parent file with its own
-imports and import prefixes. The combined import scope of a dart files is
+which extends the combined import scope of its parent file with its own
+imports and import prefixes. _It is this combined import scope that
+sub-parts inherit and extend_. The combined import scope of a Dart file *F* is
 defined as:
```
eernstg: 'Extending' a namespace is a new concept. We have otherwise considered unions of namespaces, taking care to eliminate duplicate bindings of the same name (the union is explicitly an undefined operation when the key sets aren't disjoint). Can't we use the notion of enclosing scopes rather than "extended" or "combined" scopes?

working/parts-with-imports/feature-specification.md
```diff
@@ -183,107 +190,121 @@ been linting against its use for quite a while already. Dotted names in part-of
...
-The *top-level scope* of a Dart file is a library *declaration scope*
+        *C*<sub>*name*</sub>.
+        _A part file can use the same prefix as a prefix that it inherits,
+        because inherited imports are only suggestions.
+        If it adds to that import scope, by importing into it, that can shadow
+        existing declarations, just like in the top-level declaration scope.
+        A deferred prefix import scope cannot be extended, and cannot extend
+        another prefix scope. Deferred prefix scopes are always linked to a
+        single import directive, and a single library, which can be loaded
+        individually of any other import._
```
eernstg: 'independently'?

working/parts-with-imports/feature-specification.md
```diff
@@ -183,107 +190,121 @@ been linting against its use for quite a while already. Dotted names in part-of
...
+    *   Otherwise *P*<sub>*name*</sub> has no parent scope.
+        _If the the parent's combined import scope, *C*, has an non-prefix
+        declaration for a prefix name in *F*, it would be possible to look
+        further up in the import chain of *C*, past the shadowing import,
+        for a prefix scope to extent. We will not do that. A part file's
```
eernstg: Typo: 'extend'

working/parts-with-imports/feature-specification.md
```diff
@@ -183,107 +190,121 @@ been linting against its use for quite a while already. Dotted names in part-of
...
-        `import` s (the declarations of the export scope of each imported
-        library, minus those hidden by a `show` or `hide` operator, combined
-        such that a name conflicts of different declarations is not an error,
-        but the name is marked as conflicted in the scope, and then referencing
-        it is an error.)
-*   Let *P* be a *prefix scope* containing all the import prefixes declared by
-    the current file. The parent scope of *P* is *I*.
+*   Let *P*, the *prefix scope* of *F*, be a scope with *I* as parent scope
+    and a namespace containing every import prefix declared by an import
+    declaration of *F*.
```
eernstg, line 255: We need to specify the following explicitly, to preserve a useful sanity check:
"It is a compile-time error if an import prefix of *F* is the name of a top-level declaration in *F*."

working/parts-with-imports/feature-specification.md
```diff
@@ -183,107 +190,121 @@ been linting against its use for quite a while already. Dotted names in part-of
...
     *   The *P* scope contains an entry for each name where the current file
-        has an `import` directive with that name as prefix, `as name`. (If an
+        has an `import` directive with that name as prefix. If an
         import is `deferred`, it's a **compile-time error** if more than one
         `import` directive in the same file has that prefix name, as usual.
-        _It's not an error if two import deferred prefixes have the same name
+        _It's not an error if two deferred import prefixes have the same name
         if they occur in different files, other file's imports are only
         suggestions._)
```
eernstg:
```suggestion
        if they occur in different files._
```
The notion of 'only suggestions' is more confusing than helpful.

working/parts-with-imports/feature-specification.md
```diff
@@ -183,107 +190,121 @@ been linting against its use for quite a while already. Dotted names in part-of
...
-
-That is: The combined import scope of a Dart file is a chain of the combined
-import scopes of the file and its parent files, each step adding two scopes:
-The (unnamed) import scope of the unprefixed imports and the prefix
-scope with prefixed imports, each shadowing names further up in the chain.
-
-The *top-level scope* of a Dart file is a library *declaration scope*
+        *C*<sub>*name*</sub>.
+        _A part file can use the same prefix as a prefix that it inherits,
+        because inherited imports are only suggestions.
```
eernstg: We would then say that a top-level declaration is a mere "suggestion" in the situation where a nested scope shadows its name? It just doesn't make sense. The top-level declaration has a proper binding in the current scope for that declaration (that is, the library scope), and it's shadowed in some situations, but that's not a property of the shadowed binding, it's a property of the nested scope that this name has a different meaning.

Perhaps: "_A part file can use the same prefix as a prefix that it inherits. Names imported with that prefix will add new bindings to the namespace of that prefix, or shadow existing bindings, and the prefix also provides access to the non-shadowed bindings in that prefix in the parent._"

working/parts-with-imports/feature-specification.md
```diff
@@ -183,107 +190,121 @@ been linting against its use for quite a while already. Dotted names in part-of
...
+        If it adds to that import scope, by importing into it, that can shadow
+        existing declarations, just like in the top-level declaration scope.
```
eernstg: I don't think it's helpful to talk about adding bindings to 'that import scope', it is much more to the point to say that a certain scope has another scope as its enclosing scope, and the adding/shadowing behavior follows. So we're specifying the bindings of *P*<sub>*name*</sub> as importsOf(*S*) where *S* is the set of import directives with that prefix, and the enclosing scope of *P*<sub>*name*</sub> as the scope of the same prefix in the parent.

Presumably, we can skip a parent that has no binding for the prefix name at all and go to a grandparent, or yet further up? (Line 286 says that we won't do that when there is a binding, and it's not a prefix.)

A similar consideration arises when the import scope of *F* (not the parent) has a binding for the given prefix name. Would this shadow a binding of the prefix name to a prefix in the parent?

working/parts-with-imports/feature-specification.md
```diff
@@ -183,107 +190,121 @@ been linting against its use for quite a while already. Dotted names in part-of
...
+        A deferred prefix import scope cannot be extended, and cannot extend
+        another prefix scope. Deferred prefix scopes are always linked to a
+        single import directive, and a single library, which can be loaded
+        individually of any other import._
+    *   Otherwise *P*<sub>*name*</sub> has no parent scope.
+        _If the the parent's combined import scope, *C*, has an non-prefix
```
eernstg: Typo:

```suggestion
        _If the the parent's combined import scope, *C*, has a non-prefix
```

working/parts-with-imports/feature-specification.md
```diff
@@ -183,107 +190,121 @@ been linting against its use for quite a while already. Dotted names in part-of
...
+The *top-level namespace* of a Dart library is a *declaration namespace*
 containing every top-level library member declaration in every library or part
-file of the library. The parent scope of the top-level scope of a Dart file is
-the combined import scope of that Dart file. _Each Dart file has its own copy
-of the library declaration scope, all containing the same declarations, because
-the declaration scopes of different files have different parent scopes._
-
-**It's a '** if any file declares an import prefix with the
+file of the library. 
+The *top-level scope* of *F* has the top-level namespace of the library
```
eernstg: That's nice! I've felt for a long time that "library scope" is a little bit non-obvious, and "top-level scope" would be more natural. With libraries and parts in the mix, this new terminology comes very naturally!

working/parts-with-imports/feature-specification.md
```diff
@@ -183,107 +190,121 @@ been linting against its use for quite a while already. Dotted names in part-of
...
+and it is a normal scope-name-conflict if a top-level declaration
+has the same base name as an import prefix.
+The same scope is used by all files of the library.
+
+A _post-feature library_ splits the top-level declaration scope from the
+import prefix scope to allow a part file to inherit and override the
+import prefixes of its parent file separately from the shared top-level
+declarations.
 
 Each Dart file (library file or part file) defines a _combined import scope_
```
eernstg, line 223: The 'combined import scope' of the file is the same thing as the prefix scope. It's quite confusing that the text uses two different terms to denote the same thing.

The scopes (in particular, the top-level scope of a file *F*, the prefix scope of *F*, and the import scope of *F*) are connected by 'enclosing' edges in the scope graph, just like other scopes in Dart, and we don't otherwise use a different name for any given scope just because we're aware of the fact that it may have some enclosing scopes.

working/parts-with-imports/feature-specification.md
```diff
@@ -344,17 +365,37 @@ There is no attempt to *prioritize* available extensions based on
 where they are imported. Every extension imported or declared in the
 file's top-level scope chain is equally available.
 
+It is not possible to *hide* an extension from a part file, if 
+the extension is available in the parent file.
+That means that a parent file import *can* break a part file
+by importing another applicable extension.
+
+_Conflicts a must resolved as usual: Avoid importing one extension,
```
eernstg: Typo:
```suggestion
_Conflicts must be resolved as usual: Avoid importing an extension
```

working/parts-with-imports/feature-specification.md
```diff
@@ -344,17 +365,37 @@ There is no attempt to *prioritize* available extensions based on
...
+declarations, or be imported with a name conflict,
+and they would still be available for implicit use._
+
 ### Export directives
 
 Any Dart file can contain an `export` directive. It makes no difference which
 file an `export` is in, its exported declarations (filtered by any `hide` or
 `show` combinators) are added to the library's single export scope,
 along with those of any other `export` directives in the library and
 the all non-private declarations of the library itself. Conflicts are handled
```
eernstg: Typo:
```suggestion
all non-private declarations of the library itself. Conflicts are handled
```

working/parts-with-imports/feature-specification.md
```diff
@@ -436,11 +479,10 @@ two files also contains the entire smallest sub-tree. _A tree always has a
 least containing sub-tree for any set of nodes._
 
 *   The least containing file of *two* distinct files is either one of those
-    two files, or the two files are contained in two *distinct* included part
-    files. The least containing file is the only file which contains *both*
-    files, and not in the *same* included file.
-*   Generally, the least containing file of any number of files
-*   is the *only* file which contains all the files, and which does not contain
+    two files, or it's the only file where file which
```
eernstg: Typo:

```suggestion
    two files, or it's the only file which
```

working/parts-with-imports/feature-specification.md
```diff
@@ -657,21 +699,53 @@ All in all, there is very little expected migration since all actively
 developed code, which is expected to use and follow recommended or core lints,
 will already be compatible.
 
+All exsisting valid pre-feature code that is not affected by that
+`part of` change, should also be valid post-feature code.
```
eernstg:
```suggestion
`part of` change should also be valid post-feature code.
```

working/parts-with-imports/feature-specification.md
```diff
@@ -657,21 +699,53 @@ All in all, there is very little expected migration since all actively
...
+
+## Feature Summary
+
+*   **Breaking**: A `part of` file directive cannot use a library name any more.
+*   A part file can contain `import`, `export` and `part` directives.
+*   Part files of a library must form a tree.
+    _A part can, and must, only be part of one parent file._
+*   Exports are independent of where they occur, all works like today.
+*   Imports are local to the importing file and its subparts,
+    and can shadow imports inherited from parent file.
```
eernstg: From 'the parent file', presumably.

working/parts-with-imports/feature-specification.md
```diff
@@ -657,21 +699,53 @@ All in all, there is very little expected migration since all actively
...
+*   Import prefixes shadow normal imports.
```
eernstg: It is true already today that an import prefix will shadow an imported binding when they have the same name. Perhaps the most helpful way to handle this would be:

```suggestion
*   Import prefixes shadow normal imports _(which is also true today)_.
```

lrhn:
Nothing is intended to disagree with the diagram of https://github.com/dart-lang/language/issues/4082.

The library introduces:
- One declaration namespace for all the declarations of the library. (And not import prefixes of import directives.)

Each file introduces:
- One import namespace, for all the declarations of unprefixed imports of the file.
- One prefix namespace, for all the prefixes of the file.
- One prefixed import namespace per prefix, for all the imported declarations with that import prefix.

A file introduces one scope per per namespace that it introduces, with the following parent scopes:
- The import scope: The prefix scope of the parent file, if there is a parent file, otherwise none.
- A prefixed import scope for a prefix *p*: If the parent file's prefix namespace has a scope for the prefix *p*, then that scope, otherwise none.
- The combined import scope: Has the prefix namspace as namespace and the import scope of this file as parent scope.
    - The identifiers of the prefix namespace are bound to the prefixed import scope of this library for that prefix.

A file also has a top-level declaration scope whose namespace is the library's declaration scope, and its parent scope is the file's prefix scope.

The text uses both "prefix scope" and "combined import scope" about the same scope, usually depending on whether the text focuses on the namespace or the entire chain. That's probably confusing, it should use the same name every time.

working/parts-with-imports/feature-specification.md
lrhn: Good catch. I've tried to remove references to "macros", but missed this one.

working/parts-with-imports/feature-specification.md
lrhn: ... but "which" introduces a descriptive clause, not a restrictive one like "that".
(Should this use "that"? I agree the comma is bad.)

munificent: In this case, I believe both "which" and "that" are correct. The former reads a little more formally.

working/parts-with-imports/feature-specification.md
```diff
@@ -183,107 +190,121 @@ been linting against its use for quite a while already. Dotted names in part-of
...
-declaration scope containing all top-level declarations of the library file and
-all part files, and all import prefixes declared by the library file. The
-import prefixes are added to the same scope as library declarations, and there
-is a name conflict if a top-level declaration has the same base name as an
-import prefix.
-
-This feature splits the top-level declaration scope from the import prefix
-scope to allow a part file to override the import prefix, but not the top-level
-declaration.
+Imports, and import prefixes, of a parent file can now be shadowed by
```
munificent: Nit: No commas needed here.

working/parts-with-imports/feature-specification.md
```diff
@@ -597,9 +639,9 @@ the other hand, users may choose to order source depending on properties that
 annotations apply to. _The analyzer may want to review annotations that apply to
 a library for whether they can reasonably apply to any sub-tree of parts. For
 example `@Deprecated(…)` could apply to every member in a sub-tree, allowing a
-library to keep its deprecated API, and its necessary imports, separate from the
-rest, so that it can all be removed as a single operation, and then marking all
-that API as deprecated with one annotation._
+library to keep its deprecated API, and that APIs necessary imports,
```
munificent, line 645: Nit: "APIs" -> "API's".

working/parts-with-imports/feature-specification.md
lrhn: Rephrased.

working/parts-with-imports/feature-specification.md
lrhn: Removed the Obviously. If it's obvious, we don't need to say it's obvious. If it isn't, it's wrong.

working/parts-with-imports/feature-specification.md
lrhn: Someone should do something! Rephrased

working/parts-with-imports/feature-specification.md
lrhn: That ... seems to be correct.
_I really thought we had an import scope as the parent of the library scope. Seems like it would be easier than merging everything into one scope._

Text taken.

working/parts-with-imports/feature-specification.md
lrhn: There is no extending of namespaces, only extending of scopes, which just _means_ having the extended scope as enclosing scope. So I'll try to make the text follow that.
And I've written "parent scope" where the spec uses "enclosing scope", so I'll fix that too.

Hmm. I ended up using "extends" anyway, because it extends the scope chain of the parent's combined import scope with two scopes. It's all defined in the following points, so it should be well-defined.

working/parts-with-imports/feature-specification.md
lrhn: Better!

working/parts-with-imports/feature-specification.md
lrhn: ACK, it's not adding, it's overriding in a nested scope.

working/parts-with-imports/feature-specification.md
lrhn: ACK. It's more of a role-based name. The prefix scope is the scope of the prefix namespace.
It is used both in the role of the enclosing scope of the top-level scope of that file, and in the role of the inherited imports of a parent file. It just happens to be the same scope serving two purposes.

But "prefix scope" is also confusing because that's what we normally call the individual prefix's scopes.
I'll remove it and use "combined import scope" everywhere. Except where they mean the scope of a single prefix.

working/parts-with-imports/feature-specification.md
lrhn: This is still about pre-feature libraries. It's rewritten now.

working/parts-with-imports/feature-specification.md
lrhn: Correct.

We need to name the prefix scope because we refer to it in two places - as the enclosing scope of the top-level scope of a file, and as the "inherited" import scope from a parent file.

We don't need to name the import scope, we only use it as the enclosing scope of the combined import scope/prefix scope, but it's easier to define it as a separate scope, than it is to combine it into the same namespace as import prefixes and simulate the overriding that a scope chain does automatically.

working/parts-with-imports/feature-specification.md
lrhn: That is specified as:
> It's a **compile-time error** if any file declares an import prefix with the
> same base name as a top-level declaration of the library.

It's a few paragraphs below this item list.

working/parts-with-imports/feature-specification.md
lrhn: A quick search suggests that "most style guides" recommend not using an apostrophe with a plural initialism.
The University of Oxford Style Guide agrees.

working/parts-with-imports/feature-specification.md
lrhn: Rephrased.

lrhn:
Addressed comments. Will land on Monday if no complaints.

PR #4524, opened 2025-10-02
Update the declaring constructors feature specification

eernstg:
This PR corrects a couple of mistakes in the feature specification about declaring and primary constructors. It also introduces support for using `new` rather than the class name in declarations of non-declaring constructors (that is, all the kinds that we have today).

### DIFF
--- a/accepted/future-releases/primary-constructors/feature-specification.md
+++ b/accepted/future-releases/primary-constructors/feature-specification.md
@@ -243,9 +243,9 @@ constructor parameter is `final`.
 
 In the case where the declaration is an `extension type`, the modifier
 `final` on the representation variable can be specified or omitted. Note
-that an extension type declaration is specified to use a primary
-constructor (in that case there is no other choice, it is in the grammar
-rules):
+that an extension type declaration is specified to use a declaring
+constructor (it is not supported to declare the representation variable
+using a normal instance variable declaration):
 
 ```dart
 // Using a primary constructor.
@@ -874,8 +874,9 @@ reason for this error is that the modifier `covariant` must be specified on
 the declaration of `v` which is known to exist, not on the parameter.*
 
 A compile-time error occurs if _p_ has both of the modifiers `covariant`
-and `final`. *A final instance variable cannot be covariant, because being
-covariant is a property of the setter.*
+and `final`, also if the latter is implicitly induced. *A final instance
+variable cannot be covariant, because being covariant is a property of the
+setter.*
 
 A compile-time error occurs if _p_ has the modifier `covariant`, but
 neither `var` nor `final`. *This parameter does not induce an instance
@@ -890,30 +891,31 @@ constructors.
 
 The semantics of the declaring constructor is found in the following steps,
 where _D_ is the class, extension type, or enum declaration in the program
-that includes a declaring constructor, and _D2_ is the result of the
+that includes a declaring constructor _k_, and _D2_ is the result of the
 derivation of the semantics of _D_. The derivation step will delete
 elements that amount to the declaring constructor; it will add a new
-constructor _k_; and it will add zero or more instance variable
+constructor _k2_; and it will add zero or more instance variable
 declarations.
 
 Where no processing is mentioned below, _D2_ is identical to _D_. Changes
 occur as follows:
 
-Assume that `p` is an optional formal parameter in _D_ which has the
+Assume that `p` is an optional formal parameter in _k_ which has the
 modifier `var` or the modifier `final` *(that is, `p` is a declaring
 parameter)*.
 
-Assume that the combined member signature for a getter with the same name
-as `p` from the superinterfaces of _D_ exists, and has return type `T`. In
-that case the parameter `p` has declared type `T` as well.
+Assume that `p` has no type annotation, and the combined member signature
+for a getter with the same name as `p` from the superinterfaces of _D_
+exists and has return type `T`. In that case the parameter `p` has declared
+type `T` as well.
 
 *In other words, an instance variable introduced by a declaring parameter
 is subject to override inference, just like an explicitly declared instance
 variable.*
 
-Otherwise, assume that `p` does not have a declared type, but it does have
-a default value whose static type in the empty context is a type (not a
-type schema) `T` which is not `Null`. In that case `p` is considered to
+Otherwise, assume that `p` does not have a type annotation, but it does
+have a default value whose static type in the empty context is a type (not
+a type schema) `T` which is not `Null`. In that case `p` is considered to
 have the declared type `T`. When `T` is `Null`, `p` is considered to have
 the declared type `Object?`. If `p` does not have a declared type nor a
 default value then `p` is considered to have the declared type `Object?`.
@@ -932,25 +934,25 @@ this by specifying the current scope explicitly as the body scope, in spite
 of the fact that the declaring constructor is actually placed outside the
 braces that delimit the class body.*
 
-Next, _k_ has the modifier `const` iff the keyword `const` occurs just
+Next, _k2_ has the modifier `const` iff the keyword `const` occurs just
 before the name of _D_ or before `this`, or if _D_ is an `enum`
 declaration.
 
-Consider the case where _D_ is a declaring header constructor. If the name
+Consider the case where _k_ is a declaring header constructor. If the name
 `C` in _D_ and the type parameter list, if any, is followed by `.id` where
-`id` is an identifier then _k_ has the name `C.id`. If it is followed by
-`.new` then _k_ has the name `C`. If it is not followed by `.`  then _k_
+`id` is an identifier then _k2_ has the name `C.id`. If it is followed by
+`.new` then _k2_ has the name `C`. If it is not followed by `.`  then _k2_
 has the name `C`. If it exists, _D2_ omits the part derived from
 `'.' <identifierOrNew>` that follows the name and type parameter list, if
 any, in _D_. Moreover, _D2_ omits the formal parameter list _L_ that
 follows the name, type parameter list, if any, and `.id`, if any.
 
 Otherwise, _D_ is a declaring body constructor. If the reserved word `this`
-is followed by `.id` where `id` is an identifier then _k_ has the name
-`C.id`. If it is followed by `.new` then _k_ has the name `C`. If it is not
-followed by `.` then _k_ has the name `C`.
+is followed by `.id` where `id` is an identifier then _k2_ has the name
+`C.id`. If it is followed by `.new` then _k2_ has the name `C`. If it is not
+followed by `.` then _k2_ has the name `C`.
 
-The formal parameter list _L2_ of _k_ is identical to _L_, except that each
+The formal parameter list _L2_ of _k2_ is identical to _L_, except that each
 formal parameter is processed as follows.
 
 The formal parameters in _L_ and _L2_ occur in the same order, and
@@ -972,18 +974,19 @@ positional or named parameter remains optional; if it has a default value
   instance variable declaration of the form `T p;` or `final T p;` is added
   to _D2_. The instance variable has the modifier `final` if the parameter
   in _L_ has the modifier `final`, or _D_ is an `extension type`
-  declaration, or _D_ is an `enum` declaration. In all cases, if `p` has
-  the modifier `covariant` then this modifier is removed from the parameter
-  in _L2_, and it is added to the instance variable declaration named `p`.
+  declaration and _k_ is a declaring header constructor. In all cases, if
+  `p` has the modifier `covariant` then this modifier is removed from the
+  parameter in _L2_, and it is added to the instance variable declaration
+  named `p`.
 
 In every case, any DartDoc comments are copied along with the formal
 parameter, and in the case where an instance variable is implicitly
 induced, the DartDoc comment is also added to that instance variable.
 
 If there is an initializer list following the formal parameter list _L_
-then _k_ has an initializer list with the same elements in the same order.
+then _k2_ has an initializer list with the same elements in the same order.
 
-Finally, _k_ is added to _D2_, and _D_ is replaced by _D2_.
+Finally, _k2_ is added to _D2_, and _D_ is replaced by _D2_.
 
 ### Discussion
 
--- a/accepted/future-releases/primary-constructors/feature-specification.md
+++ b/accepted/future-releases/primary-constructors/feature-specification.md
@@ -1,10 +1,10 @@
-# Declaring and Primary Constructors
+# Declaring Constructors
 
 Author: Erik Ernst
 
 Status: Accepted
 
-Version: 1.9
+Version: 1.10
 
 Experiment flag: declaring-constructors
 
@@ -1037,6 +1037,10 @@ far removed from any syntactic hint that the constructor must be constant
 of declaration, and the constructor might be non-const).
 
 ### Changelog
+
+1.10 - October 3, 2025
+
+* Rename the feature to 'declaring constructors'. Fix several small errors.
 
 1.9 - August 8, 2025
 
--- a/accepted/future-releases/primary-constructors/feature-specification.md
+++ b/accepted/future-releases/primary-constructors/feature-specification.md
@@ -620,7 +620,7 @@ constructors as well.
    | <declaringConstantConstructorSignature>;
 
 <constructorName> ::= // Modified rule.
-     <typeIdentifierOrNew> ('.' identifierOrNew)?;     
+     <typeIdentifierOrNew> ('.' identifierOrNew)?;
 
 <typeIdentifierOrNew> ::= // New rule.
      <typeIdentifier>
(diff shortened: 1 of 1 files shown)


accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -874,8 +892,9 @@ reason for this error is that the modifier `covariant` must be specified on
 the declaration of `v` which is known to exist, not on the parameter.*
 
 A compile-time error occurs if _p_ has both of the modifiers `covariant`
-and `final`. *A final instance variable cannot be covariant, because being
-covariant is a property of the setter.*
+and `final`, also if the latter is implicitly induced. *A final instance
```
kallentu: Just trying to understand this better. Could you give me an example of this change? "if the latter is implicitly induced".

kallentu (approved):
Thank you for making these changes.

accepted/future-releases/primary-constructors/feature-specification.md
lrhn: I think that's *only* for the (single) parameter of an intializing constructor of an extension type.

That parameter must just not be `covariant`. Covariant only applies to instance fields of *classes*, and extension types have none.
I'd rather call that out explicitly instead of relying on the parameter also implicitly being `final`.

(If we ever introduce "instance fields of declaring const constructors are implicitly final", then it would apply there too, but that's not relevant today.)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -874,8 +892,9 @@ reason for this error is that the modifier `covariant` must be specified on
...
+variable cannot be covariant, because being covariant is a property of the
+setter.*
 
 A compile-time error occurs if _p_ has the modifier `covariant`, but
 neither `var` nor `final`. *This parameter does not induce an instance
```
lrhn: So error if `covariant` and neither `var` nor `final` (which are *only* allowed on field parameters), and error if both `covariant` and `final`, or implicitly final..

Maybe just: Error if _p_ is marked `covariant` and is not marked `var` _(making it a field-parameter)_.

The only valid use of `covariant` on a constructor parameter is to declare the setter of a mutable field parameter as covariant. Anything that is not a `var` parameter does not declare a setter.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -890,30 +909,31 @@ constructors.
 
 The semantics of the declaring constructor is found in the following steps,
 where _D_ is the class, extension type, or enum declaration in the program
-that includes a declaring constructor, and _D2_ is the result of the
+that includes a declaring constructor _k_, and _D2_ is the result of the
 derivation of the semantics of _D_. The derivation step will delete
 elements that amount to the declaring constructor; it will add a new
-constructor _k_; and it will add zero or more instance variable
+constructor _k2_; and it will add zero or more instance variable
 declarations.
```
lrhn, line 916: So, rewriting. Again, please try to avoid that. and make the syntactic declaration introduce a semantic declaration. A declaring constructor syntax introduces a semantic constructor declaration and zero or more semantic field declarations.

(I'm a denotationalist, and I'm not sorry! Rewrite semantics are not suited for a language that has to tell users where their errors are.)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -890,30 +909,31 @@ constructors.
...
 modifier `var` or the modifier `final` *(that is, `p` is a declaring
 parameter)*.
 
-Assume that the combined member signature for a getter with the same name
-as `p` from the superinterfaces of _D_ exists, and has return type `T`. In
-that case the parameter `p` has declared type `T` as well.
+Assume that `p` has no type annotation, and the combined member signature
+for a getter with the same name as `p` from the superinterfaces of _D_
+exists and has return type `T`. In that case the parameter `p` has declared
+type `T` as well.
```
lrhn: (What if the parameter is `var` and the superinterface has as setter, but not a getter?)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -890,30 +909,31 @@ constructors.
...
 
-Assume that `p` is an optional formal parameter in _D_ which has the
+Assume that `p` is an optional formal parameter in _k_ which has the
 modifier `var` or the modifier `final` *(that is, `p` is a declaring
 parameter)*.
 
-Assume that the combined member signature for a getter with the same name
-as `p` from the superinterfaces of _D_ exists, and has return type `T`. In
-that case the parameter `p` has declared type `T` as well.
+Assume that `p` has no type annotation, and the combined member signature
```
lrhn: I wouldn't assume here. Just: 

> If *p* has no type annotation:
> *  If the combined member signature ..., then the parameter has declared type `T` as well.
> * Otherwise if has a default value ....
> Otherwise 

I wouldn't assume above either, now that I see it. Just "If `p` is an optional formal parameter in _k_ ..." or "Let `p` be any optional formal parameter in _k_ ...".

("Assuming" just gives the wrong vibes. I can't help thinking "but what if you assume wrong?", which derails my reading. Maybe that's just me.)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -890,30 +909,31 @@ constructors.
...
 
 *In other words, an instance variable introduced by a declaring parameter
 is subject to override inference, just like an explicitly declared instance
 variable.*
 
-Otherwise, assume that `p` does not have a declared type, but it does have
-a default value whose static type in the empty context is a type (not a
-type schema) `T` which is not `Null`. In that case `p` is considered to
+Otherwise, assume that `p` does not have a type annotation, but it does
+have a default value whose static type in the empty context is a type (not
```
lrhn: This feels like it's far too soon to talk about static types. Or at least not be explicit about type inference being something that happens later, when it's needed.

We haven't even figured out what the surrounding declaration _is_ yet, and wec can't do type inference on an incomplete program.

So maybe we just have to introdue field declarations without declared types, which get their declared type inferred during type inference, just like we do for `var x = e;` anyway.

That is, rather than say *at this point* what the declared type of the field declaration is, referring to type inference, say that when we get to actual type inference, to find the declared type of the field  *that was introduced by a declaring formal with no declared or inherited type, and with a default value expression*, then do inference on that expression with no context type, and then the declared type of the variable is the inferred type of the variable.

(It's not wrong, it's just reads as happening too soon.)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -932,25 +952,25 @@ this by specifying the current scope explicitly as the body scope, in spite
...
 declaration.
 
-Consider the case where _D_ is a declaring header constructor. If the name
+Consider the case where _k_ is a declaring header constructor. If the name
 `C` in _D_ and the type parameter list, if any, is followed by `.id` where
-`id` is an identifier then _k_ has the name `C.id`. If it is followed by
-`.new` then _k_ has the name `C`. If it is not followed by `.`  then _k_
+`id` is an identifier then _k2_ has the name `C.id`. If it is followed by
+`.new` then _k2_ has the name `C`. If it is not followed by `.`  then _k2_
 has the name `C`. If it exists, _D2_ omits the part derived from
```
lrhn: If what exists? Is it the `. identiferOrNew` from the syntax, which is mentioned in the *next* part of the sentence? (Very confusing! Try to avoid forwards references using "it"!)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -932,25 +952,25 @@ this by specifying the current scope explicitly as the body scope, in spite
...
 `'.' <identifierOrNew>` that follows the name and type parameter list, if
 any, in _D_. Moreover, _D2_ omits the formal parameter list _L_ that
 follows the name, type parameter list, if any, and `.id`, if any.
```
lrhn: The last sentence is confusing. Does it omit the type parameter list? (I know how to read it, it's just not my first reading of that sentence. Maybe stop after `_L_`. Maybe *define* `_L_` first. `Let _L_ be the formal parameter list that follows the name, type parameter list (if any) and `.id` (if any).`. Then you can omit it, when it's clear what it is.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -972,18 +992,19 @@ positional or named parameter remains optional; if it has a default value
   instance variable declaration of the form `T p;` or `final T p;` is added
```
lrhn: (This is where it's a problem that we can't do type inference yet, so we can't know the type `T` that would be inferred from a default value expression yet. In that case, maybe the variable should be declared as `var p;` or `final p;`, and then *type inference*, when used to find the declared type, would look at the default value expression. *Maybe* it could be declared as `var p = e;`, where `e` is the default value expression, because that expression should be constant and not refer to any variable in the constructor's scope anyway, so the variable initializer scope would agree with it. But that's not right, and won't work with `final p = e;`, because then you can't assign to it in an initializer.

I think this is a behavior of a declaring constructor which can just not be captured by desugaring to existing syntax. It's a new kind of instance variable whose type inference depends on an expression somewhere else.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -972,18 +992,19 @@ positional or named parameter remains optional; if it has a default value
...
   to _D2_. The instance variable has the modifier `final` if the parameter
```
lrhn: Also, this is where named positional parameters would need to change something.
(And I wouldn't do *that* by desugaring.)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -972,18 +992,19 @@ positional or named parameter remains optional; if it has a default value
...
-  the modifier `covariant` then this modifier is removed from the parameter
-  in _L2_, and it is added to the instance variable declaration named `p`.
+  declaration and _k_ is a declaring header constructor. In all cases, if
+  `p` has the modifier `covariant` then this modifier is removed from the
+  parameter in _L2_, and it is added to the instance variable declaration
+  named `p`.
 
 In every case, any DartDoc comments are copied along with the formal
 parameter, and in the case where an instance variable is implicitly
 induced, the DartDoc comment is also added to that instance variable.
```
lrhn: No. It can copy any comments, DartDoc or not, along with the parameters, but that has absolutely no effect. Copying it onto the instance variable is meaningless, and suggests that _D2_ will somehow be something that a source processor will be looking at.

DartDoc tools read Dart source, and will read _D_ and find comments there. They will never see DartDoc moved to _D2_. 
No DartDoc tool should ever know that _D2_ even exists. 
No user should ever have to think about _D2_, or know that the spec mentions it. 
(Because it really *does not exist*.  It's entirely a concept introduced to specify the semantics of _D_, by desugaring, but everything a user sees should be in terms of _D_. It's entirely a semantic tool, a local variable inside the specification. And, as I've said before, not my favorite tool.)

One may want to say explicitly that every metadata annotation is moved along with the parameter declaration. But that'd still, IMO, be a mistake. Even `dart:mirrors` should not see _D2_, it should reflect on _D_, and maybe reflect that an instance variable is introduced by a declaring parameter, with a reference to that parameter's mirror, and then you can read any annotations on the parameter there.
(It won't do that, because that would mean actually doing work on `dart:mirrors`, so it'll just not understand what a declaring parameter is, and hopefully not crash whn it sees one. The `dart:mirrors` library doesn't support new features at all, for a very old meaning of "new".)

TL;DR: _D2_ is a hack inside the specification to specify the semantics of _D_. It should never leak to anything outside of the specification, it's just providing the semantics of the actual class declaration _D_ that everyone else sees. So adding comments or metadata to it is meaningless.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: `final` can be omitted from `extension type E(final int i) {}` because it would break all existing extension type declarations otherwise. So that's an implicitly induced `final`. The text in line 895 just says that it still counts when we're checking whether or not we have the error about `covariant`.

eernstg: It is now called out explicitly.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: The newly introduced program elements are now denoted with phrases that include 'semantically' or 'semantics'.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: That's better, thanks!

The language specification uses 'Assume' quite frequently. Typically, it uses 'Assume that `something` has the properties `p1`, .. `pk`.' followed by a separate sentence which is then specified to hold. The point is that this is less convoluted to read than 'If `something` has the properties `p1`, .. `pk` then ...(the true statement)', simply because the latter is twice as long.

However, the bullet list avoids this problem here.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Good point! `covariant` and not `var` is an error.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Good point! Included.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: You can't talk about override inference without relying on type information (we certainly need resolution, but we also need typing because we can't just copy the text that denotes a type).

Moreover, the type of a constant expression in the empty context does not depend on any properties of the non-top-level declarations that surround that expression.

I don't think this is more demanding than many other elements of the Dart static analysis.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Rewritten.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Named positional parameters will be the topic of some other PR.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: I removed the sentence about DartDoc comments. This is no longer a relevant topic for language specification documents.

eernstg (commented):
Great review comments, thanks!

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: This one has not been fully resolved. I think it would make sense to handle it in an issue just for that topic.

kallentu: Can either of you create the issue for this so we can track it? I'm a little muddy the details here, so maybe it'd be best to come from you two.

accepted/future-releases/primary-constructors/feature-specification.md
lrhn: This paragraph can probably work. Nobody needs to know the declared type of the variable until type inference anyway, and an inferred declared type is also a declared type.

It's the creation of the syntax for the declaration below, which includes syntax for the type, which is the real problem.

(And the usual problem that `D2` is a new declaration, which is not the same declaration as `D` and which does not occur in the program being compiled. It's unclear what that means.

> and _D2_ is the result of the derivation of the semantics of _D_.

That sounds like the semantics of _D_ is a rewrite semantics different from the semantics of Dart, so we apply the rewrite semantics, or just "the rewrite"/desugaring, to the declaring constructor declaration first, and then we use the Dart semantics on the result. Which is a bunch of problems waiting to happen, especially if the feature cannot be desugared to existing Dart code.)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -972,18 +992,19 @@ positional or named parameter remains optional; if it has a default value
...
 
 If there is an initializer list following the formal parameter list _L_
-then _k_ has an initializer list with the same elements in the same order.
+then _k2_ has an initializer list with the same elements in the same order.
 
-Finally, _k_ is added to _D2_, and _D_ is replaced by _D2_.
+Finally, _k2_ is added to _D2_, and _D_ is replaced by _D2_.
```
lrhn, line 1007: Replaced in what? ("Everywhere", probably. Any and every reference to _D_ is now a reference to _D2_ instead, there is no _D_.)

When does this replacement happen, precisely?

- After parsing, for sure.
- And after all transitivie file dependencies are parsed, probably.
- After top-level name resolution and export scope resolution, probably.
- After checking type hierarchy validity? (No cycles, no extending a non-class.) Probably. This change does not affect - type hierarchy relations, can be before or after.
- After checking class `interface`/`base`/`final` modifier rules. Again, the rewrite doesn't change anything, so can be after.
 
- After rewriting the superclass, if necessary? Probably, since it refers to superclass getters declarations, and those should not

- Before checking if a `mixin class` has a non-trivial constructor. Since it doesn't specify when a primary constructor is trivial, it must be checked after desugaring.
- Before identifier resolution in members. The name of the new instance variables need to be in scope for that.
- Before checking if class members are valid overrides of superinterface members. 
- Before checking if classes have duplicate member names. The rewrite adds declartions which should not have name clashes.
- Before type inference.

That means that some parts of the language semantics see the original syntax, and some parts of the langauge semantics see the rewritten syntax.
And every part of the semantics must know and understand which one it sees, and what that means. Are we sure that's done correctly everywhere?

PR #4533, opened 2025-10-10
Add motivation to "Augmentations" about code generation and FFI.

munificent:
We usually think of augmentations as being generated code augmenting hand-authored code. But there are use cases in the other direction as well where the code generator produces the base declaration and a user hand-writes an augmentation to tweak it.

Add some verbiage to the motivation section to capture that. There's no normative change in this PR, it's just context.

cc @brianquinlan @devoncarew

### DIFF
--- a/working/augmentations/feature-specification.md
+++ b/working/augmentations/feature-specification.md
@@ -59,8 +59,9 @@ in both directions:
 Having a mixture of hand-authored and generated code works well when the
 generated code consists of completely separate declarations from the
 hand-authored code. But if a code generator wants to, say, add a method to a
-hand-authored class, then the language is of little help. This proposal
-addresses that limitation by adding *augmentations*.
+hand-authored class, or an author wants to add a method to a generated class,
+then the language is of little help. This proposal addresses that limitation by
+adding *augmentations*.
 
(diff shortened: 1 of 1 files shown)


working/augmentations/feature-specification.md
```diff
@@ -43,10 +43,24 @@ machine-generated code in separate files so that the code generator doesn't
...
+*   Other times, a code generator produces code with default basic behavior and
+    that a human author then wants to tweak or refine it. You see this sometimes
+    with FFI where you a code generator provides a default API to some external
+    system but where you want to layer on hand-authored code to provide a more
+    natural Dart-like experience.
+
+Having a mixture of hand-authored and generated code works well when the
+generated code consists of completely separate declarations from the
+hand-authored code. But if a code generator wants to, say, add a method to a
+hand-authored class, then the language is of little help. This proposal
```
lrhn: Maybe: ", or an author wants to add a method to a generated class" 

It's interesting that augmenting generating code can depend on user written signatures, but it's probably rare that a generated class will have an empty signature that is expected to be filled in using a hand-written implementation.

Not impossible, but rarer.

Issue #3102, opened 2023-05-23
Pattern-esque syntax for terser named arguments


nex3:
The pattern-matching syntax introduces the constructs `:var foo` and `:foo` to declare a variable whose name matches a field on the object being matched. The purpose of this is to reduce duplication, to users don't have to explicitly write `foo: var foo`.

I propose using this syntax as a way to reduce duplication when _passing_ named arguments. Specifically, make `foo(:argument)` equivalent to `foo(argument: argument)`. Similarly, record literals could be declared using `(:field)` instead of `(field: field)`. This would help reduce duplication and ease acclimation to Dart for users used to JS, which supports similar behavior when using maps as named arguments (in JS, `foo({argument})` is equivalent to `foo({argument: argument})`).

lrhn:
It's definitely going to be briefer, and is likely to get used a lot.

We'd have to define which expressions counts as having a "name" that can be used as parameter/field name.
* A plain identifier is a given.
* A prefixed identifier too, like `: prefix.varName` or `: Class.staticVarName` (and `: prefix.Class.staticVarName`).
* Then we might as well allow `: expression.getterName` in general, so any expression ending in `.identifier`.
* And parenthisation should be safe, even if never necessary: `: (expression.getterName)`

Those are fairly safe. Then there are the more speculative ones:
* `:foo.v1 ?? bar.v1` - both have the same name, use that. Similar for `||`, `&&` and `?`/`:`.
* `:foo.v1 ?? (throw "Nope")` - only one branch has a name, use that.
* `:await v1` - the await doesn't have to affect the name, the name's meaning is likely to still apply.
* Other prefix operators? `: ! foo.hasBar` - probably not useful since `!` changes, flips, the meaning, so why should the name still match?
* Increment operators: `: ++varName`, `: varName++` can both work, the name and value still match conceptually.
* Assignment `: foo = 42`. Probably too weird, even if there is a name right there. 
* But maybe `:foo = bar` using `bar` as the name (or whatever the name of the RHS expression is), and ignore the LHS.
* No binary operators otherwise. Generally, we want expressions in tail-like position, so the "name of the expression" still corresponds to the value.
* More?

nex3:
The analogous feature in JS _only_ works for plain identifiers, but it would be pretty neat to be able to use at least expressions ending in `.identifier`.

munificent:
[...]

PR #4538, opened 2025-10-16
Add proposed feature specification for "implicit names".

lrhn:
For issue #3102

### DIFF
--- a/working/3102 - implied-name/feature-specification.md
+++ b/working/3102 - implied-name/feature-specification.md
@@ -15,7 +15,7 @@ named arguments:
   );
 ```
 
-To avoid redundant repetition, Dart will allow you to omit the
+To avoid redundant repetition, this feature will allow you to omit the
 argument name if it's the same name as the expression providing the value.
 
 ```dart
@@ -37,7 +37,7 @@ Color colorWithAlpha(Color color, int newAlpha) {
   return (red: red, green: green, blue: blue, alpha: newAlpha);
 }
 ```
-will become
+which this feature will allow to be written as:
 ```dart
 Color colorWithAlpha(Color color, int alpha) {
   var (:red, :green, :blue, alpha: _) = color;
@@ -115,7 +115,7 @@ Where the language specification refers to a named argument's name,
 it now uses this definition of the name of a `<namedArgument>`.
 
 Where the Record specification for a record literal refers to a field name,
-it now uses to this definition.
+it now uses this definition.
 
 ## Semantics
 
@@ -239,9 +239,6 @@ The other cases do not contain any other identifier than the one that
 provides the value of the expression. A cascade could have any expression
 after the `..`, so it would be even more important for readability that
 the reader knows to look at the very next identifier for the name.
-
-However, the places where cascades are used are often in argument position,
-which
 
 Examples of uses that could be made shorter (from `package:csslib`):
 ```dart


working/3102 - implied-name/feature-specification.md
```diff
@@ -0,0 +1,273 @@
...
+```
+will become
+```dart
+extension on Color {
+  Color withAlpha(int alpha) {
+    var (:red, :green, :blue, alpha: _) = this;
+    return (:red, :green, :blue, :alpha);
+  }
+}
+```
```
eernstg, line 46: It may be worth noting that the following would suffice. Before:

```dart
extension on Color {
  Color withAlpha(int alpha) =>
      (red: red, green: green, blue: blue, alpha: alpha);
}
```

`alpha` is resolved in the lexical scope, and the other identifier expressions are resolved as instance member references (adding `this.`). After:

```dart
extension on Color {
  Color withAlpha(int alpha) => (:red, :green, :blue, :alpha);
  }
}
```

If you want to avoid the discussion about "but is that readable?" then perhaps use plain local declarations rather than an extension method.

working/3102 - implied-name/feature-specification.md
```diff
@@ -0,0 +1,273 @@
...
+immediately after a `(` or a `,`, as the entire content until
+a following `,` or `)`.
+Starting with a `:` at that point is not currently possible.
+(Even if we allow metadata annotations inside argument lists or record literals,
+it's unambiguous whether those annotations includes a following identifier or
+not.)
+
+As a non-grammatical restriction, it's a **compile-time error**
+if a `<namedExpression>` omits the leading`<identifier>`, and the following
+expressions is not a _single identifier expression_, as defined below.
```
eernstg: Typo:

```suggestion
expression is not a _single identifier expression_, as defined below.
```

working/3102 - implied-name/feature-specification.md
```diff
@@ -0,0 +1,273 @@
...
+
+In short, an expression is a _single identifier expression_
```
eernstg:
```suggestion
An expression is a _single identifier expression with identifier_
```

Having seen line 93, it seems likely that this paragraph should be commentary (such that readers will know which one wins if the two definitions disagree). Moreover, if this is changed to commentary then it would probably be more readable if placed after the actual definition (which would then motivate a reintroduction of 'In short,')

working/3102 - implied-name/feature-specification.md
```diff
@@ -0,0 +1,273 @@
...
+which has a specific single _identifier_ if and only if
```
eernstg:
```suggestion
`id` if and only if
```

working/3102 - implied-name/feature-specification.md
```diff
@@ -0,0 +1,273 @@
...
+it is one of the following:
+*   An identifier, with that identifier.
+*   `s!`, `s as T`, `(s)`, `s..cascadeSection`/`s?..cascadeSection`
```
eernstg:
```suggestion
*   `s!`, `s as T`, `(s)`, `s..cascadeSection`, and `s?..cascadeSection`
```
I think this works fine, but it might be helpful to mention (or show in an example) that this admits constructs like `id?..m1()..m2()` because `id?..m1()` can play the role as `s`.

working/3102 - implied-name/feature-specification.md
```diff
@@ -0,0 +1,273 @@
...
+    where `s` is a single identifier expression,
+    and then it has the same identifier as `s`.
+
+More formally, An expression `e` is a single identifier expression with a certain identifier if and only if it is defined as such by the following:
```
eernstg:
```suggestion
An expression `e` is a single identifier expression with a certain identifier if and only if it is defined as such by the following:
```

working/3102 - implied-name/feature-specification.md
```diff
@@ -0,0 +1,273 @@
...
+        and the `<cascade>` is a single identifier expression,
+        then `e` is a single identifier expression with the same identifier as
+        the `<cascade>`.
+    *   If `e` a `<cascade>` of the form
+        ``<conditionalExpression> (`?..' | `..') <cascadeSection>``
+        and the `<conditionalExpression>` is a single identifier expression,
+        then `e` is a single identifier expression with the same identifier as
+        the `<conditionalExpression>`.
+
+The _name of a named expression_ is then:
```
eernstg: In this particular case it's probably more readable to keep using the actual non-terminal:


```suggestion
The _name of a_ `<namedExpression>` is then:
```

working/3102 - implied-name/feature-specification.md
```diff
@@ -0,0 +1,273 @@
...
+* If the named expression has a leading identifier, then that identifier.
+* Otherwise the following expression must be a single identifier expression
+  with an identifier *I*, and then the name of the named expression is *I*.
+
+The name of a `<namedArgument>` or a named `<recordField>` is the name of
+its `<namedExpression>`.
+
+
```
eernstg: Typo:

```suggestion
```

working/3102 - implied-name/feature-specification.md
```diff
@@ -0,0 +1,273 @@
...
+Where the language specification refers to a named argument's name,
+it now uses this definition of the name of a `<namedArgument>`.
+
+Where the Record specification for a record literal refers to a field name,
+it now uses to this definition.
```
eernstg: Typo:

```suggestion
it now uses this definition.
```

working/3102 - implied-name/feature-specification.md
```diff
@@ -0,0 +1,273 @@
...
+can be tweaked.
+
+It's currently restricted to expressions where the value of the expression
+is always the value of evaluating a single identifier.
+That identifier is always the *next* identifier, and the next non-`(` token.
+
+The identifier expression can be wrapped in casts `!` or `as T`, in parentheses,
+or can be pre-used/modified using cascade invocations, but none of those
+operations change the value from evaluating the identifier, only, potentially,
+whether it evaluates to a value at all.
```
eernstg: Perhaps mention prefixed names, and motivate why they aren't included: `(name: p.name,)` can't be abbreviated to `(:p.name,)`. Presumably it's motivated by the need to insist on "first" in "the name is the first identifier after the colon".

lrhn: Not sure why a prefixed name is more relevant to mention than, say, a static member or an instance member.

(And "first identifier" is the reason. I'd personally be happy to allow any  `prefix.id` or `typeName.id`, `expression.id` or `expression?.id`, so that's probably why I'm not trying to find an argument for why we don't allow it.)

working/3102 - implied-name/feature-specification.md
```diff
@@ -0,0 +1,273 @@
...
+```dart
+ TextSpan(text: offScreenText, recognizer: recognizer..onTap = () {})
+```
+which would become:
+```dart
+ TextSpan(text: offScreenText, :recognizer..onTap = () {})
+```
+
+It is the use that has the biggest risk of being confusing to read,
+but you can always write the identifier if you prefer it.
```
munificent: I'm definitely not sold on allowing cascades. I think the before examples here are clearer even though more verbose.

In the previous comment, you say you'd be happy to support `prefix.id`. I'm sure if we do allow cascades, then users will request `.` too. But if we were to do that and also cascades, that implies these two are equivalent:

```dart
(a: x.a, b: b..y);
(:x.a, :b..y);
```

Taking the left-most identifier of a cascade chain and the right-most identifier of a property chain makes sense semantically, but is super confusing syntactically.

I'd rather avoid this whole pitfall and not support `..` or `.`.

working/3102 - implied-name/feature-specification.md
```diff
@@ -0,0 +1,273 @@
...
+    and then it has the same identifier as `s`.
+
+More formally, An expression `e` is a single identifier expression with a certain identifier if and only if it is defined as such by the following:
+
+*   If `e` is a `<primary>` expression which is an `<identifier>`,
+    it is a single identifier expression with that `<identifier>` as identifier.
+*   `s!`: If `e` is  a `<primary> <selector>*` production where `<selector>*`
+    is the single `<selector>` `` `!' ``, and `<primary>` is
+    a single identifier expression, then `e` is a single identifier expression
+    with the same identifier as the `<primary>`,
```
munificent: In the interest of completeness, I'm compelled to ask about `(:foo++)` and `(:foo--)` as well.

(I could go either way with them. But I'd lean towards "no".)

munificent (approved):
(Except for cascades) I love it! Ship it!

working/3102 - implied-name/feature-specification.md
lrhn: Changed to just be a plain function taking `color` as argument.

Then it's _probably_ better to write:
```dart
  var (:red, :green, :blue, alpha: _) = color;
  return (:red, :green, :blue, :alpha);
```
than
```dart
  return (red: color.red, green: color.green, blue: color.blue, :alpha);
```
(But not by much).

working/3102 - implied-name/feature-specification.md
lrhn: I've rewritten and shorthened the in-short form (and the more-formally form) and put the formal specification first.

working/3102 - implied-name/feature-specification.md
lrhn: Didn't find any examples of a `id: id?..selector..selector` in the wild.
I added examples of using `?..`, but not two selectors.

working/3102 - implied-name/feature-specification.md
lrhn: Actually intended. But adding extra verical space is wasted on markdown, it won't be rendered anyway.

working/3102 - implied-name/feature-specification.md
lrhn: That is a good argument.

My argument for including cascades is that there is existing code that does, fx, `queryParameters: queryParameters..removeWhere((_, value) => value == null)`.

I guess you can still write that, so it's not a loss of expressiveness, just not a gain.
At a risk of a gain in confusion, and possible more confusion if we ever allow `:a.b`.

working/3102 - implied-name/feature-specification.md
lrhn: Arguably ... they work and stay within the bounds set out.

The value of `foo++` was the value of `foo` (which has since been incremented).
That means that the meaning of `foo`-the-parameter is likely the same as `foo`-the-variable, since the same value is used for both. That's my mental check for whether this makes sense: We write only one name to cover two things, so the name should mean "the same thing" in both uses.

The argument against `:foo++` is that we won't allow `:++foo`, and that asymmetry can be grating.

Or we could allow `:++foo` too. I'd be perfectly happy with both - they are "one identifier expressions" and the value of the expression is the value of the variable - either what it was before increment, or what it is now, after increment. As long as people don't do patological things with setters. The name is the next identifier in the expression, and it'll be prefixed at most by `(`s and then at most one `++` or `--`, the name is easy to find.

And also that if `:x++` is allowed, but `:x+=1` is not, even if it's definitionally the same, it's a little special-casey. You can use it if you need to add 1, but not if you need to add 2?

But it's a single identifier with  the same name and its value is used as the value, so it's not completely out of bounds.

(My biggest argument against cascades is that they contain more identifiers. This one doesn't.
Even if the first one is easy to find, it is still messy. (And more if we'd want `:a.b`.) Increment expressions don't suffer from that.)

lrhn:
Updated, PTAL.

working/3102 - implied-name/feature-specification.md
munificent: > Or we could allow `:++foo` too.

If we did that, then we'd also probably want `:-foo`, `:!foo`, and `:~foo`.

My inclination would to not do any of those. I'd even be OK with not allowing `foo!`, but I think that one's worth doing because (a) patterns allow it and (b) it makes nullable types easier to work with, which is always good.

working/3102 - implied-name/feature-specification.md
```diff
@@ -0,0 +1,377 @@
...
+```dart
+  var subscription = stream.listen(
+    onData,
+    onError: onError,
+    onDone: onDone,
+    cancelOnError: cancelOnError,
+  );
+```
+
+To avoid redundant repetition, Dart will allow you to omit the
```
munificent: "will" is a little vague here. Is it describing an already upcoming language change or this proposed feature? I'd say "this proposal allows you...".

working/3102 - implied-name/feature-specification.md
```diff
@@ -0,0 +1,377 @@
...
+>        and the `<conditionalExpression>` is `s`,
+>        then `e` is a single identifier expression with identifier *I*.
+
+The other cases do not contain any other identifier than the one that
+provides the value of the expression. A cascade could have any expression
+after the `..`, so it would be even more important for readability that
+the reader knows to look at the very next identifier for the name.
+
+However, the places where cascades are used are often in argument position,
+which
```
munificent: "which"...?

working/3102 - implied-name/feature-specification.md
lrhn: I moved cascades out of th e feature, into the discussion section.

working/3102 - implied-name/feature-specification.md
lrhn: I'd argue against `-foo`, `!foo` and `~foo` in that they *don't* mean the same thing as the name. They likely mean the _exact opposite_.

Doing `foo(isOpen: !isOpen)` or `foo(count: -count)` would look wrong, and allowing `foo(:!isOpen)` and `foo(:-count)` would hide that.

The only operators allowed are ones that do not change the value of the identifier expression, they just might throw before evaluating to that value (`!` or `as Type`).

For `id++` and `++id`, the value of the expression *is* the value of the variable, either before or after the variable is incremented, that's why they could be argued to fit. But as I said, a mutable variable is mutable state, being the state can mean something else than just its value. This feature is _mainly_ intended for
_forwarding values_ from one place to another with the same name.

working/3102 - implied-name/feature-specification.md
lrhn: Yeah, I probably didn't know either :)

(I did, I moved it into the first paragraph, and forgot to remove this line.)

PR #4540, opened 2025-10-16
Revise "Augmentations".

### LINKED ISSUE #4377: Should "augment" be optional?
The augmentations proposal currently says that the introductory declaration *can't* be marked `augment` and all augmenting declarations *must* be.

It's unambiguous that an augmenting declaration is augmenting: it has the same name as a previous declaration in the same augmentation context. That means we could make the keyword optional or eliminate it entirely. Should we?

Personally, I do like having it be explicit. Say you were to write:

```dart
class SomeVeryLongName { ... }

// Much later...

class SomeVeryLongName { ... }
```

You meant the latter to be an unrelated class but accidentally used the same name. If we allow implicit augmentation, instead of a simple name collision error, you'll get a cascade of errors for every part of the class that doesn't overlap properly. If, for example, they both happen to define methods with the same name but different signatures, you'll get errors in those methods. Perhaps worse, if they *don't* overlap, you'll just silently create one class that is the Frankenstein amalgamation of both.

I don't think `augment` is much ceremony and avoids those pitfalls. But I could be persuaded otherwise.

munificent:
There are no behavioral changes in this PR. It only does some copy editing, removes redundant specification, and adds clarification and motivation.

Fix #4377.

### DIFF
--- a/working/augmentations/feature-specification.md
+++ b/working/augmentations/feature-specification.md
@@ -71,16 +71,16 @@ an *introductory declaration*.
 
 An augmentation declaration doesn't create a new entity in the program. Instead,
 it adds to some existing introductory declaration with the same name.
-Augmentation declarations can modify types, functions, members, and almost any
-other kind of declaration in Dart. They can add new members to, add new values
-to enums, or to the `with` or `implements` clauses, fill in function bodies,
-etc.
+Augmentation declarations can affect types, functions, members, and almost any
+other kind of declaration in Dart. They can add members to classes, append
+values to enums, append types to `with` or `implements` clauses, fill in missing
+function bodies, etc.
 
 An augmentation can be in the same file as the introductory declaration it
 applies to, or in separate part files, but they must both be in the same
 library. Augmentations allow authoring a library in multiple separate pieces or
-files (some of which are likely generated) in a flexible manner, but they aren't
-for injecting new behavior into libraries outside of the author's control.
+files (some of which are likely generated) in a flexible manner, but they can't
+inject new behavior into libraries outside of the library author's control.
 
 ### Design principle
 
@@ -265,6 +265,9 @@ Note that only extensions *with names* allow a leading `augment`. Since
 augmentations are matched with their introductory declaration by name, unnamed
 extensions can't be augmented. *Doing so wouldn't accomplish anything anyway.
 Just make two separate unnamed extensions.*
+
+Also note that an augmentation of an extension can't specify an `on` clause. It
+always uses the same `on` clause as the introductory declaration.
 
 ## Static semantics
 
@@ -452,6 +455,17 @@ ones must be.*
 An augmentation declaration does not introduce a new name into the surrounding
 scope. *We could say that it attaches itself to the existing name.*
 
+It's a **compile-time error** if an augmentation doesn't have the same kind as
+the introductory declaration. For example, augmenting a `class` with a `mixin`,
+an `enum` with a function, a method with a getter, etc.
+
+The exception is that a variable declaration (introductory or augmenting) is
+treated as a getter declaration (and a setter declaration if non-final) for
+purposes of augmentation. These implicit declarations can augment and be
+augmented by other explicit getter and setter declarations. *In other words,
+variables are never augmented or augmenting, only the getters and possibly
+setters that they induce are.*
+
 ### Complete and incomplete declarations
 
 Augmentations aren't allowed to *replace* code, so they mostly add entirely new
@@ -461,13 +475,14 @@ augmentations can fill in a body for an augmented declaration that lacks one.
 More precisely, a function or constructor declaration (introductory or
 augmenting) is *incomplete* if all of:
 
-*   The body syntax is `;`.
+*   It has no body. *That means no `{ ... }` or `=> ...;` but only `;`.*
 
 *   The function is not marked `external`. *An `external` function is considered
     to have a body, just not one that is visible as Dart code.*
 
-*   There is no redirection, initializer list, initializing formals, or super
-    parameters. *Obviously, this only applies to constructor declarations.*
+*   There is no redirection, initializer list, initializing formals, field
+    parameters, or super parameters. *Obviously, this only applies to
+    constructor declarations.*
 
 If a declaration is not *incomplete* then it is *complete*.
 
@@ -605,8 +620,8 @@ augment enum E { v5 }
 *The resulting enum has values `v1`, `v2`, `v3`, `v4`, and `v5`, in that order.*
 
 *A consequence of application order is that an augmentation in one part file can
-augment an introductory declaration in a sibling part file. This is intuitively
-confusing. The augmentation is referring to a declaration that can't be found
+augment an introductory declaration in a sibling part file. This may be
+confusing since the augmentation affects a declaration that can't be found
 anywhere in its `part of` chain.*
 
 *However, restricting augmentations to only apply to declarations _above_ (not
@@ -629,13 +644,18 @@ augment class SomeClass {
 }
 ```
 
-Mixin application classes can't be augmented.
+Mixin application classes can't be augmented:
 
-A class, enum, extension type, mixin, or mixin class augmentation may
-specify `extends`, `implements` and `with` clauses (when generally
-supported). The types in these clauses are appended to the introductory
-declarations' clauses of the same kind, and if that clause did not exist
-previously, then it is added with the new types.
+```dart
+class C = S with M;
+augment class C = S with N; // Error.
+```
+
+A class, enum, extension type, mixin, or mixin class augmentation may specify
+`extends`, `implements` and `with` clauses (when otherwise supported). The types
+in these clauses are appended to the introductory declarations' clauses of the
+same kind, and if that clause did not exist previously, then it is added with
+the new types.
 
 *Example:*
 
@@ -652,16 +672,12 @@ including enum values, are added to the instance or static namespace of the
 corresponding type in the introductory declaration. *In other words, the
 augmentation can add new members to an existing type.*
 
-Instance and static members inside a class-like declaration may themselves
-be augmentations. In that case, they augment the corresponding members in
-the same augmentation context, according to the rules in the following
-subsections.
+An instance or static member inside a class-like declaration may itself be an
+augmentation. In that case, it augments the corresponding member (the
+introducing member with the same name) in the same augmentation context,
+according to the rules in the following subsections.
 
 It's a **compile-time** error if:
-
-*   An augmentation declaration is applied to a declaration of a different kind.
-    For example, augmenting a `class` with a `mixin`, an `enum` with a function,
-    etc.
 
 *   An augmenting class declaration has an `extends` clause and any prior
     declaration for the same class also has an `extends` clause.
@@ -679,8 +695,7 @@ It's a **compile-time** error if:
     type parameters with the exact same type parameter names (same identifiers)
     and bounds if any (same *types*, even if they may not be written exactly the
     same in case one of the declarations needs to refer to a type using an
-    import prefix). The augmenting declaration may choose to omit bounds in
-    which case they will be inherited from the augmented declaration.
+    import prefix).
 
     *Since repeating the type parameters is, by definition, redundant, this
     restriction doesn't accomplish anything semantically. It ensures that
@@ -688,6 +703,18 @@ It's a **compile-time** error if:
     parameters that it uses in its body and avoids potential confusion with
     other top-level variables that might be in scope in the library
     augmentation.*
+
+    The augmenting declaration may choose to omit the bound on any type
+    parameter, in which case it will be inherited from the augmented
+    declaration.
+
+    ```dart
+    foo<X extends num, Y extends X>();
+    augment foo<X extends num, Y>() { ... }
+    ```
+
+    *Here, `Y` in the augmentation inherits `extends X` from the introductory
+    declaration.*
 
 ### Augmenting function and constructor signatures
 
@@ -703,8 +730,8 @@ signature *matches* if:
     (same *types*, even if they may not be written exactly the same in case one
     of the declarations needs to refer to a type using an import prefix).
 
-*   The return type (after type annotation inheritance) is the same as the
-    augmented declaration's return type.
+*   The return type (if not omitted) is the same as the augmented declaration's
+    return type.
 
 *   It has the same number of positional and optional parameters as the
     augmented declaration.
@@ -713,7 +740,7 @@ signature *matches* if:
 
 *   For all corresponding pairs of parameters:
 
-    *   They have the same type (after type annotation inheritance).
+    *   They have the same type (if not omitted in the augmenting declaration).
 
     *   They have the same `required` and `covariant` modifiers.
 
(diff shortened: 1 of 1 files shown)


working/augmentations/feature-specification.md
```diff
@@ -65,35 +65,22 @@ adding *augmentations*.
...
+declaration* (or just "augmentation"). In contrast, an unmarked declaration is
+an *introductory declaration*.
 
-*In Dart without this feature there are no augmentation declarations. Now
-that we are adding augmentation declarations we need to have a term that
-denotes a declaration which is not an augmentation. That is, it is one of
-the "normal" declarations that we've had all the time.*
+An augmentation declaration doesn't create a new entity in the program. Instead,
+it adds to some existing introductory declaration with the same name.
+Augmentation declarations can modify types, functions, members, and almost any
```
lrhn: modify -> augment

("modify types" sounds like it can *change* the types of a declaration, which it can't, only augment a type declaration.)

working/augmentations/feature-specification.md
```diff
@@ -65,35 +65,22 @@ adding *augmentations*.
...
+other kind of declaration in Dart. They can add new members to, add new values
+to enums, or to the `with` or `implements` clauses, fill in function bodies,
```
lrhn: This sentence feels like it's been edited, so that the parts no longer fit together.
Consider:

> They can add new members to classes, new values to enums, new types to `with`
> or `implements` clauses, can fill in missing function bodies. They can  complete or
> amend existing declarations, allowing a declaraton to be split into multiple files.

working/augmentations/feature-specification.md
```diff
@@ -65,35 +65,22 @@ adding *augmentations*.
...
-An augmentation can fill in a body for a declared member that has no body, but
-can't *replace* an existing body or add to it. In order to allow augmentations
-to provide bodies for static methods and top-level functions, we allow
-declarations of those to be "abstract" and lack a body as long as a body is
-eventually provided by an augmenting declaration.
+An augmentation can be in the same file as the introductory declaration it
+applies to, or in separate part files, but they must both be in the same
+library. Augmentations allow authoring a library in multiple separate pieces or
+files (some of which are likely generated) in a flexible manner, but they aren't
+for injecting new behavior into libraries outside of the author's control.
```
lrhn: ... aren't intended for injecting ...
or 
... arent designed for injecting ...
or (more directly)
... cannot inject ...

lrhn: ... outside of the library author's control ...

(The person trying to inject into a library is also an author.)

working/augmentations/feature-specification.md
lrhn: (Nit: "It has no body". The `;` is not a body, it's a marker representing the absence of a body.)

lrhn: (Add "field parameter" too, this will launch after declaring constructors.)

working/augmentations/feature-specification.md
```diff
@@ -803,17 +794,27 @@ augment class C {
...
 
+*   An augmentation declaration is applied to a declaration of a different kind.
+    For example, augmenting a function with a `class`, a method with a getter, a
+    constructor with a static method, etc.
+
+    The exception is that a variable declaration (introductory or augmenting) is
+    treated as a getter declaration (and a setter declaration if non-final) for
+    purposes of augmentation. These implicit declarations can augment and be
+    augmented by other explicit getter and setter declarations. See the next
+    section.
```
eernstg: I guess we could think about this in the following way: A non-local variable declaration implicitly induces a getter and possibly a setter. The getter and potential setter add a binding in the current scope and give rise to member signatures in the interface of the enclosing class/mixin/enum/mixin-class (for instance variables). So we have an implementation and a scope binding / signature for the getter and potential setter. The variable goes nowhere, it's just the out of sight agent that pulls all the strings.

Augmentations apply to the getter and potential setter, and it makes no difference whether the accessors were explicitly declared or implicitly induced by a variable declaration. We have the following:

> It's a compile-time error if an augmentation is complete and any declaration before it in the augmentation chain is also complete. *In other words, once a declaration has acquired a body, no augmentation can replace it with another.*

This would imply that the following is an error, because the getter has a body (which will read the storage where the variable is allocated) and similarly for the setter:

```dart
int x;
augment var x; // Error: Implementation provided multiple times.
```

Certainly, we would make it an error to have the following, in spite of the fact that the implementation doesn't change, we just can't override it:

```dart
abstract final int x;
augment get x => 1;
augment get x => 1;
```

By the way, the document contains this, which would probably be a type-like missing update:

```dart
f1(int _) {}
augment f1(int x) {} // OK.
augment f1(int _) {} // OK.
augment f1(int y) {} // Error, can't change name.
augment f1(int _) {} // OK.
```

It should probably use `;` rather than `{}` in all four augmenting declarations.

eernstg (approved):
LGTM

Just one comment on the conceptualization of augmenting declarations and variables. We may want to state explicitly that one particular kind of body can actually be "overridden by the same body", such that `int x; augment int x;` isn't an error. Or, alternatively, we should probably mention this situation and the error explicitly.

working/augmentations/feature-specification.md
```diff
@@ -616,6 +604,20 @@ augment enum E { v5 }
 
 *The resulting enum has values `v1`, `v2`, `v3`, `v4`, and `v5`, in that order.*
 
+*A consequence of application order is that an augmentation in one part file can
+augment an introductory declaration in a sibling part file. This is intuitively
+confusing. The augmentation is referring to a declaration that can't be found
```
lrhn: > This is intuitively confusing

No, it's not!
(Is so! Is not!)

That depends very much on the intuition you depend on for understanding how this works.

If we don't tell people that the "below" order matters, they shouldn't _start_ with that intuition. If they know that the "After" order is important, and start thinking of part directives as defining an order, that's the order they'll use to think about it. Then it's not confusing that a *later* sibling can augment  an earlier declaration.

Maybe say "Can be confusing". Definitely drop "intuitively". No two persons can agree on what is "intuitive".

working/augmentations/feature-specification.md
```diff
@@ -616,6 +604,20 @@ augment enum E { v5 }
...
+anywhere in its `part of` chain.*
```
lrhn, line 625: We don't care that you can refer to the name of the introducing top-level declaration from any file in the library, it's in the top-level scope.

Not sure why we should expect the declaration we augment to have to be somewhere specific.

We can explain it this way:
* The introductory declaration introduces a name and declaratio into the top-level scope.
* The augmenting declarations augment that declaration of the top-level scope.
* The augmenting declarations are applied in "before/after generalized source order", because an order is needed for a few cases.
* To make this not be too confusing, the introductory declaration must be before all augmenting declarations.

(And similarly for members of a class-like scope, that scope is the same everywhere in the library, collecting all introducing declarations.)

And we define the "Before" ordering of declarations to users in an easy to remember way ("file first in source roder, then its parts in directive order, recursively", aka. "depth first, preorder").

If users *start out* with that explanation, I don't think it should confusing to them that one declaration is not below another, they have never been introduced to the idea that it _should_ be.

The library/part story has always been that a library defines a flat namespace. Its declarations can be structured into parts, for structure and (now) import modularization, but the result is still a single flat namespace. Every library declaration can be accessed from anywhere. It's not surprising that a class in one part can extend a class in another sibling part.

With augmentations, order begins to matter, but only between introducing and augmenting declarations of the same entity.
Rather than a flat _set_ of declarations, the files of a library now define a flat *list* of (introducing and augmenting) declarations.
Those, together, still define a _set_ of (effective, fully augmented, semantic, or whatever we call it) declarations.

What is important is that users (and code generators) know which code movements are pure refactorings, and which are semantically signficiant. (If they care about the order, which they can reasonably choose not to.)

They can understand that by understanding the order, and understanding that (ignoring imports), you can flatten a library and its parts into a single flat file.

If parts have imports, you may need to prefix the imports and make conflicting extension invocations explicit, but you *can* still do that.

(It's sort-of like the linearization C includes, except our parts are included at the end of the file instead of the start. That's because we want the main file to have the declarations, and the parts to have the definitions, the opposite of C style.)


The "below" ordering isn't really needed outside of the spec for enhanced parts. It's not something I'd use to explain import scoping to end-users, I'd do that more directly.
I don't think any end-user should need to be introduced to the "below" relation, the "before" is all they need.

(I better stop, I keep trying to elaborate a point that I hope I have already made!)

working/augmentations/feature-specification.md
lrhn: "Mixin application class declarations can't be augmented."

Can be confusing to a user that `class C extends S with M;` is not a mixin application class, and `augment C with M2;` does seem to augment the mixin application super-class of `C`, that's what it applies `M2` to.

(And also because you augment *declarations*, not classes.)

So maybe also give an example of the specific syntax that this is about.

"Mixin application class declarations, like `class C = S with M;`, cannot be augment.

_(And they can't be augmented mostly because the syntax sucks for it. You can't omit the `S with M`. Otherwise adding more mixin applications or interfaces should be safe. Should we say that this is why? 
I think, with primary constructor syntax allowing a `;` as a class body, we could allow `augment class C with M2 implements I2;` to augment `class C = S with M implements I;`. You just can't write a `{}` as the body then, because `class C = S ..` doesn't have a new class body.)_

lrhn: generally -> otherwise

working/augmentations/feature-specification.md
```diff
@@ -659,19 +661,7 @@ It's a **compile-time** error if:
 
 *   An augmentation declaration is applied to a declaration of a different kind.
     For example, augmenting a `class` with a `mixin`, an `enum` with a function,
```
lrhn: Consider:
```markdown
* An augmenting declaration is a different kind of declaration than the introducing 
   declaration of the same name.
   For example, the introducing declaration is a `class` declaration and an augmenting
   declaration with the same name is a `mixin` declaration, or one is an `enum` declaration 
   and the other a function declaration.
```

working/augmentations/feature-specification.md
lrhn: First occurrence of "corresponding member", so maybe 

> ... corresponding member (the introducing member with the same name)
> in the same augmentation context,

working/augmentations/feature-specification.md
```diff
@@ -684,12 +674,13 @@ It's a **compile-time** error if:
...
-    the same number of type parameters with the exact same type parameter
-    names (same identifiers) and bounds if any (same *types*, even if they
-    may not be written exactly the same in case one of the declarations
-    needs to refer to a type using an import prefix).
+*   The type parameters of the augmenting declaration do not match the augmented
+    declarations's type parameters. This means there must be the same number of
+    type parameters with the exact same type parameter names (same identifiers)
+    and bounds if any (same *types*, even if they may not be written exactly the
+    same in case one of the declarations needs to refer to a type using an
+    import prefix). The augmenting declaration may choose to omit bounds in
```
lrhn: Must it omit all bounds, or can it omit only some?
```dart
foo<X extends num, Y extends X>();
augment foo<X extends num, Y>() { ... }
```
Valid or not?

Consider allowing omitting names too:
```dart
@marker
augment foo<_, _>();
```
If specified, they must be the same. If not specified, the augmenting declaration cannot refer to the name.

working/augmentations/feature-specification.md
lrhn: No mention of `mixin` `on` clause or `extension` `on` clause.
If deliberate, maybe says it explicitly:

> A `mixin` augmentation cannot specify an `on` clause,
> and an `extension` declaration cannot add an `on` clause.

(The extension could be allowed to, or required to, repeat the `on` clause.)

lrhn: (after type annotation inheritance) -> (if not omitted)

Mainly to avoid the "type annotation inheritance" which is ambiguous with what we do to instance members during type inference. But also because it's simpler - if the return type annotation is omitted, it *will* be the return type of the augmented declaration, so there is no need to check that it's the same. If not omitted, it won't be "inherited", so it's the same thing, just more direct.

working/augmentations/feature-specification.md
```diff
@@ -803,17 +794,27 @@ augment class C {
...
-introduced implicitly with the value null in the case where the parameter
+introduced implicitly with the value `null` in the case where the parameter
 has a nullable declared type, and no default values for that parameter are
 specified in the augmentation chain.
 
 It's a **compile-time** error if:
 
+*   An augmentation declaration is applied to a declaration of a different kind.
+    For example, augmenting a function with a `class`, a method with a getter, a
+    constructor with a static method, etc.
```
lrhn: Repeats line 662, which said the same thing for class-like augmentations.
Since this error goes across that destinction, the mismatched kinds can be either, maybe specify it before the "class-like declarations" and "function-declarations" split.
Then, inside the two parts, you don't have to worry about whether it's the introducing or the augmenting declaration which is the class-like/funciton declaration, it'll always be both.

working/augmentations/feature-specification.md
lrhn: Is this the same as "if the parameter type is not omitted, it is the same type as the corresponding parameter of the introducing declaration"?

(I'd rather not imply that we *update* the declarations with an inherited type. The syntax is what it is, and does not change. If "the augmented declaration" is the prior declaration, augmenting or introducing, then it may not have specificed a type. If "the augmented declaration" is some combination of all the prior declarations, then it works, but then we have to say how it's created, and how this declaration applies to it.)

lrhn, line 963: (The name *can* be `C.factory`. 😉 )

lrhn (approved):
Feel free to ignore comments not related to this PR.
(If I read, I comment. Can't help it.)

working/augmentations/feature-specification.md
munificent: Yup, I agree with all that. Fixed the `{}` to `;` in the last example. Thanks!

working/augmentations/feature-specification.md
munificent: Changed to "affect". The point of the sentence is basically establishing the scope of the feature. Using "augment" here would be sort of confusingly tautological. :)

working/augmentations/feature-specification.md
munificent: >  I don't think it should confusing to them that one declaration is not below another, they have never been introduced to the idea that it should be.

We do have evidence that at least a couple of people on the language team have found it unpleasantly surprising that the before ordering means an augmentation can apply to an introductory declaration in a sibling part file.

But I agree that it's the semantics we want and I think most of the time it will just work out without users having to think about it much.

working/augmentations/feature-specification.md
munificent: I think the current paragraph works OK.

working/augmentations/feature-specification.md
munificent: > Must it omit all bounds, or can it omit only some?

Clarified.

> Consider allowing omitting names too:

That's an anti-goal. The reason we require augmentations to re-list the parameters and type parameters is so that when those identifiers are used in the body, you can see where they come from at the declaration.

working/augmentations/feature-specification.md
munificent: Those restrictions are handled by the grammar [here](https://github.com/dart-lang/language/blob/main/working/augmentations/feature-specification.md#class-like-declarations) and [here](https://github.com/dart-lang/language/blob/main/working/augmentations/feature-specification.md#extensions).

working/augmentations/feature-specification.md
munificent: Good point. I couldn't find a good place to put this error centrally so it sort of got half split up. Consolidated it to one error earlier in the proposal.

working/augmentations/feature-specification.md
munificent: Yes, the parenthetical is about the augmentation flavor of parameter type inference, not override inference. Changed this to "(if not omitted)" too.

Issue #4485, opened 2025-08-06
Shorter constructor declarations


munificent:
We're working on [primary constructors][]. Among many nice things about this feature is that it eliminates the verbosity of repeating the entire class name when writing a constructor, which may sometimes be quite long:

[primary constructors]: https://github.com/dart-lang/language/blob/main/working/2364%20-%20primary%20constructors/feature-specification.md

```dart
// Today:
class SomeEmbarrassinglyLongClassNameWhy {
  SomeEmbarrassinglyLongClassNameWhy() {
    print('Phew!');
  }
}

// Using an in-body declaring constructor:
class SomeEmbarrassinglyLongClassNameWhy {
  this() {
    print('Phew!');
  }
}
```

This is nice, but since a class can only have one in-body declaring constructor and it must be generative, it's of limited help for classes with multiple constructors or factory constructors.

I think there is general sentiment on the language team to come up with a shorter notation for *all* constructor declarations. We've discussed it informally but I don't think there's a tracking issue. Now there is. :)

A couple of options:

### Use `this` instead of the class name

Generalizing from in-body declaring constructors, we can use `this` for all constructors:

```
Current syntax                            Proposed
----------------------------------------- -------------------
LongClassName() {}                        this() {}
LongClassName.name() {}                   this.name() {}
const LongClassName() {}                  const this() {}
const LongClassName.name() {}             const this.name() {}
LongClassName(): this.other();            this(): this.other();
LongClassName.name(): this();             this.name(): this();
const LongClassName(): this.other();      const this(): this.other();
const LongClassName.name(): this();       const this.name(): this();
factory LongClassName() { ... }           factory this() { ... }
factory LongClassName.name() { ... }      factory this.name() { ... }
factory LongClassName() = D;              factory this() = D;
factory LongClassName.name() = D;         factory this.name() = D;
const factory LongClassName() = D;        const factory this() = D;
const factory LongClassName.name() = D;   const factory this.name() = D;
```

Literally just replace the class name with `this` in all constructors. This syntax makes generative constructors ambiguous with an in-body declaring constructor. I believe that's fine. An in-body declaring constructor with no field parameters is semantically identical to a generative constructor. So when the syntax is fully ambiguous, the semantics are the same so the distinction doesn't matter.

If a constructor has any field parameters, then their presence implies that the constructor is unambiguously a declaring one.

Put another away, with this syntax, we'd say that the definition of an in-body declaring constructor is a constructor that has at least one field parameter.

[...]

PR #4543, opened 2025-10-22
Update declaring constructors to allow the new constructor syntax

eernstg:
See #4485 for background information.

This PR changes the feature specification of declaring constructors such that its grammar rules support the new style of constructor declarations (like `new();` and `const factory name() = D;`).

It handles the ambiguity among `factory();` being a method or a factory constructor declaration whose name is the name of the enclosing class/mixin-class/enum/extension-type declaration by making `factory` a reserved word.

### DIFF
--- a/accepted/future-releases/primary-constructors/feature-specification.md
+++ b/accepted/future-releases/primary-constructors/feature-specification.md
@@ -4,7 +4,7 @@ Author: Erik Ernst
 
 Status: Accepted
 
-Version: 1.10
+Version: 1.11
 
 Experiment flag: declaring-constructors
 
@@ -638,13 +638,13 @@ constructors as well.
 <factoryConstructorSignature> ::= // Modified rule.
      'const'? 'factory' <constructorName> <formalParameterList> // Old form.
    | 'const'? <factoryConstructorHead> <formalParameterList>; // New form.
- 
+
 <redirectingFactoryConstructorSignature> ::= // Modified rule.
      'const'? 'factory' <constructorName> <formalParameterList> '='
      <constructorDesignation> // Old form.
    | 'CONST'? <factoryConstructorHead> <formalParameterList> '='
      <constructorDesignation>; // New form.
- 
+
 <constantConstructorSignature> ::= // Modified rule.
    : 'const' <constructorName> <formalParameterList> // Old form.
    | 'const' <constructorHead> <formalParameterList> // New form.
@@ -702,14 +702,6 @@ constructors as well.
      ('=' <expression>)?;
 ```
 
-The word `factory` is removed from the set of built-in identifiers and
-added to the set of reserved words.
-
-*This allows the new constructor declaration syntax such as `factory();` to
-be unambiguous.  It is also a breaking change because `factory` can no
-longer be the name of a declaration. However, this occurs only infrequently
-today.*
-
 A _declaring constructor_ declaration is a declaration that contains a
 `<declaringConstructorSignature>` with a `<declaringParameterList>`, or a
 declaration that contains a `<declaringConstantConstructorSignature>`, or
@@ -720,6 +712,25 @@ type declaration, together with a declaration in the body that contains a
 
 A class or extension type declaration whose class body is `;` is treated as
 a declaration whose body is `{}`.
+
+The grammar is ambiguous with regard to the keyword `factory`.  *For
+example, `factory() => C();` could be a method named `factory` with an
+implicitly inferred return type, or it could be a factory constructor.*
+
+This ambiguity is resolved as follows: When a Dart parser expects to parse
+a `<memberDeclaration>`, and the first token is `factory`, it proceeds to
+parse the following input as a factory constructor.
+
+*Another special exception is introduced with factory constructors in order
+to avoid breaking existing code:*
+
+A factory constructor declaration of the form `factory C(...` where `C`
+is the name of the enclosing class, mixin class, enum, or extension type is
+treated as if `C` had been omitted.
+
+*Without this special rule, such a declaration would declare a constructor
+named `C.C`. With this rule it declares a constructor named `C`, which
+is the same as today.*
 
 Let _D_ be a class, extension type, or enum declaration.
 
@@ -1029,6 +1040,15 @@ then _k2_ has an initializer list with the same elements in the same order.
 
 Finally, _k2_ is added to _D2_, and _D_ is replaced by _D2_.
 
+### Language versioning
+
+This feature is language versioned.
+
+*It introduces a breaking change in the grammar, which implies that
+developers must explicitly enable it. In particular, `factory() {}` in a
+class body used to be a method declaration. With this feature it will be
+a factory constructor declaration.*
+
 ### Discussion
 
 This proposal includes support for adding the declaring header parameters to
@@ -1060,6 +1080,13 @@ far removed from any syntactic hint that the constructor must be constant
 of declaration, and the constructor might be non-const).
 
 ### Changelog
+
+1.11 - October 30, 2025
+
+* Introduce the new syntax for the beginning of a constructor declaration
+  (`new();` rather than `ClassName();`). Specify how to handle the
+  ambiguity involving the keyword `factory`. Clarify that this feature is
+  language versioned.
 
 1.10 - October 3, 2025
 


eernstg:
Testing the libraries in `tests/language`, the fact that `factory` is now a reserved word gives rise to 23 new parsing failures (we already have some failures because of parameters declared as `var p`). Apparently, there are no methods named `factory` here.

A couple of errors arise because the test specifically tests the treatment of built-in identifiers and reserved words, so that's to be expected.

The remaining errors are all concerned with the use of `factory` as the second part of a constructor name (as in `class C { factory C.factory() => ...; }`). This could occur in real code, but it seems likely that it is a more natural choice in test code, and not so common in production.

We could try out the radical approach (making `factory` a reserved word) in internal code and see if it causes non-trivial breakage (a quick search yields about 12 constructors whose name is of the form `C.factory`, apparently almost all of them in testing code). Alternatively, we could allow `factory` everywhere except as a type name and as the name of a method (which yields minimal breakage).

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -575,64 +575,81 @@ constructors as well.
 ```ebnf
 <classDeclaration> ::= // First alternative modified.
      (<classModifiers> | <mixinClassModifiers>)
-     'class' <classNamePart> <superclass>? <interfaces>? <classBody>
+     'class' <classNameMaybePrimary> <superclass>? <interfaces>? <classBody>
```
munificent, line 578: Suggestion/nit: "maybe" implies "optional" to me but not a choice from two alternatives. Perhaps a better name is "classNameOrPrimary"?

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -575,64 +575,81 @@ constructors as well.
...
    | ...;
 
-<primaryConstructorNoConst> ::= // New rule.
-     <typeIdentifier> <typeParameters>?
-     ('.' <identifierOrNew>)? <declaringParameterList>
+<primaryConstructor> ::= // New rule.
+     'const'? <typeWithParameters> ('.' <identifierOrNew>)?
```
munificent, line 582: For what it's worth, I never thought there was value in supporting `Foo.new` in constructor declarations, and we don't necessarily have to support it for primary constructors. We could just do `('.' <identifier>)` here.

But if you think it's worth it for consistency, I could go either way. :)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -685,31 +702,39 @@ constructors as well.
      ('=' <expression>)?;
 ```
 
+The word `factory` is removed from the set of built-in identifiers and
+added to the set of reserved words.
+
+*This allows the new constructor declaration syntax such as `factory();` to
+be unambiguous.  It is also a breaking change because `factory` can no
+longer be the name of a declaration. However, this occurs only infrequently
+today.*
```
munificent: We should be explicit here that this is a language versioned change. We'd say that `factory` is a built-in identifier in pre-feature libraries and a reserved word in later versions.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -575,64 +575,81 @@ constructors as well.
...
+ 
+<redirectingFactoryConstructorSignature> ::= // Modified rule.
+     'const'? 'factory' <constructorName> <formalParameterList> '='
+     <constructorDesignation> // Old form.
+   | 'CONST'? <factoryConstructorHead> <formalParameterList> '='
+     <constructorDesignation>; // New form.
+ 
+<constantConstructorSignature> ::= // Modified rule.
+   : 'const' <constructorName> <formalParameterList> // Old form.
+   | 'const' <constructorHead> <formalParameterList> // New form.
```
munificent, line 650: All the "old form" / "new form" comments are very helpful here.

munificent (approved):
This looks great!

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: It is definitely determining the class name (so it's always `className`), and then it may also include a primary constructor (so that primary constructor is optional). So it's not a class name _or_ a primary constructor. I don't know if there's a better way to indicate this "definitely class name and possibly primary constructor" relationship.. I'll leave this thread open and keep the non-terminal for now.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: I think we should be consistent, and then aim to consistently get rid of `.new` in constructor declarations. I suspect that it is really rare.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: This paragraph has been removed now because we did not make `factory` a reserved word after all. I've added a short section near the end to say that the feature is language versioned. I haven't mentioned the actual version because this may still change (I think).

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Thanks!

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Created https://github.com/dart-lang/language/issues/4550 in order to resolve this terminology issue.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -688,28 +705,47 @@ constructors as well.
...
+This ambiguity is resolved as follows: When a Dart parser expects to parse
+a `<memberDeclaration>`, and the first token is `factory`, it proceeds to
+parse the following input as a factory constructor.
+
+*Another special exception is introduced with factory constructors in order
+to avoid breaking existing code:*
+
+A factory constructor declaration of the form `factory C(...` where `C`
+is the name of the enclosing class, mixin class, enum, or extension type is
+treated as if `C` had been omitted.
```
lrhn, line 729: Would it be more direct to say that it the grammar `factory C(...` is not allowed to parse as a new-style constructor when `C` is the name of the surrounding thing.

Then there is no grammar ambiguity, and the syntax *only* parses as an old-style factory constructor. (And if we ever remove the old-style grammar, we won't still accept `factory C()`, because we forget to remove this special-case. Better to not have it to begin with.)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -575,64 +575,81 @@ constructors as well.
...
+     'const'? 'factory' <constructorName> <formalParameterList> // Old form.
+   | 'const'? <factoryConstructorHead> <formalParameterList>; // New form.
+
+<redirectingFactoryConstructorSignature> ::= // Modified rule.
+     'const'? 'factory' <constructorName> <formalParameterList> '='
+     <constructorDesignation> // Old form.
+   | 'CONST'? <factoryConstructorHead> <formalParameterList> '='
+     <constructorDesignation>; // New form.
+
+<constantConstructorSignature> ::= // Modified rule.
```
lrhn, line 648: Also declared on line 617.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -575,64 +575,81 @@ constructors as well.
...
    | 'new'
 
+<factoryConstructorSignature> ::= // Modified rule.
+     'const'? 'factory' <constructorName> <formalParameterList> // Old form.
+   | 'const'? <factoryConstructorHead> <formalParameterList>; // New form.
+
+<redirectingFactoryConstructorSignature> ::= // Modified rule.
+     'const'? 'factory' <constructorName> <formalParameterList> '='
+     <constructorDesignation> // Old form.
+   | 'CONST'? <factoryConstructorHead> <formalParameterList> '='
```
lrhn, line 645: 'CONST' -> 'const'

Same, could then be:
```dart
  'const'? <factoryDesignator> <formalParameterList> '='
```
...  which is just `<factoryConstructorSignature> '='`, right?

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -688,28 +705,47 @@ constructors as well.
...
+implicitly inferred return type, or it could be a factory constructor.*
+
+This ambiguity is resolved as follows: When a Dart parser expects to parse
+a `<memberDeclaration>`, and the first token is `factory`, it proceeds to
+parse the following input as a factory constructor.
+
+*Another special exception is introduced with factory constructors in order
+to avoid breaking existing code:*
+
+A factory constructor declaration of the form `factory C(...` where `C`
```
lrhn, line 727: Or `const factory C(...`.

It needs to be said. 😞

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -688,28 +705,47 @@ constructors as well.
...
+is the name of the enclosing class, mixin class, enum, or extension type is
+treated as if `C` had been omitted.
+
+*Without this special rule, such a declaration would declare a constructor
+named `C.C`. With this rule it declares a constructor named `C`, which
```
lrhn, line 732: Without this special rule, the grammar would be ambiguous.
And because the rule for the old form is first, it would parse as the old form.
Which means this sentence isn't needed, the grammar would parse it that way anyway.

... until we introduce static extension constructors which can only use the new syntax,
and if that's restricted at the grammar level, the syntax would then start meaning `C.C`. So let's not do the restriction at the grammar level, allow both to be parsed, then say it's an error to use the old style syntax.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -575,64 +575,81 @@ constructors as well.
...
 
 <constructorSignature> ::= // Modified rule.
-     <constructorName> <formalParameterList>
+     <constructorName> <formalParameterList> // Old form.
+   | <constructorHead> <formalParameterList> // New form.
    | <declaringConstructorSignature>;
 
 <declaringConstructorSignature> ::= // New rule.
-     'this' ('.' <identifierOrNew>)? <declaringParameterList>?;
+     'this' <identifier>? <declaringParameterList>?;
```
lrhn, line 615: I think this should be `'this' (<identifier>? <declaringParameterList>)?`;.

You can't have an identifier without a parameter list.

The `const`, name-identifier and parameter list either all go in the class header or all in an in-body declaration.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -575,64 +575,81 @@ constructors as well.
...
+<factoryConstructorHead> ::= // New rule.
+     'factory' <identifier>?;
 
 <identifierOrNew> ::=
      <identifier>
    | 'new'
 
+<factoryConstructorSignature> ::= // Modified rule.
+     'const'? 'factory' <constructorName> <formalParameterList> // Old form.
+   | 'const'? <factoryConstructorHead> <formalParameterList>; // New form.
```
lrhn, line 640: (Would it make sense to make this:
```dart
<factoryConstructorSignature> ::=
    'const'? <factoryDesignator> <formalParameterList>

<factoryDesignator> ::=
       'factory' <constructorName>
    | <factoryConstructorHead>
```
(Strawman production name!)
It seems that the rest of the productions are the same, and mean the same thing,
so we can abstract only the part that actually is different.)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -575,64 +575,81 @@ constructors as well.
...
+<redirectingFactoryConstructorSignature> ::= // Modified rule.
+     'const'? 'factory' <constructorName> <formalParameterList> '='
+     <constructorDesignation> // Old form.
+   | 'CONST'? <factoryConstructorHead> <formalParameterList> '='
+     <constructorDesignation>; // New form.
+
+<constantConstructorSignature> ::= // Modified rule.
+   : 'const' <constructorName> <formalParameterList> // Old form.
+   | 'const' <constructorHead> <formalParameterList> // New form.
+   | <declaringConstantConstructorSignature>;
```
lrhn, line 651: If we didn't use `<declaringFormalParameterList>` for `<declaringConstrntConstructorSignature>`, but just said that it is an error to have a declaring parameter anywhere else, this could also be:
```ebnf
<constantConstructorSignature> ::= 
     'const' <constructorDesignator> <formalParameterList>;

<constructorDesignator> ::=
       <constructorName>   // old form
    | <constructorHead>    // new form
    | <declaringConstructorHead> // also new
    ;

<declaringConstructorHead> ::=
    'this' <identifier>?;
```
which also avoids repeating things that mean the same thing between different productions.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -688,28 +705,47 @@ constructors as well.
...
 A class or extension type declaration whose class body is `;` is treated as
 a declaration whose body is `{}`.
 
+The grammar is ambiguous with regard to the keyword `factory`.  *For
+example, `factory() => C();` could be a method named `factory` with an
+implicitly inferred return type, or it could be a factory constructor.*
+
+This ambiguity is resolved as follows: When a Dart parser expects to parse
+a `<memberDeclaration>`, and the first token is `factory`, it proceeds to
+parse the following input as a factory constructor.
```
lrhn, line 722: (Could add commentary like "_This is similar to how a statement starting with `switch` or `{` are parsed as a switch statement or block, and never as an expression statement._".)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -1004,6 +1040,15 @@ then _k2_ has an initializer list with the same elements in the same order.
...
 Finally, _k2_ is added to _D2_, and _D_ is replaced by _D2_.
 
+### Language versioning
+
+This feature is language versioned.
+
+*It introduces a breaking change in the grammar, which implies that
+developers must explicitly enable it. In particular, `factory() {}` in a
+class body used to be a method declaration. With this feature it will be
+a factory constructor declaration.*
```
lrhn, line 1050: The feature also disallows `var x`, `final x` and `final T x` as parameters in all non-declaring constructors (Line 122/130.), which is much more far-reaching and potentially breaking.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Thanks, fixed!

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Good catch! (We don't have a non-terminal `<factoryDesignator>`, but the simplification works). Done.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Indeed, done!

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Declaring constructor signatures do not exist now.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Done.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Done.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Good point! I changed the grammar such that the old form rule does not include support for the class name alone, it requires `ClassName.id`, and all declarations using a single identifier will be parsed according to the new rule. They will then implicitly eliminate the class name when that's the given identifier. I think that's an OK behavior to have now and in the future (I don't think we're going to _want_ constructors whose name is `ClassName.ClassName`).

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: I changed the grammar to require a two part name in order to use the "old form" rule. This means that `factory id(...`, `const factory id(...`, etc. are parsed as the new form, and the new form will (for now) treat `factory C(...` where `C` is the class name as if the class name had been omitted.

I don't think anybody will ever really need a constructor whose name is `ClassName.ClassName` and hence we can just keep this treatment for now (as long as we wish to allow developers to keep those factory declarations unchanged), and then we can make it an error to have the class name there (not a syntax error, just a compile-time error).

This implies that developers will not use `factory ClassName(...` at all, and I think that's actually better than having support for constructors named `ClassName.ClassName` at some point in the future.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: I reorganized and simplified several rules. Much better, thanks!

PR #4545, opened 2025-10-24
Clarify how augmentations affect existing compile-time errors.

### LINKED ISSUE #3690: Augmentations should specify allowed partial declarations.

munificent:
Fix #3690.

### DIFF
--- a/working/augmentations/feature-specification.md
+++ b/working/augmentations/feature-specification.md
@@ -1005,40 +1005,54 @@ the combined declaration.*
 
 ### Compile errors with augmentations
 
-Many existing compile-time errors in the language specification are only
-meaningful after semantic analysis so only make sense to check after
-augmentations are applied. For example:
+Prior to augmentations, the definition of a semantic entity is produced by a
+single syntactic declaration. That allows the language specification to refer to
+those entities interchangeably. With augmentations, that is no longer the case.
+A single semantic entity may be the product of multiple syntactic declarations
+(an introductory and any number of augmentations). This raises the question of
+whether existing compile errors apply to syntactic declarations or semantic
+definitions. For example, the specification says:
 
-> Let `C` be a concrete class declared in library `L`, with interface `I`.
-> Assume that `I` has a member signature `m` which is accessible to `L`. It is a
-> compile-time error if `C` does not have a concrete member with the same name as
-> `m` and accessible to `L`, ...
+> It is a compile-time error if two elements in the type list of the
+> `implements` clause of a class `C` specifies the same type `T`.
 
-This error says that a non-abstract class must fully implement its interface.
-We don't know what methods a class even has until semantic analysis lets us
-see inherited methods, and that happens after augmentation. Thus this error is
-checked after augmentations have been applied. Therefore, there is no error in:
+Thus this is an error:
 
 ```dart
-abstract class I {
-  m();
-}
-
-class C implements I {}
-
-augment class C {
-  m() {}
-}
+class C implements I, I {}
 ```
 
-The general rule is that compile-time errors should be checked after
-augmentations are applied when possible to do so. In other words, if the
-library is well-formed enough that augmentations *can* be applied, then they
-should be. And if doing so eliminates what would otherwise be a compile-time
-error, then that error should not be reported.
+But what about:
 
-Put another way, moving code out of an introductory declaration into an
-augmentation of it should not cause new errors to appear.
+```dart
+class C implements I {}
+
+augment class C implements I {}
+```
+
+Each syntactic declaration of `C` only mentions the interface once. But the
+resulting definition produced by applying augmentations has an `implements`
+clause with `I` in it twice. To which entity does the error apply?
+
+The general rule is that **compile-time errors apply to semantic definitions
+whenever possible.** In other words, if the library is syntactically well-formed
+enough that augmentations *can* be applied, then they should be. And if doing so
+eliminates what would otherwise be a compile-time error, then that error should
+not be reported.
+
+In the example above, there is an error because the resulting definition does
+have the same interface twice in the `implements` clause. *(Though note that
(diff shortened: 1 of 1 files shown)


working/augmentations/feature-specification.md
```diff
@@ -1003,6 +1003,43 @@ the combined declaration.*
...
+augmentations are applied. For example:
+
+> Let `C` be a concrete class declared in library `L`, with interface `I`.
+> Assume that `I` has a member signature `m` which is accessible to `L`. It is a
+> compile-time error if `C` does not have a concrete member with the same name as
+> `m` and accessible to `L`, ...
+
+This error says that a non-abstract class must fully implement its interface.
+We don't know what methods a class even has until semantic analysis lets us
+see inherited methods, and that happens after augmentation. Thus this error is
```
lrhn: We have never *formally* specified the order that things happen in.
We have definitely not defined that they must happen *globally* in the same order.
(We can do a lot of things to a superclass before we even start looking at a subclass, otherwise modular compilation wouldn't work, so we could rely on a superclass having found out which members are concrete before we even start looking at which members a subclass have.)

That means that "Augmentation happening" is not a well-defined step.
Consider, instead of using temporal language, where augmentation is something that happens at some point, as if it *changes* something, rather be declarative and just say what a set of augmenting declarations *mean*. That way they colletively define something new, instead of changing something existing. ("Changing" something is horribly complicated to get right, because now there is a before and an after *which both exist*, so we now have to define twice as many things, and specifiy precisely which of them other things refer to, and keep that straight.)

> Individual syntactic declarations, introducing or augmenting, combine to define
> a single definition. Whether a class definition has a member or not, and
> whether that member is abstract or concrete, depends on all the class declarations
> that contribute to that definition.
>
> Syntactic errors usually apply to a single syntactic declaration.
> Only if a declaration is syntactically valid does it make sense to interpret
> it in the context of the rest of the program.
>
> This feature changes some of the syntactic rules, to introduce the new `augment` 
> syntax and to allow some new incomplete declarations. 
> The feature also introduces new errors to ensure that the declaring and augmenting
> declarations for a single name are well-formed: All of the same kind, at most one
> introducing declaration, and that one must be before augmenting declarations,
> and at most one complete declaration adding implementation.
> 
> If those requirements are satisfied, then the set of declarations for the same thing
> can be combined and and the properties they define can be derived.
>
> Existing non-syntactic errors for declarations, for example how they interact with other
> definitions and with inheritance, should then be applied to the combined definitions.

Or something. The idea is: Don't say that augementations are applied, as if there is a before and after. Say that they *define* something new, which is effectively meaning of a pre-feature declaration. Then say that existing checks apply to that new thing, instead of the individual declarations.

working/augmentations/feature-specification.md
```diff
@@ -1003,6 +1003,43 @@ the combined declaration.*
...
+```
+
+The general rule is that compile-time errors should be checked after
+augmentations are applied when possible to do so. In other words, if the
+library is well-formed enough that augmentations *can* be applied, then they
+should be. And if doing so eliminates what would otherwise be a compile-time
+error, then that error should not be reported.
+
+Put another way, moving code out of an introductory declaration into an
+augmentation of it should not cause new errors to appear.
```
lrhn: (Comma after "it"?)

working/augmentations/feature-specification.md
munificent: This is a good point. I took a stab at rewriting this in a different way. Let me know what you think.

working/augmentations/feature-specification.md
munificent: Not here. (There are some places in English where you hear a slight pause in the cadence but where textually a comma isn't needed or expected like before "but" here.)

munificent:
Let me know if this looks better.

PR #4548, opened 2025-10-29
Fix a typo in Private named parameters proposal

sgrekhov:
Fix a typo (final variable already initialized).


munificent:
Thanks as always, Sergey! :)

PR #4561, opened 2025-11-05
Adjust wording in the 'Private named parameters' proposal

eernstg:
The beginning of the 'private named parameters' proposal may currently be interpreted to mean that a declaring or initializing formal with a name of the form `_id` is treated as if it had had the name `id` for all purposes other than the naming of the implicitly induced instance variable. It is in fact still accessed as `_id` in the initializer list and (if not declaring or initializing) in the constructor body. This PR adds a few words to make it explicit that the corresponding public name is only used at call sites. It could be spelled out even more. However, that's probably not necessary because the details are given later in the proposal, and we just want to get the reader on track to the overall properties of the proposal here.

### DIFF
--- a/accepted/future-releases/2509-private-named-parameters/feature-specification.md
+++ b/accepted/future-releases/2509-private-named-parameters/feature-specification.md
@@ -228,9 +228,9 @@ encapsulation.
 
 The basic idea is simple. We let users use a private name in a named parameter
 when the parameter also initializes or declares a field. The compiler removes
-the `_` from the argument name but keeps it for the corresponding field. In
-other words, we do exactly what users are doing by hand when they write an
-initializer like:
+the `_` from the argument name at call sites but keeps it for the corresponding
+field. In other words, we do exactly what users are doing by hand when they
+write an initializer like:
 
 ```dart
 class House {


munificent (approved):

munificent:
Thanks!

munificent:
> It is in fact still accessed as `_id` in the initializer list and (if declaring, not initializing) in the constructor body.

With that parenthetical, are you saying that a declaring field parameter is in scope in the constructor body? That's not what I'd expect:

```dart
class C({this._s}) {
  String? _s;
  void Function()? _capture;
  init {
    _capture = () {
      print(_s);
    }
  }
}

main() {
  var c = C(s: 'before');
  c._s = 'after';
  c._capture!.call();
}
```

I expect this program would print "after", meaning that `_s` in the constructor body refers to the instance variable, not the parameter. That's what a corresponding program using an initializing formal today does:

```dart
class C {
  String? s;
  void Function()? capture;
  C({this.s}) {
    capture = () {
      print(s);
    };
  }
}

main() {
  var c = C(s: 'before');
  c.s = 'after';
  c.capture!.call();
}
```

The primary constructors proposal says:

> The same parameter list also introduces the primary parameter scope, whose enclosing scope is also the body scope of the class. Every primary parameter which is **not declaring**, not initializing, and not a super parameter is entered into this scope.
> 
> The primary parameter scope is the current scope for the body of the body part of the declaring header constructor, if any.

My emphasis added. So I think the proposal is in line with what I'm saying. Am I just misreading your parenthetical?

eernstg:
> (if declaring, not initializing)

Oops, that was a typo. I just corrected the phrase to 'if not declaring or initializing'.

> .. are you saying that a declaring field parameter is in scope in the constructor body? That's not what I'd expect

Right, the declaring parameter is not in scope in the body, it is treated in the same way as an initializing formal.

> So I think the proposal is in line with what I'm saying.

Agreed!

Zhigan650 (commented):
spec_adjust_intro_nov25

Zhigan650:
@copilot `****`

Zhigan650:
Pull request successfully merged and closed

Zhigan650:
[](url)

PR #4562, opened 2025-11-06
Eliminate in-body declaring constructors

eernstg:
This PR eliminates the support for in-body declaring constructors in the feature specification about declaring constructors. It also changes the terminology to use 'primary' rather than 'declaring' about the newly introduced constructors, because all other kinds have been eliminated.

### DIFF
--- a/accepted/future-releases/primary-constructors/feature-specification.md
+++ b/accepted/future-releases/primary-constructors/feature-specification.md
@@ -10,11 +10,9 @@ Experiment flag: declaring-constructors
 
 This document specifies _primary constructors_. This is a feature that allows
 one constructor and a set of instance variables to be specified in a concise
-form in the header of the declaration. Some elements can still be specified in
-the class body: The primary constructor can have an initializer list, which must
-then be specified in the body. It can include assertions, instance variable
-initializers, and/or a superinitializer. The primary constructor can also have a
-constructor body, which is also specified in the class body.
+form in the header of the declaration of a class or a similar entity. If the
+primary constructor also needs an initializer list or body, those can be
+specified inside the class body.
 
 One variant of this feature has been proposed in the [struct proposal][],
 several other proposals have appeared elsewhere, and prior art exists in
@@ -55,13 +53,13 @@ class Point(var int x, var int y);
 ```
 
 A class that has a primary constructor cannot have any other
-non-redirecting generative constructors. This requirement must be upheld
-because it must be guaranteed that the primary constructor is executed on
-every newly created instance of this class.
+non-redirecting generative constructors. This ensures that the primary
+constructor is executed on every newly created instance of this class,
+which is necessary for reasons that are discussed later.
 
-In particular, every generative constructor in a declaration that has a
-primary constructor must be redirecting, and it must invoke the primary
-constructor (directly or indirectly). This can be seen as a motivation for
+In particular, every other generative constructor in a declaration that has
+a primary constructor must be redirecting, and it must invoke the primary
+constructor, directly or indirectly. This can be seen as a motivation for
 the word _primary_ because it makes all other generative constructors
 secondary in the sense that they depend on the primary one.
 
@@ -122,14 +120,16 @@ There is no way to indicate that the implicitly induced instance variable
 declarations should have the modifiers `late` or `external`. This omission
 is not seen as a problem in this proposal: They can be declared using the
 same syntax as today, and initialization, if any, can be done in a
-constructor body.
+constructor body. Note that it does not make sense to declare an instance
+variable as `late` if it is always initialized in the very first phase of
+the constructor execution.
 
 ```dart
 // Current syntax.
 class ModifierClass {
   late int x;
   external double d;
-  ModifierClass(this.x);
+  ModifierClass(this.x); // Can initialize `x`, but it preempts `late`.
 }
 
 // Using a primary constructor.
@@ -243,7 +243,7 @@ where there is no primary constructor:
 
 ```dart
 // Current syntax.
-class D<TypeVariable extends Bound> 
+class D<TypeVariable extends Bound>
     extends A with M implements B, C {
   final int x;
   final int y;
@@ -252,8 +252,8 @@ class D<TypeVariable extends Bound>
 
 // Using a primary constructor.
 class const D<TypeVariable extends Bound>.named(
-  var int x, [
-  var int y = 0,
+  final int x, [
+  final int y = 0,
 ]) extends A with M implements B, C;
 ```
 
@@ -426,27 +426,22 @@ class E({
 ```
 
 Note that the version with a primary constructor can initialize `z` in the
-declaration itself, whereas the other version need to use an element in the
-initializer list of the constructor to initialize `z`. This is necessary
-because `y` isn't in scope. Moreover, there cannot be other non-redirecting
-generative constructors when there is a primary constructor, but in the
-other version we could add another non-redirecting generative constructor
-which could initialize `w` with some other value, in which case we must
-also initialize `w` as shown.
-
-As an aside, we may get rid of all those occurrences of `required` in the
-situation where it is a compile-time error to not have them, but that is a
-separate proposal, [here][inferred-required] or [here][simpler-parameters].
-
-[inferred-required]: https://github.com/dart-lang/language/blob/main/working/0015-infer-required/feature-specification.md
-[simpler-parameters]: https://github.com/dart-lang/language/blob/main/working/simpler-parameters/feature-specification.md
+declaration itself, whereas the other version needs to use an element in
+the initializer list of the constructor to initialize `z`. This is
+necessary because `y` isn't in scope in the initializer list element in the
+non-primary-constructor class. Moreover, there cannot be other
+non-redirecting generative constructors when there is a primary
+constructor, but in the class that does not have a primary constructor we
+could add another non-redirecting generative constructor which could
+initialize `w` with some other value, in which case we must also initialize
+`w` as shown.
 
 ## Specification
 
 ### Syntax
 
-The grammar is modified as follows. Note that the changes include rules
-about extension type declarations because they're using primary
+The grammar is modified as follows. Note that the changes include grammar
+rules for extension type declarations because they're using primary
 constructors as well.
 
 ```ebnf
@@ -490,11 +485,11 @@ constructors as well.
 <constantConstructorSignature> ::= // Modified rule.
      'const' <constructorSignature>;
 
-<constructorName> ::= // Modified rule.
-     <typeIdentifier> ('.' identifierOrNew)?;
+<constructorName> ::=
+     <typeIdentifier> ('.' <identifierOrNew>)?;
 
 <constructorTwoPartName> ::= // New rule.
-     <typeIdentifier> '.' identifierOrNew;
+     <typeIdentifier> '.' <identifierOrNew>;
 
 <constructorHead> ::= // New rule.
      'new' <identifier>?;
@@ -507,19 +502,19 @@ constructors as well.
    | 'new'
 
 <factoryConstructorSignature> ::= // Modified rule.
-     'const'? 'factory' <constructorTwoPartName> 
+     'const'? 'factory' <constructorTwoPartName>
       <formalParameterList> // Old form.
-   | 'const'? <factoryConstructorHead> 
+   | 'const'? <factoryConstructorHead>
       <formalParameterList>; // New form.
 
 <redirectingFactoryConstructorSignature> ::= // Modified rule.
      <factoryConstructorSignature> '=' <constructorDesignation>;
 
 <primaryConstructorBodySignature> ::= // New rule.
-     'this' initializers?;
+     'this' <initializers>?;
 
 <methodSignature> ::= // Add one new alternative.
-     ... 
+     ...
    | <primaryConstructorBodySignature>;
 
 <declaration> ::= // Add one new alternative.
@@ -582,11 +577,12 @@ A _primary constructor_ declaration consists of a `<primaryConstructor>` in
 the declaration header plus optionally a member declaration in the body
 that starts with a `<primaryConstructorBodySignature>`.
 
-A class, mixin class, or extension type declaration whose class body is `;`
-is treated as the corresponding declaration whose body is `{}` and
-otherwise the same.
+A class, mixin class, or extension type declaration whose body is `;` is
+treated as the corresponding declaration whose body is `{}` and otherwise
+the same. This rule is not applicable to a `<mixinApplicationClass>` *(for
+instance, `class B = A with M;`)*.
 
-The grammar is ambiguous with regard to the keyword `factory`.  *For
+The grammar is ambiguous with regard to the keyword `factory`. *For
 example, `factory() => C();` could be a method named `factory` with an
 implicitly inferred return type, or it could be a factory constructor whose
 name is the name of the enclosing class.*
@@ -597,16 +593,17 @@ or one or more of the modifiers `const`, `augment`, or `external` followed
 by `factory`, it proceeds to parse the following input as a factory
 constructor.
 
-*This is similar to how a statement starting with `switch` or `{` are
-parsed as a switch statement or a block, never as an expression statement.*
+*This is similar to how a statement starting with `switch` or `{` is parsed
+as a switch statement or a block, never as an expression statement.*
 
 *Another special exception is introduced with factory constructors in order
 to avoid breaking existing code:*
 
-A factory constructor declaration of the form `factory C(...` optionally
-starting with zero or more of the modifiers `const`, `augment`, or
-`external` where `C` is the name of the enclosing class, mixin class, enum,
-or extension type is treated as if `C` had been omitted.
+Consider a factory constructor declaration of the form `factory C(...`
+optionally starting with zero or more of the modifiers `const`, `augment`,
+or `external`. Assume that `C` is the name of the enclosing class, mixin
+class, enum, or extension type. In this situation, the declaration declares
+a constructor whose name is `C`.
 
 *Without this special rule, such a declaration would declare a constructor
 named `C.C`. With this rule it declares a constructor named `C`, which
@@ -619,7 +616,7 @@ that does not contain a `<primaryConstructor>`, and the body of _D_
 contains a member declaration that starts with a
 `<primaryConstructorBodySignature>`.
 
-*It is an error to have the body part of a primary constructor in the  class
+*It is an error to have the body part of a primary constructor in the class
 body, but no primary constructor in the header.*
 
 A compile-time error occurs if a `<defaultDeclaringNamedParameter>` has the
@@ -641,7 +638,8 @@ The name of a primary constructor of the form
 
 A compile-time error occurs if a class, mixin class, enum, or extension
 type has a primary constructor whose name is also the name of a constructor
-declared in the body.
+declared in the body, or if it declares a primary constructor whose name is
+`C.n`, and the body declares a static member whose basename is `n`.
 
 Consider a class, mixin class, enum, or extension type declaration _D_ with
 a primary constructor *(note that it cannot be a `<mixinApplicationClass>`,
@@ -672,21 +670,15 @@ there is no conflict about the meaning of names in such initializing
 expressions. This means that we can allow those other non-redirecting
 generative constructors to coexist with a primary constructor.*
 
-A compile-time error occurs if the name of the primary constructor is the
-same as the name of a constructor in the body.
-
-*Moreover, it is an error if two constructor declarations in the body have
-the same name. This is just restating a compile-time error that we already
-have.*
-
 The declaring parameter list of the primary constructor introduces a new
 scope, the _primary initializer scope_, whose enclosing scope is the body
-scope of _D_. Every primary parameter is entered into this scope.
+scope of _D_. Each of the parameters in said parameter list is introduced
+into this scope.
 
 The same parameter list also introduces the _primary parameter scope_,
 whose enclosing scope is also the body scope of the class. Every primary
 parameter which is not declaring, not initializing, and not a super
-parameter is entered into this scope.
+parameter is introduced into this scope.
 
 The primary initializer scope is the current scope for the initializing
 expression, if any, of each non-late instance variable declaration. It is
@@ -799,7 +791,8 @@ Consider the situation where `p` has no type annotation:
   inference, just like an explicitly declared instance variable.*
 - otherwise, if `p` is optional and has a default value whose static type
   in the empty context is a type `T` which is not `Null` then `p` has
-  declared type `T`. When `T` is `Null`, `p` has declared type `Object?`.
+  declared type `T`. When `T` is `Null`, `p` instead has declared type
+  `Object?`.
 - otherwise, if `p` does not have a default value then `p` has declared
   type `Object?`.
 
@@ -823,7 +816,7 @@ before the name of _D_, or _D_ is an `enum` declaration.
 Consider the case where _k_ is a primary constructor. If the name `C` in
 _D_ and the type parameter list, if any, is followed by `.id` where `id` is
 an identifier then _k2_ has the name `C.id`. If it is followed by `.new`
-then _k2_ has the name `C`. If it is not followed by `.`  then _k2_ has the
+then _k2_ has the name `C`. If it is not followed by `.` then _k2_ has the
 name `C`. _D2_ omits the part derived from `'.' <identifierOrNew>` that
 follows the name and type parameter list in _D_, if said part exists.
 Moreover, _D2_ omits the formal parameter list _L_ that follows the name,
@@ -834,7 +827,7 @@ formal parameter is processed as follows.
 
 The formal parameters in _L_ and _L2_ occur in the same order, and
 mandatory positional parameters remain mandatory, and named parameters
-preserve the name and the modifier `required`, if any.  An optional
+preserve the name and the modifier `required`, if any. An optional
 positional or named parameter remains optional; if it has a default value
 `d` in _L_ then it has the default value `d` in _L2_ as well.
 
@@ -847,7 +840,7 @@ positional or named parameter remains optional; if it has a default value
   unchanged from _L_ to _L2_ *(this is a plain, non-declaring parameter)*.
 - Otherwise, a formal parameter (named or positional) of the form `var T p`
   or `final T p` where `T` is a type and `p` is an identifier is replaced
-  in _L2_ by `this.p`, along with its default value, if any.  Next, a
+  in _L2_ by `this.p`, along with its default value, if any. Next, a
   semantic instance variable declaration corresponding to the syntax `T p;`
   or `final T p;` is added to _D2_. It includes the modifier `final` if the
   parameter in _L_ has the modifier `final` and _D_ is not an `extension


eernstg:
This PR deals with the comments from @lrhn on an earlier PR #4543, and it updates the primary constructor feature specification to eliminate everything associated with in-body declaring constructors.

The terminology now uses 'primary' everywhere, because that's the only new kind of constructor which is being added by this feature. I did not change the experimental flag because that's a transient and relatively hidden thing.

eernstg:
https://dart-review.googlesource.com/c/sdk/+/460420 updates the spec parser to use the new grammar rules, which serves as a sanity check on the grammar.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -1,21 +1,20 @@
...
-allows one constructor and a set of instance variables to be specified in a
-concise form in the header of the declaration, or in the body. In the case
-where the constructor is specified in the header, some elements can still be
-specified in the class body, if present: The in-header constructor can have
-an initializer list in the body, including assertions, instance variable
-initializers, and/or a superinitializer. The in-header constructor can also
-have a body in the class body.
+This document specifies _primary constructors_. This is a feature that allows
+one constructor and a set of instance variables to be specified in a concise
+form in the header of the declaration. Some elements can still be specified in
```
lrhn: ... of the class declaration ...

There is no context here, so "the declaration" feels like it should refer to either the constructor or one of the instance variables.
(I know it's not always a class, but the next sentence uses "class" too.)

Also maybe "Some elements" -> "Some constructor elements".

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -53,38 +47,26 @@ class Point {
...
-
-A declaring header constructor is also known as a _primary constructor_,
-because all other generative constructors must invoke the primary one
-(directly or indirectly).
-
-A declaring body constructor is slightly less concise, but it allows the
-class header to remain simpler and more readable when there are many
-parameters. The previous example would look as follows using a declaring
-body constructor:
+because it must be guaranteed that the primary constructor is executed on
```
lrhn: Maybe "because it must be guaranteed" -> "to guarantee".

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -53,38 +47,26 @@ class Point {
...
+every newly created instance of this class.
 
-```dart
-// A declaration with the same meaning, using a declaring body constructor.
-class Point {
-  this(var int x, var int y);
-}
-```
+In particular, every generative constructor in a declaration that has a
```
lrhn: every *other* generative constructor

(than the primary constructor)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -53,38 +47,26 @@ class Point {
...
-A declaring header constructor is also known as a _primary constructor_,
-because all other generative constructors must invoke the primary one
-(directly or indirectly).
-
-A declaring body constructor is slightly less concise, but it allows the
-class header to remain simpler and more readable when there are many
-parameters. The previous example would look as follows using a declaring
-body constructor:
+because it must be guaranteed that the primary constructor is executed on
+every newly created instance of this class.
```
lrhn: Why do we want to guarantee that?

(Because of the scope of instance variable initializers. Or "for reasons to be divulged later". Or something.)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -53,38 +47,26 @@ class Point {
...
 
-```dart
-// A declaration with the same meaning, using a declaring body constructor.
-class Point {
-  this(var int x, var int y);
-}
-```
+In particular, every generative constructor in a declaration that has a
+primary constructor must be redirecting, and it must invoke the primary
+constructor (directly or indirectly). This can be seen as a motivation for
```
lrhn: (I'd drop the parenthese here.)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -131,24 +109,20 @@ Similarly, with this feature a regular (non-declaring) formal parameter can
...
-There is no way to indicate that the instance variable declarations should
-have the modifiers `late` or `external` (because formal parameters cannot
-have those modifiers). This omission is not seen as a problem in this
-proposal: They can be declared using the same syntax as today, and
-initialization, if any, can be done in a constructor body.
+There is no way to indicate that the implicitly induced instance variable
+declarations should have the modifiers `late` or `external`. This omission
+is not seen as a problem in this proposal: They can be declared using the
+same syntax as today, and initialization, if any, can be done in a
+constructor body.
```
lrhn: (Also: external variables cannot be initialized by a constructor, and a field parameter initialize their variable, and since a field parameter in the primary constructor is guaranteed to initialize its variable, being `late` has no effect.)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -320,17 +255,9 @@ class const D<TypeVariable extends Bound>.named(
   var int x, [
   var int y = 0,
```
lrhn: Should be final to match the class above. (Doesn't match the previous examples because this constructor is `const`.)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -518,46 +423,18 @@ class E({
...
-cannot be other non-redirecting generative constructors when there is a
-primary constructor, but in the two other versions we could add another
-non-redirecting generative constructor which could initialize `w` with some
-other value, in which case we must also initialize `w` as shown in the
-three cases.
-
-Moreover, we may get rid of all those occurrences of `required` in the
+declaration itself, whereas the other version need to use an element in the
+initializer list of the constructor to initialize `z`. This is necessary
+because `y` isn't in scope. Moreover, there cannot be other non-redirecting
```
lrhn: ... in scope in the initializer expression in  the non-primary-constructor class.

(Or something to say where "y" is not in scope.)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -518,46 +423,18 @@ class E({
...
+generative constructors when there is a primary constructor, but in the
+other version we could add another non-redirecting generative constructor
```
lrhn: Do we need a name for "non-primary-constructor-class"? Because "other version" is very vague and slightly confusing.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -518,46 +423,18 @@ class E({
...
+which could initialize `w` with some other value, in which case we must
+also initialize `w` as shown.
+
+As an aside, we may get rid of all those occurrences of `required` in the
```
lrhn: Don't put this in the feature-specification proper, there is no need to put anything here which is *not* part of the feature. Wasteful at best, confusing at worst. 

Put it in an appendix like "possible future feature improvements".

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -568,8 +445,8 @@ separate proposal, [here][inferred-required] or [here][simpler-parameters].
 
 ### Syntax
 
-The grammar is modified as follows. Note that the changes include support
-for extension type declarations, because they're intended to use declaring
+The grammar is modified as follows. Note that the changes include rules
+about extension type declarations because they're using primary
```
lrhn: Don't like "about" here at all. I don't know what it means.

"rules for" would be OK. "Grammar rules for" would be better.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -608,22 +485,16 @@ constructors as well.
...
+   | <constructorHead> <formalParameterList>; // New form.
 
 <constantConstructorSignature> ::= // Modified rule.
-     'const' <constructorName> <formalParameterList> // Old form.
-   | 'const' <constructorHead> <formalParameterList> // New form.
-   | <declaringConstantConstructorSignature>;
+     'const' <constructorSignature>;
 
 <constructorName> ::= // Modified rule.
      <typeIdentifier> ('.' identifierOrNew)?;
```
lrhn: `<...>` around `identifierOrNew`, ditto just below.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -608,22 +485,16 @@ constructors as well.
...
-     'this' <identifier>? <declaringParameterList>?;
+   | <constructorHead> <formalParameterList>; // New form.
 
 <constantConstructorSignature> ::= // Modified rule.
-     'const' <constructorName> <formalParameterList> // Old form.
-   | 'const' <constructorHead> <formalParameterList> // New form.
-   | <declaringConstantConstructorSignature>;
+     'const' <constructorSignature>;
 
 <constructorName> ::= // Modified rule.
```
lrhn: (Modified relative to what? I can see the current `dartLangSpec.tex` just says `<typeIdentifier> ('.' <identifier>?)`, but that isn't the current language that this feature is modifying.)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -636,19 +507,24 @@ constructors as well.
...
-   | 'CONST'? <factoryConstructorHead> <formalParameterList> '='
-     <constructorDesignation>; // New form.
+     <factoryConstructorSignature> '=' <constructorDesignation>;
 
-<constantConstructorSignature> ::= // Modified rule.
-   : 'const' <constructorName> <formalParameterList> // Old form.
-   | 'const' <constructorHead> <formalParameterList> // New form.
-   | <declaringConstantConstructorSignature>;
+<primaryConstructorBodySignature> ::= // New rule.
+     'this' initializers?;
```
lrhn: `<initializers>?`

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -702,91 +578,59 @@ constructors as well.
...
-`<declaringParameterList>`, because that's an error)*.
+A _primary constructor_ declaration consists of a `<primaryConstructor>` in
+the declaration header plus optionally a member declaration in the body
+that starts with a `<primaryConstructorBodySignature>`.
 
-A class or extension type declaration whose class body is `;` is treated as
-a declaration whose body is `{}`.
+A class, mixin class, or extension type declaration whose class body is `;`
+is treated as the corresponding declaration whose body is `{}` and
+otherwise the same.
```
lrhn: (Do we need to say "which is not a `<mixinApplicationClass>` declaration", or say "whose `<classBody>` is `;`", so people don't start thinking `class C = S with M;` is the same as `class C = S with M {}` ... which isn't even valid.)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -702,91 +578,59 @@ constructors as well.
...
 *Another special exception is introduced with factory constructors in order
 to avoid breaking existing code:*
 
-A factory constructor declaration of the form `factory C(...` where `C`
-is the name of the enclosing class, mixin class, enum, or extension type is
-treated as if `C` had been omitted.
+A factory constructor declaration of the form `factory C(...` optionally
+starting with zero or more of the modifiers `const`, `augment`, or
+`external` where `C` is the name of the enclosing class, mixin class, enum,
+or extension type is treated as if `C` had been omitted.
```
lrhn: (Comma before "is"?)

lrhn: "is treated as if `C` had been omitted" ... I'd prefer to _directly_ say
"declares a constructor named `C`, rather than `C.C`."

No need to treat it like anything else. It does what it does, directly.

Or say that it is treated as if it was parsed using the old syntax.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -702,91 +578,59 @@ constructors as well.
...
-to a body constructor (a regular one, not declaring) and zero or more
-instance variable declarations. This implies that there is a class body
-when there is a declaring constructor. We do not wish to define declaring
-constructors such that the absence or presence of a declaring constructor
-can change the length of the superclass chain, and hence `class C;` has a
-class body just like `class C(int i);` and just like `class C extends
-Object {}`, and all three of them have `Object` as their direct
-superclass.*
+contains a member declaration that starts with a
+`<primaryConstructorBodySignature>`.
```
lrhn, line 617: (I think this is fine, if it's understood correctly: The derivation of the body member contains a derivation of `<primaryConstructorBodySignature>`.
Purely textually, `this` is a production of `<primaryConstructorBodySignature>`, and `this(int x);` starts with `this`. That's not what "starts with" means. Do we need to be pedantic here?)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -702,91 +578,59 @@ constructors as well.
...
+
+*It is an error to have the body part of a primary constructor in the  class
```
lrhn: (Double space before "class")

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -795,32 +639,29 @@ The name of a primary constructor of the form
...
-cannot be a `<mixinApplicationClass>`, because that kind of declaration
-does not support declaring constructors, that is a syntax error)*. This
-declaration is treated as a class, enum, respectively extension type
-declaration without a declaring header constructor which is obtained as
-described in the following. This determines the dynamic semantics of a
-declaring header constructor, and similarly for a declaring body
-constructor.
+A compile-time error occurs if a class, mixin class, enum, or extension
+type has a primary constructor whose name is also the name of a constructor
+declared in the body.
```
lrhn: (Why is this not just a consequence of those two constructor declarations introducing constructors with the same name?
Do we say that ti's an error if the primary constructor's name identifier is the same as the base name of static mbmer?)

accepted/future-releases/primary-constructors/feature-specification.md
lrhn, line 667: (Are we sure there are such rules, and that they are not only defined for classes?)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -832,15 +673,15 @@ expressions. This means that we can allow those other non-redirecting
 generative constructors to coexist with a primary constructor.*
 
 A compile-time error occurs if the name of the primary constructor is the
-same as the name of a constructor (declaring or not) in the body.
+same as the name of a constructor in the body.
```
lrhn: Repeats line 642

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -832,15 +673,15 @@ expressions. This means that we can allow those other non-redirecting
...
 
-*Moreover, it is an error if two constructor declarations in the body,
-declaring or otherwise, have the same name. This is just restating a
-compile-time error that we already have.*
+*Moreover, it is an error if two constructor declarations in the body have
+the same name. This is just restating a compile-time error that we already
+have.*
```
lrhn: And the primary constructor should also just be introducing a constructor with a name, and use the same rules as every other constructor declaration for avoiding conflicts.

Using special rules depending on the syntax being used is almost a guarantee to mis a case somewhere,

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -832,15 +673,15 @@ expressions. This means that we can allow those other non-redirecting
...
 
-The declaring parameter list of the declaring header constructor introduces
-a new scope, the _primary initializer scope_, whose enclosing scope is the
-body scope of _D_. Every primary parameter is entered into this scope.
+The declaring parameter list of the primary constructor introduces a new
+scope, the _primary initializer scope_, whose enclosing scope is the body
+scope of _D_. Every primary parameter is entered into this scope.
```
lrhn: What is a "primary parameter"?
(Is it just a parameter of the primary constructor? If so, say that.)

Is "entered into ... scope" a defined concept? I don't recognize the phrasing.
(But phrasing around scopes hasn't always been very consistent.)

Do I understand this correctly:
* The _primary parameter scope_ is the parameter scope of the primary constructor, and the scope for the body.
* The _primary initializer scope_ is the scope introduced for the initializer list, which also contains (final) bindings for the values of initializing formals and super parameters, and now it's also used for field initializers and contains the argument value of field parameters.

(The initializer scope could have the parameter scope as enclosing scope, and only add the extra parameters. Then nobody would start wondering whether assigning to a non-final parameter in the initializer list is visible in the body. It is, the two scopes contain the same *variables*, not just the same values.)

accepted/future-releases/primary-constructors/feature-specification.md
lrhn: ... `p` instead has declared type `Object?`.

(Otherwise we just gave `p` two different declared types.)

accepted/future-releases/primary-constructors/feature-specification.md
munificent: Maybe instead of "Some elements ...":

> If the primary constructor also needs an initializer list or body, those can be specified inside the class body.

accepted/future-releases/primary-constructors/feature-specification.md
munificent: Or even "This requirement must be upheld because it must be guaranteed" -> "This ensures"

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -518,46 +423,18 @@ class E({
...
-in the initializer list of the constructor to initialize `z`. This is
-necessary because `y` isn't in scope in those two cases. Moreover, there
-cannot be other non-redirecting generative constructors when there is a
-primary constructor, but in the two other versions we could add another
-non-redirecting generative constructor which could initialize `w` with some
-other value, in which case we must also initialize `w` as shown in the
-three cases.
-
-Moreover, we may get rid of all those occurrences of `required` in the
+declaration itself, whereas the other version need to use an element in the
```
munificent: "version need" -> "version needs".

accepted/future-releases/primary-constructors/feature-specification.md
munificent: Maybe just "other version" -> "version without a primary constructor".

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -593,7 +470,7 @@ constructors as well.
    | ';';
 
 <extensionTypeDeclaration> ::= // Modified rule.
-     'extension' 'type' <classNameMaybePrimary> <interfaces>?
+     'extension' 'type' <primaryConstructor> <interfaces>?
```
munificent, line 468: Given that extension types *don't* use `final` for the representation variable and we don't gain any more benefit from allowing the dead in-body declaring form... does it make sense to mention extension types in this proposal at all?

It seems to me that they are their own thing with their own syntax which happens to also be in the class header but is otherwise unrelated.

accepted/future-releases/primary-constructors/feature-specification.md
munificent: Alternatively, get rid of all of the `<` and `>` on the non-terminals. Since terminals are always quoted, it's unambiguous to use bare identifiers for non-terminals.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -702,91 +578,59 @@ constructors as well.
...
-a declaration whose body is `{}`.
+A class, mixin class, or extension type declaration whose class body is `;`
+is treated as the corresponding declaration whose body is `{}` and
+otherwise the same.
 
 The grammar is ambiguous with regard to the keyword `factory`.  *For
 example, `factory() => C();` could be a method named `factory` with an
-implicitly inferred return type, or it could be a factory constructor.*
+implicitly inferred return type, or it could be a factory constructor whose
+name is the name of the enclosing class.*
```
munificent, line 588: I find this really really confusing. I understand that the spec says a constructor's name includes the class name, but I think the spec is the only place where people use that terminology. I think it's much clearer to say that a constructor may be named or unnamed. Then here I would say an "unnamed factory constructor".

I'd love it if the spec reflected that notion too. It's hard to imagine any human deriving value from this line of the spec:

> It is a compile-time error if the name of a constructor is not a constructor name.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -702,91 +578,59 @@ constructors as well.
...
 
 This ambiguity is resolved as follows: When a Dart parser expects to parse
-a `<memberDeclaration>`, and the first token is `factory`, it proceeds to
-parse the following input as a factory constructor.
+a `<memberDeclaration>`, and the beginning of the declaration is `factory`
+or one or more of the modifiers `const`, `augment`, or `external` followed
+by `factory`, it proceeds to parse the following input as a factory
+constructor.
+
+*This is similar to how a statement starting with `switch` or `{` are
```
munificent: "or `{` are" -> "or `{` is".

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -795,32 +639,29 @@ The name of a primary constructor of the form
...
+generative constructor invocation will invoke the primary constructor,
+either directly or via a series of generative redirecting constructors.
+This is required in order to allow non-late instance variable initializers
+to access the parameters.*
 
-If _D_ is an extension type, it is a compile-time error if _D_ does not
-contain a declaring constructor that has exactly one declaring parameter
-which is `final`.
+If _D_ is an extension type, it is a compile-time error if the primary
+constructor that _D_ contains does not have exactly one parameter.
```
munificent, line 662: If we don't treat extension types as having primary constructors, then this can be removed.

accepted/future-releases/primary-constructors/feature-specification.md
munificent: I'm guessing this was a find/replace error from "declaring" -> "primary" and it should read "declaring parameter". :)

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Much better, thanks!

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Done.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Done.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Done.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Good point about `late`!

I don't quite agree about `external`, though. A constructor can certainly do things like `C() { myExternalInstanceVariable = 42; }` which I'd include as "being initialized by a constructor". I don't think it would create any particular difficulties to allow `external` as a modifier on a declaring parameter, which would allow the external instance variable to be initialized earlier.

We just didn't care, or we didn't like the syntax. ;-)

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Oops, thanks!

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Done.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Done.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Removed. It used to be part of the proposal, and I'd be really happy to have it, but that is definitely not happening at this time.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Done.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Done!

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: I just fixed this occurrence. I think `<...>` is such a well-known notation for non-terminals that it's probably easier on the eyes for most readers.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Done.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Good catch, that's marked as modified by mistake, removed the comment.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Good catch, done!

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: I changed this paragraph to have several sentences, each of which is considerably easier to read.

eernstg: Done!

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: I don't think we need to be pedantic. We do this a lot. In general, "let `e` be an expression" already relies on the parsing process to have yielded something like an AST where `e` gave rise to a subtree whose top node is `Expression`. We basically never bother to say those things, but otherwise `e` could be `x` in `var x = 1;`, even though that occurrence of `x` is certainly not an expression.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Thanks! Found 5 of those.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Good catch! Added the rule about static members.

The error needs to be specified explicitly because the primary constructor name does not occur in the class/mixin/etc. body (so the existing rules don't apply), and we generally don't want to catch errors by performing a program transformation and then detecting that the resulting program has an error.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Thanks, fixed!

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Removed this, it's not necessary.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: "Primary parameter" was an imprecise phrase denoting the parameters of the primary constructor. Spelled it out more precisely now.

> entered into ... scope

Good catch again! The standard phrase in the language specification is that a binding is 'introduced' into a scope. Changed the wording to use that phrase.

> Do I understand this correctly:

Yes. The 'primary parameter scope' corresponds to the formal parameter scope of a function (any other kind than a primary constructor), and the 'primary initializer scope' corresponds to the formal parameter initializer scope of a generative constructor.

> The initializer scope could have the parameter scope as enclosing scope, and only add the extra parameters

We can consider that separately (write an issue!), I think it's best to use the same approach in both cases for now.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: 'instead' does not hurt, added that.

I don't think there's an overlap, though, because `p` has type `T` when `T` is not `Null`, and `p` has type `Object?` when `T` is `Null`.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Done!

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: We do have to mention them because it is a special exception that an extension type can have multiple generative non-redirecting constructors. That's again needed in order to have multiple arguments in a constructor, and it would be an unnecessary breaking change to take that away.

eernstg: Also, the syntax for the primary constructor is the same (all the elements of the primary constructor can be used,  including the modifier `final` on the parameter, the ability to make the parameter named and/or optional, possibly with a default value; and we can have a constructor body and assertions). I don't think we should make a bunch of exceptions for extension types, just so they _can't_ to a number of things that other primary constructors can do.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: I've seen that again and again recently. Everybody else seems to think that it's OK for a declaration to have the empty name. I think that's similar to a division by zero, and I'm sure there will be cases where we get into absurd discussions about whether or not the empty name occurs at a specific location in the program. ;-)

I'll try to think about ways to allow a different terminology, but right now I'll stick to the style which has always been used in the language specification (and, I believe, in most feature specifications as well.)

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Thx!

PR #4575, opened 2025-11-25
Specify that non-plain primary parameters are immutable

eernstg:
This PR specifies that it is a compile-time error to assign a value to a formal parameter declared by a primary constructor in the initializer list of the body part of the primary constructor, as well as in the initializing expression of a non-late instance variable declaration.

```dart
class A(int x) {
  final int y = x++; // Error.
  this: assert((x = x + 10).isEven) { // Error.
    x = y / 2; // OK.
  }
}
```

### DIFF
--- a/accepted/future-releases/primary-constructors/feature-specification.md
+++ b/accepted/future-releases/primary-constructors/feature-specification.md
@@ -194,9 +194,9 @@ In the case where the declaration is an `extension type`, the modifier
 `final` on the representation variable can be specified or omitted. It is
 an error to specify the modifier `var` on the representation variable.
 
-An extension type declaration is specified to use a primary constructor (it
-is not supported to declare the representation variable using a normal
-instance variable declaration):
+An extension type declaration must have a primary constructor and its
+single parameter is always declaring. The representation variable cannot be
+declared using a normal instance variable declaration:
 
 ```dart
 // Using a primary constructor.
@@ -222,7 +222,7 @@ We can omit the type of an optional parameter with a default value,
 in which case the type is inferred from the default value:
 
 ```dart
-// Infer the declared type from the default value.
+// Infers the declared type from the default value.
 class Point(var int x, [var y = 0]);
 ```
 
@@ -697,10 +697,10 @@ latter is the current scope for the initializing expressions of all
 non-late instance variable declarations, in addition to the initializer
 list of the body part of the constructor.*
 
-*The point is that the body part of the primary constructor should have
-access to the "regular" parameters, but it should have access to the
-instance variables rather than the declaring or initializing parameters
-with the same names. For example:*
+*The point is that the function body of the body part of the primary
+constructor should have access to the "regular" parameters, but it should
+have access to the instance variables rather than the declaring or
+initializing parameters with the same names. For example:*
 
 ```dart
 class C(var String x) {
@@ -752,13 +752,15 @@ in the initializing expression of a non-late instance variable, or in the
 initializer list of the body part of a primary constructor.
 
 *This includes expressions like `p++` where the assignment is implicit.
-The rule only applies for non-late variables because the primary parameters
-are not in scope in the initializing expression of a late variable.*
+The rule does not apply to late instance variables or (late or non-late)
+static variables. The primary constructor parameters are not in scope for
+initializer expressions of those variables.*
 
 Consider a class with a primary constructor that also has a body part with
 an initializer list. A compile-time error occurs if an instance variable
 declaration has an initializing expression, and it is also initialized by
-an element in the initializer list of the body port.
+an element in the initializer list of the body part, or by an initializing
+formal parameter of the primary constructor.
 
 *This is already an error when the instance variable is final, but no such
 error is raised when the instance variable is mutable and the initializer
@@ -771,12 +773,6 @@ useful.*
 The following errors apply to formal parameters of a primary constructor.
 Let _p_ be a formal parameter of a primary constructor in a class, mixin
 class, enum, or extension type declaration _D_ named `C`:
-
-A compile-time error occurs if _p_ contains a term of the form `this.v`, or
-`super.v` where `v` is an identifier, and _p_ has the modifier
-`covariant`. *For example, `required covariant int this.v` is an error. The
-reason for this error is that the modifier `covariant` must be specified on
-the declaration of `v` which is known to exist, not on the parameter.*
 
 A compile-time error occurs if _p_ has the modifier `covariant`, but
 not `var`. *This parameter does not induce a setter.*
@@ -833,8 +829,10 @@ specifying the current scope explicitly as the body scope, in spite of the
 fact that the primary constructor is actually placed outside the braces
 that delimit the class body.*
 
-Next, _k2_ has the modifier `const` iff the keyword `const` occurs just
-before the name of _D_, or _D_ is an `enum` declaration.
+Next, _k2_ has the modifier `const` if and only if the keyword `const`
+occurs just before the name of _D_ or _D_ is an `enum` declaration. In any
+case, such an occurrence of `const` in the header of _D_ is omitted in
+_D2_.
 
 Consider the case where _k_ is a primary constructor. If the name `C` in
 _D_ and the type parameter list, if any, is followed by `.id` where `id` is
@@ -855,27 +853,32 @@ positional or named parameter remains optional; if it has a default value
 `d` in _L_ then it has the default value `d` in _L2_ as well.
 
 - An initializing formal parameter *(e.g., `T this.x`)* is copied from _L_
-  to _L2_, along with the default value, if any, and is otherwise unchanged.
-- A super parameter is copied from _L_ to _L2_ along with the default
-  value, if any, and is otherwise unchanged.
+  to _L2_, with no changes.
+- A super parameter is copied from _L_ to _L2_ any, with no changes.
 - A formal parameter which is not covered by the previous two cases and
   which does not have the modifier `var` or the modifier `final` is copied
   unchanged from _L_ to _L2_ *(this is a plain, non-declaring parameter)*.
-- Otherwise, a formal parameter (named or positional) of the form `var T p`
-  or `final T p` where `T` is a type and `p` is an identifier is replaced
-  in _L2_ by `this.p`, along with its default value, if any. If the
-  parameter has the modifier `var` and _D_ is an extension type declaration
-  then a compile-time error occurs. Otherwise, a semantic instance variable
-  declaration corresponding to the syntax `T p;` or `final T p;` is added
-  to _D2_. It includes the modifier `final` if the parameter in _L_ has the
-  modifier `final` and _D_ is not an `extension type` decaration; if _D_ is
-  an `extension type` declaration then the name of `p` specifies the name
-  of the representation variable. In all cases, if `p` has the modifier
-  `covariant` then this modifier is removed from the parameter in _L2_, and
-  it is added to the instance variable declaration named `p`.
+- Otherwise, it is a declaring parameter. A formal parameter (named or
+  positional) of the form `var T p` or `final T p` where `T` is a type and
+  `p` is an identifier is replaced in _L2_ by `this.p`, along with its
+  default value, if any. The same is done in the case where the formal
+  parameter has the form `var p` or `final p`, and `T` is the declared type
+  of `p` which was obtained by inference. If the parameter has the modifier
+  `var` and _D_ is an extension type declaration then a compile-time error
+  occurs. Otherwise, if _D_ is not an extension type declaration, a
+  semantic instance variable declaration corresponding to the syntax `T p;`
+  or `final T p;` is added to _D2_. It includes the modifier `final` if and
+  only if the parameter in _L_ has the modifier `final` and _D_ is not an
+  `extension type` decaration.  Otherwise, if _D_ is an `extension type`
+  declaration then the name of `p` specifies the name of the representation
+  variable. In all cases, if `p` has the modifier `covariant` then this
+  modifier is removed from the parameter in _L2_, and it is added to the
+  instance variable declaration named `p`.
 
-If there is an initializer list following the formal parameter list _L_
-then _k2_ has an initializer list with the same elements in the same order.
+If there is a primary constructor body part that contains an initializer
+list then _k2_ has an initializer list with the same elements in the same
+order. If that body part has a function body then _k2_ has the same
+function body.
 
 Finally, _k2_ is added to _D2_, and _D_ is replaced by _D2_.
 


lrhn:
Do we want this?
It's a loss of expressiveness compared to existing constructors.

Example of something you *could* do using side-effects:
```dart
class StreamWrapper<T>([StreamController<T>? existingController]) {
   final Stream<T> stream;
   final StreamController<T> controller;
   init: 
      controller = existingController ??= StreamController(),
      stream = controller.stream;
}
```

Initializer list entries are (probably) run after all initializing expressions, so asssignments only affect themselves and the body block.
I feel like they are sequential enough in nature that it should be OK to assume an order.

It may be a little confusing if we have:
```dart
class C(int x) {
   final x1 = x;
   this: x3 = ++x;
   final x2 = x;
   final int x3;
}
```
because the three references to `x` are not in order.

Then don't do that. Put the `this:` after the instance variables, and everything is fine.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -745,6 +747,12 @@ main() {
 }
 ```
 
+It is a compile-time error if an assignment to a primary parameter occurs
+in the initializing expression of a non-late instance variable, or in the
```
lrhn, line 751: (And late instance variables do not have access to those parameters at all, so definitely also an error if they try to assign to it. Right?)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -745,6 +747,12 @@ main() {
...
+initializer list of the body part of a primary constructor.
+
+*This includes expressions like `p++` where the assignment is implicit.*
+
```
munificent, line 758: Do we allow assigning to primary constructor parameters in the `this` block body? I could see the answer being "no" to be consistent with initializers and the initializer list. I could also see the answer being "yes" to be consistent with assigning other parameters in function bodies.

Either way, we should probably spell it out explicitly.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: True, the primary constructor parameters are not in scope for the initializing expression of a late instance variable, so they can't assign a new value to them either. I added a sentence in the commentary paragraph that follows in order to make this distinction explicit.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Following the decision at the language team meeting on Nov 19, I only introduced errors for assignments in the initializer list and the initializing expressions of non-late instance variables. So assignments are allowed in the constructor body. I added commentary to mention this.

eernstg:
@lrhn wrote:
> Do we want this?

Oh, you're right, we did _not_ include non-declaring, non-initializing-formal, non-super parameters (we could call them _plain_), they can still be mutated also in initialization code.

I adjusted the wording to match the decision.

lrhn:
We also disallowed assigning to plain primary constructor parameters in instance variable initializer expressions, but not in initializer list expressions. So the text is still half right for those.

eernstg:
> We also disallowed assigning to plain primary constructor parameters in instance variable initializer expressions

[Edit: Corrected this paragraph] The decisions made at the meeting Nov 19 do not make this distinction, they disallow mutation of primary parameters (of any kind) in non-late instance variable initializing expressions _and_ in the initializer list of the primary constructor body part.

The rule you're proposing does make sense, though, so we should discuss it.

The decision at the language meeting was rather strict. I think it would make sense to think about it one more time.

One thing to note is that it would force a rewrite on declarations like the following:

```dart
class A([int? i]) {
  final int x = i ??= 42;
  final int y = i + 1;
}
```

This declaration would be an error because we're (possibly) assigning to `i` in the initializing expression of the non-late instance variable `x`. We can still do exactly the same thing, but we will have to use a somewhat more verbose form:

```dart
class A([int? i]) {
  final int x;
  final int y;
  this: x = i ??= 42, y = i + 1;
}
```

This _might_ improve the overall readability (look again, I'm not convinced). We might at least want the error message on the initialization of `x` to indicate that _all subsequent_ initializing expressions of non-late instance variables must be moved to the initializer list of the primary constructor body part in order to maintain the same evaluation order.

If a developer is navigating these considerations already then I tend to think that we might as well allow the first form above. We could have some kind of UI support ("hover on the initializing expression") to show the evaluation order of all the initializing expressions of a given class with a primary constructor.

In any case, developers who are using side-effecting expressions in initializing expressions already need to be aware of the exact ordering of the evaluations. We should probably recommend an ordering which is aligned with the evaluation order: "Put the primary constructor body part after the instance variable declarations", or, at least, after the ones that are non-late and have an initializing expression.

@dart-lang/language-team, WDYT? Should we make the first example above an error, and hence enforce something like the second form? I tend to prefer that we allow both forms because the improvement in readability by outlawing the first one isn't very convincing to me.

[Edit: The PR has now been landed, and it uses the strictest ruleset -- primary parameters cannot be mutated in initialization code. So both of the above examples are now errors.]

eernstg:
The update now also specifies an error for a double initialization of a non-final instance variable (in the initializing expression of the variable itself, as well as in the initializer list of the body part of a primary constructor ).

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -191,10 +191,12 @@ that the instance variable declaration which is induced by this declaring
...
 In the case where the declaration is an `extension type`, the modifier
-`final` on the representation variable can be specified or omitted. Note
-that an extension type declaration is specified to use a primary
-constructor (it is not supported to declare the representation variable
-using a normal instance variable declaration):
+`final` on the representation variable can be specified or omitted. It is
+an error to specify the modifier `var` on the representation variable.
+
+An extension type declaration is specified to use a primary constructor (it
+is not supported to declare the representation variable using a normal
```
lrhn: (The "it's not verb to verb" phrasing feels a little like a Danglisism to me. Could it be:
> An extension type declaration must have a primary constructor and its one parameter is always declaring,
> the representation variable cannot be declared
using a normal instance variable declaration.

)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -220,7 +222,7 @@ We can omit the type of an optional parameter with a default value,
 in which case the type is inferred from the default value:
 
 ```dart
-// Infer the declared type from default value.
+// Infer the declared type from the default value.
```
lrhn: "Infer" -> "Infers", maybe.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -745,6 +747,27 @@ main() {
 }
 ```
 
+A compile-time error occurs if an assignment to a primary parameter occurs
+in the initializing expression of a non-late instance variable, or in the
+initializer list of the body part of a primary constructor.
```
lrhn, line 752: _"body part" feels a little ambiguous. The "body part of a primary constructor" is not the same as "the body of the primary constructor". The body part can have a body part! It's bodies all the way down!_
_Could we consider calling it something else, like "the `this` part" or just "the in-body part"._

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -745,6 +747,27 @@ main() {
...
+
+*This includes expressions like `p++` where the assignment is implicit.
+The rule only applies for non-late variables because the primary parameters
```
lrhn: Consider:

-> The rule does not apply to late variables or static variables. The primary constructor parameters are not in scope for initializer expressions of those variables at all.

(More direct and mentions static variables, redundantly since the above says "instance variables", but the above also says "non-late", so this commentary might as well be just as complete.)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -745,6 +747,27 @@ main() {
...
+are not in scope in the initializing expression of a late variable.*
+
+Consider a class with a primary constructor that also has a body part with
+an initializer list. A compile-time error occurs if an instance variable
+declaration has an initializing expression, and it is also initialized by
+an element in the initializer list of the body port.
```
lrhn: , or by an initializing formal parameter of the parameter list.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -745,6 +747,27 @@ main() {
...
+
+*This is already an error when the instance variable is final, but no such
+error is raised when the instance variable is mutable and the initializer
+list is part of a non-primary constructor. However, with a primary
+constructor this situation will always cause the value of the initializing
+expression in the variable declaration to be overwritten by the value in
+the initializer list, which makes the situation more confusing than
+useful.*
```
lrhn, line 771: _(And also because order can be confusing:_
```dart
class C(int x) {
  this: v = x + 42;
  int v = 37;
}
```
_which looks like `v = 37` might be assigned after `v  = x + 42`. It wouldn't be, but order of evaluation being confusing is the primary reason to not allow assignment to plain parameter variables, which at least makes that order-independent.)_

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -782,13 +805,13 @@ Let `p` be a formal parameter in _k_ which has the modifier `var` or the
...
-  parameter `p` has declared type `T`. *In other words, an instance
-  variable introduced by a declaring parameter is subject to override
-  inference, just like an explicitly declared instance variable.*
+- if the combined member signature for a getter with the same name as `p`
+  from the superinterfaces of _D_ exists and has return type `T`, the
+  parameter `p` has declared type `T`. If no such getter exists, but a
+  setter with the same basename exists, with a formal parameter whose type
+  is `T`, the parameter `p` has declared type `T`. *In other words, an
+  instance variable introduced by a declaring parameter is subject to
+  override inference, just like an explicitly declared instance variable.*
```
lrhn, line 810: Is it worth commenting that this works *only* for declaring parameters? One could think that `Foo([final x = 42, y = 37])` would infer `int` for `y` as well.

_We could do that, but then it would not just be for constructor parameters, so it's a larger change. For now it's only for a declaring parameter, because it **looks like a variable declarations with an initializer expression**, and those do get inference._

accepted/future-releases/primary-constructors/feature-specification.md
lrhn: (I'd drop the comma before "or", it doesn't feel like what comes after it binds to the `iff`.)

lrhn: (Consider "iff" -> "if and only if", can't assume all readers are mathematicians.)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -840,15 +863,16 @@ positional or named parameter remains optional; if it has a default value
...
-  in _L2_ by `this.p`, along with its default value, if any. Next, a
-  semantic instance variable declaration corresponding to the syntax `T p;`
-  or `final T p;` is added to _D2_. It includes the modifier `final` if the
-  parameter in _L_ has the modifier `final` and _D_ is not an `extension
-  type` decaration; if _D_ is an `extension type` declaration then the name
-  of `p` specifies the name of the representation variable. In all cases, if
-  `p` has the modifier `covariant` then this modifier is removed from the
-  parameter in _L2_, and it is added to the instance variable declaration
-  named `p`.
+  in _L2_ by `this.p`, along with its default value, if any. If the
```
lrhn: Also doesn't need to mention default value. If it wants to be complete, then it should also mention "any required modifier". _Not wrong, just feels inconsistent._

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -840,15 +863,16 @@ positional or named parameter remains optional; if it has a default value
...
+  parameter has the modifier `var` and _D_ is an extension type declaration
+  then a compile-time error occurs. Otherwise, a semantic instance variable
+  declaration corresponding to the syntax `T p;` or `final T p;` is added
+  to _D2_. It includes the modifier `final` if the parameter in _L_ has the
+  modifier `final` and _D_ is not an `extension type` decaration; if _D_ is
```
lrhn: Move the "and _D_ is not an `extension type declaration` to the prior sentence. The entire variable declaration is not introduced if it's an extension type declaration, not just the `final`.

accepted/future-releases/primary-constructors/feature-specification.md
lrhn: This feels a little inconsistent.
The prior text sounds like it starts with _L2_ being a copy of _L_, then *modifies* it to something else.
This clause states how it creates _L2_ from _L_.

It doesn't have to say "along with the default value", the text in line 854 already said that default values are retained.

Maybe the word "processed" above is just too vague.
Consider being very explicit about whether this is *modifying* or *building* _L2_.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -840,15 +863,16 @@ positional or named parameter remains optional; if it has a default value
   unchanged from _L_ to _L2_ *(this is a plain, non-declaring parameter)*.
 - Otherwise, a formal parameter (named or positional) of the form `var T p`
   or `final T p` where `T` is a type and `p` is an identifier is replaced
```
lrhn: Nit: If the parameter is var p or final p, where its declared type is inherited, inferred from default value or defaults to Object?, then this text doesn't cover that situation.
So consider:

Otherwise it's a declaring parameter.
Let p be the declared name of the parameter and
T its declared type (potentially inferred).
If D is not an extension type declaration,
the declaring parameter's name, type and preceding var, final and or covariant is replaced in L2 by
the initializing formal this.p.
Then a semantic instance variable declaration
is added to D2 with name p and type T.
The variable is final if the declaring parameter
of L had a final modifier, otherwise it's not.
The variable iscovariant if the declaring parameter
had a `covariant` modifier, otherwise it's not.
If D is an extension type declaration,
it's a compile-time error if the declaring parameter of L
has a var modifier. It's always a compile-time error if
a declaring parameter has a covariant modifier and no var,
so a valid extension type primary constructor parameter can never have a covariant modifier.
(Is it a compile-time error if it has no type clause, even if it has a default value to infer it from?)
Otherwise the parameter in L2 is the same as in L1 with the `final` modifier removed.

accepted/future-releases/primary-constructors/feature-specification.md
lrhn, line 778: (Doesn't this make the prior paragraph redundant. An initializing formal or super parameter never has a `var`.)

lrhn, line 795: _(That is a source code transformation, transforming post-feature *code* to pre-feature *code*! 
It's just done in the spec instead of in the compiler, and it pretends to only create the desugared code in order to grab its semantics, and the apply that semantics to the original code. I have no idea where the breaks down, but I'll bet you it does. If we want to have the notion of "a semantic constructor" separate from the syntax, so that two different syntaxes can define equivalent semantic constructors, and we can define type inference on semantic declarations, we should do that. We haven't specified type inference on semantic entities, but on syntactic entities, and we haven't specified it for code that is not part of the program. It doesn't stop being syntactic desugaring just because we say it ain't so.)_

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -782,13 +805,13 @@ Let `p` be a formal parameter in _k_ which has the modifier `var` or the
...
 Consider the situation where `p` has no type annotation:
-- if combined member signature for a getter with the same name as `p` from
-  the superinterfaces of _D_ exists and has return type `T`, the parameter
-  `p` has declared type `T`. If no such getter exists, but a setter with
-  the same basename exists, with a formal parameter whose type is `T`, the
-  parameter `p` has declared type `T`. *In other words, an instance
-  variable introduced by a declaring parameter is subject to override
-  inference, just like an explicitly declared instance variable.*
+- if the combined member signature for a getter with the same name as `p`
+  from the superinterfaces of _D_ exists and has return type `T`, the
```
lrhn, line 805: (Does not apply to extension type constructors.
They can have superinterfaces, but the representation variable is unrelated to those.)

accepted/future-releases/primary-constructors/feature-specification.md
lrhn: > _D2_ omits the `const` modifier before the class name of the class declaration.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -840,15 +863,16 @@ positional or named parameter remains optional; if it has a default value
   unchanged from _L_ to _L2_ *(this is a plain, non-declaring parameter)*.
```
lrhn, line 860: (Even for extension types? I *think* it may be so, because we allow both `Type id` and `final Type id` for the extension type parameter.)

accepted/future-releases/primary-constructors/feature-specification.md
lrhn, line 790: Does not apply to extension type declarations, which *must* have a primary constructor declaration.

Maybe just move extension type primary constructors into its own section entirely, and don't try to hande it the same way as for the other type declarations, inline in the same sections where it's easy to forget a case that is *also* different.
Is there *any* way in which it's not different?

accepted/future-releases/primary-constructors/feature-specification.md
lrhn: (So we remove the `final` if it's an extension type, to be backwards compatible, but still allow optional or named parameters like `extension type Foo({int x  = 0});`. So it's not backwards compatible anyway.

We could desugare that into into:
```dart
 extension type Foo._$fresh(int x) {
   Foo({this.x = 0});
 }
```
but I don't think we should introduce a new constructor. 
The user-declared primary constructor is the primary constructor of the extension type, if that ever matters for anything.)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -840,15 +863,16 @@ positional or named parameter remains optional; if it has a default value
...
-  type` decaration; if _D_ is an `extension type` declaration then the name
-  of `p` specifies the name of the representation variable. In all cases, if
-  `p` has the modifier `covariant` then this modifier is removed from the
-  parameter in _L2_, and it is added to the instance variable declaration
-  named `p`.
+  in _L2_ by `this.p`, along with its default value, if any. If the
+  parameter has the modifier `var` and _D_ is an extension type declaration
+  then a compile-time error occurs. Otherwise, a semantic instance variable
+  declaration corresponding to the syntax `T p;` or `final T p;` is added
+  to _D2_. It includes the modifier `final` if the parameter in _L_ has the
```
lrhn: The "if not extension type" should be applied to adding the variable declaration too. We don't do that for extension types.

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -840,15 +863,16 @@ positional or named parameter remains optional; if it has a default value
...
+  modifier `final` and _D_ is not an `extension type` decaration; if _D_ is
+  an `extension type` declaration then the name of `p` specifies the name
+  of the representation variable. In all cases, if `p` has the modifier
+  `covariant` then this modifier is removed from the parameter in _L2_, and
+  it is added to the instance variable declaration named `p`.
 
 If there is an initializer list following the formal parameter list _L_
```
lrhn: That's not where the initializer is any more. And remember the body block.

So:
> If _D_ contains an primary constructor in-body part, if that in-body part contains an initializer list, then _k2_ has an initializer list with the same elements in the same order, and if that in-body part has a body block, then _k2_ has the same body block.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: True, extension types are different in many ways. I also agree that override inference is unlikely to be a natural mechanism for an extension type. It is not hard to create an example where an extension type uses override inference, but it is completely contrived:

```dart
abstract class Cons {
  Cons(this.x);
  final int x;
  Cons get next;
}

class ConsLink extends Cons {
  final Cons next;
  ConsLink(super.x, this.next);
}

class ConsLoop extends Cons {
  Cons get next => this;
  ConsLoop(super.x);
}

// The type of `next` is `Cons`, based on override inference.
extension type ExtendedCons(next) implements Cons {}

void main() {
  final c0 = ConsLink(3, ConsLoop(4));

  Cons c = c0;
  for (int i = 0; i < 5; ++i) {
    print(c.x);
    c = c.next;
  }
  print('---');

  ExtendedCons ec = .new(c0);
  for (int i = 0; i < 5; ++ i) {
    print(ec.x);
    ec = ExtendedCons(ec.next);
  }
}
```

However, I don't really think it's useful to say "let's prevent override inference with extension types just because I can't come up with a good use case right now". Perhaps there will be a good use case at some point. In the meantime, it doesn't hurt anybody that override inference is never used with a representation variable.

Moving the treatment of extension types to a separate section might be a good idea, but it will also allow for accidental inconsistencies for properties which _can_ be described just once rather than twice.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: ;-)

eernstg: Yes it does, good catch!

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Indeed, thanks!

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Done.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: We don't remove `final`, we specify that `final` is implicitly added if it is not specified. It is backward compatible because existing code will now have a declaring parameter of the form `final T name`. `extension type Foo({int x  = 0});` was previously a syntax error, so it's irrelevant to backward compatibility.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: The fact that an extension type _must_ have a primary constructor is an unnecessary inconsistency. The handling of initialization of the representation variable, the static checking of accesses to the representation variable, in fact everything about the representation variable, is the same as for a final instance variable. It is irrational and error prone to insist that we must specify all these things from scratch. "DRY".

The only thing that differs is that the representation variable is stored in a different location (namely in the slot which is also considered to be the location where the reference to the extension typed entity is stored).

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: It doesn't matter for an extension type because they infer `final` on every parameter that doesn't have `var`, and `var` is an error. In particular, it is also an error for a primary constructor of an extension type to have 2 or more formal parameters, because it is an error to have multiple declaring parameters.

So we don't have to make an exception for an extension type because this rule will never apply to code that doesn't have errors.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Good catch!

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Very nice! I revised the text rather extensively.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: It is more like `copyWith`: _L2_ is identical to _L_ except for the elements which are specified in the following. So there's no dynamic mutation of _L2_ (nor of _L_).

I remove the words about the default value, it is indeed clearer with them.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Removed.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Done.

eernstg: Done.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: It is specified 2 lines earlier that `p` is a declaring parameter. I think that's sufficient.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: The ordering issue is somewhat broader because it's got to do with initialization of different variables. This is about double initialization of a single variable.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Good catch!

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Done.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: We're using 'the body part of the primary constructor' to denote `this <initializers>? (';' | <block>)`, and we're using 'function body' to denote the block.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: Done.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: That's better, done.

Issue #4574, opened 2025-11-25
Allow `<void>[e]` and `<void, void>{e: e}` where `e` has type `void`? Records, too?


eernstg:
See https://github.com/dart-lang/sdk/issues/58456 for background discussions.

The current set of rules about errors reported for expressions of type `void` does not cover the elements of collection literals. For example `<void>[e]` is an error when `e` has static type `void`, and so is `<void, void>{e: e}`.

However, the implementations currently do not report this error. Presumably, said elements are allowed because they are somewhat similar to actual arguments, and it is allowed to pass an actual argument of type `void` to a formal parameter whose declared type is also `void`.

It would be a breaking change to start reporting these errors.

@dart-lang/language-team, WDYT? Should we add void-typed collection elements to the allowlist, or should we start reporting these errors?

lrhn:
Who will we be helping by allowing this? 

I'd rather properly disallow it, so people who are putting bad or meaningless values into collections will stop doing that. And maybe notice that their code makes no sense.

Wdestroier:
For example: I was writing `StreamBuilder<void>`, because I had to make part of the widget tree reactive, but I did not want to use the value, not even `==`, `hashCode` or `toString`. I guess I'm not the first person to think about that, because GitHub found [466](https://github.com/search?q=StreamBuilder%3Cvoid%3E&type=code) files. ~~If any generic widget creates a generic list `<T>[]` under the hood, then the code will work. If this issue is not accepted, then the code wouldn't work, I guess.~~ (Answer below about generics)

lrhn:
If the class is generic, this issue has no impact. 
A type variable is not statically the `void` type.

lrhn:
Decision is to allow the cases listed here, and disallow `[?e]`, `{?e}`, `{?e:_}` and `{_:?e}` where `e` has type `void`.
The latter are also currently allowed, but do not match the allowed use of passing a value along without looking at it, beause it looks at the value to check if it's `null`.

Seems this still requires a change (bug fix) in implementation since the front-end makes `{v: 0}` and `{0: v}` errors, even if it then infers a key/value type of `void`. Analyzer does not.

```dart
// ignore_for_file: unused_local_variable
void main() {
  const void v = null;
  // Implicit types.
  var implicitList = [v];
  var implicitSet = {v};
  var implicitMapKey = {v: 0}..st<E<Map<void, int>>>(); // CFE error.
[...]

PR #4617, opened 2026-01-23
Add rules to the void usage allowlist about elements

eernstg:
See #4574 and dart-lang/sdk#58456 for background discussions.

The language team decided at the meeting Nov 26, 2025 that collection literal elements should be given a treatment with respect to expressions of type `void` which is similar to the one which is given to actual arguments in function invocations. For example:

```dart
void f(void _) {}

final void v = null;

void main() {
  // Actual arguments.
  f(v); // OK, `v` is passed to a formal parameter whose type is `void`.
  print(v); // Error, no exception applies so we can't use the value of `v`.

  // Collection literal elements.
  <void>[v]; // OK, the element type of the list literal is `void`.
  <Object?>[v]; // Error, no exception applies so we can't use the value of `v`.

  // Nested elements will propagate the value if they don't use it.
  <void>[
    v, // OK.
    if (2 > 1) v else v, // OK.
    ?v, // Error, this element uses the value of `v` to test for null.
  ];
}
```

This PR introduces two new entries in the allowlist that makes it a non-error for an expression to occur in specific situations, such that, roughly, occurrences of void expressions in void collection literals is no longer an error.

The behavior of the analyzer and the common front end is already as specified in this CL in many cases, but the handling of null-aware elements in particular needs to be updated.

```dart
// Implemented behavior that differs from the updated spec.

void v = null;

void main() {
  <void>[?v]; // Missing compile-time error.
  <void, void>{?v: v, v: ?v, ?v: ?v}; // Multiple wrong behaviors.
}
```

The 'wrong behaviors' stand for differences among tools: The analyzer does not report any of the 4 compile-time errors; the CFE reports an error at a wrong position followed by two errors which are likely to be reported for a wrong reason. So there will be a need to make some small adjustments to the tools.

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -42,8 +42,8 @@
 % release of this document.
 %
 % Jan 2026
-% - Specify that collection literal elements are treated like formal parameters
-%   with respect to the `void` usage allowlist.
+% - Specify that collection elements are allowed to have type `void` in a
+%   collection literal where the corresponding actual type argument is `void`.
 %
 % Oct 2025
 % - Simplify the member declaration grammar by using a `memberDeclaration` to


eernstg:
@natebosch, we have the language team decision from Nov 26, 2025 that collection literal elements should be allowed to have type `void` (essentially as if `e1` and `e2` in `[e1, e2]` were treated as actual arguments to function invocations, but with the exception that the element can't test for null). Could you take a look at this PR which performs the corresponding spec update?

natebosch (approved):
It's been a while since I've done much in `.tex` so I can't do a review for syntax etc, but from a phrasing standpoint this LGTM (after going back to re-read the meeting notes)

eernstg:
Thanks! LaTeX issues are generally not reviewed (other than by running `make` and looking at the PDF), so that should not be a problem.

PR #4620, opened 2026-01-26
Update shared memory proposal

mraleph:
* Mention that `final` variables and fields can't be late when describing deeply immutable types. 
* Add event loop and ownership APIs to `Isolate`.

### DIFF
--- a/working/333 - shared memory multithreading/shared_native_memory.md
+++ b/working/333 - shared memory multithreading/shared_native_memory.md
@@ -173,14 +173,14 @@ All compile time constants are deeply immutable instances.
 Unmodifiable lists (`List.unmodifiable`) which contain deeply immutable
 instances are deeply immutable.
 
-Closures which capture only `final` variables containing deeply immutable
-instances are deeply immutable.
+Closures which capture only `final`, non-`late` variables containing deeply
+immutable instances are deeply immutable.
 
 Finally, instances of classes annotated with `@pragma('vm:deeply-immutable')`
 are deeply immutable. It is a compile error if classes annotated with this
-pragma contain non-`final` fields. It is an compile time error if static
-type of field within annotated class excludes deeply immutable instances.
-If the static type of a field in a deeply immutable class is not
+pragma contain non-`final` or `late final` fields. It is an compile time error
+if static type of field within annotated class excludes deeply immutable
+instances. If the static type of a field in a deeply immutable class is not
 deeply immutable type - then compiler must insert checks in the constructor to
 guarantee that this field is initialized to a deeply immutable value.
 
@@ -206,7 +206,8 @@ values which are deeply immutable objects.
 * If static type of a field is a super-type for both deeply immutable and
   non-deeply immutable objects then compiler will insert a runtime check
   which ensures that values assigned to such field are deeply immutable.
-* A field or variable annotated with `@pragma('vm:shared')` must be `final`.
+* A field or variable annotated with `@pragma('vm:shared')` must be `final` and
+  non-`late`.
 
 > [!NOTE]
 >
--- a/working/333 - shared memory multithreading/shared_native_memory.md
+++ b/working/333 - shared memory multithreading/shared_native_memory.md
@@ -307,20 +307,102 @@ class Isolate {
   /// to acquire exclusive access to the isolate.
   ///
   /// Throws [StateError] if target isolate is owned by another thread and
-  /// thus can't be entered from a different thread.
+  /// thus can't be entered from a different thread. See
+  /// [markOwnedByCurrentThread] and [isOwnedByCurrentThread].
+  ///
+  /// Throws [ArgumentError] if the target isolate belongs to another
+  /// isolate group.
   ///
   /// Throws [ArgumentError] if [f] is not deeply immutable.
   ///
   /// Throws [StateError] if result returned by [f] is not deeply immutable.
-  R runSync<R>(R Function() f, {Duration? timeout});
+  external R runSync<R>(R Function() f, {Duration? timeout});
+
+  /// Create a new isolate in the current isolate group.
+  ///
+  /// Similar to `Dart_CreateIsolateInGroup` Dart VM C API.
+  ///
+  /// Created isolate is in runnable state, but its event loop is not running.
+  ///
+  /// To start processing isolate's messages:
+  ///
+  /// * start isolate's event loop synchronously on the current thread
+  ///   by calling [Isolate.runEventLoopSync]
+  /// * integrate isolate's event loop with other event loop by registering
+  ///   message callback ([Isolate.onMessage]) and draining pending messages
+  ///   ([Isolate.handleMessage]).
+  external static Isolate fork({String? debugName});
+
+  /// Shutdown target isolate.
+  ///
+  /// This function will block until it acquires exclusive access to the
+  /// target isolate. Isolate can only be entered for synchronous execution
+  /// between turns of its event loop, when no other thread is
+  /// executing code in the target isolate.
+  external void shutdown();
+
+  /// Set current OS thread as owner of the isolate.
+  ///
+  /// Once an isolate is owned by some OS thread it can not be
+  /// entered by any other OS thread. An attempt to acquire
+  /// exclusive access to it from another thread will fail with
+  /// an error.
+  ///
+  /// Equivalent to `Dart_SetCurrentThreadOwnsIsolate` Dart VM C API.
+  ///
+  /// Throws [ArgumentError] if `this` is not `Isolate.current`.
+  ///
+  /// Throws [StateError] if target isolate is already owned by another thread.
+  external void markOwnedByCurrentThread();
+
+  /// Returns `true` if the isolate is owned by the current OS thread.
+  ///
+  /// Equivalent to `Dart_GetCurrentThreadOwnsIsolate` Dart VM C API.
+  external bool get isOwnedByCurrentThread;
+
+  /// Run event loop for the target isolate synchronously on the current thread.
+  ///
+  /// This function will block until it acquires exclusive access to the
+  /// target isolate. Isolate can only be entered for synchronous execution
+  /// between turns of its event loop, when no other thread is
+  /// executing code in the target isolate.
+  ///
+  /// The isolate will be marked as owned by the current thread.
+  ///
+  /// Similar to `Dart_RunLoop` Dart VM C API, but unlike `Dart_RunLoop` this
+  /// function executes isolate's event loop on the current thread instead
+  /// of delegating it into the thread-pool.
+  ///
+  /// Throws [StateError] if target isolate is owned by another thread.
+  external static void runEventLoopSync();
+
+  /// Set message notify callback for the isolate.
+  ///
+  /// Provided callback will be called once for every message added to the
+  /// isolates message queue. Pending messages can be then later be drained
+  /// by calling [Isolate.handleMessage].
+  ///
+  /// Provided [callback] must be deeply immutable and will be called
+  /// on an arbitrary thread and not necessarily within some isolate. See
+  /// [NativeCallable.isolateGroupBound].
+  ///
+  /// IMPORTANT: [Isolate.handleMessage] must *not* be called from the
+  /// `callback`.
+  ///
+  /// Similar to `Dart_SetMessageNotifyCallback` Dart VM C API.
+  external void set onMessage(void Function(Isolate) callback);
+
+  /// Handle a single pending message from isolate's message queue.
+  ///
+  /// This function will block until it acquires exclusive access to the
+  /// target isolate. Isolate can only be entered for synchronous execution
+  /// between turns of its event loop, when no other thread is
+  /// executing code in the target isolate.
+  ///
+  /// Similar to `Dart_HandleMessage` Dart VM C API.
+  external void handleMessage();
 }
 ```
-
-**TODO**: Furthermore we might want to facilitate integration with third-party
-event-loops: e.g. allow to create isolate without scheduling its event loop on
-our own thread pool and provide equivalents of `Dart_SetMessageNotifyCallback`
-and `Dart_HandleMessage`. Though maybe we should not bundle this all together
-into one update.
 
 ### Scoped thread local values
 
--- a/working/333 - shared memory multithreading/shared_native_memory.md
+++ b/working/333 - shared memory multithreading/shared_native_memory.md
@@ -389,7 +389,7 @@ class Isolate {
   ///
   /// Throws an error if target isolate is pinned to another thread or already
   /// has an event loop running.
-  external static void runEventLoopSync();
+  external void runEventLoopSync();
 
   /// Event notify callback for the isolate.
   ///
--- a/working/333 - shared memory multithreading/shared_native_memory.md
+++ b/working/333 - shared memory multithreading/shared_native_memory.md
@@ -26,9 +26,11 @@ Porting this code to Dart using `dart:ffi` is currently impossible, as FFI only
 supports two specific callback types:
 
 - [`NativeCallable.isolateLocal`][native-callable-isolate-local]: native caller
-  must have an exclusive access to an isolate in which callback was created.
-  This type of callback works if Dart calls C and C calls back into Dart
-  synchronously. It also works if caller uses VM C API for entering isolates
+  must have an exclusive access to an isolate in which callback was created - or
+  isolate must be pinned (owned) by a thread which invokes the callback. In the
+  later case FFI trampoline will take care of entering the target isolate even
+  if needed. This type of callback works if Dart calls C and C calls back into
+  Dart synchronously. It also works if caller uses VM C API for entering isolates
   (e.g.`Dart_EnterIsolate`/`Dart_ExitIsolate`).
 - [`NativeCallable.listener`][native-callable-listener]: native caller
   effectively sends a message to the isolate which created the callback and does
@@ -167,8 +169,16 @@ classes are also deeply immutable: `SendPort`, `Capability`, `RegExp`,
 
 All compile time constants are deeply immutable instances.
 
-`TypedData` and `Struct` instances are deeply immutable when backed by native
-(external) memory.
+`TypedData` and `Struct` instances are considered deeply immutable.
+
+> [!IMPORTANT]
+>
+> There is a consideration here that not all `TypedData` instances can be
+> shared on the Web, where there is separation between `ArrayBuffer` and
+> `SharedArrayBuffer` exists.
+
+[deeply immutable]: https://github.com/dart-lang/sdk/blob/bb59b5c72c52369e1b0d21940008c4be7e6d43b3/runtime/docs/deeply_immutable.md
+
 
 Unmodifiable lists (`List.unmodifiable`) which contain deeply immutable
 instances are deeply immutable.
@@ -176,51 +186,48 @@ instances are deeply immutable.
 Closures which capture only `final`, non-`late` variables containing deeply
 immutable instances are deeply immutable.
 
-Finally, instances of classes annotated with `@pragma('vm:deeply-immutable')`
-are deeply immutable. It is a compile error if classes annotated with this
-pragma contain non-`final` or `late final` fields. It is an compile time error
-if static type of field within annotated class excludes deeply immutable
-instances. If the static type of a field in a deeply immutable class is not
-deeply immutable type - then compiler must insert checks in the constructor to
-guarantee that this field is initialized to a deeply immutable value.
+Finally, instances of classes _marked as deeply immutable_ by being annotated
+with `@pragma('vm:deeply-immutable')` are deeply immutable. For any class
+which is marked as deeply immutable it is a compile time error if:
 
-> [!IMPORTANT]
->
-> **TODO** should we allow sharing of all `TypedData` (and by extension
-> `Struct`) objects? This seems very convenient. There is a consideration
-> here that not all `TypedData` instances can be shared on the Web, where
-> there is separation between `ArrayBuffer` and `SharedArrayBuffer` exists.
+* a subclass of such class which is not itself marked as deeply immutable.
+* a superclass of such class is not `Object` or a class itself marked as
+deeply immutable.
+* such class contains contains non-`final` or `late final` instance variables.
 
-[deeply immutable]: https://github.com/dart-lang/sdk/blob/bb59b5c72c52369e1b0d21940008c4be7e6d43b3/runtime/docs/deeply_immutable.md
+Compiler must ensure that instance variables in deeply immutable instances
+are initialized with deeply immutable values. If this can't be guarateed
+statically then compiler must insert appropriate checks into the constructor
+to guarantee this invariant.
 
-### Shared fields and variables (`@pragma('vm:shared')`).
+### Shared variables (`@pragma('vm:shared')`).
 
-Static fields and global variables annotated with `@pragma('vm:shared')` are
+Static and global variables annotated with `@pragma('vm:shared')` are
 shared across all isolates in the isolate group.
 
-A field or variable annotated with `@pragma('vm:shared')` can only contain
-values which are deeply immutable objects.
+A variable annotated with `@pragma('vm:shared')` can only contain values
+which are deeply immutable objects.
 
-* It is a compile time error to annotate a field or variable the static type of
+* It is a compile time error to annotate a variable the static type of
   which excludes deeply immutable objects;
-* If static type of a field is a super-type for both deeply immutable and
+* If static type of a variable is a super-type for both deeply immutable and
   non-deeply immutable objects then compiler will insert a runtime check
-  which ensures that values assigned to such field are deeply immutable.
-* A field or variable annotated with `@pragma('vm:shared')` must be `final` and
+  which ensures that values assigned to such variable are deeply immutable.
+* A variable annotated with `@pragma('vm:shared')` must be `final` and
   non-`late`.
 
 > [!NOTE]
 >
-> Restrictions imposed above are the same as ones imposed on field in deeply
-> immutable classes.
+> Restrictions imposed above are the same as ones imposed on instance
+> variables in deeply immutable classes.
 
-Shared fields must guarantee atomic initialization: if multiple threads
-access the same uninitialized field then only one thread will invoke the
-initializer and initialize the field, all other threads will block until
-initialization is complete.
+Shared static and global variables must guarantee atomic initialization: if
+multiple threads access the same uninitialized variable then only one
+thread will invoke the initializer and initialize the variable, all other
+threads will block until initialization is complete.
 
 Outside of initialization we however do **not** require strong (e.g.
-sequentially consistent) atomicity when reading or writing shared fields.
+sequentially consistent) atomicity when reading or writing shared variables.
 We only require that no thread can ever observe a partially initialized Dart
 object. See [Memory Model](#memory-model) for more details.
 
@@ -231,7 +238,7 @@ Today Dart runtime always executes Dart code within a specific isolate.
 within specific _isolate group_ but outside of a specific isolate. When Dart
 code is executed in such a way it can only access static state which is shared
 between isolates (`@pragma('vm:shared')`) and attempts to access isolated state
-will cause `FieldAccessError` to be thrown.
+will cause `AccessError` to be thrown.
 
 ```dart
   /// Constructs a [NativeCallable] that can be invoked from any thread.
@@ -240,8 +247,8 @@ will cause `FieldAccessError` to be thrown.
   /// the [callback] will be executed within the isolate group
   /// of the [Isolate] which originally constructed the callable.
   /// Specifically, this means that an attempt to access any
-  /// static or global field which is not shared between
-  /// isolates in a group will result in a [FieldAccessError].
+  /// static or global variable which is not shared between
+  /// isolates in a group will result in a [AccessError].
   ///
   /// If an exception is thrown by the [callback], the
   /// native function will return the `exceptionalReturn`,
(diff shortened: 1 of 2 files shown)


mraleph:
@aam @lrhn @mkustermann please take a look at the APIs. 

Among others, I would like to cover the following use case:

```dart
Pointer<Void> threadBody(Pointer<Void> ptr) {
  final isolate = Isolate.fork();

  isolate.runSync(() {
    // Do something on the current thread.
  });

  // Spin event loop on this thread until nothing left to do
  isolate.runEventLoopSync();

  isolate.shutdown();
}

pthread_create(..., NativeCallable.isolateGroupBound(threadBody));
```

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -173,14 +173,14 @@ All compile time constants are deeply immutable instances.
...
-instances are deeply immutable.
+Closures which capture only `final`, non-`late` variables containing deeply
+immutable instances are deeply immutable.
 
 Finally, instances of classes annotated with `@pragma('vm:deeply-immutable')`
 are deeply immutable. It is a compile error if classes annotated with this
-pragma contain non-`final` fields. It is an compile time error if static
-type of field within annotated class excludes deeply immutable instances.
-If the static type of a field in a deeply immutable class is not
+pragma contain non-`final` or `late final` fields. It is an compile time error
```
lrhn: fields -> instance variables.

I *think* that's what "field" is intended to cover, but I'm not sure whether it covers static variables or not. (Do we ever use the term "static field"?)

If I had to be pedantic here, and be explicit about the separation between syntactic declarations and semantic members, I'd probably go with:

> It's a compile-time error if a class declaration is annotated with `@pragma('vm:deeply-immutable')`
> and the class, or any of its superclasses, declare a non-`final` or `late` instance variable.

Do/should we require its superclass to be `vm:deeply-immutable` too? Then `Object` would be that, and it could be:
> If a class or mixin declaration is annotated with `@pragma('vm:deeply-immutable')`, then:
> * It's a compile-time error if the declared superclass or any declared mixins are not annotated with `@pragma('vm:deeply-immutable')`.
> * It's a compile time error if the class declare a non-`final` or `late` instance variable.

Or can you declare a class to be deeply-immutable even if its superclasses don't declare it, they just happen to be immutable.

(I assume you shouldn't inherit any mutable state Byt you don't really inherit *variables*, you inherit the getter and setter, which may have been overridden along the way. The only class declaration which knows that it's an instance variable is the class which introduces it.)

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -173,14 +173,14 @@ All compile time constants are deeply immutable instances.
 Unmodifiable lists (`List.unmodifiable`) which contain deeply immutable
 instances are deeply immutable.
 
-Closures which capture only `final` variables containing deeply immutable
-instances are deeply immutable.
+Closures which capture only `final`, non-`late` variables containing deeply
```
lrhn, line 186: (That depends very much on what closures capture. When the VM's closures over-capture variables that are not referenced by the closure's code, it's still hard to predict which variables are actually closed over.)

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
   /// to acquire exclusive access to the isolate.
   ///
   /// Throws [StateError] if target isolate is owned by another thread and
-  /// thus can't be entered from a different thread.
+  /// thus can't be entered from a different thread. See
+  /// [markOwnedByCurrentThread] and [isOwnedByCurrentThread].
+  ///
+  /// Throws [ArgumentError] if the target isolate belongs to another
+  /// isolate group.
```
lrhn, line 321: (Usually don't specify specific errors. Just say that it *must* belong to the same isolate group. If that's not detectable, and you want people to react to the throwing, throw a bespoke `Exception` instead. Same for the state error - if there is any situation where you *want* a client to catch the error, make it an exception. Otherwise don't bother specifying the exact error.)

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
   ///
   /// Throws [ArgumentError] if [f] is not deeply immutable.
   ///
   /// Throws [StateError] if result returned by [f] is not deeply immutable.
```
lrhn: (`UnsupportedError` would be more fitting.)

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
-  R runSync<R>(R Function() f, {Duration? timeout});
+  external R runSync<R>(R Function() f, {Duration? timeout});
+
+  /// Create a new isolate in the current isolate group.
+  ///
+  /// Similar to `Dart_CreateIsolateInGroup` Dart VM C API.
```
lrhn, line 330: Is this line useful to a reader? (Should I go read the docs on that function. If so, *link!*)
If not, I don't see the relevance for a Dart user reading this API.

(If it actualy *calls* that function, it might be relevant information, at some abstraction level. If it's just sort-of similar to a function that I don't know and can't be bothered to look at, that's just noise.)

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  ///
+  /// Created isolate is in runnable state, but its event loop is not running.
```
lrhn: What is "runnable state"? (First an only occurrence of that term in this file, no occurrences in `dart:isolate`.)

Maybe just "The isolate has been created, but its event loop is not yet running."

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  ///
+  /// To start processing isolate's messages:
+  ///
+  /// * start isolate's event loop synchronously on the current thread
+  ///   by calling [Isolate.runEventLoopSync]
+  /// * integrate isolate's event loop with other event loop by registering
+  ///   message callback ([Isolate.onMessage]) and draining pending messages
+  ///   ([Isolate.handleMessage]).
```
lrhn: Who does that? (Which *isolate* should call those methods?)

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  ///
+  /// Similar to `Dart_CreateIsolateInGroup` Dart VM C API.
+  ///
+  /// Created isolate is in runnable state, but its event loop is not running.
+  ///
+  /// To start processing isolate's messages:
+  ///
+  /// * start isolate's event loop synchronously on the current thread
+  ///   by calling [Isolate.runEventLoopSync]
+  /// * integrate isolate's event loop with other event loop by registering
```
lrhn: Why is the first step not enough? (What does the second step *do*?)

Which "other event loop"?
(Phrasing-wise I expected one of "another event loop", "the other event loop" or "other event loops".)

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  ///   message callback ([Isolate.onMessage]) and draining pending messages
+  ///   ([Isolate.handleMessage]).
+  external static Isolate fork({String? debugName});
+
+  /// Shutdown target isolate.
```
lrhn: Shutdown" -> "Shut down" (and name `shutDown`.)

What does it mean to shut down an isolate? (Likely the same thing `Isolate.kill` says, but it uses more words.)

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  ///
+  /// This function will block until it acquires exclusive access to the
+  /// target isolate. Isolate can only be entered for synchronous execution
+  /// between turns of its event loop, when no other thread is
+  /// executing code in the target isolate.
+  external void shutdown();
```
lrhn: How does this differ from `isolate.kill(Isolate.beforeNextEvent)`?

Could it be called `killSync()`?

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  /// * integrate isolate's event loop with other event loop by registering
+  ///   message callback ([Isolate.onMessage]) and draining pending messages
+  ///   ([Isolate.handleMessage]).
+  external static Isolate fork({String? debugName});
+
+  /// Shutdown target isolate.
+  ///
+  /// This function will block until it acquires exclusive access to the
+  /// target isolate. Isolate can only be entered for synchronous execution
+  /// between turns of its event loop, when no other thread is
```
lrhn, line 352: (Is that between full events or also between microtasks?)

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  external void shutdown();
+
+  /// Set current OS thread as owner of the isolate.
+  ///
+  /// Once an isolate is owned by some OS thread it can not be
+  /// entered by any other OS thread. An attempt to acquire
+  /// exclusive access to it from another thread will fail with
+  /// an error.
+  ///
+  /// Equivalent to `Dart_SetCurrentThreadOwnsIsolate` Dart VM C API.
```
lrhn, line 363: Again, is this actually useful?

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  ///
+  /// This function will block until it acquires exclusive access to the
+  /// target isolate. Isolate can only be entered for synchronous execution
+  /// between turns of its event loop, when no other thread is
+  /// executing code in the target isolate.
+  external void shutdown();
+
+  /// Set current OS thread as owner of the isolate.
+  ///
+  /// Once an isolate is owned by some OS thread it can not be
```
lrhn: "can not" -> "cannot"

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  /// entered by any other OS thread. An attempt to acquire
+  /// exclusive access to it from another thread will fail with
+  /// an error.
+  ///
+  /// Equivalent to `Dart_SetCurrentThreadOwnsIsolate` Dart VM C API.
+  ///
+  /// Throws [ArgumentError] if `this` is not `Isolate.current`.
```
lrhn: Just make it a static method that always works on the current isolate.

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  ///
+  /// Throws [StateError] if target isolate is already owned by another thread.
```
lrhn: Is there any way to check that? If not, maybe make it an exception.
Or just return `bool` saying whether it succeeded - if it's already owned by another thread, return false. If it's not, now you own it, and return true. If you already own it, still success, return true.

The `isOwnedByCurrentThread` doesn't help.

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  external void markOwnedByCurrentThread();
```
lrhn: Maybe `ownByCurrentThread()`.

"Mark as owned" -> "Own"!
(We don't care how it's represented, just what it means.)

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+
+  /// Returns `true` if the isolate is owned by the current OS thread.
```
lrhn: (Never start DartDoc with the word "Returns". Here it should be "Whether the ...")

working/333 - shared memory multithreading/shared_native_memory.md
lrhn: Is there a `unmark...` or `disown` method to remove ownership again, or is it permanent?

working/333 - shared memory multithreading/shared_native_memory.md
lrhn: When would you use this property?

Unlike `markAsOwnedByCurrentThread`, it doesn't require the isolate to be `Isolate.current`, so you can query other isolates for whether you own them, presumably before trying to call into them.
But you need to run code inside the isolate to get to own it, so not sure what the constraints are.

If you don't own the isolate, then you probably want to. In that case, you may need to know whether someone else owns it already. (So back to wanting a way to know that.)

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  ///
+  /// Equivalent to `Dart_GetCurrentThreadOwnsIsolate` Dart VM C API.
+  external bool get isOwnedByCurrentThread;
+
+  /// Run event loop for the target isolate synchronously on the current thread.
+  ///
+  /// This function will block until it acquires exclusive access to the
+  /// target isolate. Isolate can only be entered for synchronous execution
+  /// between turns of its event loop, when no other thread is
+  /// executing code in the target isolate.
```
lrhn, line 379: (Good! No nested event loops, that's madness!)

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  ///
+  /// The isolate will be marked as owned by the current thread.
+  ///
+  /// Similar to `Dart_RunLoop` Dart VM C API, but unlike `Dart_RunLoop` this
+  /// function executes isolate's event loop on the current thread instead
+  /// of delegating it into the thread-pool.
+  ///
+  /// Throws [StateError] if target isolate is owned by another thread.
+  external static void runEventLoopSync();
```
lrhn: Can that event loop ever exit, or will it run until the isolate shuts down?
(Reactive programs, and their event loops, generally never stop. That's the point. An event loop doesn't have an exit condition, as long as there may *ever* be another event.)

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+
+  /// Set message notify callback for the isolate.
```
lrhn: Drop the "Set". (Document properties with noun phrases.)
Or make it a function instead of a setter. 

Styles do not agree on how to handle `onFoo`s. Some uses streams. This is probably too low-level for that.
The `StreamSubscription` has methods, `sub.onData((data) { .... })`. I think it works slightly better than setters-without-getters.

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  ///
+  /// Provided callback will be called once for every message added to the
```
lrhn: The provided ...

What is this "message queue"? (First/only use in this file, no occurrences in `isolate.dart`.)

Could this be "pending event added to event loop"?

(Does it include microtasks?)

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  /// on an arbitrary thread and not necessarily within some isolate. See
+  /// [NativeCallable.isolateGroupBound].
+  ///
+  /// IMPORTANT: [Isolate.handleMessage] must *not* be called from the
+  /// `callback`.
+  ///
+  /// Similar to `Dart_SetMessageNotifyCallback` Dart VM C API.
+  external void set onMessage(void Function(Isolate) callback);
+
+  /// Handle a single pending message from isolate's message queue.
```
lrhn: What if there aren't any?
(I hope nothing happens.)

What happens if the isolate is paused?

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  external static void runEventLoopSync();
+
+  /// Set message notify callback for the isolate.
+  ///
+  /// Provided callback will be called once for every message added to the
+  /// isolates message queue. Pending messages can be then later be drained
+  /// by calling [Isolate.handleMessage].
+  ///
+  /// Provided [callback] must be deeply immutable and will be called
+  /// on an arbitrary thread and not necessarily within some isolate. See
```
lrhn: "not necessarily within some isolate"

Is that "not necessarily within a specific isolate" or "not necessarily within any isolate"?
(Guessing latter. I don't like the concept of running "outside of any isolate". I guess `Isolate.current` will throw, along with a lot of other things, but it's still *inside an isolate group*, which is weird if it's not an isolate. Terminology is hard. How about running in a "raw isolate", which doesn't have access to any of the things that require an event loop, or even being properly initialized. Sound .... "cool"! :sunglasses: ).

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  /// [NativeCallable.isolateGroupBound].
+  ///
+  /// IMPORTANT: [Isolate.handleMessage] must *not* be called from the
```
lrhn: (Normal scare typography would be emphasizing both "must" and "not": ... _must not_ be called ..., 
or even: ... **must not** be called ...

(Or "_**MUST NOT**_", just for good measure :grin: )

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  /// `callback`.
+  ///
+  /// Similar to `Dart_SetMessageNotifyCallback` Dart VM C API.
+  external void set onMessage(void Function(Isolate) callback);
```
lrhn: Can an isolate have more than once callback? It sounds like it's set as a global property on the isolate itself, not the individual `Isolate` objects that exist in different isolates themselves.

Is setting it synchronous? Can there be race conditions?

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+
+  /// Handle a single pending message from isolate's message queue.
+  ///
+  /// This function will block until it acquires exclusive access to the
+  /// target isolate. Isolate can only be entered for synchronous execution
+  /// between turns of its event loop, when no other thread is
```
lrhn, line 417: What is a "turn"? :wink:

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  /// executing code in the target isolate.
```
lrhn, line 418: So this is what an isolate's event loop would do while there are events.

Will this function work if the isolate's event loop is running - effectively inserting itself before the event loop runs the next event, to ... run the next event, but so that this call will return.

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -173,14 +173,14 @@ All compile time constants are deeply immutable instances.
...
+Closures which capture only `final`, non-`late` variables containing deeply
+immutable instances are deeply immutable.
 
 Finally, instances of classes annotated with `@pragma('vm:deeply-immutable')`
 are deeply immutable. It is a compile error if classes annotated with this
-pragma contain non-`final` fields. It is an compile time error if static
-type of field within annotated class excludes deeply immutable instances.
-If the static type of a field in a deeply immutable class is not
+pragma contain non-`final` or `late final` fields. It is an compile time error
+if static type of field within annotated class excludes deeply immutable
```
lrhn: if static type -> if the static type. 
I think. I can't actually decipher the rest of the sentence anyway.

What _does_ "excludes deeply immutable instances" mean?
The only other occurrences of "excludes" is on line 204, and it doesn't define it either.
The following text suggests that a supertype that allows both immutable and mutable subtypes is not included, probably including `Object`.

Which types *are* excluded then?

If it's a type which *cannot possibly (ever)* have an immutable subtype, I don't know what those would be. You could implement any implementable interface with no state and no superclass other than `Object`.

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  /// Provided callback will be called once for every message added to the
+  /// isolates message queue. Pending messages can be then later be drained
+  /// by calling [Isolate.handleMessage].
+  ///
+  /// Provided [callback] must be deeply immutable and will be called
+  /// on an arbitrary thread and not necessarily within some isolate. See
+  /// [NativeCallable.isolateGroupBound].
+  ///
+  /// IMPORTANT: [Isolate.handleMessage] must *not* be called from the
+  /// `callback`.
```
lrhn: (Nor must a lot of other things, probably, but I can see why this one is more relevant here.)

How do you send a message to someone who _can_ call `handleMessage`?
(If all you can do is to send a port message, then you could just say which send port to send a message to, but I guess with threads you have other options too).

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -206,7 +206,8 @@ values which are deeply immutable objects.
 * If static type of a field is a super-type for both deeply immutable and
   non-deeply immutable objects then compiler will insert a runtime check
   which ensures that values assigned to such field are deeply immutable.
-* A field or variable annotated with `@pragma('vm:shared')` must be `final`.
+* A field or variable annotated with `@pragma('vm:shared')` must be `final` and
```
lrhn: (A "field" is typically a variable, so what does "variable" cover here?)

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -206,7 +206,8 @@ values which are deeply immutable objects.
 * If static type of a field is a super-type for both deeply immutable and
   non-deeply immutable objects then compiler will insert a runtime check
   which ensures that values assigned to such field are deeply immutable.
```
lrhn: How do we know that any type is not a supertype of a deeply immutable object?

Is it a whole-program analysis, and can change when third-party code changes, or is it a property of the type in some way?

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
   /// to acquire exclusive access to the isolate.
   ///
   /// Throws [StateError] if target isolate is owned by another thread and
-  /// thus can't be entered from a different thread.
+  /// thus can't be entered from a different thread. See
```
lrhn: from a different thread -> from this thread

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
   /// to acquire exclusive access to the isolate.
```
lrhn, line 314: All these thread-based members of isolate, would it make sense to put them somewhere else than in `Isolate`, like in `dart:isolate_threads`, where they could be extension members on `Isolate` if we really want them there.

Do we expect most people who use isolates to also want to use these thread-based functions, or will they clutter the API for the "normal" uses of `Isolate` and enocurage people to use them who *really shoudn't*?

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  /// between turns of its event loop, when no other thread is
+  /// executing code in the target isolate.
+  ///
+  /// The isolate will be marked as owned by the current thread.
+  ///
+  /// Similar to `Dart_RunLoop` Dart VM C API, but unlike `Dart_RunLoop` this
+  /// function executes isolate's event loop on the current thread instead
+  /// of delegating it into the thread-pool.
+  ///
+  /// Throws [StateError] if target isolate is owned by another thread.
```
lrhn: Probably want to throw if you ever call it on `Isolate.current`.
The code that calls it will be blocked waiting for control to go back to the event loop, so you have an instant deadlock.

lrhn (approved):
I'm being pedantic because I've seen too many cases of "everybody knows what this obviously means" turning into everybody thinking it obviously means different things.

So use words and phrases consistently, internally and with existing documentation.
Don't use a new word if an existing word is already in use.
If a new concept is important, define it and give it a unique name. (And define the concepts at the user/API level, nobody else cares about the implementation.)

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
   /// to acquire exclusive access to the isolate.
   ///
   /// Throws [StateError] if target isolate is owned by another thread and
-  /// thus can't be entered from a different thread.
+  /// thus can't be entered from a different thread. See
+  /// [markOwnedByCurrentThread] and [isOwnedByCurrentThread].
+  ///
+  /// Throws [ArgumentError] if the target isolate belongs to another
```
mkustermann: How can an isolate obtain an `Isolate` object from a different group?

working/333 - shared memory multithreading/shared_native_memory.md
mkustermann: Sidenote: This may cause strange/flaky behavior: For example a user may share `File().readAsBytesSync()` and thinks it works and rely on it. Though implementation may return native memory (e.g. for large amount of data) or may not (e.g. for small amount of data).

mkustermann, line 313: If an isolate is dead but someone uses the `Isolate` object, then what happens? Will an operation timeout will it throw an error?

I guess `Isolate` is a weak reference to an actual isolate. So using it has to turn the weak reference into a strong reference using some global data structure with locking (bit like `PortMap`).

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -306,21 +307,103 @@ class Isolate {
...
+  /// Created isolate is in runnable state, but its event loop is not running.
+  ///
+  /// To start processing isolate's messages:
+  ///
+  /// * start isolate's event loop synchronously on the current thread
+  ///   by calling [Isolate.runEventLoopSync]
+  /// * integrate isolate's event loop with other event loop by registering
+  ///   message callback ([Isolate.onMessage]) and draining pending messages
+  ///   ([Isolate.handleMessage]).
+  external static Isolate fork({String? debugName});
```
mkustermann: I find the name `fork` strange as a C `fork()` call actually gives you a copy of the state of the process, there's no clearing of state to some initial state happening.

Furthermore from  the motivating example in https://github.com/dart-lang/language/pull/4620#issuecomment-3799774821 it seems that the caller may not even be an isolate with state. So when a Thread (aka shared isolate) calls `Isolate.fork()` it gets an isolate with freshly initialized state.

Why not call it `Isolate.create()` instead?

mkustermann (approved):

knopp:
> ```dart
>  isolate.runEventLoopSync();
> ```

In addition to `isolate.runEventLoopSync();` I'd like to have a method to "poll" the event loop, i.e. bock until there is event scheduled and then processes the event.

If we ever wanted to for example reimplement the embedder in Dart, we'd need this for resize synchronization (i.e. we want to be able run the dart event loop in a blocking way until surface with appropriate size is generated). Right now we have an implementation for this per platform (a custom CFRunLoopMode on macOS or [polling](https://github.com/flutter/flutter/blob/1e1df36da3654f3bec977435cee56637ab77d398/engine/src/flutter/shell/platform/windows/task_runner_window.cc#L164) the loop on windows). This only processes Flutter messages and ignores the platform events.

mraleph:
@knopp polling is implementable on top of `onMessage` and `handleMessage` AFAIK.

knopp:
Not sure what's the good way to submit feedback to the proposal, just something I've noticed:
> [NativeCallable.isolateLocal](https://api.dart.dev/stable/3.2.4/dart-ffi/NativeCallable/NativeCallable.isolateLocal.html): native caller must have an exclusive access to an isolate in which callback was created. This type of callback works if Dart calls C and C calls back into Dart synchronously. It also works if caller uses VM C API for entering isolates (e.g.Dart_EnterIsolate/Dart_ExitIsolate).

Just a nit, but after https://github.com/dart-lang/sdk/issues/61623, `NativeCallable.isolateLocal` will enter the isolate as long as current thread owns the isolate.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: There is documentation on those methods on how to use them. I am not sure if it makes sense to copy it here?

In general I would expect that these are called from outside the isolate (i.e. from the code running inside the isolate group, but not in any concrete isolate).

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: I have renamed "other event loop" to "external event loop". This path exists for integrating Dart isolate with code which already contains some sort of an event loop (e.g. imagine main platform thread on Android which has its own event loop). In this situation you use `Isolate.onMessage` to post notifications about pending messages to this external event loop - and then when that external event loop turns around you can drain the message you have been notified about using `handleMessage`.

I welcome any further suggestions.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: Tried to add a bit more information.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: It's closer to `Isolate.kill(Isolate.immediate)` but completely synchrounous.

I would prefer to align terms with existing names from VM C API.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: Only between event loop turns, because microtask processing does not release isolate access.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: I wish I could link to these - but we don't publish documentation for these. So the best I could do is to link to sources, which is a bit wonky. 

It is hard to say whether these are useful or not - if reader is familiar with C API layer they are certainly useful because most of APIs introduced here are exposing functionality which is available through C API. 

I would prefer to keep these here for the purpose of spec review / but we can strip them later when landing actual code.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: See above.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: Renamed to `pinToThread`.

mraleph: No, once it is pinned (owned) you can't change the ownership.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: You can use this to decide whether you can just enter the isolate or you need to message it.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: Our event loop exits once all keep alive receive ports are closed.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: Unless we have some specific rule prohibiting setter-without-getter I would prefer to use that. 

We already have similar code in other places: https://api.dart.dev/dart-isolate/RawReceivePort/handler.html

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: This does not include microtasks. Dart's isolate event loop is effectively a loop dispatching messages to receive ports - hence "message" here. 

Renamed to use event instead of message.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: - Nothing happens. 

- Not all isolates can be paused (e.g. `pauseCapability` can be `null`). I think currently you can't pause isolates which use custom event loops.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: Yeah, I just don't see any good naming for this, hence the reference to `isolateGroupBound`. I think names like `xyz isolate` don't work - because it is not isolate at all. The most accurate description is "no-non-shared-statics-isolate", but alas it's way too long.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: Ack.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: Correct - it's a property on the underlying isolate. If you set it twice - the last one wins. Setting is synchronous.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: > Will this function work if the isolate's event loop is running

It will work if isolate is not pinned to a thread - because unpinned isolates can _move_ between threads I see no immediate problem with allowing this.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: Fair. Removed this.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: This is for integration with external event loops. So you do something like 

```
DispatchQueue.main.async {
    // call handleMessage here
}
```

or an equivalent for your specific event-loop.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: Yeah, this sentence is clunky. Rewritten it.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: Yeah, these are fairly low-level. Maybe we need to introduce `dart:threading`?

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: I think calling it on current isolate is not a problem, but calling it on an isolate which has an event loop running is a problem. Added a sentence saying that an error will be thrown.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: `spawnUri` gives you an `Isolate` object in a different group.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: Removed requirement for them to be external. Lets consider them deeply immutable always unless there is an opposition to it.

mraleph: If isolate dies operations with it should throw an error - `Isolate` is just a port number and yeah all operations with it are based on that. We can guard against dead isolates to some extent - though if port number gets reused you will get strange effects. But I am not sure this can be fully resolved.

working/333 - shared memory multithreading/shared_native_memory.md
mraleph: Renamed to `Isolate.create`.

mraleph:
@lrhn @mkustermann I think I addressed most of comments. 

I think the main question that remains is whether or not to move low-level isolate members to a separate library. 

I am starting to think that maybe `dart:threading` is an appropriate name for it. Any other ideas?

aam (approved):

working/333 - shared memory multithreading/shared_native_memory.md
```diff
@@ -305,22 +313,115 @@ class Isolate {
...
+  ///
+  /// The isolate will be marked as pinned to the current thread.
+  ///
+  /// Similar to `Dart_RunLoop` Dart VM C API, but unlike `Dart_RunLoop` this
+  /// function executes isolate's event loop on the current thread instead
+  /// of delegating it into the thread-pool.
+  ///
+  /// Throws an error if target isolate is pinned to another thread or already
+  /// has an event loop running.
+  external static void runEventLoopSync();
```
aam: Did you mean this to be instance rather than `static` method?

mraleph:
I am going to land this and @aam is working to implementing the prototype. We have few remaining unanswered questions (e.g. where should new methods reside), but we can address them later.

PR #4622, opened 2026-01-27
Primary constructors spec - fix a few typos and update flag.

kallentu:
As title says, fixing 2 typos and then updating the flag to `primary-constructors`.


accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -177,7 +177,7 @@ declaration, which may be helpful when reading the code.
 With the primary constructor, the modifier `const` could have been
 placed on the class (`const class`) rather than on the class name. This
 feature puts it on the class name because the notion of a "constant class"
-conflicts with with actual semantics: It is the constructor which is
+conflicts with actual semantics: It is the constructor which is
```
eernstg: I probably wanted to write this:

```suggestion
conflicts with the actual semantics: It is the constructor which is
```

Issue #4623, opened 2026-01-28
Allow augmentation of getters and setters with a larger set of possible implementations


eernstg:
Consider the following program:

```dart
class A {
  get v01;
  get v02;
  set v02(_);
  int get v03;
  set v03(int _);
  static get v04;
  static get v05;
  static set v05(_);
  static int get v06;
  static set v06(int _);
  abstract covariant var v07;
  abstract covariant int v08;
  static const v09; // Can not be augmented.
  static const int v10; // Can not be augmented.
  static get v11; 
  static int get v12;
  static get v13;
  static set v13(_);
  static int get v14;
  static set v14(int _);
  get v15;
  set v15(covariant _);
  int get v16;
  set v16(covariant int _);
  get v17;
  set v17(covariant _);
  int get v18;
  set v18(covariant int _);
  get v19;
  int get v20;
  get v21;
  set v21(_);
  int get v22;
  set v22(int _);
  get v23;
  set v23(_);
  int? get v24;
  set v24(int? _);
  get v25;
  set v25(_);
  int get v26;
  set v26(int _);
}

augment class A {
  augment external final v01;
  augment external var v02;
  augment external int v03;
  augment external static final v04;
  augment external static var v05;
  augment external static int v06;
  augment external covariant var v07;
  augment external covariant int v08;
  // augment static const v09; // Error, cannot augment constant variables.
  // augment static const int v10; // Error, cannot augment constant variables.
  augment static final v11 = null;
  augment static final int v12 = 0;
  augment static late final v13;
  augment static late final int v14;
  augment covariant late final v15;
  augment covariant late final int v16;
  augment covariant late var v17; 
  augment covariant late int v18;
  augment final v19 = null;
  augment final int v20 = 0;
  augment late final v21;
  augment late final int v22;
  augment var v23;
  augment int? v24;
  augment late var v25;
  augment late int v26;
}
```

This program shows a number of augmenting declarations involving getters, setters, and variables. All declarations starting with `augment` are syntax errors according to the [augmentations feature specification](https://github.com/dart-lang/language/blob/main/working/augmentations/feature-specification.md). However, I think all of them are meaningful, and it may be a natural expectation that they can be used.

If we are going to allow them then we'd need to generalize the grammar slightly (in particular, by adding `'augment'?` to almost all grammar rules in the rule for `<declaration>`).

[...]

Issue #4624, opened 2026-01-29
Ensure augmentations don't allow uninitialized top-level and static variables


munificent:
In #4623, the idea is to widen the grammar to ensure that augmentations are allowed to contain pretty much any kind of variable/getter/setter declaration. The *semantic* restrictions around not being able to augment an already complete declaration remain, but syntactically, the grammar allows you to write just about everything.

There is an important restriction in the language that some variables must be initialized at their declaration. @eernstg can correct me if I get this wrong, but I think the rules are:

If the variable is:

* top-level or a static field,
* and it's `final` or `const`,
* and it's *not* `late`,

Then it needs to have an initializer. (Local `const` declarations also need initializers, but those aren't affected by augmentations.)

It looks like the behavior is currently captured in the spec outside of the grammar:

> A compile-time error occurs if a static or library variable has no initializing ⊖
expression and a type which is not nullable (??), unless the variable declaration
has the modifier late or the modifier external.

> A constant variable is a variable whose declaration includes the modifier △
const. A constant variable must be initialized to a constant expression (17.3),
or a compile-time error occurs. 

So I don't think there's anything we need to do for augmentations. But it might be worth pointing out in the augmentations spec that this restriction remains.

eernstg:
I adjusted https://github.com/dart-lang/language/pull/4626 to make sure this is stated explicitly in the augmentations feature spec.

PR #4626, opened 2026-01-29
Adjust the grammar to allow additional augmenting variable declarations

eernstg:
See https://github.com/dart-lang/language/issues/4623 for background discussions.

This PR simplifies and generalizes the grammar rules containing `'augment'` and deriving `<declaration>` such that 26 new forms of augmenting declaration can be derived. The rationale is that they are conceptually and semantically consistent with the specification rules on similar matters, it was just the grammar that prevented them from being expressed.

The specification already contains a rule that makes it an error for an augmenting declaration to contain a constant variable declaration:

> It's a **compile-time error** if:
> *   ...
> *   A `const` variable declaration is augmented or augmenting.

This implies that two of those 26 forms will always be an error. I preferred to handle it this way because it keeps the grammar simpler and more readable, and the desired errors will be reported even though they are not syntax errors.

As a consequence of this simplification, `<memberDeclaration>` was simplified as well, as `'augment'?` was moved up to the rule `<memberDeclarations>`.

(This PR also fixes one typo.)

### DIFF
--- a/working/augmentations/feature-specification.md
+++ b/working/augmentations/feature-specification.md
@@ -887,6 +887,12 @@ It's a **compile-time error** if:
     an instance member that can be abstract. In that case, if no declaration
     provides a body, it is considered abstract.*
 
+*Using the general [compile-time error principle], some additional situations
+are errors. In particular, a compile-time error occurs if, after application of
+all augmentations, a static or library variable has no initializing expression,
+and its type is not nullable, and the declaration does not have the modifier
+`late` nor the modifier `external`.*
+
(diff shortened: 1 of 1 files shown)


eernstg:
I added a paragraph to address #4624. I don't think there's a need to say anything extra about constant variables, because they are already excluded from being augmented by anything, and from augmenting anything.

eernstg:
Hi @davidmorgan, could you take a look at this one? @munificent is busy.

eernstg:
Thanks!

Issue #4592, opened 2025-12-11
Allow `abstract` variables in more places with augmentations?


eernstg:
Thanks to @sgrekhov for bringing this up.

The [feature specification](https://github.com/dart-lang/language/blob/main/working/augmentations/feature-specification.md#class-like-declarations) mentions a declaration of a variable starting with `static abstract`, and indicates that it should be possible to use the keyword `abstract` to introduce an incomplete variable declaration even in locations where abstract members were previously not supported (top-level, static).

However, we would need to adjust the grammar in order to actually support this.

It seems likely that we would do it, based on the fact that it is very similar in nature (if not syntactically) to the support for functions/getters/setters with no body at the top level, and as a static member.

lrhn:
I think it's potentially too confusing.

A `void foo();` siganature is a signature with no implementation. It doesn't say that the member is abstract, it just doesn't give an implementation. You might inherit an implementation!

An `abstract final int foo;` does _say_ that the member is abstract.
It's functionally identical to `int get foo;`, but it sends a slightly different signal.

I have considered allowing `final int foo;` to be taken as an incomplete declaration of a `foo` getter,
but that's also confusing. It really, really looks like an instance variable declaration, and constructors may think they need to initialize the instance variable. If you can implement it using `augment int get foo => 42;`, then it wasn't. 
So that's also not a good idea. I'd rather save that for an incomplete *actual instance variable*.

It's also annyoing to have to write `int get foo; set foo(int _);`, but that *is* what you are saying the class will have to implement using augmentations (if it's not abstract).

Nothing technical prevents using `abstract int foo;` as an incomplete declaration for a getter and setter.
It's just sending a confusing signal if the variable isn't actually going to be abstract.

eernstg:
> Nothing technical prevents using `abstract int foo;` as an incomplete declaration for a getter and setter.
> It's just sending a confusing signal if the variable isn't actually going to be abstract.

The crucial point here is that the grammar currently does not allow this to be a `static` declaration. Similarly, the grammar _does_ allow a top-level variable (`final` or not) to be `abstract`. I thought we'd want to double check this apparent inconsistency.

lrhn:
ACK. So if we allow `abstract int foo;` as an instance member to be augmented with a complete variable implementation (or even a getter + setter implementation), we should allow the same syntax for a to-be-augmented declartion for static and top-level members.

[...]

Issue #4014, opened 2024-08-06
Make duplication of superinterfaces a warning, not an error


eernstg:
Let us drop the compile-time error for the following situations (and, presumably, introduce a warning or lint):

It is currently an error in a declaration of a class, mixin class, mixin, enum, or extension type, when said parts exist, if ...
- the superclass occurs in the `implements` clause.
- the `implements` clause contains duplicates.

These situations are confusing, but harmless, and it is sufficient to flag them using a warning or lint.

For consistency, it would also make sense to flag the situation where an element in the `on` clause of a mixin declaration occurs in the `implements` clause as well. We might even warn against duplicate mixins (`S with M, M` has a different meaning than `S with M`, so that is a different kind of situation than `implements I, I`, but I have basically never seen a convincing real-world use case for duplicate mixins, and it's probably always a mistake).

@dart-lang/language-team, WDYT?

@johnniwinther, @scheglov, do you think this would be easy to do, or will the fact that `implements I, I` etc. can occur break some invariants in the implementation?

johnniwinther:
It should be easy to do in the CFE.

jakemac53:
SGTM

leafpetersen:
Fine by me.

eernstg:
Done in https://github.com/dart-lang/language/pull/4642.

johnniwinther:
@eernstg Do we have implementation issues for this change?

eernstg:
Not yet, I'm about to land the spec change, will create an implementation issue then.

eernstg:
I created this one: https://github.com/dart-lang/sdk/issues/62750.

PR #4642, opened 2026-02-19
Update the 'augmentations' feature specification

eernstg:
This PR updates the feature specification in several ways.

The grammar rule for final and constant top-level variables is modified to use a `staticFinalDeclarationList`. This forces every variable to have an initializing expression. This is necessary because (1) a constant variable can never be augmented and it must have an initializing expression, and (2) a final top-level variable declaration is necessarily complete, which means that it is an error for any other declaration in the augmentation chain to provide an initializing expression, because that other declaration would then also be complete.

An alternative approach would be to keep the grammar unchanged and then specify a compile-time (non-syntax) error in the case where there is no initializing expression. However, this does not simplify the grammar, it just makes the grammar derive some wrong programs for no reason. Hence the grammar change.

A grammar rule deriving `static abstract` variable declarations is added. This is needed because we have `abstract` top-level variables and `abstract` instance variables, and it is inconsistent that `static abstract` cannot be derived. See also #4592.

The notion of matching signatures is updated to specify that the names of parameters must always be the same. (Previously, it was possible to use `_` as a "don't care" alternative to writing the actual parameter name, that's no longer supported.)

A new section is added to specify that `implements I, I` and `extends C implements C` and `mixin M on I implements I` are no longer errors. See also #4014.

### DIFF
--- a/working/augmentations/feature-specification.md
+++ b/working/augmentations/feature-specification.md
@@ -692,12 +692,12 @@ It's a **compile-time** error if:
     what is known from reading the introductory declaration will still be true
     after augmentation.*
 
-*   The type parameters of the augmenting declaration do not match the augmented
-    declarations's type parameters. This means there must be the same number of
-    type parameters with the exact same type parameter names (same identifiers)
-    and bounds if any (same *types*, even if they may not be written exactly the
-    same in case one of the declarations needs to refer to a type using an
-    import prefix).
+*   The type parameters of the augmenting declaration do not match the
+    introductory declarations's type parameters. This means there must be the
+    same number of type parameters with the exact same type parameter names
+    (same identifiers) and bounds if any (same *types*, even if they may not be
+    written exactly the same in case one of the declarations needs to refer to a
+    type using an import prefix).
 
     *Since repeating the type parameters is, by definition, redundant, this
     restriction doesn't accomplish anything semantically. It ensures that
@@ -707,7 +707,7 @@ It's a **compile-time** error if:
     augmentation.*
 
     The augmenting declaration may choose to omit the bound on any type
-    parameter, in which case it will be inherited from the augmented
+    parameter, in which case it will be inherited from the introductory
     declaration.
 
     ```dart
@@ -725,30 +725,32 @@ It's a **compile-time** error if:
 When augmenting a function (top level, static method, instance method, etc.) or
 constructor (generative, factory, etc.) the parameter lists must be the same in
 all meaningful ways. We say that an augmenting function or constructor's
-signature *matches* if:
+signature *matches* an introductory signature if:
 
 *   It has the same number of type parameters with the same type parameter names
     (same identifiers) and bounds (after type annotation inheritance), if any
     (same *types*, even if they may not be written exactly the same in case one
     of the declarations needs to refer to a type using an import prefix).
 
-*   The return type (if not omitted) is the same as the augmented declaration's
-    return type.
+*   The return type (if not omitted) is the same as the introductory
+    declaration's return type.
 
-*   It has the same number of positional parameters as the augmented
+*   It has the same number of positional parameters as the introductory
     declaration, and the same number of those are optional.
 
-*   It has the same set of named parameter names as the augmented declaration.
+*   It has the same set of named parameter names as the introductory
+    declaration.
 
 *   For each corresponding pair of parameters:
 
     *   They have the same name. _This is trivial for named parameters, but may
         fail to hold for positional parameters._
 
-    *   They have the same type (if not omitted).
+    *   They have the same type (or the augmenting declaration omits the type).
 
-    *   They agree on having or not having the modifier `covariant` and, if
-        named, the modifier `required`.
+    *   They both have the modifier `covariant`, or none of them have it.
+
+    *   They both have the modifier `required`. or none of them have it.
 
     *For constructors, we do not require parameters to match in uses of
     initializing formals or super parameters. In fact, they are implicitly
@@ -758,6 +760,11 @@ signature *matches* if:
     Instead, at most only one of the constructors can use initializing formals
     or super parameters and all other declarations for the same constructor
     must declare the corresponding parameters as regular parameters.*
+
+In this definition, the properties of the introductory declaration may
+correspond to an explicitly declared property _(such as an explicitly stated
+return type)_ or an inferred property _(such as a parameter type which has been
+obtained by override inference)_.
 
 ### Augmenting functions
 
@@ -807,7 +814,7 @@ specified in the augmentation chain.
 It's a **compile-time** error if:
 
 *   The signature of the augmenting function does not [match][signature
-    matching] the signature of the augmented function.
+    matching] the signature of the corresponding introductory declaration.
 
 *   More than one declaration in the augmentation chain specifies a default
     value for the same optional parameter. This is an error even in the
@@ -849,7 +856,7 @@ setter with a non-abstract variable declaration.*
 It's a **compile-time error** if:
 
 *   The signature of the augmenting getter or setter does not [match][signature
-    matching] the signature of the augmented getter or setter.
+    matching] the signature of the corresponding introductory getter or setter.
 
 *   A `const` variable declaration is augmented or augmenting.
 
@@ -901,7 +908,7 @@ implement the constructor by adding a redirection or a body.*
 It's a **compile-time error** if:
 
 *   The signature of the augmenting constructor does not [match][signature
-    matching] the signature of the augmented constructor.
+    matching] the signature of the corresponding introductory constructor.
 
 *   More than one declaration in the augmentation chain specifies a default
     value for the same optional parameter. This is an error even in the
@@ -1001,9 +1008,9 @@ situations:
   type. _For example `mixin M on I implements I;` is no longer an error_.
 
 An operand of an `extends`, `with`, `on`, or `implements` clause is a
-top-level construct in the type or list of types specified by the clause.
-_In particular, these changes have no effect on situations like
-`class A extends B<I> implements C<I, I Function(String)>;`
+type that occurs in the clause which is not a subterm of another type.
+_In particular, these changes have no relevance to nested occurrences like
+`I` in `class A extends B<I> implements C<I, I Function(String)>;`
 (which were not errors, anyway)._
 
 The motivation for these changes is mainly that, in general, compile-time errors
@@ -1037,8 +1044,8 @@ definitions.
 
 For example, it is an error according to the language specification if a
 concrete class has an abstract instance member declaration _D_, and there is no
-implementation inherited from a superclass which is a correct override of the
-member signature of _D_.
+implementation inherited from a superclass whose signature is a correct override
+of the member signature of _D_.
 
 Thus this is an error in Dart without augmentations:
 
@@ -1053,15 +1060,18 @@ missing implementation in an augmenting declaration:
 
 ```dart
 augment class C {
-  augment get g => 0;
+  augment int get g => 0;
 }
 ```
 
 The general rule is that **compile-time errors apply to semantic definitions
-whenever possible.** In other words, if the library is syntactically well-formed
-enough that augmentations *can* be applied, then they should be. And if doing so
-eliminates what would otherwise be a compile-time error, then that error should
-not be reported.
+whenever possible.** For this example it is *the class as a whole*, as defined
+by all its introducing and augmenting declarations, which must have a concrete
+member whose signature is a valid override for its interface member signature.
+Having some individual member declarations which do not introduce an
+implementation does not mean that the class does not introduce an implementation
+of that member, as long as there is one concrete member declaration for the
+member.
 
 The motivation for this principle is that reorganizing code into or out of
 augmentations shouldn't affect the errors that are reported. Augmentations are
@@ -1294,12 +1304,9 @@ and assume the third point is always true.
 
 *   Adjust the definition of signature matching (using `_` as "don't care"
     parameter names is no longer supported).
-    
+
 *   Add a section to say that certain redundant superinterfaces are no longer an
     error.
-    
-*   
-
 
 ### 1.40
 


eernstg:
@lrhn, could you take a look at this one? I think the only controversial part is that it introduces support for `static abstract` variables.

working/augmentations/feature-specification.md
lrhn: Matches *what*?

In particular, is it the prior declaration or the introducing declaration?
(It's the introducing declaration, because that's the declaration that cannot omit any types or names, so there is always something to compare with.).

working/augmentations/feature-specification.md
```diff
@@ -734,16 +735,20 @@ signature *matches* if:
...
 
-*   For all corresponding pairs of parameters:
+*   For each corresponding pair of parameters:
 
-    *   They have the same type (if not omitted in the augmenting declaration).
+    *   They have the same name. _This is trivial for named parameters, but may
+        fail to hold for positional parameters._
 
-    *   They have the same `required` and `covariant` modifiers.
+    *   They have the same type (if not omitted).
```
lrhn: ... if not omitted by the augmenting declaration.

Here it's important that we're comparing to the introducing declaration.
Otherwise you could have
```dart
void foo(int x);
augment void foo(x);
augment void foo(String x);
```
and no augmenting declaration disagrees about the type with the declaration it augments, but it's still completely wrong.

working/augmentations/feature-specification.md
```diff
@@ -734,16 +735,20 @@ signature *matches* if:
...
+
+    *   They agree on having or not having the modifier `covariant` and, if
+        named, the modifier `required`.
```
lrhn: "agree" is a little anthropomorhic.
Consider:
>   * Either both or neither has the `covariant` modifier.
>   * Either both or neither has the `required` modifier.

Don't bother saying that they must be named. It's a separate error (for each!) if they have `required` and are not named, or `covariant` and not on an instance member.

working/augmentations/feature-specification.md
```diff
@@ -1010,6 +983,46 @@ the combined declaration.*
...
+With this feature, it is **no longer an error** to have the following
+situations:
+
+* An `implements` clause contains two or more type operands denoting the same
+  type. _For example, `class A implements I, I;` is no longer an error_.
+
+* An `implements` clause of a declaration _D_ contains an operand `T`, and an
+  `extends` or `with` clause of _D_ contains an operand `S`, and `T` and `S`
+  denote the same type. _For example, `class A extends I implements I;` is no
+  longer an error_.
```
lrhn, line 1004: (`with I implements I` never was an error, but it still isn't. \</meme>)

working/augmentations/feature-specification.md
```diff
@@ -1010,6 +983,46 @@ the combined declaration.*
...
+
+* An `implements` clause of a mixin declaration _D_ contains an operand `T`, and
+  an `on` clause of _D_ contains an operand `S`, and `T` and `S` denote the same
+  type. _For example `mixin M on I implements I;` is no longer an error_.
```
lrhn, line 1008: (Not sure this was validly an error, but the CFE did report it, with an error message suggesting that it treates it like `extends I implements I`. Analyzer has no error.)

working/augmentations/feature-specification.md
```diff
@@ -1010,6 +983,46 @@ the combined declaration.*
...
+
+An operand of an `extends`, `with`, `on`, or `implements` clause is a
+top-level construct in the type or list of types specified by the clause.
```
lrhn: I have no idea what that tries to say.
Is that a change? To/from what?
Is it introducing a new terminology for referring to the elements of the `implements` type list? 

If so, introduce it *before* using it, and say that that's what you are doing. "We use the term _operand_ of an `extends`, `with`, `on` or `implements` clause about ..."./

"Top-level" is overloaded, and confusing.
Maybe say that it's "... a \<typeNotVoid\> occuring directly in the `extends` clause, or in the \<typeNotVoidList\> of an `on`, `with` or `implements` clause"?

(Does `operand` refer to the syntactic clause or the type it denotes?)

working/augmentations/feature-specification.md
```diff
@@ -1010,6 +983,46 @@ the combined declaration.*
...
+_In particular, these changes have no effect on situations like
+`class A extends B<I> implements C<I, I Function(String)>;`
+(which were not errors, anyway)._
```
lrhn, line 1014: So, why say anything? I didn't even *consider* that it could be a problem, and ... now I do?
(It's a spec, _everything_ is potentially important, so overcommunicating suggests that something important is going one, and if it isn't, that just confuses the reader.)

working/augmentations/feature-specification.md
```diff
@@ -1020,48 +1033,44 @@ those entities interchangeably. With augmentations, that is no longer the case.
...
 (an introductory and any number of augmentations). This raises the question of
 whether existing compile errors apply to syntactic declarations or semantic
-definitions. For example, the specification says:
+definitions.
 
-> It is a compile-time error if two elements in the type list of the
-> `implements` clause of a class `C` specifies the same type `T`.
+For example, it is an error according to the language specification if a
+concrete class has an abstract instance member declaration _D_, and there is no
+implementation inherited from a superclass which is a correct override of the
```
lrhn: "which is" -> "whose signature is"

working/augmentations/feature-specification.md
```diff
@@ -1020,48 +1033,44 @@ those entities interchangeably. With augmentations, that is no longer the case.
...
 ```
 
-Each syntactic declaration of `C` only mentions the interface once. But the
-resulting definition produced by applying augmentations has an `implements`
-clause with `I` in it twice. To which entity does the error apply?
-
 The general rule is that **compile-time errors apply to semantic definitions
 whenever possible.** In other words, if the library is syntactically well-formed
 enough that augmentations *can* be applied, then they should be. And if doing so
 eliminates what would otherwise be a compile-time error, then that error should
```
lrhn: I worry about this phrasing. Maybe it's just the "then they should be".

There is no choice about "applying augmentations". They exist, they have a semantics, unless an error is reported for something else (which may make their existince questionable).

And applying augmentations will *not* "eliminates what would otherwise be a compile-time error". There is no "otherwise". There either is an error or there is not. We shouldn't hypothesize about potential different programs. Changing any little thing may have more complex repercussions on program semantics than what is intended.
There is only one program. It has only one semantics. We should describe what that is, not what it isnt, or could have been, or would be in a different program.

Consider:
> The general rule is that **compile-time errors apply to semantic definitions
> whenever possible**. For this example it is the _class as a whole_, as defined by all
> its introducing and augmenting declarations, which must have a concrete member
> whose signature is a valid override for its interface member signature.
> Having some individual member declarations which do not introduce an implementation
> does not mean that the class does not introduce an implementation of that member,
> as long as there is _one_ concrete member declaration for the member.

working/augmentations/feature-specification.md
```diff
@@ -1020,48 +1033,44 @@ those entities interchangeably. With augmentations, that is no longer the case.
...
-augmentations shouldn't affect the errors that are reported as much as possible.
-Augmentations are a syntactic tool for organizing code, but what the user cares
-about -- and what static analysis should thus focus on -- is the semantics of
-the resulting definitions. Also, in most cases the error relies on semantic
-information that isn't even well defined for syntactic entities and is only
-known from the resolved semantic definition which can't be produced without
-applying augmentations.
+augmentations shouldn't affect the errors that are reported. Augmentations are
+a syntactic tool for organizing code, but what the user cares about -- and what
+static analysis should thus focus on -- is the semantics of the resulting
```
lrhn, line 1079: and what static analysis should focus on -> and what defines the program's behavior

working/augmentations/feature-specification.md
```diff
@@ -1020,48 +1033,44 @@ those entities interchangeably. With augmentations, that is no longer the case.
...
+definitions. Also, in most cases the error relies on semantic information that
+isn't even well defined for syntactic entities and is only known from the
```
lrhn, line 1081: "well-defined" hyphenated? (Not sure, but reads awkwardly to me without)

working/augmentations/feature-specification.md
```diff
@@ -1276,6 +1285,22 @@ and assume the third point is always true.
...
+
+*   Add support for static abstract variable declarations in the grammar.
+
+*   Adjust the definition of signature matching (using `_` as "don't care"
+    parameter names is no longer supported).
+    
+*   Add a section to say that certain redundant superinterfaces are no longer an
+    error.
+    
+*
```
lrhn: Left-over `*`.

working/augmentations/feature-specification.md
```diff
@@ -1020,48 +1033,44 @@ those entities interchangeably. With augmentations, that is no longer the case.
...
-But what about:
+However, when augmentations are supported it is possible to provide the
+missing implementation in an augmenting declaration:
 
 ```dart
-class C implements I {}
-
-augment class C implements I {}
+augment class C {
+  augment get g => 0;
```
lrhn: (Write the `int` return type. No need to be brief, and if there is any risk it can confuse a reader, it's not worth it.)

working/augmentations/feature-specification.md
lrhn, line 1008: How about `class C extends I with I {}`?
Currently reported as error by analyzer.

A quick check shows:
```dart
mixin class I {}

class C1 implements I, I {} // Error, both.
class C2 with I implements I {}
class C3 with I, I {}
class C4 extends I implements I {} // Error, both.
class C5 extends I with I {} // Error analyzer

mixin M1 implements I, I {} // Error, both.
mixin M2 on I implements I {} // Error CFE ("extends and implements clause")
mixin M3 on I, I {} // Error both

enum E1 implements I, I {x} // Error, both.
enum E2 with I implements I {x}
enum E3 with I, I {x}

extension type X1(I _) implements I, I {} // Error, both.
```
I think the ones they agree on are currently disallowed by the spec:
* `implements I, I` (anywhere)
* `extends I implements I` (`class`)
* `on I, I` (mixin)

The `on I implements I` is CFE only (treats `on`-clause like an `extends` clause according to the error message) and the `extends I with I` is Analyzer only (with a specific error message suggesting it's intended).

I expect all of these to become/stay valid.

Maybe just say that.

Or even just:
> The following errors are removed from the specification 
> in the section "Superinterfaces":
> ```latex
> It is a compile-time error if two elements in the type list of
> the \IMPLEMENTS{} clause of a class $C$ specifies the same > type $T$.
> It is a compile-time error if the superclass of a class $C$ is
> one of the elements of the type list of the \IMPLEMENTS{} clause of $C$.
> ```
> There is no longer any error based solely on the same type
> occurring more than once among the `extends`, `with`,
> `on` or `implements` types of a `class`, `mixin`, `enum` or
> `extension type` declaration.

I don't think any similar error is introduced anywhere else (so technically it should only apply to "classes").

working/augmentations/feature-specification.md
```diff
@@ -1276,6 +1285,22 @@ and assume the third point is always true.
...
 
+### 1.41
+
+*   Adjust the grammar to enforce that some top-level constant and final
+    variable declarations have an initializing expression.
+
+*   Add support for static abstract variable declarations in the grammar.
+
+*   Adjust the definition of signature matching (using `_` as "don't care"
+    parameter names is no longer supported).
```
lrhn, line 1306: (😢 )

working/augmentations/feature-specification.md
eernstg: Good point! I adjusted the text to refer to the target of the comparison as the 'other' signature.

working/augmentations/feature-specification.md
eernstg: Clarified this.

working/augmentations/feature-specification.md
eernstg: That's better, done.

working/augmentations/feature-specification.md
eernstg: (Weird that this wasn't an error, by the way. ;-)

working/augmentations/feature-specification.md
eernstg: (OK, they can fix this now, there will be co19 tests.)

eernstg: I did not mention any rules or changes dealing with mixins applied more than once. That's a valid construct with a specific semantics. I suggested that we should warn or make it an error (who knows which optimizations that might allow), but it hasn't happened yet.

So I'll just continue to avoid mentioned that topic here.

working/augmentations/feature-specification.md
eernstg: "Occurs in a list" still allows for including cases like `int, Map<I, I>`, and we don't have a definition of 'directly', either.

Changed the wording to say 'not a subterm of another type', I think that should work.

eernstg: `operand` clearly refers to a syntactic entity, but I think it's OK to implicitly expect the reader to coerce this into a type, as opposed to insisting that we write something like "the type denoted by the `operand`".

working/augmentations/feature-specification.md
eernstg: It's just the usual difficulty associated with denoting a construct only when it occurs in a specific context. For example, a null-shorting chain needs to be subject to the syntactic transformation which is its specified meaning, but this transformation should only be applied to the outermost expression, even though it may contain any number of subterms which are also null-shorting chains.

I think "isn't a subterm of another type" will work, and that's actually less confused than the null-shorting expressions.

working/augmentations/feature-specification.md
eernstg: Done.

working/augmentations/feature-specification.md
eernstg: Done.

working/augmentations/feature-specification.md
eernstg: I'll keep this one unchanged. With the change it because a duplication: The dynamic semantics _is_ what defines the program's behavior.

working/augmentations/feature-specification.md
eernstg: Google's AI response is that 'well-defined' is a compound adjective that typically occurs immediately before a noun. When it occurs after the associated noun it typically does not have the hyphen.

In this case the associated noun would be 'semantic information'. I think it's OK to omit the hyphen. Also, Bob has worked so intensively with publishing companies that he'd know how to write most things. So I'll leave this one unchanged as well.

working/augmentations/feature-specification.md
eernstg: Fixed.

working/augmentations/feature-specification.md
eernstg: I used your proposed text.

PR #4655, opened 2026-03-06
Add rules about constant primary constructors and initialization

eernstg:
The primary constructor feature specification did not specify the treatment of constant primary constructors with respect to initialization. This PR adds a paragraph of normative text and a commentary paragraph to specify that primary parameters are considered to be potentially constant in initializers (in non-late variable declarations as well as in the initializer list), and it is an error unless said expressions are potentially constant.

If we do not specify this relaxation then the initializing expressions of non-late instance variables must be constant, which would cause the primary parameters to be useless for any class with constant constructors in those initializing expressions.

The corresponding clarification about the initializer list of the body part of a constant primary constructor is needed because we previously left it unspecified, but it should be unsurprising because it just means that we're treating the primary constructor body part the same as a normal in-body constructor.

### DIFF
--- a/accepted/future-releases/primary-constructors/feature-specification.md
+++ b/accepted/future-releases/primary-constructors/feature-specification.md
@@ -4,7 +4,7 @@ Author: Erik Ernst
 
 Status: Accepted
 
-Version: 1.14
+Version: 1.15
 
 Experiment flag: primary-constructors
 
@@ -680,10 +680,14 @@ contains a body part for the primary constructor, and it has an initializer
 list, and the initializer list contains an expression which is not potentially
 constant.
 
-Moreover, for every `<constObjectExpression>` invoking a constructor of _D_, a
-compile-time error occurs if the result of substituting actual arguments of the
-constructor invocation into one of the above mentioned initializing expressions
-or initializer list elements yields an expression which is not constant.
+Moreover, for every constant expression which is an instance creation that
+invokes a constructor of _D_, a compile-time error occurs if the result of
+substituting actual arguments of the constructor invocation into one of the
+above mentioned initializing expressions or initializer list elements yields an
+expression which is not constant.
+
+*This is the same as the existing check on constant constructor invocations, but
+it applies to two new pieces of syntax.*
 
 Consider a class, mixin class, enum, or extension type declaration _D_ with
 a primary constructor *(note that it cannot be a `<mixinApplicationClass>`,
@@ -967,6 +971,11 @@ far removed from any syntactic hint that the constructor must be constant
 of declaration, and the constructor might be non-const).
 
 ### Changelog
+
+1.15 - March 11, 2026
+
+* Generalize potentially constant expressions. Specify compile-time errors
+  associated with constant expressions and initialization expressions.
 
 1.14 - March 4, 2026
 


accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -660,6 +660,25 @@ type has a primary constructor whose name is also the name of a constructor
 declared in the body, or if it declares a primary constructor whose name is
 `C.n`, and the body declares a static member whose basename is `n`.
 
+A compile-time error occurs if a class, mixin class, enum, or extension type
+declaration _D_ has a constant primary constructor, and a non-late instance
+variable declaration in the body of _D_ has an initializing expression which is
+not potentially constant when the primary parameters are considered to be
+potentially constant. An error also occurs if the body of _D_ contains a body
```
lrhn: Not "when ... are considered to be". 
They *are* potentially constant, always.

Potentially constant expressions are defined in the "constants" section of the specification.
They include the parameters of of a constant generative constructor.
We may want to make sure it's explicit about including _all_ the local variables introduced into the parameter scope or initializer list scope by such constructors' parameters, including the variables of initializing formals, declaring parameters and super parameters of a non-redirecting generative constructor.

Just remove from "when" to the end of the sentence. The expression must be potentially constant, full stop. What that means is defined elsewhere.

lrhn: I'd be explicit and say "A compile-time error also occurs ...".

(I occassionally search the spec for "compile-time error". I'd like that to work!)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -660,6 +660,25 @@ type has a primary constructor whose name is also the name of a constructor
...
+part for the primary constructor, and it has an initializer list, and the
```
lrhn: ("body part" is the name for the entire `this:...` block, right? Not the body part of the body part? Not at all  confusing!)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -660,6 +660,25 @@ type has a primary constructor whose name is also the name of a constructor
...
+initializer list contains an element of the form `this.v = e` or `v = e`, and
+`e` is not a potentially constant expression when the primary parameters are
```
lrhn: Again, just drop from "when".

It's also a compile-time error if:
* The initializer list contains an `assert(e1)` or `assert(e1, e2)` and `e1` is not a potentially constant expression
* The initializer list ends with a super constructor invocation, `super(...)` or `super.name(...)`, and any argument expression in the argument list `(...)` is not a potentially constant expression.

(Or if the body part has a body part, but I don't know if that's already said elsewhere, and this is just about potentially constant-ness.)

That said, other than the instance variable initializers, this is just the *same rules* as any other constant initializing constructor.
It would be nice if we could abstract over the syntax and refer to:
* The parameter list of the constructor.
* The parameter and initializer list scopes of the constructor.
* The initializier list entries of the constructor.
* The super-constructor invocation of the constructor.
* The body-part of the constructor.

independently of whether it's a primary constructor or not, so we can just state these rules *once*. And assign semantics only once.
That should reduce the risk of specifying an unitended difference between primary and non-primary initializing constructors.
(The only difference there should be is that primary constructors *can* have declaring parameters, but after they have declared their variable, they should be indistinguishable from initializing formals anyway, and that primary constructors disallow double-initializing, which can be handled very locally where the constructor initializes a variable, with an "if this constructor is a primary constructor, then it's a compile-time error if the instance variable has an initializer expression.")

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -660,6 +660,25 @@ type has a primary constructor whose name is also the name of a constructor
...
+considered to be potentially constant.
+
+*This is a relaxation of the existing rule which says that the initializing
+expressions in variable declarations must be constant when a class contains any
+constant constructors. However, with support for primary constructors it is
+possible to refer to the primary parameters in those initializing expressions,
+and this should be equally expressive as the corresponding declaration that uses
+a non-primary constructor and an initializer list. The rule about initializer
```
lrhn: Could we just say that if a class declaration has a constant generative constructor, then any instance variable declaration must be `final`, and if such an instance variable has an initializer expression, that expression must be a potentially constant expression.

No special rule for primary constructors. It's just that if the constructor isn't primary, there are no potentially constant parameter variables in scope for the initializer expression, so it's _almost_ the same as just having to be constant.

The one difference is that the class's type parameters are in scope, and those are potentially constant types. Still not a lot you can use that for - you can't use them as type arguments, object creation expressions are not potentially constant, and with `const` they can't use potentially constant types. But it's there if you want to store the `Type` object.

And if we don't say that, and still require a constant expression when there is no primary constructor, then  (we have twice as many rules to manage, and) there is a difference between:
```dart
class C<T> {
  const C();
  final type = T;
}
```

```dart
class const C<T>() {
  final type = T;
}
```
and 
```dart
class C<T> {
  C();
  final type = T;
}
```

The last two are allowed, the first is not, even though `T` *is in scope anyway*.
It could just work in all cases.

(Maybe we should say that the default constructor is a primary constructor, so you don't add `ClassName();` to the body, you add `()` to the `class ClassName`. That has no impact on `const` constructors, though, just a stray thought.)

accepted/future-releases/primary-constructors/feature-specification.md
lrhn: This doesn't say that it replaces the existing rule that it's an error if the expression isn't constant. Probably should say that, otherwise that rule would still apply and make a potentially-constant-but-not-constant initializer expression an error.

(It should just replace the existing rule entirely with a rule saying that "if a class has a constant constructor and an instance variable declaration with an initializer expression, it's a compile-time error if that initializer expression is not a potentially constant expression."
This should apply even if the constructor isn't primary. See  below for more arguments.)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -660,6 +660,25 @@ type has a primary constructor whose name is also the name of a constructor
...
+lists just spells out explicitly that they should be treated "the same" as the
+initializer lists of non-primary constructors.*
```
lrhn: We shouldn't try to spell out rules explicitly if they are intended to be "the same".

If they are actually _the same_ (without scare quotes), we should refer to the existing defintion. That avoids quoting the rules incorrectly or missing cases, and if something changes those rules while this feature is in flight, it's clear that this feature should inherit those changes too. If we copied the behavior into the feature spec, then it's much less clear what is intended.

If they aren't actually _the same_, then we are introducing ambiguity by saying "the rules are this _and_ they are the same", and then leaving it to the reader to determine which actual differences are intended, which are bugs, and which are just not considered part of the thing that is "the same".

Or we can introduce a complete generalized rewrite of those rules, so readers don't have to try to guess.

eernstg:
About type variables as potentially constant: I created https://github.com/dart-lang/language/issues/4661 in order to clarify this proposal. Currently the type parameters are not potentially constant.

PR #4663, opened 2026-03-17
Fix #4652: allow augmenting setters to choose parameter name

### LINKED ISSUE #4652: Restore the ability to choose the parameter name of a setter in an augmenting declaration
Thanks to @sgrekhov for bringing up this unwanted change!

We used to allow an augmentation chain to have a parameter name of choice, and then also to use the name `_` for some elements in an augmentation chain. For example:

```dart
void foo(int _);
augment foo(myName) => print(myName.isEven);
```

In this declaration chain, `foo` has a formal positional parameter named `myName` of type `int`. It was OK for the augmentation chain to specify the type in the introductory declaration "without a name", and the augmenting declaration could then choose whatever it wanted.

With https://github.com/dart-lang/language/commit/bafdffc7377bb1aaae1db1c0aa42a2093647da58, it is no longer supported to use `_` as a "whatever you want" name.

This implies that [a different rule](https://github.com/dart-lang/language/blob/main/working/augmentations/feature-specification.md#augmenting-variables-getters-and-setters) changes its meaning in a way that is inconvenient:

> For purposes of augmentation, a variable declaration is treated as implicitly defining a getter whose return type is the type of the variable. If the variable is not `final`, or is `late` without an initializer, then the variable declaration also implicitly defines a setter with a parameter named `_` whose type is the type of the variable.

In particular, the following example shows a thing that we used to be able to do, which is now an error:

```dart
// Introduce an incomplete variable, allowing augmenting
// declarations to choose an implementation.
abstract int i;

// Implement it using a backing variable.
int _i;
augment get i => _i;
augment set i(myName) {
  _i = myName;
}
```

The error occurs because we use `myName` as the name of the formal parameter of the setter, but the rule mentioned above already made the choice that this parameter must have the name `_`.

It is of course possible to change the `abstract int i;` declaration to a separate getter and setter, and choose some other name which will then make it possible to write a working augmentation of the setter, but this feels like a path around an approach that _ought_ to work, namely the approach which is shown above.

- We might choose a fixed value other than `_` as the parameter name (say, `newValue`), but that's somewhat magical.
- We might also choose to say that the introductory declaration uses a magical name for the setter which happens to be exactly the name that is used by the first augmenting declaration of the setter. It is then just as flexible as previously, and we never need to support the actual semantics that `_` used to have.

Darkildo:
This PR updates the augmentations specification so that the implicit setter
induced by a variable declaration no longer has a fixed parameter name `_`.

Previously, the spec said that for augmentation purposes a non-`final` variable
(or a `late` variable without an initializer) implicitly defines a setter with a
parameter named `_`. After the recent changes around `_`, this made it
impossible for an augmenting setter declaration to choose its own parameter
name.

This PR changes that wording so that the implicit setter is specified only in
terms of its shape and type:

- it has a single required positional parameter
- the parameter type is the type of the variable
- the variable declaration does not determine the parameter name

This restores the intended ability for an augmenting setter declaration to use a
source-level parameter name such as:

```dart
augment set i(myName) {
  _i = myName;
}

### DIFF
--- a/working/augmentations/feature-specification.md
+++ b/working/augmentations/feature-specification.md
@@ -839,8 +839,9 @@ It's a **compile-time** error if:
 For purposes of augmentation, a variable declaration is treated as implicitly
 defining a getter whose return type is the type of the variable. If the variable
 is not `final`, or is `late` without an initializer, then the variable
-declaration also implicitly defines a setter with a parameter named `_` whose
-type is the type of the variable.
+declaration also implicitly defines a setter with a single required positional
+parameter whose type is the type of the variable. The variable declaration does
+not determine the name of that parameter.
 
 If the variable is `abstract`, then the getter and setter are incomplete,
 otherwise they are complete. *For non-abstract variables, the compiler

### CODE REVIEW

File: working/augmentations/feature-specification.md [hunk 1]
```diff
@@ -839,8 +839,9 @@ It's a **compile-time** error if:
 For purposes of augmentation, a variable declaration is treated as implicitly
 defining a getter whose return type is the type of the variable. If the variable
 is not `final`, or is `late` without an initializer, then the variable
-declaration also implicitly defines a setter with a parameter named `_` whose
-type is the type of the variable.
+declaration also implicitly defines a setter with a single required positional
+parameter whose type is the type of the variable. The variable declaration does
+not determine the name of that parameter.
```
eernstg, line 844: I think you may have been OOO at the time, but the ability to use a parameter declaration whose name is specified as `_` to mean "I don't care" was reintroduced with https://github.com/dart-lang/language/pull/4669 and https://github.com/dart-lang/language/pull/4670. It should not be necessary to introduce the notion of "does not determine the name" for a parameter declaration.

I don't think it actually creates any problems with the specification to rely on "does not determine the name", but it seems simpler to rely on the ability of `_` to do the same thing, now that this feature has been restored.

eernstg (approved):
LGTM, but I do think it would be slightly better to not do this. ;-)

Issue #932, opened 2020-04-17
Make it an error to `await voidExpression`


eernstg:
For the release of Dart 2, it turned out to be a more seriously breaking change than anticipated to make it an error to `await e` in the case where `e` has static type `void`.

Shouldn't we fix this now where null-safety is being introduced?

```dart
void foo() => null;

void main() async {
  await foo(); // Error, cannot await an expression of type `void`.
}
```

@leafpetersen, @lrhn, @munificent, WDYT?

leafpetersen:
It's fine by me.  If @lrhn agrees with this change, we'd need a volunteer to spec it, write tests and file implementation issues.

lrhn:
I'm always fine with disallowing use of `void`!

eernstg:
Spec update in https://github.com/dart-lang/language/pull/933.

eernstg:
Tests added in https://dart-review.googlesource.com/c/sdk/+/144140.

eernstg:
Created issues for the analyzer and CFE implementation.

leafpetersen:
Tests and spec have landed, issues filed.  I'm going to close this out.

PR #4668, opened 2026-03-23
[specification] Remove await operand from void allowlist.

stereotype441:
Operands of await expressions haven't been allowed to be `void` since the introduction of null safety. See
https://github.com/dart-lang/language/issues/932.

---

<details>

Many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.

**Note**: The Dart team is trialing Gemini Code Assist. Don't take its comments as final Dart team feedback. Use the suggestions if they're helpful; otherwise, wait for a human reviewer.

</details>

### DIFF
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -22928,14 +22928,6 @@ In the rules, every type under consideration is a static type.
     which in turn restricts where it can occur.%
   }
 \item
-  In an expression of the form \code{\AWAIT\,\,$e$}, $e$ may have type \VOID.
-  \rationale{%
-    This rule was adopted because it was a substantial breaking change
-    to turn this situation into an error
-    at the time where the treatment of \VOID{} was changed.
-    Tools may choose to give a hint in such cases.%
-  }
-\item
   \commentary{%
     In a return statement \code{\RETURN\,$e$;},
     $e$ may have type \VOID{} in a number of situations


stereotype441:
> This is actually already done in #2605, which is the PR that contains all the updates needed in order to integrate null safety into the language specification ([the change occurs here](https://github.com/dart-lang/language/pull/2605/changes#diff-0e2f05538272f3fb682a361d7738d860125088af3ac0044e5472cac965cfdb61L22925)).
> 
> I think we should proceed with the current approach for this kind of change: Make sure #2605 is as complete and correct as possible, and then landing #2605 subsets such that dartLangSpec.tex will gradually be updated with all of null safety.
> 
> I think it gets harder to understand what's going on if we update single details directly in dartLangSpec.tex.

I definitely don't want to get in the way of your process. At the same time, I'm a bit concerned because #2605 has been open for over 3 years; is there a good plan in place for getting #2605 finished in a reasonable amount of time? I lost a nontrivial amount of time last week due to confusion about whether await operands were supposed to be allowed to be void or not, and I really enjoy being able to take the opportunity to "scratch my own itch" by making helpful updates when the spec confuses me. (If you're curious, the reason I ran into this is that I'm refactoring how the front end and analyzer handle await expressions, in preparation for fixing https://github.com/dart-lang/sdk/issues/62889, so I wanted to double check the spec to make sure I wasn't making a mistake).

I'm also a bit confused because it seems like this PR matches up perfectly with your plan of landing #2605 subsets to gradually bring dartLangSpec.text into line with null safety--the change I'm making here precisely matches the diff hunk you linked, so this PR is in fact a subset of #2605.

To be clear, I'm not trying to be difficult here; I'm just trying to understand how this PR differs from how you'd like to proceed, so that I can modify it to be compatible with your plans.

eernstg:
> I lost a nontrivial amount of time last week due to confusion about whether await operands were supposed to be allowed to be void or not,

Oh, that's unfortunate!

However, it is also exactly the reason why we should get #2605 landed.

> is there a good plan in place for getting https://github.com/dart-lang/language/pull/2605 finished in a reasonable amount of time?

It has already been unreasonable for quite a while, there's a reason why the branch has the name `specify_null_safety_sep21`. ;-) 

> it seems like this PR matches up perfectly with your plan of landing https://github.com/dart-lang/language/pull/2605 subsets to gradually bring dartLangSpec.text into line with null safety

You could say so, but I've been aiming at landing it as 10 subsets rather than 250...

You might be right that it is actually more realistic to get many small updates landed, but it does also give rise to some potential difficulties. In particular, it may be difficult to keep track of the consistency of the document if it contains a large number of tiny updates at a highly fragmented set of locations. It isn't trivial to determine whether any given piece of text in the document is pre- or post-nnbd, and any attempt to make the whole document consistent might semi-update a pre-nnbd rule or it might semi-regress a post-nnbd rule.

But it's great that you're pushing on this topic, I'd really love to get the spec to a more recent level!

eernstg:
Just FYI, here's the latest slice: https://github.com/dart-lang/language/pull/3515.

stereotype441:
> > I lost a nontrivial amount of time last week due to confusion about whether await operands were supposed to be allowed to be void or not,
> 
> Oh, that's unfortunate!
> 
> However, it is also exactly the reason why we should get #2605 landed.
> 
> > is there a good plan in place for getting #2605 finished in a reasonable amount of time?
> 
> It has already been unreasonable for quite a while, there's a reason why the branch has the name `specify_null_safety_sep21`. ;-)
> 
> > it seems like this PR matches up perfectly with your plan of landing #2605 subsets to gradually bring dartLangSpec.text into line with null safety
> 
> You could say so, but I've been aiming at landing it as 10 subsets rather than 250...

Yeah, I can definitely appreciate the desire to land #2605 in bigger chunks. In the analyzer and front end code bases where I do most of my work, we often don't have that luxury, and it can be hard to keep track of what's completed and what's still left to implement. And we have the benefit of automated tests, which the spec doesn't.

Another argument for trying to do this in bigger chunks is that since I'm refactoring void handling in the analyzer and front end right now, it wouldn't be surprising for other issues like this one to arise in the next few days. It could save a lot of time to batch up a bunch of changes to the void allowlist into a single PR for easier review.

> You might be right that it is actually more realistic to get many small updates landed, but it does also give rise to some potential difficulties. In particular, it may be difficult to keep track of the consistency of the document if it contains a large number of tiny updates at a highly fragmented set of locations. It isn't trivial to determine whether any given piece of text in the document is pre- or post-nnbd, and any attempt to make the whole document consistent might semi-update a pre-nnbd rule or it might semi-regress a post-nnbd rule.
> 
> But it's great that you're pushing on this topic, I'd really love to get the spec to a more recent level!

Yeah, me too. And I'd love to help if I can!

Considering that this is where I'm working right now, how about if I took responsibility for the "void allowlist" portion of #2605? I would do the following:

- Compare the void allowlist spec (both in its current state and with #2605 applied) against the current analyzer and front end implementations, considering all parts of the grammar in which an expression can appear (not just `await`).

- Review all other mentions of `void` in the spec to identify areas where changing the void allowlist might impact the consistency of the document.

- Bring an informal proposal to you and the rest of the language team of what I believe needs to be changed to bring the analyzer, front end, and the spec into alignment. This might include both spec changes and implementation changes.

  - If I'm going to recommend any implementation changes that are breaking, this would include prototyping those breaking changes in google3 to measure the impact.

  - This would also include a proposal for what to do about https://github.com/dart-lang/sdk/issues/60861.

- Once we've decided what to do, make changes to the analyzer and front end, and make a PR that updates the spec accordingly, for you to review.

What do you think?

eernstg:
OK, let's try to go in the direction you describe! It would be great if we can do it in a way that would just make the rebased version of #2605 smaller (that is, if possible, we should select updates from #2605 rather than writing completely new text). If the update should be different from #2605 then I'm not quite sure how we'd get a smooth rebase, but it would be good if it doesn't differ in cases where it doesn't have to differ.

stereotype441:
> OK, let's try to go in the direction you describe! It would be great if we can do it in a way that would just make the rebased version of #2605 smaller (that is, if possible, we should select updates from #2605 rather than writing completely new text). If the update should be different from #2605 then I'm not quite sure how we'd get a smooth rebase, but it would be good if it doesn't differ in cases where it doesn't have to differ.

Sounds good. I'm going to go ahead and close this PR for now. I'll reach out when I have more information to share.

PR #4669, opened 2026-03-26
Restore the "do not care" semantics of _ as a parameter name

eernstg:
In version 1.37 of this feature specification, a formal parameter whose declaration used a wildcard (`_`) as the name could be used to avoid mentioning the name of this parameter in the given declaration. If all declarations of this parameter in the augmentation chain uses `_` then the parameter is a wildcard. Otherwise, if at least one declaration uses a different name `n` then the parameter name is `n`. Finally, a compile-time error occurs if two or more distint non-`_` names are specified. This PR restores that rule, using the same text as in version 1.37.

Resolves https://github.com/dart-lang/language/pull/4652.

### DIFF
--- a/working/augmentations/feature-specification.md
+++ b/working/augmentations/feature-specification.md
@@ -753,7 +753,7 @@ signature *matches* an introductory signature if:
 
     *   They both have the modifier `covariant`, or none of them have it.
 
-    *   They both have the modifier `required`. or none of them have it.
+    *   They both have the modifier `required`, or none of them have it.
 
     *For constructors, we do not require parameters to match in uses of
     initializing formals or super parameters. In fact, they are implicitly
@@ -766,15 +766,15 @@ signature *matches* an introductory signature if:
 
 *   For all positional parameters:
 
-    *   The augmenting function's parameter name is `_`, or
+    *   The name of the augmenting parameter declaration is `_`, or
 
-    *   The augmenting function's parameter name is the same as the name of the
-        corresponding positional parameter in every preceding declaration that
-        doesn't have `_` as its name.
+    *   The name of the augmenting parameter declaration is the same as the name
+        of the corresponding parameter declaration in every preceding
+        declaration that doesn't have `_` as its name.
 
     *In other words, a declaration can ignore a positional parameter's name by
-    using `_`, but all declarations in the chain that specify a name have to
-    agree on it.
+    using `_`, but all declarations in the chain that specify a name which is
+    not `_` must agree on it.*
 
     ```dart
     f1(int _) {} // OK, this declaration doesn't care about the name.
@@ -784,12 +784,10 @@ signature *matches* an introductory signature if:
     augment f1(int _); // OK.
     ```
 
-    *Note that this is a transitive property.*
-
-    *If an augmentation uses `_` for a parameter name, the name is not
-    "inherited" from a preceding declaration for use in the augmentation's
-    body. The name of the parameter for that augmentation is `_`, which can't be
-    used because it's a wildcard:
+    *If an augmentation specifies `_` as the name of a parameter, a non-`_` name
+    is not "inherited" from a preceding declaration for use in the
+    augmentation's body. The name of the parameter for that augmentation is `_`,
+    which can't be accessed because it's a wildcard:
 
     ```dart
     f(int x);
@@ -806,7 +804,9 @@ name `n` is not in scope and may be resolved elsewhere. *For example:*
 
 ```dart
 int y = 42;
+
 void g(int y);
+
 augment g(_) {
   print(y); // OK, prints '42'.
 }


working/augmentations/feature-specification.md
```diff
@@ -764,6 +764,38 @@ signature *matches* an introductory signature if:
...
+    *If an augmentation uses `_` for a parameter name, the name is not
+    "inherited" from a preceding declaration for use in the augmentation's
+    body. The name of the parameter for that augmentation is `_`, which can't be
+    used because it's a wildcard:
+
+    ```dart
+    f(int x);
+    augment f(int _) { print(x); } // Error.
+    ```
+
```
sgrekhov, line 796: May be it makes sense to add one more example explaining the below?
```dart
int x = 42;

void foo(int _) {
  print(x);
}

augment foo(int x); // What foo prints now?
```

eernstg: Good point! Done.

eernstg:
Added a paragraph about scoping around parameters declared as `_` and named elsewhere.

working/augmentations/feature-specification.md
```diff
@@ -764,11 +764,54 @@ signature *matches* an introductory signature if:
...
+    *   The augmenting function's parameter name is `_`, or
+
+    *   The augmenting function's parameter name is the same as the name of the
+        corresponding positional parameter in every preceding declaration that
+        doesn't have `_` as its name.
+
+    *In other words, a declaration can ignore a positional parameter's name by
+    using `_`, but all declarations in the chain that specify a name have to
+    agree on it.
```
lrhn: Missing trailing `*`.

working/augmentations/feature-specification.md
lrhn, line 749: Need to update here? 

> * They have the same name, if both have a name. _This is trivially satisfied for named
    parameters. Positional parameters may have different names, and may have no name._

lrhn: "." -> ","

working/augmentations/feature-specification.md
```diff
@@ -764,11 +764,54 @@ signature *matches* an introductory signature if:
...
+    augment f1(int y); // Error, can't change the name.
+    augment f1(int _); // OK.
+    ```
+
+    *Note that this is a transitive property.*
+
+    *If an augmentation uses `_` for a parameter name, the name is not
+    "inherited" from a preceding declaration for use in the augmentation's
+    body. The name of the parameter for that augmentation is `_`, which can't be
+    used because it's a wildcard:
```
lrhn: (Should we decide, once and for all, whether the parameter's has the name `_`, which does not get added to any scope, or if it has _no_ name? I'm leaning towards the latter, so I get wary when `_` referred to as a "name".
The "can't be used because it's a wildcard" is doubly confusing - it's not that you can't use _the name_ (assuming there is one) - you can write `_`, it just won't denote the parameter.
So ... feels inconsistent. But I'm not sure we are generally consistent about this.)

working/augmentations/feature-specification.md
```diff
@@ -764,11 +764,54 @@ signature *matches* an introductory signature if:
     or super parameters and all other declarations for the same constructor
     must declare the corresponding parameters as regular parameters.*
 
+*   For all positional parameters:
+
+    *   The augmenting function's parameter name is `_`, or
```
lrhn: augmenting function -> augmenting function declaration
?

Or is that a given, since there are no "augmenting function" that isn't a declaration?

working/augmentations/feature-specification.md
eernstg: Thanks, done!

working/augmentations/feature-specification.md
eernstg: Right, this was actually present in version 1.37, but it should just be removed: Named parameters have the same name because that's how they get paired up, and names of positional parameters are handled below. So this item should be removed.

eernstg: Good catch! This wasn't changed, but we can fix it now.

working/augmentations/feature-specification.md
eernstg: Agreed, I had the same uneasy feeling about "the name `_`". But that's a somewhat bigger issue (in multiple documents, I think).

I do think there's a good amount of consistency in the perspective where the formal parameter declaration has a specific location where the name of the parameter is specified, and this can be `_`, and in that case the name of the parameter is not added to any scopes.

This works, e.g., in cases where a wildcarded parameter would have shadowed a more global declaration (say, a top-level variable), and it also means that we don't have to invent a different explanation when it comes to declarations named `_` which are not local variables (including formal parameters).

If we say that the parameter does not have a name at all then we need to apply this principle to local variables, but we have to make sure that we mention other declarations, because they _do_ have a name which is `_` if that's what we have written.

Let's change this in a more global manner if needed. I've adjusted this document to use the "doesn't get added to any scopes" perspective for now.

working/augmentations/feature-specification.md
eernstg: I changed that to talk about preceding/introductory/augmenting formal parameter declarations, which is something that we are already doing in some other places.

working/augmentations/feature-specification.md
lrhn: But it wasn't removed?

eernstg: Thanks for the heads-up, I don't know how that happened. Done in https://github.com/dart-lang/language/pull/4670.

PR #4670, opened 2026-03-30
Perform a change which was part of PR 4669, but omitted by mistake

eernstg:
This is an update that should have been performed as part of #4669, but was omitted by mistake: The requirement that parameters (in general) in an augmentation chain should have the same name is obsolete; this requirement is satisfied by construction with named parameters, and it is superseded by new rules for positional parameters. Hence, this PR removes that requirement.


eernstg:
@lrhn, friendly ping.

Issue #4672, opened 2026-03-30
Downwards inference of yield* is underspecified


stereotype441:
How should we expect this code to be analyzed?
```dart
dynamic f() sync* {
  yield* g()..foo();
}

T g<T>() => throw 'T=$T';

main() {
  for (var x in f()) {
    print('x=$x');
  }
}
```

Currently, the analyzer and CFE both accepts this code without issuing any compile-time errors. The runtime behavior is to print:
```
Unhandled exception:
T=dynamic
#0      g (file:///home/paulberry/tmp/proj/test.dart:5:13)
#1      f (file:///home/paulberry/tmp/proj/test.dart:2:10)
#2      _SyncStarIterator.moveNext (dart:async-patch/async_patch.dart:595:13)
#3      main (file:///home/paulberry/tmp/proj/test.dart:8:17)
#4      _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:313:19)
#5      _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:192:12)
```

This indicates that the context type being passed to the operand of `yield*` is either the unknown type (`_`) or `dynamic`. Reading through the implementation, I'm pretty certain it's the unknown type.

But [resources/type-system/inference.md](https://github.com/dart-lang/language/blob/main/resources/type-system/inference.md) says:

> - If the enclosing function is marked `sync*`, then for each `yield* e;` statement in the block, let `S` be the inferred type of `e`, using the local type inference algorithm described below with a typing context of `Iterable<K>`; let `E` be the type such that `Iterable<E>` is a super-interface of `S`; and update `T` to be `UP(E, T)`.

Which suggests that the context should be `Iterable<K>` for some `K`.

`K` is defined to be "the typing context for the function body as computed above from the imposed return type schema." It's not entirely clear what this refers to, but the definition of the "imposed return type schema" is as follows:

> The return type of the context function type is used at several points during inference.  We refer to this type as the **imposed return type schema**. Inference for each returned or yielded expression in the body of the function literal is done using a context type derived from the imposed return type schema `S` as follows:
> - If the function expression is neither `async` nor a generator, then the context type is `S`.
> - If the function expression is declared `async*` and `S` is of the form `Stream<S1>` for some `S1`, then the context type is `S1`.
> - If the function expression is declared `sync*` and `S` is of the form `Iterable<S1>` for some `S1`, then the context type is `S1`.
[...]

PR #4673, opened 2026-03-31
Adjust the function literal return type inference section to handle "weird" types

eernstg:
Some background discussion can be found in https://github.com/dart-lang/language/issues/4672.

```dart
void main() {
  dynamic Function() f = () sync* { yield e; };
}
```

The [rules](https://github.com/dart-lang/language/blob/main/resources/type-system/inference.md#function-literal-return-type-inference) in `inference.md` do not have a matching case for the computation of the context type for `e` because the imposed return type schema is `dynamic`, which is a type that isn't of the form `Iterable<S1>` for any type `S1`. Given that the `sync*` case doesn't match, we proceed to the last case 'Otherwise, ...', but in this case the context type schema is specified as `FutureOr<futureValueTypeSchema(S)>`, which is clearly only intended to be used when the function body is marked `async`.

This PR inserts a couple of sanity checks such that we will recognize return types that are not of the form `Iterable<...>` or `Stream<...>` for generator functions, and we will only use the `FutureOr<futureValueTypeSchema(S)>` rule when the function is marked `async`, and we will use the context type schema `_` when the imposed return type schema is completely off track (e.g., when a `sync*` function literal has context type of `int Function()` even though it could never return an `int`). The new wording relies on the notion of 'the element type' of a generator function, which is specified in the language specification.


resources/type-system/inference.md
```diff
@@ -283,17 +291,24 @@ having type `dynamic`.
...
-  - Otherwise, without null safety, the context type is `FutureOr<flatten(T)>`
-    where `T` is the imposed return type schema; with null safety, the context
-    type is `FutureOr<futureValueTypeSchema(S)>`.
+    context type schema is `S`.
+  - If the function expression is declared `async*` and the element type 
+    schema of `S` as the return type schema of an asynchronous function is
+    `S1`, then the context type schema is `S1`.
+  - If the function expression is declared `sync*` and the element type 
+    schema of `S` as the return type schema of a synchronous function is
+    `S1`, then the context type schema is `S1`.
```
stereotype441: This feels imprecise to me, because the spec’s notion of "element type of a generator function" is a mapping from a generator _function_ to a _type_, based on the function’s type. Whereas what we need here is a mapping from one type schema to another, and we need to do the mapping at a time when the function’s type hasn’t been determined yet.

How about if we extend the spec’s notion of "a type S implements a type T" (`interfaceSuperinterfaces` section of the spec) so that it applies to type schemas as well as types, and then change these two bullets to simply:

- If the function expression is declared `async*`, and `S` implements `Stream<U>` for some `U`, then the context type schema is `U`.
- If the function expression is declared `sync*`, and `S` implements `Iterable<U>` for some `U`, then the context type schema is `U`.

resources/type-system/inference.md
```diff
@@ -283,17 +291,24 @@ having type `dynamic`.
 The return type of the context function type is used at several points during
 inference.  We refer to this type as the **imposed return type
```
stereotype441: As I mentioned in https://github.com/dart-lang/language/issues/4672#issuecomment-4164265666, I would love to find a way to unify the specification of how we handle function literals more with how we handle function declarations, to match how the implementations unify them.

Looking in particular at the analyzer (because I believe it's closer to matching the existing spec), the analyzer defines the notion of "imposed return type schema" for both function expressions and function declarations (though it calls it `BodyInferenceContext.imposedType`). Its definition will look like this (after applying modifications I propose in https://github.com/dart-lang/language/issues/4672#issuecomment-4164265666):
- For a top level function, method, getter, setter, constructor, or local function declaration, let `R` be its return type. Then the imposed return type schema is:
  - `_` if `R` is `dynamic`,
  - Otherwise, `R`.
- For a function expression, let `S` be the context in which the function expression is type inferred. Then the imposed return type schema is:
  - `_` if `S` is not a function type schema, or if `S` is a function type schema with a return type schema of `dynamic`.
  - Otherwise, `R`, where `R` is the return type schema of `S`.

I think we should make the spec match this.

stereotype441 (commented):
I like the direction this is heading. In addition to my comments below, there's one other thing that I would want to address before landing this. Since this is a potentially breaking change, it would be nice to test it out in google3 and go through the [breaking change process](https://github.com/dart-lang/sdk/blob/main/docs/process/breaking-changes.md) before landing it. Once we agree on what spec text we're aiming for, I'd be happy to do a prototype implementation and run it through Google3 to make sure it doesn't cause unexpected breakages.

resources/type-system/inference.md
eernstg: Agreed. I adjusted the language specification to define 'the (a)synchronous element type of a type'. This document then generalizes it to handle type schemas, and uses this notion in the rules in line 299-304.

resources/type-system/inference.md
eernstg: Did that. Note that I changed the `futureValueTypeSchema` to map `dynamic` to `_`, in order to maintain consistency with the corresponding rules about non-async functions.

eernstg:
Made various adjustments, PTAL!

For readers who haven't followed the process in detail: Please note that this PR introduces some breaking changes, and it will not be landed unless and until a corresponding breaking change process has been conducted to the extent needed.

resources/type-system/inference.md
```diff
@@ -104,7 +110,7 @@ and which this proposal is designed to satisfy:
   violates this principle.
 * The inference for local variables, top-level variables, and fields, should
   either agree or error out.  The same expression should not be inferred
-  differently at different syntactic positions.  It’s ok for an expression to be
+  differently at different syntactic positions.  It's ok for an expression to be
```
eernstg, line 113: I had to change the apostrophe from '\x80\x99' to a plain single quote because Github (presumably because of some update today) suddenly insisted that this file is binary and cannot be shown. It returned to normal rendering when I made this change.

resources/type-system/inference.md
```diff
@@ -260,62 +266,103 @@ Note that `late` fields are inferred exactly as non-`late` fields.  However,
...
-  - If the function expression is declared `sync*` and `S` is of the form
-    `Iterable<S1>` for some `S1`, then the context type is `S1`.
-  - Otherwise, without null safety, the context type is `FutureOr<flatten(T)>`
-    where `T` is the imposed return type schema; with null safety, the context
-    type is `FutureOr<futureValueTypeSchema(S)>`.
+Assume that _D_ is a top-level, static, or instance declaration of an
+asynchronous non-generator function, method, or getter; or it is a local
+asynchronous non-generator function declaration; assume that the return
+type of _D_ is `R`. The imposed return type schema of _D_ is
+`FutureOr<futureValueTypeSchema(S)>`.
```
stereotype441, line 289: I think you mean `FutureOr<futureValueTypeSchema(R)>`.

resources/type-system/inference.md
```diff
@@ -260,62 +266,103 @@ Note that `late` fields are inferred exactly as non-`late` fields.  However,
...
 
 - **futureValueTypeSchema**(`S?`) = **futureValueTypeSchema**(`S`), for all `S`.
-- **futureValueTypeSchema**(`S*`) = **futureValueTypeSchema**(`S`), for all `S`.
-- **futureValueTypeSchema**(`Future<S>`) = `S`, for all `S`.
-- **futureValueTypeSchema**(`FutureOr<S>`) = `S`, for all `S`.
+- **futureValueTypeSchema**(`Future<S>`) = **eraseDynamic**(`S`), for all `S`.
+- **futureValueTypeSchema**(`FutureOr<S>`) = **eraseDynamic**(`S`), for all `S`.
 - **futureValueTypeSchema**(`void`) = `void`.
-- **futureValueTypeSchema**(`dynamic`) = `dynamic`.
+- **futureValueTypeSchema**(`dynamic`) = **eraseDynamic**(`dynamic`).
```
stereotype441, line 297: Why not save the reader the trouble of evaluating **eraseDynamic**(`dynamic`) and just say this?
- **futureValueTypeSchema**(`dynamic`) = `_`.

resources/type-system/inference.md
```diff
@@ -260,62 +266,103 @@ Note that `late` fields are inferred exactly as non-`late` fields.  However,
...
+
+Assume that _D_ is a non-generator function literal which is being inferred
+with context type schema `T`.
+
+If `T` is not a function type schema, or `T` is a function type schema with
+return type `dynamic`, the imposed return type schema of _D_ is `_`.
+Otherwise, let `S` be the return type schema of `T`. *Note that `S` may
+contain references to type variables declared by _D_ itself, which is not a
+problem.* In this case the imposed return type schema of _D_ is `S`.
+
```
stereotype441, line 327: I don't see any definition of _imposed return type schema_ for a generator function literal. I've addressed this in my comment on line 310.

resources/type-system/inference.md
```diff
@@ -260,62 +266,103 @@ Note that `late` fields are inferred exactly as non-`late` fields.  However,
...
+Otherwise, the imposed return type schema of _D_ is `S`.
+
+Assume that _D_ is a non-generator function literal which is being inferred
+with context type schema `T`.
+
+If `T` is not a function type schema, or `T` is a function type schema with
+return type `dynamic`, the imposed return type schema of _D_ is `_`.
+Otherwise, let `S` be the return type schema of `T`. *Note that `S` may
+contain references to type variables declared by _D_ itself, which is not a
+problem.* In this case the imposed return type schema of _D_ is `S`.
```
stereotype441, line 326: The paragraph above doesn't say anything about asynchronous vs synchronous, so I'm assuming that _this_ paragraph is intended to apply to both asynchronous and synchronous non-generator function literals. In the case of asynchronous non-generator function literals, shouldn't the imposed return type schema be `FutureOr<futureValueTypeSchema(S)>`?

I've addressed this in my comment on line 310.

resources/type-system/inference.md
```diff
@@ -260,62 +266,103 @@ Note that `late` fields are inferred exactly as non-`late` fields.  However,
...
-Inference for each returned expression in the body of the function literal is
-done in an empty typing context (see below).
+*This section assumes that Dart has null safety (older versions of the
+language are ignored).*
 
-Function literals which are inferred in an non-empty typing context where the
-context type is a function type are inferred as described below.
+In the following we refer to the asynchronous or synchronous element type
+schema of a type. This is defined from the asynchronous respectively
+synchronous element type of a type by treating `_` as a type.
```
stereotype441, line 276: Two concerns here:
- The definition of synchronous/asynchronous element type behaves differently depending on whether the type in question is a supertype of `Object`. When extending this definition to schemas, should we treat `_` as a type that extends `Object` or a type that doesn't extend `Object`?
- Synchronous/asynchronous element type of `T` is undefined in the case where `T` neither implements `Iterable<S>` (respectively `Stream<S>`) for some `S`, nor is a supertype of `Object`. In your edit to the spec, when you use asynchronous/synchronous element type to define element type of a generator function, you've explained that the undefined case only occurs if there's a compile-time error. But I believe that during type inference of a function literal, that undefined case _is_ possible to reach (because there are situations where it's ok for the static type of an expression to fail to satisfy the expression's schema). So we need to define how we handle that case.

resources/type-system/inference.md
```diff
@@ -260,62 +266,103 @@ Note that `late` fields are inferred exactly as non-`late` fields.  However,
...
 _Note that it is a compile-time error unless the return type of an asynchronous
 non-generator function is a supertype of `Future<Never>`, which means that
 the last case will only be applied when `S` is `Object` or a top type._
 
+The helper function **eraseDynamic** is defined as follows:
+
+- **eraseDynamic**(`dynamic`) = `_`.
+- Otherwise, for all `S`, **eraseDynamic**(`S`) = `S`.
+
+Assume that _D_ is a top-level, static, or instance declaration of a
```
stereotype441, line 310: The order is a bit confusing here, because we've interrupted the list of cases that define _imposed return type schema_ in order to define **futureValueTypeSchema** and **eraseDynamic**, and now we're returning to add more cases to _imposed return type schema_. I'd recommend defining each term in its own section, with bullets for cases, e.g.:

Assume that _D_ is a top-level, static, or instance declaration of a function, method, getter, setter, or constructor; or it is a local function declaration or a function literal. Then the _imposed return type schema_ of _D_ is defined as follows:
- Define `R` as follows:
  - If _D_ is a function literal which is being inferred with context type schema `T`:
    - If `T` is a function type schema, let `R` be the return type of `T`.
    - Otherwise let `R` be `_`.
  - Otherwise, let `R` be the return type of _D_. *Note that `R` may contain references to type variables declared by _D_ itself, which is not a problem.*
- If _D_ is a synchronous non-generator, the imposed return type schema of _D_ is **eraseDynamic**(`R`).
- If _D_ is an asynchronous non-generator, the imposed return type schema of _D_ is `FutureOr<futureValueTypeSchema(R)>`.
- If _D_ is a synchronous generator, the imposed return type schema of _D_ is **eraseDynamic**(`S`), where `S` is the synchronous element type of `R`.
- If _D_ is an asynchronous generator, the imposed return type schema of _D_ is **eraseDynamic**(`S`), where `S` is the asynchronous element type of `R`.

The function **futureValueTypeSchema** is defined as follows:

...

The helper function **eraseDynamic** is defined as follows:

...

stereotype441 (commented):
Returning to this after a long time away (sorry!)

I have some suggestions about how to re-order things, and there are some cases which are not covered.

Once we get that taken care of, I still want to test the change in google3 to verify that it's not too breaking.

resources/type-system/inference.md
lrhn: Presumably it was `\xe2\x80\x99`, which is UTF-8 for U+2019, `’`.

resources/type-system/inference.md
```diff
@@ -260,62 +266,103 @@ Note that `late` fields are inferred exactly as non-`late` fields.  However,
...
-## Function literal return type inference.
+## The imposed return type schema
 
-Function literals which are inferred in an empty typing context (see below) are
-inferred using the declared type for all of their parameters.  If a parameter
-has no declared type, it is treated as if it was declared with type `dynamic`.
-Inference for each returned expression in the body of the function literal is
-done in an empty typing context (see below).
+*This section assumes that Dart has null safety (older versions of the
+language are ignored).*
```
lrhn, line 272: That's no longer necessary to say. Those older versions are no longer supported at all.

resources/type-system/inference.md
```diff
@@ -260,62 +266,103 @@ Note that `late` fields are inferred exactly as non-`late` fields.  However,
...
-Each parameter is assumed to have its declared type if present.  If no type is
-declared for a parameter and there is a corresponding parameter in the context
-type schema with type schema `K`, the parameter is given an inferred type `T`
-where `T` is derived from `K` as follows.  If the greatest closure of `K` is `S`
-and `S` is a subtype of `Null`, then without null safety `T` is `dynamic`, and
-with null safety `T` is `Object?`. Otherwise, `T` is `S`. If there is no
-corresponding parameter in the context type schema, the variable is treated as
-having type `dynamic`.
+Assume that _D_ is a top-level, static, or instance declaration of a
+synchronous non-generator function, method, getter, setter, or constructor;
```
lrhn, line 279: (What's the difference between a "function" and a "method"?)

resources/type-system/inference.md
```diff
@@ -260,62 +266,103 @@ Note that `late` fields are inferred exactly as non-`late` fields.  However,
...
+or it is a local synchronous non-generator function declaration; assume
+that the return type of _D_ is `R`. If `R` is `dynamic`, the _imposed
+return type schema_ of _D_ is `_`.  Otherwise, the imposed return type
+schema of _D_ is `R`.
```
lrhn, line 283: I would use a different name for different kinds of functions.
Instead of "imposed return type scheme", I'd have "return type" for a sync-non-generator, "future value type" for an async-non-generator, "iterable element type" for a sync-generator and "stream element type" for an async-generator.

Then refer to the relevant one at `return` or `yield[*]` statements.

That reduces the risk of accidentally mixing up things, or assuming something locally that isn't true globally.
And it avoids calling the `yield`-type a "return type", which it most certianly isn't.
(It's derived _from_ a return type, but it's an _element type_.)

Which means this section isn't necessary.
At a `return` statement, "if the current function is not an async function, the context type is the return type. if the current function is an async function, the context type is `FutureOr<T>` where `T` is the future value type of the return type of the current function."

Similarly for `yield`s.

resources/type-system/inference.md
lrhn: _Why_ make this change?

If it makes any difference whether the future-value-type-schema is `_` or `dynamic`, then actually writing `dynamic` as the return type may suggest that the user wanted the `dynamic` type.

resources/type-system/inference.md
```diff
@@ -260,62 +266,103 @@ Note that `late` fields are inferred exactly as non-`late` fields.  However,
...
-- **futureValueTypeSchema**(`S*`) = **futureValueTypeSchema**(`S`), for all `S`.
-- **futureValueTypeSchema**(`Future<S>`) = `S`, for all `S`.
-- **futureValueTypeSchema**(`FutureOr<S>`) = `S`, for all `S`.
+- **futureValueTypeSchema**(`Future<S>`) = **eraseDynamic**(`S`), for all `S`.
+- **futureValueTypeSchema**(`FutureOr<S>`) = **eraseDynamic**(`S`), for all `S`.
 - **futureValueTypeSchema**(`void`) = `void`.
-- **futureValueTypeSchema**(`dynamic`) = `dynamic`.
+- **futureValueTypeSchema**(`dynamic`) = **eraseDynamic**(`dynamic`).
 - **futureValueTypeSchema**(`_`) = `_`.
 - Otherwise, for all `S`, **futureValueTypeSchema**(`S`) = `Object?`.
```
lrhn, line 299: Why not make that `_`?
I'd argue that `dynamic foo() async { ... }` is more suggestive that the user actually want `dynamic` semantics, and `Object foo() async {...}` feels more like they don't care, and `_` would be more relevant.

But in any case, I still don't see why we want to introduce `_` at all.

resources/type-system/inference.md
lrhn: (Also shows the incongruence going from "synchronous element type", which is very precise and understandable, to "imposed return type", which has nothing to do with returns.)

resources/type-system/inference.md
```diff
@@ -260,62 +266,103 @@ Note that `late` fields are inferred exactly as non-`late` fields.  However,
...
+- Otherwise, for all `S`, **eraseDynamic**(`S`) = `S`.
+
+Assume that _D_ is a top-level, static, or instance declaration of a
+synchronous respectively asynchronous generator function, method, or
+getter; or it is a local synchronous respectively asynchronous generator
+function declaration; assume that the return type of _D_ is `R`.
+
+Let `S` be the synchronous respectively asynchronous element type of `R`.
+If `S` is `dynamic`, the imposed return type schema of _D_ is `_`.
+Otherwise, the imposed return type schema of _D_ is `S`.
```
lrhn, line 317: Why is that not just _eraseDynamic_(`S`)?

resources/type-system/inference.md
```diff
@@ -260,62 +266,103 @@ Note that `late` fields are inferred exactly as non-`late` fields.  However,
...
+with context type schema `T`.
+
+If `T` is not a function type schema, or `T` is a function type schema with
+return type `dynamic`, the imposed return type schema of _D_ is `_`.
+Otherwise, let `S` be the return type schema of `T`. *Note that `S` may
+contain references to type variables declared by _D_ itself, which is not a
+problem.* In this case the imposed return type schema of _D_ is `S`.
+
+Any imposed return type schema can be designated as the _imposed return
+type_ in a situation where it is a type.
```
lrhn, line 329: Can be ... will it? And what if it isn't?

resources/type-system/inference.md
```diff
@@ -260,62 +266,103 @@ Note that `late` fields are inferred exactly as non-`late` fields.  However,
...
+body of the function literal is done in an empty typing context (see
+below).
+
+Function literals which are inferred in a non-empty typing context where the
+context type is not a function type are inferred in the empty typing context.
+
+Function literals which are inferred in an non-empty typing context where the
+context type is a function type are inferred as described below.
+
+Each parameter is assumed to have its declared type if present.  If no type is
```
lrhn, line 347: "if present" -> "if it has a declared type". It's not clear whether it's the declared type or the parameter which must be present.

resources/type-system/inference.md
```diff
@@ -260,62 +266,103 @@ Note that `late` fields are inferred exactly as non-`late` fields.  However,
...
+declared for a parameter and there is a corresponding parameter in the context
+type schema with type schema `K`, the parameter is given an inferred type `T`
+where `T` is derived from `K` as follows.  If the greatest closure of `K` is `S`
+and `S` is a subtype of `Null`, then without null safety `T` is `dynamic`, and
+with null safety `T` is `Object?`. Otherwise, `T` is `S`. If there is no
+corresponding parameter in the context type schema, the variable is treated as
+having type `dynamic`.
```
lrhn, line 354: What if the function is generic, is that covered?

PR #4688, opened 2026-05-07
Add introductory section plus specification of abbreviated constructor declarations

eernstg:
This PR adds a short introductory section about the abbreviated constructor syntax which is part of the primary constructor feature bundle, and it specifies the meaning of the abbreviated constructor declarations. For example:

```dart
class MyClass(final int i) {
  new(int j) : this(j + 1); // Rather than `MyClass(int j) ...`.
}
```

### DIFF
--- a/accepted/future-releases/primary-constructors/feature-specification.md
+++ b/accepted/future-releases/primary-constructors/feature-specification.md
@@ -462,7 +462,7 @@ declare constructors that work exactly the same:
 
 ```dart
 class MyClass {
-  const new ();
+  const new();
   new name();
   new redir(): this.name();
   factory fact() => .new();
@@ -472,10 +472,6 @@ class MyClass {
 
 In short, the class name and the period are replaced my the keyword `new`
 (in a generative constructor) or simply removed (in a factory constructor).
-
-*As a matter of formatting, note that the keyword `new` and the formal
-parameter list `()` are separated by a space to indicate that `new` is a
-keyword rather than an identifer. Similarly for `factory`.*
 
 ## Specification
 
--- a/accepted/future-releases/primary-constructors/feature-specification.md
+++ b/accepted/future-releases/primary-constructors/feature-specification.md
@@ -440,11 +440,10 @@ initialize `w` with some other value, in which case we must also initialize
 
 ### Abbreviations of in-body constructor declarations
 
-This feature includes an element which is technically independent of
-primary constructors, but it is related in that it also allows for
-declaring constructors more concisely, except that this is concerned with
-regular (non-primary) constructors in the body of the enclosing
-declaration.
+This feature includes a subfeature which is technically independent of
+primary constructors, but it is related in that it also allows for more
+concise constructor declarations. This subfeature is concerned with regular
+(non-primary) constructors in the body of the enclosing declaration.
 
 Here are some examples of today's constructor declaration syntax:
 
@@ -454,7 +453,7 @@ class MyClass {
   MyClass.name();
   MyClass.redir(): this.name();
   factory MyClass.fact() => .new();
-  factory MyClass.redirFact() = MyClass;
+  const factory MyClass.redirFact() = MyClass;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -438,6 +438,45 @@ could add another non-redirecting generative constructor which could
...
+  const factory redirFact() = MyClass;
+}
+```
+
+In short, the class name and the period are replaced my the keyword `new`
+(in a generative constructor) or simply removed (in a factory constructor).
+
+*As a matter of formatting, note that the keyword `new` and the formal
+parameter list `()` are separated by a space to indicate that `new` is a
+keyword rather than an identifer. Similarly for `factory`.*
```
lrhn: (Did we agree on this formatting?
I'd expect to write `factory()` and `new()` without space for the "unnamed" constrructor.
The `const new ()` format is _very_ confusion. In no other place is there a parameter list with a space before it, so having a space makes it look like it's not a parameter list.)

accepted/future-releases/primary-constructors/feature-specification.md
```diff
@@ -588,6 +627,23 @@ constructors as well.
...
+
+A constructor declaration containing tokens `new id` derived from
+`<constructorHead>` *(i.e., `id` is an `<identifier>` or absent)* in a
+membered, type introducing declaration named `C` has the name `C.id` when
+`id` is present, and `C` when it is absent.
+
+Similarly, a constructor declaration containing tokens `factory id` derived
+from `<factoryConstructorHead>` in a membered, type introducing declaration
+named `C` has the name `C.id` when `id` is present, and `C` when it is
+absent.
```
lrhn, line 641: Worth saying explicitly that `new id()` and `factory id()` have the same name, and will conflict if both written in the same declaration?

lrhn (approved):

accepted/future-releases/primary-constructors/feature-specification.md
munificent: I believe the formatting should be `new()` with no space and likewise `factory()` with no space. That's what I implemented in the formatter.

This is definitely a subjective choice and there are valid arguments either way. My rationale is a combination of:

* It's shorter. One less space.

* In no other place do we put a space before parameter lists. Not even operators where the thing before `(` isn't an arbitrary user-defined identifier.

* In other languages that use a reserved word or special identifier (as opposed to the class name) to define constructors, they don't use a space before `(`: [Scala with `this(...)`](https://docs.scala-lang.org/scala3/book/domain-modeling-tools.html#auxiliary-constructors), [Kotlin with `constructor(...)`](https://kotlinlang.org/docs/classes.html#secondary-constructors), [JavaScript with `constructor(...)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/constructor), [Swift with `init(...)`](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/initialization/), [Python with `__init__(...)`](https://typing.python.org/en/latest/spec/constructors.html), [Ruby with `initialize(...)`](https://www.geeksforgeeks.org/ruby/ruby-constructors/).

I understand that it's conventional to have a space between a reserved word and `(` in other places in the language, but I think those are all confined to statements (and elements). We don't put a space between a keyword and `(` in:

* Redirecting constructor calls like `this(...)`.
* Super constructor calls like `super(...)`.
* Assert statements and initializers. (I would argue that we probably *should* have a space here.)
* Function types (assuming you consider `Function` something like a keyword).

munificent (approved):
Aside from a style nit about the space before `(`, LGTM!

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: I think we discussed it, and it is also the approach taken by `dart format`.

eernstg: Just tried it again, and `dart format` of version 3.12.0-221.0.dev does include the space between `new`/`factory` and `(`.

eernstg:
I built the tools from the HEAD of SDK, and `dart format` suddenly does not include the space in `new ();` and `factory ();`. That's surprising because I thought we had discussed this, and the behavior of the formatter confirmed that assumption. Anyway, I removed the space and the commentary about why it was there.

accepted/future-releases/primary-constructors/feature-specification.md
eernstg: We could mention it, but I think it is unsurprising based on the syntax used at the call site.

PR #4690, opened 2026-05-13
Update augmentation specification with primary constructors.

lrhn:
Defines interaction between primary constructors and augmentations as:
* Each declaration of a constructor can choose between being primary or in-body.
* If any declaration is primary, that constructor is primary for the class.
* If a class has a primary constructor, it cannot have other initializing constructors, and
* Instance variable initializers can refer to constructor parameters (if the current fragment (re-)declares the primary constructor signature, includin that parameter name). Initializers are resolved in the same scope in either case, but it's an error if the name isn't visibly in scope.

Clean up "before"/"after" in the presence of declarations with no names, like the new constructor syntax `factory(x) => ...`.

Remove incorrect restriction on declaring enum members named `toString`, `hashCode` and `==`.
You always could declare those, as long as they were abstract.

(I changed that to saying that the declarations are introduced by the introductory `enum` as complete declarations in the enum. That means you you _must_ put `augment` in front of abstract declarations. I think it's an improvement, and definitely better than the blanket restriction, but it's still more restrictive than current Dart, and we may want to decide if that's what we want. I think it is.)

### DIFF
--- a/working/augmentations/feature-specification.md
+++ b/working/augmentations/feature-specification.md
@@ -254,8 +254,8 @@ When there are no values, the enum still requires a leading `;` before the first
 member to avoid ambiguity.
 
 ```
-enumType ::= 'augment'? 'enum' classNameMaybePrimary typeParameters?
-    mixins? interfaces? enumBody?
+enumType ::=
+    'augment'? 'enum' classNameMaybePrimary mixins? interfaces? enumBody
 
 enumBody ::=
     '{' (enumEntry (',' enumEntry)* (',')?)? (';' memberDeclarations)? '}'
@@ -290,53 +290,27 @@ always uses the same `on` clause as the introductory declaration.
 Both `class` and `enum` declarations can use the primary constructor syntax
 for declaring an initializing _(non-redirecting generative)_ constructor.
 
-The primary constructor syntax does not allow an extra `augment` to be added
-to the class header. As an exception, a primary constructor declaration does
-not need an extra `augment`, on top of the one on the class or enum declaration
-itself, if it declares an augmenting constructor using the primary constructor
-syntax, and it does not have an in-body part. If the declaration _does_ haven
-an in-body part (`this;` or `this: ...`), that in-body part must be prefixed by
-`augment` if the constructor declaration is augmenting.
-
-A primary constructor declaration introduces or augments an initializing
-constructor, and makes it so that that constructor is a primary constructor
-for the entire class or enum.
-
-If a complete primary constructor has a declaring parameter, that parameter
-declaration introduces a complete introducing or augmenting variable
-declaration, which is `final` if the declaring parameter has the `final`
-modifier, and is implicitly augmenting if there is a prior declaration of
-the same getter and/or setter. *(The `augment` modifier does not apply
-to parameters, so the declaring parameter cannot be marked as augmenting
-a variable.)*
-
-It's a **compile-time error** if a class or enum has two distinct
-initializing constructors, and either of them is declared as a
-primary constructor.
-*If the class or enum has a primary constructor, it cannot have any other
-initializing constructors, and therefore no other primary constructors,
-since those are always initializing.*
-That also implies that it is a compile-time error if a class or enum has two
-distinct constructors which are both declared as primary.
-
-The declarations of a class or enum can contain both primary and non-primary
-declarations of the same constructor. The class has a primary constructor
-if at least one declaration uses the primary constructor syntax.
+See [Generative constructors](#generative_constructor_declarations).
 
 ### Instance variable initializer expressions
 
 If a class or enum has a primary constructor, then the scope of non-`late`
-instance variable initializer expressions are the initializer list scope
+instance variable initializer expressions is the initializer list scope
 of that constructor.
 
 It's a compile-time error if a non-`late` instance variable initializer
-expressions refers to a varible introduced by the constructor's
-initializer list scope if the surrounding class/enum declaration does
+expressions refers to a variable introduced by the constructor's
+initializer list scope, and the surrounding class or enum declaration does
 _not_ have a primary constructor declaration which declares the
 corresponding parameter's name.
+_A primary constructor must declare all parameters, but it can omit declaring
+a positional parameter's name by using `_` instead of the name.
+If it does so, that parameter's name cannot be used by instance variable
+initializers of that class or enum declaration._
+
 _The scope exists and is used whether this particular declaration writes
-a primary constructor or not, but for readability reasons, code can only
-refer to veriables of that scope if there appears to be a declaration of
+a primary constructor or not, but for readability reasons, code may only
+refer to variables of that scope, if there appears to be a declaration of
 the variable's name in the surrounding code._
 
 ## Static semantics
@@ -615,24 +589,24 @@ files.
 
 Some terminology:
 
-*   The *source code* of a declaration is the span of character from the
-    first non-whitspace character of the declaration to the last non-whitspace
+*   The *source code* of a declaration is the span of characters from the
+    first non-whitespace character of the declaration to the last non-whitespace
     character of the declaration, as matched by the relevant grammar production.
     *Metadata is not considered part of the declaration, it precedes the
     declaration in the source.*
 
-*   A syntactic declaration *occurs in* a Dart file if the declaration's source
+*   A syntactic declaration, *D*, *occurs in* a Dart file *F* if *D*'s source
     code occurs in that file.
 
-*   A Dart file *includes* a part file if the Dart file has a `part` directive
-    with a URI denoting another part file.
+*   A Dart file, *F*, *includes* a part file *P* if the Dart file *F*
+    has a `part` directive whose URI denotes *P*.
 
-*   A Dart file *contains* (aka. transitively includes) a part file
-    if the Dart file includes the part file, or if the Dart file includes
-    another part file, and that other part file contains the part file.
+*   A Dart file, *F*, *contains* (aka. transitively includes) a part file *P*
+    if *F* (directly) *includes* *P*, or if *F* *includes* a part file *Q*
+    and *Q* *contains* *P*.
 
-*   A Dart file *contains* a declaration if the declaration occurs in the file
-    itself or in any of the part files that the Dart file contains.
+*   A Dart file *contains* a declaration *D* if the declaration *D* *occurs in*
+    the file *F* itself, or if *D* occurs in a file *P* and *F* *contains* *P*.
 
 For any two distinct syntactic declarations *A* and *B*:
 
@@ -652,7 +626,8 @@ For any two distinct syntactic declarations *A* and *B*:
             the position is the start position of that identifier
             in the source.
             *(Includes an unnamed local variable declared with a `_` as
-            identifier.)*
+            identifier. A primary constructor declaration's position is
+            that of the class name.)*
         *   If the declaration is an unnamed `extension` declaration,
             its position is the position of the `on` keyword.
 
@@ -1039,7 +1014,7 @@ For ordering purposes, these implicit declarations are _before_ any members
 declared in the declaration.
 
 *Any declaration of the same members must be augmenting (they are not
-introdoctory) and must not be complete.*
+introductory) and must not be complete.*
 *Declaring an instance member named `values` will conflict with the
 static `values` declaration as a normal scope name conflict.*
 
@@ -1048,130 +1023,125 @@ It's a **compile-time error** if:
 *   An enum doesn't have any values after all augmentations are applied. *The
     grammar allows an enum declaration to not have any values so that other
     declarations of the same enum can add them, but ultimately the enum must
-    end up with some.*
+    end up with at least one value.*
 
 ### Augmenting constructors
 
-Augmenting constructors works similar to augmenting a function, with some extra
-rules to handle features unique to constructors like redirections and
-initializer lists, and special handling of primary constructors.
+Augmenting a constructor works similarly to augmenting a function, with some
+extra rules to handle features unique to constructors, like redirections and
+initializer lists, and the primary constructor syntax.
 
-It is **not** a compile-time error for an incomplete factory constructor to
+
+<!-- It is **not** a compile-time error for an incomplete factory constructor to
 omit default values. *That is, they are treated similarly to abstract
 instance methods in this respect. This allows the augmenting declaration to
-implement the constructor by adding a redirection or a body.*
+implement the constructor by adding a redirection or a body.* -->
 
-It's a **compile-time error** if:
+A constructor declaration is a _factory constructor declaration_ if it
+has a `factory` keyword, and it is a _generative constructor declaration_ if
+it does not, including when it is a primary constructor declaration.
 
-*   The signature of the augmenting constructor does not [match][signature
-    matching] the signature of the corresponding introductory constructor.
+A constructor declaration is a _const constructor declaration_ if, and only if,
+it has a `const` keyword.
+_For a primary constructor, the `const` keyword goes before the class name
+in the header._
 
-*   A `class` or `enum` constructor is declared as a primary constructor,
-    using primary constructor syntax for at least one declaration of that
-    constructor, and the class or enum has any other initializing constructor.
+#### Factory constructor declarations
 
-*   More than one declaration in the augmentation chain specifies a default
-    value for the same optional parameter. This is an error even in the
-    case where all of them are identical. *Default values are defined by
-    the introductory declaration or an augmentation, but at most once.*
+A factory constructor declaration is a _non-redirecting factory constructor
+declaration_, and is a _complete_ declaration, if it has a body (`{...}`
+or `=> ...;`) or an `external` keyword.
 
-*   The augmentation chain has exactly one specification of a default value
-    for an optional parameter, and the constructor is a redirecting factory.
+A factory constructor declaration is a _redirecting factory constructor
+declaration_ and is _complete_ if it has a redirection clause (`= TargetType;`).
 
-*   No declaration in the augmentation chain specifies a default value for
-    an optional parameter whose declared type is potentially non-nullable,
-    and the constructor is not a redirecting factory.
+_A factory constructor declaration with no body and no redirection clause,
+ended by a single `;`, is not a complete declaration, and does not decide
+whether the factory constructor being defined will is redirecting or not.
+Only the (single) complete declaration forces that decision._
 
-*   The introductory constructor is `const` and the augmenting constructor
-    is not, or vice versa. *An augmentation can't change whether or not a
-    constructor is const because that affects whether users are allowed to use
-    the constructor in a const context.*
+#### Generative constructor declarations.
 
-*   A redirecting generative constructor augments an initializing generative
-    constructor.
+A **primary constructor** declaration causes the constructor it defines
+be a primary constructor, and the class to have a primary constructor.
+_This means the class cannot have any other initializing constructors,
+which is reflected by the consistency rules below, and it changes the
+scope used for instance variable initializer expressions._
 
-*   An initializing generative constructor augments a redirecting generative
-    constructor.
+_A primary constructor declaration always has an in-header part which
+contains a parameter list, and optionally a `const` and a constructor-
+name identifier, (`class const C.id(int x)`). It may also have an
+in-body `this`-part which can contain an initializer list or body,
+and can have metadata attached (`this: super(x) { print('done'); }`)._
 
-*   The introductory constructor is marked `factory` and the augmenting
-    constructor is not, or vice versa. *An augmentation can't change whether or
-    not a constructor is generative because that affects whether users are
-    allowed to call the constructor in a subclass's initializer list.*
+A primary constructor declaration is an _initializing constructor declaration_.
+A primary constructor declaration is _complete_ if (any of):
+  * It has an initializing formal parameter (`this.name`).
+  * It has a super parameter (`super.name`).
+  * It has a declaring parameter (`final int name`).
+  * It has an in-body `this`-part that:
+      * has a body (`{...}`) and/or
+      * has an initializer list (`: ...`).
+A primary constructor declaration is an _augmenting declaration_ if
+(either of):
+  * It has an in-body `this`-part with an `augment` keyword
+    (`augment this ...`).
+  * It has no in-body `this`-part, and there is another declaration of
+    the same constructor which occurs _before_ this constructor.
+Otherwise it is an _introductory declaration_.
+_An augmenting primary constructor with no in-body `this`-part does not
+have a separate `augment` keyword. Such a constructor is necessarily part
+of an augmenting class or enum declaration,
+and its header-part is likely on the same line as the `augment` keyword
+of that declaration. This is considered sufficient, rather than forcing
+a the author of a class declaration like `augment class C(final int x);`
+to add a body and in-body `this`-part just to write an `augment` on it,
+or alternatively to require a second `augment` keyword in the header as
+`augment class augment C(final int x);`._
 
-*   A factory constructor has no complete declaration.
+If a complete primary constructor has a declaring parameter, that parameter
+declaration also counts as a complete introductory or augmenting variable
+declaration, which is `final` if the declaring parameter has the `final`
+modifier, and is augmenting if there is a prior declaration of
+the same getter and/or setter. *(The `augment` modifier does not apply
+to parameters, so the declaring parameter cannot be marked as augmenting
+a variable.)*
 
-    *A factory constructor which has `;` instead of a body or redirection,
-    is incomplete. This allows, and requires, an augmentation to fill in an
-    implementation.
-    A valid non-redirecting generative constructor can have no body,
-    also written as `;`. Thus it is not an error for a _generative_
-    constructor to be incomplete after all augmentation are applied. It is
-    simply an initializing constructor with no body.*
+An in-body generative constructor declaration is an _initializing constructor
+declaration_ and is _complete_ if (any of):
+  * It has an `external` keyword.
+  * It has an initializing formal parameter (`this.name`).
+  * It has a super parameter (`super.name`).
+  * It has a body (`{...}` instead of).
+  * It has an initializer list (`: ...`).
 
-#### Factory constructors
-A factory constructor has a `factory` modifier in its declaration.
+*(An in-body initializing constructor cannot have declaring parameters,
+those are only available to primary constructor declarations.
+Had they been allowed, they'd also make the constructor complete.)*
 
-A factory constructor declaration is incomplete if it has no body,
-and is ended by a `;`.
+An in-body generative constructor declaration is a _redirecting generative
+constructor declaration_ and is _complete_ if it has a redirection clause
+(`: this(args);` or (`: this.name(args);`).
 
-A factory contructor is complete if it has a redirection, `= ConstructorName;`,
-in which case it's a redirecting factory constructor,
-or it has a function body, `=> expression` or `{...}`, in which case
-it's a non-redirecting factory constructor.
+*The declarations of a class or enum can contain both primary and non-primary
+declarations of the same constructor. The class has a primary constructor
+if at least one declaration uses the primary constructor syntax.*
 
-It's a **compile-time error** if:
-
-* All declarations of a factory constructor are incomplete.
-
-#### Generative constructors
-
-A generative constructor is _complete_ and _redirecting_ if contains
-a redirection clause, like `: this.name(...);`.
-
-A generative constructor is _complete_ and _initializing_ if it contains
-any of:
-
-* An `external` modifier.
-
-* An initializing formal parameter (`this.name`).
-
-* A `super` parameter (`super.name`).
-
-* A declaring parameter (`final int name`), only valid in a primary constructor.
-
-* An initializer list (like `: param = expr, ...` or `: super(args)`).
-
-* A body (does not end in `;`).
-
-_This applies both to primary constructor declarations and
-non-primary constructor declarations, where the primary constructor
-declaration declares its initializer list and body in an in-body
-constructor part, like `this: ... { ... }`. An in-body constructor
-part of `this;` does not make the declaration complete._
-
-A generative constructor declaration is _incomplete_ if it is not
-_complete_.
-
-A generative constructor declaration is _redirecting_ if it's
-a complete redirecting declaration, or if it augments a declaration
-which is redirecting.
-
-A generative constructor declaration is _initializing_ (not redirecting)
-if it's a primary constructor declaration, a complete and initializing
-declartion, or if it augments a declaration which is initializing.
-
-_The introductory declaration may not decide whether a generative
-constructor is initializing or redirecting. It's decided by the
-first declaration which is complete or which is a primary constructor
-(which must be initializing), and all later augmentations must not
-contradict that decision._
+*An in-body generative constructor declaration with no special parameters,
+initializer list, body or redirection, like `ClassName()`, does not decide
+whether the constructor is redirecting or initializing. That is decided by
+any declaration which is complete, or which is a primary constructor
+since those must be initializing.*
 
 If all _declarations_ of a generative constructor are incomplete,
-then the constructor is an _initializing_ constructor with the
-implementation defined by the introductory declaration.
+*(and therefore contains no redirecting generative constructor declarations,
+since those are all complete)*,
+then the constructor being defined is an _initializing_ constructor
+with the signature of the introductory declaration, the default values
+introduced by any declaration, no initializer list and no body.
 *(A declaration of `C();` is defined as incomplete, but is also
-a valid concrete implementation of a trivial constructor.)*
-
+historically a valid concrete implementation of a trivial constructor,
+and this ensures that it keeps working that way.)*
 
 Example:
 ```dart
@@ -1187,7 +1157,8 @@ augment class C {
   factory augment C.fact() = C.other;
 }
 ```
-and with a primary constructor:
+
+Example with a primary constructor:
 ```dart
 class const D(int x) {
   /// Creates a `D`.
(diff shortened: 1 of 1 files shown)


lrhn:
Merged to head, PTAL.

working/augmentations/feature-specification.md
```diff
@@ -243,8 +254,8 @@ When there are no values, the enum still requires a leading `;` before the first
 member to avoid ambiguity.
 
 ```
-enumType ::= 'augment'? 'enum' typeIdentifier typeParameters?
-    mixins? interfaces? enumBody
+enumType ::= 'augment'? 'enum' classNameMaybePrimary typeParameters?
```
eernstg: `typeParameters` are now part of `classNameMaybePrimary` so they shouldn't be included one more time:
```suggestion
enumType ::= 'augment'? 'enum' classNameMaybePrimary
```

working/augmentations/feature-specification.md
```diff
@@ -243,8 +254,8 @@ When there are no values, the enum still requires a leading `;` before the first
...
+    mixins? interfaces? enumBody?
```
eernstg: We shouldn't allow `enumBody` to be omitted (this would enable things like `enum E`, it should at least be `enum E;`):
```suggestion
    mixins? interfaces? enumBody
```

working/augmentations/feature-specification.md
```diff
@@ -274,6 +285,60 @@ Just make two separate unnamed extensions.*
...
+Both `class` and `enum` declarations can use the primary constructor syntax
+for declaring an initializing _(non-redirecting generative)_ constructor.
+
+The primary constructor syntax does not allow an extra `augment` to be added
+to the class header. As an exception, a primary constructor declaration does
+not need an extra `augment`, on top of the one on the class or enum declaration
+itself, if it declares an augmenting constructor using the primary constructor
+syntax, and it does not have an in-body part. If the declaration _does_ haven
+an in-body part (`this;` or `this: ...`), that in-body part must be prefixed by
+`augment` if the constructor declaration is augmenting.
```
eernstg: I think it would be helpful for the typical reader if the text specifies the rules directly, without more or less explicit references to design alternatives (like "why don't we just add that extra `augment` to the grammar?"). Perhaps:

```suggestion
As an exception, a primary constructor declaration _D_ does not include the
modifier `augment` in the class header, even in the case where it is an
augmenting declaration. _D_ is an augmenting declaration if and only if there
is a declaration of a constructor with the same name in the same augmentation
context that comes before _D_. The body part of the primary constructor
*(e.g., `this;`)* must include the modifier `augment` if and only if _D_ is an
augmenting declaration *(that is, `augment this;`)*.
```

working/augmentations/feature-specification.md
```diff
@@ -274,6 +285,60 @@ Just make two separate unnamed extensions.*
...
+
+A primary constructor declaration introduces or augments an initializing
+constructor, and makes it so that that constructor is a primary constructor
```
eernstg:
```suggestion
constructor, and causes that constructor to be a primary constructor
```

working/augmentations/feature-specification.md
```diff
@@ -274,6 +285,60 @@ Just make two separate unnamed extensions.*
...
+for the entire class or enum.
+
+If a complete primary constructor has a declaring parameter, that parameter
+declaration introduces a complete introducing or augmenting variable
+declaration, which is `final` if the declaring parameter has the `final`
+modifier, and is implicitly augmenting if there is a prior declaration of
```
eernstg: The source code does not contain the implicitly induced instance variable declaration corresponding to a given declaring parameter, so it doesn't make sense to say that any particular part of this variable declaration is implicit:
```suggestion
modifier, and is augmenting if there is a prior declaration of
```

working/augmentations/feature-specification.md
```diff
@@ -274,6 +285,60 @@ Just make two separate unnamed extensions.*
...
+the same getter and/or setter. *(The `augment` modifier does not apply
+to parameters, so the declaring parameter cannot be marked as augmenting
+a variable.)*
+
+It's a **compile-time error** if a class or enum has two distinct
+initializing constructors, and either of them is declared as a
```
eernstg: We should include the case where both are declared as primary (with different names):
```suggestion
initializing constructors, and at least one of them is declared as a
```

working/augmentations/feature-specification.md
```diff
@@ -274,6 +285,60 @@ Just make two separate unnamed extensions.*
...
+primary constructor.
+*If the class or enum has a primary constructor, it cannot have any other
+initializing constructors, and therefore no other primary constructors,
+since those are always initializing.*
+That also implies that it is a compile-time error if a class or enum has two
+distinct constructors which are both declared as primary.
```
eernstg: This would be redundant:
```suggestion
```

working/augmentations/feature-specification.md
```diff
@@ -274,6 +285,60 @@ Just make two separate unnamed extensions.*
...
+
+The declarations of a class or enum can contain both primary and non-primary
+declarations of the same constructor. The class has a primary constructor
+if at least one declaration uses the primary constructor syntax.
```
eernstg: The last sentence is already covered in line 301-302:
```suggestion
declarations of the same constructor.
```

working/augmentations/feature-specification.md
```diff
@@ -274,6 +285,60 @@ Just make two separate unnamed extensions.*
...
+
+### Instance variable initializer expressions
+
+If a class or enum has a primary constructor, then the scope of non-`late`
+instance variable initializer expressions are the initializer list scope
```
eernstg:
```suggestion
instance variable initializer expressions is the initializer list scope
```

working/augmentations/feature-specification.md
```diff
@@ -274,6 +285,60 @@ Just make two separate unnamed extensions.*
...
+of that constructor.
+
+It's a compile-time error if a non-`late` instance variable initializer
+expressions refers to a varible introduced by the constructor's
```
eernstg:
```suggestion
expression refers to a variable introduced by the constructor's
```

working/augmentations/feature-specification.md
```diff
@@ -274,6 +285,60 @@ Just make two separate unnamed extensions.*
...
+initializer list scope if the surrounding class/enum declaration does
```
eernstg:
```suggestion
initializer list scope, and the surrounding class or enum declaration does
```

working/augmentations/feature-specification.md
```diff
@@ -274,6 +285,60 @@ Just make two separate unnamed extensions.*
...
+_not_ have a primary constructor declaration which declares the
+corresponding parameter's name.
+_The scope exists and is used whether this particular declaration writes
+a primary constructor or not, but for readability reasons, code can only
+refer to veriables of that scope if there appears to be a declaration of
```
eernstg:
```suggestion
refer to variables of that scope if there is a declaration of
```
A semantic entity like a constructor is introduced by an augmentation chain whose elements are constructor declarations. Each of those would declare a formal parameter list which would in turn contain formal parameter declarations. So they "are" in the surrounding code and we don't have to worry about the meaning of "appearing to be".

working/augmentations/feature-specification.md
```diff
@@ -550,42 +615,80 @@ files.
 
 Some terminology:
 
+*   The *source code* of a declaration is the span of character from the
```
eernstg:
```suggestion
*   The *source code* of a declaration is the span of characters from the
```

working/augmentations/feature-specification.md
```diff
@@ -550,42 +615,80 @@ files.
...
+    first non-whitspace character of the declaration to the last non-whitspace
```
eernstg:
```suggestion
    first non-whitespace character of the declaration to the last non-whitespace
```

working/augmentations/feature-specification.md
```diff
@@ -550,42 +615,80 @@ files.
...
+    character of the declaration, as matched by the relevant grammar production.
+    *Metadata is not considered part of the declaration, it precedes the
+    declaration in the source.*
```
eernstg, line 604: Is this a useful rule? Does the language as such have an opinion about this at all?

working/augmentations/feature-specification.md
```diff
@@ -550,42 +615,80 @@ files.
...
+
 *   A syntactic declaration *occurs in* a Dart file if the declaration's source
     code occurs in that file.
 
-*   A Dart file *includes* a part file, if the Dart file has a `part` directive
-    with a URI denoting that part file.
+*   A Dart file *includes* a part file if the Dart file has a `part` directive
+    with a URI denoting another part file.
```
eernstg:
```suggestion
*   A Dart file _F_ *includes* a part file _P_ if _F_ has a `part` directive
    with a URI denoting _P_.
```

working/augmentations/feature-specification.md
```diff
@@ -550,42 +615,80 @@ files.
...
+
+*   A Dart file *contains* (aka. transitively includes) a part file
+    if the Dart file includes the part file, or if the Dart file includes
+    another part file, and that other part file contains the part file.
```
eernstg:
```suggestion
*   A Dart file _F_ *contains* (aka. transitively includes) a part file _P_
    if _F_ includes _P_, or if _F_ includes
    another part file _P2_, and _P2_ contains _P_.
```

working/augmentations/feature-specification.md
```diff
@@ -924,13 +1027,23 @@ enum values. Enum values are appended in augmentation application order.
...
-*   A declaration inside an augmenting enum declaration has the name `values`,
-    `index`, `hashCode`, or `==`. *It has always been an error for an enum
-    declaration to declare a member named `index`, `hashCode`, `==`, or
-    `values`, and this rule just clarifies that this error is applicable for
-    augmenting declarations as well.*
+For ordering purposes, these implicit declarations are _before_ any members
+declared in the declaration.
+
+*Any declaration of the same members must be augmenting (they are not
+introdoctory) and must not be complete.*
```
eernstg:
```suggestion
introductory) and must not be complete.*
```

working/augmentations/feature-specification.md
```diff
@@ -924,13 +1027,23 @@ enum values. Enum values are appended in augmentation application order.
...
 variables and constant variables can't be augmented.
 
-It's a **compile-time error** if:
+An introductory `enum` declaration introduces implicit introductory and
+complete declarations of:
+ * `String toString()`
+ * `int get index`
+ * `int get hashCode`
+ * `bool operator ==(Object)`
+ * `static const List<E> values;` where `E` is the enum type.
```
eernstg, line 1018: Nice!

working/augmentations/feature-specification.md
```diff
@@ -941,7 +1054,7 @@ It's a **compile-time error** if:
 
 Augmenting constructors works similar to augmenting a function, with some extra
```
eernstg: Might as well clarify this even though it hasn't been edited:

```suggestion
Augmenting a constructor works similarly to augmenting a function, with some extra
```

working/augmentations/feature-specification.md
```diff
@@ -941,7 +1054,7 @@ It's a **compile-time error** if:
...
 rules to handle features unique to constructors like redirections and
-initializer lists.
+initializer lists, and special handling of primary constructors.
```
eernstg: With the addition we can make it a list:
```suggestion
rules to handle features unique to constructors like redirections,
initializer lists, and the primary constructor syntax.
```

working/augmentations/feature-specification.md
```diff
@@ -970,23 +1087,93 @@ It's a **compile-time error** if:
     constructor is const because that affects whether users are allowed to use
     the constructor in a const context.*
 
+*   A redirecting generative constructor augments an initializing generative
+    constructor.
+
+*   An initializing generative constructor augments a redirecting generative
+    constructor.
```
eernstg: I don't think we can use this wording (lines 1090-1094).

The problem is that being an initializing generative constructor is a property of the (semantic) constructor, not a property of each of the constructor declarations that make up the augmentation chain that induces the constructor. In particular, `new();` is not initializing, we have intentionally allowed it to become a redirecting or a non-redirecting constructor, based on augmenting declarations.

The new thing here is that a primary constructor is initializing, unambiguously, and it can be incomplete.

So we could say something like this:

*   A redirecting generative constructor declaration occurs in a primary constructor.

This relies on the characterization of a primary constructor as a constructor whose augmentation chain contains at least one primary constructor declaration. It covers both directions.

working/augmentations/feature-specification.md
```diff
@@ -970,23 +1087,93 @@ It's a **compile-time error** if:
...
-*   A factory constructor is incomplete after all augmentations are applied.
+*   A factory constructor has no complete declaration.
+
+    *A factory constructor which has `;` instead of a body or redirection,
+    is incomplete. This allows, and requires, an augmentation to fill in an
+    implementation.
+    A valid non-redirecting generative constructor can have no body,
+    also written as `;`. Thus it is not an error for a _generative_
+    constructor to be incomplete after all augmentation are applied. It is
+    simply an initializing constructor with no body.*
```
eernstg: Nice!

working/augmentations/feature-specification.md
```diff
@@ -970,23 +1087,93 @@ It's a **compile-time error** if:
...
+
+#### Factory constructors
```
eernstg: Vertical space:
```suggestion
#### Factory constructors

```

working/augmentations/feature-specification.md
```diff
@@ -970,23 +1087,93 @@ It's a **compile-time error** if:
...
+A factory constructor has a `factory` modifier in its declaration.
```
eernstg: Perhaps:
```suggestion
A factory constructor declaration includes the modifier `factory`.
```
.. just so we avoid thinking about "the declaration of a constructor".

working/augmentations/feature-specification.md
```diff
@@ -970,23 +1087,93 @@ It's a **compile-time error** if:
...
+or it has a function body, `=> expression` or `{...}`, in which case
+it's a non-redirecting factory constructor.
+
+It's a **compile-time error** if:
+
+* All declarations of a factory constructor are incomplete.
+
+#### Generative constructors
+
+A generative constructor is _complete_ and _redirecting_ if contains
```
eernstg: Typo:
```suggestion
A generative constructor is _complete_ and _redirecting_ if it contains
```

working/augmentations/feature-specification.md
```diff
@@ -970,23 +1087,93 @@ It's a **compile-time error** if:
...
+
+* A `super` parameter (`super.name`).
 
-    *A constructor whose body is `;` is considered incomplete. This allows an
-    augmentation to fill in a more interesting body if it wants to. But Dart
-    also allows a generative constructor to have a body of `;` which is treated
-    as equivalent to `{}`. Thus it is not an error for a _generative_
-    constructor to be incomplete after all augmentations are applied. It is
-    simply treated as having an empty body.*
+* A declaring parameter (`final int name`), only valid in a primary constructor.
```
eernstg:
```suggestion
* A declaring parameter (e.g, `final int name`). *Such parameters can
  only occur in a primary constructor.*
```

working/augmentations/feature-specification.md
```diff
@@ -970,23 +1087,93 @@ It's a **compile-time error** if:
...
 
-An incomplete constructor can be completed by adding an initializer list and/or
-a body, or by adding a redirection:
+* An initializer list (like `: param = expr, ...` or `: super(args)`).
```
eernstg: I think we should avoid implying that the left hand side of an initializer list element is a parameter:
```suggestion
* An initializer list (like `: name = expr, ...` or `: super(args)`).
```

working/augmentations/feature-specification.md
```diff
@@ -970,23 +1087,93 @@ It's a **compile-time error** if:
...
 
+* A body (does not end in `;`).
+
+_This applies both to primary constructor declarations and
+non-primary constructor declarations, where the primary constructor
+declaration declares its initializer list and body in an in-body
+constructor part, like `this: ... { ... }`. An in-body constructor
+part of `this;` does not make the declaration complete._
```
eernstg: We have used the terminology 'body part' to denote member declarations of the form `this ...`. "in-body" anything is more like a reference to the abandoned feature known as in-body primary constructors. So, perhaps:
```suggestion
declaration declares its initializer list and constructor body in
a body part (`this: ... { ... }`). A body part of the form `this;` does not make
the declaration complete._
```

working/augmentations/feature-specification.md
```diff
@@ -970,23 +1087,93 @@ It's a **compile-time error** if:
...
+A generative constructor declaration is _incomplete_ if it is not
+_complete_.
+
+A generative constructor declaration is _redirecting_ if it's
+a complete redirecting declaration, or if it augments a declaration
+which is redirecting.
+
+A generative constructor declaration is _initializing_ (not redirecting)
+if it's a primary constructor declaration, a complete and initializing
+declartion, or if it augments a declaration which is initializing.
```
eernstg: Typo:
```suggestion
declaration, or if it augments a declaration which is initializing.
```

working/augmentations/feature-specification.md
```diff
@@ -1000,15 +1187,68 @@ augment class C {
...
+}
+
+augment class D(final int x) {
+  augment this { print("Initialized D"); }
+
+  @Deprecated("Was a bad idea anyway");
+  augment D(int _);
+}
+```
+#### Instance variable initialization during constructor invocation
```
eernstg, line 1286: Vertical space:
```suggestion

#### Instance variable initialization during constructor invocation
```

working/augmentations/feature-specification.md
```diff
@@ -1000,15 +1187,68 @@ augment class C {
...
+Then all non-`late` instance variable declarations of the class which have
+an initializer expression are processed in their source order.
+
+Each instance variable is initialized in turn, by evaluating its initializer expression. If the class has a primary constructor, the initializer expression
+is evaluated in the initializer list scope, otherwise it's evaluated in the
+class scope, and if the evaluation completes with a value, the instance
+variable is initialized to that value.
+
+After all instance variable initializers have been executed,
+constructor execution continues with executing as normal,
```
eernstg:
```suggestion
constructor execution continues as in Dart before this feature,
```

working/augmentations/feature-specification.md
```diff
@@ -1000,15 +1187,68 @@ augment class C {
...
+starting with variable initialization by initializing formals
+and declaring parameters.
+
+_This is how primary constructors already work. The only difference is
+that because a single constructor can have more than one declaration,
+the complete declaration of a primary constructor may not be a primary
+constructor declaration._
```
eernstg: The constructor entity which is introduced by an augmentation chain of constructor declarations is primary if at least one of the declarations in the chain is primary. It does not make sense to say that 'a constructor .. is a constructor declaration', but it does make sense to say that the constructor is primary. Perhaps:


```suggestion
_This is how primary constructors already work. The only difference is
that because a constructor can be introduced by more than one
declaration, some declarations contributing to a given primary
constructor might not be primary constructor declarations._
```

eernstg (approved):
Very nice! It's definitely good that it is now possible to have multiple elements in the augmentation chain for a primary constructor.

I noted a bunch of small issues, and a couple of hard discrepancies (for instance, a few grammar rules must be adjusted).

working/augmentations/feature-specification.md
lrhn: That's very, very confusing.
We may never have referred to the actual body block/`;` of a constructor as a "body part", but I'd call it that in a heartbeat in normal conversation.

Every time I see "body part" about a primary constructor, I have to stop and think what it means.
I'd very much like to _not_ use "body part' about the in-body part of the primary constructor.

eernstg: So here's the current terminology:

```dart
class C.name(final int i) {
  //  ^^^^^^^^^^^^^^^^^^^
  // The "header part" of the constructor `C.name`.

  this { print('New C!'); }
//^^^^^^^^^^^^^^^^^^^^^^^^^
// The "body part" of the constructor.
  //   ^^^^^^^^^^^^^^^^^^^^
  // The "body" of the constructor.
}
```

We could say that they are the "in-header part", the "in-body part", and the "body". Would you have to stop and think about that, too? If it's better then we should do it consistently.

working/augmentations/feature-specification.md
```diff
@@ -274,6 +285,60 @@ Just make two separate unnamed extensions.*
...
+## Primary constructors
+
+Both `class` and `enum` declarations can use the primary constructor syntax
+for declaring an initializing _(non-redirecting generative)_ constructor.
+
+The primary constructor syntax does not allow an extra `augment` to be added
+to the class header. As an exception, a primary constructor declaration does
+not need an extra `augment`, on top of the one on the class or enum declaration
+itself, if it declares an augmenting constructor using the primary constructor
+syntax, and it does not have an in-body part. If the declaration _does_ haven
```
munificent: Nit: "haven" -> "have an".

working/augmentations/feature-specification.md
munificent: FWIW, I have referred to the syntax as a "this block". Some canonical name would definitely be good. The tech writers will need to know what to call it.

munificent (approved):

working/augmentations/feature-specification.md
lrhn: A specification doesn't have to say _why_ it does something. It can, as commentary, but it should work just as well without.
It also shouldn't say what it doesn't do, so this text was bad in that regard.

I've rewritten it to be more direct, and separating the commentary into such.

working/augmentations/feature-specification.md
lrhn: "Either" doesn't (necessarily) preclude it being both. But let's not give the reader any chance of misunderstanding.
But this is not a new rule, that's the rule from primary constructors. Rephrased to make it more precise.

working/augmentations/feature-specification.md
lrhn: It is. Changed to commentary and put closer to rule.

working/augmentations/feature-specification.md
lrhn: Rewritten slightly.

working/augmentations/feature-specification.md
lrhn: The language does not, but this refers to individual syntactic "declarations", so it needs to say what it means by "declaration".

I think I wrote this because the "relevant grammar production" for a class declaration should be `<classDeclaration>`, which does not contain `<metadata>`. 
Metadata is attached to "the declaration". As such, it is not itself part of that "the declaration".

Also, it's not a rule, it's commentary.


In this text we have `memberDeclarations ::= (metadata 'augment'? memberDeclaration)*`, which
makes metadata something that's before a `memberDeclaration`. But so is the `augment`, which feels weird.
(The grammar isn't organized purely by syntactic entity. It reuses `'augment'?` here because it gives a shorter grammar for the same language as putting it on every `memberDeclaration`, not because it considers the `augment?` not part of the "member declaration". Right? So referring to "the declaration" syntactically is a little iffy, when the syntax itself doesn't do that. Even splitting the above into
```ebnf
memberDeclarations ::= (metadata augmentableMemberDeclaration)*;
augmentableMemberDeclaration ::= 'augment'? memberDeclaration;
```
would allow referring to "augmentable member declaration", whether it's marked as augmented or not, and the member declaration independent of whether it's augmented or not.)

I don't think it _matters_ whether metadata is included or not.
I'm not sure the span of a declaration matters at all, the ordering is based on the position of the
name-token of the declaration.

Also, I didn't want to worry about whether doc-comments are metadata or whitespace.

working/augmentations/feature-specification.md
lrhn: Reworded.

working/augmentations/feature-specification.md
lrhn: Not sure a redirecting generative constructor can occur "in" a primary constructor.

Anyway, I completly rewrite the entire section.
Let's see if I threw something out with the bath water.

working/augmentations/feature-specification.md
lrhn: Ack. Still doesn't have a proper name.

Using "this block" is confusing when it's not a block.
It makes sense when its `this { something here; }`, but not really for  `this: assert(bananas.areGreat);`.

I'll use "in-body `this`-part".
It's a mouthful, but it's precise _and_ understandable. (I think. Hope. Pray!)

It is a _part_ of a primary constructor declaration, ☑️. (Unlike most single declarations those can occur in non-contiguous source spans.)

It's _in the body_, as opposed to the _in-header part_ of the primary constructor declaration, ☑️.

The `this` is just to remind people what the "in-body part" of a primary constructor is, ☑️.

Checks all my boxes!

working/augmentations/feature-specification.md
lrhn: Something like that.

lrhn:
Updated (significantly in places). I think it has better flow, but not sure if it has better consistency with the rest.

PTAL!

working/augmentations/feature-specification.md
```diff
@@ -274,6 +285,34 @@ Just make two separate unnamed extensions.*
...
+Both `class` and `enum` declarations can use the primary constructor syntax
+for declaring an initializing _(non-redirecting generative)_ constructor.
+
+See [Generative constructors](#generative_constructor_declarations).
+
+### Instance variable initializer expressions
+
+If a class or enum has a primary constructor, then the scope of non-`late`
+instance variable initializer expressions is the initializer list scope
+of that constructor.
```
eernstg: A couple of small edits to use the same terminology as other spec documents:
```suggestion
If a class or enum has a primary constructor, then the current scope of the
initializing expression of a non-`late` instance variable is the primary initializer
scope.
```

working/augmentations/feature-specification.md
```diff
@@ -274,6 +285,34 @@ Just make two separate unnamed extensions.*
...
+
+It's a compile-time error if a non-`late` instance variable initializer
+expressions refers to a variable introduced by the constructor's
+initializer list scope, and the surrounding class or enum declaration does
+_not_ have a primary constructor declaration which declares the
+corresponding parameter's name.
```
eernstg, line 313: A couple of terminology adjustments:
```suggestion
It's a compile-time error if the initializing expression of a non-`late`
instance variable refers to a variable introduced by the primary
initializer scope, and the surrounding class or enum declaration does
_not_ have a primary constructor declaration.
```

However, this is the point that we need to clarify: If we accept the wording that a name even has the ability to 'refer to' such a parameter then it is implied that it is in scope (because some other element in the class/enum augmentation chain of declarations has a primary constructor with a parameter of that name), and it is an error to refer to it.

I like that interpretation, but I believe that the language team voted it down.

working/augmentations/feature-specification.md
```diff
@@ -274,6 +285,34 @@ Just make two separate unnamed extensions.*
...
+_A primary constructor must declare all parameters, but it can omit declaring
+a positional parameter's name by using `_` instead of the name.
+If it does so, that parameter's name cannot be used by instance variable
+initializers of that class or enum declaration._
```
eernstg, line 317: Yes! I think there's no doubt that we want this outcome.

Whether the name itself is forbidden, or it just resolves to whatever declaration we may find in outer scopes, that's the question which needs to be clarified.

working/augmentations/feature-specification.md
```diff
@@ -274,6 +285,34 @@ Just make two separate unnamed extensions.*
...
+
+_The scope exists and is used whether this particular declaration writes
+a primary constructor or not, but for readability reasons, code may only
+refer to variables of that scope, if there appears to be a declaration of
+the variable's name in the surrounding code._
```
eernstg, line 322: Which is true exactly if that's the semantics which is accepted by the language team.

working/augmentations/feature-specification.md
eernstg: Closing: No need to fight over this one. ;)

working/augmentations/feature-specification.md
eernstg: Sounds good.

working/augmentations/feature-specification.md
eernstg: Thanks!

working/augmentations/feature-specification.md
```diff
@@ -924,69 +1002,148 @@ enum values. Enum values are appended in augmentation application order.
...
+
+_A factory constructor declaration with no body and no redirection clause,
+ended by a single `;`, is not a complete declaration, and does not decide
+whether the factory constructor being defined will is redirecting or not.
+Only the (single) complete declaration forces that decision._
+
+#### Generative constructor declarations.
+
+A **primary constructor** declaration causes the constructor it defines
+be a primary constructor, and the class to have a primary constructor.
```
eernstg:
```suggestion
to be a primary constructor, and the class to have a primary constructor.
```

working/augmentations/feature-specification.md
```diff
@@ -924,69 +1002,148 @@ enum values. Enum values are appended in augmentation application order.
...
+contains a parameter list, and optionally a `const` and a constructor-
+name identifier, (`class const C.id(int x)`). It may also have an
+in-body `this`-part which can contain an initializer list or body,
+and can have metadata attached (`this: super(x) { print('done'); }`)._
+
+A primary constructor declaration is an _initializing constructor declaration_.
+A primary constructor declaration is _complete_ if (any of):
+  * It has an initializing formal parameter (`this.name`).
+  * It has a super parameter (`super.name`).
+  * It has a declaring parameter (`final int name`).
```
eernstg, line 1083: This might be helpful for a reader who doesn't quite know the answer:
```suggestion
  * It has a declaring parameter (`final int name` or `var int name`).
```

working/augmentations/feature-specification.md
```diff
@@ -924,69 +1002,148 @@ enum values. Enum values are appended in augmentation application order.
...
+  * It has an in-body `this`-part that:
+      * has a body (`{...}`) and/or
+      * has an initializer list (`: ...`).
+A primary constructor declaration is an _augmenting declaration_ if
+(either of):
+  * It has an in-body `this`-part with an `augment` keyword
+    (`augment this ...`).
+  * It has no in-body `this`-part, and there is another declaration of
+    the same constructor which occurs _before_ this constructor.
```
eernstg: Perhaps 'a constructor with the same name' rather than 'the same constructor'?

It would then be up to various error checks to enforce that it's not just the same name, it is actually a non-erroneous augmentation chain for "the same" constructor.

working/augmentations/feature-specification.md
```diff
@@ -924,69 +1002,148 @@ enum values. Enum values are appended in augmentation application order.
...
+or alternatively to require a second `augment` keyword in the header as
+`augment class augment C(final int x);`._
+
+If a complete primary constructor has a declaring parameter, that parameter
+declaration also counts as a complete introductory or augmenting variable
+declaration, which is `final` if the declaring parameter has the `final`
+modifier, and is augmenting if there is a prior declaration of
+the same getter and/or setter. *(The `augment` modifier does not apply
+to parameters, so the declaring parameter cannot be marked as augmenting
+a variable.)*
```
eernstg, line 1112: Nice!

working/augmentations/feature-specification.md
```diff
@@ -924,69 +1002,148 @@ enum values. Enum values are appended in augmentation application order.
...
+declaration_ and is _complete_ if (any of):
+  * It has an `external` keyword.
+  * It has an initializing formal parameter (`this.name`).
+  * It has a super parameter (`super.name`).
+  * It has a body (`{...}` instead of).
+  * It has an initializer list (`: ...`).
+
+*(An in-body initializing constructor cannot have declaring parameters,
+those are only available to primary constructor declarations.
+Had they been allowed, they'd also make the constructor complete.)*
```
eernstg: 'Had they been allowed ... ' is probably not going to help any readers. ;-)

working/augmentations/feature-specification.md
```diff
@@ -1001,14 +1158,146 @@ augment class C {
...
+    report every declaration that has an error compared to the introductory
+    declaration._
+
+*   The signature of an augmenting constructor does not [match][signature
+    matching] the signature of the corresponding introductory constructor.
+    _The signature of a constructor using the privately-named-parameters
+    feature uses the public name for that parameter._
+
+*   There is a primary constructor declaration, and there is an initializing
+    constructor declaration for any other constructor.
```
eernstg: ... and the enclosing declaration is a `class` or an `enum`. _An extension type declaration does admit initializing constructors in the body even though it must have a primary constructor._

eernstg (approved):
This is great! Added a couple of comments.

The only thing which is still in need of clarification is the distinction:

- A positional primary parameter declared as `_` which has a different name `id` is in scope in all elements of the augmentation chain of the enclosing class/enum declaration, and usages of `id` are resolved as references to the parameter, and this is a compile-time error. OR
- A positional primary parameter declared as `_` which has a different name `id` is ignored during lexical lookups. This means that `id` can be used if it resolves to some other declaration (e.g., a top-level declaration).

I still prefer the former approach, but I believe the language team preferred the latter when it comes to positional parameters in general being accessed in the function body (so `int x = 1; class A { foo(bool x); } augment class A { augment foo(_) => print(x); }` would print 1).

eernstg (approved):
Closed several threads.

lrhn:
Well, before I fix those comments, Gemini found some bugs that I have fixed :blush: .

working/augmentations/feature-specification.md
lrhn: I believe they voted for it. So ... I guess we'll have to ask what the rest of them what they think they agreed to.

working/augmentations/feature-specification.md
lrhn: I get the point. "The constructor" does not exist until we have checked that the declarations are coherent enough to actually define _something_.

If anything, "another constructor _declaration_ with the same name".

It gets annoying quick not have a word for "two declarations that _should_ combine to define the same thing, whether they successfully do so or not".

Maybe "another declaration _for_ the same constructor?"

working/augmentations/feature-specification.md
lrhn: This is intended as a not to future language implementors.

We've had cases where a spec didn't say that something should not be allowed for one reason, because it was already disallowed for another reason. When the second reason was removed, the implementations started allowing something they shouldn't, because nobody ever said there was two reasons to disallow it.

This is me trying to avoid that situation.

But I can change it to "If they're ever allowed, they'll also ...", to be precise about that.
Precise is good. Me like.

working/augmentations/feature-specification.md
lrhn: ACK. This spec hadn't been updated with the primary constructor grammar changes for extension types, so they didn't have "primary construtors". I've fixed that now, but this place may need to say "non-extension type", aka "class or enum" when it comes to primary constructors.

lrhn: Also, the entire section starts with "of a class or enum".
This all needs to be said for extension types as well (at which point this particular case doesn't apply.)

eernstg: Right, I missed that the whole section is class/enum only. Closing.

working/augmentations/feature-specification.md
eernstg: Closing: This is handled via https://github.com/dart-lang/language/issues/4723.

eernstg:
Closed several additional threads.

eernstg (approved):
Closing threads.

lrhn:
I'll land this, knowing that https://github.com/dart-lang/language/issues/4727 is still open, and start working on something covering that problem.

PR #4700, opened 2026-06-04
Infer `const` on generative constructors in an `enum` declaration

eernstg:
The language team decided that `const` would be optional, inferred if not specified, on generative constructors in `enum` declarations. This PR adjusts the enhanced-enums feature specification accordingly.

The PR also removes trailing whitespace from a number of lines, and it adds some missing `const` modifiers on constructors in examples showing the desugared meaning of a given `enum` declaration.

### DIFF
--- a/accepted/2.17/enhanced-enums/feature-specification.md
+++ b/accepted/2.17/enhanced-enums/feature-specification.md
@@ -102,7 +102,7 @@ The semantics of such an enum declaration, *E* with name *N*, is defined as intr
 
   * If `E` is declared as `enum Name with Mixin1, Mixin2 implements Type1, Type2 { … }` then the immediate superinterfaces of *C* are the interfaces of `Enum with Mixin1, Mixin2`, `Type1` and `Type2`.
 
-- **Declared members**: For each member declaration of the `enum` declaration *E*, the same member is added to the class *C*. This includes constructors (which must be `const` generative or non-`const` factory constructors.)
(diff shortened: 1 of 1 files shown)


accepted/2.17/enhanced-enums/feature-specification.md
```diff
@@ -47,9 +47,9 @@ The grammar of the `enum` declaration becomes:
     <arguments>
 ```
 
-It is a **compile-time error** if the enum declaration contains any generative constructor which is not `const`.
-
-_We_ could _allow omitting the `const` on constructors, since it’s required, so we could just assume it’s always there. That’s a convenience we can also add at any later point. For now we require the `const`._
+With a generative `enum` constructor that does not have the modifier `const`,
+this modifier is implicitly induced. *That is, `const` can be omitted, but the
+constructor is constant anyway.*
```
lrhn: (Maybe add ", and must satisfy all the requirements of a generative `const` constructor."
Just so nobody starts assuming you can have a body since it's not _declared_ `const`, and something, somewhere, accidentally used a phrasing like "a declaration declared `const` cannot have a body.".

The _constructor_ is a constant constructor. The _constructor declaration_ may or may not contain the `const` modifier. The two must not be confused.)

eernstg: Yes, that's better. Done.

eernstg (commented):
Adjusted according to review, PTAL.

eernstg:
I'll land this. @lrhn, if you have further comments then let's handle it in an additional PR.

PR #4705, opened 2026-06-08
Move pattern exhaustiveness prototype and fix URLs

eernstg:
This PR moves the exhaustiveness_prototype directory from accepted/future-releases to accepted/3.0/patterns, such that it is located together the other pattern related documents. The PR also fixes some broken URLs in other specification documents.


munificent (approved):
LGTM! Thanks for keeping the repo clean!

To fix the analyzer CI failures, you'll also want to update the paths in `.github/workflows/dart.yml` to point to where the prototype now lives.

PR #4724, opened 2026-07-10
Change inference.md as needed for 4718

eernstg:
This PR changes the specification of type inference such that the inferred return type of a generator function literal does not include a spurious `Null`, which is otherwise added in the case where the end of the body of the function literal can be reached. For example, `() sync* {}` has inferred return type `Iterable<Null>`, but with this change it will be `Iterable<Never>`, as it should be.

### DIFF
--- a/resources/type-system/inference.md
+++ b/resources/type-system/inference.md
@@ -332,7 +332,7 @@ follows.
     - Let `T` be `Null` if any control path reaches the end of the block
       without exiting the block, as computed by the **definite completion**
       analysis specified elsewhere.
-  
+
 Let `K` be the typing context for the function body as computed above from the
 imposed return type schema.
 
--- a/resources/type-system/inference.md
+++ b/resources/type-system/inference.md
@@ -6,6 +6,10 @@ Status: Draft
 
 ## CHANGELOG
 
+2026.07.09
+  - Remove a spurious `Null` from the element type inferred for a generator
+    function literal when the end of the body is reachable.
+
 2024.12.17
   - Change the function literal return type inference rules to ignore
     `return;` statements in generators (it doesn't actually cause null to be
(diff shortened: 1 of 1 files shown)


resources/type-system/inference.md
```diff
@@ -318,13 +322,20 @@ inferred type of the expression body, using the local type inference algorithm
...
+follows.
+
+  - When the function literal is a generator, let `T` be `Never`.
+  - When the function literal is not a generator:
+    - Let `T` be `Never` if every control path through the block exits the
+      block without reaching the end of the block, as computed by the
+      **definite completion** analysis specified elsewhere.
+    - Let `T` be `Null` if any control path reaches the end of the block
+      without exiting the block, as computed by the **definite completion**
+      analysis specified elsewhere.
```
lrhn, line 334: (You can probably guess I'd prefer that case to have type `void` 😁 . Not returning something is not the same as returning `null`. It's the same as not returning anything useful. Alas, not where we are today.)

resources/type-system/inference.md
```diff
@@ -318,13 +322,20 @@ inferred type of the expression body, using the local type inference algorithm
...
-schema.
+follows.
+
+  - When the function literal is a generator, let `T` be `Never`.
+  - When the function literal is not a generator:
+    - Let `T` be `Never` if every control path through the block exits the
+      block without reaching the end of the block, as computed by the
+      **definite completion** analysis specified elsewhere.
+    - Let `T` be `Null` if any control path reaches the end of the block
+      without exiting the block, as computed by the **definite completion**
```
lrhn, line 333: Is "without exiting the block" redundant? Or just confusing?
If control reaches the end, then it definitely didn't exit before that.
On the other hand, it _will_ exit as the very next thing, so does it _really_ "reach then end of the block without exiting" when it "reaches the block and exits"?

lrhn (approved):

Issue #4727, opened 2026-07-15
Augmenting primary constructors and instance variable initializer scopes.


lrhn:
Primary constructors ships with the behavior that:
```dart
class C({final int _x = 0}) {
  this: assert(_x >= 0);
  final int y = _x + 1;
}
```
declares a constructor with a signature `C Function({int x})` and `_x + 1` is evaluated in the initializer list scope of that constructor which has a binding for `_x`.

For augmentations this raises the complexity of constructor overrides.

The currently proposed behavior is that:
```dart
class C({int x});
augment class C({final int _x = 0});
```
is valid since the first declaration's constructor is incomplete, and the two constructor declarations have compatible signatures. The `_x` is an "implementation detail".
However, that detail leaks.

You can write:
```dart
class C({int x}) {
  final int y = x;
}
augment class C({int x = 0}) {
  this: _x = x;
  final int _x;
}
```
and it will work, but, the second primary constructor declaration is complete (it has an initializer list).

However if you try to write:
```dart
class C({int x}) {
  final int y = x;
}
augment class C({final int _x = 0});
```
it _won't_ work, because `x` in `final int y = x;` is not in the initializer list scope for the primary constructor, that contains `_x`.

The augmentation, written later and using an _idiomatic_ instance variable declaring parameter (and valid by itself) introduces an error in the earlier class declaration.

Introducing errors in other fragments isn't special, that can happen with augmentations when everybody have to agree. What's annoying here is that they're not really disagreeing on what happens, only on how to refer to the constructor parameter, and only because one of them uses the supposedly helpful private-named-parameters feature.

**Possible ways to handle this**
1. Don't use the initializer list scope for instance variable initialziers. Instead use a scope derived from the parameter list of the primary constructor on the _same_ fragment that contains the instance variable initializer. So the class declaration with `C({int x})` would see `x` in that scope, the class with `C({final int _x = 0})` would see `_x`.
    * It introduces a _new_ scope. Per class declaration.
    * You can't (as easily) move field initializers between the field and an initializer list if they have different scopes. (But they don't if they're inside the same class declaration.)
    * It's an error to refer to the public name of a parameter in an initializer of a class fragment that doesn't have a primary constructor declaring that parameter with that name. (There is only one declaration which can possibly have a private-named parameter, since that's necessarily a complete declaration, and inside that class declaration field initializers can then refer to that private name, and not the public one.)

[...]

PR #4729, opened 2026-07-20
Scopes for field initializer expressions with augmentations and primary constructors

lrhn:
Adds the "field initializer scope" which as the scope for instance variable initializers.

It's based on the names and types of the primary constructor's signature, and can be used whether there is a constructor or not. See: https://github.com/dart-lang/language/issues/4727

### DIFF
--- a/working/augmentations/feature-specification.md
+++ b/working/augmentations/feature-specification.md
@@ -293,7 +293,7 @@ always uses the same `on` clause as the introductory declaration.
 
 ## Primary constructors
 
-A `class`, `enum` or `extension type` declarationscan use the primary
+A `class`, `enum` or `extension type` declarations can use the primary
 constructor syntax for declaring an initializing _(non-redirecting
 generative)_ constructor.
 
@@ -311,6 +311,9 @@ expressions refers to a variable introduced by the constructor's
 initializer list scope, and the surrounding class or enum declaration does
 _not_ have a primary constructor declaration which declares the
 corresponding parameter's name.
+See more details in the [Instance variable initializer](instance-variable-initialization-during-constructor-invocation)
+section.
+
 _A primary constructor must declare all parameters, but it can omit declaring
 a positional parameter's name by using `_` instead of the name.
 If it does so, that parameter's name cannot be used by instance variable
@@ -768,7 +771,7 @@ augmentation. In that case, it augments the corresponding member (the
 introducing member with the same name) in the same augmentation context,
 according to the rules in the following subsections.
 
-It's a **compile-time** error if:
+It's a **compile-time error** if:
 
 *   An augmenting class declaration has an `extends` clause and any prior
     declaration for the same class also has an `extends` clause.
@@ -940,7 +943,7 @@ introduced implicitly with the value `null` in the case where the parameter
 has a nullable declared type, and no default values for that parameter are
 specified in the augmentation chain.
 
-It's a **compile-time** error if:
+It's a **compile-time error** if:
 
 *   The signature of the augmenting function does not [match][signature matching]
     the signature of the corresponding introductory declaration.
@@ -1241,6 +1244,12 @@ contain constructor declarations where:
     report every declaration that has an error compared to the introductory
     declaration._
 
+*   There is a `const` initializing constructor declaration and:
+    *   A complete instance variable declaration which `late`,
+        or which is neither `final` nor `external`.
+    *   A complete instance variable declaration with an initializer
+        expression which is not a potentially constant expression.
+
 *   The signature of an augmenting constructor does not [match][signature
     matching] the signature of the corresponding introductory constructor.
     _The signature of a constructor using the privately-named-parameters
@@ -1302,7 +1311,7 @@ To perform instance variable initialization on a class or enum declaration:
   the lexical scope for instance variable initializers is the class/enum body
   scope.
   At initialization time, for each non-`late` instance variable with
-  an initializer expression, in source order, evaluate the expression initializer
+  an initializer expression, in source order, evaluate the initializer expression
   in the runtime body scope, then initialize the variable to the result.
 
 * If the class or enum has a primary constructor, each class or enum declaration
@@ -1323,15 +1332,16 @@ To perform instance variable initialization on a class or enum declaration:
   * None of these entries are assignable.
   * An identifier which resolves to a name in the field initializer scope
     is a compile-time error unless the surrounding class or enum declaration
-    has a primary constructor declaration with has a corresponding parameter
+    has a primary constructor declaration which has a corresponding parameter
     declaration which has that identifier as name.
     * _For a private named parameter, only the private name satisfies this._
 
-  This is the lexical scope for instance variable the instance variable
-  initializers.
+  This is the lexical scope for instance variable initializers. The parent
+  scope of the field initializer scope is the body scope of the surrounding
+  class or enum.
 
-  At initialization time, each of the entries are bound to the corresponding
-  actual argument value.
+  When the constructor is invoked, each of the entries are bound to the
+  corresponding actual argument value.
   If there is no corresponding argument, then the parameter must be optional.
   If any declaration of the parameter has a default value, the entry is bound
   to that value, otherwise the entry is bound to `null` _and the parameter
@@ -1341,6 +1351,10 @@ To perform instance variable initialization on a class or enum declaration:
   class or enum declaration, in source order, has its initializer expression
   is evaluated in that runtime field initializer scope, and the variable is
   initialized to the result.
+
+Whether the evaluation uses the body scope or the field initializer scope,
+it is still a compile-time error if the expression refers to any instance member
+in the body scope, or to `this`.
 
 After having run all non-`late` instance variable initializers in all
 declarations of the class or enum, the initializing constructor's implementation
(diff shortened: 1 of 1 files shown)


working/augmentations/feature-specification.md
```diff
@@ -293,7 +293,7 @@ always uses the same `on` clause as the introductory declaration.
 
 ## Primary constructors
 
-A `class`, `enum` or `extension type` declarationscan use the primary
+A `class`, `enum` or `extension type` declarations can use the primary
```
munificent, line 296: "declarations can" -> "declaration can".

working/augmentations/feature-specification.md
```diff
@@ -1241,6 +1244,12 @@ contain constructor declarations where:
     report every declaration that has an error compared to the introductory
     declaration._
 
+*   There is a `const` initializing constructor declaration and:
+    *   A complete instance variable declaration which is `late`,
+        or which is neither `final` nor `external`.
+    *   A complete instance variable declaration with an initializer
+        expression which is not a potentially constant expression.
```
munificent, line 1251: Is it not already an error to have a `const` constructor with fields that aren't `final` with potentially const initializers? What is this section adding to the language?

working/augmentations/feature-specification.md
```diff
@@ -1286,45 +1295,111 @@ contain constructor declarations where:
...
+each declaration of the class or enum in source order, with that
+given argument list.
+
+To perform instance variable initialization on a class or enum declaration:
+
+* If the class or enum does not have a primary constructor,
+  the lexical scope for instance variable initializers is the class/enum body
+  scope.
+  At initialization time, for each non-`late` instance variable with
+  an initializer expression, in source order, evaluate the initializer expression
```
munificent, line 1314: I don't think "in source" order is sufficiently precise to cover how instance variable initializers from different augmentations are ordered. Probably need to say something like:

---

At initialization time:

*   For each declaration of the class or enum, in augmentation application order:
    *   For each non-`late` instance variable in the declaration with an initializer expression, in source order:
        * ...

working/augmentations/feature-specification.md
```diff
@@ -1286,45 +1295,111 @@ contain constructor declarations where:
...
+* If the class or enum has a primary constructor, each class or enum declaration
+  introduces a _field initializer scope_ which:
+  * Has an entry for each parameter with a name in the combined constructor
+    signature of that constructor.
+    _No entries for positional parameters where all declarations use `_`
+    instead of a name, all other parameters represented by their name._
+    * If, and only if, that class or enum declaration has a complete primary
+      constructor declaration which has one or more private named parameters
+      _(which are initializing formals or declaring parameters)_, then both
+      the private and the public names are in the field initializer scope.
```
munificent, line 1326: I'm still not a fan of this approach.

working/augmentations/feature-specification.md
```diff
@@ -1286,45 +1295,111 @@ contain constructor declarations where:
...
+To perform instance variable initialization on a class or enum declaration:
+
+* If the class or enum does not have a primary constructor,
+  the lexical scope for instance variable initializers is the class/enum body
+  scope.
+  At initialization time, for each non-`late` instance variable with
+  an initializer expression, in source order, evaluate the initializer expression
+  in the runtime body scope, then initialize the variable to the result.
+
+* If the class or enum has a primary constructor, each class or enum declaration
```
munificent, line 1317: I think this would be a little clearer as:

```
* Otherwise (the class or enum has a primary constructor): Each class or enum declaration introduces...
```

working/augmentations/feature-specification.md
```diff
@@ -1286,45 +1295,111 @@ contain constructor declarations where:
...
+  corresponding actual argument value.
+  If there is no corresponding argument, then the parameter must be optional.
+  If any declaration of the parameter has a default value, the entry is bound
+  to that value, otherwise the entry is bound to `null` _and the parameter
+  must be nullable_.
+
+  Then each non-`late` instance variable with an initializer expression in that
+  class or enum declaration, in source order, has its initializer expression
+  evaluated in that runtime field initializer scope, and the variable is
+  initialized to the result.
```
munificent, line 1353: I don't like that both branches of the top-level if describe the same behavior here. I think it would be simpler if we organize it like:

```
1.  Each class or enum declaration introduces a _field initializer scope_. This
    is the lexical scope for instance variable initializers. The parent scope is
    the body scope of the surrounding class or enum.

2.  If the class or enum has a primary constructor:

    1.  For each parameter in combined constructor signature of the
        constructor...

    *[More specification here to populate the scope with the primary constructor
    parameters...]*

    *Otherwise if there is no primary constructor, the field initializer scope
    is left empty.*

3.  For each declaration of the class or enum, in augmentation application order:

    1.  For each non-`late` instance variable in the declaration with an
        initializer expression, in source order:

        1.  Evaluate the initializer expression and initialize the variable
            with the result.
```

This way, everything that can be left the same between both branches is shared.

munificent (approved):
I'm still really not a fan of having both the public and private names in scope for private named parameters. But I'm OK with iterating on that in issues and revising this later. 

Some other suggestions but otherwise LGTM! :)

working/augmentations/feature-specification.md
eernstg: I'd also say that these errors are the existing ones in pre-augmentations Dart plus the general principle that static analysis (including emission of compile-time errors) should be performed according to the semantic declarations, not according to one fragment at a time.

working/augmentations/feature-specification.md
```diff
@@ -1286,45 +1295,111 @@ contain constructor declarations where:
...
-constructor execution continues with executing as in Dart before this feature,
-starting with the variable initialization of the parameter list, by initializing
-formals and declaring parameters.
-
-_This is how primary constructors already work. The only difference is that
-because a single constructor can be introduced by more than one declaration,
-the complete declaration, the actual implementation, of a primary constructor
-might not be a primary constructor declaration._
+When invoking the initializing constructor with a valid argument list to
+initialize a new object, perform instance variable initialization on
```
eernstg, line 1304: Just to be safe, we could spell out lateness here again:
```suggestion
initialize a new object, perform non-`late` instance variable initialization on
```

working/augmentations/feature-specification.md
```diff
@@ -1286,45 +1295,111 @@ contain constructor declarations where:
...
+each declaration of the class or enum in source order, with that
+given argument list.
```
eernstg, line 1306: 'given' where? Perhaps there should be an extra sentence above introducing the invocation and in particular its actual argument list.

working/augmentations/feature-specification.md
eernstg: Same here. I'd prefer to treat the declaration using a private name consistently as having this private name for all lexical lookup operations. The only location where the corresponding public name is used is as the label of a named argument at a call site.

working/augmentations/feature-specification.md
```diff
@@ -1286,45 +1295,111 @@ contain constructor declarations where:
 #### Instance variable initialization during constructor invocation
```
eernstg, line 1295: We need to settle a terminology that clearly distinguishes the fragments that are represented as syntactic class/enum/... declarations from the semantic declarations that are obtained by applying all augmentations.

We could say 'declaration fragments' when we mean the syntactic declarations, and 'semantic declarations' when that's what we mean, but we really need to keep those two concepts strictly separated.

working/augmentations/feature-specification.md
```diff
@@ -1286,45 +1295,111 @@ contain constructor declarations where:
...
+  the lexical scope for instance variable initializers is the class/enum body
+  scope.
+  At initialization time, for each non-`late` instance variable with
+  an initializer expression, in source order, evaluate the initializer expression
+  in the runtime body scope, then initialize the variable to the result.
+
+* If the class or enum has a primary constructor, each class or enum declaration
+  introduces a _field initializer scope_ which:
+  * Has an entry for each parameter with a name in the combined constructor
+    signature of that constructor.
```
eernstg, line 1320: I don't think we have defined 'combined constructor signature'. It should suffice to talk about 'the constructor declaration' in 'the semantic class declaration', or just 'the semantic constructor declaration'.

One reason why we might want to avoid 'combined' here is that we already use 'combined member signature' to talk about the handling of non-identical member signatures in superinterfaces.

Another reason is that the semantic constructor declaration has a signature which isn't _combined_ in the usual sense of that word, it is the semantic signature which is obtained by gathering pieces from the entire augmentation chain of the declaration, but those pieces are never allowed to disagree on anything, they are only allowed to omit certain elements (like the type or name), which will then be taken from other syntactic declarations.

working/augmentations/feature-specification.md
```diff
@@ -1286,45 +1295,111 @@ contain constructor declarations where:
...
+    * If, and only if, that class or enum declaration has a complete primary
+      constructor declaration which has one or more private named parameters
+      _(which are initializing formals or declaring parameters)_, then both
+      the private and the public names are in the field initializer scope.
+  * That entry has the type of the parameter in the constructor signature as
+    its declared type.
+    _If no declaration of a constructor has an explicit type, then a
+    type may have been inferred from a default value or it may have defaulted
+    to `dynamic`._
+  * None of these entries are assignable.
```
eernstg, line 1332: This is not quite congruent with "a field initializer scope which:". Some other entries in this list have the same issue.

eernstg (commented):
Looks good in general!

A couple of comments: I do think we need to choose an unambiguous terminology for the syntactic declarations of classes/enums/etc, and for the semantic declaration which is the result of applying all the augmenting declarations. Another question to consider is whether the corresponding public name should be in scope together with the private name, in the case of a private named parameter in a primary constructor.

PR #4731, opened 2026-07-22
Remove the getter/setter signature mismatch error

eernstg:
This is the specification update that corresponds to the implementation issue https://github.com/dart-lang/sdk/issues/58578. It updates the NNBD feature specification to remove the getter/setter type mismatch error, and it updates dartLangSpec.tex as well.

The feature has been implemented and released, that is, there is no need to change any tools in response to this spec update.

### DIFF
--- a/accepted/2.12/nnbd/feature-specification.md
+++ b/accepted/2.12/nnbd/feature-specification.md
@@ -6,6 +6,10 @@ Status: Draft
 
 ## CHANGELOG
 
+2026.07.22
+  - Remove the compile-time error for a getter/setter pair where the return type
+    of the getter is not a subtype of the parameter type of the setter.
+
 2025.06.16
   - Add an exception for extension members in the error about nullable
     receivers.
@@ -613,7 +617,7 @@ It is an error for the initializer expression of a `late` local variable to use
 a prefix `await` expression that is not nested inside of another function
 expression.
 
-It is an error for a class with a generative `const` constructor to have a 
+It is an error for a class with a generative `const` constructor to have a
 `late final` instance variable.
 
 It is not a compile time error to write to a `final` non-local or instance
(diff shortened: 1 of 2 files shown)


accepted/2.12/nnbd/feature-specification.md
lrhn: "body of a ... field" -> "initializer expression of a ... instance variable"
?

lrhn: (Not sure why it should be. Why do we need to say what isn't? Should it be commentary?
Is it an error to have a `final` non-local or instance variable if it's not declared `late` or does not have an initializer?
How does the "or" in "non-local or instance" associate? Instance variables are non-local, so can `or instance` be removed?)

lrhn (approved):
LGTM, STV, SHIP IT!

(Unrelated drive-by comments only.)

accepted/2.12/nnbd/feature-specification.md
eernstg: Done.

eernstg: Changed it to commentary.

I suppose it may have been included because it seems reasonable to have a rule like "it's a compile-time error to write to a `final` variable, period", and this is a reminder that some `final` variables are intended to allow write operations because they aren't initialized at the declaration.

> Is it an error to have a `final` non-local or instance variable if it's not declared `late` or does not have an initializer?

If it is not `late` _and_ does not have an initializer (after augmentations, if any), a top-level and static variable is a compile-time error. An instance variable must be initialized, but there are several other ways to do it.

I think it's OK to just make it commentary and then leave it as is.

eernstg (commented):
Thanks, adjusted the wording in a couple of locations.

PR #4732, opened 2026-07-22
Fix a couple of potential security issues in download_packages.dart.

munificent:
The script accessed the pub API over HTTP and not HTTPS, which could allow a MITM attack. Further, the package name and version in the response could contain path characters like "../../../" which would allow a malicious response to write elsewhere in the file system.

(In practice, this is a one-off rarely run utility so it's fine, but a security tool flagged it, so we may as well fix it.)

See b/528556622.


stereotype441 (approved):
lgtm!

Issue #4735, opened 2026-07-27
[augmentations] General principle and names of implicitly induced instance variables


eernstg:
Thanks to @sgrekhov and @fshcheglov for bringing up this case. Consider the following program:

```dart
class C2([final int? _]) {}

augment class C2 {
  augment C2([int? x]);	
}

// If we decide that primary constructors can only be augmented by primary constructors we'd have this:
augment class C2([int? x]);
```

Following the [general principle](https://github.com/dart-lang/language/blob/main/working/augmentations/feature-specification.md?rgh-link-date=2026-07-27T14%3A28%3A12.000Z#compile-errors-with-augmentations) about static analysis of augmentations, we should apply augmentations whenever possible, and then perform the static checks on the semantic declaration.

This implies that we determine the name of the parameter by checking that the augmentation chain for the parameter contains a single identifier (specified any positive number of times) which is not `_` together with zero or more occurrences of `_`, or it specifies `_` everywhere. In the latter case, the parameter is wildcarded. In the former case, the parameter name is said identifier. In the example above this means that the parameter has the name `x`.

Consequently, the parameter will implicitly induce an instance variable whose name is also `x`, and there are no compile-time errors in the program.

However, the fact that a parameter is explicitly declaring in the introductory declaration could also be taken to imply that it introduces an instance variable whose name is `_` along with a wildcarded parameter, and the parameter name `x` in the augmenting declaration is a compile-time error. This means that we'd introduce a special exception with declaring parameters, that they can't use `_` as a "don't care" parameter, or at least not in the declaration that carries the keyword `final` or `var`.

@dart-lang/language-team, WDYT?

lrhn:
The declaring parameter `final int? _` is the only declaration of the instance variable. 
That's not (just) a parameter name, and instance variable/getter names are not optional.
It introduces an instance variable named `_`.

The only question to me is whether that should be an error, to have declaring parameter whose name is not the actual name of the parameter.
We do allow it for private named parameters, but this feels different. 

We allow omitting the names of positional parameters _because_ an omitted name isn't used. Here it is used. 

I think I'd say that if a parameter is declaring (or initializing), then it it's name cannot be omitted, so whatever (public) name it uses must be the name that all declarations agree on. Even if it's `_`. That `_` does not count as omitting a name.

> that they can't use `_` as a "don't care" parameter, or at least not in the declaration that carries the keyword `final` or `var`.

[...]

PR #4741, opened 2026-08-06
Declaring parameters can't be don't-care

eernstg:
This PR adds a new paragraph in the section 'Augmenting function and constructor signatures', specifying that it is a compile-time error for a declaring parameter to use a "don't care" name. This ensures that it is always possible to see the name of the implicitly induced instance variable by looking at the formal parameter declaration fragment that induces it.

For more background and discussions, see https://github.com/dart-lang/language/issues/4735.


stereotype441 (approved):
lgtm

munificent (approved):
LGTM!

Consider also updating the CHANGELOG in the proposal.

Issue #433, opened 2019-06-28
Should we restrict cycles of F-bounded types through union types? 


leafpetersen:
We don't allow type variable declarations of the form `T extends T`, or `T extends S, S extends T`.  The analyzer currently accepts type variable declarations of the form `T extends FutureOr<T>`, but the CFE crashes on such declarations.  With NNBD, we will also have the ability to write `T extends T?` or `T extends S, S extends T?` unless we specify otherwise.

Should we consider making these an error?

My inclination is yes.  In general, this feels to be pushing up against (or crossing) an important boundary in the equational theory that may bite us if we eventually add union types and attempt to re-interpret `?` and `FutureOr` as union types.  Note that for example `T extends T | Never` is essentially another way of writing `T extends T`, which suggests the kinds of problems that may arise.

I also haven't found any uses for this kind of equation, and since the CFE doesn't accept the `FutureOr` example, there is almost certainly no existing code that uses this feature.

One way of specifying this is to say that given a set of type variable declarations of the form `X0 extends B0, ..., Xn extends Bn`, construct a graph with one node for every `Xi`, and edges from `Xi` to `Xj` iff `Xi` occurs in a top level union in `Bj`, and say that it is an error if that graph has any cycles.  We say that `Xi` occurs in a top level union in a type `T` iff:
  - `T` is `Xi
  - `T` is `S?` and `Xi` occurs in a top level union in `S`
  - `T` is `FutureOr<S>` and `Xi` occurs in a top level union in `S`

@eernstg @lrhn @munificent thoughts?

cc @stereotype441

lrhn:
Since `T extends T` is satisfied by any type, it's a useless bound.
The other ones are similarly useless as a bound because `T` always occurs directly in the union.

So, there should be no *harm* in disallowing those cases.

Is there harm in allowing them? That is, can we *recognize* that these are meaningless bounds that are trivially satisfied by any type, and just ignore them? If we have to detect them anyway to make them errors, it seems equally possible to ignore them.

The safe bet is to disallow. I'll defer to your experience here, so do whatever you find best. It shouldn't affect anybody negatively.

(The moment we get union types, we can write recursive equations with type variables, like this one that will occur alomost immediately: `Json extends num|Null|String|bool|List<Json>|Map<String,Json>`. I'm not sure whether this is an argument for or against union types :smile:.)

eernstg:
@leafpetersen wrote:
> Should we consider making these an error? .. My inclination is yes.

[...]

PR #4743, opened 2026-08-07
Specify error for pseudo-cyclic type parameter bounds

eernstg:
For background and discussion, please see dart-lang/language#433.

This PR introduces a compile-time error for type parameters and their bounds when they form a union type based cycle. For example, not just `X extends X` and `Y extends Z, Z extends Y` are errors (as they have always been), but also `X extends X?` and `Y extends FutureOr<Z>, Z extends Y`. That is, we make it an error to form a cycle even when union types are stripped off first. This is done because it allows us to avoid some known sources of infinite looping in the subtype and standard upper bound algorithms.

Following https://github.com/dart-lang/language/issues/433#issuecomment-2415818156, this PR also renames "union-free type derived from a type" to "union base type of a type".

### DIFF
--- a/specification/dart.sty
+++ b/specification/dart.sty
@@ -478,10 +478,6 @@
 \newcommand{\MutualSubtypeStd}[2]{\MutualSubtype{\Delta}{#1}{#2}}
 \newcommand{\MutualSubtypeNE}[2]{\ensuremath{{#1}\,<:>\,{#2}}}
 
-% Judgment expressing that a type is a subtype of the union base type
-% of another.
-\newcommand\UnionBasedSubtype[2]{\ensuremath{#1\,\cup\!\EXTENDS\,\,\,#2}}
-
 % Judgment expressing that a supertype relation exists.
 \newcommand{\Supertype}[3]{\ensuremath{{#1}\vdash{#2}\,:>\,{#3}}}
 \newcommand{\SupertypeStd}[2]{\Supertype{\Delta}{#1}{#2}}
--- a/specification/dart.sty
+++ b/specification/dart.sty
@@ -478,6 +478,10 @@
 \newcommand{\MutualSubtypeStd}[2]{\MutualSubtype{\Delta}{#1}{#2}}
 \newcommand{\MutualSubtypeNE}[2]{\ensuremath{{#1}\,<:>\,{#2}}}
 
+% Judgment expressing that a type is a subtype of the union base type
+% of another.
+\newcommand\UnionBasedSubtype[2]{\ensuremath{#1\,\cup\!\EXTENDS\,\,\,#2}}
+
 % Judgment expressing that a supertype relation exists.
 \newcommand{\Supertype}[3]{\ensuremath{{#1}\vdash{#2}\,:>\,{#3}}}
 \newcommand{\SupertypeStd}[2]{\Supertype{\Delta}{#1}{#2}}
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -7474,7 +7474,8 @@ It is a \Error{compile-time error} if a generic declaration $D$ has
 type parameters and bounds
 \code{$X_1$\,\,\EXTENDS\,\,$B_1$,\,\dots,\,$X_s$\,\,\EXTENDS\,\,$B_s$} such that
 \UnionBasedSubtype{X_j}{X_{j+1}} for all $j \in 1 .. s - 1$,
-and \UnionBasedSubtype{X_s}{X_1}.
+and \UnionBasedSubtype{X_s}{X_1}
+\commentary{(that is, $B_s$ is $X_1$, possibly wrapped in some unions)}.
 These type parameters can be declared in any order,
 and there may be additional type parameters declared by $D$,
 the requirement is just that the cycle exists.
--- a/specification/dartLangSpec.tex
+++ b/specification/dartLangSpec.tex
@@ -7491,8 +7491,9 @@ the requirement is just that the cycle exists.
   could just as well have omitted the bound because it is always satisfied.
   Similarly, a pseudo-circular declaration like
   \code{X\,\,\EXTENDS\,\,FutureOr<Y>, Y\,\,\EXTENDS\,\,X}
-  is satisfied when all type variables in the cycle have the same value,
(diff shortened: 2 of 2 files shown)


specification/dart.sty
```diff
@@ -478,6 +478,10 @@
 \newcommand{\MutualSubtypeStd}[2]{\MutualSubtype{\Delta}{#1}{#2}}
 \newcommand{\MutualSubtypeNE}[2]{\ensuremath{{#1}\,<:>\,{#2}}}
 
+% Judgment expressing that a type is a subtype of the union base type
+% of another.
+\newcommand\UnionBasedSubtype[2]{\ensuremath{#1\,\cup\!\EXTENDS\,\,\,#2}}
```
lrhn: What does that look like? Is it basically "A &Union;&zwj;extends B"?

eernstg: Yes. Improvement suggestions would be very welcome, though!

specification/dartLangSpec.tex
```diff
@@ -7439,15 +7442,61 @@ \section{Generics}
 \LMHash{}%
 A type parameter $T$ may be suffixed with an \EXTENDS{} clause
 that specifies the \Index{upper bound} for $T$.
-If no \EXTENDS{} clause is present, the upper bound is \code{Object}.
-It is a compile-time error if a type parameter is a supertype of its upper bound
-when that upper bound is itself a type variable.
+If no \EXTENDS{} clause is present, the upper bound is \code{Object?}.
+
+\LMHash{}%
+We need to define a special subset of the clauses \code{$X$\,\,\EXTENDS\,\,$B$}
```
lrhn: No need to tell the reader that we need to tell them something. Just do it: "We define ..."

specification/dartLangSpec.tex
```diff
@@ -7439,15 +7442,61 @@ \section{Generics}
...
+in order to avoid a certain kind of cycles.
+Let
+\IndexCustom{\UnionBasedSubtype{X}{Y}}{$\cup$@$X\,\,\,\textsf{U}\EXTENDS\,\,Y$}
+be the relation among type variables $X$ and $Y$ where
+\code{$X$\,\,\EXTENDS\,\,$Y$}, or
```
lrhn: Can we say this non-syntactically? And this line is probably redundant with the next.

Something like:
> if the bound of `X` is _B_ and the union base type of _B_ is `Y`.

specification/dartLangSpec.tex
lrhn: Is this a subset of type variables, type variable declarations, or syntactic type parameter clauses?

(Augmentations will want to know.)

specification/dartLangSpec.tex
```diff
@@ -7439,15 +7442,61 @@ \section{Generics}
...
+}
 
+\LMHash{}%
+It is a \Error{compile-time error} if a generic declaration $D$ has
+type parameters and bounds
+\code{$X_1$\,\,\EXTENDS\,\,$B_1$,\,\dots,\,$X_s$\,\,\EXTENDS\,\,$B_s$} such that
+\UnionBasedSubtype{X_j}{X_{j+1}} for all $j \in 1 .. s - 1$,
+and \UnionBasedSubtype{X_s}{X_1}
+\commentary{(that is, $B_s$ is $X_1$, possibly wrapped in some unions)}.
+These type parameters can be declared in any order,
```
lrhn: Are you telling me the indices above are not the syntactic order?
May want to say that up front, it's a hard realignment to have to reinterpret that after having read the text in the most common way.

specification/dartLangSpec.tex
```diff
@@ -7439,15 +7442,61 @@ \section{Generics}
...
+  For example, \UnionBasedSubtype{X}{X}
+  when \code{$X$\,\,\EXTENDS\,\,$X$},
+  when \code{$X$\,\,\EXTENDS\,\,$X$?},
+  and when \code{$X$\,\,\EXTENDS\,\,FutureOr<$X$>}, and
+  \UnionBasedSubtype{X}{Y} when
+  \code{$X$\,\,\EXTENDS\,\,FutureOr<FutureOr<$Y$?>{}>?}.%
+}
 
+\LMHash{}%
+It is a \Error{compile-time error} if a generic declaration $D$ has
```
lrhn: Maybe give the condition a name. Makes it easier to understand and talk about. 

Say 
> A generic declaration has cyclic bounds if
> a subset of its type parameters can be ordered as `X`<sub>1</sub>..`X`<sub>_s_</sub>, has corresponding bounds `N`<sub>1</sub>..`B`<sub>_s_</sub>, and for each (I,j) in {(n, n+ 1| n &in; 1.._s_-1}&Union;{(s, 1)}, the union base type of `B`<sub>i</sub> is `X`<sub>j</sub>.
> It's a compile time error if a declaration has cyclic bounds.

specification/dartLangSpec.tex
```diff
@@ -7439,15 +7442,61 @@ \section{Generics}
...
+
+\rationale{%
+  A type parameter declaration like
+  \code{X\,\,\EXTENDS\,\,X} or \code{X\,\,\EXTENDS\,\,X?}
+  could just as well have omitted the bound because it is always satisfied.
+  Similarly, a pseudo-circular declaration like
+  \code{X\,\,\EXTENDS\,\,FutureOr<Y>, Y\,\,\EXTENDS\,\,X}
+  is satisfied when all type variables are bound to a top type,
+  or a couple of other cases involving bottom types and \code{Object},
+  which is not likely to be useful.
```
lrhn, line 7499: Maybe not. 

You can always _satisfy_ the bounds, but if you have `<X extends Y, Y extends FutureOf<X>>`, you can assign `Y` to `X`, and `await (x as X)` to `Y`.  You couldn't do that if you removed the bounds, and you can't just collapse them to one type variable, that wouldn't have the two different bounds.

You could likely have reduced all of the variables to a single type variable with no bound, _and_ added the `FutureOr` and `?` on occurrences as needed. Maybe.

But more relevantly `FutureOr` and `?` are structural,  so `X extends FutureOf<X>` has an infinitive expansion if one tries to be eager. That's bad. 
(A cycle with no union type shouldn't be a problem if it can be detected, it's just useless.)

lrhn (approved):

PR #4745, opened 2026-08-11
Remove expression statement restriction from dot shorthands spec

### LINKED ISSUE #64009: CFE and analyzer allow dot shorthand at beginning of expression statement
The dot shorthand feature specification [specifies](https://github.com/dart-lang/language/blob/011935ae97dc3d6828d6827fea52374ea271cfff/accepted/3.10/dot-shorthands/feature-specification.md?plain=1#L57-L60):

> We also add `.` to the tokens that an expression statement cannot start with. This doesn't
> affect starting with a double literal like `.42`, since that's a different token than a single `.`.
> _(Not sure this is *necessary*, but it will possibly make parser recovery easier/
> So mainly disallow this as an abundance of caution.)_

However, the analyzer and CFE both seem to accept expression statements starting with dot shorthands:

```dart
void main() {
  .parse('true') && false;
}
```

Perhaps there was a decision that this restriction wasn't necessary?

parlough:
Removes the restriction that suggested preventing expression statements from beginning with a dot (`.`) and therefore static member shorthands, to match the existing implementations which didn't implement the restriction.

Resolves https://github.com/dart-lang/sdk/issues/64009

### DIFF
--- a/accepted/3.10/dot-shorthands/feature-specification.md
+++ b/accepted/3.10/dot-shorthands/feature-specification.md
@@ -4,7 +4,7 @@ Author: lrn@google.com
 
 Status: Accepted
 
-Version: 1.4.1 (2025-04-02)
+Version: 1.4.2 (2026-08-11)
 
 Experiment Flag: dot-shorthands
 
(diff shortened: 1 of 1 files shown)


parlough:
@lrhn @kallentu Thoughts on this?

kallentu (approved):
Looks fine by me if that's the decision we're okay with. @eernstg and @lrhn for the final call though.

PR #4749, opened 2026-08-21
Restrict permissions in Dart CI workflow

athomas:
Declare `contents: read` permissions explicitly in `.github/workflows/dart.yml` to satisfy least-privilege security checks (zizmor `excessive-permissions`).


eernstg (approved):
LGTM

Issue #2824, opened 2023-02-07
Should the use of both show and hide on an import be a syntax error?


alexeyinkin:
In this example `hide` adds no meaning because if an identifier is not in the list of `show` it is not visible anyway:

```dart
import 'package:collection/collection.dart' show ListEquality hide MapEquality;

void main() {
  ListEquality();
  //MapEquality(); // undefined
}
```

Moreover an identifier can be in both `show` and `hide` resulting in it being hidden:

```dart
import 'package:collection/collection.dart' show ListEquality hide ListEquality; // ignore: unused_import

void main() {
  //ListEquality(); // undefined
}
```

Both of these examples build and produce no linter issues.

Is there a valid case for using both `show` and `hide` on a single import? If yes, then there should be a lint that flags the meaningless cases like above. If no, should this be a syntax error? With Dart 3 allowed to introduce breaking changes, should this opportunity be used to add this syntax change?

incendial:
I'd also add that it's currently possible to have multiple show / hide with the same identifier, like:

```dart
import 'package:collection/collection.dart' show ListEquality show ListEquality show ListEquality show ListEquality;

void main() {
  ListEquality();
}
```
and it produces no errors, but looks strange.

eernstg:
Agreed! - the semantics of multiple show and hide clauses on imports and exports is convoluted and even counter-intuitive.

I argued in https://github.com/dart-lang/sdk/issues/32878 that `show ... show ...` should provide access to the union of the lists of names in those `show` clauses, rather than the intersection (which is actually the outcome according to the specification). The behavior of `show ... hide ...` is similarly counter-intuitive.

However, these rules have been around for more than a decade, and there wasn't much appetite in 2018 (or indeed at any time) to change anything in this area. This was probably so because nobody felt a strong urge to write powerful and complex `show ... hide ...` clauses anyway, and there's always a cost associated with the introduction of breaking changes.

I've created a lint proposal, https://github.com/dart-lang/linter/issues/4052.

lrhn:
I'd happily support only allowing one `hide` or `show` on each import. There is never need for anything more.

If we ever introduce a "rename" combinator, we might need to reconsider, but until then one is always enough. We're not helping anyone by allowing more.

(That `show` works like intersection is a consequence of it really meaning "hide-everything-but". Showing everything is the default, so we only need hide-operators. One of them is called `show`.)

[...]

PR #4753, opened 2026-08-24
Working propsal for "single-combinators" language experiment

srawlins:
Work towards https://github.com/dart-lang/language/issues/2824

A candidate for a language "feature" (purely a restriction) in a possible Dart 4 release.

---

### DIFF
--- a/working/single-combinators/feature-specification.md
+++ b/working/single-combinators/feature-specification.md
@@ -1,0 +1,106 @@
+# Single Combinators
+
+Author: Sam Rawlins
+Status: Proposed
+Version: 1.0
+
+Specification for issue
+[#2824](https://github.com/dart-lang/language/issues/2824), restricting the
+specifying of combinator clauses to one combinator per import/export directive.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


munificent:
I love it. Do you want a reviewer for this PR yet?

srawlins:
Sure! I think this spec is complete. I've got an implementation too, but thought I would try to go one step at a time :D

munificent:
The spec looks good to me. It would be nice to have a section on migration. We already have [a lint that warns on multiple combinators](https://dart.dev/tools/diagnostics/multiple_combinators) (nice!) but it doesn't look like it's in the core or recommended sets as far as I can tell.

A good first step would probably be enabling that lint and rolling it out.

The change itself should probably be a language-versioned breaking change. If we do that, we probably won't get to completely remove support for multiple combinators until some future Dart 5.0 release, but it's a start. Or, if we really think the breakage is low enough, we could try to ship it in Dart 4.0, but that's probably hard.

munificent (approved):

srawlins:
It is not a lint, but a static warning. `multipleCombinators` is being reported unless someone opts out of it.

PR #4754, opened 2026-08-27
[flow analysis] Clean up documentation of promotion chains, add Lean model.

stereotype441:
This PR adds a top level directory `lean`, containing the beginnings of a Lean model of the behavior of flow analysis. Currently the model includes:

- An abstraction of the Dart type system, defining some of the operations needed by flow analysis (I will expand on this abstaction as needed).

- A concrete simplified model of the Dart type system, omitting complications that are irrelevant to flow analysis (such as support for f-bounded types). This model is defined as an instantiation of the above abstraction.

- A model of promotion chains and their properties.

I intend to build out this model as part of documenting (and improving) the flow analysis specification.

The Lean code builds locally without errors (which validates all the theorems contained in it). In a follow-up PR I intend to add a github check that will verify that the code continues to build without errors.

The PR also makes the following improvements to `flow-analysis.md`:

- The notation for lists has been changed to use `++` for list concatenation. This is easier to precisely explain than the old notation using `...`, and is more consistent with the Lean model.

- The notion of a subsequence is defined, using the symbol `<+` (which Lean also uses, though it uses the terminology "sublist"). This is used to prove that the join of two promotion chains is a promotion chain (see below).

- The notion of strict subtyping is defined, using the symbol `<<:`. Note that the Lean model uses `≤` for subtyping and `<` for strict subtyping, since this allows a number of convenient theorems to come into play.

- The definition of promotion chains has been moved to its own section.

- The notion of a promotion chain being "valid for declared type `T`" has been renamed to "strictly bounded by `T`". This lays the groundwork for future PRs that will specify the behavior of private field promotion. (I will want to avoid talking about the "declared type" of a field, because of ambiguities that arise when a field is declared with one type in one class and then overridden with another type in a subclass.)

- The join algorithm for promotion chains is now fully specified. A sketch is given of the proof that this algorithm is idempotent and commutative, and a counterexample illustrates that it is not associative. I will file an issue to discuss whether to change to an algorithm that _is_ associative.


resources/type-system/flow-analysis.md
```diff
@@ -106,18 +106,16 @@ that assignment).
...
-    beginning with all the elements of `l` and followed by `a`.
-  - A list of types `p` is called a _promotion chain_ iff, for all `i < j`,
-    `p[j] <: p[i]`. _Note that since the subtyping relation is transitive, in
-    order to establish that `p` is a promotion chain, it is sufficient to check
-    the `p[j] <: p[i]` relation for each adjacent pair of types._
-  - A promotion chain `p` is said to be _valid for declared type `T`_ iff every
-    type in `p` is a subtype of `T`. _Note that since the subtyping relation is
-    transitive, in order to establish that `p` is valid for declared type `T`,
-    it is sufficient to check that the first type in `p` is a subtype of `T`._
+  - We use the notation `l₁ ++ l₂` to denote the concatenation of lists.
```
lrhn, line 111: Very nice use of Unicode!
I don't think there is anything looking like `<+`, but how about `⪿`? Or any of `⋲`, `⋳` or just plain old `∊`? (I'll stop now 😁 )

resources/type-system/flow-analysis.md
```diff
@@ -126,9 +124,83 @@ that assignment).
     the top element of `s`.  If `s` is empty, the result is undefined.
   - We use the notation `top(s)` to mean the top element of the stack `s`.  If
     `s` is empty, the result is undefined.
-  - Informally, we also use `[...t, a]` to describe a stack `s` such that
+  - Informally, we also use `t ++ [a]` to describe a stack `s` such that
```
lrhn: Maybe "describe" -> "act as".

resources/type-system/flow-analysis.md
```diff
@@ -126,9 +124,83 @@ that assignment).
...
     `top(s)` is `a` and `pop(s)` is `t`.
 
+### Strict subtyping
+
+A type `T` is said to be a _strict subtype_ of `U` (denoted `T <<: U`), iff `T
+<: U` and not `U <: T`.
+
+The strict subtyping relation is irreflexive (`¬ T <<: T`), asymmetric (`¬(T <<:
+U ∧ U <<: T)`), and transitive (`T <<: U ∧ U <<: V → T <<: V`).
```
lrhn: (Use `⇒` or `&rArr;` for implication. Probably the former since this is inside a code-block.)

resources/type-system/flow-analysis.md
```diff
@@ -126,9 +124,83 @@ that assignment).
...
+
+_Note that mutual subtypes are excluded; for example `¬ dynamic <<:
```
lrhn: (Should I have thought they would be included? I'll admit I read the definition directly as:  "T \<: U and not-mutual-subtypes", so this paragraph feels like it's explaining a tautology.)

resources/type-system/flow-analysis.md
```diff
@@ -126,9 +124,83 @@ that assignment).
...
+Object?`. This is necessary for transitivity._
+
+### Promotion chains
+
+A list of types `c` is called a _promotion chain_ iff, for all `i < c.length -
```
lrhn: ... for all `1 < i < c.length - 1` ...

This index use seems to starts at 1. (There are `c.length - 1` elements in the range 1 <= i < c.length, which is what there must for the (c[i], c[i+1])s to be valid and cover all adjacent pairs.

So indexes start at 1 and end at `length`.

resources/type-system/flow-analysis.md
```diff
@@ -126,9 +124,83 @@ that assignment).
...
+1`, `c[i + 1] <<: c[i]`.
```
lrhn: (Complete side-track, because I was looking at it for another reason. If we have 
```dart
void foo<X extends Never, Y extends X>(X value) { 
   if (value is Y) {
     // Not promoting to X&Y because X and Y are mutual subtypes (both bottom).
   }
}
```
Should we promote `value` to `X&Y` here? Or will flow analysis recognize that we never take the branch anyway, and not promote because of that. (Not that we can ever call the function.)

It _probably_ doesn't matter in any way, because the result is assignable to `Y` whether we promote or not (it's a bottom type), and it has all members (it's a bottom type).
And because the code can't possibly be run either way.
That means we can't ever have an `X&R` where `X` is bottom-bounded, and UP of any two bottom types can just be `Never`, you can't tell the difference.)

resources/type-system/flow-analysis.md
```diff
@@ -126,9 +124,83 @@ that assignment).
...
+
+_We will use promotion chains to represent the state of a variable that has been
+promoted zero or more times (via an `is` test, `as` cast, or an operation that
```
lrhn: Or any other way of promoting, so maybe don't try to list them, or have a "for example".
(Most patterns could probably count as `is` or `as` operations, but promotion by exhaustion doesn't really fit. if anything, it's the "other" branch of a _series_ of implicit `is` tests.)

resources/type-system/flow-analysis.md
```diff
@@ -106,18 +106,16 @@ that assignment).
...
-  - A promotion chain `p` is said to be _valid for declared type `T`_ iff every
-    type in `p` is a subtype of `T`. _Note that since the subtyping relation is
-    transitive, in order to establish that `p` is valid for declared type `T`,
-    it is sufficient to check that the first type in `p` is a subtype of `T`._
+  - We use the notation `l₁ ++ l₂` to denote the concatenation of lists.
+
+    _For example, `l ++ [a]` means the list beginning with all the elements of
+    `l` and followed by `a`._
+  - We use the notation `l₁ <+ l₂` to denote that `l₁` is a subsequence of
+    `l₂`. That is, `l₁ = [l₂[k₀], l₂[k₁], ... l₂[kₙ]]` where `k₀ < k₁ < ... <
```
lrhn: Haven't defined index operation (yet?), so we don't know that the `k`s here are integers.
Or `.length`, which is really needed for index bounds.

> - We use the notation `l.length` for the number of values in a list. If `l` is a list `[v₁, ..., vₙ]`, then `l.length` is n. _Length satisfies that `[].length` is zero, `[v].length` is one, and `(l₁ ++ l₂).length = l₁.length + l₂.length`._
> - We use the notation `l[k]` for a single element of a list where `k` is an integer so that if `l` is `[v₁, ..., vₙ]`, and 1 &le; k &le; n = `l.length`, then `l[k]` is `vₖ`. This operation satisfies that `(l₁ ++ l₂)[k]` is `l₁[k]` if `1 ≤ k ≤ l₁.length` and `l₂[k - l₁.length]` if `l₁.length < k ≤ l₁.length + l₂.length`.
> - We use the notation `l₁ <+ l₂` to denote that `l₁` is a subsequence of
>   `l₂`. That is, `l₁ = [l₂[k₁], l₂[k₂], ... l₂[kₙ]]` where for some k₁,... ,  kₙ with `1 ≤ k₁ < ... <  kₙ ≤ l₂.length`.

Need to decide whether indices go from 0..length-1 or 1..length.
(The use of Unicode subscript characters for subscripts here make me vary of using `-1` in them, so I'd go for 1..length.)

resources/type-system/flow-analysis.md
```diff
@@ -126,9 +124,83 @@ that assignment).
...
+1`, `c[i + 1] <<: c[i]`.
+
+_We will use promotion chains to represent the state of a variable that has been
+promoted zero or more times (via an `is` test, `as` cast, or an operation that
+proves that the variable's value is not `null`). We require `c[i + 1] <<: c[i]`
+to ensure that each successive promotion is a refinement of the previous one,
+which helps make the promotion behavior predictable._
+
+Equivalently, `c` is a promotion chain iff, for all `0 ≤ i < j < c.length`,
+`c[j] <<: c[i]`. This follows from transitivity of `<<:`.
```
lrhn, line 157: Here indices start at 0 and go to length-1. Pick one, either will work.

(Is zero a natural number? Discuss!)

resources/type-system/flow-analysis.md
```diff
@@ -126,9 +124,83 @@ that assignment).
...
+
+Consequently, any subsequence of a promotion chain is also a promotion
+chain. _This fact will be used to prove that the `join` operation (defined
+below) produces a valid promotion chain._
+
+A promotion chain `c` is said to be _strictly bounded by_ `T` iff `[T]++c` is
+also a promotion chain.
```
lrhn, line 164: (Aka, `c.first <<: T`.)

resources/type-system/flow-analysis.md
```diff
@@ -126,9 +124,83 @@ that assignment).
...
+
+#### Joining of promotion chains
+
+The _join of promotion chains_ `c₁` and `c₂`, denoted `c₁.join c₂`, is given by
+the following recursive definition:
+
+```
+([T₁]++c₁).join ([T₂]++c₂) =
```
lrhn: Confusing notation. Using `.join` to suggest it's an operation on the first operand, like a dart method, but then _not_ using parentheses around the second operand, puts it in a notational uncanny valley. The space after `join` really breaks up the reading too.

I'd do either: `join(c₁, c₂)` or `c₁ join c₂`. Or find a nice Unicode character and do `c₁ ⨝ c₂`.
(That character is the "join" from relational algrebra, but I'd go for aestethics over precision.)

resources/type-system/flow-analysis.md
```diff
@@ -126,9 +124,83 @@ that assignment).
...
+    [T₁] ++ c₁.join c₂ if T₁ = T₂
```
lrhn: Associativity is not obvious. I can see the result should be `[T1] ++ (c1.join c2)`, but the parens are needed.

resources/type-system/flow-analysis.md
```diff
@@ -126,9 +124,83 @@ that assignment).
...
+    c₁.join c₂         if T₁ ≠ T₂ ∧   T₁ <: T₂ ∧   T₂ <: T₁
+    ([T₁]++c₁).join c₂ if T₁ ≠ T₂ ∧   T₁ <: T₂ ∧ ¬ T₂ <: T₁
+    c₁.join ([T₂]++c₂) if T₁ ≠ T₂ ∧ ¬ T₁ <: T₂ ∧   T₂ <: T₁
+    []                 if T₁ ≠ T₂ ∧ ¬ T₁ <: T₂ ∧ ¬ T₂ <: T₁
+[].join c₂ = []
+c₁.join [] = []
+```
```
lrhn: Doing math using single-space characters doesn't make the lines distinguishable. I want something signifying each new item. I'd go for:

> The `join` operation on promotions chains is defined as:
> * `join([], c2)` = `[]`
> * `join(c1, [])` = `[]`
> * `join([T1]++c1, T2++c2)` = 
>     * `[T1] ++ join(c1, c2)` if `T1` = `T2`
>     * `join(c1, c2)` if `T1` &ne; `T2` &wedge; `T1` <: `T2` &wedge; `T2` <: `T1`
>     * `join([T1]++c1,c2)` if `T1` &ne; `T2`, &wedge; `T1` <: `T2` &wedge; &not; `T2` <: `T1`
>     * `join(c1,[T2]++c2)` if `T1` &ne; `T2`, &wedge; &not;  `T1` <: `T2` &wedge; `T2` <: `T1`
>     * `join(c1, c2)` if `T1` &ne; `T2`, &wedge; &not;  `T1` <: `T2` &wedge; &not; `T2` <: `T1`

resources/type-system/flow-analysis.md
```diff
@@ -126,9 +124,83 @@ that assignment).
...
+
+By construction, the join of two promotion chains is always a subsequence of
+each of the input chains. That is, `c₁.join c₂ <+ c₁` and `c₁.join c₂ <+
+c₂`. Therefore, if `c₁` and `c₂` are promotion chains, then so is `c₁.join c₂`.
```
lrhn: (Maybe avoid line-break inside ``` ` ```-sections. It might work, but the review tool gets confused. Or it might not work, I don't actually know._)

lrhn: That is: the join is the maximal common subsubsequence. That could be a valid definition by itself, at least if we can argue that it's always well defined and unique, which shouldn't be hard. "Make a sequence from all elements that occur in both chains. That's definitely maximal (there can be no further elements that occur in both), and because `<<:` is asymmetric, the order of those elements _must_ be the same in both sequences, so it's a subsequence of both."
Then the algorithm above is a way to find that sequence, not its definition. ... I think I like it better, to define what things _are_, not just how to constructor something.)

resources/type-system/flow-analysis.md
```diff
@@ -126,9 +124,83 @@ that assignment).
...
+
+_We will use the join operation to combine the promotion chains at the point
+where two separate control flow paths rejoin (e.g., at the end of an `if`
+statement). Informally, the join of two promotion chains is a promotion chain
+that keeps whatever promotions are present in both control flow paths, with the
+exception that if at any point the types in the two chains become unrelated
+(neither `T₁ <: T₂` nor `T₂ <: T₁`), then further promotions are dropped._
```
lrhn, line 189: That doesn't sound right.
Which means the definition doesn't sound right either.

If we have a diamond struture (yep, always that old chestnu!):
D <: (B, C) <: A, with no relation between B and C,
then the promotion chains `[A, B, D]` and `[A, C, D]` should join to `[A, D]`, not stop when it seees `[B, D]` and `[C, D]` with `B` and `C` unrelated.

Sooo, what are we implementing?
We seem to be implementing what you've specified. So we should fix the implementation too.

```dart
sealed class A;
class B implements A {
  bool get imB => true;
}
class C implements A {
  bool get imC => true;
}
class D implements B, C {
  bool get imD => true;
  bool get imB => false;
  bool get imC => false;
}

void foo(A v, bool yes) {
  if (yes) {
    switch (v) {
      case B():  // Promotion chain [A,B]
         v.imB;
         if (v is! D) return;
         // Promotion chain [A, B, D].
         v.imD;
      case C(): // Doesn't get promoted by exahustion if just using `default:`?
         v.imC;
         // Promotion chain [A, C].
         if (v is! D) return;
         // Promotion chain [A, C, D]. 
         v.imD;
    }
    v.imD; // Not valid.
    print(v.me); // D
    // Promotion chain is [A, D]
  } else {
    if (v is! B) return;
    // Promotion chain is [A, B];
    v.imB;
    // if (v is! C) return; // Also works.
  } // Join [A,D] and [A, B]
  // Promotion chain should only be [A] here.
  print(v.me); // A
}

void main() {
  foo(D(), DateTime.now().millisecondsSinceEpoch > 0);
}

extension on A {
  String get me => "A";
}
extension on D {
  String get me => "D";
}
```

resources/type-system/flow-analysis.md
```diff
@@ -126,9 +124,83 @@ that assignment).
...
+The _join of promotion chains_ `c₁` and `c₂`, denoted `c₁.join c₂`, is given by
+the following recursive definition:
+
+```
+([T₁]++c₁).join ([T₂]++c₂) =
+    [T₁] ++ c₁.join c₂ if T₁ = T₂
+    c₁.join c₂         if T₁ ≠ T₂ ∧   T₁ <: T₂ ∧   T₂ <: T₁
+    ([T₁]++c₁).join c₂ if T₁ ≠ T₂ ∧   T₁ <: T₂ ∧ ¬ T₂ <: T₁
+    c₁.join ([T₂]++c₂) if T₁ ≠ T₂ ∧ ¬ T₁ <: T₂ ∧   T₂ <: T₁
+    []                 if T₁ ≠ T₂ ∧ ¬ T₁ <: T₂ ∧ ¬ T₂ <: T₁
```
lrhn: See below. This should be `c1.join c2`, same as the second case, because both prove that neither `T1` nor `T2` can be occur anywhere in the other chain.

That is: for `[T1]++c1` and `[T2]++c2`:
* if `T1`=`T2`, it's in the common subset.
* otherwise
  * if &not; `T1` <: `T2` then `T1` (is not `T2` from above) is not greater than any element of `c2`, because all of those are strictly subtypes of `T2`. So we can drop `T1.
  * if &not; `T2` <: `T1` then we can similarly drop `T2`.
  * but if both `T1` <: `T2` and `T2` <: `T1` (and not equal), then we can also drop both,
     because every element in `c1` is a strict subtype of `T1` and so cannot be equal `T2`,
     and vice versa.

(The join should be symmetric and work from either end, whether starting with the super-est type in each chain or with the subbest.)

Or more directly:
join(`[T1]++c1`,`[T2]++c2`) ::=
* If `T1` == `T2`, `[T1]++join(`c1`, `c2`)`
* else if `T1`<:`T2`, join(`[T1]++c1`, `c2`)  // T1 _might_ be in the intersection.
* else if `T2`<:`T1`, join(`c1`, `[T2]++c2`)  // T2 _might_ be in the intersection.
* else join(`c1`, `c2`)

That will skip mutual subtypes in two steps, first removing `T2` in the second item and then removing `T1` in the third item of the iteration recursion.

resources/type-system/flow-analysis.md
```diff
@@ -126,9 +124,83 @@ that assignment).
...
+
+_We will use the join operation to combine the promotion chains at the point
+where two separate control flow paths rejoin (e.g., at the end of an `if`
+statement). Informally, the join of two promotion chains is a promotion chain
+that keeps whatever promotions are present in both control flow paths, with the
+exception that if at any point the types in the two chains become unrelated
+(neither `T₁ <: T₂` nor `T₂ <: T₁`), then further promotions are dropped._
+
+The `join` relation is idempotent and commutative by construction. It is not
+associative.
```
lrhn, line 195: If it was just "maximal common subsequence" it would be associative too. Doing something less precise is what makes it depend on what the _other_ elements in the chain are, those that aren't used by the result anyway.

resources/type-system/flow-analysis.md
```diff
@@ -126,9 +124,83 @@ that assignment).
...
+
+_For a counterexample to associativity, consider:_
+
+```
+c₁ = [Map<Object?, int>, Map<int, int>]
+c₂ = [Map<dynamic, int>, Map<int, int>]
+c₃ = [Map<int, Object>,  Map<int, int>]
+c₁.join c₂ = [Map<int, int>]
+(c₁.join c₂).join c₃ = [Map<int, int>]
+c₂.join c₃ = []
```
lrhn, line 205: And with the change above, that too would be `[Map<int, int>]`, when it drops both of the leading unrelated elements and continues with the rest.

resources/type-system/flow-analysis.md
```diff
@@ -263,7 +328,7 @@ We also make use of the following auxiliary functions:
   where `r2` is `r` with `true` pushed as the top element of the stack.
 
 - `unsplit(M)`, where `M = FlowModel(r, VM)` is defined as `M1 = FlowModel(r1,
-  VM)` where `r` is of the form `[...s, n1, n0]` and `r1 = [...s, n0&&n1]`. The
+  VM)` where `r` is of the form `s ++ [n1, n0]` and `r1 = s ++ [n0&&n1]`. The
```
lrhn: Spaces around the `&&`? (I don't know what they mean, but it's hard to read.
(I should make a script for changing `[a-z]+(\d)\b` into a subscript of the digit.
The `Sn` is harder :)

resources/type-system/flow-analysis.md
lrhn: "deleting" sounds mutating. More like "omitting". Which is still negative, it says what's not in there, instead of what is.

So maybe instead be direct and positive:

> is a subsequence of `promotionChain` which contains every element `T` where `written <: `T`.

(Which still means it may contain `written`. Can't easily see where it's used, but if we assignment-promote to `written`, we should check that we don't add the type if it's already there.)

resources/type-system/flow-analysis.md
```diff
@@ -444,13 +509,13 @@ Definitions:
...
 
 - `toi_promote(declared, promotionChain, tested, written)`, where `declared` and
   `written` are types satisfying `written <: declared`, `promotionChain` is
-  valid for declared type `declared`, and all types `T` in `promotionChain`
-  satisfy `written <: T`, is the promotion chain `newPromotionChain`, defined as
+  strictly bounded by `declared`, and all types `T` in `promotionChain` satisfy
+  `written <: T`, is the promotion chain `newPromotionChain`, defined as
   follows. _("toi" stands for "type of interest".)_
   - Let `provisionalType` be the last type in `promotionChain`, or `declared` if
     `promotionChain` is empty. _(This is the type of the variable after
```
lrhn, line 526: (Would it be easier, as in "fewer special cases", to always have the declared type in the promotion chain? We don't have to keep it physically in the list in the implementation if we want to save space.)

resources/type-system/flow-analysis.md
```diff
@@ -472,25 +537,24 @@ Definitions:
   - Let `p2` be the set `p1 \ { provisionalType }` _(where `\` denotes set
     difference)_.
```
lrhn, line 543: Unicode has a real symbol for that, `∖`. (I think it's slightly longer then `\` in this particular font.)

resources/type-system/flow-analysis.md
```diff
@@ -444,13 +509,13 @@ Definitions:
...
-    that `demote(promotionChain, written)` is also valid for declared type `T`._
+  - _Note that if `promotionChain` is strictly bounded by `T`, it follows that
+    `demote(promotionChain, written)` is also strictly bounded by `T`._
 
 - `toi_promote(declared, promotionChain, tested, written)`, where `declared` and
   `written` are types satisfying `written <: declared`, `promotionChain` is
-  valid for declared type `declared`, and all types `T` in `promotionChain`
-  satisfy `written <: T`, is the promotion chain `newPromotionChain`, defined as
+  strictly bounded by `declared`, and all types `T` in `promotionChain` satisfy
+  `written <: T`, is the promotion chain `newPromotionChain`, defined as
```
lrhn, line 523: `<:`, not `<<:` ....

resources/type-system/flow-analysis.md
```diff
@@ -472,25 +537,24 @@ Definitions:
   - Let `p2` be the set `p1 \ { provisionalType }` _(where `\` denotes set
     difference)_.
   - If the `written` type is in `p2` then `newPromotionChain` is
-    `[...promotionChain, written]`. _Writing a value whose static type is a
+    `promotionChain ++ [written]`. _Writing a value whose static type is a
```
lrhn, line 545: .... and here we push `written` even though the last element of `promotionChain` _could_ be a mutual subtype of it.
(We only check whether they're the _same_ type in the case above, not whether it's a mutual subtype, and the constraints on `promotionChain` only requires `written <: T`, not `<<:`.)

resources/type-system/flow-analysis.md
```diff
@@ -472,25 +537,24 @@ Definitions:
...
     type of interest promotes to that type._
     - _By precondition, `written <: declared` and `written <: T` for all types
       in `promotionChain`. Therefore, `newPromotionChain` satisfies the
-      definition of a promotion chain, and is valid for declared type
-      `declared`._
+      definition of a promotion chain, and is strictly bounded by `declared`._
```
lrhn, line 549: Except it doesn't because `written <: T` is not sufficient, it needs `written <<: T`.
(Is this what's implemented, and can we make something break?)

resources/type-system/flow-analysis.md
```diff
@@ -472,25 +537,24 @@ Definitions:
...
       provisionalType`.
     - If `p3` contains exactly one type `T` that is a subtype of all the others,
-      then `promoted` is `[...promotionChain, T]`. _Writing a value whose static
+      then `promoted` is `promotionChain ++ [T]`. _Writing a value whose static
       type is a subtype of a type of interest promotes to that type of interest,
       provided there is a single "best" type of interest available to promote
       to._
       - _Since `T <: provisionalType <: declared`, and all types `U` in
         `promotionChain` satisfy `provisionalType <: U`, it follows that all
         types `U` in `promotionChain` satisfy `T <: U`. Therefore
```
lrhn, line 560: Probably can't break this. If the provisional type is a mutual subtype of `written`, then any type `T` in `p2` where `written <: T <: provisionalType` is also a mutual subtype, and since `provisionalType` is _probably_ also a type of interest, there won't be a least.

resources/type-system/flow-analysis.md
lrhn, line 577: (Only place `demote` is used, right?)

lrhn:
I think the promotion chain join can be improved. 
(And there are some places where I don't think we're handling mutual subtypes correctly in assingment promotion, but not sure if it's something that's a real problem.)

resources/type-system/flow-analysis.md
stereotype441: I could possibly get behind `⪿`, though I worry a little that it people will interpret it as an exclusive subsequence (by analogy to `⊂`). The relation I'm trying to define is _inclusive_.

`⋲`, `⋳`, and `∊` worry me because of potential confusion with `∈`, which I would like to reserve for denoting list membership (i.e. `x ∈ l` iff `∃k, l[k] = x`).

`<+` has the advantage of being consistent with the Lean code.

I asked Gemini for suggestions and it offered some more alternatives:
- `⊑` (apparently common in computer science papers, although some people use this to mean "contiguous prefix")
- `≼`
- `⊆ₒ` ("subset ordered")

🤷 I could probably be happy with any of these.

resources/type-system/flow-analysis.md
stereotype441: When I first used Lean's `Preorder` typeclass to define `≤` as subtyping, it gave me `<` for free, defining `T < U` as `T ≤ U ∧ ¬ U ≤ T` (the definition I'm using here), and I thought to myself, "wait, why not `T ≤ U ∧ T ≠ U`?" Then I realized that that definition wouldn't be very useful because it would prevent transitivity. So I thought it might be nice to capture that in an informative comment.

I suppose I could say:

> _It is tempting to instead define `T <<: U` iff `T <: U` and `T ≠ U`, but that definition would be less useful since it wouldn't satisfy transitivity._

Would that be clearer?

But maybe it's better not to lead the reader down the garden path at all. I would also be fine dropping this paragraph entirely.

resources/type-system/flow-analysis.md
stereotype441: Done!

resources/type-system/flow-analysis.md
stereotype441: I was intending for indexing to be zero-based, in which case I believe my definition is correct as written.

resources/type-system/flow-analysis.md
stereotype441: Good point. I've changed this to just "(e.g. via an `is` test)".

resources/type-system/flow-analysis.md
stereotype441: Oh can we _please_ use zero-based indices? Lean uses zero-based indexing (and so does Dart!), and it would hurt my brain _so_ much to try to use zero-based indexing when working on the proofs and the Dart implementation, and one-based indexing when reading and writing the spec.

I can understand your wariness about using `-1` in subscripts. "vₙ₋₁" looks fine in Github's variable-width font but kinda wonky in its typewriter font (`vₙ₋₁` ☹️). But I'll gladly put up with that wonkiness to avoid having to mentally add and subtract one all the time when moving between `flow_analysis.md` and the Lean/Dart code.

Other than the zero-based vs. one-based question, I've updated my PR to follow your suggested wording.

resources/type-system/flow-analysis.md
stereotype441: I'm trying to consistently use zero-based indexing.

> (Is zero a natural number? Discuss!)

Yes. 😃

resources/type-system/flow-analysis.md
stereotype441: I'd prefer not to say `c.first` because (a) it's not defined, and (b) any definition of it would have to explain what to do when `c` is the empty list. I suppose we could say:

> A promotion chain `c` is said to be _strictly bounded by_ `T` iff `c = []` or `c[0] <<: T`.

But at that point "`[T]++c` is also a promotion chain" feels simpler.

resources/type-system/flow-analysis.md
stereotype441: Yeah, fair enough. I've switched to `join(c₁, c₂)` everywhere.

resources/type-system/flow-analysis.md
stereotype441: Fixed.

resources/type-system/flow-analysis.md
stereotype441: Yeah, good point. I've updated the PR to match your recommendation (except that the last case is `[]`, to match the current behavior of the implementation).

resources/type-system/flow-analysis.md
stereotype441: I believe it works fine, but I don't mind changing it. Done.

(Apologies in advance if I screw this up in other places. I'm in the habit of typing `M-q` in EMACS to reflow paragraphs, and that puts line breaks inside <tt>`</tt>-sections.)

resources/type-system/flow-analysis.md
stereotype441: The `&&`s just mean boolean "and" (this section is talking about the "reachability" stacks, which are stacks of booleans).

For now I've added spaces around the `&&`, as you suggested.

I believe there's an equivalent way to specify all this stuff without using the boolean stacks at all, but I need to do some more figuring to prove to myself that's the case. So I'll clean this up further in a future PR if that's ok with you.

resources/type-system/flow-analysis.md
stereotype441: > ...maybe instead be direct and positive:
> 
> > is a subsequence of `promotionChain` which contains every element `T` where `written <: `T`.

I like your suggested wording, thanks!

> (Which still means it may contain `written`. Can't easily see where it's used, but if we assignment-promote to `written`, we should check that we don't add the type if it's already there.)

Yeah, we definitely need to make sure we don't break the promotion chain invariant when subsequently do type-of-interest promotion.

In practice, the implementation _does_ make sure not to break the promotion chain invariant when doing type-of-interest promotion. I haven't reached this part of the spec yet in my revision process, so I'm not 100% sure what the spec says.

I will return to this in a subsequent PR if that's ok with you.

resources/type-system/flow-analysis.md
stereotype441: Fixed, thanks!

resources/type-system/flow-analysis.md
lrhn: I'd drop it. 

I'm not seeing why that definition would be tempting in a system where we have mutual sub-types that are not equal.

The goal is to have an asymmetric and non-reflexive relation, so that a chain contains at most one element per equivalence group. 
Using `=` for anything in that context feels out of place.

resources/type-system/flow-analysis.md
lrhn: Ack, but do add the "0 &le; ".

resources/type-system/flow-analysis.md
lrhn: Go Unicode! Didn't know it had subscript + and -.

resources/type-system/flow-analysis.md
lrhn: Ack. I did that here to avoid making a decision on whether to start at 0 or 1. With that decided, out just means `[0]`.

resources/type-system/flow-analysis.md
munificent: I like using operators for operations that are extremely common and have a well-known familiar operator already. For other operations, I would be inclined to use a readable name.

I think `++` for concatenation is familiar enough to work well (and I like that it disambiguates from addition).

I'm less enthused about using a Unicode operator for "subsequence" which is already a fairly unfamiliar operation. Maybe just `subseq(l₁, l₂)` or `l₁ subseq l₂` if you think the latter helps clarify which argument is a subsequence of the other?

resources/type-system/flow-analysis.md
```diff
@@ -108,16 +108,22 @@ that assignment).
...
+    satisfies that `[].length = 0`, `[v].length = 1`, and `(l₁ ++ l₂).length =
+    l₁.length + l₂.length`._
+  - We use the notation `l[k]` for a single element of a list where `k` is an
+    integer so that if `l = [v₀, ... vₙ₋₁]`, and `0 ≤ k < n`, then `l[k] =
+    vₖ`. _This operation satisfies that `(l₁ ++ l₂)[k]` is `l₁[k]` if `0 ≤ k <
+    l₁.length` and `l₂[k - l₁.length]` if `l₁.length ≤ k < l₁.length +
+    l₂.length`._
+  - We use the notation `l₁ <+ l₂` to denote that `l₁` is a subsequence of
+    `l₂`. That is, `l₁ = [l₂[k₀], l₂[k₁], ... l₂[kₙ₋₁]]` for some `k₀ < k₁ <
+    ... < kₙ₋₁`. _Note that subsequences need not be contiguous._
```
munificent, line 126: The non-contiguous callout is helpful here. I definitely didn't assume that at first.

munificent (approved):

resources/type-system/flow-analysis.md
stereotype441: You're right, that would be much better. It would be a change to the existing functionality, and I want to keep the spec in line with the implementation, so I don't want to change the spec _quite_ yet, but I've filed https://github.com/dart-lang/language/issues/4757 to track the idea, and I hope to get to implementing it (as a language versioned change) soon!

resources/type-system/flow-analysis.md
stereotype441: Acknowledged. See https://github.com/dart-lang/language/issues/4757.

resources/type-system/flow-analysis.md
stereotype441: Acknowleged. See https://github.com/dart-lang/language/issues/4757.

resources/type-system/flow-analysis.md
stereotype441: Acknowledged. See https://github.com/dart-lang/language/issues/4757.

resources/type-system/flow-analysis.md
stereotype441: Acknowleged. See https://github.com/dart-lang/language/issues/4757.

resources/type-system/flow-analysis.md
stereotype441: Yeah, this section is a bit of a mess, and definitely doesn't line up with the current implementation behavior. I'd prefer to leave it as is for now, and come back to it in a later PR so I can think about it as a whole and get it right.

resources/type-system/flow-analysis.md
stereotype441: Acknowledged. I will address this in a follow-up PR.

resources/type-system/flow-analysis.md
stereotype441: Acknowledged. The current implementation doesn't have this mistake. I will address this in a follow-up PR.

resources/type-system/flow-analysis.md
stereotype441: Acknowledged. I will clean this up in a follow-up PR.

resources/type-system/flow-analysis.md
stereotype441: Correct.

resources/type-system/flow-analysis.md
stereotype441: Ok, I'm convinced that using a custom operator for "subsequence" is more clever than necessary. I've changed to `subseq(l₁, l₂)`.

resources/type-system/flow-analysis.md
stereotype441: Makes sense. I've dropped it.

resources/type-system/flow-analysis.md
stereotype441: Done

stereotype441:
> I think the promotion chain join can be improved. (And there are some places where I don't think we're handling mutual subtypes correctly in assingment promotion, but not sure if it's something that's a real problem.)

To summarize my comments above (since Github makes it hard to thread conversations):
- Regarding your proposed improvement to promotion chain join, I love it! I've filed https://github.com/dart-lang/language/issues/4757 to track it. I'll leave this PR as is since it correctly documents the current behavior (though I did add a link). Let me know if https://github.com/dart-lang/language/issues/4757 is missing any crucial information.
- Regarding the mishandling of mutual subtypes correctly in assignment promotion, the text of the spec definitely doesn't match the implementation. Possibly in benign ways, possibly not. I'm not sure. I'd prefer to leave it as is and return to it in a later PR so I can think carefully about it and get it right.

resources/type-system/flow-analysis.md
stereotype441: Currently we don't promote to `X&Y` because, as you say, `X` and `Y` are mutual subtypes. Despite the messy state of the spec, I believe the implementation is solidly consistent in maintaining the invariants that
- Every type in a promotion chain is a strict subtype of the previous.
- The promotion chain for a local variable is always strictly bounded by the local variable's (declared or inferred) unpromoted type. _(Private field promotion is a whole other story that I'll get to documenting in a later PR.)_

I definitely want to keep the first invariant, because without it, `join` gets messy. I'd prefer to keep second invariant too, because it feels nicely consistent with the first. So my vote would be to keep the current behavior of not promoting to `X&Y` in this case.

resources/type-system/flow-analysis.md
stereotype441: For local variables, yes. For private field promotion things get trickier, because the "declared" type can change due to other promotions. Consider:
```dart
class A {
  final Object? _x;
  A(this._x);
}
class B implements A {
  final num _x;
  B(this._x);
}
f(A a) {
  // Declared type of `a._x` is `Object?`; promotion chain is `[]`
  a._x.expectStaticType<Exactly<Object?>>;

  // Promote `a._x`
  a._x as Object;

  // Promotion chain for `a._x` is now `[Object]`
  a._x.expectStaticType<Exactly<Object>>;

  // Promote `a`
  if (a is B) {

    // Declared type of `a._x` is now `num`; promotion chain is still `[Object]`
    // When you read from `a._x`, we notice that the last type in the promotion
    // chain is not a subtype of the declared type, so we ignore the promotion
    // and give you the declared type.
    a._x.expectStaticType<Exactly<num>>;

    // Cancel the promotion of `a` by joining control flow paths
  }

  // Declared type of `a._x` is now `Object?` again. Promotion chain is still
  // `[Object]`.
  a._x.expectStaticType<Exactly<Object>>;
}
```

Because of this, I'm inclined to keep declared type separate from the promotion chain.

lrhn (approved):

PR #4756, opened 2026-08-27
Language feature lifecycle: Add info about template project and breaking changes.

kallentu:
Some extra updates to the lifecycle of a language feature doc.
- Ensure we're using the project template and not aviary.
- Ensure we're writing a CHANGELOG entry with breaking changes introduced by the feature.


munificent (approved):
I love it. The "Include any breaking changes" reminder is going to be a gift from our past selves to our future selves. :D

PR #4765, opened 2026-09-03
Simplify augmentations even more.

### LINKED ISSUE #4565: How do primary constructors and augmentations interact?
We need to figure out all of the various ways that augmentations and primary constructors can play nicely or poorly together:

1. Does an augmentation of a class with a primary constructor have to repeat the primary constructor in its header? If yes, it seems like it's a redundant member declaration, which doesn't make much sense. We don't require you to repeat any other members and in fact explicitly forbid it. If no, then the primary constructor parameters are in lexical scope in the augmentation's class body but their declarations are not textually visible. With type parameters, we specifically require the augmentation to repeat them to avoid that syntactic confusion.

2. I'm assuming in 1 that the primary constructor parameters are in scope inside the body of the augmentation. But that's a choice too. Should they be?

3. Can an augmentation augment a primary constructor with a body by writing a generative constructor declaration with the same name? Conversely, can an augmentation augment a generative constructor with a `;` in the introductory declaration by having a primary constructor with the same name in the augmentation?

4. Can an augmentation of a class with a primary constructor add a `this` block if the introductory declaration doesn't have one? Or if it does?

5. What else am I not thinking of?

I haven't thought about this much, but here's one proposal:

We treat primary constructors *including the scoping of their parameters* as syntactic sugar effectively applied before augmentation. So a primary constructor is treated like an in-body constructor declaration. Any instance field initialized at its declaration that refers to a primary constructor parameter is treated like an explicit constructor initializer. Then augmentations apply to the result of that.

So, if you write:

```dart
class C(final int x, int y) {
  int z = y;
  this : assert(x != y);
}
```

As far as augmentation is concerned, it's as if you wrote:

```dart
class C {
  final int x;
  int z;
  new (this.x, int y) : z = y, assert(x != y);
}
```

And then an augmentation can modify that according to the existing rules.

### LINKED ISSUE #4573: Specify order that instance field initializers run with augmentations
The language says that the initializers for instance fields run in the order their declarations appear:

```dart
int i = 0;

class C {
  int a = i++;
  int b = i++;
}

main() {
  var c = C();
  print('${c.a} ${c.b}');
}
```

This prints `0 1`.

Now consider:

```dart
int i = 0;

class C {}

augment class C {
  int a = i++;
}

augment class C {
  int b = i++;
}
```

Does it behave the same? Consider that these augmentations may be in separate part files.

I think the obvious answer is that field initializers run in augmentation application order. Or, said another way, applying an augmentation *appends* new members to a type. That's consistent with how enum values and mixin `with` clauses are handled.

I don't think there's anything scary here, the proposal just needs to actually specify this.

### LINKED ISSUE #4662: Clarify interactions among primary constructors and augmentation
The [augmentations feature](https://github.com/dart-lang/language/blob/main/working/augmentations/feature-specification.md) and the [primary constructors feature](https://github.com/dart-lang/language/blob/main/accepted/future-releases/primary-constructors/feature-specification.md) have been developed concurrently, and a couple of issues about their interactions have not yet been clarified.

Note that https://github.com/dart-lang/language/issues/4565 contains older discussions about this topic. This issue refers to proposals from there several times, but otherwise starts afresh in order to keep it simple and avoid revisiting arguments made relative to older versions of both features.

This proposal is considerably simpler than the one I wrote [here](https://github.com/dart-lang/language/issues/4565#issuecomment-3575291431). I think this topic calls for simplifications, which is also conceptually aligned with other changes to the augmentations feature that we've decided on recently.

### Proposed decision 1: No `augment` keyword on primary constructors

We could allow a primary constructor to have the keyword `augment`, which would allow it to be an augmenting declaration. For example:

```dart
class A(int i);
augment class augment A(final int i);
```

I'd recommend that we do not introduce support for the `augment` keyword on primary constructors (that is, _after_ keywords like `class` or `enum`). We already decided that we do not want to allow `@metadata` at this location, and I believe we agreed that it is not helpful to interpret metadata on the class as being applicable to a primary constructor as well, if any; similarly, I don't think it's useful to say that an `augment` keyword on a class/mixin class/enum/extension type declaration is applicable to the primary constructor in its header as well, if any.

Many elements in the proposals in #4565 seem to assume that a primary constructor declaration is augmenting even though it doesn't have the keyword `augment` (whereas the enclosing class may or may not have it). Let's call this an _implicitly augmenting_ declaration. I'd recommend that we do not introduce any such thing, for consistency with the rest of the augmentations feature.

This implies that a primary constructor declaration cannot be augmenting, so it has to be introductory. Many proposals in #4565 agree with this choice, but many others disagree.

Similarly, there is no support for marking a declaring parameter of a primary constructor with the modifier `augment`. This could have been used to augment the implicitly induced instance variable declaration (as implied in some parts of #4565). However, there is no need to support this kind of augmentation, the instance variable can be augmented just like other instance variables even though it has been implicitly induced.

### Proposed decision 2: The header and body part go together

We could allow an augmentation chain for a constructor to have a primary constructor header part in one declaration and a body part in a different declaration. For example:

```dart
class B(int i);

augment class B {
  final int j;
  /*augment?*/ this : j = i {
    ...
  }
}
```

We already decided that we do not want to allow access to the primary parameters (like `i`) in a declaration that does not include the primary parameter declarations. This implies that it would be an error for the body part to have an initializer list entry like `j = i` (because `i` cannot be accessed without the parameter list in the header, and we can't write an augmenting declaration in the header, according to decision 1). This implies that the body part can add metadata and documentation, but it can't have a useful initializer list, and it can't have a useful body (at least, it can't access the constructor parameters).

Another difficulty is the keyword `augment`, does the body part have that or not? You could claim that it should only have `augment` if it is augmenting another _body part_ of a primary constructor, but this would then have to be an incomplete body part (only adding metadata or documentation). Also, we'd need to discuss whether or not we could have a primary constructor body part _first_, and then the header part in some later augmenting declaration.

```dart
class C {
  @metadata
  this;
}

augment class C();
```

This certainly makes it harder to determine what it means for a sequence of declarations to be an augmentation chain for the given constructor.

Arguably, it violates the decision 1 because it makes the header part of the primary constructor declaration augmenting (even though it's very unclear what it is augmenting, because the lone `this;` declaration fails to determine most of the relevant properties of the primary constructor, including its signature).

Many proposals in #4565 assume that we can have a header part and a body part of a primary constructor in different declarations, with some constraints. For example, they include declarations of the form `augment this ...` and show examples where one class declaration has a body part and no header part of a primary constructor. This gives rise to a lot of complexity.

I'd recommend that we make it a compile-time error to have a body part unless it occurs in the same declaration as a header part, no exceptions. A primary constructor can have a body part or not, but they can't be brought together from two (or more!) distinct declarations in an augmentation chain.

This also implies that there will not be any `augment this ...` declarations. Hence, the keyword `augment` is simply a syntax error in a body part declaration of a primary constructor.

As mentioned, with this it is guaranteed that every primary constructor declaration is introductory, not augmenting. However, nothing prevents it from occurring in an augmenting class/mixin class/enum/extension type declaration.

### Proposed decision 3: A primary constructor is always complete

It is possible to declare a primary constructor that does not have any of the elements which are otherwise used to determine that the declaration is complete (it's complete if it has any declaring parameters, any initializing formals, any super parameters, any initializer list elements, or a body, or if any non-late instance variable initializers refer to a primary parameter). This kind of declaration could presumably be used to introduce a constructor with a certain signature which could then be made complete in an augmenting declaration. For example:

```dart
class D(int i);

augment class D {
  final int i;
  new(this.i) {
    print("We can even add a body!");
  }
}
```

We're using a regular constructor declaration to augment the primary constructor, as decided [here](https://github.com/dart-lang/language/issues/3177#issuecomment-2007848454).

However, it should be noted that the primary constructor declared in the introductory declaration of `D` does not have any of the special powers that primary constructors otherwise provide: It cannot have a declaring parameter (because that would make it complete, and we're only considering incomplete ones here), and the same goes for initializing formals and super parameters. Also, non-late instance variable initializers cannot refer to any of the parameters of the primary constructor (this would also make it complete).

This implies that the primary constructor may be slightly more concise, but otherwise there is no reason whatsoever why it couldn't just be declared in the body of `D` as an incomplete, normal, in-body constructor.

I'd recommend that we specify that every primary constructor is complete.

### Omitted elements

This implies that certain features discussed in #4565 are not supported (this proposal is among the strictest):

- A primary constructor cannot be augmenting (but it can be augmented by an in-body constructor).
- A primary constructor cannot be "just a signature"; many proposals in #4565 agree on this choice and describe it using phrases like "primary-ness is significant".
- A primary constructor is _always_ complete (even if it "doesn't do anything").
- Initializing expressions of non-late instance variables can refer to primary parameters, _but only_ in the augmentation chain element that declares the primary constructor as well.

The last restriction is probably the most controversial. The discussions in #4565 generally tend to support a much larger amount of locations where those initializing expressions can refer to a primary parameter.

However, I consider this ability to be part of the _implementation_ of the constructor (just like `this.x` is considered to be part of the implementation of any constructor), and the fact that we've decided to avoid combining pieces of an implementation (all of the implementation must be provided by a single declaration) makes me think that this is the consistent choice.

### Proposal

A primary constructor can be declared on a membered, type-introducing declaration (class, mixin class, or enum), be it introductory or augmenting. *This makes the primary constructor proposal a pre-condition for the augmentations proposal. Extension types have their own rules, we will adjust them as needed when the augmentations feature spec is updated.*

A primary constructor is an introductory declaration. A primary constructor is a complete constructor declaration.

A primary constructor can be augmented by an in-body constructor. *It follows that this must be an incomplete declaration, with the same parameter names and types etc. It can add DartDoc comments, metadata, or default values.*

The initializing expression in a non-late instance variable declaration _D_ can refer to parameters of a primary constructor _P_, but only in the case where _D_ and _P_ occur in the same element in the augmentation chain for the type-introducing declaration *(e.g., in the same syntactic `class` declaration)*.

References to primary parameters in another element of the augmentation chain are reported as compile-time errors *(in particular, they do not get resolved in the top-level scope)*.

```dart
class A(int x) {
  final int y = x + 1; // OK
}

String x = "Global";

augment class A {
  final int z = x.length; // Error, `x` is the parameter which isn't accessible.
}
```

### Answers

Here are the answers to the questions posed in the OP of #4565:

> - Does an augmentation of a class with a primary constructor have to repeat the primary constructor in its header? If yes, it seems like it's a redundant member declaration, which doesn't make much sense. We don't require you to repeat any other members and in fact explicitly forbid it. If no, then the primary constructor parameters are in lexical scope in the augmentation's class body but their declarations are not textually visible. With type parameters, we specifically require the augmentation to repeat them to avoid that syntactic confusion.

No, a primary constructor declaration is never augmenting and hence there's no need to discuss how an augmenting primary constructor declaration should be written.

> - I'm assuming in 1 that the primary constructor parameters are in scope inside the body of the augmentation. But that's a choice too. Should they be?

Yes, they are in scope, but it is only allowed to use them if the primary constructor is specified in the current introductory-or-augmenting declaration. "You can access a primary parameter, but only if you can see it."

> - Can an augmentation augment a primary constructor with a body by writing a generative constructor declaration with the same name? Conversely, can an augmentation augment a generative constructor with a ; in the introductory declaration by having a primary constructor with the same name in the augmentation?

No, a primary constructor is always complete, so we can't add a body in an augmentation.

> - Can an augmentation of a class with a primary constructor add a this block if the introductory declaration doesn't have one? Or if it does?

No, the header and body part of a primary constructor are always declared in the same element of the augmentation chain of the enclosing type-introducing declaration. "Implementations are specified in one location."

### LINKED ISSUE #4723: Clarify scoping of primary constructor parameters in augmentations
@scheglov raised a question about:

```dart
class A(int p1);

augment class A(int _) {
  final int value = p1; // <-
}
```

Is the marked line an error because `p1` is not in scope, or OK because it is "inherited" from the main declaration? I believe the answer should be that it's an error. The principle we have with augmentations is:

* In an augmentation, *members* from other declarations are accessible.
* But *lexically scoped* identifiers from other declarations are not.

In this case, since `p1` is a parameter and the augmentation doesn't redeclare it with the name `p1`, the second case applies. If `p1` had been a declaring parameter (like `final int p1`), then the `p1` *member* would be in scope inside the augmentation inside things like instance method bodies, but the `p1` *parameter* would still not be in scope inside other instance field initializers.

@scheglov says the proposal isn't clear on this point, so I'm filing this issue to track pinning that down in the proposal.

cc @lrhn @eernstg

### LINKED ISSUE #4727: Augmenting primary constructors and instance variable initializer scopes.

### LINKED ISSUE #4735: [augmentations] General principle and names of implicitly induced instance variables

### LINKED ISSUE #4737: In augmentations, don't combine scopes before lexical resolution
# Background

The [current proposal for augmentations](https://github.com/dart-lang/language/blob/main/working/augmentations/feature-specification.md#scoping) says:

> The static and instance member namespaces for an augmented type or extension declaration include the declarations of all members in the introductory and augmenting declarations. Identifiers in the bodies of members are resolved against that complete merged namespace. *In other words, augmentations are applied before identifiers inside members are resolved.*

I believe one of the main motivations for this is to allow code in one fragment of, say, a class, to access static members declared in another fragment, e.g.:
```dart
class C {
  static f() {}
}
augment class C {
  static g() {
    f(); // correctly resolves to C.f()
  }
}
```

But it has a consequence for name lookup that I don't like. Quoting from the [augmentations spec](https://github.com/dart-lang/language/blob/main/working/augmentations/feature-specification.md#scoping):
```dart
// Library "main.dart":
part 'other.dart';

const name = 'top level';

class C {
  test() {
    print(name); // Treated as `this.name`, so it refers to the declaration in `other.dart`
  }
}

main() {
  C().test();
}

// Part file "other.dart":
part of 'main.dart';

augment class C {
  String get name => 'member';
}
```

I don't like this because when I'm looking at the file `main.dart`, I can't see `other.dart`, so I look for an enclosing construct that contains a declaration of `name`, I find the top level declaration, and assume that is what is being referred to.

# Proposal

The compilation pipeline consists of these stages:
1. Parsing: converts input text into unresolved fragments.
2. Lexical resolution: resolves bare identifiers to a declaration in an enclosing fragment when possible.
3. Model building: merges fragments, constructs a notion of what types exist in the program and what their members are.
4. Compilation: resolves all bare identifiers that couldn't be resolved in step 2, does flow analysis, code generation, etc.

In step 2, we resolve members that are defined in the same fragment:
```dart
class C {
  static f() {}
  g() {}
  h() {
    f(); // Translated to `C.f()` in step 2
    g(); // Translated to `this.g()` in step 2
  }
}
```

In step 4, we resolve members that are defined in other fragments, as well as references to inherited members via implicit `this`:
```dart
class B {
  inherited() {}
}
class C extends B {
  h() {
    inherited(); // Translated to `this.inherited()` in step 4
    f(); // Translated to `C.f()` in step 4
    g(); // Translated to `this.f()` in step 4
  }
}
augment class C {
  static f() {}
  g() {}
}
```

Note that the breakdown of compilation into these 4 steps is already reflected in the implementation; this part is not new. The only new thing I'm proposing (compared to Dart 3.13) is to change how step 4 resolves a bare identifier <tt>_n_</tt> that couldn't be resolved in step 2. Previously the rule was: "interpret the identifier as <tt>this._n_</tt>." The new rule is: "if the current (merged) class contains a static declaration named _n_, then statically resolve the identifier to that declaration. Otherwise, interpret the identifier as <tt>this._n_</tt>."

This change is fully backward compatible, because in the absence of augmentations, references to static declarations are always resolved during step 2.

# Consequences

The example above from the augmentations spec now behaves differently:
```dart
// Library "main.dart":
part 'other.dart';

const name = 'top level';

class C {
  test() {
    print(name); // Resolved to top level `name` in step 2, so the program prints `top level`.
  }
}

main() {
  C().test();
}

// Part file "other.dart":
part of 'main.dart';

augment class C {
  String get name => 'member';
}
```

Previously, we were considering adopting a special rule to prevent accidentally referring to a primary constructor parameter that's declared in another fragment. Now we get the desired behavior without needing any special rules:
```dart
class C(int x) {}
augment class C(int _) {
  final int y = x; // ERROR
}
```

munificent:
- Don't allow augmentations to augment primary constructors or declare them.

- Primary constructor parameters are in scope in all fragments.

There is some loss of expressiveness, but not that much and I think the simplicity in return is worth it.

Fix #4565.
Fix #4573.
Fix #4662.
Fix #4723.
Fix #4727.
Fix #4735.
Fix #4737.

### DIFF
--- a/working/augmentations/feature-specification.md
+++ b/working/augmentations/feature-specification.md
@@ -554,8 +554,8 @@ augmenting) is *incomplete* if all of:
 
 *   There is no redirection, initializer list, initializing formals, or super
     parameters. *Obviously, this only applies to constructor declarations. We
-    don't consider field parameters here because primary constructors can't be
-    augmented anyway.*
+    don't consider declaring parameters here because those only occur in primary
+    constructors which can't be augmented.*
 
 If a declaration is not *incomplete* then it is *complete*.
 
@@ -993,7 +993,7 @@ For purposes of augmentation, a variable declaration is treated as implicitly
 defining a getter whose return type is the type of the variable. If the variable
 is not `final`, or is `late` without an initializer, then the variable
 declaration also implicitly defines a setter with a parameter named `_` whose
-type is the type of the variable. Likewise, a field parameter in a primary
+type is the type of the variable. Likewise, a declaring parameter in a primary
 constructor induces an introductory instance variable declaration which in turn
 has a complete getter and a complete setter if not `final`.
 
@@ -1004,7 +1004,8 @@ it, so these members have bodies.*
 
 A getter can be augmented by another getter, and likewise a setter can be
 augmented by a setter. This is true whether the getter or setter is explicitly
-declared or implicitly declared using a variable declaration or field parameter.
+declared or implicitly declared using a variable declaration or declaring
+parameter.
 
 *Since non-abstract variables are complete, that implies that it is an error to
 augment a non-abstract variable declaration with a complete getter, setter, or
@@ -1106,7 +1107,7 @@ It's a **compile-time error** if:
     default values.*
 
 *   The augmenting declaration and augmented declaration do not have the
-    same `const` and `factor` modifiers. *Augmentations can't change whether a
+    same `const` and `factory` modifiers. *Augmentations can't change whether a
     constructor is const or not, or whether it is generative or not. Note that
     the `new` keyword is omitted here. Augmentations do not have to agree on
     whether they use the old new constructor syntax:*
--- a/working/augmentations/feature-specification.md
+++ b/working/augmentations/feature-specification.md
@@ -2,7 +2,7 @@
 
 Authors: rnystrom@google.com, jakemac@google.com, lrn@google.com, eernst@google.com
 
-Version: 1.45 (see [Changelog](#Changelog) at end)
+Version: 1.46 (see [Changelog](#Changelog) at end)
 
 Experiment flag: augmentations
 
@@ -249,6 +249,15 @@ Likewise, the grammar for an augmenting `mixin` declaration does not allow
 specifying an `on` clause. Only the introductory declaration permits that. We
 could relax this restriction if compelling use cases arise.
 
+It's a **compile-time error** if a class marked `augment` has a primary
+constructor (`primaryConstructor`) or contains a primary constructor initializer
+block (`primaryConstructorBodySignature`). *Only the introductory declaration of
+a class may declare a primary constructor or its initializer block. Allowing
+primary constructors in augmentations raises tricky questions around how field
+parameters behave and what it means to augment a primary constructor. We could
+allow this in a future version, but for now at least, we avoid that complexity
+by limiting primary constructors to the introductory declaration.*
+
 ### Enums
 
 For enum declarations, in addition to the `augment` modifier, we allow declaring
@@ -291,36 +300,6 @@ Just make two separate unnamed extensions.*
 Also note that an augmentation of an extension can't specify an `on` clause. It
 always uses the same `on` clause as the introductory declaration.
 
-## Primary constructors
-
-A `class`, `enum` or `extension type` declarationscan use the primary
-constructor syntax for declaring an initializing _(non-redirecting
-generative)_ constructor.
-
-See [Generative constructors](#generative-constructor-declarations).
-
-### Instance variable initializer expressions
-
-If a class or enum has a primary constructor, then the current scope of
-the initializer expression of a non-`late` instance variable is the
-primary initializer scope, rather than the body scope of the
-surrounding class or enum declaration.
-
-It's a compile-time error if a non-`late` instance variable initializer
-expressions refers to a variable introduced by the constructor's
-initializer list scope, and the surrounding class or enum declaration does
-_not_ have a primary constructor declaration which declares the
-corresponding parameter's name.
-_A primary constructor must declare all parameters, but it can omit declaring
-a positional parameter's name by using `_` instead of the name.
-If it does so, that parameter's name cannot be used by instance variable
-initializers of that class or enum declaration._
-
-_The scope exists and is used whether this particular declaration writes
-a primary constructor or not, but for readability reasons, code may only
-refer to variables of that scope, if there appears to be a declaration of
-the variable's name in the surrounding code._
-
 ## Static semantics
 
 ### Augmentation context
@@ -345,12 +324,6 @@ variable declarations, because those aren't subject to augmentation.*
 
 ### Scoping
 
-The static and instance member namespaces for an augmented type or extension
-declaration include the declarations of all members in the introductory and
-augmenting declarations. Identifiers in the bodies of members are resolved
-against that complete merged namespace. *In other words, augmentations are
-applied before identifiers inside members are resolved.*
-
 It is already a **compile-time error** for multiple declarations to have the
 same name in the same scope. This error is checked *after* part files and
 augmentations have been applied. *In other words, it's an error to declare the
@@ -358,6 +331,14 @@ same top-level name in a library and a part, the same top-level name in two
 parts, the same static or instance name inside an introductory declaration and
 an augmentation on that declaration, or the same static or instance name inside
 two augmentations of the same declaration.*
+
+#### Member namespaces
+
+The static and instance member namespaces for an augmented type or extension
+declaration include the declarations of all members in the introductory and
+augmenting declarations. Identifiers in the bodies of members are resolved
+against that complete merged namespace. *In other words, augmentations are
+applied before identifiers inside members are resolved.*
 
 *For example:*
 
@@ -441,6 +422,43 @@ from all of the augmentations and parts, then through the import scopes which
 may be different for each part file, and finally to the import scope of the main
 library.
 
+#### Primary constructor parameters
+
+When a class or enum has a primary constructor, the current scope of non-`late`
+instance variable initializers is the primary initializer scope. This is true
+even for instance variables declared in other augmentations. *In other words,
+again, augmentations are merged before references to identifiers are resolved.*
+*For example:*
+
+```dart
+// main.dart:
+part 'other.dart';
+
+class C(int a);
+
+// other.dart
+part of 'main.dart';
+
+import 'third.dart' show a;
+
+augment class C {
+  var c = a; // Refers to constructor parameter, not imported `a`.
+}
+```
+
+If a primary constructor parameter is a private named parameter, then references
+to it in instance field initializers in augmentations use the private name, just
+as they would in the introductory declaration. *For example:*
+
+```dart
+class C({final int _a}) {
+}
+
+augment class C {
+  var b = _a; // Refers to constructor parameter.
+}
+```
+
 ### Type annotation inheritance
 
 An augmenting declaration may omit type annotations for a return type, variable
@@ -486,37 +504,39 @@ An augmentation declaration *D* is a declaration marked with the built-in
 identifier `augment`. We add `augment` as a built-in identifier as a language
 versioned change, to avoid breaking pre-feature code.
 
+An augmentation declaration does not introduce a new name into the surrounding
+scope. *We could say that it attaches itself to the existing name.*
+
 *D* augments a declaration *I* with the same name and in the same augmentation
 context as *D*. There may be multiple augmentations in the augmentation context
 of *D*. More precisely, *I* is the declaration before *D* and after every other
 declaration before *D*. *(See "Application order" for the definition of *before*
-and *after*.)*
+and *after*.)* We say that *I* is the declaration which is *augmented by* *D*.
 
-It's a **compile-time error** if there is no matching declaration *I*. *In other
-words, it's an error to have a declaration marked `augment` with no declaration
-to apply it to.*
+It's a **compile-time error** if:
 
-We say that *I* is the declaration which is *augmented by* *D*.
+*   There is an augmentation *D* and there is no declaration with the same name
+    in the same augmentation context that is before *D*.
 
-*In other words, take all of the declarations with the same name in some
-augmentation context, order them according to the "after" relation, and each
-augments the result of all the prior augmentations applied to the original
-declaration. The first one must not be marked `augment` and all the subsequent
-ones must be.*
+*   There is an introductory declaration *I* and an augmentation *D* with the
+    same name in the same context and *D* is before *I*.
 
-An augmentation declaration does not introduce a new name into the surrounding
-scope. *We could say that it attaches itself to the existing name.*
+    *In other words, take all of the declarations with the same name in some
+    augmentation context, order them according to the "after" relation, and each
+    augments the result of all the prior augmentations applied to the original
+    declaration. The first one must not be marked `augment` and all the
+    subsequent ones must be.*
 
-It's a **compile-time error** if an augmentation doesn't have the same kind as
-the introductory declaration. For example, augmenting a `class` with a `mixin`,
-an `enum` with a function, a method with a getter, etc.
+*   An augmentation doesn't have the same kind as the introductory declaration.
+    For example, augmenting a `class` with a `mixin`, an `enum` with a function,
+    a method with a getter, etc.
 
-The exception is that a variable declaration (introductory or augmenting) is
-treated as a getter declaration (and a setter declaration if non-final) for
-purposes of augmentation. These implicit declarations can augment and be
-augmented by other explicit getter and setter declarations. *In other words,
-variables are never augmented or augmenting, only the getters and possibly
-setters that they induce are.*
+    The exception is that a variable declaration (introductory or augmenting) is
+    treated as a getter declaration (and a setter declaration if non-`final`)
+    for purposes of augmentation. These implicit declarations can augment and be
+    augmented by other explicit getter and setter declarations. *In other words,
+    variables are never augmented or augmenting, only the getters and possibly
+    setters that they induce are.*
 
 ### Complete and incomplete declarations
 
@@ -532,9 +552,10 @@ augmenting) is *incomplete* if all of:
 *   The function is not marked `external`. *An `external` function is considered
     to have a body, just not one that is visible as Dart code.*
 
-*   There is no redirection, initializer list, initializing formals, field
-    parameters, or super parameters. *Obviously, this only applies to
-    constructor declarations.*
+*   There is no redirection, initializer list, initializing formals, or super
+    parameters. *Obviously, this only applies to constructor declarations. We
+    don't consider field parameters here because primary constructors can't be
+    augmented anyway.*
 
 If a declaration is not *incomplete* then it is *complete*.
 
@@ -542,8 +563,8 @@ It's a **compile-time error** if an augmentation is complete and any declaration
 before it in the augmentation chain is also complete. *In other words, once a
 declaration has acquired a body, no augmentation can replace it with another.*
 
-*It is allowed to augment a complete declaration as long as the augmentation
-itself is incomplete. This can be useful for an augmentation to add metadata.*
+*A complete declaration can be augmented as long as the augmentation itself is
+incomplete. This can be useful for an augmentation to add metadata.*
 
 *Examples:*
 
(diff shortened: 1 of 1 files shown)


working/augmentations/feature-specification.md
```diff
@@ -345,12 +324,6 @@ variable declarations, because those aren't subject to augmentation.*
 
 ### Scoping
 
-The static and instance member namespaces for an augmented type or extension
-declaration include the declarations of all members in the introductory and
-augmenting declarations. Identifiers in the bodies of members are resolved
-against that complete merged namespace. *In other words, augmentations are
-applied before identifiers inside members are resolved.*
```
eernstg, line 352: Note to reader: This paragraph has been moved, not deleted. Which is fine!

working/augmentations/feature-specification.md
```diff
@@ -597,61 +618,30 @@ files.
...
 Some terminology:
 
-*   The *source code* of a declaration is the span of characters from the
-    first non-whitespace character of the declaration to the last non-whitespace
-    character of the declaration, as matched by the relevant grammar production.
-    *Metadata is not considered part of the declaration, it precedes the
-    declaration in the source.*
+*   The *source location* of a declaration is the offset of the first
+    non-whitespace character of the declaration, as matched by the relevant
+    grammar production.
```
eernstg, line 623: I think we had a discussion about the inability to order the declaration of `i` and `j` when we're looking at `int i, j;`.

Arguably, the new text is no worse than the old one. ;-)

However, I think the safe bet would be to say that it is the offset of the declared name. Declarations that have no name would use the offset of the first non-whitespace character of the declaration (they can't declare multiple things anyway).

working/augmentations/feature-specification.md
```diff
@@ -597,61 +618,30 @@ files.
...
-        declaration, then *A* and *B* must be a primary constructor declaration
-        and its surrounding type-introducing declaration.
-        In that case, the type introducing declaration is before the
-        primary constructor declaration.
-        *(The same effect can be achieved by making the position of a
-        primary constructor be at the end of the class name identifier
-        instead of at the start.)*
+    *   If *A*'s *source location* is before *B*'s, then *A* is before *B*.
+
+    *   Else *B* is before *A*.
```
eernstg, line 644: I think this will suffice (that is, the old text can indeed be removed) if we use the "beginning of name" rule.

A constructor declaration like `new();` would be covered by the case "has no declared name".

(Yes, I'm getting used to it, slowly. ;-)

working/augmentations/feature-specification.md
```diff
@@ -987,7 +993,9 @@ For purposes of augmentation, a variable declaration is treated as implicitly
 defining a getter whose return type is the type of the variable. If the variable
 is not `final`, or is `late` without an initializer, then the variable
 declaration also implicitly defines a setter with a parameter named `_` whose
-type is the type of the variable.
+type is the type of the variable. Likewise, a field parameter in a primary
```
eernstg:
```suggestion
type is the type of the variable. Likewise, a declaring parameter in a primary
```

working/augmentations/feature-specification.md
```diff
@@ -996,7 +1004,7 @@ it, so these members have bodies.*
 
 A getter can be augmented by another getter, and likewise a setter can be
 augmented by a setter. This is true whether the getter or setter is explicitly
-declared or implicitly declared using a variable declaration.
+declared or implicitly declared using a variable declaration or field parameter.
```
eernstg:
```suggestion
declared or implicitly declared using a variable declaration or a
declaring parameter.
```

working/augmentations/feature-specification.md
```diff
@@ -1059,306 +1067,88 @@ Augmenting a constructor works similarly to augmenting a function, with some
...
-constructor execution continues with executing as in Dart before this feature,
-starting with the variable initialization of the parameter list, by initializing
-formals and declaring parameters.
-
-_This is how primary constructors already work. The only difference is that
-because a single constructor can be introduced by more than one declaration,
-the complete declaration, the actual implementation, of a primary constructor
-might not be a primary constructor declaration._
+*   The augmenting declaration and augmented declaration do not have the
+    same `const` and `factor` modifiers. *Augmentations can't change whether a
```
eernstg:
```suggestion
    same `const` and `factory` modifiers. *Augmentations can't change whether a
```

working/augmentations/feature-specification.md
```diff
@@ -532,18 +552,19 @@ augmenting) is *incomplete* if all of:
 *   The function is not marked `external`. *An `external` function is considered
     to have a body, just not one that is visible as Dart code.*
 
-*   There is no redirection, initializer list, initializing formals, field
-    parameters, or super parameters. *Obviously, this only applies to
-    constructor declarations.*
+*   There is no redirection, initializer list, initializing formals, or super
+    parameters. *Obviously, this only applies to constructor declarations. We
+    don't consider field parameters here because primary constructors can't be
```
eernstg: We do have the non-terminal `<fieldFormalParameter>`, but they are usually described as 'initializing formals', and they are not guaranteed to only occur in a primary constructor declaration (also, initializing formals have already been mentioned). So I think 'field parameters' in this context would be intended to mean 'declaring parameters':
```suggestion
    don't consider declaring parameters here because primary constructors can't be
```

eernstg (approved):
This is a very nice simplification!

A couple of rather long text blocks have been removed (e.g., line 1236-1354), and it appears to contain rules about primary constructors which are not valid any more, in addition to rules about other things which are already specified elsewhere. There might be a need to keep some bits and pieces of this text, but I haven't spotted anything specific.

working/augmentations/feature-specification.md
munificent: > I think we had a discussion about the inability to order the declaration of `i` and `j` when we're looking at `int i, j;`.

Do we need to be able to order them? All that matters is the application order of the augmentations of `i` and the application order of the augmentations of `j`. As far as I know, there is no need to have an ordering between unrelated declarations.

working/augmentations/feature-specification.md
munificent: Applied the fix myself with other changes.

working/augmentations/feature-specification.md
munificent: Applied locally.

working/augmentations/feature-specification.md
munificent: I believe we don't even need the "beginning of name" rule (which as you note gets annoying with unnamed constructors). Is there a case where what I have isn't sufficient? I'm trying to keep the ordering algorithm simple because I feel like we spent a lot of specification text on what is a pretty simple concept: textual order within a file, pre-order across the part tree with part directives traversed after the declarations in the file.

munificent:
Thanks for the feedback! :D

working/augmentations/feature-specification.md
eernstg: I think that's true. You could say that it's a sanity check on the ordering that it will yield a total order on all declarations, but it should also be correct if it's a partial order and we're only comparing declarations which are totally ordered.

Closing!

working/augmentations/feature-specification.md
eernstg: I do think it's sufficient. We may or may not wish to say something about it.

eernstg (approved):
Still LGTM

munificent:
I'll wait until @lrhn takes a look too in case he has any feedback.

munificent:
OK, it's been long enough, I'm gonna just merge this. I am, of course, happy to iterate on it more afterwards if @lrhn or anyone else has feedback. :)

